PR #101, opened 2013-02-18
Support a new option for returning raw predictions via socket

claytonhoefer:
We have a use case where we want to know the "confidence level" of a prediction, which means we need to look at the raw predictions output of vw.  However, only the final prediction is sent back to the socket connection when running vw in daemon mode.

This pull request adds support for a new command line option, --return_raw, which makes vw return the raw predictions, in place of the final ones, to the socket.  You can also still specify output files for both prediction types.

We thought this might be useful to others as well.

Edit: as a further point of reference, this discussion is very similar to our use case: http://fastml.com/predicting-closed-questions-on-stack-overflow/

### DIFF
--- a/vowpalwabbit/csoaa.cc
+++ b/vowpalwabbit/csoaa.cc
@@ -261,7 +261,7 @@ namespace CSOAA {
     for (int* sink = all.final_prediction_sink.begin; sink != all.final_prediction_sink.end; sink++)
       all.print((int)*sink, ec->final_prediction, 0, ec->tag);
 
-    if (all.raw_prediction > 0) {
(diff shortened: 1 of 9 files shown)


ehtycuact:
Hello,
I came around this topic when trying to see if I could get all predictions from a --oaa model. My idea is to use the probabilities as confidence measure and then multiply it by a payout generated from another function to get the final "rank" for my class. However, when running the model as a daemon I don't get neither all classes or the raw probabilities. I know its a topic discussed before but is there any workaround that I can do to get all the raw probabilities without having to use a file, as my application requires async begaviour?

Cheers

PR #458, opened 2014-11-27
A new algorithm: FTRL http://www.eecs.tufts.edu/~dsculley/papers/ad-click-prediction.pdf

grafke:
A new algorithm: FTRL http://www.eecs.tufts.edu/~dsculley/papers/ad-click-prediction.pdf

### DIFF
--- a/vowpalwabbit/global_data.h
+++ b/vowpalwabbit/global_data.h
@@ -215,6 +215,8 @@ struct vw {
 
   float l1_lambda; //the level of l_1 regularization to impose.
   float l2_lambda; //the level of l_2 regularization to impose.
+  float ftrl_alpha;
+  float ftrl_beta;
   float power_t;//the power on learning rate decay.
   int reg_mode;
 


leakyMirror:
I tried above command and get this:

```
Enabling FTRL-Proximal based optimization
ftrl_alpha = 0
ftrl_beta = 0.1
Num weight bits = 23
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = test.vw
num sources = 1
average    since         example     example  current  current  current
loss       last          counter      weight    label  predict features
0.693147   0.693147          1      1.0     1.0000   0.0000        5
NAN prediction in example 2, forcing 0.0
0.693147   0.693147          2      2.0     1.0000   0.0000        5
NAN prediction in example 3, forcing 0.0
0.693147   0.693147          3      3.0     1.0000   0.0000        5
NAN prediction in example 4, forcing 0.0
0.693147   0.693147          4      4.0     1.0000   0.0000        5

finished run
number of examples = 4
weighted example sum = 4
weighted label sum = 4
average loss = 0.693147
best constant = 1
best constant's loss = 0.313262
total feature number = 20

```
I am using current version from master branch

grafke:
Hello, Justas,

Could you attach the dataset used for training?

leakyMirror:
I used the same data as trufanov-nok in his previous post:

```
1 |d a b c
-1 |d a b d
1 |d a c
```

JohnLangford:
Confirmed.    In fact, even this:
vw test_f --ftrl

shows the problem.

JohnLangford:
The problem was that ftrl_alpha defaulted to 0.  I changed it to default to 1.  If there is a better default, please update.

grafke:
For most cases alpha values of 0.01 - 0.1 should fit.

JohnLangford:
I set it to 0.005, because that is what optimizes performance on the 
string version of RCV1 CCAT-or-not.

When optimized, the performance appears comparable to the default update 
rule for VW.  I am curious what a more thorough comparison would show.

-John

snowzjy:
I don't know the features interaction is true in ftrl? Is it compatible with --stage_poly and --l1 and --l2 ?

JohnLangford:
Yes for -q --cubic, etc...  You can try with --stage_poly.  There is 
explicit l1/l2 support in the ftrl code.

-John

PR #465, opened 2014-12-04
Java

jmorra:
This is the JNI work that we had discussed earlier.  I modified the JNI C++ layer a little bit to make the packages as shallow as possible.  I used Maven for the Java build engine.  In order to create a Jar:
1.  Make sure JAVA_HOME is set to the Java home directory.  This is critical as resources under that directory are necessary for compiling the JNI layer.  On my Mac the directory is /Library/Java/JavaVirtualMachines/jdk1.7.0_51.jdk/Contents/Home.  This directory must have include underneath it and a platform specific directory with include under that.
2.  Install Apache Maven
3.  Navigate to the java directory and type mvn package.  In some of my tests a few of the Java tests were failing.  In this case, try running mvn package again and see if they pass the second time.
4.  After this is done there should be a directory under java called target.  This directory should have a file called vw-jni-${os.version}-${os.arch}-1.0-SNAPSHOT.jar.  This is a platform specific jar for your OS.  This Jar contains the native VW libraries compiled for your system and will only be usable for people on the same platform.

I have tested this only with the simple example from the house so far, but I think it should work in general.  We can certainly write a bunch more tests in Java if so desired.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ compiler:
 before_install:
   - sudo apt-get update -qq
   - sudo apt-get install -qq libboost-all-dev
+  - sudo apt-get install maven
 install: 
   - make
+  - cd java; mvn test
 script: make test


fulmicoton:
Is there a chance to see that land in Maven Central Repository?

PR #479, opened 2014-12-23
add .PHONY targets as needed

sam-s:
these are simple non-destructive cosmetic changes


JohnLangford:
I merged this one by hand.   Several of the line-ending changes were lost due to conflict with the giant hackfest that I did over the holidays.  Feel free to resubmit.

PR #483, opened 2015-01-02
Fixing Java build

jmorra:
I have just fixed the Java build.  The issue was that I was storing a model for testing purposes.  It would seem that model files are not compatible across VW versions.  When the VW model was trying to read the old file it was hanging.  I changed my logic at test time and now create a new VW model for each test.  Also, I updated the Maven pom file.  There was an issue before whereby the resultant jar sometimes would not contain the native VW library.  This has been fixed.  Additionally I made the main Makefile detect the number of cores on the machine and use that for building.  This has made my builds a lot faster.

Let me know what you think and if you want me to modify anything.


JohnLangford:
You're quite on the ball, but something remains unhappy.

https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/45729140

Can you look into it?

-John

jmorra:
That is really weird, it looks like the results are now slightly
different.  I'll lesson the precision of the tests.

JohnLangford:
Variation in low order bits seems to be fairly common across platforms.  If
nothing else, a different library with a different order of operations
internally can do it.

-John

jmorra:
OK, I think it all looks good now.

JohnLangford:
Thanks, merged in.

PR #487, opened 2015-01-08
lrq arg bugfix

pmineiro:
bugfix: lrq should not always use dropout for testing.


JohnLangford:
My bad, obviously.  Thanks for fixing.

PR #488, opened 2015-01-12
remove classhing redefinition of fmax and fmin

nicknussbaum:
remove clashing redefinition of fmax and fmin. This causes a build error

change configuration of explore to x64 to avoid incompatible project errors


JohnLangford:
Odd, I could have swore I fixed the fmax/fmin issue in my pass...  Anyways, thanks.

PR #490, opened 2015-01-15
Nn pure reduction

pmineiro:
As discussed, nn not accessing weight vector directly.


JohnLangford:
This fails on test 20.  Can you inspect and fix?

pmineiro:
Changed random seed on test, works now.

pmineiro:
Test 38 is a real bug, which I'm fixing, and then I want to re-run the demos (which will take a while).

pmineiro:
Ok, demos seems stable.

PR #494, opened 2015-01-21
speedups to python interface

hal3:


hal3:
added a few more minor changes

PR #501, opened 2015-01-22
New demo using conversions logs from Criteo

ochapelle:


JohnLangford:
Thanks, incorporated.

In general, I agree that we need a much better global testing 
discipline/infrastructure.  That's definitely something that I want to 
work on.

-John

PR #504, opened 2015-01-29
mnist reduction experiments modifications

pmineiro:
*Not* an actual pull request, more an FYI.

I had to make some modifications to the code to get (-oaa 10 --binary) to correspond to "reduction to classification with ties broken randomly".

### DIFF
--- a/vowpalwabbit/binary.cc
+++ b/vowpalwabbit/binary.cc
@@ -12,6 +12,9 @@ void predict_or_learn(char&, LEARNER::base_learner& base, example& ec) {
     ec.pred.scalar = 1;
   else
     ec.pred.scalar = -1;
+
+  // to force OAA to use the discrete label
+  ec.partial_prediction = ec.pred.scalar;
   
   if (ec.l.simple.label != FLT_MAX)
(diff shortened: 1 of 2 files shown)


arielf:
Hi Paul,

Do we need to update the reference stderr of --oaa test 10?
See:  https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/48822140

pmineiro:
This is not a branch that is meant to be pulled into master _ever_.

We were writing a paper and I needed to make some modifications in order to
generate the graph for an experiment.  Part of the modifications was making
OAA a bit dumber (an error transform) to illustrate a point about
reductions.

p

arielf:
Thanks. Makes sense.

So what shoud John do?
- Open a separate mnist branch and pull into it, (not to master as this pull is) and close?
- Just close (all that's needed to reproduce is already on your branch)?

pmineiro:
Yep just close methinks, it'll live forever on github as is (funny enough I
had to consult a "paper branch" the other day to field a question about an
older paper so it can be useful).

p

PR #506, opened 2015-02-02
Enable TCP KeepAlive

tdeconin:
In our test and production environments we see server slots in use by clients that have long disappeared (by normal program termination). Somehow the TCP connections are not properly closed and the TCP connection on the server remains in an ESTABLISHED state. 

If TCP keepalive is enabled these stale connection slots will be reused, preventing the need for a vw restart.


JohnLangford:
I'm unclear on why things remain in an ESTABLISHED state, but keepalive appears harmless.

JohnLangford:
(merged in)

PR #509, opened 2015-02-05
Mac OSX and clang++

bscottm:
Make configure enforce C++11, needed in cbify.cc: std::to_string, std::unique_ptr. Convince clang++ that C++11 features are OK to use (-stdlib=libc++) on Mac OSX.


JohnLangford:
Why does the destructor need to be declared and virtual in the exploration
library?   I'm uncomfortable with forcing people to write noop functions.

I don't fully follow all the automake stuff, but I'm willing to experiment
with it unless I hear otherwise.

(Louie, maybe you can look at the exploration library relevant aspects of
this pull request?)

-John

lhoang29:
I'm also not clear why the destructors need to be defined and marked virtual? (I think the existing destructor for EpsilonGreedy was just dead code). Virtual destructors are normally needed only on base classes.

JohnLangford:
Merged, then I'll remove the virtual destructors by hand.

bscottm:
John:

The clang++/libc++ version of std::unique_ptr uses new and delete to maintain pointer unique ownership. Hygienically, that implies a virtual destructor, even though the destructor does nothing and the class isn't subclasses. It's just a good practice (and code should compile without warnings.)

automake: the code used deprecated variables, these are just updates. I also meant to find which autoconf macro sets _FP_OFFSET_BITS; that shouldn't get set explicitly unless the platform needs it (might look at that tonight.)

What I'm doing with VW on my Mac: I've been able to find a good combination of features related to canine heat stress, based on research vet experimental data.  Important problem in ONR's Navy Expeditionary Canine Sciences Program to improve the military working dog's quality of life. In real life, I lead a CS research department for a think tank.


-scooter

bscottm:
It's a good practice for dynamically allocated classes.

lhoang29:
Just so I understand better, can you explain why this is good practice? I don't see any benefits to doing this since there are no derived types. Is it to ensure proper destruction in case of future inheritance?

bscottm:
I'll have to look up the actual number from the "Effective C++" series, but, basically, yes, the virtual dtor ensures proper destruction in case of unknown/unseen subclasses. Sure, **you** don't have any derived classes, but the **clang++** compiler doesn't know that.

Consequently, clang++ issues warnings. Code should compile without warnings. Newbies don't necessarily grok that the warning is harmless, so another benefit is fewer bug reports misattributed to a compiler warning.

bscottm:
Ok, so here's the output from the merged commit, because you need to see
the warnings to believe that it might just be a good idea to have the
virtual destructors and eliminate newbie misattribution:

cbify.cc:329:64: warning: delete called on 'vw_policy' that has virtual
functions but non-virtual destructor
      [-Wdelete-non-virtual-dtor]
template<class T> inline void delete_it(T* p) { if (p != NULL) delete p; }
                                                               ^
cbify.cc:335:5: note: in instantiation of function template specialization
'delete_it<vw_policy>' requested here
    delete_it(data.policy);
    ^
cbify.cc:329:64: warning: delete called on
'MultiWorldTesting::TauFirstExplorer<vw_context>' that has virtual
functions but
      non-virtual destructor [-Wdelete-non-virtual-dtor]
template<class T> inline void delete_it(T* p) { if (p != NULL) delete p; }
                                                               ^
cbify.cc:336:5: note: in instantiation of function template specialization
      'delete_it<MultiWorldTesting::TauFirstExplorer<vw_context> >'
requested here
    delete_it(data.tau_explorer);
    ^
cbify.cc:329:64: warning: delete called on
'MultiWorldTesting::EpsilonGreedyExplorer<vw_context>' that has virtual
functions but
      non-virtual destructor [-Wdelete-non-virtual-dtor]
template<class T> inline void delete_it(T* p) { if (p != NULL) delete p; }
                                                               ^
cbify.cc:337:5: note: in instantiation of function template specialization
      'delete_it<MultiWorldTesting::EpsilonGreedyExplorer<vw_context> >'
requested here
    delete_it(data.greedy_explorer);
    ^
cbify.cc:329:64: warning: delete called on
'MultiWorldTesting::BootstrapExplorer<vw_context>' that has virtual
functions but
      non-virtual destructor [-Wdelete-non-virtual-dtor]
template<class T> inline void delete_it(T* p) { if (p != NULL) delete p; }
                                                               ^
cbify.cc:338:5: note: in instantiation of function template specialization
      'delete_it<MultiWorldTesting::BootstrapExplorer<vw_context> >'
requested here
    delete_it(data.bootstrap_explorer);
    ^
cbify.cc:329:64: warning: delete called on
'MultiWorldTesting::GenericExplorer<vw_context>' that has virtual functions
but
      non-virtual destructor [-Wdelete-non-virtual-dtor]
template<class T> inline void delete_it(T* p) { if (p != NULL) delete p; }
                                                               ^
cbify.cc:339:5: note: in instantiation of function template specialization
'delete_it<MultiWorldTesting::GenericExplorer<vw_context>
      >' requested here
    delete_it(data.generic_explorer);
    ^
cbify.cc:329:64: warning: delete called on 'vw_recorder' that has virtual
functions but non-virtual destructor
      [-Wdelete-non-virtual-dtor]
template<class T> inline void delete_it(T* p) { if (p != NULL) delete p; }
                                                               ^
cbify.cc:341:5: note: in instantiation of function template specialization
'delete_it<vw_recorder>' requested here
    delete_it(data.recorder);
    ^
cbify.cc:329:64: warning: delete called on 'vw_cover' that has virtual
functions but non-virtual destructor
      [-Wdelete-non-virtual-dtor]
template<class T> inline void delete_it(T* p) { if (p != NULL) delete p; }
                                                               ^
cbify.cc:347:5: note: in instantiation of function template specialization
'delete_it<vw_cover>' requested here
    delete_it(data.cover);
    ^
7 warnings generated.

JohnLangford:
This is 'yuck' I had not realized. It's yet another reason that virtual 
is iffy as a keyword.

I'd prefer extra code to extra warnings.

-John

JohnLangford:
Good point, thanks for the clarifications. Misattribution is indeed a concern. It seems that in this case, we can also remove the virtual keywords from the derived classes. The only downside of doing this that I know of is less readability. On the other hand however we can remove extra warnings & extra code. What do you think?

-----Original Message-----

bscottm:
If the code can eliminate virtual functions, then there's no need for a virtual destructor. I can probably take a look at that (caveat: STL containers might have an issue, but it's worth a shot to simplify the code.)

JohnLangford:
I'd certainly prefer neither extra code nor extra warnings.

-John

JohnLangford:
Eliminating the virtual functions is somewhat nontrivial, because the 
library interfaces into C#.

-John

JohnLangford:
We should only remove the virtual keywords from functions of derived classes so C# isn't affected.

Louie

JohnLangford:
Go ahead.

-John

lhoang29:
No problem, @bscottm just so we don't end up sending duplicate PR, I can take over from here.

lhoang29:
So the situation is a little different than I thought. The existing derived classes already do not use any virtual keywords, but the base abstract interfaces use virtual functions without virtual destructors (which is an oversight). I'm guessing clang++ detects this warning through the inheritance chain. I would think adding virtual dtors to the **base** classes is enough to fix the warnings (so we can avoid extra noop dtors in children types). @bscottm did you try this before? If not would you mind testing it out? Unfortunately I cannot get these warnings on Windows no matter what I try so I can't test if this is true.

bscottm:
@lhoang29: It gets worse. gcc 4.7 will tell you that non-virtual destructors will cause undefined behavior. And yes, I'll give the base class fix a shot.

bscottm:
@lhoang29: Derived classes need virtual destructors too. Those are the semantics. Creating new pull request with patch.

PR #513, opened 2015-02-12
minor fixes here and there, added graph prediction for search, initial fix for beam search

hal3:


JohnLangford:
This seems fine, except for the multipredict hacks in oaa.cc which break modularity.  

I think we need to instead create an appropriate interface to support multipredict in the learner and invoke it.  Can you remove?

hal3:
forgot about that -- reverted!

JohnLangford:
Merged, and removed 'ret' from the oaa datastructure.

Issue #510, opened 2015-02-10
Another long feedback


trufanov-nok:
Hi,

Below is my feedback for last 3 moths of VW usage. It's mostly feature requests, some ideas and only few non-critical bugs.

# I. VW

I spent last 3 months competing in kaggle's [avazu](http://www.kaggle.com/c/avazu-ctr-prediction) contest using only VW and end up 29th on private leaderboard. I suspect many VW users got better result, but I used only one model, no ensembles and was short of RAM, so it seems to be max what I can do.  I've used a fork of VW made 3 months ago and implemented dozens of heuristics in it. Some of them may be interesting and few definitely worth discussion. On the other hand I realised that several bugs I found (mostly with new_mf) were already fixed and in the main branch recently so I just wasted a time with them thus I paid my price for hiding info till competition ends.

## 1. namspaces redefinition
The first thing I have implemented in my VW fork is *--redefine* parameter which allows me to rename any namespace or set of namespaces. I gave  a new namespaces to each feature of original dataset taking their names were from [a-Z] interval. Then I declared a parameter *--redefine* which takes string argument in form of *x:=n* where *x* is a list of old namespaces, *:=* is an operator and *n* is a new namespace to which they should be assigned. That allowed me to put something like following in my vw command line:

``--redefine abcdefgihjklmnoprstquwz:=L --redefine abcdegf:=a --ignore L  --q aa``

In example above we assign all our namespaces to namespace *L* and then (order is matter) reassign some of them to namespace *a*. As all features got their own namespaces the namespace *a* will contain 7 features and *L* 23-7=16. Then namespace a is used to generate quadratic features.

Why it's convenient? I can now experiment with different features used for quadratic features generation by just addition or removal literals in  `--redefine abcdegf:=a `.
On practice I wrote a bash script which started with `abcdefgihjklmnoprstquwz:=a` as a baseline and used leave-one-out with validation to find features which removal could improve the result.

This parameter also allows you to do something like

``--redefine abc:=a --ignore a --redefine def:=b --redefine klmn:=c -q bc``

All namespaces which were not redefined will be used as before - as ordinary feature.

Another thing why I'm so enthusiastic about this function is because it's very easy to implement without any performance penalties. My implementation is described in appendix.

## 2. Simple combinations in -q and --cubic
[...]

PR #515, opened 2015-02-13
Multiple regressor initialisation fix (12)

trufanov-nok:
Fix for https://github.com/JohnLangford/vowpal_wabbit/issues/510
*if* clauses are reordered and *else* are added between them.
The copy of original bug report is:

## 12. Multiple regressor initialisation

In `initialize_regressor()` in `parse_regerssor.cc` it's worth adding else between these blocks
```
  if (all.random_weights)
{...}
   if (all.random_positive_weights)
{...}
   if (all.initial_weight != 0.)
{...}
```
 
 The `random_positive_weights` is set to true automatically with `--new_mf` mode in `mf.cc`. I found that these random values ( `0.1 * frand48()` ) are too big for my dataset (9000 features per example). So I initialised regressor with help of `--initial_weight` with something like 1e-5 and got better convergence. But now regressor initialises twice. I would put *else* clauses between these 3 *if*'s and reorder them so `all.initial_weight` comes first.

### DIFF
--- a/vowpalwabbit/parse_regressor.cc
+++ b/vowpalwabbit/parse_regressor.cc
@@ -36,20 +36,22 @@ void initialize_regressor(vw& all)
     {
       cerr << all.program_name << ": Failed to allocate weight array with " << all.num_bits << " bits: try decreasing -b <bits>" << endl;
       throw exception();
-    }
+    } else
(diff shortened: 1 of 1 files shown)


JohnLangford:
Sounds good, merged.

PR #516, opened 2015-02-13
fix for typo in initial_weight help (14)

trufanov-nok:
Fixing help for initial_weight as described in 14 of
https://github.com/JohnLangford/vowpal_wabbit/issues/510

Copy of original bug report
## 14. Typos.
The help line for initial_weight says:
``--initial_weight arg              Set all weights to an initial value of 1.``

Shall " be to an initial value of **arg**." I suppose.


JohnLangford:
Good, thanks.

PR #518, opened 2015-02-13
lrq empty namespace test + fix

pmineiro:
Fixes reported bug.


arielf:
Thanks for fixing this so quickly.  Highly appreciated.

PR #519, opened 2015-02-15
Redefine namespace param implementation (1)

trufanov-nok:
According to https://github.com/JohnLangford/vowpal_wabbit/issues/510
This pull request contains implementation of --redefine param.

Original proposal:

## 1. namspaces redefinition
The first thing I have implemented in my VW fork is *--redefine* parameter which allows me to rename any namespace or set of namespaces. I gave  a new namespaces to each feature of original dataset taking their names were from [a-Z] interval. Then I declared a parameter *--redefine* which takes string argument in form of *x:=n* where *x* is a list of old namespaces, *:=* is an operator and *n* is a new namespace to which they should be assigned. That allowed me to put something like following in my vw command line:

``--redefine abcdefgihjklmnoprstquwz:=L --redefine abcdegf:=a --ignore L  --q aa``

In example above we assign all our namespaces to namespace *L* and then (order is matter) reassign some of them to namespace *a*. As all features got their own namespaces the namespace *a* will contain 7 features and *L* 23-7=16. Then namespace a is used to generate quadratic features.

Why it's convenient? I can now experiment with different features used for quadratic features generation by just addition or removal literals in  `--redefine abcdegf:=a `.
On practice I wrote a bash script which started with `abcdefgihjklmnoprstquwz:=a` as a baseline and used leave-one-out with validation to find features which removal could improve the result.

This parameter also allows you to do something like

``--redefine abc:=a --ignore a --redefine def:=b --redefine klmn:=c -q bc``

All namespaces which were not redefined will be used as before - as ordinary feature.

Another thing why I'm so enthusiastic about this function is because it's very easy to implement without any performance penalties.

### DIFF
--- a/vowpalwabbit/global_data.h
+++ b/vowpalwabbit/global_data.h
@@ -293,6 +293,9 @@ struct vw {
   bool ignore_some;
   bool ignore[256];//a set of namespaces to ignore
 
+  bool redefine_some;          // --redefine param was used
+  unsigned char redefine[256]; // keeps new chars for amespaces
+
   std::vector<std::string> ngram_strings; // pairs of features to cross.
   std::vector<std::string> skip_strings; // triples of features to cross.
   uint32_t ngram[256];//ngrams to generate.
--- a/vowpalwabbit/parse_args.cc
+++ b/vowpalwabbit/parse_args.cc
@@ -311,7 +311,7 @@ void parse_feature_tweaks(vw& all)
     ("hash", po::value< string > (), "how to hash the features. Available options: strings, all")
     ("ignore", po::value< vector<unsigned char> >(), "ignore namespaces beginning with character <arg>")
     ("keep", po::value< vector<unsigned char> >(), "keep namespaces beginning with character <arg>")
-    ("redefine", po::value< vector<string> >(), "redefine namespaces beginning with characters from string <s>... as namespace <n>. <arg> shall be in form 's:=n' where := is operator. Use ':=n' to redefine all namespaces.")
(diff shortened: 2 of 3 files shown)


trufanov-nok:
The `-redefine <arg>` processing has been changed
Now `<arg>` is in form `N:=S` where `:=` is operator, `N` - new namespace, `S`- list of old namespaces.  
The behaviour is following:  
`A:=abc` - redefine namespaces a, b and c as A
`ABC:=ABC` - same as above as only 1 character of N part is used
`A:=` - default namespace (' ') is redefined as A
`:=ABC` - namespaces a, b and c are redefined as default namespace (' ')
`:=` - default namespace (' ') is redefined as default namespace (' '), so no effect
`:=:` - wildcard ':' is used - all namespaces except default one (' ')  are redefined as default namespace (' ')
`:=:=` - same as above. First occurrence of operator has effect.
`:=abc:def` - same as above. S part processing will be stopped after reach `:` wildcard.  all namespaces except default one (' ')  will be redefined as default namespace (' ') at this point.
`--redefine A:= --redefine A:=:` - all namespaces including default one (' ') are redefined as A namespace.
`::=` - default namespace (' ') is redefined as namespace ':'.

JohnLangford:
This looks good, except for one minor request: can you make ABC:=abc generate a warning?  Using that indicates someone does not understand what is going on.

trufanov-nok:
It's done. Please check warning wording.

JohnLangford:
Merged, thanks.

PR #524, opened 2015-02-18
--save_resume fix (9)

trufanov-nok:
This shall fix --save_resume
I'll add a sample a bit later

As I can see line
``float sparse_l2;``
Isn't in gd.cc anymore. Was it deleted? Shall I reflect this in my pull request?
I also quite unsure in confiure.ac changes. Is it ok or there is other way to handle VW version change?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,4 +1,4 @@
-AC_INIT([vowpal_wabbit], [7.8], [jl@hunch.net])
+AC_INIT([vowpal_wabbit], [7.10.1], [jl@hunch.net])
 AC_DEFINE([PACKAGE_URL],["https://github.com/JohnLangford/vowpal_wabbit"],[project url])
 AC_CONFIG_HEADERS(vowpalwabbit/config.h)
 AM_INIT_AUTOMAKE()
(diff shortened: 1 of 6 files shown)


JohnLangford:
It's complaining that it can't automerge.  Can you reconcile with current master?

--sparse_l2 is something experimental that I added this morning.

trufanov-nok:
Few notes on this one:

This patch affects GD and FTRL
I found 3 problems in GD if its state is restored from save_resume model:
a). average loss is different 
b). average loss since last is different in print_update() for first outputted value
c). number of examples per pass is wrong in vw' s summary

a) could be fixed by restoring `g.total_weight value`, b) - by restoring `sd->old_weighted_examples` and c). by restoring `all.current_pass values`.

as `g` object was removed from `save_load_online_state` arguments list (bcs of its call from ftrl, suppose)I've added `g*` as an optional argument. That required `save_load_online_state()` declaration change in `gd.h` and forward declaration of `struct gd;` before it. (As `struct gd` is declared and implemented solely in gd.cc. Moving it to .h caused some missing declarations for its fields)

I've defined new constant 
``#define VERSION_SAVE_RESUME_FIX "7.10.1"``
to compare while loading model. Current value in config.am is 7.8 which confused me a little, as latest ver should be 7.10. Perheps i'm doing something wrong here.

Warnings are shown only if both `--save_resume` and `-i` parameters are used and model file format is old. I'm checking save_resume field of file instead of related parameter of vw command line as it may be missing.

As model file version is checked in one .cc file and save_resume flag of model file in another .cc file I was forced to store model file version in all.model_file_version filed. This in its turn required a small modification in constructor of `version_struct``

trufanov-nok:
Please delete this pull request - i've send another one

PR #531, opened 2015-02-21
autotools_cleanups

bscottm:
Add tests for large file support, eliminating the hacky  _FILE_OFFSET_BITS=64 command line definition. Large file support is enabled on platforms where this is an option.

Disentangle profiling option from compiler optimization. Adds a new configure flag, "--enable-debug", that sets the compiler's optimization to "-g -O1". The default is the highly optimized compiler setting.

NOTE: Regenerate Makefile(s) after the pull request is merged for the default platform.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,8 +51,8 @@ $ make install
 
 Note that ``./autogen.sh`` requires *automake* (see the prerequisites, above.)
 
-Options that were passed to `./configure` in 7.6 and earlier may now be passed
-to `./autogen.sh`.
+`./autogen.sh`'s command line arguments are passed directly to `configure` as
+if they were `configure` arguments and flags.
 
 Be sure to read the wiki: https://github.com/JohnLangford/vowpal_wabbit/wiki
 for the tutorial, command line options, etc.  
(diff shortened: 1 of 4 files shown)


arielf:
Thanks for this.  There was an issue report on the mailing list (by Dan Swanson) of failure to build on CentOS, possibly because autotools introduces `-std=gnu++11` flag into the auto generated Makefile.  The default Makefile actually forces `-std=c++0x` - can we force this in autotools as well?

bscottm:
It's not autotools that introduces the flag, per se. The problem is that VW distributes Makefiles, which captures whatever configure deduced when the Makefile was written.

Probably better to simply go suggest that users do the usual:

```
$ ./configure               # (or ./autogen.sh)
$ make
$ make install
```

bscottm:
There's a lot of old cruft leftover in the VW build system. Doesn't look like the library/Makefile has been regenerated in ages -- implying that anyone who runs `autogen.sh` will never be able to `make test` successfully because `Makefile.am` doesn't have those extra targets, like `ezexample_predict`.

PR #533, opened 2015-02-23
You can use brew for a complete VW installation

zachmayer:


kamal94:
Is it possible to pleeeeaaase add the installation of the python wrapper to the full installation? It is a nightmare trying to install it separately, especially on Mac.

JohnLangford:
We need someone to master and maintain the Mac installation.  A core 
issue here is that most of us working on VW don't have access to a Mac.  
Any volunteers?  This is the sort of thing which should require a day or 
two to setup carefully but then be quite easy to keep current.

-John

gramhagen:
I'm not an expert on building on Mac, but I've got one and have wrestled
with VW a bit, so happy to help out.  Also, putting the python wrapper as
an actual package could help with the install.  Though you might not want
to automatically install it given people typically work in virtual envs.
But adding a make option or more documentation and packaging could make
that much easier.

scott

PR #534, opened 2015-02-24
prefer_cxx_io

bscottm:
Remove some vestiges of C printf formatting. Reduces warnings on non-x86
platforms, like ARM, where size_t isn't necessarily unsigned long and formatting
warnings are enabled.

### DIFF
--- a/vowpalwabbit/cb_algs.cc
+++ b/vowpalwabbit/cb_algs.cc
@@ -303,15 +303,13 @@ using namespace CB;
   {
     if (all.sd->weighted_examples >= all.sd->dump_interval && !all.quiet && !all.bfgs)
       {
-        char label_buf[32];
+	std::string label_buf;
         if (is_test)
-          strcpy(label_buf," unknown");
+          label_buf = " unknown";
         else
-          sprintf(label_buf," known");
-	char pred_buf[32];
-	sprintf(pred_buf,"%8lu",(long unsigned int)ec.pred.multiclass);
+          label_buf = " known";
 
-	all.sd->print_update(all.holdout_set_off, all.current_pass, label_buf, pred_buf, 
+	all.sd->print_update(all.holdout_set_off, all.current_pass, label_buf, ec.pred.multiclass, 
 			     ec.num_features, all.progress_add, all.progress_arg);
       }
   }
--- a/vowpalwabbit/cost_sensitive.cc
+++ b/vowpalwabbit/cost_sensitive.cc
@@ -193,15 +193,13 @@ namespace COST_SENSITIVE {
 	    num_current_features += (*ecc)->num_features;
         }
 
-        char label_buf[32];
+	std::string label_buf;
         if (is_test)
-          strcpy(label_buf," unknown");
+          label_buf = " unknown";
         else
-          sprintf(label_buf," known");
-	char pred_buf[32];
-	sprintf(pred_buf,"%8lu",(long unsigned int)ec.pred.multiclass);
+          label_buf = " known";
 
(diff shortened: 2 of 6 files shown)


bscottm:
One thing I could do is make the column widths "constexpr" constants, which would reduce future risk of column alignment issues.

Running the test now… really warm Mac laptop. :-)

arielf:
This looks almost perfect. Thanks!
- Can you also add `std::fixed`, to align the decimal dot and add trailing zeroes where needed?  The current-label and current-predict columns look wobbly due to the right adjustment.  Same should be applied to the 1st two floating point columns.  See http://stackoverflow.com/questions/9131090/c-output-align-and-decimal-place-holder
- Can you change the title of the push to the more descriptive:  "fix progress alignment and prefer C++ io" ?

Thanks again for all your recent contributions!

bscottm:
The current-label and current-predict columns are passed in as char[] arrays. They aren't passed in as their native types because in multilabel, those quantities are integral, whereas for simple labels, those quantities are floating point.

But that's why we have function overloading, isn't it? :-)

bscottm:
Squashed the commit for easier reading.

arielf:
@bscottm
The travis auto-build failed after the last commit:
https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/52318742
Seems like `--holdout_off` wasn't respected @ test #35.  Suggest to always run `make test` before a push, one may never know...  Thanks again!

bscottm:
I'll look into it.

bscottm:
<groan> Changing precision back to 4.

bscottm:
It's not really a build fail or a test fail, per se. It's white space differences. Do I need to clean that up as well?

arielf:
Thanks again.
`test/RunTests` explicitly ignores pure white-space differences by calling `diff` with the `--ignore-all-space` option so spaces alone cannot fail a test.  Test 35 stderr has higher precision on some of the summary numbers which is actually cool/fine, but it may require a refresh of the reference-stderr file.

Another thing I noticed is that there's a new _leading_ space on the progress lines. This difference is actually/normally ignored, but here it appears because the diff fails on a non-space difference later.  Is this redundant leading space really necessary? Perhaps we should remove it.  Thanks.

bscottm:
Are you sure the leading space that isn't due to diff context? :-)

arielf:
Yes, in the original code, the two leftmost (loss) columns were left justified so there was no leading space.

Some utilities like `utl/vw-convergence` rely on this (finding a number right at the start of the line in the progress output.) so I think we shouldn't change this.

`loss` column values are always non-negative (so they don't need space for sign) and not big (except for deliberately pathological cases) so 6 digits of precision should fit in a width=8, precision=6 left-justified field.  We need only 2-chars for a leading `0.` - No need to give them extra space like we do for `example counter`, plus keeping them left-justified is better for backward compatibility.

bscottm:
Ok, so I frobbed the formatting. I left the numeric parts of the column right justified, but that shouldn't really matter at this point.

arielf:
Great.
Looks like we have a perfect fix, more maintainable code, and full passing of the tests.
Thanks so much for all the work on this.

bscottm:
I'm sure there are other places where C-style formatting can be removed; something I can tackle incrementally, e.g., search.cc.

More important problem at the moment is the build system -- a choice needs to be made whether the handcrafted Makefiles are the build system or whether the project should go fully over to autotools?

JohnLangford:
Thanks much --- cleaning these little things up is very helpful in the long term.

PR #535, opened 2015-02-24
assume_stricter_alignment

bscottm:
Don't assume x86-like lax alignment or bus errors occur. Use memcpy() to
ensure unaligned data is copied correctly on all platforms. (Found on ARM).


arielf:
Thanks for the extra legwork.  I used to work on compilers in the past so this optimization isn't surprising. Good to know.  The `movss` sequence might still be slower for IA64 which has efficient non-aligned access but I'm guessing not by much.  You may time the two versions when using `-k -c` on a reasonable sized data-set on AMD64/IA64.  Anyway, hope John accepts this pull request.

bscottm:
Found a few more places where memcpy() is needed. The reinterpret_cast<> isn't strictly necessary, but gives the code a C++ look-and-feel.

PR #539, opened 2015-02-27
arm_neon_simd

bscottm:
Consolidate SIMD used in gd.cc to unify the different implementations of inverse square root. Add ARM NEON implementation.

Rename: VW_LDA_NO_SSE2 -> VW_NO_INLINE_SIMD (i.e., -DVW_NO_INLINE_SIMD on the compiler's command line to disable SIMD, SSE2 or NEON).

NOTE: "static inline" tells g++ and clang++ that InvSqrt() is a leaf function, so it can be aggressively inlined.

### DIFF
--- a/vowpalwabbit/gd.cc
+++ b/vowpalwabbit/gd.cc
@@ -10,8 +10,12 @@ license as described in the file LICENSE.
 #include <netdb.h>
 #endif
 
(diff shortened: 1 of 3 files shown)


JohnLangford:
Merged, thanks.

PR #544, opened 2015-03-03
autotool_project_fixups

bscottm:
Update the autotools support, deprecate the older handcrafted Makefile
structure. Updated aux autoconf macros in acinclude.d to the latest
versions.

Developers should only need to run 'autogen.sh' if 'configure.ac' or
any of the 'Makefile.am' automake templates are modified.

Ordinary users:
```
	$ ./configure <configure-flags> CXX=<something> CC=<something>
	$ make
	$ make test
	$ make install
```

MAJOR CHANGES:
(a) Building the Python module: Enable this explicitly via configure:

	configure --enable-python

(b) Build the Java JNI module: Enable this explicitly via configure

	configure --enable-java-jni


JohnLangford:
There are about 3 issues here.  The first is that travis complains.  The second is that I'm not ready to abandon the old Makefile---it's far too convenient.  The third is that this this is a giant checkin with many thousands of lines.  It would be nice to have a build system with somewhat fewer lines than the project itself.

bscottm:
Any change to the build system will be giant. But I did learn how to get Travis to cooperate, mostly, off my own repo instead of spamming yours.

If you want, reject the pull and I'll resubmit a pull request at a later time when the patch is more robust.

PR #549, opened 2015-03-04
added multipredict, threaded through scorer, used in oaa

hal3:


hal3:
this currently has multiupdate in it, which is of dubious use. i can yank it if you want.

JohnLangford:
Yes, probably best to yank.

-John

hal3:
strangely, some other commit that i integrated completely removed multipredict *AND* multiupdate, so neither of them is in there right now, really (at least they're not in gd.cc which means that they're never actually being used). it'll take some effort to undo that git error.

hal3:
ok it's fixed.

hal3:
just fixed build error.

JohnLangford:
Why the change to do-lrq-hogwild?

Are you sure that you want --audit in test.py?

Do we need beam.h in the present code?

Why are class_weight_multipliers useful?

This doesn't matter in the long term, but the indent change on csoaa.cc generates lots of lines of diff...

Why is there a multiupdate_f in learner.h?

Why is vw_is_main needed?  Any why init_time?  I'm generally trying to reduce the use of the global datastructure.

Instead of alloca, can you allocate during setup?

You are trying to use multiupdate in oaa.cc.  Is that intentional?

Why do we want print out final stats in finish()? 

Is the sqrtmc() code worth including?

hal3:
> Why the change to do-lrq-hogwild?

this was something paul and I worked out when he was here because 
something was dying. i don't remember anymore what :)

> Are you sure that you want --audit in test.py?

nope, removed

> Do we need beam.h in the present code?

not right now but once search_meta.cc gets added back, it'll be required.

> Why are class_weight_multipliers useful?

it might be worth yanking this out. the reason i put it in was to test 
out some stuff where it was inconvenient to specify per-example the 
costs for each class because they were global costs.

> This doesn't matter in the long term, but the indent change on csoaa.cc
> generates lots of lines of diff...

i know... i noticed that after the fact... i must have accidentally hit 
tab with a lot of stuff selected and not noticed.

> Why is there a multiupdate_f in learner.h?

straggler code -- removed!

> Why is vw_is_main needed? Any why init_time? I'm generally trying to
> reduce the use of the global datastructure.

vw_is_main -- this came up when figuring out stuff with paul. basically 
the issue is that when vw is the main executor, you can avoid running 
some code that _is_ necessary when you're in library mode.

a specific example is whether search needs to produce an output for all 
training data. as an extreme example, if you're running with --quiet, 
then you never need to print accuracy or anything like that, which means 
you can avoid running test passes, which speeds things up (especially if 
REF is used as one of the rolls).

init_time was so that we could do speed tests for hogwild. i can remove it.

> Instead of alloca, can you allocate during setup?

where is this?

> You are trying to use multiupdate in oaa.cc. Is that intentional?

nope, yanked

> Why do we want print out final stats in finish()?

where is this?

> Is the sqrtmc() code worth including?

probably not. happy to just keep it in my own branch.

  -h

JohnLangford:
alloca: https://github.com/JohnLangford/vowpal_wabbit/pull/549/files#diff-b939599be895a4b4d610b6a0a68c6797R58

finish: https://github.com/JohnLangford/vowpal_wabbit/pull/549/files#diff-37715d20272c8511f892a5777571886aR1049

There seems to still be a fair amount of multiupdate stuff in the diff. 

I'd suggest pulling sqrtmc() and class_weight_multipliers for now. 

(Also, sync to master)

(Also, the merge step would be easier if independent things were in independent branches).

hal3:
> finish:
> https://github.com/JohnLangford/vowpal_wabbit/pull/549/files#diff-37715d20272c8511f892a5777571886aR1049

ah, this was something paul and i did so that when you run in hogwild 
mode you only get one output rather than a bunch of bashed together outputs.

> There seems to still be a fair amount of multiupdate stuff in the diff.

yup, i'm kiling it

> I'd suggest pulling sqrtmc() and class_weight_multipliers for now.

ok

> (Also, sync to master)
>
> (Also, the merge step would be easier if independent things were in
> independent branches).

that would require me learning more than pull/commit/push as my git 
vernacular :)

  -h

hal3:
ok all of these items are addressed in the latest version

JohnLangford:
> that would require me learning more than pull/commit/push as my git
> vernacular :)

Probably worth your time.  Git is very good at branch & merge, which means
you can have multiple threads of work open and in progress simultaneously.
In addition to making for easier pulls, this means that you can more freely
interrupt yourself if something comes.

-John

JohnLangford:
Test 70 fails.

https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/58950661

hal3:
urgh, this looks like some floating point precision issue.... it doesn't 
fail locally for me. give me a sec and i'll address.

hal3:
fixed

JohnLangford:
There's still an alloca:
https://github.com/JohnLangford/vowpal_wabbit/pull/549/files#diff-d19573de0801985b45f745e703dcfc66R141

Factor into the nn datastructure?

Why is the accounting getting moved?
https://github.com/JohnLangford/vowpal_wabbit/pull/549/files#diff-f411c38884e8d38a5120bbfa2a5e3d1cL86

hal3:
i can remove the alloca's -- those are pauls fault though :P.

the acocunting getting moved for the reason i said in the last comment: 
so that only a single output comes out of hogwild mode

hal3:
done

JohnLangford:
Got it.  

But there is still an alloca

https://github.com/JohnLangford/vowpal_wabbit/pull/549/files#diff-d19573de0801985b45f745e703dcfc66R144

Fix?

hal3:
fixed

PR #552, opened 2015-03-09
prefer_nullptr

bscottm:
[Primarily stylistic]
Fix compile error in multilabel.cc: '\0' is not a pointer type. It is a character and shouldn't be used as a subsitute for NULL, even if that was previously valid in C. (clang++ 3.7 flagged this as an error.) (Note: it's also possible that the fix in multilabel.cc is incorrect and should have had a pointer dereference.)

Stylistic consistency: Prefer the 'nullptr' keyword over 'NULL' because it is more type safe than NULL (0). 'nullptr' was introduced as part of the C++0x standard; VW requires at least C++0x support to compile.

### DIFF
--- a/vowpalwabbit/active.cc
+++ b/vowpalwabbit/active.cc
@@ -128,7 +128,7 @@ float get_active_coin_bias(float k, float avg_loss, float g, float c0)
   
 base_learner* active_setup(vw& all)
 {//parse and set arguments
-  if(missing_option(all, false, "active", "enable active learning")) return NULL;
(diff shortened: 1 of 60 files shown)


arielf:
John, please accept this. Currently top of tree fails to compile with clang:
```
multilabel.cc:108:27: error: assigning to 'char *' from incompatible type 'char'
            p->parse_name[i].end = '\0';
                                 ^ ~~~~
```
Thanks.

JohnLangford:
This is merged in.  The \0 was supposed to be an assignment, will fix.

PR #553, opened 2015-03-09
Initial pass at a basic SVRG implementation.

froystig:


JohnLangford:
All, we now have an initial version of SVRG in VW as a base learning 
algorithm.  For those who don't know, Variance Reduced Gradient 
plausibly helps convergence to a good solution significantly in some 
situations.  Any feedback about where this performs well or poorly is 
appreciated.

This implementation doesn't yet merge with the various elements of the 
default VW update rule, but we are looking in that direction.

-John

vowpalwabbit/svrg.cc
```diff
@@ -0,0 +1,182 @@
...
+  GD::foreach_feature<float, update_stable_feature>(*s.all, ec, g);
+}
+
+void learn(svrg& s, base_learner& base, example& ec)
+{
+  assert(ec.in_use);
+
+  predict(s, base, ec);
+
+  const int pass = s.all->passes_complete;
```
zjf: Is this the correct way to get the No. of passes? I think it will be bigger than the actual pass, because the parser thread is usually faster than the learner thread.

froystig: Your concern seems reasonable -- I've noticed the same. I have replaced this logic with something more correct in an incoming changeset, using the `end_pass()` interface.

(The change is sitting between other parts of work-in-progress that I have, on a local branch that's nearly complete. I'll aim to arrive at a pull request point soon. Is the parser/learner discrepancy causing issues for you in the meantime?)

zjf:
I am doing some benchmark between SGD and SVRG, using the RCV1 dataset. The final loss of SVRG (0.47) is much bigger than that of SGD (0.04). So I take a peek into the source code and try to get out what is going on. Btw, I guess the incorrect calculation of pass will not cause a significant decrease in performance. Is the following command for SVRG correct?

```
vw rcv1/rcv1.train.vw --cache_file cache_train -f sgd.model --passes 20
vw rcv1/rcv1.train.vw --cache_file cache_train -f svrg.model --passes 20 --svrg --stage_size 2 
```

arielf:
The default update in `vw` in not 'classic' SGD.  It is a significantly enhanced version of SGD which is scale, sparsity & importance adjusted.  It is very hard to beat it in my experience. To get 'classic' SGD you need to explicitly pass the `--sgd` option.  This will turn off the 3 default enhancements: `--normalized`, `--invariant`, `--adaptive`

zjf:
@arielf I'm aware that there are many optimization tricks in the enhanced SGD implementation, but for this dataset, the classic SGD also gives a loss of 0.04. So I'm wondering whether I use SVRG in a right way.

arielf:
Could `--svrg` simply be broken or not fully ready, in the current top of tree?

When I run vanilla `vw` on the tree `test/train-sets/0002.dat` (one pass) I get 0.003847 avg loss and reasonable predictions.  When I add `--svrg` I get avg loss 0.283062 and all online progress predictions are 0.0000. Here's the full output:

```
vw --svrg test/train-sets/0002.dat
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = test/train-sets/0002.dat
num sources = 1
svrg pass 0: committing stable point
svrg pass 0: computing exact gradient
0.271591 0.271591            1            1.0   0.5211   0.0000       15
0.279042 0.286494            2            2.0   0.5353   0.0000       15
0.300773 0.322504            4            4.0   0.5854   0.0000       15
0.309289 0.317804            8            8.0   0.5575   0.0000       15
0.295603 0.281916           16           16.0   0.5878   0.0000       15
0.298948 0.302294           32           32.0   0.6038   0.0000       15
0.293966 0.288983           64           64.0   0.5683   0.0000       15
0.290021 0.286077          128          128.0   0.5351   0.0000       15
0.295309 0.300597          256          256.0   0.5385   0.0000       15
0.279036 0.262763          512          512.0   0.5053   0.0000       15
svrg pass 1: taking steps

finished run
number of examples = 1000
weighted example sum = 1000.000000
weighted label sum = 526.517586
average loss = 0.283062
best constant = 0.526518
total feature number = 14996
```

zjf:
@arielf The output is in expectation. The first step of SVRG algorithm, according to the description in the paper, is looping through all the examples using the initial weight (which happens to be zero in vw's implementation). I think we could replace this step with the enhanced SGD in vw and use SRVG in the next steps. Anyway, we are accustomed to the fact that vw could generate fairly well model in one pass.

froystig:
Correct, @arielf -- "isn't fully ready" is certainly accurate. In response to questions and points so far:

- It's always possible that the version currently in the tree has issues. It hasn't seen much use yet.
- One needs to have VW take several passes at present. Right now, the first pass is spent computing just a single stable gradient and not making any updates. This means that the main regressor's parameters are still all 0 at the end of the first pass -- error rates above 40% are no surprise in light of that.
- More generally, this implementation interacts poorly with VW's standard incremental evaluation. It does not perform updates at every step, for one. @arielf is correct that comparisons are only "fair" against plain SGD, yet even then they are still not exactly comparable by reading the incremental output from VW. Perhaps a better test for now is to print out predictions to a file in the final pass and compare them.

I have these and a couple of other design-level issues in mind in my next set of changes (e.g. beginning with small stages that double in size until reaching the full-pass mark).

At any rate, the above might explain the behavior that you are seeing. @zjf, I would not guess that the parser/learner discrepancy is the central problem; pass lengths can be estimated more correctly (and will be soon), but this should just be small noise on top of, say, incorrect evaluation numbers, or not updating the regressor at all. :)

Thanks to both of you for taking a look!

arielf:
Thanks @froystig .  Another input: after adding `-c --passes 20` the predictions are mostly `0.0000` with an occasional `1.0000`.  Actual labels in this small data-set (1000 examples) are real values between 0.0 and 1.0 centered at around 0.5. If I add `--initial_weight 0.5` I get predictions mostly `1.0000` . Seems to require trial-and-error with `--initial-weight` to get any values between 0.0 and 1.0.

JohnLangford:
Sounds like there is a bug in SVRG.  Roy, if you have some time to look
into it, that would be helpful.

-John

froystig:
Indeed. Thanks, @arielf. I'll have a look.

PR #555, opened 2015-03-12
lda_core.cc SIMD refactoring

bscottm:
[Sorry if the change set looks huge, caused in large part by running the code through clang-format to make it more readable.]

Refactor the LDA inline SIMD code into two blocks: one block for fast approximations vs. accuracy, the second block for SIMD code vs. serial code.

Update x86 SIMD code and use SSE3, SSE4.2 intrinsics, if the compiler has them enabled. Vector element extraction is template code with a specialization for the 0'th element. All of this results in better instruction selection in clang, gcc. [GCC 4.6.2 supports SSE4.1 with the "-msse4.1" flag. Not sure if this is desired in the Travis builds.]

Also prefer STL algorithms in a few places where they make sense, show examples of how C++ lambdas work (it's a C++0x feature.)

### DIFF
--- a/vowpalwabbit/lda_core.cc
+++ b/vowpalwabbit/lda_core.cc
@@ -7,7 +7,8 @@ license as described in the file LICENSE.
 #include <vector>
 #include <algorithm>
 #include <numeric>
-#include <boost/align/is_aligned.hpp>
+#include <cmath>
+
 #include <boost/math/special_functions/digamma.hpp>
 #include <boost/math/special_functions/gamma.hpp>
 
@@ -23,6 +24,13 @@ license as described in the file LICENSE.
 #include "gd.h"
 #include "rand48.h"
 #include "reductions.h"
+
+#include <boost/version.hpp>
+
+#if BOOST_VERSION >= 105600
+#include <boost/align/is_aligned.hpp>
+#endif
+  
 
 enum lda_math_mode { USE_SIMD, USE_PRECISE, USE_FAST_APPROX };
 
@@ -66,6 +74,18 @@ struct lda {
 };
 
 // #define VW_NO_INLINE_SIMD
+
+namespace
+{
+  inline bool is_aligned16(void *ptr)
+  {
+#if BOOST_VERSION >= 105600
+    return boost::alignment::is_aligned(16, ptr);
+#else
+    return (reinterpret_cast<uintptr_t>(ptr) & 0x0f == 0);
+#endif
+  }
+}
 
 namespace ldamath
 {
@@ -147,7 +167,7 @@ namespace ldamath
   // Extract v[idx]
   template <const int idx> float v4sf_index(const v4sf x)
   {
-#ifdef __SSE4_1__
+#if defined(__SSE4_1__)
     float ret;
     uint32_t val;
 
@@ -227,22 +247,20 @@ namespace ldamath
 
   void vexpdigammify(vw &all, float *gamma)
   {
-    unsigned int n = all.lda;
     float extra_sum = 0.0f;
     v4sf sum = v4sfl(0.0f);
-    size_t i;
     float *fp;
     const float *fpend = gamma + all.lda;
 
     // Iterate through the initial part of the array that isn't 128-bit SIMD
     // aligned.
-    for (fp = gamma; fp < fpend && !boost::alignment::is_aligned(16, fp); ++fp) {
+    for (fp = gamma; fp < fpend && !is_aligned16(fp); ++fp) {
       extra_sum += *fp;
       *fp = fastdigamma(*fp);
     }
 
     // Rip through the aligned portion...
-    for (; boost::alignment::is_aligned(16, fp) && fp + 4 < fpend; fp += 4) {
+    for (; is_aligned16(fp) && fp + 4 < fpend; fp += 4) {
       v4sf arg = _mm_load_ps(fp);
       sum += arg;
       arg = vfastdigamma(arg);
@@ -254,7 +272,7 @@ namespace ldamath
       *fp = fastdigamma(*fp);
     }
 
-#if defined(__SSE3__)
+#if defined(__SSE3__) || defined(__SSE4_1__)
     // Do two horizontal adds on sum, extract the total from the 0 element:
     sum = _mm_hadd_ps(sum, sum);
     sum = _mm_hadd_ps(sum, sum);
@@ -266,11 +284,11 @@ namespace ldamath
     extra_sum = fastdigamma(extra_sum);
     sum = v4sfl(extra_sum);
 
-    for (fp = gamma; fp < fpend && !boost::alignment::is_aligned(16, fp); ++fp) {
+    for (fp = gamma; fp < fpend && !is_aligned16(fp); ++fp) {
       *fp = fmax(1e-6f, fastexp(*fp - extra_sum));
     }
 
-    for (; boost::alignment::is_aligned(16, fp) && fp + 4 < fpend; fp += 4) {
+    for (; is_aligned16(fp) && fp + 4 < fpend; fp += 4) {
       v4sf arg = _mm_load_ps(fp);
       arg -= sum;
       arg = vfastexp(arg);
@@ -289,11 +307,11 @@ namespace ldamath
     const float *np;
     const float *fpend = gamma + all.lda;
 
-    for (fp = gamma, np = norm; fp < fpend && !boost::alignment::is_aligned(16, fp); ++fp, ++np) {
+    for (fp = gamma, np = norm; fp < fpend && !is_aligned16(fp); ++fp, ++np) {
       *fp = fmax(1e-6f, fastexp(fastdigamma(*fp) - *np));
     }
 
-   for (; boost::alignment::is_aligned(16, fp) && fp + 4 < fpend; fp += 4, np += 4) {
+   for (; is_aligned16(fp) && fp + 4 < fpend; fp += 4, np += 4) {
       v4sf arg = _mm_load_ps(fp);
       arg = vfastdigamma(arg);
       v4sf vnorm = _mm_loadu_ps(np);
@@ -328,34 +346,25 @@ namespace ldamath
   // Log gamma:
   template <typename T, const lda_math_mode mtype> inline T lgamma(T x)
   {
-    // This is an intentional crash for generic types.
-    std::cerr << "Using generic lgamma<T, bool> template, aborting." << std::endl;
-    abort();
+    BOOST_STATIC_ASSERT_MSG(true, "ldamath::lgamma is not defined for this type and math mode.");
   }
 
   // Digamma:
   template <typename T, const lda_math_mode mtype> inline T digamma(T x)
   {
-    // This is an intentional crash for generic types.
-    std::cerr << "Using generic digamma<T, lda_math_mode> template, aborting." << std::endl;
-    abort();
-    return static_cast<T>(0);
+    BOOST_STATIC_ASSERT_MSG(true, "ldamath::digamma is not defined for this type and math mode.");
   }
 
   // Exponential
   template <typename T, lda_math_mode mtype> inline T exponential(T x)
   {
-    // This is an intentional crash for generic types.
-    std::cerr << "Using generic exponential<T, bool> template, aborting." << std::endl;
-    abort();
+    BOOST_STATIC_ASSERT_MSG(true, "ldamath::exponential is not defined for this type and math mode.");
   }
 
   // Powf
   template <typename T, lda_math_mode mtype> inline T powf(T x, T p)
   {
-    // This is an intentional crash for generic types.
-    std::cerr << "Using generic powf<T, bool> template, aborting." << std::endl;
-    abort();
+    BOOST_STATIC_ASSERT_MSG(true, "ldamath::powf is not defined for this type and math mode.");
   }
 
   // High accuracy float specializations:
@@ -363,7 +372,7 @@ namespace ldamath
   template <> inline float lgamma<float, USE_PRECISE>(float x) { return boost::math::lgamma(x); }
   template <> inline float digamma<float, USE_PRECISE>(float x) { return boost::math::digamma(x); }
   template <> inline float exponential<float, USE_PRECISE>(float x) { return std::exp(x); }
-  template <> inline float powf<float, USE_PRECISE>(float x, float p) { return std::powf(x, p); }
+  template <> inline float powf<float, USE_PRECISE>(float x, float p) { return std::pow(x, p); }
 
   // Fast approximation float specializations:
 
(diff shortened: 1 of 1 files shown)


bscottm:
If Boost.SIMD ever becomes a real thing (or the Github repo actually has working code), that will be a far better option than the hand-rolled intrinsics. Until then... this refactoring does allow me to add ARM NEON intrinsics for that port.

arielf:
Scott, this looks nice.  Just to make sure: have you verified we don't have a runtime regression from this change.  I recall in the past Paul Mineiro switched to using some Intel SIMD instructions and we got a pretty significant degradation in speed.  Granted, it may have been a different change. Thanks!

bscottm:
I haven't been particularly rigorous about collecting LDA SIMD vs. serial timing, although I did notice that I went from ~11 seconds for the entire test suite to ~9 seconds.

Is there a reasonable "soak test" for LDA?

Regression: There's little or no checking to ensure that the underlying arrays are vector aligned, so the compilers have to be pessimistic about loads and stores. That's were I'd put my money if I were to guess the origins of any SIMD-related regressions.

My patch shouldn't be quite as pessimistic if "-msse4.2" is passed to gcc. There's more XMMS register pressure, but that's not as awful as memory pressure from unnecessary spills to convert from __m128 to an array of floats.

bscottm:
There are a few more tweaks I could look into that deal with alignment. And one more union to remove.

BTW: The SIMD code is enabled by default in both the pre- and post- patch code. If there are any regressions, they'll show up fairly quickly.

bscottm:
Ariel: I've hacked the code so that all three (fast approx, accurate and SIMD) can be chosen at runtime, with a little help from templates. No more ifdef-ing. All code is inlined. Added a flag, "--math-mode <fast-approx|approx|accurate|accuracy|simd" to choose which math is used by the LDA core code (could probably be generalized across all of VW.)

Interesting regression, but I'm not sure how far back it goes: SIMD and Boost functions produce the same output, modulo 1e-4/1e-5 precision errors. The fast approximations are off by a factor of 2. Am investigating.

arielf:
Two small, perhaps obvious comments:

Timing on the whole test-suite is misleading.  The more "precision differences ignored" lines you have, the slower the run-time because the ~10x slower fuzzy-FP match is skipped if the original diff succeeds.   The best test for speed of a certain test is to `time` an individual run.

I'm fine with converting `ifdef`'s to runtime switches where the benefit is greater than the increase in complexity and code size.  These routines are very small but (for any architechure) one of the options is faster so it should be the default without burdening the user with the additional complexity.

In any case, many thanks for doing all this research on what is really the fastest, and for all the other deep dives into the code trying to make it better.

JohnLangford:
Paul, can you take a look at this in detail?

-John

bscottm:
@John, @pmineiro, @arielf:

I pushed a newer version of the code that tidies the SIMD refactoring up. I've found that 'git diff -w' is the best way to compare the versions, since there are a lot of white space changes.

You can select between the different modes via the command line flag "--math-mode [fast-approx|approx|precise|simd]", with SIMD being the default. They produce similar results, although the test case output is likely to need regenerating if this is incorporated into the release.

At some point in the near future (next two weeks), I'll go and regenerate the test output for the three different modes and corresponding outputs.

bscottm:
@JohnLangford @pmineiro @arielf:

Looks like I managed to get past the Boost.Alignment issue. Test is failing because the output needs regeneration. Please have a look at the build output and let me know how to proceed.

arielf:
Hi Scott,

Thanks again for the work.

Looking at the travis failure:  https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/55075029

Test 17 `diff -u --minimal train-sets/ref/wiki1K.stderr stderr.tmp`
is failing due to a significant difference in output.

This is not a floaing-point precision difference but something fundamentally different.

e.g: average loss after 256 examples is much lower in the new code.

This may actually be good (lower loss) but it needs to be verified that the current implementation is the correct one.  It looks to me that one of the implementations must be incorrect given the big difference in
output.  The incorrect one _may_ be the existing (old) one, meaning we had some accuracy bug all along, but this is less likely.

The question is which of the 3 implementations USE_SIMD, USE_PRECISE, USE_FAST_APPROX is causing such a big diff and is it actually correct.

arielf:
Sorry, I realize I may not have been clear enough.  I understand this change switches from the present default (fast approx?) to using SIMD.  The main point is I don't expect such a change to have such a *large* effect on the output.  Ergo there has to be a bug somewhere.  The main purpose of `make test` is to catch these cases which is what happened here.  Changing the reference output is easy, but is it the correct fix?

bscottm:
At least it's a consistent "feature":

```
$ ../vowpalwabbit/vw -k --lda 100 --lda_alpha 0.01 --lda_rho 0.01 --lda_D 1000 -l 1 -b 13 --minibatch 128 --math-mode simd /Users/michel/play/vowpal_wabbit/test/train-sets/wiki256.dat
Num weight bits = 13
learning rate = 1
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = /Users/michel/play/vowpal_wabbit/test/train-sets/wiki256.dat
num sources = 1
10.149698 10.149698            1            1.0  unknown   0.0000      732
10.369990 10.590282            2            2.0  unknown   0.0000       27
10.325998 10.282005            4            4.0  unknown   0.0000       53
10.401808 10.477618            8            8.0  unknown   0.0000       60
10.356298 10.310789           16           16.0  unknown   0.0000       26
10.473010 10.589722           32           32.0  unknown   0.0000      125
10.474932 10.476854           64           64.0  unknown   0.0000      313
10.425374 10.375817          128          128.0  unknown   0.0000       50
9.620762 8.816150          256          256.0  unknown   0.0000       33

finished run
number of examples = 256
weighted example sum = 256.000000
weighted label sum = 0.000000
average loss = 9.620762
total feature number = 22158
```

```
$ ../vowpalwabbit/vw -k --lda 100 --lda_alpha 0.01 --lda_rho 0.01 --lda_D 1000 -l 1 -b 13 --minibatch 128 --math-mode fast-approx /Users/michel/play/vowpal_wabbit/test/train-sets/wiki256.datNum weight bits = 13
learning rate = 1
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = /Users/michel/play/vowpal_wabbit/test/train-sets/wiki256.dat
num sources = 1
10.149614 10.149614            1            1.0  unknown   0.0000      732
10.369901 10.590187            2            2.0  unknown   0.0000       27
10.325905 10.281909            4            4.0  unknown   0.0000       53
10.401716 10.477528            8            8.0  unknown   0.0000       60
10.356205 10.310694           16           16.0  unknown   0.0000       26
10.472919 10.589634           32           32.0  unknown   0.0000      125
10.474844 10.476768           64           64.0  unknown   0.0000      313
10.425280 10.375717          128          128.0  unknown   0.0000       50
9.620675 8.816070          256          256.0  unknown   0.0000       33

finished run
number of examples = 256
weighted example sum = 256.000000
weighted label sum = 0.000000
average loss = 9.620675
total feature number = 22158
```

```
$ ../vowpalwabbit/vw -k --lda 100 --lda_alpha 0.01 --lda_rho 0.01 --lda_D 1000 -l 1 -b 13 --minibatch 128 --math-mode precise /Users/michel/play/vowpal_wabbit/test/train-sets/wiki256.datNum weight
Num weight bits = 13
learning rate = 1
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = /Users/michel/play/vowpal_wabbit/test/train-sets/wiki256.dat
num sources = 1
10.149390 10.149390            1            1.0  unknown   0.0000      732
10.369712 10.590034            2            2.0  unknown   0.0000       27
10.325692 10.281672            4            4.0  unknown   0.0000       53
10.401501 10.477310            8            8.0  unknown   0.0000       60
10.355986 10.310471           16           16.0  unknown   0.0000       26
10.472720 10.589454           32           32.0  unknown   0.0000      125
10.474631 10.476542           64           64.0  unknown   0.0000      313
10.425073 10.375516          128          128.0  unknown   0.0000       50
9.620526 8.815979          256          256.0  unknown   0.0000       33

finished run
number of examples = 256
weighted example sum = 256.000000
weighted label sum = 0.000000
average loss = 9.620526
total feature number = 22158
```

arielf:
Thanks!

This consistency lends credence to the hypothesis that the new implementations are correct, and we had a (much worse) implementation inaccuracy before.  The small differences (< 0.001) in FP output between the 3 implementations are about what I expect from "FP implementation differences".

I'm leaning towards fixing the reference output now :)

@pmineiro if you get a chance to confirm and add your vote, it would be great.

Thanks.

bscottm:
@arielf:

I was also more "stringent" about namespace qualification, especially when using functions in the std:: namespace. There are header files that have "using namespace std" but there is potential pollution in the source code that overrides and therefore hide functions that std:: provides. Don't ask me for a specific instance off the top of my head, though -- although some of the "mydigamma" and "mylgamma" preprocessor definitions come to mind.

I specifically qualifiy calls to std::exp, std::fmin and std::pow to ensure that the correct function is being called. Mostly out of paranoia -- I've been burned one too many times by namespace issues.

bscottm:
@arielf:

Actually, this may be the reason: The threshold for expdigammify/vexpdigammify and expdgammify_2/vexdigammify_2 are consistently 1.0e-6. The previous code used 1.0e-6 for expdigammify and expdigammify_2 (non-simd) vs. 1.0e-10 for vexpdigammify/vexpdigammify_2 (the simd code).

Not hard to tweak if there's an urgent desire to go back to the previous threshold (although that'd have to wait until week after next.)

pmineiro:
We should test on a large amount of input.  When I first implemented this way back when, I had underflow/overflow issues that only surfaced on larger data sets.  It's hard for me to tell from the diff :), but it'd be nice to get some confidence that we haven't reintroduced them somehow.

bscottm:
@pmineiro: Was that the reason for the fmin(1.e-10, ...) in vexpdigammify?

pmineiro:
@bscottm: details are hazy, but that sounds plausible.

bscottm:
@pmineiro: That explains the inconsistency. The fmax value for expdigammify_2 is 1e-6 and 1e-10 for vexpdigammify_2 in the old code. They are now both set to 1e-6.

Also, the diff is easier to read if you look at it using the unified view, not the split view.

bscottm:
Adjusted the underflow threshold to 1.0e-10. Test results align as expected on my simple local Mac. Let's see what Travis CI has to say… :-)

bscottm:
And that's a good sign. :-)

bscottm:
Just for completeness for @arielf:

```
$ ../vowpalwabbit/vw -k --lda 100 --lda_alpha 0.01 --lda_rho 0.01 --lda_D 1000 -l 1 -b 13 --minibatch 128 --math-mode simd ../test/train-sets/wiki256.dat
Num weight bits = 13
learning rate = 1
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = ../test/train-sets/wiki256.dat
num sources = 1
10.149310 10.149310            1            1.0  unknown   0.0000      732
10.369817 10.590323            2            2.0  unknown   0.0000       27
10.325925 10.282034            4            4.0  unknown   0.0000       53
10.401768 10.477611            8            8.0  unknown   0.0000       60
10.356288 10.310807           16           16.0  unknown   0.0000       26
10.472954 10.589620           32           32.0  unknown   0.0000      125
10.474859 10.476765           64           64.0  unknown   0.0000      313
10.425314 10.375769          128          128.0  unknown   0.0000       50
10.005566 9.585817          256          256.0  unknown   0.0000       33

finished run
number of examples = 256
weighted example sum = 256.000000
weighted label sum = 0.000000
average loss = 10.005566
total feature number = 22158
```

```
$ ../vowpalwabbit/vw -k --lda 100 --lda_alpha 0.01 --lda_rho 0.01 --lda_D 1000 -l 1 -b 13 --minibatch 128 --math-mode approx ../test/train-sets/wiki256.dat
Num weight bits = 13
learning rate = 1
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = ../test/train-sets/wiki256.dat
num sources = 1
10.149236 10.149236            1            1.0  unknown   0.0000      732
10.369734 10.590232            2            2.0  unknown   0.0000       27
10.325835 10.281937            4            4.0  unknown   0.0000       53
10.401677 10.477518            8            8.0  unknown   0.0000       60
10.356195 10.310714           16           16.0  unknown   0.0000       26
10.472864 10.589532           32           32.0  unknown   0.0000      125
10.474771 10.476678           64           64.0  unknown   0.0000      313
10.425222 10.375674          128          128.0  unknown   0.0000       50
10.005482 9.585742          256          256.0  unknown   0.0000       33

finished run
number of examples = 256
weighted example sum = 256.000000
weighted label sum = 0.000000
average loss = 10.005482
total feature number = 22158
```

```
$ ../vowpalwabbit/vw -k --lda 100 --lda_alpha 0.01 --lda_rho 0.01 --lda_D 1000 -l 1 -b 13 --minibatch 128 --math-mode precise ../test/train-sets/wiki256.dat
Num weight bits = 13
learning rate = 1
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = ../test/train-sets/wiki256.dat
num sources = 1
10.148996 10.148996            1            1.0  unknown   0.0000      732
10.369544 10.590092            2            2.0  unknown   0.0000       27
10.325634 10.281723            4            4.0  unknown   0.0000       53
10.401487 10.477340            8            8.0  unknown   0.0000       60
10.356004 10.310521           16           16.0  unknown   0.0000       26
10.472686 10.589368           32           32.0  unknown   0.0000      125
10.474590 10.476495           64           64.0  unknown   0.0000      313
10.425045 10.375500          128          128.0  unknown   0.0000       50
10.005359 9.585672          256          256.0  unknown   0.0000       33

finished run
number of examples = 256
weighted example sum = 256.000000
weighted label sum = 0.000000
average loss = 10.005359
total feature number = 22158
```

JohnLangford:
I merged this in.  

The one thing that I find funny here is the use of memcpy instead of union.  Is that really required?

bscottm:
@JohnLangford:

Just got back from an OCONUS trip and was out of touch for the duration. The memcpy is more intentional than the union and should tell future developers what's going on -- bit conversion from 32-bit floating point to 32-bit integer for the purposes of bit banging/munging. memcpy also avoids alignment issues: the union relies on integer and float alignment being the same, which can be a poor assumption.

gcc and clang's optimizers are pretty smart about reducing memcpy from a function call to an instruction or two. I'll post the assembly if that would help.

bscottm:
@JohnLangford: Been doing software development for a while and have seen many strange things, so I tend to code defensively.

BTW: There are other places in the lda_core.cc code where loops could be vectorized. I don't want to recreate what the Boost.SIMD library (if it ever comes to fruition), but there are other vectorization opportunities.

PR #556, opened 2015-03-13
Simple combinations, interactions greater than cubic & caching [fix 2,3,4]

trufanov-nok:
Referring https://github.com/JohnLangford/vowpal_wabbit/issues/510

Please don't merge - experimental feature, PR made mainly to start discussion regarding implementation details. It won't pass Travis CI.
I'll post a detailed description re implementation here in a few hours.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -32,6 +32,7 @@ noinst_HEADERS = vowpalwabbit/accumulate.h \
 	vowpalwabbit/ect.h \
 	vowpalwabbit/gd.h \
 	vowpalwabbit/gd_mf.h \
+	vowpalwabbit/interactions.h \
 	vowpalwabbit/kernel_svm.h \
 	vowpalwabbit/lda_core.h \
 	vowpalwabbit/log_multi.h \
--- a/vowpalwabbit/Makefile.am
+++ b/vowpalwabbit/Makefile.am
@@ -4,7 +4,7 @@ liballreduce_la_SOURCES = allreduce.cc
 
 bin_PROGRAMS = vw active_interactor
 
-libvw_la_SOURCES = hash.cc global_data.cc io_buf.cc parse_regressor.cc parse_primitives.cc unique_sort.cc cache.cc rand48.cc simple_label.cc multiclass.cc oaa.cc multilabel_oaa.cc ect.cc autolink.cc binary.cc lrq.cc cost_sensitive.cc multilabel.cc csoaa.cc cb.cc cb_algs.cc search.cc search_sequencetask.cc search_dep_parser.cc search_hooktask.cc search_multiclasstask.cc search_entityrelationtask.cc search_graph.cc parse_example.cc scorer.cc network.cc parse_args.cc accumulate.cc gd.cc learner.cc lda_core.cc gd_mf.cc mf.cc bfgs.cc noop.cc print.cc example.cc parser.cc loss_functions.cc sender.cc nn.cc bs.cc cbify.cc topk.cc stagewise_poly.cc log_multi.cc active.cc kernel_svm.cc best_constant.cc ftrl_proximal.cc
+libvw_la_SOURCES = hash.cc global_data.cc io_buf.cc parse_regressor.cc parse_primitives.cc unique_sort.cc cache.cc rand48.cc simple_label.cc multiclass.cc oaa.cc multilabel_oaa.cc ect.cc autolink.cc binary.cc lrq.cc cost_sensitive.cc multilabel.cc csoaa.cc cb.cc cb_algs.cc search.cc search_sequencetask.cc search_dep_parser.cc search_hooktask.cc search_multiclasstask.cc search_entityrelationtask.cc search_graph.cc parse_example.cc scorer.cc network.cc parse_args.cc accumulate.cc gd.cc interactions.cc learner.cc lda_core.cc gd_mf.cc mf.cc bfgs.cc noop.cc print.cc example.cc parser.cc loss_functions.cc sender.cc nn.cc bs.cc cbify.cc topk.cc stagewise_poly.cc log_multi.cc active.cc kernel_svm.cc best_constant.cc ftrl_proximal.cc
 
 libvw_c_wrapper_la_SOURCES = vwdll.cpp
 
--- a/vowpalwabbit/example.h
+++ b/vowpalwabbit/example.h
@@ -22,6 +22,7 @@ const size_t affix_namespace     = 132;   // this is \x84
 const size_t spelling_namespace  = 133;   // this is \x85
 const size_t conditioning_namespace = 134;
 const size_t dictionary_namespace  = 135;
+const size_t interactions_namespace  = 136; // stores generated fatures if all.cache_interactions is enabled
 
 struct feature {
   float x;
--- a/vowpalwabbit/example.h
+++ b/vowpalwabbit/example.h
@@ -116,3 +116,8 @@ inline int example_is_newline(example& ec)
           ((ec.indices.size() == 1) &&
            (ec.indices.last() == constant_namespace)));
 }
+
+inline bool valid_ns(char c)
+{
+  return !(c == '|' || c == ':');
+}
--- a/vowpalwabbit/gd.h
+++ b/vowpalwabbit/gd.h
@@ -10,6 +10,8 @@
 
 #include "parse_regressor.h"
 #include "constant.h"
+#include "interactions.h"
+
 
 namespace GD{
   LEARNER::base_learner* setup(vw& all);
@@ -21,59 +23,64 @@ namespace GD{
   void save_load_regressor(vw& all, io_buf& model_file, bool read, bool text);
   void save_load_online_state(vw& all, io_buf& model_file, bool read, bool text, GD::gd *g = NULL);
 
+
   // iterate through one namespace (or its part), callback function T(some_data_R, feature_value_x, feature_weight)
   template <class R, void (*T)(R&, const float, float&)>
   inline void foreach_feature(weight* weight_vector, size_t weight_mask, feature* begin, feature* end, R& dat, uint32_t offset=0, float mult=1.)
   {
-    for (feature* f = begin; f!= end; f++)
+    for (feature* f = begin; f!= end; ++f)
       T(dat, mult*f->x, weight_vector[(f->weight_index + offset) & weight_mask]);
   }
 
   // iterate through one namespace (or its part), callback function T(some_data_R, feature_value_x, feature_index)
   template <class R, void (*T)(R&, float, uint32_t)>
-   void foreach_feature(weight* weight_vector, size_t weight_mask, feature* begin, feature* end, R&dat, uint32_t offset=0, float mult=1.)
+   void foreach_feature(weight* /*weight_vector*/, size_t /*weight_mask*/, feature* begin, feature* end, R&dat, uint32_t offset=0, float mult=1.)
    {
-     for (feature* f = begin; f!= end; f++)
+     for (feature* f = begin; f!= end; ++f)
        T(dat, mult*f->x, f->weight_index + offset);
    }
  
-  // iterate through all namespaces and quadratic&cubic features, callback function T(some_data_R, feature_value_x, S)
-  // where S is EITHER float& feature_weight OR uint32_t feature_index
-  template <class R, class S, void (*T)(R&, float, S)>
-  inline void foreach_feature(vw& all, example& ec, R& dat)
-  {
-    uint32_t offset = ec.ft_offset;
+   // iterate through all namespaces and quadratic&cubic features, callback function T(some_data_R, feature_value_x, S)
+   // where S is EITHER float& feature_weight OR uint32_t feature_index
+   template <class R, class S, void (*T)(R&, float, S)>
+   inline void foreach_feature(vw& all, example& ec, R& dat)
+   {
+       uint32_t offset = ec.ft_offset;
 
-    for (unsigned char* i = ec.indices.begin; i != ec.indices.end; i++)
-      foreach_feature<R,T>(all.reg.weight_vector, all.reg.weight_mask, ec.atomics[*i].begin, ec.atomics[*i].end, dat, offset);
-     
-    for (vector<string>::iterator i = all.pairs.begin(); i != all.pairs.end();i++) {
-      if (ec.atomics[(unsigned char)(*i)[0]].size() > 0) {
-        v_array<feature> temp = ec.atomics[(unsigned char)(*i)[0]];
-        for (; temp.begin != temp.end; temp.begin++)
-        {
-          uint32_t halfhash = quadratic_constant * (temp.begin->weight_index + offset);
-       
-          foreach_feature<R,T>(all.reg.weight_vector, all.reg.weight_mask, ec.atomics[(unsigned char)(*i)[1]].begin, ec.atomics[(unsigned char)(*i)[1]].end, dat, 
-                               halfhash, temp.begin->x);
-        }
-      }
-    }
-     
-    for (vector<string>::iterator i = all.triples.begin(); i != all.triples.end();i++) {
-      if ((ec.atomics[(unsigned char)(*i)[0]].size() == 0) || (ec.atomics[(unsigned char)(*i)[1]].size() == 0) || (ec.atomics[(unsigned char)(*i)[2]].size() == 0)) { continue; }
-      v_array<feature> temp1 = ec.atomics[(unsigned char)(*i)[0]];
-      for (; temp1.begin != temp1.end; temp1.begin++) {
-        v_array<feature> temp2 = ec.atomics[(unsigned char)(*i)[1]];
-        for (; temp2.begin != temp2.end; temp2.begin++) {
-           
-          uint32_t halfhash = cubic_constant2 * (cubic_constant * (temp1.begin->weight_index + offset) + temp2.begin->weight_index + offset);
-          float mult = temp1.begin->x * temp2.begin->x;
-          foreach_feature<R,T>(all.reg.weight_vector, all.reg.weight_mask, ec.atomics[(unsigned char)(*i)[2]].begin, ec.atomics[(unsigned char)(*i)[2]].end, dat, halfhash, mult);
-        }
-      }
-    }
-  }
+       for (unsigned char* i = ec.indices.begin; i != ec.indices.end; ++i)
+           foreach_feature<R,T>(all.reg.weight_vector, all.reg.weight_mask, ec.atomics[*i].begin, ec.atomics[*i].end, dat, offset);
+
+#ifndef USE_INTERACTIONS
+//if USE_INTERACTIONS is defined - the pairs and triples have already been generated and stored in atomics[interactions_namespace]
+
+       for (vector<string>::iterator i = all.pairs.begin(); i != all.pairs.end();i++) {
+           if (ec.atomics[(unsigned char)(*i)[0]].size() > 0) {
+               v_array<feature> temp = ec.atomics[(unsigned char)(*i)[0]];
+               for (; temp.begin != temp.end; temp.begin++)
+               {
+                   uint32_t halfhash = quadratic_constant * (temp.begin->weight_index + offset);
+
+                   foreach_feature<R,T>(all.reg.weight_vector, all.reg.weight_mask, ec.atomics[(unsigned char)(*i)[1]].begin, ec.atomics[(unsigned char)(*i)[1]].end, dat,
+                           halfhash, temp.begin->x);
+               }
+           }
+       }
+
+       for (vector<string>::iterator i = all.triples.begin(); i != all.triples.end();i++) {
+           if ((ec.atomics[(unsigned char)(*i)[0]].size() == 0) || (ec.atomics[(unsigned char)(*i)[1]].size() == 0) || (ec.atomics[(unsigned char)(*i)[2]].size() == 0)) { continue; }
+           v_array<feature> temp1 = ec.atomics[(unsigned char)(*i)[0]];
+           for (; temp1.begin != temp1.end; temp1.begin++) {
+               v_array<feature> temp2 = ec.atomics[(unsigned char)(*i)[1]];
+               for (; temp2.begin != temp2.end; temp2.begin++) {
+
+                   uint32_t halfhash = cubic_constant2 * (cubic_constant * (temp1.begin->weight_index + offset) + temp2.begin->weight_index + offset);
+                   float mult = temp1.begin->x * temp2.begin->x;
+                   foreach_feature<R,T>(all.reg.weight_vector, all.reg.weight_mask, ec.atomics[(unsigned char)(*i)[2]].begin, ec.atomics[(unsigned char)(*i)[2]].end, dat, halfhash, mult);
+               }
+           }
+       }
+#endif
+   }
 
   // iterate through all namespaces and quadratic&cubic features, callback function T(some_data_R, feature_value_x, feature_weight)
   template <class R, void (*T)(R&, float, float&)>
--- a/vowpalwabbit/gd.h
+++ b/vowpalwabbit/gd.h
@@ -49,37 +49,6 @@ namespace GD{
 
        for (unsigned char* i = ec.indices.begin; i != ec.indices.end; ++i)
            foreach_feature<R,T>(all.reg.weight_vector, all.reg.weight_mask, ec.atomics[*i].begin, ec.atomics[*i].end, dat, offset);
-
-#ifndef USE_INTERACTIONS
-//if USE_INTERACTIONS is defined - the pairs and triples have already been generated and stored in atomics[interactions_namespace]
-
-       for (vector<string>::iterator i = all.pairs.begin(); i != all.pairs.end();i++) {
-           if (ec.atomics[(unsigned char)(*i)[0]].size() > 0) {
-               v_array<feature> temp = ec.atomics[(unsigned char)(*i)[0]];
-               for (; temp.begin != temp.end; temp.begin++)
-               {
-                   uint32_t halfhash = quadratic_constant * (temp.begin->weight_index + offset);
-
-                   foreach_feature<R,T>(all.reg.weight_vector, all.reg.weight_mask, ec.atomics[(unsigned char)(*i)[1]].begin, ec.atomics[(unsigned char)(*i)[1]].end, dat,
-                           halfhash, temp.begin->x);
-               }
-           }
-       }
-
-       for (vector<string>::iterator i = all.triples.begin(); i != all.triples.end();i++) {
-           if ((ec.atomics[(unsigned char)(*i)[0]].size() == 0) || (ec.atomics[(unsigned char)(*i)[1]].size() == 0) || (ec.atomics[(unsigned char)(*i)[2]].size() == 0)) { continue; }
-           v_array<feature> temp1 = ec.atomics[(unsigned char)(*i)[0]];
-           for (; temp1.begin != temp1.end; temp1.begin++) {
-               v_array<feature> temp2 = ec.atomics[(unsigned char)(*i)[1]];
-               for (; temp2.begin != temp2.end; temp2.begin++) {
-
-                   uint32_t halfhash = cubic_constant2 * (cubic_constant * (temp1.begin->weight_index + offset) + temp2.begin->weight_index + offset);
-                   float mult = temp1.begin->x * temp2.begin->x;
-                   foreach_feature<R,T>(all.reg.weight_vector, all.reg.weight_mask, ec.atomics[(unsigned char)(*i)[2]].begin, ec.atomics[(unsigned char)(*i)[2]].end, dat, halfhash, mult);
-               }
-           }
-       }
-#endif
    }
 
   // iterate through all namespaces and quadratic&cubic features, callback function T(some_data_R, feature_value_x, feature_weight)
--- a/vowpalwabbit/gd.h
+++ b/vowpalwabbit/gd.h
@@ -49,6 +49,8 @@ namespace GD{
 
        for (unsigned char* i = ec.indices.begin; i != ec.indices.end; ++i)
            foreach_feature<R,T>(all.reg.weight_vector, all.reg.weight_mask, ec.atomics[*i].begin, ec.atomics[*i].end, dat, offset);
+
+       INTERACTIONS::generate_interactions<R,S,T>(all, ec, dat);
    }
 
   // iterate through all namespaces and quadratic&cubic features, callback function T(some_data_R, feature_value_x, feature_weight)
--- a/vowpalwabbit/global_data.cc
+++ b/vowpalwabbit/global_data.cc
@@ -340,6 +340,8 @@ vw::vw()
       spelling_features[i] = 0;
     }
 
+  interactions = v_init<v_string>();
+
   //by default use invariant normalized adaptive updates
   adaptive = true;
   normalized_updates = true;
--- a/vowpalwabbit/global_data.h
+++ b/vowpalwabbit/global_data.h
@@ -383,6 +383,8 @@ struct vw {
   size_t numpasses;
   size_t passes_complete;
   size_t parse_mask; // 1 << num_bits -1
+  bool permutations;
+  std::vector<std::string> interactions; // interactions of namespaces to cross.
   std::vector<std::string> pairs; // pairs of features to cross.
   std::vector<std::string> triples; // triples of features to cross.
   bool ignore_some;
@@ -391,8 +393,8 @@ struct vw {
   bool redefine_some;          // --redefine param was used
   unsigned char redefine[256]; // keeps new chars for amespaces
 
-  std::vector<std::string> ngram_strings; // pairs of features to cross.
-  std::vector<std::string> skip_strings; // triples of features to cross.
+  std::vector<std::string> ngram_strings;
+  std::vector<std::string> skip_strings;
   uint32_t ngram[256];//ngrams to generate.
   uint32_t skips[256];//skips in ngrams.
   std::vector<std::string> limit_strings; // descriptor of feature limits
--- a/vowpalwabbit/global_data.h
+++ b/vowpalwabbit/global_data.h
@@ -384,7 +384,7 @@ struct vw {
   size_t passes_complete;
   size_t parse_mask; // 1 << num_bits -1
   bool permutations;
-  std::vector<std::string> interactions; // interactions of namespaces to cross.
+  v_array<v_string> interactions; // interactions of namespaces to cross.
   std::vector<std::string> pairs; // pairs of features to cross.
   std::vector<std::string> triples; // triples of features to cross.
   bool ignore_some;
--- a/vowpalwabbit/interactions.cc
+++ b/vowpalwabbit/interactions.cc
@@ -233,4 +233,74 @@ void generate_interactions(vw& all, example& ec)
 
 }
 
+//some helper functions
+
+inline bool valid_ns(char c)
+{
+    return !(c == '|' || c == ':');
+}
+
+// expand namespace interactions if contain wildcards
+
+void expand_namespace_depth(string& ns, vector<string>& res, string val,  size_t pos)
+{
+    assert (pos <= ns.length());
+
+    if (pos == ns.length())
+    {
+        res.push_back(val);
+    }
+    else
+        if (ns[pos] != ':')
+        {
+            val.push_back(ns[pos]);
+            expand_namespace_depth(ns, res, val, pos+1);
+        }
+        else
+        {
+            res.reserve(res.size() + valid_ns_size);
+            for (unsigned char j = printable_start; j <= printable_end; j++)
+            {
+                if(valid_ns(j))
+                {
+                    val.push_back(j);
+                    expand_namespace_depth(ns, res, val, pos+1);
+                    val.pop_back();
+                }
+            }
+        }
+}
+
+bool compare_interactions (string a, string b) {
+    std::sort(a.begin(), a.end());
+    std::sort(b.begin(), b.end());
+    return (a < b);
+}
+
+bool compare_sorted_interactions (string a, string b) {
+    return (a < b);
+}
+
+void filter_duplicate_interactions(vector<string>& vec, bool leave_elements_sorted)
+{
+    bool(*fn_cmp)(string,string);
+    if (leave_elements_sorted)
+    {
+        for (vector<string>::iterator i = vec.begin(); i != vec.end(); ++i)
+        {  // sort namespaces in each interaction ascendantly to group equal namespaces in them
+            string& s = *i;
+            std::sort(s.begin(), s.end());
+        }
+        fn_cmp = compare_sorted_interactions;
+    } else
+        fn_cmp = compare_interactions;
+    // can't use unique in case of !
+    // so use fastest std::set (http://stackoverflow.com/a/1041939/841424) with proper comparision function
+    set<string, bool(*)(string,string)> temp_set(fn_cmp);
+    size_t size = vec.size();
+    for ( size_t i = 0; i < size; ++i )
+        temp_set.insert( vec[i] );
+    vec.assign( temp_set.begin(), temp_set.end() );
+}
+
 }
--- a/vowpalwabbit/interactions.cc
+++ b/vowpalwabbit/interactions.cc
@@ -1,0 +1,228 @@
+#include "interactions.h"
+
+namespace INTERACTIONS
+{
+
+// generates features for all.interations and store them in cache ( ec.atomics[interactions_namespace] )
+void generate_interactions(vw& all, example& ec)
+{
+  if (!all.interactions.empty() && !v_array_contains(ec.indices, (unsigned char)interactions_namespace) )
+    {   // if cache wasn't created yet
+
+        v_array<feature>& cache = ec.atomics[interactions_namespace];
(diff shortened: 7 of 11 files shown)


trufanov-nok:
Well' that will be a complicated one.

## The concept
As John correctly pointed out: the implementation is crucial for fix 2 [simple combinations] affects 3 and 4, so I've decided to try to make all of them as a whole.

My initial idea was to create `gd.h`-like module with own `foreach_feature()` implementation. There should be also optional functionality (switched on by default) for storing generated features in a special namespace instead of immediately calling function on them (caching, described in proposal 3). Plus new parameter `--interactions` for passing interaction of any length including levels greater than cubic.

Unfortunately I've run into trouble with this initial concept. The reason is the existence of one special case noted by Martin in our yahoo discussion: the interaction of a feature with itself has sense sometimes: if its weight is other than 1. Such features might model non-linearity. Thus I decided to implement not just the simple combinations of features from namespaces, but simple combinations + this special case. In it's turn this makes impossible to estimate number of new features which will be generated by particular interaction (like it's done in `parser.cc::setup_example``) without iterating them and checking their weights. In other words If we have interaction `-q aa` with `n` features in `a` I can't say how many new features will be generated without preliminary counting all features with weights != 1. in `a`.

As I wanted to make my code as fast as possible I didn't like the idea of iterating all features in each interaction for every example. Thus I've decided to do the following trick:

I've decided to implement proposal 3 (caching of generated features) as non-optional. This allows to separate `fereach feature` iterators code from *new features generation* code bcs we can use cache to pass results of second to the first. So wrote a function that just generate features from interactions and stores them in the cache namespace. In a process it set ups corect number of features and sum of their weights for example. It shall be called only once just after parsing the example. Then we just need to say `gd::foreach_feature` that it shouldn't process `all.pairs` and `all.triples` as they are already generated and stored these features as atomic features.

## Rough performance estimations
I also didn't reject the idea to do this with support of interactions longer than cubic. So the function made with a big non-recursive loop inside. And still it was slower than original code. John was right and cache by itself doesn't give you some performance handicap that you can spend on other useful things and still be on par with master branch performance. Another challenge for performance was caused by my intention to make generated feature hashes backward compatible with original gd.h's code. E.g. pairs shall use hash function for pairs, triples - old hash code for triples, and for longer interactions I can use my own. I did that too.
Initially performance of my code was 10-20% slower than VW's master code. After many hours of tweaking I could improve it. I didn't made 100% reliable tests yet, but my expectation that it's now ~5% slower in user mode and ~5 % faster in real mode (clock time). By *user* and *real*  I mean measurements printed out by unix' `time` command. And so I can assume my implementation after all tweaks still takes more CPU time but it seems to be better paralleled and works a bit faster or on par with vanilla VW if you have  CPU with several cores. I have 4 cores in my laptop. VW usually takes 25% of CPU which is exactly one core. With my implementation it consumes 29-30% of CPU. Which is good, but I have no clue which piece of code is paralleled exactly.

TBC

TBC

trufanov-nok:
Implementation could be split in 2 parts: the code to process and store interaction values (`--interaction` param) of any length and the feature generation code.

## Interactions processor. (proposal 4)

Ok, I decided to work with interactions of any length. Current `all.pairs` and `all.triples` work with length 2 and 3 respectively. Of course, I realised that I can't remove these 2 arrays without breaking almost every module in VW. So my idea was to add a new vector `all.interactions` and store all interactions in it. In this case `pairs` and `triples` could be just subsets of its content. Unfortunately, I didn't find a good way to do that and left it for future. Currently `all.interactions` contains copies of content of other two arrays and vice versa.

I also wrote a small recursive function to expand any number of wildcards `:` in string of any length.

So, what do we have now? We have 3 arrays  `all.pairs` , `all.triples` and `all.interactions`. They get the data passed by `-q`, `--cubic` and `--interactions`. Only last one accepts strings of arbitrary length (but > 2). Then all data in these 3 arrays are expanded if needed (wildcards are replaced with all printed characters). Then all strings with length 2 from `interactions` are added to `pairs`. Strings with length 3 are added from `interactions` to `triples`. And original data from `pairs` and `triples` is copied to `interactions`.
Finally `interactions` contains everything and other two arrays only content of required length.
Now if simple combinations mode is enabled (it's so, as by default `all.permutations == false`) the letters in each string in `interactions` are sorted ascendently. This helps to find duplicate namespaces in interaction with one pass.

After that I've additionally remove all duplicates from `interactions` that might be generated for example by `-q ::`. Feature generation code takes care of *feature* interactions duplicates *inside* interaction but not duplicates of interactions itself. *I would like to ask your opinion if this shall be scaled to `al.permutations` mode too?* Right now in vanilla VW the argument `-q ::` will give you interactions `ab` and `ba`. This is not `ab` and `ab` as in case of simple combinations after letter sorting, but still seems pretty useless for me.

That's all. We'll generate features based on `all.interactions` and won't work with `pairs` and `triples` at all. Other modules (except gd.h, which we'll modify) may continue to work with `pairs` and `triples` as usual. That works.

Notes: 
1. Ideally data shouldn't be duplicated and `pairs` & `triples` shall contains references to `interactions`. Or something like that. That's possible with `v_array<>::begin/end` if content of `interactions` will be sorted by it's length.
2. I don't like current approach of expanding wildcards. Even `--cubic :::` will generate s lot of triples and takes a few minutes to complete. We might think about something smarter later.

The described functionality is implemented in parse_arguments.cc and global_data.h (array declaration). I've also deleted two wrong comments in the global_data.h by the go.

TBC

trufanov-nok:
## Feature generator
# Generator incorporation
For simplicity let's assume we've already implemented function ``void generate_interactions(vw& all, example& ec)`` in `interactions.cc` and discuss it's incorporation into VW code. The tricky thing here that we should not just insert it's call somewhere but disable few other functions from being called. There are few ways to do that. Currently I've used `#define USE_INTERACTIONS ` keyword to exclude functionality replaced by feature generator, but it could also be a static variable or constant evaluated at runtime or even VW param accessible throw command line arguments. *How do you think what is the best?*
Then I write about excluded functionality I mean 2 pieces of code: 
1. Estimation of `num_features` and `total_sum_feat_sq` in `parser.cc::setup_example()`. This now is done inside generator. In fact, this is a right place for generator function to be invoked.
2. Everything except iteration throw atomic features in `gd.h::foreach_feature`. We have stored features in cache namespace, so no need iterate `pairs` or `triples` anymore. Most of `gd.h` content is highlighted in github comparison window - don't believe it ). There are only few lines of code and comments.

NOTE:
a. Currently `USE_INTERACTIONS ` is defined in `interactions.h` which might be not a best choice. Once you include header file you're getting this in effect. *Any better ideas for its placement?*
b. I've made a few more changes in gd.h - replaced postfix increment of iterators (`for(;;f++)`) with prefix increment (`for(;;++f)`) as recommended for performance. Plus commented unused function arguments.

# Generator implementation.

First of all, I've defined `bool permutations` in `global_data.h` which is initialised in `parse_args.cc`. It's `false` by default. Then I've added declaration of a new `interactions_namespace = 136` in `examples.h`
After few attempts I admit that such function can't be inline, so I made 2 files for my code: `interactions.h` and `interactions.cc`. References to both are added to `Makefile.am` and `vowpalwabbit\Makefile.am`. *I'm not sure if  `interactions.h` is in a right place of `Makefile.am`. * And seems I've forgotten to add .cc file to VS project file. Will do later.

As for function itself - it's pretty well commented. It uses non-recursive algorithm to iterate features in namespaces. state is saved in array of `feature_gen_data`. Each `feature_gen_data`contains metadata for one inner loop. It also exploits the fact that namespaces are grouped and same namespaces are always neighbours. 

# Hashing
There are old hash algorithms support. I've chosen variation of Bernstein hash suggested by Josh Bloch in Effective Java http://stackoverflow.com/a/1646913/841424 for interactions longer than cubic. Mostly bcs it's very similar to algorithms used for pairs and cubic. I also prefer algorythm where constant isn't depend on number of hashed integers like in cubic hash. It was not easy to integrate it efficiently. Current implementation of these 3 hashes are heuristically chosen as fastest for my GCC compiler. 

# Performance optimisation

 spend a lot of time trying different performance optimisation tricks. Many of them make things worse and were removed. Some were pretty useless, but still in the code, as they at least shouldn't harm the performance. And few listed below were useful. (I'm not talking now about changes in algorithms or code logic - only over-optimisation tricks).
1. I've moved a block of some variable declarations from loop to parent scope. And also found that these variables mostly constant for `all.permutations == true`.
2. I've found with Valgrind that accessing to `std::string` has significant cost. It was converted to `const string` before calling `operator[]()`. http://stackoverflow.com/a/19920071/841424 Not sure how much I gained with this trick. But I would recommend to consider it, as VW has non-constant strings in all.pairs and a..triples too. Perhaps they better be stored as `const std::string`. Perhaps replacement with `v_array<unsigned char>` would be even better.
3. fields in `struct feature_gen_data` are all length 8.
4. I would like to mention that all my attempts to store something as static object to avoid unnecessary memory allocations lead to decrease in performance. Perhaps that prevented some parallelisation hidden from my eyes.

# Why make test and Travis CI test fails?

There are 4 errors in make test. The output could be found here: http://pastebin.com/jku1nxAe
This was obtained by running `RunTests.pl` without `-e` param.
The first 3 errors most probably caused by the fact, that VW now works in simple combinations mode by default. The average losses for them are better. I can adopt test results for new functionality. But there is 4th error in `Test case 59`. Frankly, I don't understand what is going on in this test. I found that some VW modes are sensitive to the number of `ec.indices` and use them for own purpose. This case seems to manipulate with interaction pairs and numbers of generated features in it's code. *I need an advice on what is it and how it's better to fix.*

I think that's all for this PR.

JohnLangford:
You have clearly put quite a bit of work into this---I'm impressed.

Questions first:
1. Having -q :: do 'ab' and 'ba' seems undesirable.

2. Interactions should always be enabled if they are supported. In other 
words, no #define USE_INTERACTIONS  This implies that interactions must 
subsume -q and --cubic.

3. I'd suggest removing the #define entirely.

Reading through the code here's a core dump of thoughts.

(1) push_back always makes sure there is enough memory, so you do not 
need ensure_could_push.

(2) The feature generation becomes quite expensive in the --cubic ::: 
case, because it must check each combination even if most are never 
used.  There is a faster approach.  In essence, you could preprocess to 
create an array of size 256, each entry of which contains either nothing 
(signifying the end of an interaction) or an array of size 256 
(signifying the next element of the interaction), recursing this 
appropriately to handle all interactions.

When an example is encountered, starting with a sorted list of indicies, 
you can look up the first index: empty -> no interaction, something -> 
interaction with another element so loop other elements empty -> no 
interaction, something -> nontrivial interaction so loop again with 
empty-> end interaction while something -> a higher order interaction...

(3) I'd generally like to _avoid_ adding stuff to global.h, because 
global.h needs to shrink.  An alternative would be making interactions 
be a reduction element.  This would shift your computation from 
parser.cc to the learning thread.  I'm unclear on whether this is 
helpful or harmful computationally.

(4) I'm not a fan of having a dozen different ways to interact 
features.  I'd much rather have one approach which we know is right and 
works well.  I'm skeptical about the Bernstein hash, because it does not 
take into account the fact that many of our features have small 
values.   The interaction function in stagewise_poly.cc is probably the 
most tested one that we have at the moment.

(5) Some of our base learning algorithms (gd_mf) and reductions (lrq, 
mf, and some learning to search tasks) require knowledge of namespaces 
because they do something different than a complete expansion.  What to 
do?   This suggests that we should expand the interactions lower in the 
reduction stack near the base learning algorithms _or_ do this is as an 
actual cache rather than a preprocess step.  Imagine defining a 
foreach_feature which _either_ looped through the cache _or_ created a 
cache if one did not exist.  Is there a reason why this could not work well?

(6) Test 59 failing suggests that there is some effective change in the 
library interface.

Your thoughts?

-John

trufanov-nok:
(1). Ok it's doable.
(2&3). I'm not sure I got your point. The core idea of this implementation of interactions is to replace `-q` and `--cubic` with more generic function instead of adding one more function for interactions higher than cubic. Otherwise I would need to implement 'simple combinations' enhancement in all 3 functions. Thus old code never invoked even there is no `--interactions` specified and only old arguments `-q` or `--cubic` are used. Thus usage of #define has sense: we don't have useless code in binary, still one can roll back to old codebase by recompiling VW without it. The alternative could be #define removal and boolean flag in VW class to compile both implementations and switch between them without recompilation

(1). It's just over-optimization. If you push elements into an empty array one by one then your vector will call realloc to allocate memory for 5, 13, 29, 61, 125 elements etc. I wanted to jump over first N allocations in case I know M elements shall be pushed into array. For example - jump over first 4 allocations if you need to push 100 elements.  
I don't want to use just `resize(100)` as it sets **exact ** size of allocated memory even there  is enough already. Also there might be more push_back() calls in next iteration. This means. every call to `resize()` will result in a call to `realloc()`. Technically, `ensure_could_push()` is equal to `resize()` with
 ``if ((size_t)(end_array-begin) < length)`` 
instead of
 ``if ((size_t)(end_array-begin) != length)`` 
on [line 37](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/v_array.h#L37) In other words, `push_back` is intended for pushing elements one by one, `resize()` for pushing total N elements and `ensure_could_push()` intended for pushing chunks of elements without knowing total N.  
On practice this could result in better performance for namespaces with big number of features. I thought about case `--redefine f:=: -q ff` while doing that. But I don't think any significant changes in performance were observed in my tests.    

(2). Interesting. Shall fought about this later. I want to left this out of this PR scope.  

(3). Ok, I'll try to find what effect on performance it has.  But it's connected with parser thread as described in (5). Will try to check if everything is fine.

(4). Hm, I worrying about performance. In fact current hashes used in `-q` and `--cubic` seems pretty Bernstein's for me as can be represented in form of `f(x) = k*x+b`. Just `k` and `b` are different. So I assume another Bernsteins won't be mush worse than current ones. And still I'm missing an intuition about what restrictions which are applied to such constants in VW? You've mentioned that `many of our features have small values`. I think this is about feature hashes (positions in regressor array) rather than weights (values in regressor array). If so, murmur3 seems to distribute hashes uniformly in uint32, so there shouldn't be any imbalance in them that must be addressed with special constant.
As for stagewise_poly.cc - it seems to be not easy to understand the details from code, could you point me to ant article or give some keywords to google some info regarding hashing implemented there?

(5). *Imagine defining a foreach_feature which _either_ looped through the cache _or_ created a cache if one did not exist. Is there a reason why this could not work well?* There is at least one. With new feature generation logic depending on feature weight (case X*X with weight != 1.) you can't easily calculate total number of generated features and can't just increase `ec.num_features` in [parser.cc](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parser.cc#L840). Thus `num_features` shall be updated in the code where such features are generated. So, either useless `foreach_feature` is need to be called in parser.cc just to generated features or we must be sure that there is no any code depending on `ec.num_features` called before first `foreach_feature` is used. Otherwise such code will work with incorrect `ec.num_features` value. As I have no comprehensive knowledge of VW codebase to be sure in second I've preferred first. This actually might affect also (3). 

As for gd_mf etc. I need to take a look into them. They shouldn't depend on num_features as it was updated in parcer.cc for -q and --cubic  for all algorithms with no exceptions. They could depend on namespaces and perhaps I just need to hide interactions_namespace from ec.indices. But I don't want them to do double work if they could reuse values from interactions_namespace.  Need to think about it.

(6) Let's leave it for now.

JohnLangford:
(apologies, been traveling)

For 2&3, I believe that anyone who wants the old code should just check out the old code.  I see no reason to continue supporting (in any fashion) an obsolete approach.

W.r.t. ensure_could_push, the extra realloc's only happen if you have a fresh v_array.  Most of the time you should not, implying that ensure_could_push slows you down.

W.r.t. efficient feature generation, can this be left out?  Efficiency and the location in the processing stack seem inherently related.

for (3) Did you have a chance to do perf tests?

for (4), '| 1:3.4' has one feature with index 1.  Effectively, the default approach hashes features with numeric feature names to the index applied by the name.  I'd prefer to avoid an approach with poor performance on the default hash function.  This particular hash has that problems because 17 and 31 are not large enough to exercise even the default parameter space if all seed features have small values.

For (5), I'm willing to accept duplicate features for Interact(a,a) for efficiency reasons.  I expect they will have little impact in practice because they are always only a very small minority.  

num_features is primarily used for reporting after the learning process.

trufanov-nok:
Hi,
.1. Regarding `Having -q :: do 'ab' and 'ba' seems undesirable.`  
I've included related changes in this PR. But now I realise that this code may not be enough as  I can't answer one question: what shall be the result of duplicates removal from `--cubic cba --cubic cab`? Shall it be `abc` or `cba`? In other words - is it allowed to sort interactions without a reason. Previously in `--interactions` i always sorted them and used `abc`. But If I would be a user I would expect VW to remove one of duplicated interactions but wouldn't expect it to make any permutations in the one that was left. So, perhaps I shall sort namespaces in interactions if only `all.permutations == false` (we want simple combinations, it's by default) and interactions have at least one duplicate namespace inside of it. 

It seems i can formulate this as a rule and suggest following concept:
a. Duplicated interactions should be **always** removed from `-q, --cubic and --interactions`.
b. After duplicated interactions were removed it's allowed to sort indexes in interactions ascendantly if only `all.permutations==false` (it's false by default) and there is at least one duplicate namespace in interaction. Otherwise it should be left as is.

For example: 
```
vw  --cubic bca cba --interactions cabcc" -> "bca abccc" ## cba removed,  cabcc sorted, bca isn't
vw --permutations --cubic bca cba --interactions cabcc" -> "bca cabcc" ## cba removed, nothing sorted as all.permutations == true
vw  --cubic aba baa --interactions cabcc" -> "aab abccc"
```
Ideally it would be better to process duplicate namespaces in interactions without sorting indexes in interaction string at all, but It will definitely be more difficult to do and  hit the performance.


.2. Regarding ` I see no reason to continue supporting (in any fashion) an obsolete approach.` - I've deleted all code which I put into `#ifndef USE_INTERACTIONS` with definition itself. Commit included in PR.

.3. Regarding `W.r.t. ensure_could_push, the extra realloc's only happen if you have a fresh v_array. Most of the time you should not, implying that ensure_could_push slows you down.`

I've tried to profile VW with Valgrind with `ensure_could_push()` and without it. As I expected `ensure_could_push` usage decreases `realloc` invocations by 1-2% (tested on 100k examples with 20k features in each). But surprisingly Valgrind shows that 101% of `reallocs`in case  there is no  `ensure_could_push`  has 30% smaller "cost". Even more strange that invocations of `free()` (which are called 100k times in both cases) costs 30% less without `ensure_could_push`. 
So, I've decided to remove `ensure_could_push` as I couldn't get any strong evidence of its usefulness.

4. Regarding `W.r.t. efficient feature generation, can this be left out? Efficiency and the location in the processing stack seem inherently related.` - will discuss this in next post.

(3).`Did you have a chance to do perf tests?`
I've performed a simple experiment yesterday by moving call of `generate_interactions()` from `parser.cc:setup_example()` [case VW1] to `learner.cc:dispatch_example()` [case VW2]. Plus tested original VW more precisely [case VW3]. The test file could be found [here]( https://yadi.sk/d/9I4n1f6LfddoR).  The command line was ``time vw test --redefine q:=: -q qq --cubic qqq --passes 7 -c -k --permutations`` for [VW1, VW2] and ``time vw test --redefine q:=: -q qq --cubic qqq --passes 7 -c -k`` for [VW3] (as no --permutations key implemented in it).  
The average (3 test launches) results:
```
VW1
real 136.6053ms sd = 4.786588
user 189.57ms sd = 8.423677
sys  1.884667ms sd = 0.0872143

VW2
real 168.3103ms sd = 1.674398
user 176.269ms sd = 2.034453
sys 2.527667ms sd = 0.0925869

VW3
real 127.5957ms sd = 9.859673
user 131.0027ms sd = 10.89027
sys  2.084667ms sd = 0.08119319
```

It looks like calling `generate_interactions` in parser thread (VW1) is a bit faster than calling it in learning thread on multicore CPU although requires more CPU time. The problem is it's 10% worse than VW3 (original) that i didn't expect to see. It seems I need to stop and think about performance again. Perhaps your approach of efficient feature generation could be useful, and i need to change hash function anyway, and i want to improve performance.

trufanov-nok:
Regarding performance: bad news - I've failed to find any optimization that noticeably improves my code performance. The only idea left is to incorporate a simple `for` loops for case of paris or cubic features and use current code only for bigger level of interactions. But I don't want to do that yet, bcs: 
Good news - I've retested everything more carefully and the [results](http://pastebin.com/uVG3gZLh) shows that:

* VW with feature generation in learning thread is 10-15% slower than original VW.
* VW with feature generation in parser thread is only 2-3% slower than original VW.

Thus: 1. I'll keep feature generation in parser thread. 2. I'm ok with this performance. Will see what happens when I change hash function.

arielf:
Why not use the current, in the code (murmur_hash32 v3.0)?  it is one of the fastest hashes available.

trufanov-nok:
Hi Ariel,
The hash we are discussing is used for interactions only. Their number grows non-linearly. On practice I've used 12k interactions made from a dataset with only 30 features. So even if it's 10% slower than simple `a*x+b` this will hit performance dramatically. Also it feels quite overcautious to perform whole murmurhash over 2 operands which are already murmurhashed. I suspect that simple xor would suite for this purpose if only not features with implicitly set hash which John refers to.

arielf:
Thanks Alex. Got it. Sorry.

trufanov-nok:
Regarding hashing.
* I've got John's point regarding implicitly setted hash values. Once I was going to add all my features to a dictionary and use their ids instead of string values to avoid collisions but found that `-b` param doesn't affect my results much enough to make this work reasonable.
* It seems new hash function (Bernstein’s) has one drawback - for any set of values `a,b..y,z` the `hash(a,b..y,z) == hash(a,b..y+1,z+527)` and 527 is too small - less than expected number of unique features in dataset if someone decides to use dictionary approach. I suppose its constants were adjusted to hash ASCII strings.
* Looks like I need to take a look on FNV hash implemented in stagewise_poly.cc. I suspect it will be slower than simpler `ax+b` hashes. So it will be used for interaction levels > 3 only with no hope to make a hash for all cases. Thus we'll end up with 3 hashing algorithms for interactions. And as FNV significantly differs from pair and cubic hashes I don't think I'll be able to implement all of them in one generic non-recursive loop without breaking last gcc loop optimisation efforts. Thus I'm going to implement backoff model mentioned in previous post: 
```
for (all interactions)
if (interaction_len == 2) {
   for { for { pairs_hash }} //old code
 } else if (interaction_len == 3) {
  for {for { for { cubic_hash}}} //old code
} else {
 wile loop for any length {FNV_hash} //new code
}
```
I've already tried it. Such approach isn't equal to the current VW implementation bcs the code is called from parser thread + generated features are stored in cache. I've tested performance and it [looks comparable enough](http://pastebin.com/U6MyHEH7).
Also in case i fail with FNV and high levels of interactions I'll copy paste these pieces into gd.h:FOREACH_FEATURE to add simple combinations support for pairs and triples only. I already have this code so it always was a minimum goal.

trufanov-nok:
Hi, I've taken a look into `stagewise_poly.cc`. I wouldn't say that I understood everything but it looks like their hash function (omitting stride_shift unrolling) is [just a](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/stagewise_poly.cc#L221) : ``feature1 xor 16777619*feature2``. And that makes me diffident because of:  
1. This approach would be good bcs it's only 2 operations.
2. I wouldn't say it's a FNV hash in any matter. 

It looks like [FNV-1a](http://www.isthe.com/chongo/tech/comp/fnv/#FNV-1a) hash. But FNV-1a designed to hash one argument and looks like:
```
hash = offset_basis
for each octet_of_data to be hashed
        hash = hash xor octet_of_data * FNV_prime
return hash
```
where `32 bit FNV_prime = 16777619` and `32 bit offset_basis = 2166136261`. Values are listed [here](http://www.isthe.com/chongo/tech/comp/fnv/#FNV-param). Note 16777619 usage in stagewise_poly.  
In `stagewise_poly` second feature value is used instead of `offset_basis`. Also `xor` is performed for whole integers instead by octets. According to documentation:
``The xor is performed on the low order octet (8 bits) of hash.``
So each octet of value should be xor'd with hash[0] in a loop.
Thus makes me think that `stagewise_poly`'s approach is too different from FNV hash to expect it to have any known FNV hash properties.  
I think original FNV still could be used to interact 2 features. All we need is to just make 64 bit integer from two 32bit integers and use `64 bit offset_basis = 1469598103934665603` + `64 bit FNV_prime = 1099511628211` instead of 32bit parameters and than xor-fold it to 32bit as [recommended](http://www.isthe.com/chongo/tech/comp/fnv/#xor-fold). For ex, my pseudocode: 
```
int32 xorfold(int64 val) {
    int32* x = &val;
    return *(x+1) xor *x;}
```
But it will be much slower than currently used hash bcs of `for` loop used for octets which doubtfully could be vectorized by compiler.

So I'm going to use ``feature1 xor 16777619*feature2``, but is there reasons to be sure that it's good enough?

trufanov-nok:
Last commit contains `for` loops for pairs and triples instead of processing all levels of interaction in one generic loop. FNV-like hash is used for levels of interactions > 3.  
There is a `//#define FNV_HASH_TEST` keyword. If you uncomment it then pairs and triples will be processed in generic loop with FNV-like hash. My impression is that FNV-like performance is same with Bernstain hash.  
Moreover average loss seems to be lower than if qubic/pair constants used for same dataset (0.567055 vs 0.570414 with 7 passes in test dataset). So, perhaps FNV-like hash is better than current hashes used for pairs and triples. If someone want to test by himself I'm leaving `FNV_HASH_TEST` in the code.
By default i'll keep use the old hashes for pairs and triples and FNV only for other level of interactions.

trufanov-nok:
The commit above adjusts the interactions preprocessing code with proposed rules:
```
1. Duplicated interactions should be always removed from -q, --cubic and --interactions.
2. After duplicated interactions were removed it's allowed to sort indexes in interactions ascendantly if only all.permutations==false (it's false by default) and there is at least one duplicate namespace in interaction. Otherwise it should be left as is.  
```
If that scheme will be accepted I'll also simplify preprocessing code in parse_args.cc

trufanov-nok:
There were quite a lot of text since last John's comment, so I'll wait for new comments and continue after that.
Even if changes above will not rise any questions there are few big problems unsolved:
1. moving feature generation to learning thread as reduction should solve its conflicts with other vw modes like `--rank` but may hit performance.
2. efficient expansion of namespace wildcards in interactions.
I'll work on that one by one after at least hashing problem solution (Bernsteins -> FNV) will be accepted.

bscottm:
@trufanov-nok: Have you considered manually unrolling loops by factors of 2 or 4? Often times, this will have a beneficial impact on loop performance.

trufanov-nok:
@bscottm , Yes. Currently loops for pairs and cubics are enrolled.

bscottm:
@trufanov-nok: "enrolled" or "unrolled"? By what factor? 2? 4? 8?

trufanov-nok:
@bscottm , OK, now I got what you mean. Googled [it](http://en.wikipedia.org/wiki/Loop_unrolling). Unfortunately I can't make any assumptions on array lengths at compilation time. I've tried it for pairs with factor 4 in following manner. Replaced:
```
for (i:n)
 for (j:m)
  do_pair(i,j)
```
with 
```
for (i:n)
 for (j:m)
{
  do_pair(i,j)
  ++j;
  if (j > m) break;

  do_pair(i,j)
  ++j;
  if (j > m) break;

  do_pair(i,j)
  ++j;
  if (j > m) break;

  do_pair(i,j)
  ++j;
  if (j > m) break;
}
```
inspired by [unrolling wile loops](http://en.wikipedia.org/wiki/Loop_unrolling#Unrolling_WHILE_loops).
I've tested it with --q qq where length(q) == 27 and didn't observe any performance increase.

I don't feel that it could improve performance. I didn't observe any perf changes with replacing indexes with pointers too.

bscottm:
@trufanov-nok: Here's where high performance computing experience comes in for the win... :-)

I'm assuming that this is numeric code and not string manipulation code. Otherwise, you're right, unrolling is not a win. Unrolling doesn't help string manipulation much.

You _do not_ want to have 'break' statements in the middle of the loop, however, because that'll cause pipeline bubbles in the processor pipeline (bubbles == stalls, stalls are a bad thing and the likely reason you're not seeing any performance improvement.) 

Basically, the unrolled loop should operate on 4 successive values, almost like writing SIMD vector code. If there are values left over after the unrolled loop, then operate on them in a separate loop.

Also, if you have vectorization turned on as a gcc or clang optimization, unrolled loops make it easier for the compiler to spot vectorization opportunities (if you leave out the break statements.) Think of the compiler's optimizer as a very interesting tree pattern matcher. :-) :-) :-)

Something similar to:

```
for (i:n) {
  for (j:(m&3)) {
    do_pair(i,j)
    do_pair(i,j+1)
    do_pair(i,j+2)
    do_pair(i,j+3)
    j += 4;
  }
  for (j:(m&3)..m) {
    do_pair(i, j)
  }
}
```

JohnLangford:
(been overloaded, but less so now)

For .1: The sort-or-not decision seems counterintuitive.  What if we do no sort, but just warn when there are duplicates via the sort test?  It is _possible_ that people want duplicates.  When using the hashing trick, it can be beneficial to have multiple features because some may have bad collisions while others do not.

>So I'm going to use feature1 xor 16777619*feature2, but is there reasons to be sure that it's good >enough?

Various choices were tested on about 30 datasets and this was found to be the best.  I have some confidence that it's a good approach for us, but that's subject to revision by evidence.

>But it looks like formula h(feature1,feature2) {feature2 xor 16777619*feature1} works.

That's rather interesting.   I'll send an email around to the stagepoly authors about it.

> By default i'll keep use the old hashes for pairs and triples and FNV only for other level of interactions.

I'd rather have the same hash used everywhere unless there is some confirmed reason to not do so.  Simplicity is a virtue.  Does this have a negative impact on performance?

>1. moving feature generation to learning thread as reduction should solve its conflicts with other vw >modes like --rank but may hit performance.

Yeah, I am concerned that we're pondering an update which slows things down and breaks compatibility.  Doing it late in the foreach_feature template has some real virtues.   Why do we want to early expand again?

JohnLangford:
Incidentally, I am curious about whether bscottm's approach is helpful.

trufanov-nok:
> For .1: The sort-or-not decision seems counterintuitive. What if we do no sort, but just warn when there are duplicates via the sort test? It is possible that people want duplicates. When using the hashing trick, it can be beneficial to have multiple features because some may have bad collisions while others do not.

We need to be careful here with terms. We talk about interactions (strings) which contain namespaces (characters). We do not remove duplicate namespaces but they are processed in a special manner to generate simple combinations instead of permutations of features. Due to asymmetry of hash the permutations would allow to get 2 different hashes for same pair. This might give us a chance to bypass hash collision but on practice it doesn't worth it. This is why we are trying to replace old permutations with simple combinations.  
Now we are considering interactions level. There are 2 questions: do we need to sort namespaces in interaction (chars in strings) & do we need to remove duplicate interactions (duplicated unordered strings, not duplicated chars in string).  

As for sort - I just must do it. Otherwise i can't group duplicate namesaces in interaction and process them in a previously mentioned special manner. The nuance is: shall I sort namespaces always or only when really needed (there is at least one duplicate namespace in interaction).  To keep it simple you might recommend to sort always. (as we can't without sorting at all)

As for duplicate interactions - they better be removed for very same reason we are replacing permutations with simple combinations - it doesn't worth to bypass hash collisions in a such way. Moreover, if you really will recommend to sort always then both interactions `ba` and `ab` become `ab` and this even won't generate different hashes.  
Also I expect wildcards like `-q ::` to be a reason of duplicates existence in most cases. I believe there should be a programmatic (VW's feature) way to remove such duplicates. Just because if user forced to remove them manually he have to stop using wildcards and write a script that generate him all possible pairs of letters without duplicates and that will be a lot of parameters. And it will be even worse for `--cubic :::`. I tried that once - got a very long command line.

To sum these arguments: if we want make it more intuitive we have not much options: it shall sort always + remove duplicate interactions. I can add a warning if some interactions got removed. And if you wish - optional command line param to disable duplicate removal. Or vice versa - warning + param to enable duplicates removal. I just sure there shall be this feature (duplicates removal).

trufanov-nok:
>I'd rather have the same hash used everywhere unless there is some confirmed reason to not do so. Simplicity is a virtue. Does this have a negative impact on performance?

I tried that and don't think it has strong effect on performance. Will test FNV-like hash performance as default hash for all cases.

>Yeah, I am concerned that we're pondering an update which slows things down and breaks compatibility. Doing it late in the foreach_feature template has some real virtues. Why do we want to early expand again?

Because of estimation of `num_features` and `total_sum_feat_sq` in `parser.cc::setup_example()`. It seems to be vital to count pairs and triples at this point. In master it estimated just by [multiplying](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parser.cc#L840) count of features. In case permutations replaced with simple combinations it a bit harder. And if we keep features x*x when weight of x != 1 it requires iterating throw all features which makes it worth to use cache at this point. I don't ready yet to refuse from keeping such x*x yet. And as I managed to prove that it works faster in parser thread I would leave it there.
The real problem we are trying to solve by returning feature generation back to learning thread is: `manipulation with namespaces cause problems with some VW modes` and it could be solved by implementing feature generation as reduction.

I suggest to suspend this problem - I need to think about it. I believe I can solve his problem without using reductions. Give me a week to play with it.

trufanov-nok:
Quoting myself:
>As for sort - I just must do it. Otherwise i can't group duplicate namesaces in interaction and process them in a previously mentioned special manner. 

I just thought that I possible can do simple combinations without sorting of namespaces in interaction. (need to preprocess interactions and save some indexes in vw object among interactions' namespaces) So there might be another option that I could suggest: not to sort + warning on duplicate interactions +  duplicates removal by default/optional.
I suppose you'll prefer this option? With default or optional duplicates removal?

JohnLangford:
This seems a convincing argument.  Let's go with sort always + remove
duplicate interactions + warn + option to disable sort/removal.

-John

JohnLangford:
My understanding is that the _most_effective_ solution found for feature generation is what exists at present in foreach_feature.  Given this, perhaps we should be generalizing that for higher order combinations rather than struggling with the extra computation imposed by doing it elsewhere?

The issue with num_features and total_sum_feat_sq could be handled at the bottom rather than the top of the reduction stack.  num_features is primarily for reporting and total_sum_feat_sq is only used by sgd, so it's wasted computation in almost all cases.

A nonsorting approach to duplicates removal is desirable, because there are computational implications to the order of expansion.   Consider a 2 x 1000 outer product.  Done one way, you have (effectively) 1000 size-2 operations.  Done the other way, you have 2 size 1000 operations.  I would not be surprised if these differ computationally.

bscottm:
Most assuredly, 2x1000 and 1000x2 are different. And they optimize differently -- you have a lot of algorithmic tricks in the 1000x2 case (if that weren't the case, Jack Dongarra at UTK [aka "The Wizard of Matrices at Knoxville"] wouldn't have made a good career out of ATLAS.)

bscottm:
There's also the slightly more complicated "lazy" approach, which would mean saving state in a structure (aka the closure approach) and computing 8 (or 16 or 32 or however many) at a time, as needed. That tends to amortize the cost of computing the pairs by increasing code execution locality. The fast case is a pop off the front of a queue 8/9ths of the time, 1/9 of the time, you're generating a new set of 8 pairs.

One of the tricks from purely functional data structures school of computing.

trufanov-nok:
Hi,
Sorry for a delay - I had a problems with update to Kubuntu 15. Latest commits do 2 things:
1. Now FNV-like hash s used for pairs and triples too. It seems to work slightly better than original approach.
2. Now namespaces in interactions are always sorted. If there are duplicates - they are removed. If something were removed - a warning displayed with suggestion to disable duplicates filtration if needed. Sample output:
```
creating quadratic features for pairs: :: 
WARNING: duplicate namespace interactions were found. Removed: 4186.
You can use --leave_duplicate_interactions to disable this behaviour.
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
```

Unfortunately there is no param to disable namespaces sorting itself - only duplicates filtration in them. I couldn't think out how to implement simple combinations without preliminary sorting.

trufanov-nok:
It was a really long and interesting discussion and I hope it's going to its end. So I would like to make some summary.

What we are able to offer VW users in scope of this PR request:

1. A better hash function for future interactions - FNV-like hash
2. Automatic future interactions duplicates removal - "simple combinations mode" which speed ups VW in case namespace is interacted with itself (`-q aa`). Could be disabled with `--permutations`
3. Automatic interactions duplicates removal which speed ups VW in case such duplicates is generated (by `-q ::`). Could be disabled with `--leave_duplicate_interactions`.
4. Support of interactions of higher order than triples. With help of `--interactions abcd_etc`

Drawbacks are

1. Namespaces in interactions are now always sorted (`-q aB` will be processed as `-q Ba`)
2. Usage of many namespace wildcards in one interaction is still slow at VW parse_params stage. (`--cubic :::` or `--interactions ::::::`)
3. There are question about feature generation performance for higher than triples order of interaction.

Few notes on performance: 

* I can always make pairs and triples processing as fast as in vanilla VW by implementing them as embedded `for` loops (as in master branch) instead of generic non-recursive loop. In fact, I can do this for quadruples etc. if needed.
* I think I shall try inline recursion to automatize generation of such fast nested `for` loops. But it's compiler specific (for ex., `#pragma inline_recursion, inline_depth` for MSVC compiler).

And we still have one conceptual problem:
I'm creating additional namespace to store generated features and this breaks several VW modes.
I'm suggesting cutting this knot by refusing from new namespace creation. It can be done in a few ways:

1. We leave feature generation code in parser thread but features will be stored in `v_array ec.cache` -  a new array member of example class. This class member will be additionally processed in `GD::foreach_feature()` after `atomics`.
2. Move feature generation to learning thread. Don't store features - just call a function for newly generated feature and forget it. 

Both are doable. I would advocate for first one: It's performance is better tested and almost equal to original VW. Also for me is natural that cache generation is a part of parser thread.
As for 2nd approach - I've tried it. It may resolve in more bulky code bcs of templates. It's performance is equal to master branch due to nested `for` loops usage for pairs and triples.  But it could be slower than 1st approach for higher levels of interactions when it switches to non-recursive generic loop. Also I didn't implement the code for counting ` total_sum_feat_sq ` for it but it shouldn’t affect performance.

In other words I can make both and both will be on par by performance with original VW in modes (`-q` `--cubic`) that original VW supports. The question is how fast will be higher orders of interactions processing (`--interaction abcd`). And I would prefer 1st approach as it's ready, requires less code and tested.

My plan is following:
I'll wait for John's design on problem above. Then I'm going to create a new PR and adapt all new features to current master branch to make it merge-able. And then we can polish it by fixing non-conceptual problems and making final perf tests.

JohnLangford:
It does seem that we're close to landing here.

Here is a trick to avoid reordering things.
Start: an array of strings [s1, s2, s3, ...]
Create: a new array [(1,s1'), (2,s2'), (3,s3'), ... ] where the ' denotes sorting the characters.
Sort the new array by the second field of each pair: [ (4,s4'), (8,s8'), (2,s2'), ...]
Loop through the new array checking uniqueness.  
If (i,si') is unique, keep si (which you can look up with i).
otherwise warn for si.

I'm more comfortable with choice (2) on three grounds:
(a) It is an approach more compatible with GPUs as may matter going forward.  This is minor, but it might plausibly become major in the future.
(b) Benchmarking seems to reveal that while (1) takes similar time, it takes more cpu-time.  This is a minor disadvantage.
(c) I believe that (2) will be more compatible with some of the existing algorithms (MF, poly, etc...) otherwise left in the cold.  This seems major.

The only advantage that I see to (1) is that it's already done, but that doesn't seem important in the long run---getting it right is important.

trufanov-nok:
Ok, I'll do (2).

Regarding sorting - when i was talking about drawbacks i meant order of namespaces (characters) in single interaction (string). As hash(a,b) != hash(b,a) this might have some impact on result due to different number of collisions. This sorting is done to decrease complicity of duplicate namespaces detection which is needed to make sure we generate simple combinations, not permutations. Doing this without sorting severely hit performance.

Your suggestion could be applied only to order of interactions (strings) themselves as i really sort them too before duplicate interactions removal. For example `-q ab -q ab -q aa` become `-q aa -q ab`. Frankly I didn't even thought about keeping their order as in contrary with previously described it affects only order of processing of newly generated features, not their index in regressor. If you think it's better to keep this order, i can do that.

JohnLangford:
I was thinking about -q ab -q ba which might happen implicitly via -q :: .

But, you have a point about -q aa .  Perhaps _only_ reorder things if you discover duplicate characters in the interactions?

JohnLangford:
To expand on that: "An interaction sequence consists of a set of characters that are reordered only if some of the characters are duplicates"?

trufanov-nok:
Hm, how about this in your terms? Rule of thumb # 2:
1. VW removes interaction sequences which consists of a same set of characters leaving only one of them.
2. VW reorder interaction sequence if some of the characters are duplicates in it".

So in case we had `--cubic bab --cubic abb -q aa --cubic bab`
After step one applied: `--cubic bab -q aa`
After step two applied: `--cubic abb -q aa`

Additionally I can print a warning with a number of interactions which characters order was really changed. And even make it close eyes on -q aa and --cubic aaa as sorting change nothing in them.

So for case described above it will be:

`WARNING: some interactions contained duplicated characters and were reordered. Interactions affected: 1`

JohnLangford:
Closing this one as the primary is now merged.

PR #560, opened 2015-03-19
fix Visual Studio Build Errors and warning

nicknussbaum:
Add missing files to static lib project

fix build errors and some warnings

### DIFF
--- a/c_test/sample.c
+++ b/c_test/sample.c
@@ -1,7 +1,14 @@
 #include <stdio.h>
 
+
+#ifndef OMIT_MS				// a gcc sample would omit these two defines
+#define MS_CONV				// use Microsoft libraray convention
+#define USE_CODECVT			// enable wide character converesion enabled API
+#endif
+
+
 /*
-To compule with gcc use this command
+To compile with gcc use this command
 LD_LIBRARY_PATH=../vowpalwabbit/.libs ; gcc sample_gcc.c -I./../vowpalwabbit/ -L./../vowpalwabbit/.libs -lvw -lvw_c_wrapper -lallreduce -o sample_gcc
 */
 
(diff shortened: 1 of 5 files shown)


JohnLangford:
Thanks.

trufanov-nok:
Hi, I've found a problem with this PR. It's described here: https://github.com/JohnLangford/vowpal_wabbit/issues/574
There is one line to fix. Could someone pls reverse its change?

nicknussbaum:
Sorry about that.

 

Can you verify that the following change works for you?

 

      for (size_t i = 0; i < 256; i++)

          all.redefine[i] = (unsigned char)i;

trufanov-nok:
Yes, it works.
Btw, i've also [googled](http://stackoverflow.com/questions/13139621/looping-with-an-unsigned-char) the "elegant" solution for such problem:
```
      unsigned char i = 0;
      do {
          all.redefine[i] = i;
      } while (++i != 0);
```
But old one is shorter and more expectable. So unless someone think it faster I would stick with old loop.

nicknussbaum:
The fix it’s to avoid the warning for the cast

 

I’ve submitted a pull request for the cast version.

 

I’m not wild about depending on the max size of char as the form you suggest.

I’d probably prefer the slightly slower

 

unsigned char  c = 0;

For ( int I = 0; I <255; ii++) {
                all.redefine[i] = c++;

}

 

Anyhow, sorry for bug.

bscottm:
@JohnLangford, @trufanov-nok: I ran into this as well earlier. It's also in my pull request.

bscottm:
It's not a good idea, defensively, to use unsigned char or char as an array index, which ought to be obvious by now.

bscottm:
@JohnLangford: There were also undefined behavior warnings in parse_args.cc. Also in my pull request.

PR #567, opened 2015-03-24
Added PiSTOL algorithm to the FTRL family

bremen79:
Hi John,
I have implemented my algorithm PiSTOL in the FTRL file (renamed from ftrl_proximal to ftrl).
The reason to use the same file is that both are FTRL algorithms. Ideally, any other FTRL algorithm can be simply added there, just implementing two functions (pre and post update) and a switch.

There is now an option (ftrl_algo) to call Proximal FTRL or PiSTOL FTRL. The default is PiSTOL.
The options ftrl_alpha and ftrl_beta controls similar numbers in the two algorithms. The default values are different for each algorithm.
I also added a couple of tests for PiSTOL.

Let me know if there is anything you want me to change.


JohnLangford:
Two thoughts: 

1) Do you really need to make 3 loops over the features rather than 2?  I had FTRL down to 2.
2) Instead of adding the --ftrl_algo flag, can you just reuse the --ftrl flag to pass the algo?

bremen79:
1) Still thinking about it, probably I cannot change it unless I add an approximation in the update
2) This is more tricky than expected and I need you opinion here. It seems that in boost if I specify a default value for an option, the function count will always return true. (BTW this seems to be not taken into account in scorer.cc). Hence, if I define --ftrl with default value pistol, I cannot check if the flag is present or not. On the other hand, if I don't specify a default value, you are forced to write each time which ftrl you want.
Which behavior would you like?

bremen79:
Nope, it does not work. If I specify an implicit value, and type

```
vw --ftrl filename
```

filename will be considered the name of the algorithm.
An alternative is not to have a default value, but in this case to get the help for the FTRL options I would have to specify

```
vw --ftrl anything --help
```

rather than just

```
vw --ftrl --help
```

Note that other switches have the same behaviour, for example for the LDA options I have to type

```
vw --lda any_number --help
```

and I get and error with

```
vw --lda --help
```

Is this what you prefer?

JohnLangford:
Hmm.  How about having two valid options: --ftrl and --pistol?   Similar to here: 
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/bfgs.cc#L966

JohnLangford:
If you want (this isn't necessary), you can also clean up the code a bit.  
1) It would be nice if the update_data structure was a part of the ftrl data structure so it did not need to be reassigned on every example.  
2) There is very little shared code between proximal and pistol.  Simply having a learn_pistol and learn_proximal is probably epsilon clearer and epsilon^2 faster.

bremen79:
Done everything. Better now?

JohnLangford:
Yes.

JohnLangford:
Yep.

All, I just merged in Francesco's Pistol algorithm (--pistol).  That means,
we now have:

(a) The default importance-weighted adaptive normalized approach.
(b) A svrg variant.
(c) FTRL-proximal.
(d) Pistol

-John


On Mon, Mar 30, 2015 at 1:47 PM, Francesco Orabona <notifications@github.com

PR #571, opened 2015-03-26
whitespace between struct or ) and {

zenogantner:
Cosmetic issues I came across while studying the source code.
Let me know if this kind of pull request makes sense.


arielf:
Hi @zenogantner 

I think this change is ok, except it includes a huge diff (https://github.com/zenogantner/vowpal_wabbit/commit/f6e581e2f8c5340106e92813a1d3bdedafa1511f) because it was being applied to an old version.
This has two unwanted side-effects:
- The diff shows much more than the change you intended to make
- hard to tell if there aren't issues with the large merge/commit https://github.com/zenogantner/vowpal_wabbit/commit/f6e581e2f8c5340106e92813a1d3bdedafa1511f

If you could close this, do a full merge with latest first, then redo, maybe John would be more inclined to accept this pull request.  Thanks!

JohnLangford:
I have no real opinion either way here, so this seems fine if it is preferred.

zenogantner:
Hi @arielf, thank you for looking into this. Next time I will make sure that I work on the latest version of the master repo/branch, to avoid unnecessary confusion.

bscottm:
@arielf, @JohnLangford, @zenogantner: How about using astyle or clang-format to make code style consistent?

Upside: Consistent code style, easy(ier) for others to read.
Downside: Large delta in github.

arielf:
Or use trusty old `indent` (`apt-get install indent` on Debian/Ubuntu/Mintmap) in batch.
It supports a `~/.indent.pro` config file for various styles) like:
```
--indent-level4
--k-and-r-style
--blank-lines-after-procedures
--braces-on-if-line
--braces-on-struct-decl-line
--case-indentation 2
--cuddle-do-while
--cuddle-else
--no-tabs
```
So by adding a spec like the above to the source tree and a `Makefile` rule to apply it, we can recommend (and easily enforce) a preferable common style.

I never fancied the inconsistencies but that's part of the price you pay when letting many people contribute without establishing ground rules like the above from the start.

John any strong opinions here?

bscottm:
@arielf: astyle seems better maintained than trusty 'ol indent and better suited for C++ and C++-like code. Just my $0.02, discounted for inflation.

arielf:
Touche.  The 80's have called and they want `indent` back :)

I just tried `astyle` and it looks great.  As log as we pick one of:
```
--style=kr                      # aka -A3
--style=stroustrup         # aka -A4
--style=linux                  # aka -A8
--style=1tbs -A10          # "One True Brace Style"
```
I'm in.
My favorite is -A10.

Please no Lisp, GNU, Horstmann or the other visual horrors.

bscottm:
@arielf: "1tbs" should be renamed "highlander", because there can be only one. :-)

PR #572, opened 2015-03-28
Visual Studio doesn't support constexpr in VS2013

nicknussbaum:
The windows build was failing. I've put some fixes in so that it builds and removed a bunch of warnings.

Visual Studio doesn't support constexpr in VS2013
static const member doesn't work with floats
Replace it with an inline function. (Could be a macro if performance critical)
 
conditionally define functions which call other conditionally defined functiosn
	expdigammify calls the undefined vexpdigammify
	expdigammify_2 calls the undefined  vexpdigammify_2
so they aren't called when the v functions are not defined

Many case statements don't have a default case. Added one which defaults to case USE_PRECISE:

### DIFF
--- a/vowpalwabbit/lda_core.cc
+++ b/vowpalwabbit/lda_core.cc
@@ -66,7 +66,11 @@ struct lda {
   double example_t;
   vw *all; // regressor, lda
 
-  static constexpr float underflow_threshold = 1.0e-10f;
+// constexpr is not supported in VS2013 (except in one CTP release)
+// if it makes it into VS 2015 change the next ifdef to check Visual Studio Release
+
+  // static constexpr float underflow_threshold = 1.0e-10f;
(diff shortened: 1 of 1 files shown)


arielf:
Thanks Nick,

Looks like there's some issue with travis/rabbitmq (unrelated to this change).
I wonder if there's a way to ask travis to attempt the same build again?

JohnLangford:
The same thing happened to a different build.   

Nick, can you bump and retest to see if this is persistent for some reason?

JohnLangford:
This looks like it should just work.   I'll risk merging.

nicknussbaum:
Sorry, I don’t know how to engage your Travis tests.

I didn’t think I had permissions for that.

Is there documentation?

 

The code changes do seem minor. It would be good to have the author of those changes make sure it’s not messing up what looks like some heavy duty optimization using constexpression.  

 

Regards

Nick

PR #573, opened 2015-03-28
add --lrqfa option for field aware low rank quadratic features

### LINKED ISSUE #551: "field aware" low rank quadratic features
A new machine learning library called "LIBFFM: A Library for Field-aware Factorization Machines" [just won](https://www.kaggle.com/c/criteo-display-ad-challenge) two [Kaggle competitions](https://www.kaggle.com/c/avazu-ctr-prediction) for predicting ad click-through-rates.

[Reading through their slides](http://www.csie.ntu.edu.tw/~r01922136/slides/ffm.pdf) and [forum posts](https://www.kaggle.com/c/avazu-ctr-prediction/forums/t/12608/4-idiots-solution-libffm), it seem that what they refer to as [factorization machines](http://www.ismll.uni-hildesheim.de/pub/pdfs/Rendle2010FM.pdf) is the same thing as what VW means by "low rank quadratic features."  

I think the difference "factorization machines" and "field aware factorization machines" is as follows:  let's say you're interacting two namespaces: `--lrq=ab1`.  A factorization machine would build a single weight matrix for all the interactions in those two namespaces, while a field-aware factorization machine would build a separate weight matrix for each pair of features in both namespaces.

Am I correct that VW does the latter, and is therefore "field aware?"

zjf:
fix #551 
The new codes are based on lrq.cc and the mainly differences are:
1. The command line argument for lrq is like: `--lrq ab2 --lrq cd2`, and it is like `--lrqfa abcd2` for lrqfa, because this new algorithm is natural to handle interactions among multiple fields. Also, when there exists a lot fields, the latter form is easier to write.
2. The initialization for weight is scaled by a factor of 0.5/sqrt(k), which comes from the implementation of LIBFFM.
3. The dropout option is not fully tested and I will add it later.
4. Better indentation :)

Any comments are welcomed.


JohnLangford:
It looks like this failed travis due to some network failure.  Can you bump and retest it?

JohnLangford:
Paul, can you take a look at this?

-John

zjf:
John, `apt-get update` caused the failure. I don't know travis system quite well. What should I bump and retest exactly? Open a new pull request with the same commits?

pmineiro:
I want to take an extensive look at this, with some additions in the demo/ directory to really kick the tires.  However, I'm heading to the Grand Canyon for 10 days, so there will be some delay.

JohnLangford:
I think it was a transient fail on travis.  Just checkin a whitespace change to trigger a retest.

zjf:
Catch up with master branch and make travis happy now.

JohnLangford:
Paul, I merged this in because it's nicely separated and modular.  I'd
still appreciate any examination you can do.

All, this is code for 'field aware' lrq which has been shown to be useful
on various competitions.

-John

pmineiro:
Just getting to this.  Sadly, there are differences between lrq.cc and lrqfa.cc that do not seem germaine to field-awareness.  Eliminating these differences will ultimately let us merge the two implementations.

I can take a stab at this, but first, I need test cases that help me determine that I have not "gone off the rails".

@zjf Do you have to time to add some test examples to the test/ directory?  Ideally, something where field-aware generates very different results than non-field-aware.  If the test/ directory totally confuses you, you can make a demo/ somewhere in the demo/ directory, that would also work.

Sandy4321:
any progress still 2015

seems to be it is great option to add to VW?

Sandy4321:
also it can be useful 
https://github.com/VowpalWabbit/vowpal_wabbit/tree/master/demo/movielens

PR #576, opened 2015-03-30
simd_fixes_from_ms

bscottm:
Minor tweaks to MSR-annotated fixes (constexpr not supported yet, a couple of places where the SIMD code should be conditionally compiled.)

Basically, when I drop the ARM SIMD code in, shouldn't have issues related to x86-specific conditional compilation.


bscottm:
And some warning reduction related to clang and gcc (some of which were generated in the Travis CI 'make' output, BTW)

JohnLangford:
Nick, can you take a look at this?

-John

JohnLangford:
This patch seems good to go, except I don't understand what the changes to vwdll are buying us.  Why do them?

JohnLangford:
I've merged this, except for the vwdll changes.

PR #577, opened 2015-03-30
Fix #574 redefine functionality is broken

### LINKED ISSUE #574: The redefine functionality is broken
I have just pulled current VW to compare its performance with mine and found out that my implementation of --redefine is broken by [recent](https://github.com/JohnLangford/vowpal_wabbit/commit/7419f80b245c1df4d02d648252c99c5557998bd3) commit.
``for (size_t i = 0; i < 256; i++)``
was replaced with
``for (unsigned char i = 0; i < 256; i++)``
at line [542](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parse_args.cc#L542).

This cause endless loop whenever you use --redefine as unsigned char can't be > 255.

Could someone fix this asap? Seems I can't create second VW fork, can't delete it yet and it's too different from master branch.

*  I'll add a test for --redefine later to make sure this won't happen again.
* Looks like the line was changed to fix Visual Studio build error - pls let me know what error is, perhaps I could workaround it in a safe manner.

nicknussbaum:
( unsigned char iterate counter doesn't terminate)

This fixes a bug introduced in eliminated a Visual Studio warning.

### DIFF
--- a/vowpalwabbit/parse_args.cc
+++ b/vowpalwabbit/parse_args.cc
@@ -539,8 +539,8 @@ void parse_feature_tweaks(vw& all)
   if (vm.count("redefine"))
   {
       // initail values: i-th namespace is redefined to i itself
-      for (unsigned char i = 0; i < 256; i++)
-          all.redefine[i] = i;
+      for (size_t i = 0; i < 256; i++)
+          all.redefine[i] = (unsigned char)i;
 
       // note: --redefine declaration order is matter
       // so --redefine :=L --redefine ab:=M  --ignore L  will ignore all except a and b under new M namspace


trufanov-nok:
Great, thanks!

JohnLangford:
The same bugfix is part of a larger patch.  Nick, can you take a look at that one?

nicknussbaum:
The unsigned char bug should be fixed by the fix in the patch

This might be a little simpler  ( outlook is screwing with capitalization

 

unsigned char c  = 0;

for (size_t I = 0; I < 256; i++) {

                all.redefine[i] = c++;

}

 

 

The other  changes to the wdll.h I’m not sure about. Is there a way for me to pull the patch and test it in visual studio? 

 

If it is going to be done this way then the c_test/sample.c should be fixed to not define the MS_CONV definition

 

 

Nick

bscottm:
I can change the patch to remove the tautological constant issue, if that's solved elsewhere. The symbol export defs macro forest is reused code from another of my projects adapted for VW.

bscottm:
Sample.c probably doesn't need to change, if it's just imploring symbols.

JohnLangford:
I'm merging this one in despite the conflict, because it's obviously needed.

PR #581, opened 2015-04-02
renamed --ftrl_proximal to --ftrl for consistence with documentation

KristianHolsheimer:
The documentation suggests to use the `--ftrl` flag in order to use the FTRL-proximal update rule. I noticed that the flag that was actually supported is `--ftrl_proximal`. I also noticed that there have been a couple of name changes from *ftrl_proximal* to *ftrl* in the code base. So, assuming that you'd want to extend this to the actual command line arguments, I updated those as well. (Conversely, you might want to change `--ftrl` to `--ftrl_proximal` in the docs.)

### DIFF
--- a/test/RunTests
+++ b/test/RunTests
@@ -1079,7 +1079,7 @@ __DATA__
     train-sets/ref/0002.autolink.predict
 
 # Test 72: train FTRL-Proximal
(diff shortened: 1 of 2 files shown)


JohnLangford:
Thanks much!

PR #585, opened 2015-04-10
Deallocate string memory and local reference in JNI layer

vsuthichai:
During a performance test, a memory leak was discovered in the JNI layer.  This is to fix the string deallocation issue that wasn't done in the original JNI implementation here at eHarmony.


JohnLangford:
Thanks---tricky bug.

PR #595, opened 2015-04-23
Add wrapping of C++ exceptions into Java exceptions, and remove finalize() method, as it is not reliable

EmergentOrder:


JohnLangford:
Jonathan, can you review please?

-John

JohnLangford:
Jonathan is concerned that:
a) the pom file contains information specific to their dev environment.
b) the finalizer deletion makes things annoying.  From his response:
... it forces the Java user to clean up after themselves.  I hope that they did this because they had some memory issues (or else why touch it?).  In that case, I would recommend a hybrid approach that allows the calling of closeInstance as many times as possible (will probably require some C++ changes) without doing any work and making closeInstance public.  In that case closeInstance will still be called by finalize but it can also be called manually for high memory situations.  Honestly this seems kinda weird that they would care about this because I would envision they should only be making 1 VWScorer Java object per program and using that a lot.

EmergentOrder:
I made the changes to the pom.xml not for my own environment, but to enable publishing to the public Sonatype Central Repository. The only thing specific to me that it contains is the groupid com.github.EmergentOrder, which was following the convention they suggested. However, I'd be happy to change that to com.github.JohnLangford or to just vw , if you folks would like to take over ownership of the Sonatype project entry etc. If not, I'll remove those changes and publish myself.

EmergentOrder:
Re: the finalizer change, I did in fact encounter memory issues, in a setting where there are many threads each concurrently handling different models, with different params, etc. Hence, needed many VWScorer objects. I can roll that part back for now, and see about making the changes from Jonathon above at a later date.

JohnLangford:
The exception stuff here seems obviously desirable.  

Publishing it on Sonatype seems obviously desirable.  It should probably be your name given that you are doing it.

The finalizer I'm unclear on.  Obviously, we don't want to have a system with memory leaks, but also obviously we don't want to force java people to do memory management as they aren't used to it.   Is the finalizer working in a single-thread environment?  If so, it seems like Jonathan's suggestion is probably best, and if not just going with this patch seems best.

EmergentOrder:
Finalizers in general are not guaranteed to be called by any point in time.
One discussion of this can be found here: http://stackoverflow.com/questions/2506488/when-is-the-finalize-method-called-in-java
Or just search: "java finalize not guaranteed". 
My feeling is that memory errors would be much less likely to occur in a single-threaded environment, ( especially with a single VWScorer object), but no guarantees.

EmergentOrder:
Should I remove the finalizer change and put it in another PR so we can close this one ?

JohnLangford:
It's a tough call.  Let's keep finalizer for now, because I expect most
people use the library single-thread.

But a hybrid fix does seem pretty desirable in the long run.

EmergentOrder:
brought the finalizer back. That Travis fail looks unrelated.

EmergentOrder:
Should be ready to merge, the Travis fail is not on the Java wrapper part.

PR #604, opened 2015-04-29
Consume boost and zlib through nugets instead of sources.

sharatsc:
Source dependency on zlib and boost was a sore point in the windows build process. Changed to consume nugets instead which will be downloaded at build time. 
windows version can now simply be built from command line using 
msbuild /p:Configuration=Release/Debug /platform:x64/Win32 vw.sln

Note: This is a pre-requisite to enabling continous integration on a build server.
would appreciate if somebody could validate the patch


JohnLangford:
Something broke in the build.   https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/60589797 

I'm not sure it's any change that you made, but maybe bump the pull request to retrigger?

JohnLangford:
All, Sharat tweaked the windows build to use NuGet.  This looks pretty
desirable, because you get much easier builds on windows via sucking in
appropriate binaries for zlib and boost from NuGet.  However, I want to
invite comment in case there are any significant issues with NuGet---I'm
not familiar enough with it.

-John

sharatsc:
The build was broken earlier due to https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/60436748 (travis-ci runs AMD64)

sharatsc:
Any comments?

IDisposable:
I'm strongly in favor of using NuGet when possible. It's the moral equivalent of using apt-get.

JohnLangford:
I'm  confused about the SSE2 build breakage---nothing should have happened from the change that I made for Linux builds.  If something did, then I'll need to fix this in gd.cc rather than by updating the test files.

sharatsc:
Looking at the build history on travis https://travis-ci.org/JohnLangford/vowpal_wabbit/builds it appears that the build broke with the SSE2 fix which changed gd.cc outputs for AMD architecture. It so happens that the travis build agents run amd64 
From the build logs, 
OS name: "linux", version: "2.6.32-042stab090.5", arch: "amd64", family: "unix"

JohnLangford:
Something needs to be fixed again in the SSE then.  I'll try to address
shortly.

In the meantime, can you rollback the last commit?

-John

sharatsc:
Rolled back commit related to train-sets/ref

JohnLangford:
This is merged in now, so everyone building on windows will need to
revisit.  Hopefully, it should be much easier now.

-John

PR #608, opened 2015-05-04
add API to save model

hmmroger:
Add a way for saving model at any moment. This is needed when the input
format method does not apply. The input method requires example goes
through generic_driver().


JohnLangford:
Looks good, thanks.

PR #610, opened 2015-05-04
Simple combinations, interactions greater than cubic & caching [fix 2,3,4] - 2

trufanov-nok:
Referring #510

This is remastered version of old PR to make it mergable:
https://github.com/JohnLangford/vowpal_wabbit/pull/556  

I was going to add new 'make test' files but realized that several tests are already broken in master branch. I can update them and then update with changes my code makes but not sure if I should. Perhaps tests results are not outdated but demonstrate a real problem that must be fixed beforehand.

Does someone know something about current state of Travis CI testing?


JohnLangford:
The travis CI testing was my bug, fixed now.

JohnLangford:
Overall, this looks like quite a solid patch now.  

1) How much difference does GEN_INTER_LOOP make timing-wise?  
2) Also how does this compare timing-wise to the existing quadratic code?  
3) Can you bump the patch so that it will retest on travis?  I want to check that everything is working with the new multipredict code.

trufanov-nok:
Ok, last few commits include:
* bugfix for FNV-like hash implementation in special loops for pairs and triples - my mistake.
* code for saving and restoring interactions in model files. This required VW version increment to 7.10.2. Currently I just save interactions after pairs and triples. Technically pairs and triples are not needed in model files anymore...
* There is a change in code for `--search_task dep_parser` mode. They generate their own pairs and triples in this module. I reflected that in my interactions array. Corresponding file with test results (search_dep_parser.stderr) was adjusted. It was changed twice as I found FNV bug after first change.
* 2 more test files were chenged. Both for GD: 0002a.stderr and redefine.stderr.
* twice I've synced my fork with upstream codebase.

TODO:
* I need to adapt audit_pairs() audit_triples for interactions. Most probably that will be a callback function for GD::foreach_feature(). 
* I shall do good performance tests after that. I expect tough designs at this step.

trufanov-nok:
Ok, good news is that I found out that I can omit stride_shift usage in FNV hash and now it almost as fast as the original hash. I've tested Modified and Original VW on previously mentioned [dataset](https://yadi.sk/d/9I4n1f6LfddoR) (90000 examples). The command line was:
```
(time ./vowpal_wabbit/vowpalwabbit/vw data/test --redefine q:=: -q qq --cubic qqq --passes 10 --early_terminate 100 -c -k --permutations) |& tail -n 3 > test.res
(time ./vowpal_wabbit_current/vowpalwabbit/vw data/test --redefine q:=: -q qq --cubic qqq --passes 10 --early_terminate 100 -c -k) |& tail -n 3 >> test.res
.. repeated total 5 times
```
So it runs modified and original vw one after another on same dataset with --permutations for 900.000 examples. That takes approx 30 min to complete.
The results (in sec):
```
Modified VW		1		2		3		4		5
real	180,703	167,807	169,195	169,731	167,353
user	184,252	170,888	172,392	172,644	169,92
sys	3,256	3,288	3,112	3,18	3,448
					
Original VW					
real	171,213	170,243	170,107	170,389	168,193
user	173,964	172,688	172,832	172,956	170,94
sys	3,38	3,192	2,972	3,156	3,376
					
TOTAL:
Modified VW	Mean	StDev			
real	170,9578	5,5340660639			
user	174,0192	5,8278856543			
sys	3,2568	0,1268826229			
					
					
Original VW		Mean	StDev				
real	170,029	1,1129052071			
user	172,676	1,0926481593			
sys	3,2152	0,1704441257			
```

As for me they looks equal enough.  

Bad news is that generic non-recursieve loop under same test case gives about 250 sec and 235 after microoptimization. This is 30% worse than nested `for` loops.
I tried cache that filled during first function launch - it made things even worse.
Looks like only cache generated at parser thread allows to improve performance.

Another bad news is that inline recursion gives same ~250 sec. I tried to play with gcc optimization params like `max-inline-recursive-depth` - they either got ignored or didn't improve result significantly. Or I made something wrong.

My proposal is following:
I can hardcode nested loops for interaction lengths from 2 to 5-10 and left microoptimized non-recursive lop for longer interactions. The code might be a bit bulky, but at least we'll be sure it's max fast. Or - I can even remove generic for loop and add limit on max interaction length acceptable by --interactions, for ex. <10. What do you think?

JohnLangford:
The timings do look good---we can accept a 1% hit for a better interaction function generally.  W.r.t loop unrolling, I'd suggest just doing pairs and triples explicitly, then having a slower generic function for other choices.  This covers the common case well and avoids much code bloat.

The semantics of --permutations is not clear.   We need a better way to describe it.

I think you need to change the way offset is used from: 
https://github.com/JohnLangford/vowpal_wabbit/pull/610/files#diff-067e797a2cbf8bc6771249b261fac82bR146
to:
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/gd.h#L77

so that multipredict:
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/gd.h#L26
works properly.

multipredict is a significant optimization (factor 7?) improvement when you need to make multipredictions with the same features at different offsets.

trufanov-nok:
>I think you need to change the way offset is used

Will do. Looks like offset usage has changed since Feb and I didn't noticed that.

>The semantics of --permutations is not clear

We already have this new parameter:
```
("leave_duplicate_interactions", "Don't remove interactions with duplicate combinations of 
namespaces. For ex. this is a duplicate: '-q ab -q ba' and a lot more in '-q ::'.")
```
How about renaming --permutations to:
```
("leave_duplicate_features", "Generate synthetic features from interactions that differ only
 by order of interacting features. For example -'q aa' will generate 16 synthetic features if
 namespace 'a' contains 4 features of weight 1. If not specified only 6 new features will be
 generated (10 if all weights != 1). Details could be found in documentation.")
```
Better description is welcomed as english is not my native language.

>I'd suggest just doing pairs and triples explicitly

Then all i need is to implement audit_interactions() function and doubletest everything.
Btw, how undesirable is addition of a new field to [feature struct](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/example.h#L26) (it currently consists only 2 fields: index and weight). Thinking about audition I found out that it could be done with passing of additional function to existing foreach_feature() if every features will hold a pointer to corresponding audit_data or NULL. Otherwise it seems i'll need to fork foreach_feature() and tweak it for audit_data.

JohnLangford:
Adding a new element to the feature struct is pretty undesirable in general.   Yeah, this is pretty crappy for foreach_feature().  Maybe instead of forking, you can abstract audit vs. actual features via an additional template parameter?

trufanov-nok:
The audit functions were replaced.
Current benchmark is:
```
		VW_new_mean	VW_new_SD	VW_old_mean	VW_old_sd
real	172,5126	7,5205410244	169,2312	10,9915115976
user	175,6224	7,9026216409	172,1072	11,2056505032
sys	3,2856	0,1017683644	3,2392	0,1324658447
```

Audit uses same `generate_features` function passing additional parameters. The function was generalised to iterate both `v_array<feature>*` and `v_array<audit_data>*`.
Additional function should be called in case of audit_data and eliminated by compiler optimizer in case of features. So in theory there shouldn't be any new code added in `feature` variant of template.


As for audit functionality itself:   
* I'm adding audit_data for Constant feature in parser.cc. 
```
if (all.audit || all.hash_inv) ae->audit_features[constant_namespace].push_back({nullptr,(char*)"Constant",(uint32_t)constant, 1.,false});
```
It doesn't own string memory and simplify audit function code a lot as now I can assume that all features has audit_data and I don't need to have feature* for audit_data == NULL (which was a case of consant feature).  
* Instead of `^` I'm using `*` as interaction between features. Like 'ns1^ft1*ns2^ft2:addr:weight:val@val'. Previously `^` was used in both cases: to separate ns and ft and to separate two interacting features.
* Namespaces titles are printed as is - ignoring redefinitions applied. I think it's more clear.

Examples.

New audit output for `--redefine q:=: -q qq --cubic qqq`:
```
0.000000 1000009418151094273
        b^banner_pos_0*m^device_conn_type_2*w^C21_79:42950:1:0@0        b^banner_pos_0*l^device_type_1*s^C18_0:17271:1:0@0      b^banner_pos_0*l^device_type_1*t^C19_35:76597:1:0@0     b^banner_pos_0*l^device_type_1*q^C20_-1:27036:1:0@0     b^banner_pos_0*l^device_type_1*w^C21_79:147488:1:0@0    b^banner_pos_0*m^device_conn_type_2*n^C14_15706:107359:1:0@0    b^banner_pos_0*m^device_conn_type_2*o^C15_320:54207:1:0@0       b^banner_pos_0*m^device_conn_type_2*p^C16_50:95738:1:0@0
```
Old audit output:
```
0.000000 1000009418151094273
        l^device_type_1^w^C21_79^j^device_ip_ddd2926e:208970:1:0@0      l^device_type_1^w^C21_79^s^C18_0:73957:1:0@0    l^device_type_1^w^C21_79^r^C17_1722:122269:1:0@0        l^device_type_1^w^C21_79^p^C16_50:127374:1:0@0  l^device_type_1^w^C21_79^o^C15_320:217035:1:0@0 l^device_type_1^w^C21_79^n^C14_15706:139051:1:0@0
       l^device_type_1^w^C21_79^m^device_conn_type_2:242856:1:0@0      l^device_type_1^w^C21_79^l^device_type_1:126962:1:0@0   l^device_type_1^w^C21_79^k^device_model_44956a24:139826:1:0@0   l^device_type_1^w^C21_79^t^C19_35:14503:1:0@0 
```

Features order and hash differ as hash function was changed to FNV.

New `--invert_hash`

```
Version 7.10.2
Min label:-1.000000
Max label:1.000000
bits:18
1 pairs: qq 
1 triples: qqq 
2 interactions: len: 2 interaction: qq len: 3 interaction: qqq 
lda:0
0 ngram: 
0 skip: 
options:
:0
Constant:116060:-0.000286
U^holiday:6519:0.000286
U^holiday*U^holiday:3874:-0.000286
U^holiday*U^holiday*U^holiday:248817:0.000286
U^holiday*U^holiday*a^C1_1002:236169:-0.000317
U^holiday*U^holiday*a^C1_1005:170532:-0.000182
```

old `--invert_hash`

```
Version 7.10.1
Min label:-1.000000
Max label:1.000000
bits:18
1 pairs: qq 
1 triples: qqq 
lda:0
0 ngram: 
0 skip: 
options:
:0
Constant:116060:-0.000039
U^holiday:6519:0.000039
U^holiday^U^holiday:133138:-0.000039
U^holiday^U^holiday^U^holiday:148375:0.000023
U^holiday^U^holiday^Y^wday:102473:-0.000019
U^holiday^U^holiday^a^C1_1002:160303:-0.000055
U^holiday^U^holiday^a^C1_1005:225986:0.000007
```


It looks It's only left to polish code and add few more comments. Perhaps benchmark it for few more cases.
Btw, what about semantics of --permutations proposed in previous post?

trufanov-nok:
Well, it seems I've done with `eval_count_of_generated_ft()` The combinatorics used to evaluate number of generated features with new interactions algorithm and recurrent formulas to calc their sum of weights are described in this blog [post](http://trufml.blogspot.ru/2015/05/math-behind-evalcountofgeneratedft-in-vw.html). Its URL is included in code comments.  
I've also added definition DEBUG_EVAL_COUNT_OF_GEN_FT in interactions.cc that will automatically compare results obtained with help of `eval_count_of_generated_ft()` with results got by simple launching `generate_interactions` with counting callback function. The results match now for all auto tests. The flag is commented out by default for performance reasons.

JohnLangford:
I wanted to pull this down and do some testing myself because this is a pretty large patch.  Unfortunately, the testing did not go well--I'm seeing a factor of 2 hit for quadratic features.  To replicate:

wget http://hunch.net/~jl/rcv1.gz
vw -c rcv1.gz
time vw -c rcv1.gz -q ff 

With the current master on my desktop the time is 25s.  On the other hand, when I incorporate this patch, it becomes 48s.  Can you replicate this?  And if so, do you know what's happening?  

I don't want to delay this patch any further than necessary.  Using the existing system for quadratic and cubic features and the new system for general features is a fine outcome.

trufanov-nok:
Will check tomorrow.

trufanov-nok:
Perhaps its an effect of `-c`. Could you doublecheck with `-c -k`?

JohnLangford:
-c -k causes the parser to take more time because it must recreate the 
cache file.  Using just -c stresses the learning core more and is the 
faster option.

-John

trufanov-nok:
I mean: is there no possibility that you tested both versions with one
cache file? In this case second launch will benefit from cache while first
be slower as it creates cache.

2015-05-24 1:37 GMT+04:00 John <notifications@github.com>:

> -c -k causes the parser to take more time because it must recreate the
> cache file. Using just -c stresses the learning core more and is the
> faster option.
>
> -John
>
> On 05/23/2015 05:17 PM, Alexander Trufanov wrote:
> >
> > Perhaps its an effect of |-c|. Could you doublecheck with |-c -k|?
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/JohnLangford/vowpal_wabbit/pull/610#issuecomment-104945112
> >.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/JohnLangford/vowpal_wabbit/pull/610#issuecomment-104946215>
> .
>



-- 
With best regards,
Alexander Trufanov

JohnLangford:
I tested both with the same cache file _after_ I created the cache 
file.  cache file creation time was not included in the timings.

-John

trufanov-nok:
Well, it's interesting, If you're using the same command line "time vw -c
rcv1.gz -q ff " for both then patched vw is supposed to be twice faster
than the old code. Bcs if you don't use --permutations flag it will
generate twice less interactions. And I'm getting 53 sec vs 24 sec with
this dataset that is what i'm expecting to find. Patched is faster.
Could you publish your console output on pastebin.com? I wan't make sure
the number of features is twice smaller for patched case.

2015-05-24 1:45 GMT+04:00 John <notifications@github.com>:

> I tested both with the same cache file _after_ I created the cache
> file. cache file creation time was not included in the timings.
>
> -John
>
>
> On 05/23/2015 05:41 PM, Alexander Trufanov wrote:
> > I mean: is there no possibility that you tested both versions with one
> > cache file? In this case second launch will benefit from cache while
> first
> > be slower as it creates cache.
> >
> > 2015-05-24 1:37 GMT+04:00 John <notifications@github.com>:
> >
> > > -c -k causes the parser to take more time because it must recreate the
> > > cache file. Using just -c stresses the learning core more and is the
> > > faster option.
> > >
> > > -John
> > >
> > > On 05/23/2015 05:17 PM, Alexander Trufanov wrote:
> > > >
> > > > Perhaps its an effect of |-c|. Could you doublecheck with |-c -k|?
> > > >
> > > > —
> > > > Reply to this email directly or view it on GitHub
> > > > <
> > >
> >
> https://github.com/JohnLangford/vowpal_wabbit/pull/610#issuecomment-104945112
> > > >.
> > >
> > > >
> > >
> > > —
> > > Reply to this email directly or view it on GitHub
> > >
> > <
> https://github.com/JohnLangford/vowpal_wabbit/pull/610#issuecomment-104946215
> >
> > > .
> > >
> >
> >
> >
> > --
> > With best regards,
> > Alexander Trufanov
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/JohnLangford/vowpal_wabbit/pull/610#issuecomment-104946980
> >.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/JohnLangford/vowpal_wabbit/pull/610#issuecomment-104947136>
> .
>



-- 
With best regards,
Alexander Trufanov

JohnLangford:
Huh, you're right.  I accidentally switched the cases in my head.   
Merged in.

All, I just accepted a _big_ patch which is a systematic redo of 
interactions (-q, --cubic, and now all higher order interactions).  This 
means that models using these things will be incompatible.  The 
advantages of this patch are:
(a) interactions run faster in many cases by eliminating redundant pairs.
(b) interactions work a little bit better in many cases via a slightly 
better hash.
(c) interactions can be done to any order.

-John

PR #612, opened 2015-05-05
Online boosting

beygel:
This pull request provides an implementation of online boosting algorithms in the following two papers: Beygelzimer, Kale, Luo: Optimal and adaptive algorithms for online boosting, ICML-2015, and Chen, Lin, and Lu: An online boosting algorithm with theoretical justifications, ICML-2012.


JohnLangford:
Can you do some tweaks?

(1) Remove -O2 from the Makefile.  We have -O3 in OPTIM_FLAGS.  Also, remove -ffast-math.  We decided that the minor speed improvement wasn't worth the cross-platform variance.

(2) Remove the extra printout on line 794 of gd.cc

(3) What kinds of task is the online boosting code supposed to address?  It's not clear that it's placed in the right position in the reduction stack.

(4) There is a reference to decision_stump.h in Makefile.am which is not met.  Please remove.

PR #613, opened 2015-05-06
added static overloads for hash_space & hash_feature to remove dependenc...

lhoang29:
...e on vw instance

### DIFF
--- a/cs_test/Program.cs
+++ b/cs_test/Program.cs
@@ -49,26 +49,26 @@ namespace cs_test
 
             IntPtr featureSpacePtr = pinnedFeatureSpace.AddrOfPinnedObject();
 
-            uint snum = VowpalWabbitInterface.HashSpace(vw, "s");
+            uint snum = VowpalWabbitInterface.HashSpace("s");
             featureSpace[0].name = (byte)'s';
-            sfeatures[0].weight_index = VowpalWabbitInterface.HashFeature(vw, "p^the_man", snum);
+            sfeatures[0].weight_index = VowpalWabbitInterface.HashFeature("p^the_man", snum);
             sfeatures[0].x = 1;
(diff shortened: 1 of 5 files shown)


JohnLangford:
Something is breaking in the build.  Can you track it down?

https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/61406830

lhoang29:
ok it should be fixed now. python function export didn't like 2 different function overloads. I just renamed the new one.

JohnLangford:
The interface is fine, but I'm worried about the computational structure.  getHasher is called for every single feature hashed, right?   That's very different from calling it once per program invocation.  What is the computational hit compared to using a predefined hash function?  

https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parse_primitives.cc#L56

lhoang29:
The hit seems nonexistent based on my timing below (microsecs). This makes sense since getHasher is simply an if else statement (vs indexing into vw structure). I'm actually more interested in the cost of marshaling data between language boundaries that exists in both of these variants. However, let's see whether this is the bottleneck first before trying to optimize further. 

calls = 100,000

hash_space before: 830872
hash_space after: 835672
hash_feature before: 835193
hash_feature after: 828950

Some more details on the code:

**hash_space before**: uint snum2 = VowpalWabbitInterface.HashSpace(**vw**, "s");
**hash_space after**: uint snum1 = VowpalWabbitInterface.HashSpace("s");

const string s = "w^thew^man\u0394";
**hash_feature before**: uint weight_index1 = VowpalWabbitInterface.HashFeature(**vw**, s, snum);
**hash_feature after**: uint weight_index2 = VowpalWabbitInterface.HashFeature(s, snum);

PR #621, opened 2015-05-11
Interact

alekh:


JohnLangford:
Some suggestions:

1) Remove all references to stride & stride_shift, because the logic is sound without them. 
2) Use weight_mask to make things work with hashes.
3) Use push_many() rather than push_back()
4) Move setup() in the reduction stack to just before the ldf setup is defined.

JohnLangford:
The build is breaking. https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/62253292  

Can you fix?

alekh:
Ah, sorry I committed too quickly. Will fix tomorrow morning.

Alekh

alekh:
Fixed.

Alekh

JohnLangford:
This has changes to csoaa.cc now?

alekh:
Yes, I pushed the changes I mentioned in the earlier email to you. For
returning multipredict. Basically this branch now has code in the state I
mentioned in my other email to you. Presumably you can just pull changes to
interact.cc if you don't want the others right now?

Alekh

JohnLangford:
It's not so easy to pull a subset of changes.  Can you send me the precise
email again?  And are you doing further edits for multilabel returns?

-John

alekh:
Here is the email I sent earlier. The thing I am unsure about below is
whether the newline example is really being pushed into ec_seq. I also have
some other doubts which blocked me from proceeding. But this is mostly
done, and all that is needed is to grab the indices from the sorted list
(in scores v_array in ldf struct) and return them. Maybe this is quick for
you? Leon will be here soon and I will probably spend an hour or two with
him.


Original email below.


I am not sure where it happens, because I don’t see the newline example
being pushed into ec_seq in predict_or_learn.



I think I also don’t know what is the right thing to do following learn in
do_actual_learning.



I suspect it might be best for you to take a look. I have code which
compiles currently, and checked into my fork on github under interact
branch. It implements the flag in ldf, maintains the predicted scores and
sorts on them. All that is needed is to return them action indices based on
the sorted order in the right way.



Btw, I might have to chat with Leon this afternoon to work on the ICML
camera ready, so I might be away for a bit.



Alekh

alekh:
PS: if you aren't quite sure about what I mean, the best thing might be to
have a quick 5 minute phone call now, before Leon gets here.

Alekh

JohnLangford:
There are some other issues with multilabel that I'm working on now and
want to deal with before adding multilabel to csoaa.

-John

JohnLangford:
(If you back out your changes or drop them into a separate branch, I think
the remainder is ready to go.)

-John

alekh:
Ah, so you were able to merge it? Sorry I had to step away to talk to Leon.

Alekh

JohnLangford:
Yep.  I just needed to read through and understand it.

-John

trufanov-nok:
Hi, Alekh
Cpp static analyzer is barking on interact.cc [line 110](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/interact.cc#L110) as calloc'ed `data` variable won't be freed in case this `if` statement break function execution. Could you pls fix?

PR #635, opened 2015-05-15
fixed windows build due to missing files

lhoang29:


lhoang29:
hmm, not sure why this is failing. built for me locally

PR #637, opened 2015-05-15
Enable parallel build and verbose logging

sharatsc:
Couple of config optimizations
a) shallow copy to speed up build start
b) parallel builds to enable faster run time
c) detailed logging.
d) removed my e-mail since README.md has a build badge.


sharatsc:
Tweaking it still. Will resend it.

PR #639, opened 2015-05-19
vw-hypersearch: make VW detection on Windows more safe/robust ('win' …

arielf:
Problem with previous implementations:
1) `win` matches `Darwin` too - which we don't want
2) `:` matches drive prefixes like `C:`
Also, to make it more less prone to Windows vs Unix differences, we now proceed to try calling `vw` anyway even when we can't find the executable.  By changing the `die` to a `warn`.


JohnLangford:
Merging in, thanks for staying on this.

PR #644, opened 2015-05-24
Interact

alekh:


JohnLangford:
I'd rather keep struct cb in the .cc file for information containment reasons.  Can you do that?

alekh:
Hey John,

I have pushed an update. There was an issue in my function definition. I
was trying to take as input an argument of type cb, which required exposing
the structure in the header. I now realize that would be incompatible with
its use in cb_adf, so I have exposed inner objects which are common to the
two. An alternative would be for both of them to use a common substructure.

There was a correctness question which came up for me while looking at
cb_adf. The class index created in the cost sensitive label for ips is 0 in
gen_example_ips. Is this correct? I suppose since we use the same regressor
for each action, the action id is meaningless in this case and all that
matters is the cost on that line?

Alekh

JohnLangford:
I'm skeptical this will work effectively.  

(1) The semantics of the input parameters are too unclear, so application errors are expected.
(2) The definition of 'action' is nominal in the multiline version
(3) The definition of cb_type is not shared across the multiline and single line versions.

Can you unify the definition of CB_TYPE also to prevent bugs?

PR #645, opened 2015-05-25
updates+fixes in README.md for Ubuntu builds

arielf:


trufanov-nok:
btw, regarding unused things. I'm was working with vw source with QtCreator and its toolchain. Which uses gcc with default params. And was getting ~400 warnings every build. Most of them are unused function arguments. I think there shall be about 50 such functions. And their effect multiplied as some of them are templated or inline.  
I tremendously tempted to fix these warnings and decrease their number to a dozen or so. By commenting out argument names. Like that:  
`inline void noop(void* data) {} -> inline void noop(void* /*data*/) {}`  
But there are 2 problems:  
1. That should be manual work as some warnings are generated by arguments which are not used in Linux code but used in Win code (code switched with typedef). Actually, not so much work.
2. That will affect dozen cpp files that could make any pending PRs unmergable.

How do you think, would it be a desirable patch? Or it's better to not touch this?

JohnLangford:
I'd say 'go for it', and would suggest simple removing unused function
arguments rather than commenting out.  Code cleanups like this a valuable
for sanity.

-John

PR #653, opened 2015-05-26
Fix 2 clang warnings: unsigned comparison to int in svrg.cc & char ar…

arielf:
…ray index in parse_args.cc


JohnLangford:
Is the change to svrg.cc valid?  I believe it assigns a -1.

arielf:
Thanks for the note.  I missed that. `svrg.cc` indeed initializes `s.prev_pass = -1`.

The problem is that it also assigns `pass = s.all->passes_complete` where RHS is a `size_t`.

When assigning -1 to an unsigned int we get `(UINT_MAX - 1)` which may work too for 'first time' in loop detection.

Anyway, just to be safe, I fixed the warning differently (preserving the `int` type in `struct svrg: pass`, and casting to `int` in the later assignment). This should preserve all the current semantics.  Sorry!

PR #656, opened 2015-05-27
Making JNI layer better

jmorra:
In this pull request I have made the JNI layer more Java like by having a close method and extending Closable.  This more naturally follows the Java paradigm of other classes that need to be closed.  This also allows for try with resources in Java 7.

In addition I have made the Java layer thread safe by making use of an AtomicBoolean so that closing can only happen once.

Finally I have removed the company specific information in the pom and pointed instead to the original VW repository.


EmergentOrder:
Can you publish this via Sonatype OSS?

JohnLangford:
Having things in Sonatype does seem desirable.  Suggestions?

-John

jmorra:
I totally agree, but I think this is trickier than your typical build because the jar itself is OS dependent.  Here would be my suggestions
1.  John (or his designee) control the release cycle for Sonatype
2.  Designate one or more OS's that we are supporting
3.  Find some way to build for those OS's with a CI (maybe Travis) using the command mvn clean deploy -Dos.version=$(uname).  This will allow for artifact names that are much cleaner than they are currently.  The reason I chose to use os.version and not os.name in the artifact ID is because os.name has spaces and produces an invalid artifact.

I think we should also agree upon a naming schema.  I am partial to no company specific group or artifact name as it is an open source project.

JohnLangford:
I will not be doing a release cycle for Sonatype---I'm both too 
overloaded and not familiar enough with it.

So, the question is: Who wants to?

-John

jmorra:
I just had a long talk with John, and here is our current plan.  First of all, I think it's important that there be one entity which has the private key to push to Sonatype and I think that John should control that key as this is his project.  I also think it's important that we make the artifact not OS dependent.  I'm currently working on a build script that will make use of Docker to try and do that.  The idea is that we'll build from an OSX machine and then compile the library with static linking on a 32 Docker VM.  I'll include both of these inside the generated Maven artifact and then load the appropriate library based on the system OS.  This will require Travis to give us access to an OSX box, which hopefully we can get them to do.  If this is all successful then we can follow the post outlined [here|https://gist.github.com/neothemachine/4060735] and get Travis to push snapshot builds to Sonatype.  We can then devise some kind of release cycle with Maven.  I think this is our best option because it removes the infectious nature of this dependency.  Once I have something that I'm confident works I'll commit more files here.

JohnLangford:
An OS-independent library seems clearly desirable. 

A fully automated solution seems clearly desirable.  

It may be desirable to include part of this patch now (?), depending on how long the above take (seems uncertain).

jmorra:
I think I might be able to get a bunch done today.  Let me see what I can finish today and then I'll let you know more by the end of the day.

EmergentOrder:
I was thinking Maven classifiers at first, but I like the idea of a single artifact. Sounds like a good plan, thanks!

jmorra:
I've been working on this for the most part of the day and have some disappointing news to report.  According to [here](http://stackoverflow.com/questions/24493337/linking-static-library-with-jni) and [here](http://openjdk.java.net/jeps/178) it does not appear possible to use JNI with a statically linked library.  I was able to get some semblance of a build script going that I've reproduced below, but the JNI layer still doesn't work for me.  The below script is my static build script.  It's trivial to convert this to a shared build script by removing some lines.

```
brew tap phinze/homebrew-cask
brew install brew-cask
brew cask install virtualbox
brew install boot2docker
brew install docker
boot2docker init
boot2docker up
# The boot2docker command outputs some exports that you have to run here that are machine dependent

docker pull 32bit/ubuntu:14.04
docker pull ubuntu:14.04

cat Makefile | perl -pe 's/-fPIC//g' | perl -pe 's/cd java(.+)things/cd java\1-f Makefile.static things/g' > Makefile.static
cat java/Makefile | perl -pe 's/-fPIC//g' | perl -pe 's/\-shared/-static/g' | perl -pe 's/\$\</\$\^/g' > java/Makefile.static

docker run -v $(pwd):/vowpal_wabbit 32bit/ubuntu:14.04 /bin/bash -c "\
apt-get install -qq software-properties-common; \
add-apt-repository -y ppa:ubuntu-toolchain-r/test; \
apt-get update; \
apt-get install -qq g++; \
apt-get install -qq make; \
apt-get install -qq libboost-all-dev; \
apt-get install -qq default-jdk; \
export JAVA_HOME=/usr/lib/jvm/java-7-openjdk-i386; \
cd /vowpal_wabbit; \
make clean; \
make -f Makefile.static; \
mv java/target/vw_jni.lib java/target/vw_jni.linux.i386.lib"

docker run -v $(pwd):/vowpal_wabbit ubuntu:14.04 /bin/bash -c "\
apt-get install -qq software-properties-common; \
add-apt-repository -y ppa:ubuntu-toolchain-r/test; \
apt-get update; \
apt-get install -qq g++; \
apt-get install -qq make; \
apt-get install -qq libboost-all-dev; \
apt-get install -qq default-jdk; \
export JAVA_HOME=/usr/lib/jvm/java-7-openjdk-amd64; \
cd /vowpal_wabbit; \
make clean; \
make -f Makefile.static; \
mv java/target/vw_jni.lib java/target/vw_jni.linux.amd64.lib"

rm Makefile.static java/Makefile.static

make clean
make
mv java/target/vw_jni.lib java/target/vw_jni.mac.x86_64.lib
```

When trying these artifacts the JNI layer complains about the static linking.  When trying the build script with the shared linking the JNI layer complains that the wrong libboost program options library is present.

I purposefully tested this on an older Scientific Linux version to see if that would succeed and it does not.  At this point I'm not sure what to do short of trying Java 8.  Does anyone have any suggestions?

JohnLangford:
I take it Java-8 isn't standard enough yet?

If so, what is plan B for Java 7?  Punt on an OS-independent library and 
have multiple versions?

-John

jmorra:
I think requiring Java 8 is preferable to OS dependence. Another thing I
read last night is that it seems possible to build a dynamic library from a
static one. If I have time over the weekend I will try this as this seems
like a perfect solution. In the meantime if anyone else can take my build
script and add a line to build a dynamic library from the static one that
might close the loop on this process.

Otherwise I'm on vacation this week and will look more at this when I get
back.

jmorra:
I have some good news.  It looks like I’ve “succeeded” in making the VW JNI layer platform agnostic.  I put succeeded in quotes because I haven’t set out to do exactly what I want, but I feel pretty good about how it stands now (and I don’t know how to make it any better).  If you look at the pull request you’ll see a file called java/src/main/bin/build.sh.  This file contains the information for how I build VW for multiple platforms at once.  Because the JNI layer is required to dynamically link, a new version must be built for not only every Linux distro, but also every version of every distro.  Luckily Docker makes this process relatively easy.  As you can see from the script I have builds for ubuntu 12, ubuntu 14 (32 bit as well), red hat 6 and red hat 7 in addition to the native Mac build.  This script is designed to ONLY be run on a mac (and will fail otherwise).

The process is to now make as many of these Docker builds as necessary and package them all into the jar.  The jar now has no platform specific naming, so it’s infectious naming scheme is gone which is a big plus.  Also, if someone wants to add a new Linux distro, it should be pretty easy to do following the logic I’ve already laid out.  The big wholes in this project as I see it now are

1.  No guarantee this will work on all mac versions (and I don’t know how to do that, nor even test it).
2.  No Windows support

I honestly cannot afford to spend any more time on this now, but I would encourage the community to try and extend my process to make this more robust and support more platforms.

In addition to the build process I've updated the JNI layer a lot itself.  It now supports 2 forms of bulk data processing (arrays and lists), along with better exception handling, and most importantly, it now works with more than one instance.  I had a really nasty bug before where I had a global vw instance.  This is now handled by letting the Java side keep track of a pointer to the vw object.  This means that close must be called or a memory leak will occur, but this is made more obvious by implementing Closable.

I have also added a lot more tests to show that everything is working.

One more note is that during the creation of this I was getting frustrated and thought of abandoning it in favor of just including a statically linked vw binary and calling that from Java.  In order to determine if this was a good idea I made a simple performance test and these are the results on 1,000,000 very simple examples:

Milliseconds for JNI layer: 1187
Milliseconds for JNI batch list layer: 1109
Milliseconds for JNI batch layer: 1086
Milliseconds for external process: 12927

The last thing to decide is what the versioning scheme is (I think it should be identical to VW) and how this is going to get built and pushed to sonatype.  I really think that John should control that process, but we should discuss it too.

Also, I'd like to thank @deaktator for his help in this whole process.

jmorra:
I fixed the failing tests and now do the Maven versioning from configure.ac through the java Makefile.  I think at this point this is ready to be merged in.  The only outstanding work here is to check the parse_label in the predict function to make sure we only attempt to work with the currently valid types.  I'm still doing a little benchmarking tests on my side, so the code may change a little bit, but this is mostly finalized now.

jmorra:
@deaktator did some benchmarking and it appears that the bulk interfaces were in fact [slower](https://microbenchmarks.appspot.com/runs/817d246a-5f90-478a-bc27-d5912d2ff874#r:scenario.benchmarkSpec.methodName,scenario.benchmarkSpec.parameters.nExamples,scenario.benchmarkSpec.parameters.loss,scenario.benchmarkSpec.parameters.mutabilityPolicy) than the single run.  In lieu of this I'm removing the references to the bulk interface.  This makes the code simpler and apparently better.

JohnLangford:
This seems clearly better than what we have now for the Java interface, 
so I'm going to accept it.

This has the implication that Jonathan will be handling releases to 
Sonatype, at least in the short term---I hope that's ok with everyone.  
He has, created a nice script that does this in a pushbutton way, 
although it requires a mac to run.

-John

PR #657, opened 2015-05-27
fixing illegal memory access in generated features count evaluation func

trufanov-nok:
With help of valgrind and `-fmudflap` i've found a bug in my own module. It's in function that evaluates 
count of generated features and sum of their squared weights. It didn't affect its results.

It seems sometimes I can compare last element of v_array with v_array->end in a loop like
`for(i; i < arr->end;++i) compare(i, i+1)`
And accessing data of v_array->end is dangerous.
On practice that didn't cause any problems as (v_array->end-1 != v_array->end) with high probability. So I wouldn't notice it without valgrind.

Correctness of function output is checked by testing with enabled DEBUG_EVAL_COUNT_OF_GEN_FT


JohnLangford:
Another subtle one---thanks for catching.

-John

PR #661, opened 2015-05-28
fixes bug getting feature names when namespace is null

StuHorsman:
When namespace is null and a weight is specified, the weight is pasted as a continuous variable.


StuHorsman:
@slymore can you review?

slymore:
all good

StuHorsman:
I thought I'd split them out into seperate PRs as my second change was a new feature.

PR #663, opened 2015-05-29
Merged Alekh's fix for seg fault when double robust method is used

yitanms:


JohnLangford:
This is failing the cb_adf test.  Can you fix? 
https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/64516123

yitanms:
Thanks, John. I am looking at this. I am not clear about what this test (#84) is about. When I debugged it in Visual Studio using command line arguments: vw --cb_adf train-sets/cb_test.ldf, it works. How can I reproduce the failure in visual studio?

yitanms:
I kind of get some sense out of this test. Is it about comparing the actual output (on stderr) with the expected output ( in file train-sets/ref/cb_adf.stderr)? We made some changes in do_actual_learning() in cb_adf. which is probably the cause. Should I revert just this change below?

  if (isTest || !is_learn)
    {
      gen_cs_example_ips(data.ec_seq, data.cs_labels);//create test labels.
      call_predict_or_learn<false>(data, base, data.ec_seq);
    }
  else 
	  learn<CB_TYPE_DR>(data, base, data.ec_seq);  // the template parameter changed from CB_TYPE_IPS to CB_TYPE_DR

yitanms:
With cb_type specified as ips, the actual output matches from the expected. Is this test failure a false alarm?

yitanms:
Another way to fix this test failure is to change default reduction type to ips in cb_adf_setup. Currently DR is the default, decided as Alekh and me discussed the changes. Comments?

JohnLangford:
This looks decent, but can you fix a detail? 

Lines 492-512 are messed up because there is code duplication.  In particular, you should _always_ do: 
if (count(all.args.begin(), all.args.end(), "--csoaa_ldf") == 0 && count(all.args.begin(), all.args.end(), "--wap_ldf") == 0)
	{
 all.args.push_back("--csoaa_ldf");
 all.args.push_back("multiline");
}

Issue #651, opened 2015-05-26
A sufficiently large number seems to disable learning


amosr:
From what I can tell, if the first row has a number larger than 18446745900000000000, learning just doesn't happen.

Given the following input file:
```
1 | out:1       ignore:18446745900000000000
1 | out:1
1 | out:1
1 | out:1
1 | out:1
1 | out:1
1 | out:1
1 | out:1
```
Now, I'd expect to ignore "ignore" and just use out, or maybe a constant 1.

However when you run it, it just seems to give up and keep the model at 0.
```
$ vw data4/test
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = data4/test
num sources = 1
average    since         example     example  current  current  current
loss       last          counter      weight    label  predict features
1.000000   1.000000          1      1.0     1.0000   0.0000        2
1.000000   1.000000          2      2.0     1.0000   0.0000        2
1.000000   1.000000          4      4.0     1.0000   0.0000        2
1.000000   1.000000          8      8.0     1.0000   0.0000        2

finished run
number of examples per pass = 8
passes used = 1
weighted example sum = 8
weighted label sum = 8
average loss = 1
best constant = 1
best constant's loss = 0
total feature number = 16
```

And if you output the model file, we just get
```
Version 7.10.0
Min label:0.000000
Max label:1.000000
bits:18
0 pairs:
0 triples:
lda:0
0 ngram:
0 skip:
options:
:0
```

I should note that I haven't tried with latest master yet, only the version I got off homebrew (7.10.0). I figure this is worth documenting either way though, as my searching didn't turn up anything about it.

JohnLangford:
When I run this example, I see nan cropping up (as below).

Tracing, the problem is that large*large = inf and the normalized update 
makes this inf*0 = nan.

You should be warned that something is going wrong.  Why don't you see 
NAN in the printout?

-John

10:36AM humpty-219: ./vw temp
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = temp
num sources = 1
average  since         example        example  current  current current
loss     last          counter         weight    label  predict features
1.000000 1.000000            1            1.0   1.0000 0.0000        3
NAN prediction in example 2, forcing 0.000000
1.000000 1.000000            2            2.0   1.0000 0.0000        2
NAN prediction in example 3, forcing 0.000000
[...]

PR #667, opened 2015-05-30
Show NAN warning when update is not possible

amosr:
For some reason I am not getting NAN warnings for the predictions in #651. The actual predictions are being returned as proper 0 for me, but after some digging I noticed that pred_per_update is NAN in get_pred_per_update, so maybe putting an error there is OK too?
It looks like returning 0 instead of NaN here has the dubious property of letting training occur later. In some ways it's desirable to have it error for good, but I can also imagine arguments for "recovering" to some extent.

### DIFF
--- a/vowpalwabbit/gd.cc
+++ b/vowpalwabbit/gd.cc
@@ -467,6 +467,13 @@ template<bool sqrt_rate, bool feature_mask_off, size_t adaptive, size_t normaliz
     norm_data nd = {grad_squared, 0., 0., {g.neg_power_t, g.neg_norm_power}};
     
     foreach_feature<norm_data,pred_per_update_feature<sqrt_rate, feature_mask_off, adaptive, normalized, spare> >(all, ec, nd);
+
+    if (nanpattern(nd.pred_per_update)) {
+      nd.pred_per_update = 0;
+      cerr << "NAN prediction update in example " << all.sd->example_number + 1 << ", forcing " << nd.pred_per_update << endl;
+      return nd.pred_per_update;
+    }
+
     if(normalized) {
       g.all->normalized_sum_norm_x += ld.weight * nd.norm_x;
       g.total_weight += ld.weight;


JohnLangford:
I'm hesitant to include this because:
(a) This is a 'hot' part of the pipeline.
(b) This does not get at the root cause.  Why did the NAN warning for 
the prediction not trigger for you?

If we end up wanting to armor the update, the right place to do it is 
probably just before this line:
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/gd.cc#L522 

so that all update computations are checked.

-John

amosr:
Yes, fair enough.
Ah!
Ok, so -ffast-math in clang is doing some questionable optimisations. If I run make with
```
make CXXOPTIMIZE="-O2 -g"
```
I get the NAN prediction warnings, but with default or -ffast-math, I lose the warnings.

What do you think about removing -ffast-math for clang? Disabling it actually made one of my test runs faster: a 5GB file went from 19s to 17s, but I wouldn't put much weight on that.

As an aside, are you aware that "make test; make test" fails, because, I suppose, a daemon is left running and the next make test attempts to bind the same port? Do you know if this is OSX specific too?

JohnLangford:
Taking out -ffast-math seems like a good idea.  That's what we did with 
the default Makefile because the results were just too inconsistent to 
test across different platforms.

"make test; make test" works fine on linux.  Can you run "make test; 
netstat -tunlp; make test" (or equivalent) on a mac and see what's up 
with the daemon?

-John

JohnLangford:
Ping?

amosr:
Sorry, I've been overwhelmed with other stuff.
Umm - maybe I'm crazy. I swear "make test; make test" wasn't working before, but I just tried again and it's fine. I can't imagine anything would have changed since then..

JohnLangford:
There was a small tweak to the daemon mode test that maybe addressed this?

-John

amosr:
No, I haven't actually pulled since! So it must have just been something strange mucking up locally. Some weird OSX state hanging around, I guess, I don't know

JohnLangford:
OSX deprecates daemon() which we use, but there seems to be no 
alternative interface which prevents race conditions.

-John

amosr:
Ok, strange. I'll have a proper try to reproduce it later tonight and let you know how I go

amosr:
Hey, I wasn't able to reproduce the make test thing, so I'm not sure what was going on. I didn't try much, just a make clean and (make test && make test) a whole bunch of times, and it always worked. Not sure, maybe just ignore it unless it comes up again.

JohnLangford:
On 06/16/2015 06:06 PM, Amos Robinson wrote:
>
> Hey, I wasn't able to reproduce the make test thing, so I'm not sure 
> what was going on. I didn't try much, just a make clean and (make test 
> && make test) a whole bunch of times, and it always worked. Not sure, 
> maybe just ignore it unless it comes up again.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/JohnLangford/vowpal_wabbit/pull/667#issuecomment-112583600>.
>

Ok, thanks.

-John

PR #669, opened 2015-05-31
fix bug on Makefile for mac

marcovzla:
Currently there is a bug in the makefile that prevents us from building pyvw correctly
on mac os.

The problem is that BOOST_LIBRARY is set to "-L /usr/lib -L /usr/local/lib"
which is a problem because there is a libpython in /usr/lib that is not the one used to
compile boost, and that libpython hides the one in /usr/local/lib, which is the one we want.

The solution is to set BOOST_LIBRARY to "-L /usr/local/lib" so that the correct
libpython can be found.


JohnLangford:
Merged in, thanks.

-John

PR #670, opened 2015-06-01
Osx python make

plandes:
adding fixes to install on OSX with the python wrapper

done at NAACL conference


JohnLangford:
Please check the master---I believe this is fixed as of a few minutes ago.  (Also, this patch is huge---if it isn't fixed, I'd much appreciate a smaller patch.)

plandes:
I somehow missed this. Just to clarify I believe it looked big because it was done on the 7.10 tag and when you merged it brought back that old tag.  I wasn't surgical--I was on a mission to get it to work for the tutorial.

Thanks.

JohnLangford:
Does the master work for you now?

-John

plandes:
No.

$ make python
make: Nothing to be done for `python'.
$ cd python
$ make
cd ..
make python
make[1]: Nothing to be done for `python'.
$ make things
c++ -std=c++0x -I/System/Library/Frameworks/Python.framework/Versions/2.7/include/python2.7 -I/System/Library/Frameworks/Python.framework/Versions/2.7/include/python2.7  -fPIC -c pylibvw.cc -o pylibvw.o
In file included from pylibvw.cc:1:
In file included from ./../vowpalwabbit/vw.h:7:
./../vowpalwabbit/global_data.h:14:10: fatal error: 'boost/program_options.hpp'
      file not found
#include <boost/program_options.hpp>
         ^
1 error generated.
make: *** [pylibvw.o] Error 1

marcovzla:
can you try `make clean` first?

plandes:
I wiped everything out and did a brand new clone for this test.  Your problem is your configure setup isn't creating the makefile in the python directory, which is why I modified it.

marcovzla:
isn't it already there?
https://github.com/JohnLangford/vowpal_wabbit/blob/master/python/Makefile

plandes:
Yes it is there, which is the problem.  You want the auto*/configure to generate that make file like the others, which I am pretty sure is how the root directory make files are being generated.  So autoconf/automake usually generates a ./configure script (which yours does) and then configure generates the make files.  python/Makefile is NOT generated and thus doesn't get the config it needs--in this case where boost lives.

Hope this helps.
Cheers!

JohnLangford:
There are multiple build systems in VW (why?  it's complex).   If you 
just use the default Makefile, does it work?

-John

plandes:
No. That's the output I sent.  I recommend applying the patches or copy the files in place I attached a few days ago.

JohnLangford:
I don't understand how 'make python' can say:

make: Nothing to be done for `python'.

with a clean git clone of the master.  That's certainly quite different 
from what my make does.

-John

plandes:
I couldn't tell you but this is what I got before I made any changes and my setup is fairly typical (i.e. I use brew, which is a common well supported package manager) and built the code for the first time.

I won't be very helpful with debugging someone else's solution.  I recommend you either work with them or implement my solution by copying the three files I attached in the earlier email.

Good luck.

Issue #673, opened 2015-06-04
Tricky segfaut when using --random_seed and -O3


trufanov-nok:
Thanks to @stephen-hoover and a big luck we've found another memory corruption.
It seems that while working on previous issue something  had damaged --quiet functionality (that behaviour disappeared after `make clean`) which let us realize that `make tests` 18 and 19 are actually cause segfaults.
That wasn't detected because these tests use `--quiet` and compared with empty output files.

After some debugging I was able to reproduce segfault in Linux and found out that:
1. segfaut happens only in presence of `--random_seed`
2. segfaut happens only if `-O3` flag was passed to GCC (another thing that hide this from me as I’m not using such flags in my QtCreator debug setup.)
3. segfaut happens after a call to `po::notify(vm);` in [best_constant.cc](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/best_constant.cc#L36)

It seems that `po::notify` must throw exceptions if command line is malformed and don't match options configuration but it didn't. It just corrupts memory.

It looks like i'm reparsing command line at best_constant.cc and this approach was implemented before stack of reductions. Nowadays I can just take a reference to all.vw and use it. This fixes the segfaut (at least for me, currently Stephen checks if this fixes it on Mac too. In his case segfaut happens during learning, so it might not). 

As it was in my code I put some efforts to find out how I made this mistake and found the real cause (imho). The problem happens if only `--random_seed` is passed to command line. It's not depend on other params, `random_seed` param name or its description. But it depends on fact that local variable address is [passed](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parse_args.cc#L864) to its constructor:
```
  size_t random_seed = 0;
  new_options(all, "VW options")
    ("random_seed", po::value<size_t>(&random_seed), "seed random number generator");
```

If you replace it with some member of `all` it won't segfault.

My hypothesis is that in case `--random_seed` value specified it tries to push its value to address in local stack (to local variable). It's fine for `parse_args` but cause memory corruption if `all.opts` is used for parsing in any other function. So either `all.opts` content should be modified before such usage or address from heap should be passed to `random_seed`. And somehow bad things didn't happen if you don't use `-O3`. I suspect memory corruption happens always but don't cause segfault.

[...]

PR #674, opened 2015-06-04
fix tricky segfaut when using --random_seed

trufanov-nok:
Reference #673


trufanov-nok:
Looks like there are more bugs. Stephen reports that his sample segfauts from time to time, but i can't reproduce that on my Kubuntu yet. And these segfaults are definitely related to boost::program_options. Perhaps it's boost version dependant. Will check.

JohnLangford:
This looks like an improvement as-is, so I'm merging what you've done so far.

PR #675, opened 2015-06-04
fixed neighbor feature bugs, removed NANs, edited notebook

hal3:


JohnLangford:
This is doing something nasty to the compiler:
https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/65588204

Can you look into it and fix?

-John

hal3:
very very weird. i made a tiny edit and it seems fine now.

hal3:
Ok I'm having a heck of a time getting the build to work on Windows. I finally got the compile to work, but there's a linking problem that I'm kind of stumped on. John -- do you have any insight?

hal3:
blech, finally working!!!

JohnLangford:
Something is missing on the linux side build now? 
https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/66749778

hal3:
fixed

JohnLangford:
It looks like there is some debugging code in ftrl.cc?  Using wasNAN?  
And in gd.cc?  And in log_multi.cc?

Having labeldict.h and label_dictionary.h also seems awkward.  Why both?

The use of a label dictionary for input classes seems nice.  Is there a 
way to make it work in parallel?

The use of recompute_hidden in nn.cc might be premature.   Is that 
working properly?

What's going on in oaa.cc?

-John

hal3:
> It looks like there is some debugging code in ftrl.cc? Using wasNAN?
> And in gd.cc? And in log_multi.cc?

cleaned.

> Having labeldict.h and label_dictionary.h also seems awkward. Why both?

renamed labeldict to namedlabels and removed redundancy.

> The use of a label dictionary for input classes seems nice. Is there a
> way to make it work in parallel?

it does work in parallel :)

> The use of recompute_hidden in nn.cc might be premature. Is that
> working properly?

yes it's working. it's only active for multilabel, which is when there's 
a lot of repeated computation when being reduced to from, eg, oaa. 
basically only used in multipredict.

> What's going on in oaa.cc?

what do you mean? which part?

JohnLangford:
Why is so much code going away in nn.cc?

I don't see allreduce getting called, so I'm skeptical that the labels will be consistent in a cluster setting.

hal3:
> Why is so much code going away in nn.cc?

there was "backup" code while i was debugging multipredict; it's all 
good so i killed the extra.

> I don't see allreduce getting called, so I'm skeptical that the labels
> will be consistent in a cluster setting.

yeah i was worried about this exact issue. the solution is:

labels are specified on the command line.

JohnLangford:
Allright, merged.

-John

PR #676, opened 2015-06-06
fix vw[<=7.8] model file compatibility

trufanov-nok:
This commit should fix model files compatibility broken in 7.9. It was broken at [this](https://github.com/JohnLangford/vowpal_wabbit/commit/c316ebe62a2a428ab13841195bf7dc5420f5a1db) commit.
just before 7.8 got tag 7.9. The problem is that `all.rank` field was removed from `vw` class and was excluded from saving in model header.
If we read dummy `uint32` at this point - at least simple 7.8 model could be successfully read in 7.10.2

Current solution reads dummy 32bit at place where rank was stored in model files with version <= 7.8.
Note: in case read value != 0 the warning is displayed. The user is asked to pass this value manually via command line.

### DIFF
--- a/vowpalwabbit/parse_regressor.cc
+++ b/vowpalwabbit/parse_regressor.cc
@@ -21,6 +21,7 @@ using namespace std;
 /* Define the last version where files are backward compatible. */
 #define LAST_COMPATIBLE_VERSION "6.1.3"
 #define VERSION_FILE_WITH_CUBIC "6.1.3"
+#define VERSION_FILE_WITH_RANK_IN_HEADER "7.8.0"
 #define VERSION_FILE_WITH_INTERACTIONS "7.10.2"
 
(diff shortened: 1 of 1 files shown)


JohnLangford:
This seems good.  For extra credit, you could add this to the all.file_options as per: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/global_data.h#L507 which is what the code currently does.

Windows seems to lack u_int_32.  https://ci.appveyor.com/project/JohnLangford/vowpal-wabbit/build/1.0.106  Can you fix?

trufanov-nok:
> For extra credit, you could add this to the all.file_options as per: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/global_data.h#L507 which is what the code currently does.

Pls take a look at #677

PR #677, opened 2015-06-06
better old model files rank processing

trufanov-nok:
Addition to #676 
In case rank: N is found in old model file and N != 0 vw will check if any --rank M was specified in command line. If it's not - it will add --rank N to it. `gd_mf_setup` with `missing_options` is called later so it will process parameter as usual.
If there is already --rank M in command line it just warns the user. I decided not to overwrite command line value. Frankly, vw will crash if N != M anyway.


JohnLangford:
Merged, thanks.

-John

Issue #649, opened 2015-05-25
The default namespace is not included in --quadratic=::


martinpopel:
While `echo '1 |f a b c' | vw -q :: --audit` works as expected and reports 7 features (Constant, a, b, c, a*b, a*c, b*c),  `echo '1 | a b c' | vw -q :: --audit` reports only 4 features (Constant, a, b, c).

Before merging PR #610, `echo '1 | a b c' | vw -q :: --audit` reported 13 features (as is now reported with `echo '1 |f a b c' | vw -q :: --audit --permutations`).

arielf:
Thanks for catching. Looks like some iterators over name-spaces+features need readjust/rewrite after #610.
Need a few tests for `--audit` as well.

trufanov-nok:
I'm quite surprised that 
> Before merging PR #610, echo '1 | a b c' | vw -q :: --audit reported 13 

The reason of 4 results is that in example '1 | a b c'  features belong to empty namespace (in fact it's namspace with character ' ' (space 0x20)). Which is treated as absence of any namespace. And namespace wildcard ':' doesn't include ' '. 

wildcard includes everything [between '!' and '~'](https://github.com/JohnLangford/vowpal_wabbit/blob/88bcd410103345d431bcdb5b80dc72157f84277b/vowpalwabbit/parse_args.cc#L429) except for ['|' and ':'](https://github.com/JohnLangford/vowpal_wabbit/blob/88bcd410103345d431bcdb5b80dc72157f84277b/vowpalwabbit/parse_args.cc#L63). Note: I'm referencing branch state at commit merged just before mine.  
And according to [ASCII table](https://en.wikipedia.org/wiki/ASCII#ASCII_printable_code_chart) space is just before '!'.

So old VW should generate 4 features for `echo '1 | a b c' | vw -q :: ` even if `-q ::` specified as well as current VW. Could you please check that?
If there are 4 features and 13 with `--audit` than it was a bug in old VW code which disappeared now.

The only way to interact features which are not belong to any namespace is to reassign them to some namespace: ` vw    --redefine a:=    -q aa`

trufanov-nok:
> So old VW should generate 4 features for echo '1 | a b c' | vw -q :: even if -q :: specified as well as current VW. Could you please check that?

I recall i have old VW copy. It generates 13 features. I'll take a look into it tomorrow as it's 2:30 AM here.

trufanov-nok:
[...]

Issue #648, opened 2015-05-25
--invert_hash with --oaa reports only the features for the first class


martinpopel:
The famous iris dataset contains 4 features for each example:
```
3 |f s_length:7.9 s_width:3.8 p_length:6.4 p_width:2.0
2 |f s_length:6.8 s_width:2.8 p_length:4.8 p_width:1.4
1 |f s_length:5.7 s_width:4.4 p_length:1.5 p_width:0.4
```
When using `-q ff`, there should be up to 16=4^2 quadratic features (or less with VW after #610) plus 4 simple features plus 1 constant features. With `--oaa=3` there should be three variants of each feature, thus (16+4+1)*3 = 63 features in total.

```
wget http://archive.ics.uci.edu/ml/machine-learning-databases/iris/iris.data
cat iris.data | perl -nle '@f=split /,/,$_; $class = /setosa/ ? 1 : /versi/ ? 2 : /virgi/ ? 3 : exit for($f[4]); printf "$class |f s_length:%s s_width:%s p_length:%s p_width:%s\n", @f;' > iris.vw
shuf iris.vw > iris.shuf.vw
vw -d iris.shuf.vw -c -k --loss_function=logistic --oaa=3 -q ff --passes=3 -f oaa.model
vw -d iris.shuf.vw -t -i oaa.model --invert_hash=model.invert --readable_model=model.readable
```

When I tried the experiment above with one-month-old VW (7d0ec30), I got 63 features in model.readable as expected. However, model.invert contained only 31 features, because the 16 quadratic features were present only in one version (for class1, because their hashes were divisible by 4).

Now I tried the same experiment with the newest VW from GitHub (003b50b), which contains the new mechanism for interactions ( #610 ), where interactions f1*f2 and f2*f1 are considered the same, so I would expect (4 + 4*3/2 + 4 + 1) * 3 = 45 total features.
model.readable contains 45 features.
model.invert contains 15 features (all divisible by 4, i.e. also the constant feature and simple features are only in one version):

```
Constant:202096:0.784281
f^p_length:106408:-0.163726
f^p_length*f^p_length:248528:-0.038985
f^p_length*f^p_width:158456:-0.126031
f^p_width:16256:-0.706956
f^p_width*f^p_width:248064:-0.285858
f^s_length:140528:0.059708
f^s_length*f^p_length:113272:-0.023495
f^s_length*f^p_width:6736:-0.096329
f^s_length*f^s_length:131360:0.002369
[...]

PR #679, opened 2015-06-07
Test for -q :: & --invert_hash + support of comparision of arbitrary files in RunTests

trufanov-nok:
I've tried to create a test case to cover 2 issues:
#648 - we don't have any test to compare --invert_hash results
#649 -  we don't have any test that uses namespaces with wildcards like -q ::

So I've added following command line as a test:
`echo -e '3 |f a b c |e x y z\n2 |f a y c |e x' | {VW} --oaa 3 -q :: --invert_hash inv_hash.cmp`

Notice, it uses `echo` and pipeline instead of file as dataset.
Still I've added 2 files - for stdout and  inv_hash.cmp.

Regarding the last one - it looks like RunTests script decides what to compare by filename extension and parameter name. For example, `*.predict` is compared with whatever is used after `-p`. `*.stderr` with vw stderr etc.
I've decided not to specify new file extension for `--invert_hash` (as it's the only case among 88 tests) but declare one general use extension.
Now RunTests supports *.cmp extension. It will find whatever ends with .cmp in VW arguments list (as it's generic I cant rely on parameter name) and compare it with *.cmp in list of reference files.

As I don't know Perl I would like to ask @arielf  to verify the code if John confirms this is the enhancement script needs. I especially worry about case when both `-p sometheng.predict` and `--whatever *.cmd` are declared in VW command line. Won't `-p` regexp empty command line before `cmd` regexp process it? I'm not sure how =~ works.


trufanov-nok:
It seems that Travis CI demonstrates inconsistence of my interactions filtering function. In one cases interaction '' A" is left and "A " in other. I'll try to fix that and include patch in this PR. In case this PR won't be possible to merge (there are might be also problems with weights comparison in model files that could fluctuates) I'll submit it as separate PR.

arielf:
Thanks Alex,

My fault for not abstracting the reference-files in a nicer way.  There are very few vw options that produce a user-specified readable text file that can be compared: (`-p, --readable_model, --invert_hash, --output_feature_regularizer_text `) .  I feel it would be cleaner if we could 1) support all of them in a consistent way and 2) support multiple instances on the same line so all of them can be compared if they appear in the test.

What would be a clean abstraction to support this?  Is the `*.cmp` naming convention the best choice?  What about multiple `*.cmp` files?

To answer your questions:
`if ($string =~ /regexp/) ...` in perl is simply a pattern-matching statement, where the first (leftmost) match applies, but you can also do it in some loop to support multiple matches:
   `while ($string =~ /regexp/g) ...`
The `g` modifier will iterate on all matches so you can capture sub-matches (say in `$1`) and use them in the loop.

trufanov-nok:
>  Is the *.cmp naming convention the best choice? 
nope
> What about multiple *.cmp files?
not supported.
My Perl knowledge doesn't allow me to implement better solution, I just though *.cmp approach is a bit better than extension supporting just invert_hash with same amount of code.

As for real generic approach: how about matching all files with reference files by extension. For example:
{VW} -d dataset.dat -p output.xxx --invert_hash 123.yyy
prediction_ref.xxx
invert_hash_ref.yyy

So .dat will be ignored as has no pair. xxx and yyy will be compared with their ref. .stderr and .stdout might be left as standard hardcoded extensions for VW's stderr and stdout.

trufanov-nok:
And now i'm running into a problem of comparing files containing weights which might differ in 10^-5 - Travis CI failed. That somehow solved for stderr will take a look into code.

trufanov-nok:
Hey, Ariel
Is it possible to allow control over lines slit in diff() before smart comparison (that takes precession into account) ? Suppose this would require passing additional argument (regexp string) to diff() and to diff_lenient_float() in it. By default it could be "/\s*\|\s*/". I would like to pass "/\s*\:\s*/" or even "/\s*\:[.+-0-9]*$/ in my last test case. My lines are in form of:
```
-e^x:125152:-0.078247
+e^x:125152:-0.078248
```
So only last piece should be compared. Of  course mismatch in 125152 will be more than precession so splitting just by ':' would be fine too.
And most inconvenient - such solution will require specification of new keyword in tests section that could let user to pass non-default regexp.
For example {SPLIT}[/\s*\:\s*/"] that could be parsed by `$line =~ /{SPLIT}[(\S+)]\b/`

arielf:
Alex,

I'm not sure we need a different split regexp.
If there's a difference in any outputs (quick default diff), `RunTests` falls back to 'fuzzy-compare' where small precision differences are ignored.  This second fuzzy diff is very slow, and (always) calling `diff` with:
```
 -y, --side-by-side
              output in two columns
```
which always uses the `    |    ` separator between the two versions.

What may be a problem here is the _width_ of the line (not sure) which can be increased with `-W <somenumber>` See this line in `RunTests`:
```
my $DiffOpts = '-N --minimal --suppress-common-lines --ignore-all-space --strip-trailing-cr --side-by-side -W 160';
```
I hope to have more time to work on this the coming week. A bit swamped at the moment.

trufanov-nok:
Ok, then I misunderstood the code. Then the word separators is the param I'm looking for. I'll try to do the adjustment by myself.
```
            # Break lines into tokens/words
            my (@w1) = split(' ', $line1);
            my (@w2) = split(' ', $line2);
```

trufanov-nok:
Heh, it works!
Just using ':' instead of space as words separator was enough to apply diff() with precession adjustments for float values to text model file. I've added an optional `{WORD_SPLIT=x}` param that parsed in case `*.cmp` file was found. Where x is a split value which is = ' ' by default.

Would be great if you take a look and approve the code.

Fell free to modify these parameters later if you wish.

trufanov-nok:
Ok, let's try global variable. I'm pretty fine with that.

JohnLangford:
I'm going to merge this one in, as it seems clearly superior.  W.r.t. 
wildcard pairs, truncation has the potential to ignore some change in 
the definition of the wildcard set which seems potentially unfortunate.

Logically, we really only want a hard diff on this line.  Is that doable?

-John

PR #681, opened 2015-06-08
New utility: utl/csv2vw: better CSV/TSV to vw converter

arielf:
Hi John,

I've seen attempts at converting csv to vw format in the past, but all of them fell short in one way or  another, so I finally wrote my own that does everything I needed correctly. I found it very useful in a recent Kaggle competition.

This new utility supports:
   - Optionally use header line for feature names (`-h` option)
   - If no  `-h` is used, will number features as `1..k` based on column number
   - Multiclass labels will be auto-converted to `1..k` if they are non-numeric e.g. `species: setosa,versicolor,virginica`
   - Categorical features are auto-converted to vw boolean `name=value`
   - Numerical features will use name:value
   - Numeric command-line arg allows specifiying the label column number (also: negative numbers conveniently support "from end of line" e.g: -1 is last column)
   - Auto-splitting on commas and/or tabs
   - Allows specifying (and overriding) the input separator as perl regexp (`-s <regexp>` option)

Thanks.

### DIFF
--- a/test/daemon-test.sh
+++ b/test/daemon-test.sh
@@ -45,14 +45,17 @@ else
 fi
 
 
-# A command and pattern that will unlikely to match anything but our own test
+# A command (+pattern) that is unlikely to match anything but our own test
 DaemonCmd="$VW -t -i $MODEL --daemon --num_children 1 --quiet --port $PORT"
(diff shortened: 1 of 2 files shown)


JohnLangford:
This looks good, thanks.  Merging in.

-John

Issue #680, opened 2015-06-07
Test 61 failure


trufanov-nok:
Few weeks ago I've noticed a small bug.  `make test` completes without a problem after first launch, but 2 `lt-vw` processes keeps running as background processes after that. So if you execute `make test` again you'll get 
```
RunTests: test 61: lenient-diff.tmp: line 1: fuzzy-match missing data on one of the sides. Can't compare
                                                                              > bind: Address already in use
RunTests: test 61: saving lenient diff in 'test-61.lenient-diff' for later inspection
RunTests: test 61: FAILED: ref(/dev/null) != stderr(stderr.tmp)
        cmd: /usr/bin/timeout 30 ./daemon-test.sh
```
You'll get this every time till you kill lt-vw processes manually.  As first launch goes fine Travis CI ignores this problem. That's quite annoying to kill these processes every-time if you run make tests often, could someone fix?

arielf:
I was looking at this and wasn't able to reproduce in my own (Ubuntu, desktop) env.

Also: I cannot find any 'lt-vw' string anywhere in the vowpalwabbit code or tests

Is this happening only in Travis?

Anyway, I will try to modify `test/daemon-test.sh` so killing of the daemon is more reliable (with a pid file)

arielf:
If I use --pid_file ... and only send SIGTERM (via kill) to the process matching the pid in the pidfile, I always get this:
```
11:18 go 455 vowpal_wabbit/arielf> ( cd test/ ; ./daemon-test.sh )
vw-daemon-test: OK
*** Error in `../vowpalwabbit/vw': free(): invalid pointer: 0x00007f215eb8a000 ***
```
So looks like we're still not shutting down properly on SIGTERM to the parent process.  Need to fix this first before we can try fixing `daemon-test.sh`.  Perhaps the two issues are related.

trufanov-nok:
> I cannot find any 'lt-vw' string anywhere

I suppose this is an effect of calling system `fork()` function.
I'm observing 2 `lt-vw` processes if I execute
`vw --daemon --num_children 1 --port 10000`   
and 10 processes if just ``vw --daemon``

arielf:
Thanks!
That explains it.  Maybe 'lt-' has something to do with linux threads

Although it doesn't happen on my system (Ubuntu 14.04).
When I call:

    `vw --daemon --num_children 1 --port 10000`

I only see two processes both called `vw`.  No `lt-vw` anywhere.

[...]

PR #682, opened 2015-06-09
daemon-test.sh: make stop_daemon() work nicely with libtool

arielf:
When using autoconf/libtool `vw` may be wrapped and appear as `.../.libs/lt-vw` - this causes `stop_deamon()` to fail to kill the process properly.

This pull-request solves issue https://github.com/JohnLangford/vowpal_wabbit/issues/680 (Test 61 failure) reported by Alex.

### DIFF
--- a/test/daemon-test.sh
+++ b/test/daemon-test.sh
@@ -45,14 +45,17 @@ else
 fi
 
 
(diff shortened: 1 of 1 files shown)


JohnLangford:
Thanks.

PR #683, opened 2015-06-09
Cb adf

alekh:


JohnLangford:
Instead of changing cb_algs.h, should we just be protecting the function call?

alekh:
I think it is the responsibility of the function in cb_algs to not mess up
the prediction of the example, since it has an interface that returns the
prediction as a float. It just seems confusing to me that a function that
returns a prediction would also mutate the example in the process, so I am
in the favor of the changed version.

Alekh

JohnLangford:
That's a fair point.... there should be just one interface for the 
prediction.  I'll go with this for now.

-John

PR #686, opened 2015-06-15
Finalizing JNI layer for first release

jmorra:
This is a finalization of the JNI layer for it's first release in version 7.11.  I am currently testing this layer extensively and there will be a few more issues that come up during that test.  The first ones I found here were

1.  Naming everything according to the correct maven structure
2.  Refactoring some classes
3.  Making boost exception handling more robust


JohnLangford:
Merged, thanks.

-John

PR #688, opened 2015-06-16
updated argument (#581) in demo

djstrong:
Changed `ftrl` argument and removed duplicated variables.


JohnLangford:
Merged, thanks.

-John

PR #694, opened 2015-06-16
JNI Safety Issues

jmorra:
I just realized that I can make a segfault by using multiple threads to access the same VW object.  This pull request fixes that by wrapping all calls in a [ReentrantLock](http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantLock.html).  After doing my basic load testing I realized that this does not add a lot of overhead (less than 10% speed decrease) but it adds a lot of safety.  In addition I added some safety against trying to return multi labels in the constructor.  John suggested that other types could produce multi labels as well, but my tests seemed to indicate otherwise, so I'm not sure what to do about that.


deaktator:
I think that you should use a `java.util.concurrent.locks.ReentrantReadWriteLock`.  This should allow you to have higher throughput on situations with much higher read volume than write volume because many threads calling predict can operate concurrently.

```java
import java.util.concurrent.locks.ReentrantReadWriteLock;

// ...

public class VW {
  private final ReentrantReadWriteLock lock;

  public VW(String command) {
    isOpen = new AtomicBoolean(true);
    isOpen = true;
    lock = new ReentrantReadWriteLock(true);  // Read about fairness policy to see if you want it.
    nativePointer = initialize(command);
  }

// ...

  public float predict(String example) {
    lock.readLock().lock();
    try {
      if (isOpen)
        return predict(example, false, nativePointer);
      else throw new IllegalStateException("Already closed.");
    }
    finally {
      lock.readLock().unlock();
    }
  }

// ...

  public float learn(String example) {
    lock.writeLock().lock();
    try {
      if (isOpen) {
        return predict(example, true, nativePointer);
      }
      else throw new IllegalStateException("Already closed.");
    }
    finally {
      lock.readLock().unlock();
    }
  }

// ... 

  public void close() {
    lock.writeLock().lock();
    try {
      if (isOpen) {
        isOpen = false;
        closeInstance(nativePointer);
      }
    }
    finally {
      lock.unlock();
    }
  }

// ... 

}
```

JohnLangford:
Not safe in general---even predict is not multithreaded, because 
individual reductions sometimes modify reduction dependent state in the 
process of doing prediction.

-John

jmorra:
In that case I would advocate for leaving this as is as it locks all multithreaded access.

JohnLangford:
Merging now on the theory that this is better than what we have.

PR #695, opened 2015-06-16
C# Wrapper

eisber:
C# Wrapper supporting
- opt-in style serialization from C# classes to VW
- Multi-threaded by VW instance pooling and model sharing

Nuget creation script
Unit tests


JohnLangford:
136 files changed?  Oh boy :-)

Can you get rid of lines like this?
 /// <summary> 

Why do we need an object factory?

Why have all the interfaces smeared across many different files? 

Why is MoreEnumerable required?  And I haven't read apache close enough, but is it compatible with BSD licensing elsewhere?  Having multiple licenses in the codebase is generally iffy.

Code like this:
 {
this.rwLockSlim.ExitWriteLock();
}

Is better written like this:
this.rwLockSlim.ExitWriteLock();

in my opinion because I like to be able to see the algorithm with fewer pixels.  When things are dispersed all over, it becomes harder to follow.

Instead of: Microsoft.Research.MachineLearning maybe we should have a more semantic namespae?

What is a "visitable namespace"?

<More in a bit>

JohnLangford:
In cs_test/Program.cs() why are the existing test cases disabled?

What is app.config?

Why add:  <Prefer32Bit>false</Prefer32Bit> 

The overhead required for the testing framework seems too high?  It's hard to imagine this approach catching up to the 90ish tests and staying caught up.  Is there a way to more directly reuse the existing tests?

What is Antlr?

<more in a bit>

eisber:
The object factory is used so we can safely swap out the model. Initially we just had a lambda that creates new VowpalWabbit instances from a shared model, but we realized that we actually need to dispose the model the lambda points too as well. So we introduced a disposable object factory.

@ ///: are you referring to code like:

    /// <summary>
    /// Thread-safe object pool supporting versioned updates.
    /// </summary>

If so this actually triggers the IntelliSense compatible documentation. So I’d rather keep the ///

@ interfaces: common C# coding guidelines. 1 file = 1 interface/class.

@ MoreEnumerable: convenience. The code is rather simple. I can replace it with my own implementation if you want. I found http://programmers.stackexchange.com/questions/40561/is-bsd-license-compatible-with-apache

@ rwLockSlim: I just checked and the { } are actually required by C#. You cannot omit them.

I think the fewer pixels go very much against common C# coding guidelines (1 file = 1 interface/class)

@ Microsoft.Research.MachineLearning: Any suggestions?

“visitable namespace”: as for the serializer I construct something similar to an abstract syntax tree (AST) of the classes to be serialized. The nodes in the syntax tress are Namespaces and Features. To support multiple front-ends for serialization (e.g. Native library interfaces vs. string vs. reflection based) I’m utilizing the visitor pattern.
As part of the pattern it is the obligation of the visitor implementor to call into the nodes for further recursion into the tree (e.g. the Visit() calls).
So if you have a look at IVowpalWabbitVisitor.cs you’ll see that I’m passing different Namespace nodes. Depending if the implementor has the obligation to call Visit() on the namespace or not I’m passing either IVisitableNamespace or INamespaceSparse.

From: John
Date: Friday, June 19, 2015 at 13:09
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] C# Wrapper (#695)


136 files changed? Oh boy :-)

Can you get rid of lines like this?
///

Why do we need an object factory?

Why have all the interfaces smeared across many different files?

Why is MoreEnumerable required? And I haven't read apache close enough, but is it compatible with BSD licensing elsewhere? Having multiple licenses in the codebase is generally iffy.

Code like this:
{
this.rwLockSlim.ExitWriteLock();
}

Is better written like this:
this.rwLockSlim.ExitWriteLock();

in my opinion because I like to be able to see the algorithm with fewer pixels. When things are dispersed all over, it becomes harder to follow.

Instead of: Microsoft.Research.MachineLearning maybe we should have a more semantic namespae?

What is a "visitable namespace"?

eisber:
I didn’t realized that this actually unit tests. I re-enabled the test cases.

App.config allows you to configure the runtime and library bindings.

What file? "Why add: false”

So the lines required for new test cases are the following:

        [TestMethod]
        [Description("using normalized adaptive updates and a low --power_t")]
        [DeploymentItem(@"train-sets\0002.dat", "train-sets")]
        [DeploymentItem(@"train-sets\ref\0002c.stderr", @"train-sets\ref")]
        [DeploymentItem(@"pred-sets\ref\0002c.predict", @"pred-sets\ref")]
        public void Test7and8()
        {
            VWTestHelper.Learn<Data, DataListener>(
                "-k --power_t 0.45 -f models/0002c.model",
                @"train-sets\0002.dat",
                @"train-sets\ref\0002c.stderr");

            VWTestHelper.Predict<Data, DataListener>(
                "-k -t -i models/0002c.model",
                @"train-sets\0002.dat",
                @"pred-sets\ref\0002c.predict");
        }

But only if the input file format matches the Data type. So the problem is that I actually try to convert the input data sets to something non-general, but rather something that is a bit customized for the data file to achieve good coverage of the serializer.

Consider rcv1 data: it has f1,…,fN features and a “const” feature. I could model this key-value pairs and just stick in f1,…,fN + const, but I chose to model this as a key-value pair structure for f1,…,fN and a separate property for “const”.

If we just want to test that VW behaves the same when using a more general data structure, I guess I can simplify (or even automate) the unit tests. When it comes to testing the various paths through the serializer I have to follow the current model.

Antlr is my parser generator so I don’t have to manually parse the VW string lines (Antlr=yacc/lex). It helps me to easily constructs different data types I’m using to test the serialization.

Markus

From: John
Date: Friday, June 19, 2015 at 13:27
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] C# Wrapper (#695)


In cs_test/Program.cs() why are the existing test cases disabled?

What is app.config?

Why add: false

The overhead required for the testing framework seems too high? It's hard to imagine this approach catching up to the 90ish tests and staying caught up. Is there a way to more directly reuse the existing tests?

What is Antlr?

JohnLangford:
We should stick with C# styles for C# code, but ... argh.

W.r.t. licenses, one small helper file doesn't seem like enough motivation to complicate the license file in the root directory.  So, rewrite or avoid seems like the best approach.

W.r.t. Microsoft.Research.MachineLearning, it seems this namespace is used for multiple semantically distinct things, so perhaps it should be broken apart more deeply.

Why did library_example.cc change?

In interactions.h, why are you casting a uint32_t to a uint32_t?

There is a large amount of code of this sort now:
 stringstream msg;
msg << "error: cannot find dictionary '" << s << "' in path; try adding --dictionary_path";
cerr << msg.str() << endl;
throw runtime_error(msg.str().c_str());

Can you instead define an object "throw_runtime_error" with an overloaded << operator which does this?  Collapsing syntactic things like this into a single line substantially aids readability in the long term.

I'm not sure this:
+ if (all.interactions.size() == 0)

in parse_args.cc is correct.  You can specify --quadratic on the commandline.

<more in a bit>

eisber:
:)

I’ll re-write the MinBy

@ Namespace: s/Microsoft.Research.MachineLearning/Microsoft.Research.MachineLearning.VowpalWabbit/ ?

library_example.cc: I unified the library interface to only allow arrays and not vectors anymore. As the vector code path was buggy and to some extend duplicate we agreed that I’ll remove it (at least from what I remember).

Interactions.h: I only introduced casts to reduce the number of warnings. Which line?

@ runtime_error: So the complete block should look like:

{
throw_runtime_error err << "error: cannot find dictionary '" << s << "' in path; try adding —dictionary_path”;
throw_runtime_error.throw();
}

@ interactions: consider the following flow:

Load Model
  Populate Interactions
  Push —interactions ab on args stack (this is done to be able to seed of another vw instance from a shared model as we’d other miss out on the parameters)
ParseArgs
  Interactions are already populated, but since we previously pushed the —interaction parameter this code will try to perform the expansion again

I’m not sure which case you worry about. If we start without a blank model, the interactions won’t be populated by Load Model.

From: John
Date: Friday, June 19, 2015 at 13:57
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] C# Wrapper (#695)


We should stick with C# styles for C# code, but ... argh.

W.r.t. licenses, one small helper file doesn't seem like enough motivation to complicate the license file in the root directory. So, rewrite or avoid seems like the best approach.

W.r.t. Microsoft.Research.MachineLearning, it seems this namespace is used for multiple semantically distinct things, so perhaps it should be broken apart more deeply.

Why did library_example.cc change?

In interactions.h, why are you casting a uint32_t to a uint32_t?

There is a large amount of code of this sort now:
stringstream msg;
msg << "error: cannot find dictionary '" << s << "' in path; try adding --dictionary_path";
cerr << msg.str() << endl;
throw runtime_error(msg.str().c_str());

Can you instead define an object "throw_runtime_error" with an overloaded << operator which does this? Collapsing syntactic things like this into a single line substantially aids readability in the long term.

I'm not sure this:

  *   if (all.interactions.size() == 0)

in parse_args.cc is correct. You can specify --quadratic on the commandline.

JohnLangford:
There is something subtle with min_prediction and max_prediction.  If you specify them on the commandline, they are _fixed_.  If you don't specify them, they adjust to the range of the labels.  Hence, by shifting from the value to the argument, you make them fixed.

Within the C++ code, it would be good to reduce this to one line of code:
{
 throw ex;
}

Why is x64 changed to Win32 in vw.sln?

I'm a bit surprised that vwdll.h has so little changed.  

github is dropping anything in bra/ket combinations.
Why add Prefer32bit: false in  cs_test/cs_test.csproj?

Having no lines of code per test case seems like the right way to go in the long term.  Can you do that?

Line 158 in interactions.h.

Ideally, it will look like:

throw_runtime_error err << "error: cannot find dictionary '" << s << "' in path; try adding —dictionary_path” << endl;

That is better than what we have now.  I could live with your proposal---that would be on parity.

W.r.t. interactions, I'm worried about the case where someone specifies -q on the commandline.  Presumably, it won't work.  That's probably why  test/train-sets/ref/ezexample_predict.stderr has a diff.

JohnLangford:
For the future it is _much_ preferred to have a patch for each systematic change.  That's much easier to digest than a giant conglomeration.

eisber:
Sure. Although we saved you from the refactoring changes :)

From: John
Date: Friday, June 19, 2015 at 14:25
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] C# Wrapper (#695)


For the future it is much preferred to have a patch for each systematic change. That's much easier to digest than a giant conglomeration.

JohnLangford:
I don't understand why the object factory is buying us anything.

-John

JohnLangford:
Maybe just VowpalWabbit for the namespace?

-John

eisber:
And the assembly? Just VowpalWabbit.dll? Fine with me.

From: John
Date: Friday, June 19, 2015 at 14:50
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] C# Wrapper (#695)

Maybe just VowpalWabbit for the namespace?

-John

JohnLangford:
It's mostly a psychological issue on my end.  Seeing a giant patch is a 
"yikes, that's many lines of code to review".  If it comes in coherent 
pieces, the hill to climb is not so formidable to begin.

-John

JohnLangford:
Sounds reasonable.

-John

eisber:
@ min/max_predictions: how can I detect if on the command line —min_prediction 0 was specified or if it just to default? Should I omit pushing —min_prediction 0 (and only push != 0)

@ Runtime error: I can trigger the throw on << endl (thought about this too), but then we cannot have multi-line messages for exception and I think it’s a bit unexpected to throw on endl. I’ll implement my .throw() variant.

I actually removed Win32 – added it back right now. The C# wrapper only works with x64.

Having no lines per test cases might be interesting. Not entirely sure. Will try.

The VowpalWabbit namespace is not that optimal since it collides with the object VowpalWabbit. I also don’t want to name the namespace VW as it will collide on the native side.


From: John
Date: Friday, June 19, 2015 at 14:24
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] C# Wrapper (#695)


There is something subtle with min_prediction and max_prediction. If you specify them on the commandline, they are fixed. If you don't specify them, they adjust to the range of the labels. Hence, by shifting from the value to the argument, you make them fixed.

Within the C++ code, it would be good to reduce this to one line of code:
{
throw ex;
}

Why is x64 changed to Win32 in vw.sln?

I'm a bit surprised that vwdll.h has so little changed.

github is dropping anything in bra/ket combinations.
Why add Prefer32bit: false in cs_test/cs_test.csproj?

Having no lines of code per test case seems like the right way to go in the long term. Can you do that?

Line 158 in interactions.h.

Ideally, it will look like:

throw_runtime_error err << "error: cannot find dictionary '" << s << "' in path; try adding —dictionary_path” << endl;

That is better than what we have now. I could live with your proposal---that would be on parity.

W.r.t. interactions, I'm worried about the case where someone specifies -q on the commandline. Presumably, it won't work. That's probably why test/train-sets/ref/ezexample_predict.stderr has a diff.

eisber:
@ Interactions.h: I get a warning from:

template <class R, void (*T)(R&, float, uint32_t)>
inline void call_T( R& dat, weight* /*weight_vector*/, const size_t /*weight_mask*/, const float ft_weight, const size_t ft_idx)
{
    T(dat, ft_weight, ft_idx);
}

1>c:\work\myvw\vowpalwabbit\interactions.h(69): warning C4267: 'argument' : conversion from 'size_t' to 'uint32_t', possible loss of data


From: John
Date: Friday, June 19, 2015 at 14:49
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] C# Wrapper (#695)

I don't understand why the object factory is buying us anything.

-John

eisber:
Do you see value in the ObjectPool to start with?

eisber:
Opted for namespace VW to avoid namespace/type name collision.

eisber:
Do you mind if I get rid of the cerr << msg in each file and we just rely on the cerr that’s in the global catch in main.cc?

Markus

JohnLangford:
On 06/19/2015 03:11 PM, Markus Cozowicz wrote:
> @ min/max_predictions: how can I detect if on the command line 
> —min_prediction 0 was specified or if it just to default? Should I 
> omit pushing —min_prediction 0 (and only push != 0)

To do this properly, I think we need an extra flag which is something 
like --autominmax.

> @ Runtime error: I can trigger the throw on << endl (thought about 
> this too), but then we cannot have multi-line messages for exception 
> and I think it’s a bit unexpected to throw on endl. I’ll implement my 
> .throw() variant.

You can have a multiline message by using "\n".  I'd prefer the shorter 
variant, although I'm flexible about the name.  I'm always happier when 
lines of rote code disappear.

> I actually removed Win32 – added it back right now. The C# wrapper 
> only works with x64.
>
> Having no lines per test cases might be interesting. Not entirely 
> sure. Will try.
>
> The VowpalWabbit namespace is not that optimal since it collides with 
> the object VowpalWabbit. I also don’t want to name the namespace VW as 
> it will collide on the native side.

Vowpal_Wabbit?

-John

JohnLangford:
It should be uint32_t ft_idx.

-John

JohnLangford:
I don't follow.  cerr is a system object.

-John

eisber:
So couple of facts upfront:


-          throw is a keyword, so void throw() ain’t possible

-          also vw_exception err << “message”  is invalid syntax

so here are the options:

Option 1)
THROW("socket: " << strerror(errno));

Implemented using Macro.

Option 2)

                             vw_exception ex;
                             ex.stream() << "socket: " << strerror(errno);
                             throw ex;

Option 3)
                             vw_exception ex;
                             ex << "socket: " << strerror(errno);
                             throw ex;

The downside here is that I need to overload operator<< for every data type right of <<

I’d opt for Option 1.

Markus

eisber:
Right now almost all place that throw an exception also log to cerr (directly in the catch block).
I’d like to remove that and just log to cerr in the global catch block in main.cc

Markus

From: John
Date: Friday, June 19, 2015 at 16:32
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] C# Wrapper (#695)

I don't follow. cerr is a system object.

-John

JohnLangford:
Got it.  That seems fine.

-John

eisber:
@ Exception throwing:

If we go for the macro we can also include filename and line number in the error message.

Markus

From: John
Date: Friday, June 19, 2015 at 16:32
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] C# Wrapper (#695)

I don't follow. cerr is a system object.

-John

JohnLangford:
macros are unappealing, because it's using a bigger language.

You should be able to chain operators as per the way that new_options() 
works. 
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/bfgs.cc#L966

It does require some finesse.

-John

JohnLangford:
That does have some appeal.  But why do you need a macro to get that?

-John

eisber:
Because _FILE_ and_LINE_ can be expanded at the right place.

You’re suggestion of bgfs.cc means that I essentially replace operator<<() with operator(), right?

So we’d end up having

throw vw_exception(“message”)(5)(“ “)

Hmm maybe I can chain << after that.

From: John
Date: Friday, June 19, 2015 at 16:41
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] C# Wrapper (#695)

That does have some appeal. But why do you need a macro to get that?

-John

eisber:
So quick update:

I can do

throw vw_exception("error: you can't combine lda and active learning") << 5 << "markus";

OR

throw vw_exception() << "error you can't combine lda …";

OR
THROW(“error: “ << 123 << “…”) // including filename and line numbers

Markus

From: John
Date: Friday, June 19, 2015 at 16:41
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] C# Wrapper (#695)

That does have some appeal. But why do you need a macro to get that?

-John

JohnLangford:
Option 2 seems best syntactically.  Can you use a macro with _FILE_ and
_LINE_ to end the message?

(Also, what happens to _FILE_ and _LINE_ if you put them into an inline
function?)

-John

eisber:
It’s not getting expanded at the right place. Since macro are really string replace it’s moved to the right place. Inline is a compiler optimization.

You can do

Throw vw_exception() << _FILE_ << “:” _LINE__ << “your own message”;

But I think it’s nicer if we hide this and then actually expose this as members of vw_exception.

Markus

From: John
Date: Friday, June 19, 2015 at 17:02
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] C# Wrapper (#695)

Option 2 seems best syntactically. Can you use a macro with _FILE_ and
_LINE_ to end the message?

(Also, what happens to _FILE_ and _LINE_ if you put them into an inline
function?)

-John

JohnLangford:
That makes me lean towards the macro then.

-John

JohnLangford:
I think we are much closer now.  Hopefully you can fix the build soon?

JohnLangford:
Merged in.   Please do more incremental requests in the future.   

There are a few minor issues we can/should address at leisure.  
(a) There are a few places where THROW() is not used. 
(b) vw_exceptions looks like it should be header only.
(c) vw_exceptions.h looks like it should be a standard header rather than added to every file.

JohnLangford:
All, I just merged in a much more mature C# wrapper from Markus based on
CLI.  There's much more to the patch, including better exception handling,
C# side testing, and a library for uploading VW into a running process
without interruption.

-John

eisber:
(a)    Which one do you refer too? I just looked through “throw “. I just found things like “throw curv_ex” and as it was a very specific exception I did not want to change it

(b)    I initially had everything in the header, but any changes triggered a full rebuild.

(c)     Should I integrate vw_exceptions.h into vw.h?

Markus

JohnLangford:
I went through and changed all the THROW() instances inline with what I 
wanted.  vw_exception seems like it should be just a (separate) header 
file, and then never change.

-John

PR #697, opened 2015-06-17
Cb adf

alekh:
This version has agreement on rcv1 in cb_algs and cb_adf.


JohnLangford:
Lines 135-139 look like a memory leak to me.  Yes?

alekh:
Why is it a memory leak? wc is not a pointer, so it should be allocated and
de-allocated on the stack automatically?

Alekh

PR #701, opened 2015-06-20
experience replay

hal3:


JohnLangford:
Why copy_array_memcpy()?

hal3:
it's much faster

JohnLangford:
Why?

Another question: For prediction you call predict(), and for learning you call predict() then learn() on a different example.  This means that the prediction being returned after a learn() has a very different semantics from what happens on a predict().  Why?

hal3:
Why is it faster? Because it avoids a bunch of explicit pushes and separate memory copying. The difference is not huge, but it did make a difference for me.

Semantics: isn't this true in general? (that "the prediction returned after learn() has a different sementics from what happens on predict()") i don't see why this is much different. is there an obviously better approach? i don't think you want to be calling learn()s when the person above you has only called predict()...

JohnLangford:
W.r.t. speed, copy_array() calls push_many() which uses memcpy() underneath---there are no separate pushes.    What is the test used to see the advantage of one over the other?

W.r.t. learn(), it typically returns a prediction which is at least similar to what predict() would have returned, and several algorithms rely on this behavior.  Here, it returns a prediction for some random example.   Perhaps you should preserve the prediction after predict() and return that instead?

hal3:
speed: huh, weird. my test must've been broken in some way. i'll revert 
it back to just using copy_array.

learn: what does "returns" mean?

PR #702, opened 2015-06-23
size_t on 32bit systems cannot be so high

martinpopel:
Compilation on 32bit systems (and g++ 4.6.3) was broken:

    interactions.cc:301:116: error: narrowing conversion of ‘2432902008176640000ll’
    from ‘long long int’ to ‘size_t {aka unsigned int}’

According to C99, 7.18.4, limit of size_t is (at least) 65535.
Relying on anything bigger is implementation-specific.


JohnLangford:
Thanks, merged.

-John

PR #704, opened 2015-06-23
Rolling back my "fixes"

sharatsc:
These fixed the test-mode bug but apparently changed the prediction-model values. The latter is a bigger problem based on use case. Rolling back until I can get a correct fix and validate it


JohnLangford:
Fixed.  It would be nice to get this settled, so if you need help ask.

-John

PR #705, opened 2015-06-24
faster C(n,k) for interactions number estimator

### LINKED ISSUE #703: Faster C(n,k) in interactions.cc
I guess that the number of combinations (N choose K) at
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/interactions.cc#L440-L443
would be slightly faster if computed without explicitly doing the factorials. See
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/boosting.cc#L30
CC: @trufanov-nok

trufanov-nok:
Switched to faster implementation of C(n,k) as proposed by Martin.
C(n,k) is used in estimation of number of generated features for interacting namespaces that called once for each example in parser.cc.
Resolves #703

### DIFF
--- a/vowpalwabbit/interactions.cc
+++ b/vowpalwabbit/interactions.cc
@@ -437,19 +437,18 @@ void eval_count_of_generated_ft(vw& all, example& ec, size_t& new_features_cnt, 
 
 
                     size_t n;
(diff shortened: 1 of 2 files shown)


JohnLangford:
Merged in, thanks.

-John

PR #707, opened 2015-06-25
self-interacting features regardless of their weight

### LINKED ISSUE #698: Self-interactions are not included when the feature has value 1
When using quadratic or cubic interaction terms, self interactions are not created from features which have a value of 1. Example:

`echo '1 |f a b' | vw` shows 3 features (correct: "a", "b", and the constant).
`echo '1 |f a:2 b:2' | vw -q ff` shows 6 features, which is also correct: "a", "b", constant, "a * a", "b * b", and "a * b".
`echo '1 |f a b' | vw -q ff` shows 4 features. If I add an "--invert_hash", the model output is
```
Constant:116060:0.158008
f^a:14355:0.158008
f^a*f^b:161933:0.158008
f^b:146788:0.158008
```
which is missing the "a * a" and "b * b" terms.

This behavior also happens at test time. If I extract model coefficients from an invert hash model file and compare predictions generated from those to predictions generated directly from VW, the predictions match everywhere except where a feature has a value of exactly 1.

I found this error in VW version 7.10.2, commit hash dc5532fad7223f1ec48965aa892aec9f58da2b61, running on OS X 10.10.3. I know that the error was not present at commit hash bb688074f82142cf340d12c2d4ae8e0e290756d6, but I haven't done a bisection to locate exactly where it was introduced.

trufanov-nok:
Resolves #698 
One thing that bothers me is unexpected decrease of total number of features for test 65:
test/train-sets/ref/search_dep_parser.stderr
from 696025 to 695754


trufanov-nok:
I believe that num_features is calculated wrongly in search_dep_parser as it calculated as in old permutations mode. I've tried to fix it and the total number of features in test 65 is dropped to 593481.
Still i'm not sure if this is enough and if this can explain previous number changes.

I think Hal  should know this piece of VW?
@hal3 , Could you please take a look and confirm that this total number of features is has sense?

hal3:
+cc kaiwei and He He who know more about this code 

-- (brief: thumb-typed on a tiny keyboard)
Hal Daumé III me@hal3.name @haldaume3 http://hal3.name
"Arrest this man, he talks in maths." http://nlpers.blogspot.com

JohnLangford:
I'm going to merge this now, because I want to drop a version and it seems that getting self-interactions right is more important than accounting.  Nevertheless, it's good to get the accounting right when we have a moment.

PR #709, opened 2015-06-26
ignore files generated in tests

martinpopel:
.gitignore already contained
  test/*.predict
so we just need to use this extension consistently.
I added
  test/*.cmp


JohnLangford:
Merged, thanks.

PR #714, opened 2015-06-27
interactions settings are now saved in model files as string

### LINKED ISSUE #711: Interactions bug
There seems to be a bug in the interactions code:

echo ' ' | ../vowpalwabbit/vw -q :: -f foo; echo ' ' | ../vowpalwabbit/vw -i foo

yields:
vw: the required argument for option '--interactions' is missing

trufanov-nok:
This shall resolve #711 
I've done following:
1. VW version is increased to 7.10.3 as we are changing model file format again.
2. I've removed interactions settings copying to `all.file_options` for old model files. This was introduced in #695. It removed bcs contained already expanded namespaces and for few other reasons.
3. Since 7.10.3 arguments of `-q, --cubic and --interactions` are saved and restored via `all.file_options` **unexpanded**. This reduces readable model filesize.
4. I found a bug - VW may successfully restore pairs and triples for model files <7.10.2 (before  interactions were introduced) but these settings weren't copied to `all.interactions` and thus were ignored by VW - as `GD::foreach_feature` now works only with `all.interactions`. Now this is fixed.
5. Fixed memory leak when `all.interactions` is directly initialized from old model file and then overwritten by command line settings.
6. Now we have two cases when users command line settings may overwrite VW models settings: 
  * interaction arrays are directly initialized with expanded namespaces from  old model files
  * model files settings are added to `all.args` with `all.file_options`.

In 1st case I can just check arrays size, print warning if user specified new interaction settings in command line and reinitialize arrays.
In 2nd case I must compare `all.args` with restored `all.file_options` and if both contains one of ("-q","--quadratic", "--cubic", "--interactions") i should remove such settings from `all.file_options` . Otherwise they'll be copied to `all.args` and duplicated. Thus I've modified `add_to_args()` function - now it accepts a list of parameters to exclude from addition. The local static bool is introduced to signal about that from `parse_modules()` to `parse_feature_tweaks()`.


trufanov-nok:
Warning message text is
```
"WARNING: model file has set of {-q, --cubic, --interactions} settings stored, but they'll be OVERRIDEN by set of {-q, --cubic, --interactions} settings from command line.\n"
```

JohnLangford:
Merged in, thanks.

PR #715, opened 2015-06-27
added missing variable to example

marcovzla:
The `prev_word` variable is missing from the `SequenceLabeler` example in the Learning_to_Search notebook. This causes a `NameError` when calling `learn()`.

I replaced it with `prevPred` based on the `SequenceLabeler2` example.


JohnLangford:
Hal, can you take a look?

-John

JohnLangford:
fixed and pushed

marcovzla:
closing because this is the wrong solution

PR #717, opened 2015-06-29
temporary fix for segfault in word_alignment

hal3:


marcovzla:
This segfault should have been fixed in 8d56ede968a2bb73cce181f2655d74b71a850b57

JohnLangford:
There is an unreconciled patch in the diff.

hal3:
sorry fixed!

JohnLangford:
What is search_generate? 

Why do all of the tests use --search_linear_ordering?  Does that work better?  If so maybe that should be the default?  If not, maybe the tests should shift?

Where are the magic numbers in passthrough coming from?  Did passthrough prove useful?  Maybe the magic numbers should be encapsulated in a function call so they aren't all over the place?

hal3:
search_generate is playing around with text generation. i can remove it but it's a reasonable example of using the library to roll your own search problem. let me know what you prefer.

the tests use --search_linear_ordering so that their output remained the same. in general our expectation is that it's better to leave this off.

magic numbers: they're just random large numbers. so far i haven't found passthrough to be particularly useful, but i wanted to try it on dep parsing because the other problems are very easy. i can move this all into a function call.

JohnLangford:
A fair number of people use the test examples as documentation, so it seems better to update the test examples rather than use --search_linear_ordering.  

W.r.t. magic numbers, it seems better to use the standard feature generation technique.  Look here:
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/interactions.h#L167 and on line 177.  Essentially, it is FNV_PRIME * first ^ second.  This one has been tested pretty thoroughly with evidence that it is better than a random number.

It seems even better to just use the existing feature manipulation.  Can you just add a pair for -q instead?

hal3:
this is all fixed now. i did some __FILE__ magic in passthrough. let me know if you prefer another solution. the issue is that different reductions add their own passthrough info, but the feature ids need to not collide across reductions --> i needed some way to identify the reduction the passthrough was coming from.

hal3:
not sure why appveyor failed -- it seems to be timing out.

JohnLangford:
@eisber Do you know why AppVeyor is dieing here?

eisber:
I had a similar issue. Just re-running the build fixed it.
We should actually looking into switching to VSO…

From: John
Date: Friday, October 2, 2015 at 14:54
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] temporary fix for segfault in word_alignment (#717)


@eisber<https://na01.safelinks.protection.outlook.com/?url=https%3a%2f%2fgithub.com%2feisber&data=01%7c01%7cmarcozo%40064d.mgd.microsoft.com%7ce2c2309df2234d69e3b308d2cb5ae346%7c72f988bf86f141af91ab2d7cd011db47%7c1&sdata=43DBq0Fr0EUMofeZ52FFfjpRc7TFM6cPhLZJNUOAnHc%3d> Do you know why AppVeyor is dieing here?

JohnLangford:
Appveyor does seem pretty flaky.

Hal, can you bump it to cause a rebuild?

-John

eisber:
I don’t have permissions so I don’t see the button, but for my pull request in appveyor it had a re-run button.

Markus

From: John
Date: Friday, October 2, 2015 at 15:06
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] temporary fix for segfault in word_alignment (#717)

Appveyor does seem pretty flaky.

Hal, can you bump it to cause a rebuild?

-John

hal3:
done

hal3:
urgh no it won't work because there were edits in the interim.

hal3:
ok tried again

JohnLangford:
Should we remove the --search_linear_ordering from the test cases (and update the outputs)?

Basically, do we want to see that forever after when people try things?

JohnLangford:
Wow, the FILE hack is pretty impressive :-)

But why use (i+magic)*quadratic_constant?  This seems a worse index generation than the old quadratic features?

Why not (FNV_PRIME * magic) ^ i?

hal3:
I'm not sure I should consider "pretty impressive" as positive or negative sentiment :). Really what's needed is for reductions to be able to identify themselves. I think this comes up in other places too -- like search should be able to check whether you have an LDF-based or non-LDF-based reduction underneath it. Anyway, end of tangent, this hack works for now.

I've changed it to FNV_prime and the xor business. I have to move the FNV_prime definition from interations.h to constant.h but that seems pretty reasonable.

Will re-push in aminute

JohnLangford:
Cool.  Now should the --search_linear_ordering be left in?

hal3:
ugh i thought I had changed that. I think git changed it back. give me another quick second :)

hal3:
looks like it's good now

JohnLangford:
Merged in, thanks.

-John

PR #718, opened 2015-06-30
Issue693

KaiWeiChang:
@hal3 
- Remember arguments in dep parsing and relation extraction tasks.
- rename srn -> sch in dep parsing task
- temporarily remove --search_metatask selective_branching --search_max_branch 5 in dep_parser demo (it causes the parser hangs during training, and I haven't found the reason).


KaiWeiChang:
hm...fixing it..

KaiWeiChang:
I cannot reproduce this compiling error on my computer... will try to find another computer to debug tomorrow...

PR #719, opened 2015-06-30
Updating Java release plan

jmorra:
It turns out that doing a maven release of this is very complicated.  I cannot use the maven release plugin and instead must write the script manually.  It is checked in as java/src/main/bin/release.sh.  Note that this script works, but ONLY when the Makefile's are checked in.  The reason for this is because the Java module does not build from a release tag.  As a result when I just released version 8.0.0 to maven central I had to use the commit tag of 191e64dd76bac23ddc1e5b2e1b9c7fe1d3eab392.  Hopefully we can fix this so that the Java module does build from a release tag then the release.sh script can reference that tag instead of a commit slightly after that tag.


JohnLangford:
Let's try to work out a better approach for the future, but thanks much for handling this.

PR #720, opened 2015-07-01
Documentation & Nuget ready

eisber:


JohnLangford:
Is there a semantics to having '///' '////' or '/////'?  If not, can you reduce it to just use '//'?

JohnLangford:
Why

 +#define _CRT_SECURE_NO_WARNINGS 
?

eisber:
Getting rid of annoying warnings.

Markus




On Thu, Jul 2, 2015 at 9:10 AM -0700, "John" <notifications@github.com<mailto:notifications@github.com>> wrote:


Why

+#define _CRT_SECURE_NO_WARNINGS
?

-

eisber:
Yes there is a semantic too out: /// allows for intellisense compatible documentation to be picked up. I'm getting the corresponding xml output and package it together with nuget

Markus




On Thu, Jul 2, 2015 at 9:08 AM -0700, "John" <notifications@github.com<mailto:notifications@github.com>> wrote:


Is there a semantics to having '///' '////' or '/////'? If not, can you reduce it to just use '//'?

-

JohnLangford:
Why is 4 //// needed?  And 5?

-John

JohnLangford:
But why not address the warnings in the code?

-John

JohnLangford:
Ping.  (W.r.t. // vs. /// I want to know if obvious code cleanups will break something, so I must know the semantics.  W.r.t. warnings, I've almost always preferred addressing them rather than hiding them.)

eisber:
@ documentation: https://msdn.microsoft.com/en-us/library/ms177246.aspx

Markus




On Sun, Jul 5, 2015 at 3:14 AM -0700, "John" <notifications@github.com<mailto:notifications@github.com>> wrote:


Ping. (W.r.t. // vs. /// I want to know if obvious code cleanups will break something, so I must know the semantics. W.r.t. warnings, I've almost always preferred addressing them rather than hiding them.)

-

JohnLangford:
The URL has no semantics for more than 3 '///'s.  Can you tweak to minimize these?

I'm not following the change here:  https://github.com/eisber/vowpal_wabbit/commit/ba485f2bc0b4981bc4c441a2cdf42aa9c8a6c422  
Can you describe what you are trying to do?

Why do we need an additional implementation of the hash function? https://github.com/eisber/vowpal_wabbit/commit/a0b0f8ffa2eea341c47099b3dcc13f61f0c62976  Seems inelegant to have two?

eisber:
I don’t get the 3 ‘///’ issue. What part of the source do you refer too?

So for “fixed --interact in combination with –cbadf" the issue was that the namespace order changed and this broke calling code.
The new version just re-inserts the removed namespace at the same location.

@ hash: Louie and myself measured performance and we got a big improvement by not performing string marshalling, but rather staying in c# land.

Markus

JohnLangford:
Got it w.r.t. interact.

Consider this line for example:
https://github.com/JohnLangford/vowpal_wabbit/pull/720/files#diff-20fa8b6910ad642ead90f6800cf77660R18
which has '/////'.  As far as I can tell, there is no reason to have more
than '///'?

The extra hash implementation is in C++ code though, right?  Why do you
need two C++ implementations of the same thing?

-John

eisber:
Oh I see. That was unfinished code. It’s gone in the next drop.

The second hash is in C++/CLI. So it’s actually compiled to managed code. (e.g. this function: uint32_t hashall(String^ s, unsigned long u))

JohnLangford:
Ok, can you remove it from the pull request?

-John

eisber:
Sure.

JohnLangford:
(This: 

https://github.com/JohnLangford/vowpal_wabbit/pull/720/files#diff-ebd13e93aa34dd3c861ff50a3d82fa2aR256
) is the extra hash implementation.

eisber:
Yes. This is the C++/CLI version. Lync?

JohnLangford:
Merged at last!

PR #721, opened 2015-07-02
active cover code

tzukuoh:
Added active_cover.cc and active_cover.h
Modified active.cc and active.h to include Oracular CAL
Added several test cases


JohnLangford:
Can you fix the windows build?
https://ci.appveyor.com/project/JohnLangford/vowpal-wabbit/build/1.0.351

tzukuoh:
Windows build fixed

JohnLangford:
Hmm, seems to still be barfing?

tzukuoh:
Ok, it works now.

tzukuoh:
ok it passed both tests.

JohnLangford:
I'm finally getting to this :-)

Can you shorten up the new test cases?  They are a bit long to read.  The goal for the test cases is "long enough to exercise the code".  Much longer versions which exhibit statistical performance can go into the demo directory.  

It looks like there are many spurious editing changes?   This makes reading the diff a bit harder.

Can you push all the stuff except for the function call in active_header.h into the .cc file?  Why is there a virtual destructor?

Why do you need n_in_dis in the shared data structure?

I haven't checked the particular bits of the algorithm itself yet, but maybe after your next pass.

tzukuoh:
Updated active_cover.cc and active_cover.h

JohnLangford:
Can you resolve the conflicts?  It's not merging automatically.

tzukuoh:
Conflicts resolved

python/mini_vw.py
```diff
@@ -18,8 +18,11 @@ def mini_vw(inputFile, numPasses, otherArgs):
                 ex = vw.example(l)
                 vw.learn(ex)
                 ex.finish()
+		print ex.get_loss()
```
JohnLangford: Printing something for every example seems undesirable.

test/RunTests
```diff
@@ -1298,3 +1298,16 @@ printf '3 |f a b c |e x y z\n2 |f a y c |e x\n' | \
 {VW} train-sets/0001.dat -i models/0097.model -p 0098.predict
     test-sets/ref/0098.stderr
     pred-sets/ref/0098.predict
+
+# Test 99: active cover
```
JohnLangford: Can you bump this to tests 101-103?

vowpalwabbit/Makefile.am
```diff
@@ -4,7 +4,7 @@ liballreduce_la_SOURCES = allreduce_sockets.cc allreduce_threads.cc vw_exception
...
+libvw_la_SOURCES = hash.cc global_data.cc io_buf.cc parse_regressor.cc parse_primitives.cc unique_sort.cc cache.cc rand48.cc simple_label.cc multiclass.cc oaa.cc multilabel_oaa.cc boosting.cc ect.cc autolink.cc binary.cc lrq.cc cost_sensitive.cc multilabel.cc label_dictionary.cc csoaa.cc cb.cc cb_adf.cc cb_algs.cc search.cc search_meta.cc search_sequencetask.cc search_dep_parser.cc search_hooktask.cc search_multiclasstask.cc search_entityrelationtask.cc search_graph.cc parse_example.cc scorer.cc network.cc parse_args.cc accumulate.cc gd.cc learner.cc lda_core.cc gd_mf.cc mf.cc bfgs.cc noop.cc print.cc example.cc parser.cc loss_functions.cc sender.cc nn.cc bs.cc cbify.cc topk.cc stagewise_poly.cc log_multi.cc active.cc active_cover.cc kernel_svm.cc best_constant.cc ftrl.cc svrg.cc lrqfa.cc interact.cc comp_io.cc interactions.cc vw_exception.cc allreduce_sockets.cc allreduce_threads.cc
```
JohnLangford: @eisber Why is there still an allreduce_sockets.cc here?

JohnLangford:
Are the defaults set to what tests reveal are the best defaults?

Also, active_cover_weights seems like quite a bit of code to avoid declaring 3 arrays rather than 1, no?

vowpalwabbit/Makefile.am
eisber: That’s part of the Model hardening & example… pull request…

tzukuoh:
Yes, default values are set to the best according to experiments.
Replaced active_cover_weights with 2 arrays.

JohnLangford:
Can you also set defaults for IWAL?

-John

tzukuoh:
For OAC/IWAL, the best fixed hyper-parameter setting constrained to the default learning rate is 35%/50% worse than the one using the optimized learning rate in terms of the relative auc gain.

JohnLangford:
This is merged in.

All, we now have a new kind of active learning algorithm in the VW 
codebase.  See here:
http://arxiv.org/abs/1506.08669
for details.

-John

PR #725, opened 2015-07-09
Update pylibvw.cc

gramhagen:
currently calling ex.learn() using the python interface updates the model even after calling ex.set_test_only(True)

now the bit is checked when calling learn and the learner's predict method is called instead of learn if the example is a test only example

### DIFF
--- a/python/pylibvw.cc
+++ b/python/pylibvw.cc
@@ -105,7 +105,11 @@ void my_finish_example(vw_ptr all, example_ptr ec) {
 }
 
 void my_learn(vw_ptr all, example_ptr ec) {
-  all->learn(ec.get());
+  if (ec->test_only) {
+    all->l->predict(*ec);
+  } else {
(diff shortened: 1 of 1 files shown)


gramhagen:
seems to be an unrelated problem with appveyor causing the build to fail:

```bash
Build started
Fetching repository commit (340ad63)...OK
Total: 15.2 MB in 660 files
cd c:\vowpal_wabbit
vowpalwabbit\.nuget\nuget install -o vowpalwabbit\packages cs_unittest\packages.config
Installing 'Antlr4.Runtime 4.3.0'.
Successfully installed 'Antlr4.Runtime 4.3.0'.
WARNING: The operation has timed out
Unable to find version '4.3.0' of package 'Antlr4'.
Command exited with code 1
```

JohnLangford:
Thanks.  Not sure what the freakish build failure is, but appveyor has not always been stable.

Issue #726, opened 2015-07-12
A few small suggestions


trufanov-nok:
I. I've played with --readable_model and was quite surprised to find a lot of `0.000000` feature values in its output. VW doesn't save features with 0 weight to regressor. In a moment I realized that they're not zeroes but just very small. `gd.cc::save_load_regressor()` uses simple `:%f \n` for sprintf to save weights and `%f` has precession 6 by default. So you can't see values smaller than 1e-6 - you'll get only leading zeroes in readable model. My suggestion is to use `%g` instead:
```
f	Decimal floating point, lowercase	392.65
F	Decimal floating point, uppercase	392.65
e	Scientific notation (mantissa/exponent), lowercase	3.9265e+2
E	Scientific notation (mantissa/exponent), uppercase	3.9265E+2
g	Use the shortest representation: %e or %f	392.65
G	Use the shortest representation: %E or %F	392.65
```
This results in following output:
```
...
8246:-2.64074e-10
8264:0.605609
8265:-0.495252
8266:-0.596221
8288:-1.95944e-12
8289:-1.63608e-12
...
```

There is a drawback - if someone parses readable model files automatically he might not be ready to parse scientific format of real values. Also `%g` shall be slower than `%f`.

II. I found error messages `malformed example !\n'|' , ':' , space or EOL expected after : xxx` quite useless bcs they display not the whole example but its piece up to current `reading_head`. And this piece is perfectly fine bcs the problem happened just after reading_head. For example for `-1 |f val1:1.0val2:2.0` (space between features is missing) you'll get  `malformed example !\n'|' , ':' , space or EOL expected after : -1 |f val1:1.0`. Thus you have to open dataset, find this example and check the rest of example line to find the problem. I changed this to simply print whole example line in error message in my fork. There may be other solutions too.

trufanov-nok:
III
 I'm writing a reduction and found out that `parser.cc::set_done(all)` is seems to be standard way to manually stop learning and exit VW.  Surprisingly I realized that reduction's end_examples() callback function is never invoked in this case. 
There is a bug at [this line](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/learner.cc#L55)
```
  if (all.early_terminate) //drain any extra examples from parser and call end_examples
      while ( all.early_terminate == false )
{
}
```
while loop body won't be ever called.

trufanov-nok:
[...]

PR #727, opened 2015-07-13
fix broken while in generic_driver

trufanov-nok:
refers to suggestion 3 in  https://github.com/JohnLangford/vowpal_wabbit/issues/726


JohnLangford:
Thanks, merged in.    That code fragment was a clear bug, so I'm unclear on how it even happened in the first place.

PR #731, opened 2015-07-20
Adding the ability to load JNI externally

jmorra:
The point of these changes are twofold
1.  Add the ability to load the JNI library externally.  This will allow better debugging and easier hot fixes of the C code is required.
2.  Expose more methods in NativeUtils to help debugging.

@deaktator, I would love any feedback on this.


java/src/main/java/vw/VW.java
```diff
@@ -54,7 +47,44 @@ public static void main(String[] args) {
     public VW(String command) {
         isOpen = true;
         lock = new ReentrantLock();
-        nativePointer = initialize(command);
+        // This allows the user to instead of using the prepackaged JNI layer load their own external JNI layer.
```
deaktator: This comment should be a javadoc comment above the method so that people can refer to the API documentation rather than the code for an understanding of how to load their own natively compiled JNI extension.  I would also put this is a README.md so that people know it's possible.

java/src/main/java/vw/VW.java
```diff
@@ -54,7 +47,44 @@ public static void main(String[] args) {
...
+        // This means that if a user wants to use this code with an OS that is not supported they would follow the following steps
+        // 1.  Download VW
+        // 2.  Build VW for the OS they wish to support
+        // 3.  Call either System.load(xxx) or System.loadLibrary(xxx)
+        // After doing this then the first UnsatisfiedLinkError will never occur.  If a user wishes to use the prepackaed
+        // JNI layer then the first time this object is constructed the UnsatisfiedLinkError will throw and the native libraries
+        // will be loaded and every other time it will not.
+        long currentNativePointer;
+        try {
+            currentNativePointer = initialize(command);
```
deaktator: Based on the above comment, I imagine it's possible to load the native library via: 

```java
System.loadLibrary(...)
```

but the `loadedNativeLibrary` function will still be false.  Is the meaning of the `loadedNativeLibrary` variable really *my load native library procedure was successfully invoked*? Or is should it really mean *someone successfully loaded a valid native VW library*?  If it's the second, I would lock with `STATIC_LOCK` then set `loadedNativeLibrary` to true.  Otherwise, maybe it would be beneficial to have a second variable stating a user-defined native VW library was loaded.

deaktator:
I think this looks good with the exception of the `loadedNativeLibrary` variable that I described above.  Maybe it would be beneficial to have a static function in the VW class that describes how the library was loaded.

```java
class VW { 
// ...
  public static enum LibType {
    NONE, // Not loaded yet.
    INTERNAL,  // Loaded using one of the library-defined, precompiled native libraries
    USER_DEFINED // Loaded using a user-defined native library
  }

  public static LibType libType() {
    return ...
  }
// ...
}
```

java/src/main/java/vw/VW.java
jmorra: That is true, but if System.loadLibrary was called then the loadNativeLibrary method will never be called.  In order to make this more clear, I will set loadedNativeLibrary to true in the try block after a successful load.  Also, loadedNativeLibrary is set to true inside of the static lock so it will work correctly.

deaktator:
Nice to have the initialization string getter!

JohnLangford:
Merged in, thanks.

PR #736, opened 2015-07-30
loss func shouldn't be called for never observed label

### LINKED ISSUE #735: Spurious error output with hinge loss
I've found that if I use VW with hinge loss, the screen output contains an incorrect error message, "You are using label 0 not -1 or 1 as loss function expects!" when the input to VW has only a single class. Example:

The command `echo "1 | a b" | vw --loss_function hinge` produces output
```
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = 
num sources = 1
1.000000 1.000000            1            1.0   1.0000   0.0000        3

finished run
number of examples per pass = 1
passes used = 1
weighted label sum = 1.000000
average loss = 1.000000You are using label 0 not -1 or 1 as loss function expects!

best constant = 1.000000
best constant's loss = 0.000000
total feature number = 3
```

Similarly, the command `echo -e "1 | a b\n1 | a" | vw --loss_function hinge` produces output:
```
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = 
num sources = 1
1.000000 1.000000            1            1.0   1.0000   0.0000        3
0.711395 0.422791            2            2.0   1.0000   0.5772        2

finished run
number of examples per pass = 2
passes used = 1
weighted label sum = 2.000000
average loss = 0.711395You are using label 0 not -1 or 1 as loss function expects!

best constant = 1.000000
best constant's loss = 0.000000
total feature number = 5
```

But if the input contains both "-1" and "1" examples, the error message does not occur:
`echo -e "1 | a b\n-1 | a" | vw --loss_function hinge` -->
```
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = 
num sources = 1
1.000000 1.000000            1            1.0   1.0000   0.0000        3
1.288605 1.577209            2            2.0  -1.0000   0.5772        2

finished run
number of examples per pass = 2
passes used = 1
weighted label sum = 0.000000
average loss = 1.288605
best constant = -1.000000
best constant's loss = 1.000000
total feature number = 5
```

I'm running the current master branch of VW (commit 62fbc5553a1cdf2f3f800cae7fdead3b00e7fd30) on OS X 10.10.4 .

Why I care: VW uses stdout for error messages, so I'm reading stdout to check for errors. Having an error message when there's not an error is confusing. I format examples with a dummy label (all "1"s) when I'm predicting, so this extra error message appears there.

trufanov-nok:
resolves #735

### DIFF
--- a/vowpalwabbit/best_constant.cc
+++ b/vowpalwabbit/best_constant.cc
@@ -71,9 +71,11 @@ bool get_best_constant(vw& all, float& best_constant, float& best_constant_loss)
         return false;
 
     if (!is_more_than_two_labels_observed)
-    best_constant_loss = ( all.loss->getLoss(all.sd, best_constant, label1) * label1_cnt +
-                           all.loss->getLoss(all.sd, best_constant, label2) * label2_cnt )
-            / (label1_cnt + label2_cnt);
+    {
+      best_constant_loss =  (label1_cnt>0)?all.loss->getLoss(all.sd, best_constant, label1) * label1_cnt:0.0;
+      best_constant_loss += (label2_cnt>0)?all.loss->getLoss(all.sd, best_constant, label2) * label2_cnt:0.0;
+      best_constant_loss /= label1_cnt + label2_cnt;
+    }
     else best_constant_loss = FLT_MIN;
 
     return true;


JohnLangford:
Merged in, thanks.

stephen-hoover:
Thanks, @trufanov-nok !

PR #739, opened 2015-08-02
synthetic feature weight calculation in macros

trufanov-nok:
I would like to propose this little change.
It's just a macros `INTER_WEIGHT_OP(weight1, weight2) weight1*weight2` that replaces `weight1*weight2` expressions everywhere where new feature weight is calculated for features generated by interactions.
As macros are preprocessed this shouldn't affect neither performance nor results. Regression tests shows that everything is ok.

I found out that for some tasks the ways of new feature weights calculation other than simple multiplication   may let you achieve significantly better results. Particularly `fabs(weight1-weight2)` or even `weight1>weight2?1.0:-1.0` for features generated by interacting pairs of different fund market indicators. This may be explained by that: for human traders intersection or convergence of price moving averages means more than their absolute values.

Anyway, I'm not suggesting new feature weight operations - just a macros to let developer play with them more easily.

### DIFF
--- a/vowpalwabbit/interactions.h
+++ b/vowpalwabbit/interactions.h
@@ -99,6 +99,9 @@ struct feature_gen_data
 //    feature_gen_data(): loop_idx(0), x(1.), loop_end(0), self_interaction(false) {}
 };
 
+// macros below may be adjusted to change the way synthetic feature weight is calculated
+// beware - its result must be non-zero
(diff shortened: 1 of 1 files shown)


JohnLangford:
Merged in.  It seems fine, although if you truly find this useful, making it a user-visible option becomes desirable.

JohnLangford:
Also, I tend to prefer inline functions to macros so I switched it to an inline function.  Why?  Because inline functions don't require a new language.

PR #740, opened 2015-08-05
add more info printed when parsing malformed example

followyourheart:
Print more info when the example format does not match the required format. So one can debug more efficiently.


JohnLangford:
Thanks both. 

At this particular point in the code an error must be happening because there are more than 4 tokens in the simple label, so the token number of the error is kind of vague.  Something like: 

Error: too many token in simple label: <token1> <token2> ... 

Seems the most explanatory possible given the location.  The exact line number is difficult to recover (and may be meaningless if the data is coming of a network).

JohnLangford:
I've tweaked the code along the lines suggested.

PR #742, opened 2015-08-06
Assume adaptive and normalized are off when g is NULL

pmylund:
I noticed a segfault when using --ftrl/--pistol and --save_resume stemming from g being NULL in save_load_online_state.

I added a null check to enter the (!adaptive && !normalized) block when g is NULL, and `vw --ftrl -i 1000lines.model -f 1000lines.model --save_resume 1000lines` now doesn't segfault, and produces a lower average loss each time it is run, indicating that things are working.

I'm new to this codebase, so it's not clear to me whether this is the right fix. Would appreciate your feedback.

### DIFF
--- a/vowpalwabbit/gd.cc
+++ b/vowpalwabbit/gd.cc
@@ -756,7 +756,7 @@ void save_load_online_state(vw& all, io_buf& model_file, bool read, bool text, g
 	    {
 	      assert (i< length);		
 	      v = &(all.reg.weight_vector[stride*i]);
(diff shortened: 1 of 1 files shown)


JohnLangford:
Looks correct, merging in.  The use of save_load_online_state is a bit of a
hack in this case.

-John

JohnLangford:
(Thanks)

-John

pmylund:
Happy to help!

PR #743, opened 2015-08-09
fix vw-hypersearch -e 'command argument'

denik:
Previously it did not work unless 'command' is found the current directory:

  master ~/work/vowpal_wabbit $ utl/vw-hypersearch -e 'cat utl/vw-hypersearch' 0 1 echo %
  -e cat utl/vw-hypersearch: No such file or directory
  Usage: vw-hypersearch [options] <lower_bound> <upper_bound> [tolerance] vw-command...
  ...

With this patch it works fine:

  master ~/work/vowpal_wabbit $ utl/vw-hypersearch -e 'cat utl/vw-hypersearch' 0 1 echo %
  Using external evaluation plug-in: 'cat utl/vw-hypersearch'
  trying 0.618033988749895 . 1 (best)
  ...

### DIFF
--- a/utl/vw-hypersearch
+++ b/utl/vw-hypersearch
@@ -581,7 +581,8 @@ sub process_args {
         # Allow users to pass args to the eval script,
         # assume first non-space sequence is the command
         my ($first_word) = ($opt_e =~ /^\s*(\S+)/);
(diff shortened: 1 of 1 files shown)


arielf:
Looks good, thanks for catching this!

PR #753, opened 2015-08-19
Threaded AllReduce, C# Refactoring, Minor CB ADF Fix

eisber:


JohnLangford:
Some partial comments:

There are many pieces of code like this:
+            // support anchor feature
+            if(namespaceDense.DenseFeature.AddAnchor)
+            {
+                this.namespaceBuilder.AddFeature(this.namespaceHash, 1);
+                i++;
+            }
+
Why isn't this one piece of code?


Visit here does not seem to be a good descriptor of what is happening.
cs/Serializer/Visitors/VowpalWabbitStringVisitor.cs 
Instead of Visit, maybe Output?

eisber:
2 thinks: first I’m implementing a visitor pattern for the serializer. The graph to be visited is the description of properties and associated behavior (similar to an AST) and the different visitor implementations (native and string) visit that graph and produce the corresponding output.
You’d rather see a separate method with a ref parameter? Fine with me.

JohnLangford:
It's ok to use the word 'visitor' in the description of the pattern, but 
individual visitors should probably have more meaningful names.

Yeah, having a separate method to compress the code seems pretty 
desirable.  Deduplication of code seems a near-universal virtue.

-John

eisber:
I had another look. The code only occurs in 2 places VowpalWabbitNativeVisitor.cs and VowpalWabbitNativeVisitorExt.tt. The many occurrences you see in VowpalWabbitNativeVisitorExt.cs are actually generated code. I’d rather have it expanded out as Roger observed some performance impact already.

eisber:
@ Visit Methods: how about MarshalFeature instead of Visit or Output (see https://msdn.microsoft.com/en-us/library/bb384865.aspx) - it feels more descriptive

JohnLangford:
Marshal certainly sounds more descriptive.  

But, why are we checking in generated code?  Why don't we check in the generator and then let a compiler generate code?

eisber:
I’ll rename to Marshal then.

@ generated code: It’s common VS practice to check in the generated .cs files. I believe the reason for it is, that one will have to confirm a dialog when the .tt file is executed (security). Not sure how the confirmation works with an automated build.
I can look into it, but rather stick to common practice.

Markus

JohnLangford:
Please look into it.  My experience with checking in generated code is 
that it's very annoying since small changes can generate massive diffs.

-John

JohnLangford:
I'd like to get this one dealt with today. 

Can you change StringVisitor to StringMarshall?  And visit to marshall?  Here: 
https://github.com/JohnLangford/vowpal_wabbit/pull/753/files#diff-20fa8b6910ad642ead90f6800cf77660R17

Can you fix the indentation issues here? https://github.com/JohnLangford/vowpal_wabbit/pull/753/files#diff-67143e7959b24496f986b523eefeb4acL43  And here: https://github.com/JohnLangford/vowpal_wabbit/pull/753/files#diff-c6d0ae977a5016c155b717d7a8fa0118R449  ?  Note that there is now a .editorconfig file which should set the indentation style to something which does this automatically with the right plugin.

Also, why did you make any changes to bfgs.cc?  It should be unnecesssary.   Nuking spurious changes is welcome.

Another issue: why did you shift the allreduce library into object files?  The library seems desirable as a separate entity because it is easy to imagine use in many other settings.

(more after I eat lunch)

JohnLangford:
There is something off with the spacing here:
https://github.com/JohnLangford/vowpal_wabbit/pull/753/files#diff-d479359b05a8ea144a9b12c13044983cR97  Please fix so that I can see what has really changed.

Why has gd.cc been altered?  Why has io_buf.cc been altered? kernel_svm.cc? main.cc? parse_args.cc?

JohnLangford:
Merged, finally.  Shorter pull requests would be good.  Also, try to avoid changing whitespace unnecessarily please.

PR #754, opened 2015-08-19
Documentation

eisber:
if merge AllReduce first, this should be a lot less.


JohnLangford:
Can you resolve the conflicts here?

eisber:
Done.

JohnLangford:
Cool.   Louie can you look at the C# code here?

-John

cs/ObjectPool.cs
```diff
@@ -157,6 +160,8 @@ public void UpdateFactory(ObjectFactory<TSource, TObject> factory)
 
         internal void ReturnObject(PooledObject<TSource, TObject> pooledObject)
         {
+            Contract.Ensures(pooledObject != null);
```
lhoang29: remind me whether this will throw a runtime exception? I think the last time I tried it didn't do anything...

eisber: depends on compile-time defines. since I don't specify anything it doesn't throw. That's the reason I'm sometimes using throw ArgumentNullException and Contract.EndBlock()

cs/Serializer/Visitors/VowpalWabbitStringVisitor.cs
```diff
@@ -8,12 +8,15 @@
...
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.Contracts;
 using System.Globalization;
 using System.Linq;
 using System.Text;
 using VW.Interfaces;
 using VW.Serializer.Interfaces;
 
+#pragma warning disable 1591
```
lhoang29: which warning is this?

cs/Serializer/VowpalWabbitSerializer.cs
```diff
@@ -137,13 +181,32 @@ private void Dispose(bool disposing)
...
             }
 
+#if DEBUG
+            CacheEntry cacheEntry;
+            if (!this.reverseLookup.TryGetValue(example, out cacheEntry))
+            {
+                throw new ArgumentException("Example is not found in pool");
+            }
+
+            if (!cacheEntry.InUse)
```
lhoang29: can cacheEntry be null here?

cs/Serializer/VowpalWabbitSerializerFactory.cs
```diff
@@ -47,25 +42,34 @@ public static class VowpalWabbitSerializerFactory
...
+
+#if DEBUG
+            var stringSerializerFunc = CreateSerializer<TExample, VowpalWabbitStringVisitor, string>();
+
+            Func<VowpalWabbit, TExample, ILabel, VowpalWabbitExample> wrappedSerializerFunc;
             if (serializerFunc == null)
             {
                 // if no features are found, no serializer is generated
-                serializerFunc = (_,__,___) => null;
+                wrappedSerializerFunc = (_,__,___) => null;
```
lhoang29: i found that the _ causes stylecop (or maybe fxcop) to crash, so it might be good to use dummy names instead.

cs/Serializer/Visitors/VowpalWabbitStringVisitor.cs
eisber: http://stackoverflow.com/questions/25846205/compiler-warning-cs1591-how-to-show-that-warning-only-for-undocumented-methods

cs/VowpalWabbitAsync.cs
```diff
@@ -9,14 +18,29 @@
...
         private VowpalWabbitThreadedLearning manager;
 
+        /// <summary>
+        /// The serializers are not thread-safe. Thus we need to allocate one for each VW instance.
+        /// </summary>
         private VowpalWabbitSerializer<TExample>[] serializers;
 
         internal VowpalWabbitAsync(VowpalWabbitThreadedLearning manager)
         {
+            Contract.Requires(manager != null);
```
lhoang29: what's the difference between Requires and Ensures?

cs/Serializer/VowpalWabbitSerializerFactory.cs
eisber: done

cs/VowpalWabbitAsync.cs
```diff
@@ -96,16 +170,54 @@ private void Dispose(bool disposing)
...
+        /// <summary>
+        /// The serializers are not thread-safe. Thus we need to allocate one for each VW instance.
+        /// </summary>
         private VowpalWabbitSerializer<TActionDependentFeature>[] actionDependentFeatureSerializers;
 
         internal VowpalWabbitAsync(VowpalWabbitThreadedLearning manager)
         {
+            if (manager == null)
+            {
+                throw new ArgumentNullException("manager");
```
lhoang29: why doesn't Contract.Requires (or other methods) work here?

cs/Serializer/VowpalWabbitSerializer.cs
eisber: CacheEntries are only added a few lines below. Never null.

cs/VowpalWabbitAsync.cs
eisber: Requires are preconditions (e.g. manager != null)
Ensures are postconditions that must be met on method exit.

cs/cs.csproj
```diff
@@ -15,16 +15,60 @@
...
     <DebugSymbols>true</DebugSymbols>
     <OutputPath>bin\x64\Debug\</OutputPath>
-    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <DefineConstants>TRACE;DEBUG</DefineConstants>
     <DebugType>full</DebugType>
     <PlatformTarget>x64</PlatformTarget>
     <ErrorReport>prompt</ErrorReport>
     <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
     <DocumentationFile>bin\x64\Debug\VowpalWabbit.XML</DocumentationFile>
+    <CodeContractsEnableRuntimeChecking>True</CodeContractsEnableRuntimeChecking>
```
lhoang29: Seems weird that the contract settings only apply to Debug mode... why is that?

cs/VowpalWabbitAsync.cs
eisber: Contract.Requires doesn't throw always (just if compiled correctly). I use the direct checks in places that aren't on the hot path

cs_unittest/TestModelLoading.cs
```diff
@@ -73,11 +73,11 @@ private void InternalTestModelRandomCorrupt(string modelFile)
                         Assert.IsTrue(false);
                     }
                 }
-                catch (VowpalWabbitException ex)
+                catch (VowpalWabbitException)
```
lhoang29: can you remove this catch? the one below is enough. I think this was accidentally checked in while I was debugging.

vw_clr/vowpalwabbit.cpp
```diff
@@ -209,6 +213,11 @@ namespace VW
 
 	void VowpalWabbit::Learn(VowpalWabbitExample^ ex)
 	{
+#if _DEBUG
+		if (ex == nullptr)
```
lhoang29: nit: wrap curly braces, this will throw stylecop errors

vw_clr/vw_base.cpp
```diff
@@ -22,14 +22,8 @@ using namespace System::Text;
 namespace VW
 {
 	VowpalWabbitBase::VowpalWabbitBase(VowpalWabbitSettings^ settings)
-		: m_examples(nullptr), m_vw(nullptr), m_model(nullptr)
+		: m_examples(nullptr), m_vw(nullptr), m_model(nullptr), m_settings(settings != nullptr ? settings : gcnew VowpalWabbitSettings)
```
lhoang29: is ?? not valid in c++/cli?

cs/cs.csproj
eisber: I do fine grained checking even on the hot path. Don’t want to have that in debug mode.

cs_unittest/TestModelLoading.cs
eisber: done

cs/ObjectPool.cs
lhoang29: why is this better than manually throwing? I understand that code doesn't have to deal with contract violations, but because of the silent failures, won't it be hard to debug?

vw_clr/vowpalwabbit.cpp
eisber: done

vw_clr/vw_base.cpp
eisber: nope. just tried.

cs/ObjectPool.cs
eisber: Since this is on the hot path I didn’t want to add the additional check here.

lhoang29:
looks good

JohnLangford:
Why is spacing changing again in .cc files?  Can you fix it so that only things that actually change change?

lhoang29:
I think editorconfig automatically normalizes spacing in each file if it doesn't have the proper indentation. We want this though so at the end every file has the right format.

JohnLangford:
Editorconfig was applied to the previous checkin, so I believe this is 
uneditorconfig.

-John

eisber:
So I think this was introduced by the merge. I’m trying to revert to jcl/master right now.

JohnLangford:
Great, merged.

PR #757, opened 2015-08-20
Model hardening & example reuse for efficient hyper-parameter search

eisber:


JohnLangford:
Now, can you tweak this to merge?  And then get @lhoang29 to check over the C# side.

eisber:
Again some whitespace hell.

JohnLangford:
Fixable?

JohnLangford:
@lhoang29 Can you look over the csharp part?

JohnLangford:
Why is allreduce_sockets.cc in libvw?  Shouldn't it be in allreduce?

What is the goal of the process_example refactoring in learner.cc?

Why is split_commandline needed?

lhoang29:
c# side looks good

eisber:
Why is allreduce_sockets.cc in libvw? Shouldn't it be in allreduce?: **fixed**

 What is the goal of the process_example refactoring in learner.cc?
--> Efficient hyper-parameter search by having multiple vw instances in memory and feeding the same example to all instances (serial, not parallel). 

Why is split_commandline needed?
--> the vw setup function expects argv, argc parameters. For the hyper-parameter search I read the cmd line from a single file and need to transfer it into argv, argc format.

vowpalwabbit/vw_validate.cc
```diff
@@ -0,0 +1,54 @@
...
+{
+    void validate_version(vw& all)
+    {
+        if (all.model_file_ver < LAST_COMPATIBLE_VERSION || all.model_file_ver > PACKAGE_VERSION)
+        {
+            THROW("Model has possibly incompatible version! " << all.model_file_ver.to_string());
+        }
+    }
+
+    void validate_unexpected_eof(size_t nbytes)
```
JohnLangford: This seems problematic in two ways.  

The first is that the name is unnecessarily obscure.   Maybe something like "positive_bytes" or "validate_positive_bytes".

The second is that this specializes the function to model io, which seems undesirable.  At points in the future, we may plausibly want these functions to do other things.

JohnLangford:
I'd like to have just one function for converting a commandline into an argv.   Why isn't this one adequate?
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parse_args.cc#L1190

vowpalwabbit/main.cc
```diff
@@ -14,12 +14,123 @@ license as described in the file LICENSE.
...
+
+       fail:
+               if (wargs) LocalFree(wargs);
+               if (cmdlinew) free(cmdlinew);
+       }
+#endif // WIN32
+
+       if (argv)
+       {
+               for (i = 0; i < *argc; i++)
```
JohnLangford: This is a more minor stylistic thing.   Blocks of length 1 are generally not necessary and getting rid of them makes the code substantially more concise.

vowpalwabbit/parse_regressor.cc
```diff
@@ -54,287 +49,307 @@ const size_t buf_size = 512;
...
         {
-	all.file_options->str(buff2);
-    }
+            all.file_options->str(buff2);
+        }
+
+        // Read/write checksum if required by version
+        if (all.model_file_ver >= VERSION_FILE_WITH_HEADER_HASH)
+        {
+            uint32_t check_sum = uniform_hash(model_file.space.begin, model_file.space.size(), 0);
```
JohnLangford: This seems like it might cause a very subtle bug.  The problem is that size() is the size of the number of bytes in the buffer rather than the number of bytes in the header.  This is generally OS and FS dependent.

vowpalwabbit/vw_validate.cc
lhoang29: I can rename this. The check is not supposed to specialize to model only, although the error message certainly needs updating.

vowpalwabbit/parse_regressor.cc
lhoang29: Didn't realize the buffer depends on OS & FS. How about I keep a local byte buffer of the header to compute the hash?

eisber:
Didn’t know about it. I quickly looked through and the main difference seems to be environment variable expansion. When reading the string from a model this might seem undesirable, not so sure on the hyper-parameter search option. It might be useful to expand variables in there…

Let me know what you prefer, I’m happy to switch to the existing implementation.

Markus

vowpalwabbit/main.cc
eisber: Agree. So depending on the other thread on environment variable I'll either drop this code completely or adapt the formatting.

JohnLangford:
We should definitely only have one implementation.  Can you turn off 
environment variable expansion easily?  If so, using the new one seems 
fine.  If not, let's just use the old one.  Also, it should probably be 
in parse_args for modularity reasons.

-John

vowpalwabbit/parse_regressor.cc
JohnLangford: The problem starts here:

https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/io_buf.cc#L123

read() is not guaranteed to return the requested number of bytes. From 
the man page:

"It is not  an
        error  if  this  number  is smaller than the number of bytes 
requested;
        this may happen for example because fewer bytes are actually  
available
        right  now  (maybe  because we were close to end-of-file, or 
because we
        are reading from a pipe, or from a terminal),  or  because 
read()  was
        interrupted  by  a  signal."

Hence read_file:
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/io_buf.h#L137 


may not fill the buffer.  Hence, if the number of calls to fill() differ 
from one run to another, you'll get different results.  In my 
experience, this only matters when reading from a pipe or socket, but 
given the unclear semantics...it feels iffy to trust that all OS/FS 
combinations read a complete blob when it happens to be available.

As far as what to do instead, I'd suggest adding a 'hash' flag to the 
io_buf datastructure, then lazily computing a chained hash when that 
flag is on.  This should allow interruptions at any point while 
maintaining consistency over a chain.

-John

eisber:
The expansion wasn't switchable by a flag or so. So I'm using the existing function.

JohnLangford:
What is the status on this?

eisber:
I’m about to review Louies changes.

cs_unittest/TestModelLoading.cs
```diff
@@ -10,41 +10,55 @@
...
                 using (var vw = new VowpalWabbitModel(string.Format("--quiet -t -i {0}", modelFile)))
                 {
                     // should only reach this point if model is valid
-                    Assert.IsTrue(shouldPass);
+                    passed = true;
                 }
             }
             catch (VowpalWabbitException ex)
             {
                 Assert.IsTrue(ex.Message.Contains("corrupted"));
```
eisber: use https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.testtools.unittesting.expectedexceptionattribute.aspx

lhoang29: Looks neat, although I wanted to inspect the exception message before asserting. I guess the ideal thing would be to throw specific exception type and use the annotation here, but I don't think that is possible from C++ side, so we'd need to catch in CLI, inspect message, and rethrow, which seems messy...

eisber: You can inspect the message in the 2nd argument.

lhoang29: Hmm, don't think that is intended for checking content of the exception message. Also, for the noExceptionMessage doc (which is the 2nd param), it says:
This API supports the product infrastructure and is not intended to be used directly from your code. (Inherited from ExpectedExceptionBaseAttribute.)

eisber: Yep. It’s just the message. Got confused myself.

JohnLangford:
Status now?

lhoang29:
My changes are done, not sure about other parts of it

eisber:
Upstream merge done.

JohnLangford:
I'm reviewing again.  What does the refactoring of learner.cc buy us?

eisber:
savings on example parsing. If one supplies an arguments file, the code parses the example once and then feeds it into multiple VW instances.

JohnLangford:
Perhaps there is a better name than this: bin_text_read_write_validate_eof ? Maybe "validated"?  You aren't validating an EOF---you are validating that some bytes are read or written.

JohnLangford:
Or maybe no name.  Is there ever a reason to _not_ use these functions?

JohnLangford:
I don't have any more comments.  The checksum handling seems cool.

lhoang29:
Renamed. I tried no name before and it broke some use cases, there are some instances of this (in gd.cc if I remember correctly) that allow no bytes read.

JohnLangford:
Can you name one?

-John

lhoang29:
looks like there's one case here: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/gd.cc#L610

JohnLangford:
I don't follow---the size of i is 4 bytes.

-John

lhoang29:
I was trying to point out that this code is ok with buf_read returning 0 bytes after reading, even though 4 bytes are requested. The reason is it is ok to have no more data in the file at this point. Thus there are valid scenarios for both the _validate and not overloads.

JohnLangford:
Got it.

-John

JohnLangford:
This is merged in now, thanks to Louie and Markus.

(a) Models are hardened with a chained hash over the header so that 
corrupt models do not cause segfaults.
(b) Markus tweaked things so that you run several VW _models_ with 
varying parameters (and the same parser).  In situations where parsing 
time is a limiting factor in hyperparameter search, this can yield a 
substantial speedup.

-John

PR #764, opened 2015-08-25
python pip installable vowpal wabbit package

trbs:
Initial setup.py that creates a pip installable package for Vowpal Wabbit.

Currently the Python module is statically linked with Vowpal Wabbit.
It would be very nice if we would be able to build the Python module and have it link to the installed libvw.so as well in the future. (result being a ```pyvw``` package that works like the ```pyzmq``` package, using the system libvw.so if available otherwise building it from source)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,3 +93,9 @@ VW.creator.user
 VW.files
 VW.includes
 
+# Python package
+build/**
+dist/**
+pyvw.egg-info/**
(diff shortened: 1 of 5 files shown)


JohnLangford:
This looks good to me.  Hal, any comments?

-John

JohnLangford:
Looks good to me too!

JohnLangford:
Great, merged.

-John

PR #765, opened 2015-08-25
Cleaner code

martinpopel:
"Just" renaming variables, so the code is easier to read and understand.


arielf:
i hope this is accepted.  Many parts of the code have become too complex to follow so anything that make them easier to understand is welcome.

JohnLangford:
This all seems great.  Getting the right names for things is a huge help.

-John

Issue #761, opened 2015-08-22
R/dt2vw.R error when used with many columns


martinbel:
The dt2vw function returns this error: "sprintf: only 100 arguments allowed" 
when called with a data.table of +100 columns. 

Is there any other way to sort this out? I've checked the source code, but couldn't find a solution. This would be quite useful to have working as can improve significantly the VW formating. 

Also, am I using the weight argument correctly?

```
# Here is a gist with the same function. Just added a print of the format string.

library(data.table)
source('https://gist.githubusercontent.com/martinbel/a013c2bd2333b85eeb76/raw/7b68ede0385f3d8684a07854e8d6ca527b9cd441/dt2vw.R')

# Generate a table with 101 columns
X = cbind(matrix(rnorm(1:1010), nrow=10, ncol=101), sample(c(1, -1), 10, T))
X = as.data.table(X)
X[, weight:=ifelse(V102 == 1, 3, 1)]

namespaces = list(a = list(varName = sprintf('V%s',1:50), keepSpace=F),
                  b = list(varName = sprintf('V%s',51:101), keepSpace=F))

dt2vw(data=X, fileName='../out.vw', namespaces = namespaces, 
      target='V102', weight = 'weight', 
      tag = NULL, hard_parse = F)

### error

[...]

PR #766, opened 2015-08-25
R/dt2vw.R : possible to have more than 100 variables

SelimRaboudi:
solves #761 
New features:
--> possible to process a data table with more than 100 variables
--> appends the name of a categorical variable to its value
--> possibility to write the vw data file by chuncks: added the argument append = T

### DIFF
--- a/R/dt2vw.R
+++ b/R/dt2vw.R
@@ -17,7 +17,7 @@
                                     #variables like '_', or same variables perceived differently life "_var" and "var" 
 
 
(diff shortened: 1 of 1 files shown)


JohnLangford:
This looks good although my understanding of R is quite limited.    Anyone
else have comments?

-John

JohnLangford:
Merged.

All, this should unlock some VW functionality in R.

-John

PR #771, opened 2015-09-01
JNI support for CentOS 5 x86_64

jmorra:
It has become necessary for us at eHarmony to run the VW JNI layer on CentOS 5.  This pull request should build the JNI layer targeting CentOS 5.  A big problem with targeting CentOS 5 is that the default Boost that comes with CentOS 5 is too old to support VW.  As a result we need to install an upgraded Boost version.  I have chosen Boost 1.41 because that is the default on CentOS 6.  This means that in order to use this on CentOS (or any Red Hat variant including Scientific Linux) the user must execute the following command

```
sudo yum install epel-release
sudo yum install boost141-devel
```
More information can be found [here](http://www.rackspace.com/knowledge_center/article/install-epel-and-additional-repositories-on-centos-and-red-hat).  After installation a directory called `/usr/lib64/boost141` must exist.  If Boost 1.41 is installed then everything should work as expected.


JohnLangford:
Merged in, thanks.

-John

PR #772, opened 2015-09-03
Adding an example of using VW cluster mode on Apache Spark

jmorra:
This is a first pass at getting VW to work in cluster mode on Apache Spark.  There are a number of tricks in here (some that require insight into how the Spark engine works) that make this work.  I've seen it work on some of my data with Spark 1.2.0.  I'm not including this as a whole standalone maven project because it's meant to be template code that someone can take and do what they want with.  @deaktator if you have any comments about how this is written I'd love to hear about that.  This is written as one file because it's designed to be copy paste-able into the spark-shell with and it can be used right from there.

### DIFF
--- a/cluster/spark/VwSparkCluster.scala
+++ b/cluster/spark/VwSparkCluster.scala
@@ -7,7 +7,7 @@ import org.apache.spark.Logging
  * done asynchronously so data can be read and written to at the same time.
  * Created by jmorra on 1/22/15.
  */
-object PipeUtils {
+class PipeUtils(bufferSize: Int = 1 << 20) {
   import java.io._
 
   import scala.concurrent.ExecutionContext.Implicits.global
@@ -15,7 +15,6 @@ object PipeUtils {
   import scala.language.postfixOps
   import scala.sys.process._
 
-  private[this] val DEFAULT_BUFFER_SIZE = 1 << 20
   /**
    * This implicit class will allow easy access to streaming through external processes.  This
    * should work on a line by line basis just like Spark's pipe command.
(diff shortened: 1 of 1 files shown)


deaktator:
I'd allow for varying the buffer size. This could be done a variety of ways. For instance, change PipeUtils to a `class` parameterized by `bufferSize` which has a default value of `1 <<20`.  Then you can instantiate PipeUtils and import the implicit class. 

I would comment more on why the design for PipeUtils is as it is.  

Also, is asynchronous writing to the `InputStream` necessary in this use case, or should each line pulled from the `TraversableOnce` as it is requested from the `InputStream`? I know it is when all data is pushed to one node, I honestly just don't know here.

deaktator:
Also, maybe a try / finally in the async future code.

jmorra:
@deaktator you do realize that this is code you wrote, right.  You're more critical of your own code then mine :).

deaktator:
What can I say, I'm a perfectionist.

jmorra:
I've taken all of your suggestions.  PipeUtils now has a configurable bufferedSize that is accepted as a parameter when creating VwSparkCluster with a default value.  I've also added a try/finally and some more comments.

deaktator:
Beautiful. Well done.

JohnLangford:
Merged, thanks.

All, we now have an example of how to use VW with Spark.

-John

amit024003:
@jmorra 

Please can you help me to understand. How did you setup the vowpal library for Apache Spark? what command did you use to run it?

Thanks

deaktator:
Hey @amit024003, did you look at the VwSparkCluster class in

 https://github.com/jmorra/vowpal_wabbit/blob/e83a0da8d2507d18a78c53be43cfea3d3a39cd00/cluster/spark/VwSparkCluster.scala

It might answer your question. I don't know if that code is up to date, but that's my best guess.

amit024003:
@deaktator 

Thank you for quick response. I am looking how we can run this code on spark with "spark-submit".
I am looking for parameters that I need to set to run it.

jmorra:
This code was never intended to be run directly within `spark-submit`.  The goal of this code was to be able to give an idea of how it can be integrated into a spark project.  For instance at ZEFR we use it as part of our training pipeline.  We create an `RDD[String]` that is the VW formatted data.  Once we have data in memory we can call this class to actually train a model in parallel.

You could contribute a wrapper around this that simply loads data from some location (HDFS, NFS, S3, etc.) and then calls this, but for all of our work actually converting our raw data to VW format is expensive and costly.  Because of that we want this as part of a pipeline.

amit024003:
@jmorra 

Thank you for explaining it. I understood that I can use this class in pipeline as I am also going to use for training purpose. But I am struggling to run vw in spark cluster (setup with Hadoop and spark-2.1). I have tried to set env variables  (As spark.yarn.appMasterEnv.[EnvironmentVariableName]) after attaching vw package as archive to spark-submit. I am wondering how did you test your code. Please provide me parameters that you used to this code.

Issue #665, opened 2015-05-29
Some options are missing in vw --help


martinpopel:
According to active.cc, there are parameters `--simulation` and `--mellowness`, but the only way how to discover it is inspecting the source code. 

Problem 1:
`vw -h` prints only `--active`, but no  "simulation" and "mellowness". There is no indication that more info will be shown with `vw --active --help`.

Problem 2:
`vw --active --help` shows no info, it waits for stdin.

Problem 3:
https://github.com/JohnLangford/vowpal_wabbit/wiki/Command-line-arguments#active-learning-options is outdated. (I would fix it now, but I have never used active learning.)

Problem 4:
There is no check that after changing parameter names (refactorization), https://github.com/JohnLangford/vowpal_wabbit/wiki/Command-line-arguments will be updated. Due to problem1 and problem2, the wiki is the only place where we can point users.

JohnLangford:
Fixed #3.

How is it possible to fix #4?

-John

martinpopel:
> Fixed # 3.

OK, I've now substituted the remaining (in the description) `--active_simulation` with `--simulation`.

> How is it possible to fix # 4?

I think it is not worth to make any automation (wiki text automatically generated/validated from the C++ source codes). The wiki should include more details on the parameters (usage hints, links to separate wiki pages and published papers etc.) than `vw -h`. Just keep this wiki page in mind when merging pull requests that change/add the parameters. And/or ask volunteers to check&update the wiki once per several months.

trufanov-nok:
It looks like `use [option] --help for more info` is correct only for reductions that takes no additional arguments. For others you have to specify an argument that let them complete setup() function successfully. Example: `--nn 1 --help`. But some reductions won't work even with that or has additional requirements to complete their setup(). For example: 
```
vw --bfgs --help
enabling BFGS based optimization **without** curvature calculation
vw (bfgs.cc:947): you must make at least 2 passes to use BFGS
```
I think one of the ways to fix this is to move the reduction-specific command line options initialization code from setup() to some new reduction class function (ex, `bool options_init(vw&)`) that could be called from setup(). For example nn setup would be modified from:
```
base_learner* nn_setup(vw& all)
{
  if (missing_option<size_t, true>(all, "nn", "Sigmoidal feedforward network with <k> hidden units"))
    return nullptr;
  new_options(all, "Neural Network options")
  ("inpass", "Train or test sigmoidal feedforward network with input passthrough.")
[...]

PR #774, opened 2015-09-04
--probabilities

### LINKED ISSUE #566: multi class oaa progressive loss
Before looking at trying to produce a reproducible example (since currently this issue is showing up experimenting on a large kaggle data set) I wanted to check on the logic being employed and make sure there is a good chance the issue I am seeing is really an issue,

The data set is multi-class (9 classes). I compiled the current VW today from github. When I build model with 

vw.exe -d train.vw --oaa 9 -b 24 --loss_function logistic -f  VW.mod

The progressive loss is exceptionally good (0.26). Should this represent the multi-clas log loss?

When predicting a test set using raw_predictions and passing through  a sigmoid I find that the predictions (probabilities) are either all equal or one class will be essentially 1 and the other classes essentially zero. The "hand computed" multi class log loss ends up being quite horrible,

vw.exe -d test.vw -t -i  modVW.mod -r raw_pred.txt 

Is this an issue already known - I did find some discussions of issues with oaa raw_prediction but sounded like it was recently patched.

### LINKED ISSUE #724: Trying to recover probabilities from command line  in daemon mode not working.
My goal is to recover the probabilities from the command line when i am in daemon mode using python.

Based on the response given  below, which  I verified it works (thanks)
http://stackoverflow.com/questions/30764810/how-do-i-get-the-raw-predictions-r-from-vowpal-wabbit-when-running-in-daemon

I do the next:

#I start my process in Daemon
                                    os.system('vw -t -i model.vw --daemon --port 26542 --link=logistic -r /dev/stdout')

Then i start sending examples with the model above in order to get probabilities.
command='echo "2 | landstar system downgraded to sell at zacks lstr |a word_count:8" | nc localhost 26542'

                                      proc = subprocess.Popen(command,shell=True,stdout=subprocess.PIPE)

then I tried to read the output
                                      proc.stdout.read()
but the above instruction returns

'1.000000\n'  

It only returns the prediction :( which is correct. I  can see how the probabilities  were created  properly in the command line but I cannot  extract them . I tried different ways but none of them worked. I only can extract the prediction, which is useless for the task than i want to do.

Could you be so kind to tell me how I can get the probabilities?

I am using mac OS X Yosemite, version 10.10.3.
VW version 7.10.2

Thank you,

ja

martinpopel:
This new parameter `--probabilities` can be used in combination with `--loss_function=logistic` and either `--oaa` or `--csoaa_ldf=mc`. VW predictions (`-p`) will be directly class probabilities (summing to one). There is no need for specifying any `--link` option.

This PR fixes #566 and it should also fix #724.
See #566 for a detailed discussion.


martinpopel:
The Windows (appveyor) test is complaining [here](https://ci.appveyor.com/project/JohnLangford/vowpal-wabbit/build/1.0.452#L1095), but I have no idea why.
```
Test method cs_unittest.Test3Class.Test7and8 threw exception: 
VW.VowpalWabbitException: --probabilities can only be used with --oaa=N or --csoaa_ldf=mc
```
This [exception is thrown only if](https://github.com/martinpopel/vowpal_wabbit/blob/3a043da85437c8b4717089141913cbdaf1fcf204/vowpalwabbit/parse_args.cc#L990) `--probabilities` are on the command line, but [the reported source](https://github.com/JohnLangford/vowpal_wabbit/blob/master/cs_unittest/Test3.cs#L59-L70) of the problem seems to have no `--probabilities`.

Any hints? @eisber

JohnLangford:
I'm hesitating with this pull request.  To me, outputting probabilities
changes the type of the output, implying a change in the type of the
prediction, implying some alternate type definition parallel to multiclass
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/multiclass.h
or multilabel
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/multilabel.h
and an option to support these where appropriate.    So, instead of having
a magic flag in global_data.h, there would be a 'multiprobability' type
with standardized interfaces, and the flag for switching to
'multiprobability' mode would exist in oaa.cc and csoaa.cc rather than in
parse_args.cc.   Is that clear?

-John

martinpopel:
Thanks for the feedback. I agree `--probabilities` changes the type of output (unlike most other VW options), but that is exactly the purpose of this option. So let me react on the two remaining comments.

* Originally, I included the probabilities (`float* probs`) inside `polyprediction` (and I could easily add `multiprobability` struct type, so we can add more attributes later). The problem was that I need **both** the best class prediction and the probabilities for all classes. (I want to report both 0-1 loss and multiclass-log loss). So I moved `probs` outside of `polyprediction` union. Also for `csoaa_ldf` I need just one probability number (not an array) because each class has its own (sub)example.

* Currently, the `probabilities` flag is stored in `global_data.h` because it is needed in two reductions (`oaa` and `csoaa_ldf`). I can move the `--probabilities` option to `oaa.cc` and `csoaa.cc` (i.e. duplicate its definition), but probably I will still need to store it inside `all` (similarly, e.q. `csoaa_rank` aka `multilabel_prediction` is stored inside `all` although it is valid only for some reductions). The problem would be that `vw -h` would not list the `--probabilities` options (see #665).

eisber:
I tried to reproduce the issue (your fork/master). But didn’t hit the issue. The test that fails simply passes -k --power_t 0.45 -f models/0002c.model.

Markus

JohnLangford:
Yeah, there is nothing wrong with changing the output type.

My fear is that there are many prediction types, each of which could 
take a field in the example datastructure.  But, having each prediction 
type take it's own field means that the example datastructure is left 
with a bunch of dead fields in any individual use.    This seems 
undesirable in three ways:
(1) there is a creeping cruft of wasted RAM.
(2) other people might reuse the datastructure in any of several 
surprising/unpleasant ways.
(3) keeping track of which fields are associated with prediction becomes 
more difficult.  This matters when refactoring comes into play.

Can you just infer the best class from the probabilities per class?  
This seems amortized by the writing of the class predictions themselves 
so it should be ok computationally.

Wrt listing --probabilities, this seems a separate issue.  Perhaps we 
should modify things to print out all options all the time?

W.r.t. having something in the global datastructure, it seems we need 
the type of the input label and the output prediction to be 
(effectively) stored in the global structure so that appropriate parsing 
and printing can occur.  I have not effectively modularized that, but 
want to do so.  Doing so is beyond the scope of what you are trying to 
do so a flag to specify what the output type is in the 'all' 
datastructure seems reasonable.

-John

martinpopel:
So moving `float* probs` and `float prob` inside `polyprediction` is the suggested way to go? I will try it.

Ad (1) In online learning we have always just one example in memory at a time. Empty pointer and unused float take just few bytes, so it should not be any concern. So the only problem with wasted RAM is when using `--bfgs --probabilities`? (I admit I have not tested nor even considered this combination.)
BTW: regarding memory wasting, I would rather focus on the default case with stride=4, but only 3 floats are used (and maybe the four-byte precision is needed only for the weight and half precision would be enough for the remaining two floats, but I am not sure about the speed).

Ad (2) OK, but reusing the members of `union` may be even more unpleasant and surprising (especially if someone stores e.g. `scalar` and someone else reads it as a `multiclass`). I mean we need some documentation with best practices anyway.

Ad (3) OK (cf. `partial_prediction` and `updated_prediction` and `topic_predictions`, which are outside of `polyprediction`).

One more questions: Currently, I calloc_or_die the `probs` array [inside oaa_setup](https://github.com/JohnLangford/vowpal_wabbit/pull/774/files#diff-4ab1ab4410cdc0261cf020c9b02ba7abR179), i.e. just once when VW starts.
Is it OK, or should I allocate a new array for each example (and free it inside finish_example)? I'm now concerned with compatibility with batch learning (and speed in online learning).

JohnLangford:
Yes, thank you.

The wasted RAM is not a great concern---every individual instance of it 
is minor.  It's just the kind of thing that can creep in from many 
sources and is hard to reverse when the total becomes a problem.

(1) The number of examples live in the system is ring_size, not 1. With 
the default update, the extra piece of memory actually is used as a 
cache between the pred_per_update() pass and the update() pass. 
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/gd.cc#L101 


(2) Stepping on the union the wrong way is indeed very bad.  We have a 
weak typing in effect with polyprediction, and there is a good case for 
stronger typing.  (This is better than a year ago when we effectively 
had no typing with void*...)  Suggestions?

(3) Yeah, I want to get rid of those.  Unifying and strongly typing all 
the predictions seems pretty desirable.

You should let v_array<> alloc everything inside of polyprediction and 
then have a deleter that is passed in as here: 
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parser.cc#L1185 
Why?  This keeps all the memory associated with an example rather than 
an element of reduction stack.  The first works as expected under 
composition, but the second would not.

-John

martinpopel:
Finally, I found time to get back to `--probabilities`.
I tried moving `float* probs` and `float prob` inside the `polyprediction`. The problem is that with `--probabilities --oaa`, I need both the predicted label and the probabilities of all labels. The label is needed for computing zero-one loss (I decided to print the zero-one loss in the progress reports in *average loss* and *since last* columns) and also for the *current predict* column. So I would need to recompute the predicted label at least three times on various places in the code (oaa.cc:predict_or_learn, oaa.cc:finish_example_probabilities, multiclass.cc:finish_example).

> Can you just infer the best class from the probabilities per class? This seems amortized by the writing of the class predictions themselves so it should be ok computationally.

Yes, I can infer the predicted label (best class) from the probabilities per class or from `o.pred[i].scalar`. Should I recompute it three times per example, save few bytes of memory (default ring_size is 256, uint32_t takes 4 bytes, that is 1kB) at the cost of slower learning and predictions? Or should I store the predicted label somewhere within the example? Should I create new struct for probabilities which would contain both the predicted class as well as the probabilities?

> You should let v_array<> alloc everything inside of polyprediction and then have a deleter that is passed in as here: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parser.cc#L1185

Thanks for pointing this line. Originally, I had tried to mimic the multilabel's implementations, but I was getting segfaults. As I see it, the general problem is that when implementing one reduction, I need to change about a dozen places in different files, so I need to have quite a deep knowledge of VW. This is just one example (parser.cc contains `if (all.multilabel_prediction)` although `label_v.erase()` is called also multilabel.h:parse_label().)

JohnLangford:
On 09/29/2015 10:32 AM, Martin Popel wrote:
>
> Finally, I found time to get back to |--probabilities|.
> I tried moving |float* probs| and |float prob| inside the 
> |polyprediction|. The problem is that with |--probabilities --oaa|, I 
> need both the predicted label and the probabilities of all labels. The 
> label is needed for computing zero-one loss (I decided to print the 
> zero-one loss in the progress reports in /average loss/ and /since 
> last/ columns) and also for the /current predict/ column. So I would 
> need to recompute the predicted label at least three times on various 
> places in the code (oaa.cc:predict_or_learn, 
> oaa.cc:finish_example_probabilities, multiclass.cc:finish_example).
>
>     Can you just infer the best class from the probabilities per
>     class? This seems amortized by the writing of the class
>     predictions themselves so it should be ok computationally.
>
> Yes, I can infer the predicted label (best class) from the 
> probabilities per class or from |o.pred[i].scalar|. Should I recompute 
> it three times per example, save few bytes of memory (default 
> ring_size is 256, uint32_t takes 4 bytes, that is 1kB) at the cost of 
> slower learning and predictions? Or should I store the predicted label 
> somewhere within the example? Should I create new struct for 
> probabilities which would contain both the predicted class as well as 
> the probabilities?
>

I think infering the label on the fly is probably (?) better.  There are 
many other algorithms which would like to output a vector of float 
predictions---see for example the LDA code.  Hence, supporting a 'vector 
of floats as a prediction' type as first class seems inherently 
desirable.   Given this, the added value of also supporting a 'vector of 
floats plus an int' prediction type seems less clear.

>     You should let v_array<> alloc everything inside of polyprediction
>     and then have a deleter that is passed in as here:
>     https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parser.cc#L1185
>
> Thanks for pointing this line. Originally, I had tried to mimic the 
> multilabel's implementations, but I was getting segfaults. As I see 
> it, the general problem is that when implementing one reduction, I 
> need to change about a dozen places in different files, so I need to 
> have quite a deep knowledge of VW.
>

It's certainly desirable to isolate everything in a reduction to one 
file.  I think that's generally true of most reductions, _except_ when 
you start messing with label or prediction types which are in a messy 
nonmodular state.  I need a week or two to go through and harden 
definitions here similar to what I did for most reductions last year.

-John

JohnLangford:
Closing, as new attempt is in process.

PR #779, opened 2015-09-10
Java interface build issue fixes

marko-asplund:
disable doclint check to make build work with Java 8; rename JNI native lib file, so that the Java lib can load it


marko-asplund:
this PR was superseded by #780.

PR #780, opened 2015-09-11
Java interface build issue fixes

marko-asplund:
disable doclint check to make build work with Java 8; rename JNI native lib file, so that the Java lib can load it

### DIFF
--- a/java/Makefile
+++ b/java/Makefile
@@ -36,6 +36,7 @@ pom_version: pom.xml
 target/vw_jni.lib: src/main/c++/vw_VW.cc ../vowpalwabbit/main.o ../vowpalwabbit/libvw.a ../vowpalwabbit/liballreduce.a
 	mkdir -p target;
 	$(CXX) -shared $(FLAGS) -o $@ $< $(VWLIBS) $(STDLIBS) $(JAVA_INCLUDE)
+	osf=$$(perl script/os_info.pl osfamily) && mv target/vw_jni.lib target/vw_jni.$$osf.lib
 
 clean:
-	rm -f target/vw_jni.lib
+	rm -f target/vw_jni\.*.lib
--- a/java/script/os_info.pl
+++ b/java/script/os_info.pl
@@ -1,0 +1,49 @@
+#!/usr/bin/env perl
+
+use strict;
+use warnings;
+
+if ($#ARGV != 0) {
+  die "invalid arguments, exiting"
+}
+
+my $cmd = $ARGV[0];
+
+if ($cmd eq "osfamily") {
+  my $f = get_os_family($^O);
+  print "$f." . get_os_arch($f);
+} else {
+  print "unknown command\n";
+}
+
+sub get_lsb_release {
+  my $opt = shift;
+  my $res=(split(':', `lsb_release $opt`))[1];
+  $res =~ s/\s+//g;
+  return $res;
+}
+
+sub get_os_arch {
+  my $os_family = shift;
+  my $arch = `uname -m`;
+  if ($os_family =~ /ubuntu/i) {
+    $arch = `dpkg --print-architecture`;
+  }
+  chomp($arch);
+  return $arch;
+}
+
+sub get_os_family {
+  my $os = shift;
+  if ($os =~ /darwin/i) {
+    return "Darwin";
+  } elsif ($os =~ /linux/i) {
+    my $dist = get_lsb_release("-i");
+    $dist =~ s/ /_/g;
+    my $fullVersion = get_lsb_release("-r");
+    my $majorVersion = (split('\.', $fullVersion))[0];
+    return "$dist.$majorVersion";
+  } else {
+    return "unknown";
+  }
+}


JohnLangford:
Jonathan, can you take a look at this?

-John

jmorra:
After looking at this code it would appear that the author is replacing the behavior in [bin](https://github.com/JohnLangford/vowpal_wabbit/blob/master/java/src/main/bin/build.sh) with a Perl script.  This seems fine but it requires updating that script and ensuring that it still works as expected on all the OS's that the JNI layer claims to support.  This should be very straightforward to do but without this extra check I fear that the Java build may not be working.

JohnLangford:
Can this be setup to be automatically checked?

-John

marko-asplund:
To give you a bit more background on the changes, the use case that I was aiming to support was "to be able to easily build the VW Java interface for the particular platform I'm using without having to specify  platform information during build". There's probably other ways to support this in addition to the approach I took and an alternative would welcome, as well.

AFAIK, `build.sh` and `release.sh` seem to be geared towards building the interface for multiple platforms and creating a release to be published in an artefact repository.

jmorra:
`build.sh` is indeed written for cross platform building.  However, the way that they work is by moving a file called `vw_jni.lib` to a platform specific location.  If you rename the file to that platform specific location within the build script then `build.sh` will fail.  I think an easy fix is the following
1.  Remove all the mv lines from `build.sh` and verify that calling `build.sh` still works.  By this I mean that the exact same directory structure (with the same file names) is created for each OS specified in `build.sh`.
2.  Verify that `mvn clean install` works and doesn't have any failing tests.

If those two things are good then this seems totally reasonable.

If, on the other hand, you just want compilation to work for Java 8 then I would suggest doing a PR with just the changes to the pom (which make total sense) and not worrying about the build process.  Also, I've updated the JNI layer and you can now issue your own call to either `System.load()` or `System.loadLibrary()` before calling the constructor and it should work fine.

marko-asplund:
I created a separate PR #786 for the doclint workaround.

Doesn't the above mean that anyone wanting to build JNI by himself would need to use the `build.sh` and also have Docker installed on their system? It would be nice to be able to build for just a single system without these additional build-time dependencies.

jmorra:
I have fixed the JavaDoc instead of just masking the problem [here](https://github.com/JohnLangford/vowpal_wabbit/pull/788).  I still think your point about doing the platform specific builds at Make time could have merit.  I had envisioned it working in such a way that if you wanted to build your own JNI layer then that would imply you're using an OS that is currently not supported.  Because of this you'd want a generic library name that you could load yourself with `System.loadLibrary(String)`.  This is the reason in the constructor for [VW](https://github.com/JohnLangford/vowpal_wabbit/blob/master/java/src/main/java/vw/VW.java) it first checks if a library is already loaded.  In your case doesn't this process make more sense?  If I'm misunderstanding something, please let me know and I can update things accordingly.

JohnLangford:
There is another reason why someone might want to build: they may want to update to the latest VW master and/or be working with their own custom fork.   Doing this without doing all the docker stuff seems plausibly desirable.

JohnLangford:
I've lost track---what is the status of this PR?

-John

jmorra:
I think we might want to cherry pick the perl script if it works, but that'll require changes in `build.sh` as well.  The doclint stuff has been superseded by [788](https://github.com/JohnLangford/vowpal_wabbit/pull/788).  I think the best way to proceed is to close this one and submit a new one that just addresses the issues about the build environment.  The Java 8 stuff is no longer needed and shouldn't be added cause it just masks issues with the JavaDoc.

marko-asplund:
@jmorra I removed the doclint workaround from this PR.

JohnLangford:
Marko, What does this code do that can't already be done?

JohnLangford:
(I mean, the current version of the code).

marko-asplund:
The current version allows a user to build the VW Java interface just for the particular platform he's currently working on, without needing to install Docker. As @jmorra noted, the `build.sh` script probably would need to be updated accordingly. Also, I've tested this change on Ubuntu 13 (x64), but testing should be performed on a wider variety of different platforms.

Another approach, would be to use a Maven plugin e.g.
https://github.com/trustin/os-maven-plugin
for obtaining OS information and then the resource plugin to copy the native lib in-place (with the right name).
this way the code that does names the library file and the code that resolves the name would arguably be  more similar since both use JVM properties.

@jmorra what do you think?

JohnLangford:
My understanding is that you can use maven to build directly on any platform rather than using build.sh (and docker).  Is that correct?

jmorra:
Correct

JohnLangford:
So this patch does some OS-specific naming that is not done by the maven 
build?  How useful/necessary is that?

-John

marko-asplund:
AFAIK, you can't get Maven to do OS-specific naming out-of-the-box.
You would need something like the `os-maven-plugin` I mentioned for this.
I think this would be a better approach than using the Perl script and I'm currently experimenting with it. The Maven part of this is straight-forward, but additionally we have to synchronize the library loading code (name resolution).

jmorra:
@marko-asplund it would appear this is attempting to accomplish the same thing as #822.  Do you think this one can be closed in favor of #822?

marko-asplund:
@jmorra yes, closing this issue now.

PR #786, opened 2015-09-17
Disable doclint check to make build work with Java 8

marko-asplund:
Workaround for a doclint check failure.
The Javadoc should also be fixed.

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -14,6 +14,10 @@
     </description>
     <url>https://github.com/JohnLangford/vowpal_wabbit</url>
 
+    <properties>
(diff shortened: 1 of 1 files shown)


jmorra:
I was thinking about this more, and wouldn't a better solution be to fix the broken docs in this case?

jmorra:
I have gone ahead and fixed the JavaDoc to be compatible with Java 8.  I think it's better practice to fix the underlying problem then just ignore the issues especially cause this is a public project.  The pull request can be found [here](https://github.com/JohnLangford/vowpal_wabbit/pull/788).

marko-asplund:
@jmorra yes, much better that way! thanks :thumbsup:

PR #792, opened 2015-09-18
Macfix

eisber:
@alekh I merged from the feature branch to master and found that there were some more asserts (on the ordering). Do you have a suggestion on what to do if we discover unordered features?

### DIFF
--- a/vowpalwabbit/interact.cc
+++ b/vowpalwabbit/interact.cc
@@ -2,7 +2,7 @@
 Copyright (c) by respective owners including Yahoo!, Microsoft, and
 individual contributors. All rights reserved.  Released under a BSD (revised)
 license as described in the file LICENSE.
-*/
+ */
 #include <sstream>
 #include <float.h>
 #include "reductions.h"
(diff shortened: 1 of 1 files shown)


alekh:
If we discover features out of order, then we should just probably just
give up on that example and return 0?

Alekh

eisber:
So it’s a bit different then the other cases as we’ve already found some features, but am fine with returning 0 too.

Is this a decision?

Ideally I’d throw an exception (so we should put a comment there).


From: alekh
Date: Friday, September 18, 2015 at 16:06
Cc: Markus Cozowicz
Subject: Re: [vowpal_wabbit] Macfix (#792)

If we discover features out of order, then we should just probably just
give up on that example and return 0?

Alekh

alekh:
I agree that we would throw an exception if we could, but we seem to have
troubles there. Should we print a message to stderr and return 0?

Alekh

JohnLangford:
The general structure of things with VW is that stderr is used for diagnostics while stdout is used for errors or requested output (like audit).

eisber:
Fine with me. Do you want to do this in release & debug, or just debug as it currently is?

eisber:
So as discussed I’d propose the use of a fancier logging infrastructure using function pointers/lambdas.

Something along:

vw.audit([f] (out) { out << “feature: “ << f.x; } );

JohnLangford:
This is merged in, although I tweaked the way that predict is called---no need to call it if is_learn.

PR #794, opened 2015-09-19
R/vw.R internal AUC and ROC plots. 

martinbel:
Also added a complete example of dt2vw.R and vw.R. Also deleted the old files, vw.R and vw_example.R.


JohnLangford:
This is complaining that it can't merge.  Can you sync with master?

martinbel:
This should fix it: https://github.com/JohnLangford/vowpal_wabbit/pull/797

JohnLangford:
I believe this is 'close because it has been superceded".   Reopen if otherwise.

PR #797, opened 2015-09-20
Changes in R folder. 

martinbel:
Added AUC - added a readme for R scripts. And the example is in rvw_example folder.


JohnLangford:
Should we be including .png and .pdf results in the repository?  My 
inclination is 'no', because:

(a) it makes download be relatively slow.
(b) These are derived files, and hence they could be overwritten just 
because of an OS change rather than any semantic change.

-John

martinbel:
I understand, I wasn't aware of the policy you had. I sounds reasonable. Would you like to remove them and make another pull request?

JohnLangford:
You can probably just remove them from the branch on the pull request.

-John

martinbel:
@JohnLangford Done.

martinpopel:
> You can probably just remove them from the branch on the pull request.

Note that this way the relatively big png/pdf files still remain in the git history.
So the argument

> (a) it makes download be relatively slow.

still applies. (The history is compressed, but that does not help for png.)
If `git clone` should be fast, we need to get the files out of history.
The easiest way is to delete the branch and recreate again.

martinpopel:
@martinbel and thanks for the examples and documentation. Well done.

martinbel:
@martinpopel Ok. I see. I'll do that, no problem.

JohnLangford:
Good point.

-John

JohnLangford:
Closing as superceded.

R/rvw_example/dt2vw.R
```diff
@@ -1,251 +0,0 @@
...
-  parsingVar <- function(x, keepSpace, hard_parse)
-  {
-    #remove leading and trailing spaces, then remove special characters then remove isolated underscores.
-    if(!keepSpace)
-      spch = specCharSpace
-    else
-      spch = specChar
-    
-    if(hard_parse)
-      gsub('(^_( *|_*)+)|(^_$)|(( *|_*)+_$)|( +_+ +)',' ', gsub(specChar,'_', gsub('(^ +)|( +$)', '',x)))
```
zachmayer: I think the regexes for cleaning all the data types should be user-specifiable.  Also, consider using the [stringi package](https://cran.r-project.org/web/packages/stringi/index.html) for much faster regexes (like orders of magnitude faster on some data)

R/rvw_example/dt2vw.R
```diff
@@ -1,251 +0,0 @@
...
-  #required packages
-  require(data.table)
-  
-  if(class(data)[1] != "data.table")
-    data = data.table(data)
-  
-  #change target if its boolean to take values in {-1,1}
-  if(is.logical(data[[target]]) | sum(levels(factor(data[[target]])) == levels(factor(c(0,1)))) == 2)
-  {
-    data[[target]][data[[target]] == TRUE] = 1
```
zachmayer: using `=` to change data will force a copy of the whole data.table.  Try the [`data.table::set`](http://www.inside-r.org/packages/cran/data.table/docs/set) function instead.

R/rvw_example/dt2vw.R
```diff
@@ -1,251 +0,0 @@
...
-    if(hard_parse)
-      gsub('(^_( *|_*)+)|(^_$)|(( *|_*)+_$)|( +_+ +)',' ', gsub(specChar,'_', gsub('(^ +)|( +$)', '',x)))
-    else
-      gsub(spch, '_', x)
-  }
-  
-  ###   NAMESPACE LOAD WITH A YAML FILE
-  if(typeof(namespaces) == "character" && length(namespaces) == 1 && str_sub(namespaces, -4, -1) == "yaml")
-  {
-    print("###############  USING YAML FILE FOR LOADING THE NAMESPACES  ###############")
```
zachmayer: Use `message` instead of `print` so the user can suppress it with `suppressMessages` if they desire.

R/rvw_example/dt2vw.R
```diff
@@ -1,251 +0,0 @@
...
-  ###   FULL VW DATA STRING (NOT FORMATTED YET) : (%target %weight |A num1:%f %s |B num2:%f %s)
-  if (!is.null(tag)){
-    formatDataVW = paste0(formatDataVW, collapse = '|')
-  } else {
-    formatDataVW = paste0(formatDataVW, collapse = ' |')
-  }
-  
-  formatDataVW = paste0("sprintf2('", formatDataVW, "',",argexpr, ")")
-  ###   FORMATTING USING THE DATA.TABLE DYNAMICS TO OBTAIN THE FINAL VW DATA STRING
-  temp = data[, eval(parse(text = formatDataVW))]
```
zachmayer: This is very clever, I like it!

zachmayer:
I left a few comments on your code too.  I couldn't find the commit where the `dt2vw.R` file was created, so I commented on this PR instead, where it looks like you removed a duplicate copy.

JohnLangford:
These are good ideas, but I don't have time to investigate personally.  
Can you implement and send a pull request?

-John

zachmayer:
Will do

PR #798, opened 2015-09-21
changes in R folder - AUC in vw.R (no png-pdf)

martinbel:
Equivalent to #797 but without the extra png-pdf files.


zachmayer:
Wouldn't it make sense to make `vw.R` and `dt2vw.R` an R package, perhaps as a [submodule](http://git-scm.com/book/en/v2/Git-Tools-Submodules) with it's own repository?  Then we can do things like use travis to run R CMD check and run unit tests, etc.

zachmayer:
I can help setup the submodule/package/testing

JohnLangford:
I'd certainly welcome help.  I don't know or use R, so someone more 
familiar would need to handle it.

-John

martinbel:
I've proposed this previously, but the argument not to do it was that these are more "informal" scripts and a package should have a more cohesive functionality, a plan to implement an entire API, etc. I think the submodule idead is great, and will make it easier to expand the current functions functionality.

zachmayer:
Personally, I think that any informal script that's being used by more than one person is a great candidate to turn into a package.  I like `devtools::install_github()` a lot more than cut and paste =D.

I also think that any package on github only basically comes with no promises.  It might not work, it might be incomplete, it might never get finished.  Once you put a package on CRAN, that's where the promises of completeness and correctness come into play.

Anyways, this is just my personal opinion, but I think we could make a pretty clean little package with a function for writing VW datasets and a function for running VW models on those datasets.

PR #799, opened 2015-09-22
Extend VW Java interface to support topic predictions

marko-asplund:


java/vw-java-demo/src/main/java/org/vw/demo/VWDemo.java
```diff
@@ -0,0 +1,18 @@
+
+package org.vw.demo;
+
+import vw.VW;
+
+public class VWDemo {
+  private static String VW_DATA_DIR = "/home/masplund/vw/5cb7389b4b/ok/data";
+
+  public static void main(String ... args) {
+    String cmd = String.format("-i %s/predictor.vw --lda_epsilon 100000 -t --quiet", VW_DATA_DIR);
```
sharatsc: please leave out the --lda_epsilon parameter specification in the demo code.

java/src/main/java/vw/VW.java
```diff
@@ -115,6 +115,18 @@ public float predict(String example) {
         }
     }
 
+  public float[] getDocTopicDistribution(String example) {
```
sharatsc: May name it getTopicPredictions to reflect c++ api?

marko-asplund:
@sharatsc PR updated based on your comments, thanks!
I'm still getting an empty prediction vector, though.

marko-asplund:
PR updated and seems to work now.

Please review.

marko-asplund:
@jmorra Could you please review this PR?

jmorra:
I just had my second child two days ago so I princely can't look at this for a little bit. Maybe @deakator can have a look. If this doesn't get any response in 3 weeks please message me again. I should be back at work by then.

deaktator:
@marko-asplund Looking right now.

marko-asplund:
@jmorra congratulations!

deaktator:
Hey @marko-asplund.  I looked at your PR and I have a few suggestions.  

## Convert example usage to a unit test

My first suggestion is turning your example usage into a unit test to show that it works.

```java
  String cmd = String.format("-i %s/predictor.vw -t --quiet", vwDataDir);
  System.out.println("cmd: " + cmd);
  VW vw = new VW(cmd);
```

The reason I suggest this is that when I run the following, I get an error on VW 8.0.0 (installed via *brew*).

```bash
(paste -d' ' <(jot -b '|' 6) <(jot -b 0 6) <(jot 6 2); 
 paste -d' ' <(jot -b '|' 6) <(jot -b 1 6) <(jot 6 2)) > test.dat

vw --quiet \
   --lda 2 \
   -d test.dat \
   -b 3
   --passes 100 \
   --cache_file vw.cache \
   -f vw.lda.model \
   --readable_model vw.lda.readable.model \

echo "| 1 7" | vw -i vw.lda.model -t --quiet
```

Specifically the error is `vw: option '--lda' cannot be specified more than once` which occurs when trying to load `vw.lda.model`.  I get the same thing when I try to do it from Java using:

```xml
<dependency>
  <groupId>com.github.johnlangford</groupId>
  <artifactId>vw-jni</artifactId>
  <version>8.0.0</version>
</dependency>
```

I included a test.  I want to make sure it works for you.  It should probably be put in with the VW JNI project code and not in the example usage. 

```java
package vw;

import java.io.File;

import org.junit.Test;
import org.junit.runner.RunWith;
import org.junit.runners.BlockJUnit4ClassRunner;
import static org.junit.Assert.*;

@RunWith(BlockJUnit4ClassRunner.clas)
public final class LdaTest {
  private static final int REPS = 100;

  public LdaTest() {}

  @Test
  public void testLda() throws Exception {
    final String[] data = new String[] {
      "| 0 2", 
      "| 0 3",
      "| 0 4",
      "| 0 5",
      "| 0 6",
      "| 0 7",
      "| 1 2",
      "| 1 3",
      "| 1 4",
      "| 1 5",
      "| 1 6",
      "| 1 7"
    };
    
    final String model = getModelPath(false);
    final String readableModel = getModelPath(true);

    System.out.println("model: " + model);
    System.out.println("readable model: " + readableModel);

    writeVwModelToDisk(model, readableModel);
    assertTrue(rehydratedModelOk(model));
  }

  private static void writeVwModelToDisk(final String model, final String readableModel) throws Exception {
    final VW vwModel = new VW("--quiet -b 3 --lda 2 -f " + model + " --readable_model " + readableModel);

    for (int i = 0; i < REPS; ++i) {
      for (String d: data) {
        vwModel.learn(d);
      }
    }

    vwModel.close();
  }

  /** 
   * Instantiate model to see if it's possible or whether we get an exception.
   * @param model path to binary VW model.
   */
  private static boolean rehydratedModelOk(final String model) {
    VW v = null;
    try {
      v = new VW("-i " + model + " -t --quiet");
      return true;
    }
    catch (Throwable e) {
      return false;
    }
    finally {
      try { 
        v.close(); 
      } 
      catch (Throwable ignored) {}
    }
  }

  private static String getModelPath(final boolean readable) throws Exception {
    val f = File.createTempFile("vwtest", (readable ? ".readable" : "") + ".lda.model");
    val p = f.getAbsolutePath();
    f.deleteOnExit();
    f.delete();
    return p;
  }
}
```

## Interface is too specific

My second thought is that the interface seems too specific.  It would be nice to support a single interface where a vector of predictions is produced.  Something like 

```
public float[] multipredict(String example);
```

OR 

```
public List<Float> multipredict(String example);
```

@jmorra and I have gone back and forth about the array vs. `java.util.Collection` issue.  I'm not sure which is better.  At least with a collection, we can enforce immutability by passing the result to `java.util.Collections.unmodifiableList` before returning it.  That's my preference but it was be slightly slower due to boxing each value and wrapping.  Another possibility is to make a custom `java.util.List` type that extends `RandomAccess` but stores the values in an array.  It would also have a method `getRaw` or something that doesn't need to box the floats.  This would be the same in spirit as [fastutil](http://fastutil.di.unimi.it) but would not have the bloat of including the whole library.


*What do you think?*

marko-asplund:
@deaktator thanks for your review comments!

I've updated the PR to address the comments - please re-review and test.
The unit test is included with small modifications.

Also, it would be great, if you could check Java_vw_VW_multipredict for potential memory leaks.

Wrt. array vs. collection discussion, I'd suggest we use float[] for now, since that way the user's hands wouldn't be tied, and it's quite trivial for users to wrap the interface, if they want to.

deaktator:
Hey @marko-asplund You're build is actually failing.  I noticed there is a bug #804 that doesn't tell you but you should look at the travis logs to see the issue.  I created the issue #804 and created a pull request to fix it.  For now, check the travis reports after you check in to make sure the build is working.


[https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/82099343](https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/82099343)

marko-asplund:
Java version downgraded back to version 6.

@deaktator looks like the CI server has an old Java version (`1.7.0_76`), so upgrading source level to version 8 made the CI build fail.

deaktator:
@marko-asplund Unfortunately, I don't think that's the extent of the problem.  I think (see #804) that the builds were failing but silently succeeding.  For instance, see the [last java-related  PR](https://github.com/JohnLangford/vowpal_wabbit/pull/788).  If you look at the [travis build log](https://s3.amazonaws.com/archive.travis-ci.org/jobs/80869604/log.txt) and search "failure", you'll see that the build actually failed even though Travis thought it passed.  I created #804 and created the pull request to update the bug in the `.travis.yml` file that fixes this.  Unfortunately, the build is broken because the previous merge actually seems to be broken.  

Did you run the following locally with success?

```bash
make clean
make
cd java
mvn test
```

marko-asplund:
The build is still failing, but now for a different reason:
https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/82221191

The problem seems to caused by the native library not getting properly loaded, possibly due to a platform specific library file naming problem (see issue #780).

marko-asplund:
here's some relevant extracts from the CI build logs:

https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/82099343

```[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.2:compile (default-compile) on project vw-jni: Fatal error compiling: invalid target release: 1.8 -> [Help 1]```

https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/82221191

```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running vw.VWTest
Tests run: 26, Failures: 0, Errors: 24, Skipped: 2, Time elapsed: 0.748 sec <<< FAILURE! - in vw.VWTest
testBadVWArgs(vw.VWTest)  Time elapsed: 0.105 sec  <<< ERROR!
java.lang.UnsatisfiedLinkError: no vw_jni in java.library.path
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1886)
	at java.lang.Runtime.loadLibrary0(Runtime.java:849)
	at java.lang.System.loadLibrary(System.java:1088)
	at vw.jni.NativeUtils.loadOSDependentLibrary(NativeUtils.java:112)
	at vw.VW.loadNativeLibrary(VW.java:76)
	at vw.VW.<init>(VW.java:63)
	at vw.VWTest.setup(VWTest.java:41)
```

**Comment by deaktator** [comment]
@marko-asplund Yup.  I've already seen those and forwarded to @jmorra.  This was happening even in the last **accepted** PR.  The issue is with the `.travis.yml`.  This needs to be fixed.  I think it should now be 

```
cd java && make && mvn clean install && cd ..
```

Testing locally to see if that's the case.

**Comment by marko-asplund** [comment]
The discussion related to the library name resolution in issue #780 is still inconclusive.
My local repo includes the following workaround:

```
diff --git a/java/src/main/java/vw/jni/NativeUtils.java b/java/src/main/java/vw/jni/NativeUtils.java
index 16c8f68..3f7e9ac 100644
--- a/java/src/main/java/vw/jni/NativeUtils.java
+++ b/java/src/main/java/vw/jni/NativeUtils.java
@@ -103,7 +103,7 @@ public class NativeUtils {
      */
         String osFamily = getOsFamily();
-        String osDependentLib = path + "." + osFamily + "." + System.getProperty("os.arch") + suffix;
+        String osDependentLib = path + suffix;
         if (NativeUtils.class.getResource(osDependentLib) != null) {
             loadLibraryFromJar(osDependentLib);
         }
```

**Comment by jmorra** [comment]
Please don't commit that last change as it won't run correctly when using `build.sh`.  I've already commented on a fix in [805|https://github.com/JohnLangford/vowpal_wabbit/pull/805].

File: java/src/main/java/vw/VW.java [hunk 3]
```diff
@@ -115,6 +115,18 @@ public float predict(String example) {
         }
     }
 
+  public float[] multipredict(String example) {
```
jmorra: Can we change this to conform with the other methods?  By throwning an IllegalStateException the caller is informed about a problem instead of silently succeeding and returning an empty array.  I would suggest something like this

```
    public float[] multipredict(String example) {
        lock.lock();
        try {
            if (isOpen) {
                return multipredict(example, nativePointer);
            }
            throw new IllegalStateException("Already closed.");
        }
        finally {
          lock.unlock();
        }
    }
```

jmorra:
Hey, I made a few comments about the PR.  I would think that vw_VW.cc needs to be updated.  I specifically wrote code that should stop multiprediction from happening.  I think that we should apply the following diff as well

```
Mac:~/git/vowpal_wabbit$git diff
diff --git a/java/src/main/c++/vw_VW.cc b/java/src/main/c++/vw_VW.cc
index 4966c7d..1624dab 100644
--- a/java/src/main/c++/vw_VW.cc
+++ b/java/src/main/c++/vw_VW.cc
@@ -37,11 +37,6 @@ JNIEXPORT jlong JNICALL Java_vw_VW_initialize(JNIEnv *env, jobject obj, jstring
     try {
         vw* vwInstance = VW::initialize(env->GetStringUTFChars(command, NULL));
         vwPtr = (jlong)vwInstance;
-        if (vwInstance->multilabel_prediction) {
-            Java_vw_VW_closeInstance(env, obj, vwPtr);
-            vwPtr = 0;
-            throw_java_exception(env, "vw/exception/IllegalVWInput", "VW JNI layer only supports simple and multiclass predictions");
-        }
     }
     catch(...) {
         rethrow_cpp_exception_as_java_exception(env);
diff --git a/java/src/test/java/vw/VWTest.java b/java/src/test/java/vw/VWTest.java
index 243e278..f26fcb5 100644
--- a/java/src/test/java/vw/VWTest.java
+++ b/java/src/test/java/vw/VWTest.java
@@ -3,7 +3,6 @@ package vw;
 import org.junit.*;
 import org.junit.rules.ExpectedException;
 import org.junit.rules.TemporaryFolder;
-import vw.exception.IllegalVWInput;
 
 import java.io.*;
 import java.util.Map;
@@ -252,14 +251,6 @@ public class VWTest {
     }
 
     @Test
-    public void testMultiLabel() {
-        thrown.expect(IllegalVWInput.class);
-        thrown.expectMessage("VW JNI layer only supports simple and multiclass predictions");
-        VW vw = new VW("--quiet --multilabel_oaa 3");
-        vw.close();
-    }
-
-    @Test
     public void testVersion() throws IOException {
         String actualVersion = VW.version();
         String expectedVersion;
```
But I think that @JohnLangford might be better to comment on this, but I think that LDA predictions should handle the case that I was trying to prevent before.  What do you guys think?

marko-asplund:
@jmorra multipredict() now throws IllegalStateException if VW object is closed.
wasn't planning to commit my local library name resolution hack :-), just posted that as info for anyone wanting to get their local build working quickly.

should I also apply your multilabel prediction prevention patch?
this is something I'm unfamiliar with.
who could comment on this?

java/src/main/java/vw/VW.java
marko-asplund: multipredict() updated to conform with predict().

java/src/main/c++/vw_VW.cc
```diff
@@ -76,6 +76,34 @@ JNIEXPORT jfloat JNICALL Java_vw_VW_predict(JNIEnv *env, jobject obj, jstring ex
...
+    r = env->NewFloatArray(vwInstance->lda);
+    float *predictions = ec->topic_predictions.begin;
+    jfloat *weights = new jfloat[vwInstance->lda];
+    for (size_t k = 0; k < vwInstance->lda; k++) {
+      weights[k] = predictions[k];
+    }
+    env->SetFloatArrayRegion(r, 0, vwInstance->lda, weights);
+    delete weights;
+
+    ec->in_use = true; // temporary workaround for issue #800
```
marko-asplund: @jmorra should this workaround be committed as part of this PR or leave it out and wait for a fix separately with issue #800?

JohnLangford:
On naming: multipredict is a little bit ambiguous because predictions have different types in general and we may want to support multiple predictions of different types in the future.  For example, VW also has vectors of integers as base types.  So, instead of multipredict, something like float_vector_predict seems better.  (I believe Java has no way to disambiguate a function call based on return type as for C++.  If that is not true, then you can do even better.)

JohnLangford:
Multilabel vs. multipredict:

multilabel prediction is predicting a set of labels in a multiclass-like setting.  This interface does not support that, because the int and float are different types.  Given this, I think the multilabel check/disable should continue to exist in the Java interface until it is enabled in the Java interface.

java/src/main/c++/vw_VW.cc
jmorra: I'm not sure about this, @JohnLangford would be better to comment on it then me.  I have not seen this issue before, however, if it really is an issue with predict then it should be updated in all the C functions within the JNI interface.

jmorra:
@JohnLangford I'm confused about this.  It would appear that this method is returning an array of floats.  Isn't that exactly what multilabel prediction is doing?  It would seem that that should necessitate the removal of the checks.  I think it makes sense to leave in the test about calling multilabel prediction on the predict Java method as that should fail, however, we should be able to write another test showing the same thing passing with calling multipredict.  Also, if we're adding this shouldn't we also add a corresponding learn method?  Is that necessary?

JohnLangford:
multilabel prediction is for an array of int's.  See here:
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/multilabel.h#L14
and here:
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/example.h#L52

The lda code instead predicts an array of floats here:
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/example.h#L72

The lda predictions should be a part of the polyprediction type, but I haven't merged them yet.

jmorra:
Thanks, in that case I think that if we're going to do this we should just extend it to support all array types.  Might this be solvable with generics?  @deaktator what do you think?

deaktator:
@JohnLangford @jmorra (I can't believe I'm saying this but) should the `float[] multipredict(String example)` interface in JNI be overloaded for @JohnLangford examples?  Do we care about `float[]` vs `int[]` in the return type?  They could all just be `float[]`.  The question is whether it's worth it to give up type safety for interface uniformity.  What do you guys think?

deaktator:
Sorry @jmorra.  You snuck that in there before my response.  I was telling @marko-asplund before that I prefer the use of generic collections rather than arrays anyway because we can enforce immutability.  That might be good.  I also mentioned using our own `java.util.List` implementation that stores the array in unboxed (primitive) form (like in fastutil).  It would have, in addition to the `List` methods, a `get` function to get the raw, unboxed, value.  

If we return 

```java
List<? extends Number> 
```
Then we could let the user cast but that sucks too.  Maybe instead: 

```java
class VW {
  <T extends Number> MulitPredictor<T> getMultiPredictor();
}

interface MulitPredictor<T> {
  public List<T> multiPredict(String example);
}
```

I just don't know if that's too convoluted.

jmorra:
@deaktator I did not expect that response from you.  This doesn't seem to be a case for generics and making use of `instanceof` here is not desirable as it'll be a lot slower.  However, I think the distinction between `int` and `float` is important because `int` would imply classification and `float` would imply regression.  For instance I think it would be better if contextual bandits returned `int` instead of `float`.  The best solution would be to parametrize the class on either `int` or `float`.

jmorra:
@deaktator, that does seem pretty complicated, and the use of `java.util.*` will slow down the interface and one of the main goals of this layer is to keep it as fast as possible.  Maybe the best thing to do is to just have multiple calls now for each type and then interleave them later?  I'm not sure.

deaktator:
@jmorra yup.  Speed trumps.  Agreed.  I would do something that's slight harder to code but very easy to invoke by the end user.  It should take into account

1. **SPEED**
1. Extensibility

Something like:

```java
/**
 * @param example
 * @param SomeVwContext this allows us to access the specific vw instance (maybe just ptr address) 
 */
public interface InternalMultiPredictor<T> {
  T predict(String example, SomeVwContext vwPtr); 
}

public interface VwMultiPredictor<T> {
  T predict(String example);
}

// package private class
LdaMultipredictor implements InternalMultiPredictor<float[]> {
  public float[] predict(String example, SomeVwContext vwPtr) {
    // call native LDA method on the vwPtr. return float[].
  }
}

class VW {
  // Could just be a static constant.
  public static InternalMultiPredictor<float[]> lda() {
    return new LdaMultipredictor();
  } 

  <T> VwMultiPredictor<T> multiPredictor(InternalMultiPredictor<T> instance) 
                          throws IncompatibleModelException {
    // create MultiPredictor<T> from InternalMultiPredictor<T> by 
    // injecting this VW instance's context information.
  }
}
```

Then you create it using:

```java
final VW vw = new VW("...");

// Keep this vwMulti instance around for many examples.
final VwMultiPredictor vwMulti = vw.multiPredictor(VW.lda());
final float[] prediction = vwMulti.predict("example here ...");
```

deaktator:
Or we could just do straight OOP.

```
LDA → VwMultiPredictor<float[]> → VwBase
```

```
VW → VwBase
```

etc.

Then, just do:

```java
final LDA lda = new LDA("params");
```

marko-asplund:
@deaktator 

It's a nice design, but will this it really be worth the added complexity from both the interface developer and user perspective? From user perspective having to obtain `vw.multiPredictor` instance seems a bit unintuitive and like extra work.

I guess we can't really expect to get much reuse across the two `multipredict` functions in JNI interface layer (`vw_VW.cc`). Also, in the Java layer we're only talking about two quite short methods, so the reuse potential is quite minimal.

deaktator:
@marko-asplund I only see one main benefit of jumping through all these hoops: the code that wraps the predict method.  For instance, the current vw predict is:

```java
public float predict(String example) {
    lock.lock();
    try {
        if (isOpen) {
            return predict(example, false, nativePointer);
        }
        throw new IllegalStateException("Already closed.");
    }
    finally {
        lock.unlock();
    }
}
```

All of that boilerplate could be written once in `VwMultiPredictor`, which doesn't need to be an interface, but rather a final generic class that takes an `InternalMultiPredictor`.  The `InternalMultiPredictor` then is extremely simple and just wraps the JNI method.

Another approach is the OOP approach where you have 

```java
abstract class VwMultiPredictor<T> {
  protected abstract T predictHelper(String example);

  public final T predict(String example) {
    // wrapper code
    final T prediction = predictHelper(example);
    // wrapper code
    return prediction;
  }
}
```

The approach to take just depends on the your stance of the [GoF](https://en.wikipedia.org/wiki/Design_Patterns) [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance) argument.  Personally, I like composition.  You can mock it better, etc.  And since composition is your friend, you could do:

```java
public interface InternalMultiPredictor<T> {
  T predict(String example, long nativePointer); 
}

class OptionVwMultiPredictor<T> extends InternalMultiPredictor<Optional<T>> {
  public static <T> InternalMultiPredictor<Optional<T>> 
  around(InternalMultiPredictor<T> p) {
    return new OptionVwMultiPredictor<T>(p);
  }

  private final VwMultiPredictor<T> p;

  public OptionVwMultiPredictor(VwMultiPredictor<T> p) {
    this.p = p;
  }

  public Optional<T> predict(String example, long nativePointer) {
    try {
      return Optional.fromNullable(T);
    }
    catch (Throwable t) {
      return Optional.absent();
    }
  }
}
```

Then you can compose

```java
final VwMultiPredictor<float[]> vwMulti = 
  vw.multiPredictor(OptionVwMultiPredictor.around(VW.lda()));
final Optional<float[]> prediction = vwMulti.predict("example here ...");
```

marko-asplund:
@deaktator I did some prototyping for the multipredict "composition" approach based on the design you proposed. Is this what you had in mind?
https://github.com/marko-asplund/vowpal_wabbit/commit/bd592bab1c02e0d737c186ad17c1c67ff40b93ee

still needs some more work: the code compiles, but I get a `UnsatisfiedLinkError` during `mvn test`.
the code is in `jni-topic-predictions-1` branch.

JohnLangford:
I don't think we should overload types.  Printing tends to be off always and the float type itself cannot encode an int larger than 10^6.

JohnLangford:
In designing an interface, simplicity for the user seems like the strongest goal with simplicity of the interface secondary.  From a user perspective something like: vw.predict<float[]>("example") seems like the best syntax we could hope for.  Is that possible?   The next best seems to be vw.predict_float_vector("example") .

marko-asplund:
@JohnLangford Overloading by method return value type is not possible in Java, so we have to 
- use different method names or
- add some dummy input parameter or
- use separate interfaces

@deaktator, @jmorra: the multipredict "composition" approach prototype PR can be found here:
https://github.com/marko-asplund/vowpal_wabbit/pull/1
The `int[]` multilabel code is still unimplemented, which would result in a few more lines of code.

The "boilerplate" approach prototype PR can be found here:
https://github.com/marko-asplund/vowpal_wabbit/pull/2

I haven't created a prototype for the "inheritance" approach.

Currently, these PRs are against my VW repo fork and meant only to illustrate the different approaches.

Since we're only talking about two methods here, I think the abstraction layer would result in additional complexity both for the API implementation as well as for the user.

I'd vote for the second approach due to its simplicity, but I'm fine with any working solution that gets merged. However, I'd like to be able to finalize the PR soon.

So, which design approach will be chosen?
Any other changes that are needed before this PR can be merged?

deaktator:
@JohnLangford I have an idea for @marko-asplund inheritance approach. Will send in a few minutes.

deaktator:
# Inheritance Approach

@marko-asplund @JohnLangford 

This will be simple and will provide a unified interface.  All VW predictors will implement `VWGeneric<T>`.  Classes that don't want the penalty of boxing will implement additional methods.  Predictor classes are responsible for their own C/C++ files.  There is not a megalithic VW.cc file.

## Calling Code

```java
LDA lda = new LDA("VW command here ...");
final float[] prediction = lda.predict("example here ...");

// AND

VW vw = new VW("VW command here ...");
final float prediction = vw.predictSpecialized("example here ...");
final float boxedPrediction = vw.predict("example here ...");
```

## Implementation Code

```java

```

```java
// VWBase.java

// Responsible for loading vw_jni.lib, lock, vw pointer, closing, etc.
public abstract class VWBase extends Closeable {
    private volatile static boolean loadedNativeLibrary = false;
    private static final Lock STATIC_LOCK = new ReentrantLock();
    private boolean isOpen;
    private final String command;

    protected final Lock lock;
    protected final long nativePointer;
    
    protected VWBase(final String command) { /* ... */ }
    private static void loadNativeLibrary() { /* ... */ }
    protected final boolean isVwOpen() { return isOpen; }
    public void close() { /* .. */ }
    
    private static native long initialize(String command);
    private static native void closeInstance(long nativePointer);
}
```

```java
// VW.java:  Current implementation: scalar return value

public final class VW extends VWGenericBase<Float> {
    public VW(final String command) {
        super(command);
        // Any model validations go here.
    }
    
    // To adhere to VWGeneric
    protected Float predict(String example, boolean learn, long nativePointer) {
        return predict_specialized(example, learn, nativePointer);
    }

    private float learnOrPredict(final String example, final boolean learn) {
        lock.lock();
        try {
            if (isVwOpen()) {
                return predict_specialized(example, learn, nativePointer);
            }
            throw new IllegalStateException("Already closed.");
        }
        finally {
            lock.unlock();
        }
    }

    // No Boxing
    public float learnSpecialized(String example) {
        return learnOrPredict(example, true);
    }

    // No Boxing
    public float predictSpecialized(String example) {
        return learnOrPredict(example, false);
    }

    public static native String version();  // To maintain compatibility
    private static native float predict_specialized(String example, boolean learn, long nativePointer);
}
```

```java
// VWGeneric.java: interface for multi predictions
public interface VWGeneric<T> {
    public T learn(String example);
    public T predict(String example);
}
```

```java
// VWGenericBase.java: abstract base class for multi predictions

public abstract class VWGenericBase<T> extends VWBase implements VWGeneric<T> {
    public VWGeneric(final String command) {
        super(command);
    }

    protected T predict(String example, boolean learn, long nativePointer); 

    private final T learnOrPredict(final String example, final boolean learn) {
        lock.lock();
        try {
            if (isVwOpen()) {
                return predict(example, learn, vw.nativePointer);
            }
            throw new IllegalStateException("Already closed.");
        }
        finally {
            lock.unlock();
        }
    }

    public final T learn(String example) {
        return learnOrPredict(example, true);
    }

    public final T predict(String example) {
        return learnOrPredict(example, false);
    }

    // NO JNI c/c++ code for this class.
}
```

```java
// LDA.java:  Specific LDA impl.  API extender only has to write a class like this.
public final class LDA extends VWGenericBase<float[]> {
    public LDA(String command) {
        super(command);
    }
    protected native float[] predict(String example, boolean learn, long nativePointer);
}
```

If that's not possible, then this:
```java
// LDA.java:  Specific LDA impl.  API extender only has to write a class like this.
public final class LDA extends VWGeneric<float[]> {
    public LDA(String command) {
        super(command);
        // Any model validations go here.
    }
    protected float[] predict(String example, boolean learn, long nativePointer) {
        return lda_multipredict(example, learn, nativePointer);
    }
    private static native float[] lda_multipredict(String example, boolean learn, long nativePointer);
}
```

deaktator:
@marko-asplund I feel your pain about this PR taking a while.  I think it is important to get the design correct so that it is extensible.  @marko-asplund @jmorra @JohnLangford I like the *Inheritance Approach* I suggested above over the composition-based idea.  I think it's the most straightforward for the average Java developer.  What do you guys think?  The unified `predict` and `learn` methods across all non-primitive implementations are nice to have.

jmorra:
@JohnLangford @marko-asplund after looking at @deaktator 's suggestion I think this makes a lot of sense.

1.  It conveys the inherited nature well to the author of the Java code.
2.  Allows the user to easily create examples
3.  Keeps the speed of not boxing any examples if the user desires to do so.

My only comment is I would remove all mention of LDA anywhere in the Java code.  My understanding is that this is for multi-prediction of which LDA is a specific instance.  Since there is no LDA specific code I would just write about LDA in the comments as an example of using this approach and not name anything in the code LDA*.  If those are changed I'm ok with committing this.

deaktator:
I tried this in my fork and c and java build best tests fail because it can't find jni.  The System.load succeeds and its visible in the loaded libraries but any call to a native function gives an UnsatisfiedLinkError. Looking into it.

JohnLangford:
How would this be done?

Logically, the I/O interface of an example depends on two types: the 
type of the input label and the type of the output prediction.  Using 
the type instead of LDA makes sense, but I don't quite get the 'without 
LDA' version of what Ryan is proposing.

-John

marko-asplund:
@deaktator Can you publish your code in a feature branch, so that we can try it out?
I started getting `UnsatisfiedLinkError`s when I had two JNI functions named like this: `Java_vw_VW_multipredict_topics` and `Java_vw_VW_multipredict_multilabel`. Things started working correctly after replacing the last underscore with camel casing. Not sure if this was just caused by `make clean` not cleaning up everything or if the naming actually had an effect.

deaktator:
@marko-asplund **deaktator/vowpal_wabbit** *jni-multi* branch.  You're probably right, I have an underscore in one of the `native` methods.  I'll look at it tomorrow, but here it is for now.

deaktator:
@jmorra I agree with @JohnLangford.  We could add a marker interface:

```java
public interface MultiPredictor extends VWGeneric<float[]>{}
```

And have `LDA` implement that interface as well as extend `VWGenericBase`.

deaktator:
@marko-asplund I'm still getting `java.lang.UnsatisfiedLinkError: no vw_jni in java.library.path`.  I tried changing the names of the native functions to remove underscores.  Didn't work.  I also added the configuration to the surefire plugin:

```xml
  <configuration>
    <redirectTestOutputToFile>true</redirectTestOutputToFile>
    <forkCount>1</forkCount>
    <argLine>-Djava.library.path=${basedir}/target</argLine>
  </configuration>
```

This didn't work.  Then I tried copying `target/vw_jni.lib` to `target/vw_jni.dylib`.  Still didn't work.  I tried using `System.loadLibrary` right after calling `System.load` in the `@BeforeClass` in the unit test.  Nope.

marko-asplund:
@deaktator tried to have a look, but your your branch seems to be missing some files:
https://github.com/deaktator/vowpal_wabbit/tree/jni-multi/java/src/main/java/vw

I'm guessing, this may have occurred due to a bug in `.gitignore`.
the following command should provide a workaround:
```sed -i.tmp 's/^vw$/#vw/' .gitignore```

not sure which files are meant to be ignored with that particular line.

deaktator:
@marko-asplund talked to @jmorra today.  It's an issue in `java/Makefile`.  I am trying to figure out how to fix the makefile.  I almost got it working and all but one `UnsatisfiedLinkError`s went away.

deaktator:
@marko-asplund committed with all but VW.version() function working.

deaktator:
@marko-asplund @jmorra @JohnLangford My latest build now compiles and the unit tests pass.  I had to modify the `java/Makefile` but now it should be sufficiently generic that it will no longer need to change just to add new JNI c/c++ code.  

@marko-asplund do you want to test your LDA test code against this build and see if things are working properly?

jmorra:
Thanks Ryan. Could we rename some classes to what we discussed this
morning? I think a more appropriate naming scheme would make this easier to
understand.

deaktator:
@jmorra sure thing.

deaktator:
@jmorra tell me what you think about it now.

jmorra:
@deaktator this looks nice except I would rename everything called LDA to something more generic like multipredict or something like that.  @JohnLangford do you have an opinion on the naming schema?

deaktator:
@JohnLangford I can only get behind @jmorra's idea if we make the *C* code smarter so that it can select the proper algorithm based on the constructor `command` parameter.  Since the *C* code doesn't currently have any real smarts, it doesn't seem to make sense as written it's currently written.  I like the idea of basing the Java class name on the shape of the return type, and like I said above, I suggest the marker interface approach, unless the *C* code changes.

We could try to make one `native` method that returns an algorithm type (**would need guidance on the *C* code to do this**) and then have an enum on the java side.  A java non-native function would query the *C* side for the type and produce the enum.  The enum would then select the proper algorithm or throw an exception if an algorithm could not be found.

If we can somehow build the smarts into the *C* code, then the suggestion would make sense.

marko-asplund:
@deaktator Yes, I'd like to test the code in your branch when you're done with the naming and other changes. Could you please let me know when you're finished with those, and I'll try it out.

deaktator:
@marko-asplund. I'll let you know when I here what @jmorra and @JohnLangford decide on naming, etc. 

Make sure to check the LDA C code. I copied it from yours, but I don't know if you changed it since.

jmorra:
I still like the naming convention that I suggested, but I'll defer to @JohnLangford.

JohnLangford:
Closing this one because we have a cleaner interface now.

PR #802, opened 2015-09-23
better malformed ex warning

trufanov-nok:
As discussed in #726 (2).
Was:
```
$ echo '1 |1
2 |sda:4.0fdfd:4' | ./vw
...
malformed example !
'|' , space or EOL expected after : "|sda:4.0"
malformed example !
'|' or EOL expected after : "|sda:4.0"
1.000000 1.000000            1            1.0   1.0000   0.0000        1
1.790589 2.581179            2            2.0   2.0000   0.3934        1

finished run
```

Proposed:
```
$ echo '1 |1                                    
2 |sda:4.0fdfd:4' | ./vw
...
malformed example !
'|' , space or EOL expected after : "|sda:4.0"
Example #1: "|sda:4.0fdfd:4"
malformed example !
'|' or EOL expected after : "|sda:4.0"
Example #1: "|sda:4.0fdfd:4"
1.000000 1.000000            1            1.0   1.0000   0.0000        1
1.790589 2.581179            2            2.0   2.0000   0.3934        1

finished run
```

I've refactored code a bit and moved warning code to inline function.
In first example user is getting only "|sda:4.0" which displays proper float weight. User doesn't see the rest of the line. In second example the rest of the line is shown. Unfortunately I can't display whole example line (label header is missing) without more serious code changes, but I can display line number which precisely addresses the example in file.

### DIFF
--- a/vowpalwabbit/parse_example.cc
+++ b/vowpalwabbit/parse_example.cc
@@ -49,6 +49,13 @@ public:
 
   ~TC_parser() { }
 
+  inline void parserWarning(const char* message, char* begin, char* pos, const char* message2)
+  {
+      cout << message << std::string(begin, pos - begin).c_str() << message2
+           << "\nExample #" << this->p->end_parsed_examples << ": \"" << std::string(this->beginLine, this->endLine).c_str() << "\""
+           << endl;
+  }
+
   inline float featureValue() {
     if(*reading_head == ' ' || *reading_head == '\t' || *reading_head == '|' || reading_head == endLine || *reading_head == '\r')
       return 1.;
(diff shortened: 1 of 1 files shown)


arielf:
The additional info (line number and token) looks good.
- Is it possible to get rid of the double print?
- errors like malformed examples are supposed to be rare/unusual so inlining them (for performance?) may actually hurt performance due to more instruction cache space use.

trufanov-nok:
As  for 1st. Let's say it's not trivial. I wouldn't touch them.
As for 2nd. I'm just not aware about "instruction cache space use" problem. I thought inline function is replaced by it's implementation or real (not inline) function call at compilation time depends on compiler decision. You mean it can hurt performance in second case?

JohnLangford:
Perhaps we should change the message to:
cout << message << std::string(begin, pos - begin).c_str() << message2
 << " in Example #" << this->p->end_parsed_examples << ": \"" << std::string(this->beginLine, this->endLine).c_str() << "\""
 << endl;

JohnLangford:
I'll merge it and do it myself.

PR #805, opened 2015-09-25
Updated .travis.yml Closes #804

### LINKED ISSUE #804: .travis.yml allows broken java code to pass
Currently, the `.travis.yml` file has the following bug: 

```bash
cd java; mvn test; cd ..
```

This allows the build to pass even when there is a maven failure because the exit code of `cd ..` is always **0** even when the exit code for `mV test` is **non-zero**.  To correct, change the line to:

```bash
cd java && mvn test && cd ..
```

deaktator:
Updated .travis.yml: `cd java; mvn test; cd ..` -> `cd java && mvn test && cd ..`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
 script:
   - make
   - make python
-  - cd java; mvn test; cd ..
+  - cd java && mvn test && cd ..
   - make test
   - make test_gcov --always-make
   - python setup.py sdist
--- a/java/src/test/java/vw/VWTest.java
+++ b/java/src/test/java/vw/VWTest.java
@@ -12,7 +12,6 @@ import java.util.TreeMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
-import java.util.regex.Pattern;
 import static org.junit.Assert.*;
 
 /**
--- a/java/src/test/java/vw/VWTest.java
+++ b/java/src/test/java/vw/VWTest.java
@@ -12,7 +12,7 @@ import java.util.TreeMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
-
+import java.util.regex.Pattern;
 import static org.junit.Assert.*;
 
 /**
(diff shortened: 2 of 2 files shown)


deaktator:
Hey @JohnLangford. This commit will break the build (I already saw it fail on *Travis*). It is failing because the builds have likely been failing all along.  I only fixed the bug that hid the failures from *Travis*.  

Should this be merged which will report that  the VW build will be broken, or do we wait for the broken code to be fixed and just not report that it is broken?  I would rather merge so we know what's failing rather than not merge and not know what's failing.

Not sure what to do here.  What do you think?

jmorra:
Hi, I would fix this with the following diff

```
Mac:~/git/vowpal_wabbit$git diff
diff --git a/java/src/test/java/vw/VWTest.java b/java/src/test/java/vw/VWTest.java
index 243e278..674f56d 100644
--- a/java/src/test/java/vw/VWTest.java
+++ b/java/src/test/java/vw/VWTest.java
@@ -23,6 +23,16 @@ public class VWTest {
     private final String heightData = "|f height:0.23 weight:0.25 width:0.05";
     private VW houseScorer;
 
+    @BeforeClass
+    public static void globalSetup() throws IOException {
+        try {
+            System.load(new File(".").getCanonicalPath() + "/target/vw_jni.lib");
+        }
+        catch (UnsatisfiedLinkError e) {
+            // Do nothing as this means that the library should be loaded as part of the jar
+        }
+    }
+
     @Rule
     public ExpectedException thrown = ExpectedException.none();
```

I updated the library loading to allow a user to specify the library externally.  I think this is correct and wouldn't change it, however, adding this to the test will allow them to pass in both travis and when using `build.sh`.

deaktator:
@arielf @JohnLangford I fixed, with the help of @jmorra, the unit test that was failing.  Now Travis is working.  Looks good.

arielf:
Nice. Thanks for all the effort!

deaktator:
No problem.  I got your build to work. I had to add @jmorra suggestion for the `BeforeClass` he put in #804 or #805.  Then I fixed the version test like I did for #805.  I didn't look too close at the test code.  Did you deserialize the model and then use it to predict at least one sample?  I think one other thinf we should do is do a -Xmx16m in the surefire params and hammer the model with 1M predictions to look for leaks.

One other thing. In the future, we should be able to reuse the same multi predict interface for other multiclass models.  Do you have any idea how to query the vw model to see what kind of multiclass model it is?

deaktator:
Sorry @arielf. I was juggling too many conversations on my cell phone and I sent you the wrong message. No problem doing the work.

JohnLangford:
Merged in, thanks.

-John

PR #809, opened 2015-10-02
Adding in support for all VW output types

jmorra:
This pull request is the culmination of work from @deaktator and @marko-asplund.  I think that this should encompass everything that everyone wants to do in a generic fashion.  The goal of this pull request is to unify everyone's ideas in a way that allows all VW types to be accessible from the JNI layer.  @marko-asplund thank you for getting the ball rolling on this and adding an initial implementation.  I've cleaned up some of your C++ layer to remove an unnecessary extra copy, and @deaktator I added a bunch more tests to make sure everything does in fact work as expected.  I would relish any and all comments to make sure everyone agrees with what is being done here.

The only thing this is missing is a layer from @JohnLangford that will allows us to determine the expected model output type upon instantiation of the model.  This will allow validation of the correct output type at instantiation and because we have different classes for the different expected return types this cannot be messed up after the model is correctly validated.

Thanks again guys


deaktator:
Looks good to me. Feel bad about screwing @marko-asplund out of his commit / PR #799.  Did you take my test suggestions from #799?

jmorra:
I actually totally agree with you @deaktator.  I don't want to screw @marko-asplund out of his commit.  In fact I tried to do a pull request for this relative to @deaktator's branch and then have him merge into @marko-asplund but I couldn't figure out how to do that.  Ryan, let me know how to do that and I'm happy to go that route to give everyone credit.

deaktator:
If it can wait until Wednesday night.  I can do it. I'll defer that decision to @marko-asplund.  I know he wanted to get this stuff merged.

marko-asplund:
Guys, it's about the content of the commit, not the commit metadata :-)
I'm currently using an old version of the interface extension, so there's no rush from my part.
I'll refactor our application code when the new interface has been merged in master.

jmorra:
If you guys are all ok with what's written then I think the only thing left is to determine if @JohnLangford is going to add the protective functions now or else I'll add a big comment about safety before this is merged.

deaktator:
If @JohnLangford doesn't add, I would suggest adding the unsafe comment to the package-info.java in one of the packages:

vw
vw.learner

jmorra:
@JohnLangford my only concern about this is the amount of duplicated code on the C side.  I'm sure this can be reduced, but I'm not sure how.  Please let me know if you have any ideas.

jmorra:
I've reduced the amount of code considerably making use of C++ lambda expressions.  @JohnLangford this is the first time that I've used lambda expressions and I want to make sure this won't be inefficient.  Could you please have a look at let me know what you think.  I like this a lot better cause the amount of code duplication is greatly reduced.

deaktator:
@jmorra looking at the structure of each function:

1)  initialize return var, R
2) TRY
3)  initialize example
4)  learn / predict
5)  populate R
6)  cleanup example and JNI
7)  CATCH
8)  return R

You could group return var initialization and population into a single function, F, pass that function to the function that sets up and tears down the example and JNI env, G.  G calls F and returns F's return value. G(F, example) might have to go inside another function that casts G(F, example) to the appropriate type.  Sorry. I am not a C guy.  I would judt say void* the hell out of it.

deaktator:
Wow! You wrote that in the time it took me to shower and write my response.  Are templates slower? I've heard, perhaps, incorrectly, they are slower.

jmorra:
Have a look at my latest commit. I figured it how to do this in a type safe
manner. I saw the same thing you did and didn't like it either.

deaktator:
I did look! Looks good to me (but I'm not the C++ guru).

jmorra:
Templates shouldn't be slower, but the way I passed the function may be
hence why I think it's best for John to comment on this solution. It's
certainly better code, but I don't want to sacrifice performance too much

java/src/test/java/vw/learner/VWFloatArrayLearnerTest.java
```diff
@@ -0,0 +1,73 @@
+package vw.learner;
```
jmorra: @marko-asplund I was going through this class and I cleaned up some redundant code in here.  There were two more tests, one that didn't assert anything and one that checked if an exception would be thrown which would have be triggered by the one test left.  I'm not too familiar with how LDA works in VW and I'm not sure exactly what this last test is doing.  Can you please advise.

marko-asplund: Basically `testLdaPredict` trains a model (+persists it), loads the model and then uses the model to get a prediction.
`testLdaPredict` tests all of these tasks. The rationale for having separate tests for each of the tasks was to make troubleshooting easier in case one of them fails.
For example if all of the 3 tests would fail then it should be straightforward to determine that the root cause is `testLdaTrainModel` failing.

JohnLangford:
Today I looked into typifying the output label so it can be queried 
across the library interface. It's harder than I realized.

The only practical way to do it at the moment is to look at the 
finish_example in the top level learner, but there are many more kinds 
of finish_example than there are label types, and they are often private 
to individual reductions.

So, for the moment, I think we need to leave type-unsafe access to 
predictions and rely on an intelligent programmer and I'll try to 
address this in my next refactoring pass.

-John

JohnLangford:
I don't see anything computationally atrocious on the C++ side.  Which lines were you worried about precisely?

deaktator:
Not atrocious sounds, at least to me, like there is room for improvement.  Are you thinking we can do better?

JohnLangford:
Which line(s) precisely?

-John

jmorra:
As far as improvements, my only concern was the speed of passing around the functions.  I read some comments that using std::function was slow, hence I removed it and use a template typename parameter now.  Is this efficient and in keeping with the VW style or is there a better way to do it?

I think passing the functions around is desirable because it reduces the total amount of code considerably.

JohnLangford:
Templates are good and common in VW.

-John

jmorra:
OK, then I'm ok with it, except as @deaktator has suggested to me offline we should annotate the classes to denote that they are not safe.  @deaktator, which annotation is most appropriate for this?

Also this will totally break the current Java API.  I'm ok with this as the interface is still relatively new and I think it's much more stable and extensible like this, but it might be a problem for some people.

JohnLangford:
There is no question that we want the newer interface.   The remaining 
question:Is the added complexity of maintaining the old interface worth 
it?   It's your decision.

-John

jmorra:
I think it might be appropriate to poll the user base and see if anyone is using the old interface yet.  If there is a lot of usage then I can set something up with the old interface, but if there are very few users then I'd like to not maintain both interfaces.

deaktator:
Maybe just leave the old one in and put a deprecated annotataion and add to the readme an expected removal date / version.

JohnLangford:
All, Jon wants to change the Java interface into something cleaner that 
will work with different prediction types which seems desirable going 
forward.  Is anyone strongly dependent on the current java interface?

-John

JohnLangford:
No comment, so this is merged in now.

-John

PR #812, opened 2015-10-03
predictions are saved as int if have no float part

### LINKED ISSUE #801: Suggestion: uint predictions shouldn't be saved as float values
This is a small wish that I would like to save here. Perhaps I'll check if it's easy to implement by myself later. Or if someone is going to rewrite related functionality (I recall some discussions about `--probabilities` mode) it may be possible to address it at the same time.

I'm working with pretty big dataset and it's validation part has 12Mln examples. I'm using `--ect` mode. The predictions file I'm getting is 108Mb. While reference file that I made from same validation part with `cut -f 1-1 -d ' '` is only 24Mb (1 character + '\n' per line). That's because VW saves predictions as floats while example labels are integers. So I'm, getting 
```
$ head -n 3 data5.pred
3.000000
3.000000
3.000000
```
instead of 
```
$ head -n 3 data5.true
3
3
3
```

It's kind of waste of disk space and network traffic. It's not a big problem for me, but may be undesirable if VW is operating over network on several nodes and used for bigger dataflows.

The problem would be possible to easily fix with just overloaded print() function for `int` 5-7 years ago, but nowadays the codebase is so big that it's not so trivial. So, I'd just save my proposal here and turn back to it whenever I'll find a spare time for it.

trufanov-nok:
resolves #801 
Simplest solution. Shouldn't affect performance much.

### DIFF
--- a/vowpalwabbit/global_data.cc
+++ b/vowpalwabbit/global_data.cc
@@ -93,7 +93,10 @@ void print_result(int f, float res, float, v_array<char> tag)
   if (f >= 0)
   {
     char temp[30];
-    sprintf(temp, "%f", res);
+    if (floorf(res) != res)
+       sprintf(temp, "%f", res);
+    else
+       sprintf(temp, "%d", (uint32_t) res);
     std::stringstream ss;
     ss << temp;
     print_tag(ss, tag);


JohnLangford:
This is a nice simple hack, thanks.

The more permanent solution is to have a new print function for uint32_t 
type predictions.

-John

PR #815, opened 2015-10-07
Created the r.vw package combining dt2vw and vw

SelimRaboudi:
This follows up #794 and  #797.

I've condensed vw and dt2vw into an R package, "r.vw", as @zachmayer suggested. Both functions are documented. The package is based on the previous code of @martinbel and mine.
I've also updated the readme file.

to install the package:
```
devtools::install_github("JohnLangford/vowpal_wabbit", subdir = "R/r.vw")
```

regarding vw, I did some slight modifications: it's now possible to directly use a R data.frame in vw(), without having to use dt2vw() beforehand. See examples/vw_example_2.R and ?vw

@zachmayer I haven't updated the code of dt2vw following your remarks, will do :)

It's good to see R/Vowpal Wabbit integration is advancing !


JohnLangford:
My knowledge of R is limited, but this looks good.  I merged it after
waiting a bit for comments.

-John

zachmayer:
Looks like a good start to me too.

PR #816, opened 2015-10-08
hex encoding of namespaces in comand line

### LINKED ISSUE #808: Suggestion: support of hex \x00 format in command line args for namespaces
VW theoretically supports up to 256 different namespaces. On practice it's a bit more limited. But I'm working with about ~160 namespaces now. (I had to adjust `valid_ns_size` constant for that). And I found out that specification of non-printable characters or characters that are not present in latin charset is a bit painful.  
It seems that it can be done in current bash with `vw --redefine $'a:=\xC0'`, but I also have an app that calls VW directly, and this command line isn't preprocessed in bash in this case. It's written in Qt and encapsulates string data in unicode QString objects, sourcefiles are in UTF-8. I had to play with system codepages to workaround this problem. It may be also a problem for scripts like `vw-hypersearch`.  
My suggestion is to declare and implement in VW an escape sequence for  namespace specification in hex format for `-q`, `--cubic`, `--ngram`, `--interactions`, `--redefine` and perhaps a few more command line parameters. It will be better to have terminal\platform\language independent way to encode such namespaces.
It can be `\x00` in this case from bash it can be called as `vw --redefine a:=abc..\\xFE\\xFF`  
The helper function may be like that:
```
string lookup_for_escape_characters(const string& arg)
{ // make a copy of arg replacing \x00 sequences in it
    string res;
    int pos = 0;
    unsigned char c; char* p;
    while (pos < (int)arg.size()-3)
    {
        if (arg[pos] == '\\' && arg[pos+1] == 'x')
          {            
            c = (unsigned char) strtoul(arg.substr(pos+2,2).c_str(), & p, 16);
            res.push_back(c);
            pos += 4;
          } else
            res.push_back(arg[pos++]);
    }

    while (pos < (int)arg.size())
        res.push_back(arg[pos++]);

    return res;
}
```

and usage:
```
before:
          string arg = *arg_iter;
          size_t arg_len = arg.length();
          // let's find operator ':=' position in N:=S
          for (size_t i = 0; i < arg_len; i++)
after:
          string arg = *arg_iter;
          arg = lookup_for_escape_characters(arg); // the change
          size_t arg_len = arg.length();
          // let's find operator ':=' position in N:=S
          for (size_t i = 0; i < arg_len; i++)
```

If it's fine I'll prepare a PR with function like that for all command line parameters that accept namespaces + case for `make test` and perhaps some notes for wiki.

trufanov-nok:
resolves #808 
New platform independent format for namespace encoding is `\xAA` where AA is hex code of char.  
In bash it will be `vw -q \\xAA` as \ must be escaped.  
Note: `x` is case sensitive in `\xAA`. Just to decrease probability of accidental use.  
Change affects following vw params: q, cubic, interactions, ignore, keep, redefine, ngram, skips, lrq, lrqfa, affix, spelling.  
As a bonus --ignore and -keep now supports strings as well as single characters:
WAS: ``vw --ignore a --ignore b --ignore c --ignore d --ignore e``
NOW: ``vw --ignore abcde`` OR ``vw --ignore ab --ignore cd --ignore e``

### DIFF
--- a/vowpalwabbit/global_data.cc
+++ b/vowpalwabbit/global_data.cc
@@ -179,7 +179,7 @@ void compile_gram(vector<string> grams, uint32_t* dest, char* descriptor, bool q
       cout << "You must specify the namespace index before the n" << endl;
     else {
       int n = atoi(ngram.c_str()+1);
-      dest[(uint32_t)ngram[0]] = n;
+      dest[(uint32_t)(unsigned char)*ngram.c_str()] = n;
       if (!quiet)
         cerr << "Generating " << n << "-" << descriptor << " for " << ngram[0] << " namespaces." << endl;
     }
--- a/vowpalwabbit/lrq.cc
+++ b/vowpalwabbit/lrq.cc
@@ -3,6 +3,7 @@
 #include "reductions.h"
 #include "rand48.h"
 #include "vw_exception.h"
+#include "parse_args.h" // for spoof_hex_encoded_namespaces
 
 using namespace LEARNER;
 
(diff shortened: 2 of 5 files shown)


JohnLangford:
Merged in.  This looks good.   Can you add some documentation to the wiki
as well?

-John


On Thu, Oct 8, 2015 at 4:44 PM, Alexander Trufanov <notifications@github.com

trufanov-nok:
>  Can you add some documentation to the wiki as well?  

OK

PR #817, opened 2015-10-09
New CSV/TSV to VW format processing tool.

thejonan:
Few improvements on csv2vw script, packaged into a new one:

- By default outputs sparse matrix, omitting zeroes
- Ability to specify splitting column, so that several datasets are written, based on different values in that column
- Ability to specify tag and label columns
- All columns are specified with name and number (positive and negative)
- Ability to provide regex pattern for excluding certain columns
- Ability to group columns into namespaces, based on provided regex patter

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,6 @@ dist/**
 pyvw.egg-info/**
 python/pyvw.egg-info/**
 
+/COPYING
+/README
+/compile
--- a/README.md
+++ b/README.md
@@ -145,10 +145,10 @@ brew install boost --with-python
 $ port install libtool autoconf automake
 
 ## Build Boost for Mac OS X 10.8 and below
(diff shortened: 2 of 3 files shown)


arielf:
Thanks. These enhancements seem useful.

One request: could you please fix indentation issues?
It seems to mix tabs and spaces (or assume non-universal tab size) in ways that are not universally consistent, both in leading space and in non-leading space comment mis-alignments.

See here:
https://raw.githubusercontent.com/thejonan/vowpal_wabbit/07a4659b8c3620b8bf4346dd8f93dcb99d2f2ad1/utl/vw-format.pl

Thanks.

thejonan:
Yes, I'm sorry for this. It should be fixed now.

JohnLangford:
Merged in, thanks.

-John

Issue #818, opened 2015-10-09
LBFGS with the "--output_feature_regularizer_*" option causes a segmentation fault


tzukuoh:
Running Test 16 with "--output_feature_regularizer_text" (as well as "--output_feature_regularizer_binary") causes a segmentation fault:

tkhuang@msrnyc-03:~/software/vowpal_wabbit/test$ ../vowpalwabbit/vw -k -c -d train-sets/rcv1_small.dat --loss_function=logistic -b 20 --bfgs --mem 7 --passes 20 --termination 0.001 --l2 1.0 --holdout_off --output_feature_regularizer_text r.txt
using l2 regularization = 1
enabling BFGS based optimization \*\*without\*\* curvature calculation
Num weight bits = 20
learning rate = 0.5
initial_t = 0
power_t = 0.5
decay_learning_rate = 1
m = 7
Allocated 72M for weights and mem
\#\# avg. loss    der. mag.       d. m. cond.      wolfe1         wolfe2          mix fraction    curvature       dir. magnitude  step size 
creating cache_file = train-sets/rcv1_small.dat.cache
Reading datafile = train-sets/rcv1_small.dat
num sources = 1
 1 0.69315      0.00266         0.87764                                                         2.24708         776.93237       0.39057   
 3 0.51357      0.00493         0.23981          0.523903       0.088793                                        40.87817        1.00000   
 4 0.48774      0.00232         0.08215          0.274548       -0.410690                                       2.64529         1.00000   
 5 0.47879      0.00006         0.00617          0.595892       0.183063                                        0.47184         1.00000   
 6 0.47750      0.00000         0.00221          0.703360       0.403715                                        0.68626         1.00000   
 7 0.47680      0.00000         0.00038          0.588395       0.175459                                        0.08911         1.00000   
 8 0.47671      0.00000         0.00002          0.568445       0.136827  
Termination condition reached in pass 8: decrease in loss less than 0.100%.
If you want to optimize further, decrease termination threshold.
                                0.00444         1.00000   
Segmentation fault

tzukuoh:
I created an end_examples function, which contains the only call to preconditioner_to_regularizer() (see https://github.com/tzukuoh/vowpal_wabbit/blob/master/vowpalwabbit/bfgs.cc)
But I still get a seg fault. Here is what gdb said:

tzukuoh:
[...]

PR #819, opened 2015-10-12
fix segmentation fault for regularizer output

eisber:
Proposed fix for #818


JohnLangford:
This doesn't seem to be the right fix.  We need this function to be called once after all examples have been processed.  Using the end_examples() hook seems like the right approach.

PR #820, opened 2015-10-13
Upgrade to VW Java interface to Java 8

marko-asplund:
Java 6 has reached it's end-of-life in February 2013. Upgrade to Java 8.
http://www.oracle.com/technetwork/java/eol-135779.html

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -97,8 +97,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.2</version>
                 <configuration>
-                    <source>1.6</source>
-                    <target>1.6</target>
+                    <source>1.8</source>
+                    <target>1.8</target>
                 </configuration>
             </plugin>
             <plugin>


marko-asplund:
@jmorra Regarding our previous discussion on upgrading Java, what would be a suitable schedule for merging this PR?

jmorra:
As I stated before we cannot do this as it would cause our processes to break terribly.  As it stands it will work fine with all Java 8 code.  I'm confused why this is required.  Once we upgrade our systems here, we can move forward with this, but doing this now would cause a huge headache on our side.

marko-asplund:
Sure, I understand and do remember our previous discussion regarding this issue.
That's the reason I created this PR as a separate issue, so that you can merge it after you've upgraded Java on your systems.

JohnLangford:
The build is breaking on travis.

Is there any advantage to upgrading?  Usually, compilers try to be backwards compatible.

jmorra:
I think having it ready to build with Java 8 is a good goal.  Building with
Java 8 on travis is also desirable.  This fix will involve edits to the
travis.yml file.

JohnLangford:
Ping on this?   The patch seems mergeable as long as backwards compatibility is maintained.

jmorra:
This would cause a real headache for me as we're just moving to Java 7 now.  Moving to Java 8 would require a lot of issues on my side.  It doesn't get us anything on the VW side so I would recommend not doing this now.

JohnLangford:
Without backwards compatibility, this seems like a no-go for now.

PR #821, opened 2015-10-13
Fix Java interface build failure due to doclint error

marko-asplund:
Fix Java interface build failure due to doclint error (with Java 8),
replace IDE generated author attribution with Javadoc compatible one.

Issue #807, opened 2015-10-01
[vw-jni] more robust OS detection


petro-rudenko:
Hi, i'm using linux mint (ubuntu derivative). Cannot run vw-jni program, because ```lsb-release -r``` shows linux mint version. Maybe distinguish binary files by glibc version, they was compiled, or use /etc/os-release where it actually points to ubuntu:
```
NAME="Ubuntu"
VERSION="14.04.3 LTS, Trusty Tahr"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 14.04.3 LTS"
VERSION_ID="14.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
```

JohnLangford:
os-release sounds like it might be good.

-John

marko-asplund:
This is related to PR #780.

JohnLangford:
I believe this is fixed.

PR #822, opened 2015-10-13
Implement Java interface native OS library name generation using Maven

marko-asplund:
This PR aims to resolve issues reported in #780 and #807 by implementing native OS library name generation during build-time using Maven.

@jmorra and @petro-rudenko: could you please review and test.

This PR depends on PR #821.

Also, created an issue for `os-maven-plugin` to provide additional OS info: https://github.com/trustin/os-maven-plugin/issues/15

### DIFF
--- a/java/src/main/java/vw/jni/NativeUtils.java
+++ b/java/src/main/java/vw/jni/NativeUtils.java
@@ -49,7 +49,7 @@ public class NativeUtils {
             Process process = Runtime.getRuntime().exec("lsb_release -r");
             reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
             String line;
-            Pattern releasePattern = Pattern.compile("Release:\\s*(\\d+).*");
+            Pattern releasePattern = Pattern.compile("Release:\\s*(\\S+).*");
             Matcher matcher;
             while ((line = reader.readLine()) != null) {
                 matcher = releasePattern.matcher(line);
@@ -76,7 +76,7 @@ public class NativeUtils {
     public static String getOsFamily() throws IOException {
         final String osName = System.getProperty("os.name");
         if (osName.toLowerCase().contains("mac")) {
-            return "Darwin";
+            return "darwin";
         }
         else if (osName.toLowerCase().contains("linux")) {
             String distro = getDistroName();
@@ -88,7 +88,7 @@ public class NativeUtils {
             if (version == null) {
                 throw new UnsupportedOperationException("Cannot determine linux version");
             }
-            return distro.trim().replaceAll(" ", "_") + "." + version;
+            return (distro.trim().replaceAll(" ", "_") + "_" + version).toLowerCase();
         }
         throw new IllegalStateException("Unsupported operating system " + osName);
     }
@@ -103,7 +103,7 @@ public class NativeUtils {
      */
     public static void loadOSDependentLibrary(String path, String suffix) throws IOException {
         String osFamily = getOsFamily();
-        String osDependentLib = path + "." + osFamily + "." + System.getProperty("os.arch") + suffix;
+        String osDependentLib = path + "-" + osFamily + "_" + System.getProperty("os.arch") + suffix;
         if (NativeUtils.class.getResource(osDependentLib) != null) {
             loadLibraryFromJar(osDependentLib);
         }
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -75,6 +75,15 @@
                 </includes>
             </testResource>
         </testResources>
+
+        <extensions>
+          <extension>
+            <groupId>kr.motd.maven</groupId>
+            <artifactId>os-maven-plugin</artifactId>
+            <version>1.4.0.Final</version>
+          </extension>
+        </extensions>
+        
         <plugins>
             <plugin>
                 <artifactId>maven-clean-plugin</artifactId>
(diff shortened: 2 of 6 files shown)


jmorra:
This doesn't build properly on a mac.  When I run I see the following snippets

```
[INFO] ------------------------------------------------------------------------
[INFO] Detecting the operating system and CPU architecture
[INFO] ------------------------------------------------------------------------
[INFO] os.detected.name: osx
[INFO] os.detected.arch: x86_64
[INFO] os.detected.classifier: osx-x86_64
[INFO]                                         
```

and

```
[INFO] --- copy-rename-maven-plugin:1.0:copy (copy-file) @ vw-jni ---
[INFO] Copied /Users/jmorra/git/vowpal_wabbit/java/target/vw_jni.lib to /Users/jmorra/git/vowpal_wabbit/java/target/classes/vw_jni-${os.detected.release}_${os.detected.release.version}_x86_64.lib
```

The file created on my machine is called `vw_jni-${os.detected.release}_${os.detected.release.version}_x86_64.lib	vw_jni-osx_${os.detected.release.version}_x86_64.lib` which isn't correct.

java/src/main/java/vw/jni/NativeUtils.java
```diff
@@ -49,7 +49,7 @@ public static String getLinuxVersion() throws IOException {
             Process process = Runtime.getRuntime().exec("lsb_release -r");
             reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
             String line;
-            Pattern releasePattern = Pattern.compile("Release:\\s*(\\d+).*");
+            Pattern releasePattern = Pattern.compile("Release:\\s*(\\S+).*");
```
jmorra: Why this change?  This regex is slower.

java/src/main/java/vw/jni/NativeUtils.java
```diff
@@ -88,7 +88,7 @@ else if (osName.toLowerCase().contains("linux")) {
             if (version == null) {
                 throw new UnsupportedOperationException("Cannot determine linux version");
             }
-            return distro.trim().replaceAll(" ", "_") + "." + version;
+            return (distro.trim().replaceAll(" ", "_") + "_" + version).toLowerCase();
```
jmorra: If you change the pattern here you have to change `build.sh` as well.  If you don't update this file at all then this change can be merged in without affecting `build.sh`.  I had intended the `.` to separate out different elements of the build environment in the name and `_` to denote word separation.  For example, in my opinion `vw_jni.Red_Hat.6.amd64.lib` is more legible than `vw_jni.Red_Hat_6_amd64.lib`.  As far as casing, this is designed to match the system calls to uname.  If you look at line 171 of `build.sh` you can see I specifically don't input `Darwin` but instead get the property directly.  I think this is more robust as I'm just reusing the system properties.

java/src/main/java/vw/jni/NativeUtils.java
```diff
@@ -103,7 +103,7 @@ else if (osName.toLowerCase().contains("linux")) {
      */
     public static void loadOSDependentLibrary(String path, String suffix) throws IOException {
         String osFamily = getOsFamily();
-        String osDependentLib = path + "." + osFamily + "." + System.getProperty("os.arch") + suffix;
```
jmorra: See above.

java/src/main/java/vw/jni/NativeUtils.java
marko-asplund: The original regex parses OS release major version, but unfortunately `os-maven-plugin` only supports getting the full version.

jmorra: This is problematic then because if we go with this we'll have to create a separate build for every minor release as well and the number of libraries packaged in the jar will explode.

java/src/main/java/vw/jni/NativeUtils.java
marko-asplund: The rationale behind proposing a naming scheme change was that the syntax would allow easier parsing of different name components. Consider the following example based on the original scheme:
`vw_jni.ubuntu.13.10.amd64.lib`
arguably this would be somewhat trickier to parse than 
`vw_jni-ubuntu_13.10_amd64.lib`

So, that was the rationale, but I don't have a problem with using the original scheme either, if you think that it would be better.

Also, `os-maven-plugin` lowercases strings, so I think we should follow that convention.

jmorra: Lowercasing should be fine, but we have to update `build.sh` then.  It should be straightforward to do, just update lines 19, 24, 29, 49, 56, 61, and 171.  The last line might be a little tricky.  I really think adding the minor version to the name is problematic.

marko-asplund: Understand your concern about major/minor OS version. 
Using the `os-maven-plugin` based approach, one way around this would be to implement  https://github.com/trustin/os-maven-plugin/issues/15.
Another option would be to list all the native libraries included in the archive at runtime and then automatically pick the best match.
Not sure which would be the better approach.
With Linux, binary compatibility it's not so much about the distribution and version as much as it is about glibc and kernel version. In that respect, the latter approach could arguably be more robust.

JohnLangford:
It sounds like build.sh needs to be fixed here.   

I'm still sketchy on why the change of naming is worthwhile?  What is the scenario which calls for this?

marko-asplund:
Here's some prototype code that would allow listing available native library files:
https://gist.github.com/marko-asplund/58f91f386c363e89fd3d

`NativeUtils.loadOSDependentLibrary` could use something like this to iterate over available library files and pick the most suitable one. I think this could be one way of getting over the major/minor issue.

However, I think determining binary compatibility is a bit problematic. OS type and CPU architecture most likely need to match, but e.g. different Linux variants can still be compatible. Should the OS release and version be included in the native library name at all?

@jmorra what do you think?

JohnLangford:
@jmorra  can you take a look?

My understanding is that the core problem is linux likes to name itself in many ways, so using the right flavor of binary is tricky.

jmorra:
That's exactly right. Not only that, but this doesn't work on Mac. I would
hold off on this until the referenced issue is resolved with the maven
plugin.

marko-asplund:
Yes, this still requires some more work, so it's not ready to be merged.
I created PR https://github.com/trustin/os-maven-plugin/pull/16 for `os-maven-plugin` to address some requirements of this PR, but that hasn't been processed yet.

JohnLangford:
I'm going to close this for now.  Please reopen when you fix the issues.

PR #823, opened 2015-10-13
Getting build.sh working

jmorra:
It would appear that build.sh is broken.  The Centos 5 build was failing because the cluster `Makefile` now requires the boost library.  This fix will supply the boost library when linking the cluster build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,7 @@ library/library_example
 library/recommend
 library/gd_mf_weights
 library/test_search
+library/search_generate
 test/RunTests.last.times
 *.dSYM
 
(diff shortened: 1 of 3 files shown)


jmorra:
I just verified that this change does in fact make `build.sh` work again.

.gitignore
```diff
@@ -52,6 +52,7 @@ library/library_example
 library/recommend
 library/gd_mf_weights
 library/test_search
+library/search_generate
```
jmorra: I'm not sure about this, but it would appear that search_generate is a new generated file.  @JohnLangford please correct me if I'm wrong.

JohnLangford: Right, the source is in search_generate.cc.

PR #824, opened 2015-10-14
Assert non empty LDA predictions vector

marko-asplund:
Demonstrate bug in the new JNI interface (PR #809).


jmorra:
Should this be closed and combined with #828?

marko-asplund:
merged in PR #828.

PR #825, opened 2015-10-14
Fixing Java build, build.sh, and adding Java test for bad model

jmorra:
This pull request should hopefully finalize the Java build process for the next release.  This will also add back in an ignored test because VW no longer segfaults on a bad model.


jmorra:
@deaktator, you might be interested in the change in VWFloatLearnerTest.java

PR #826, opened 2015-10-15
Fix recent broken build also seen in OS-X

arielf:
Turned out to be a leading space introduced by C++ FP formatting of the 1st number in a line

`RunTests` now works-around this by stripping leading white space in lenient-float compares.

### DIFF
--- a/test/RunTests
+++ b/test/RunTests
@@ -41,6 +41,8 @@ my $ErrorCount = 0;
 # These --side-by-side diff opts are used to make the
 # fuzzy-compare easier: just split on '|' and compare numeric values
 # word by word:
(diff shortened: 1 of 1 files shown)


arielf:
Also: this now passes test 16 with either `-march=native` or not.

JohnLangford:
Great, thanks Ariel.  Getting macs to pass tests is a nice bonus.

-John

arielf:
NP, sorry it took me so long to figure this out.  Once you discovered this can also be triggered by `-march=native` and I could reproduce it (with gcc only) on every try, I was able to figure out the root cause.

JohnLangford:
Yeah, reproducibility is the key to solution.

-John

Issue #827, opened 2015-10-15
New Java interface returns empty LDA predictions


marko-asplund:
It seems that the new Java interface `VWFloatArrayLearner. predict` method, introduced in PR #809, always returns an empty predictions vector.

PR #824 aims to add a test that demonstrates this problem.

@jmorra @deaktator Is this just a problem in my environment or does this also occur in yours?

marko-asplund:
Seems to be fixed with commit 70aa3ebf0ef83cb3b6bb0c65130a85e9ef8729e7 by @JohnLangford.

PR #828, opened 2015-10-15
Validate LDA predictions vector length and make test docs more readable

marko-asplund:
Propose fix for #827.

### DIFF
--- a/java/src/test/java/vw/learner/VWFloatArrayLearnerTest.java
+++ b/java/src/test/java/vw/learner/VWFloatArrayLearnerTest.java
@@ -8,11 +8,13 @@ import org.junit.rules.TemporaryFolder;
 import vw.VWTestHelper;
 
 import java.io.IOException;
+import java.util.*;
 
-import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.*;
 
 /**
- * Created by jmorra on 10/2/15.
+ * @author jmorra
+ * @author marko asplund
  */
 public class VWFloatArrayLearnerTest {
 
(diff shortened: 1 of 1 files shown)


jmorra:
This is good that you found this.  I agree that is has to work, I am just really interested in keeping the mentioning of LDA in the JNI code to a minimum as I want this to continue to work when other learner types that require a float array to be returned are introduced.  @JohnLangford is there another way to do this without referencing LDA by name?

jmorra:
I think you should merge #824 in here as well into one PR.

java/src/main/c++/vw_learner_VWFloatArrayLearner.cc
```diff
@@ -3,9 +3,9 @@
 #include "jni_base_learner.h"
 #include "vw_learner_VWFloatArrayLearner.h"
 
-jfloatArray floatArrayPredictor(example* vec, JNIEnv *env) {
+jfloatArray floatArrayPredictor(example* vec, JNIEnv *env, jlong vwPtr) {
     v_array<float> predictions = vec->topic_predictions;
-    size_t num_predictions = predictions.size();
+    size_t num_predictions = ((vw*)vwPtr)->lda;
```
jmorra: Since we already have done this cast in jni_base_learner.h it seems silly to do it here too.  Let's have predictor accept the vwInstance directly instead of the pointer.  That would require minor changes to all the files, but it should be better cause you don't do this cast twice.

marko-asplund:
@jmorra updated this PR according to your comments.

Not sure how many people are using VW as an embedded library, but I imagine to them ´vw->lda` member is part of the API and renaming it would result in breaking existing code.

java/src/test/java/vw/learner/VWFloatArrayLearnerTest.java
```diff
@@ -33,31 +35,70 @@ public void setupFiles() throws IOException {
...
 
     @Test
-    public void testLdaPredict() {
+    public void testFloatArrayLearnerPredict() {
         writeVwModelToDisk();
         VWFloatArrayLearner v = rehydrateModel();
-        float[] vector = v.predict("| 1:1 2:2 3:3");
+        float[] vector = v.predict(convertQuery("| wondering we look since"));
         assertNotNull(vector);
+        assertTrue("predictions vector length == 3", vector.length == 3);
```
jmorra: Can we change this to `assertEquals(3, vector.length)` please.

java/src/main/c++/jni_base_learner.h
```diff
@@ -31,7 +31,7 @@ T base_predict(
         else
             vwInstance->l->predict(*vec);
 
-        result = predictor(vec, env);
+        result = predictor(vec, env, (vw*)vwPtr);
```
jmorra: We've already done this, the variable is called `vwInstance`.  There is no need to cast vwPtr again, just pass `vwInstance` directly.  The line should read

```
result = predictor(vec, env, vwInstance);
```

java/src/main/c++/vw_learner_VWFloatArrayLearner.cc
```diff
@@ -3,9 +3,9 @@
 #include "jni_base_learner.h"
 #include "vw_learner_VWFloatArrayLearner.h"
 
-jfloatArray floatArrayPredictor(example* vec, JNIEnv *env) {
+jfloatArray floatArrayPredictor(example* vec, JNIEnv *env, vw *vwPtr) {
```
jmorra: Since we're calling the variable `vwInstance` in `jni_base_learner.h` can we rename the variable in all of these functions to be `vwInstance`.  Consistency makes this easier to follow.

marko-asplund:
thanks. PR updated.

jmorra:
I think this looks reasonable now except for my concerns about addressing the lda variable directly.  @JohnLangford, any ideas on this?

JohnLangford:
The topic predictions vector was not being sized correctly in lda_core.cc due to a nonstandard use.  This is hopefully fixed---please check.

In general though, some dependence on the lda code is inevitable because that is the only thing emitting a float vector and the type has not been properly generalized.   That seems ok here because this is not in the public interface, but rather in the interface layer.  When the type structure changes, we can change the interface layer without altering the interface.

JohnLangford:
(Marko, please check to see if the length is set properly now.)

jmorra:
This seems to work for me.  @marko-asplund please try rolling back all your changes outside of the test directory.  If all the tests still pass then I think this is preferable.  Your tests should prove that @JohnLangford's change is valid.  Does that make sense?

marko-asplund:
thanks! PR updated and title changed to reflect the new scope.

PR #830, opened 2015-10-16
Formatting files with editorconfig & astyle

lhoang29:
Let's use Horstmann bracket style for now


lhoang29:
Just a heads up, this PR will create quite a few merge conflicts for others

JohnLangford:
Two issues:

1) eclint does not seem to be very standard in unix land.  
jl@humpty:~/programs/vowpal_wabbit/vowpalwabbit$ eclint
eclint: command not found
jl@humpty:~/programs/vowpal_wabbit/vowpalwabbit$ sudo apt-get install eclint
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package eclint

I'm not sure what to do about this, but it almost certainly means I can't have it in new_version.

2) Using --keep-one-line-blocks seems merited.

lhoang29:
Actually eclint only runs on Linux I believe. To install it you have to first install npm:

sudo apt-get install npm
sudo npm install -g eclint

Can you try that? I ran eclint on a ubuntu VM to generate the formatting.

JohnLangford:
Seems iffy?  After running the commands I get:

  8:49PM humpty-31: eclint
/usr/bin/env: node: No such file or directory

-John

lhoang29:
Oh, how about this as well:

sudo apt-get install nodejs

JohnLangford:
Claims to already be installed, no change in behavior.

-John

lhoang29:
Ok, looks like you got similar problems to me before. Can you try:

sudo apt-get purge nodejs npm
curl -sL https://deb.nodesource.com/setup | sudo bash -
sudo apt-get install -y nodejs

You may need to rerun these commands again:

sudo apt-get install npm
sudo npm install -g eclint

JohnLangford:
Looks like this works now.

-John

lhoang29:
Cool. I'll update new_version to keep one line blocks

JohnLangford:
I'm going to handle this on my own.   Amongst other things I verified that these operations _are_ mutually idempotent, which is great.  Thanks much for figuring this out.

JohnLangford:
Louie seems to have plausibly solved our remaining formatting problems.

These two lines:

find -regextype posix-extended -regex '.*\.(cc|h|cpp)' | xargs eclint fix
find -regextype posix-extended -regex '.*\.(cc|h|cpp)' | xargs astyle -s2
--style=horstmann --lineend=linux --keep-one-line-blocks
--keep-one-line-statements

appear safe, idempotent, canonicalizing, and plausibly produce a reasonable
format.  I've updated the codebase with them.

eclint is not terribly standard.  Louie found the following to install it
on Linux:

sudo apt-get purge nodejs npm
curl -sL https://deb.nodesource.com/setup | sudo bash -
sudo apt-get install -y nodejs
sudo apt-get install npm
sudo npm install -g eclint

I don't plan to strongly enforce formatting, but it is built into the
release process now and I would encourage people to apply the above if
their editor ends up doing some form of undesired reformatting.

-John

JohnLangford:
Thanks to Markus for pointers to eclint as well.

martinpopel:
> /usr/bin/env: node: No such file or directory

I solved this by `sudo apt-get install nodejs-legacy`.
I also followed https://github.com/sindresorhus/guides/blob/master/npm-global-without-sudo.md because I have limited sudo access on my work machine (I can run `sudo apt-get`, but not `sudo bash -`).

PR #836, opened 2015-10-22
Fix RunTests

martinpopel:
The last line of RunTests was ignored, even if it ended with a newline,
because eof(DATA) is true in that case.
If the last line contained e.g.
  train-sets/ref/active_cover_oracular.stderr
this stderr was not tested.
In this case, the file was missing, so the poor guy,
who would try to add a new test and got an error.

### DIFF
--- a/test/RunTests
+++ b/test/RunTests
@@ -274,28 +274,24 @@ sub ref_file($) {
     $file;
 }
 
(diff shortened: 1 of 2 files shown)


JohnLangford:
Good catch, and thanks.

-John

PR #837, opened 2015-10-23
--probabilities (second attempt)

### LINKED ISSUE #566: multi class oaa progressive loss

### LINKED ISSUE #724: Trying to recover probabilities from command line  in daemon mode not working.

martinpopel:
This is an alternative to PR #774 (taking into account @JohnLangford 's suggestion to rather recompute the predicted label on the fly, instead of storing it in the example). I've decided to create a new PR, so I can rebase it on the current master. Original description:

This new parameter --probabilities can be used in combination with --loss_function=logistic and either --oaa or --csoaa_ldf=mc. VW predictions (-p) will be directly class probabilities (summing to one). There is no need for specifying any --link option.

This PR fixes #566 and it should also fix #724.
See #566 for a detailed discussion.


vowpalwabbit/oaa.cc
```diff
@@ -85,11 +88,94 @@ void predict_or_learn(oaa& o, LEARNER::base_learner& base, example& ec)
...
+{ free(o.pred);
+  free(o.subsample_order);
+  free(o.probs);
+}
+
+// TODO: partial code duplication with multiclass.cc:finish_example
+void finish_example_probabilities(vw& all, oaa& o, example& ec)
+{ // === Compute multiclass_log_loss
+  // TODO:
+  // What to do if the correct label is unknown, i.e. (uint32_t)-1?
```
JohnLangford: If the correct label is unknown, it seems incorrect to compute any loss.

vowpalwabbit/csoaa.cc
```diff
@@ -470,6 +470,23 @@ void do_actual_learning(ldf& data, base_learner& base)
   if (start_K > 0)
     for (size_t k=1; k<K; k++)
       LabelDict::del_example_namespaces_from_example(*data.ec_seq[k], *data.ec_seq[0], (data.all->audit || data.all->hash_inv));
+
```
JohnLangford: I'm confused about the application to csoaa.cc.   Why does this make sense?  Predicting the expected cost given features does not have a probability semantics in general.  Why not just go through the oaa.cc code path?  This seems like to doing the same thing in two places.

vowpalwabbit/parse_args.cc
```diff
@@ -1003,6 +1006,13 @@ void parse_reductions(vw& all)
   all.reduction_stack.push_back(bs_setup);
 
   all.l = setup_base(all);
+
+  if (all.probabilities)
+  { if (!all.vm.count("oaa") && !all.vm.count("csoaa_ldf"))
```
JohnLangford: How can this happen?

vowpalwabbit/parse_args.cc
```diff
@@ -1003,6 +1006,13 @@ void parse_reductions(vw& all)
...
+      THROW("--probabilities can only be used with --oaa=N or --csoaa_ldf=mc");
+    if (!all.vm.count("loss_function") || all.vm["loss_function"].as<string>() != "logistic" )
```
JohnLangford: Shouldn't this test be in oaa.cc?

vowpalwabbit/csoaa.cc
martinpopel: I need ldf, so I cannot use `--oaa` (yes, VW is missing `--oaa_ldf`, but that's a minor problem). In `parse_args.cc` I check that the command line is `--csoaa_ldf=mc --loss_function=logisitic`, so the probabilities are well defined. See also my last comment in #566 .

vowpalwabbit/parse_args.cc
martinpopel: Simply: someone uses `--probabilities` without `--oaa` or `--csoaa_ldf`. The only other reduction I am aware of, where `--probabilities` could be used is binary classification, but here we already can use `--link=logistic`, which has the same effect of reporting probabilities. Nevertheless, so far `--probabilities` are implemented only for oaa and csoaa_ldf, so I think it is better warn users if they try to use it elsewhere.

vowpalwabbit/parse_args.cc
martinpopel: I would need to repeat the test both in oaa.cc and csoaa.cc. Should I do it?
I felt the test (just two lines of code) fits here where I am testing `all.probabilities && (!oaa || !csoaa_ldf)` (and this test obviously cannot be moved to oaa.cc/csoaa.cc).

vowpalwabbit/csoaa.cc
JohnLangford: Ok.  I smell a bad factoring in our code, but that's beyond what you are 
trying to solve.

-John

vowpalwabbit/parse_args.cc
JohnLangford: I would prefer the definition of --probabilities in oaa.cc and csoaa.cc 
and doing the checks locally.  Modularity at a small cost in extra code 
seems worthwhile.  If you can remove the probabilities flag from the 
global datastructure and just rely on the local datastructure for each, 
that seems good to.

-John

vowpalwabbit/example.h
```diff
@@ -52,6 +52,8 @@ typedef union
 { float scalar;
   uint32_t multiclass;
   MULTILABEL::labels multilabels;
+  float* probs; // for --probabilities --oaa
```
JohnLangford: I don't see an appropriate free().   Is there a memory leak?

vowpalwabbit/oaa.cc
martinpopel: OK, but what should I do?
If I initialize `correct_class_prob = 1` (instead of 0, as it is now), I will get `multiclass_log_loss = 0` and thus effectively, I won't update `all.sd->multiclass_log_loss` nor `all.sd->holdout_multiclass_log_loss`. Should I change it this way?
I am still not sure, if I should increase `all.sd->holdout_multiclass_log_loss`.

vowpalwabbit/parse_args.cc
martinpopel: As I wrote in #774: I can move the `--probabilities` option to `oaa.cc` and `csoaa.cc` (i.e. duplicate its definition), but probably I will still need to store it inside `all` (similarly, e.q. `csoaa_rank` aka `multilabel_prediction` is stored inside `all` although it is valid only for some reductions). The problem would be that `vw -h` would not list the `--probabilities` options (see #665).
So should I really move the option `--probabilities` to oaa.cc and csoaa.cc?

vowpalwabbit/oaa.cc
JohnLangford: Why is it important to have a separate loss field?  Why not just use the 
existing loss field?  Then the existing accounting mechanisms should 
work fine.

-John

vowpalwabbit/parse_args.cc
JohnLangford: Yes, more modular is preferred.  It may be impossible to remove 
all.probabilities entirely, but you should not need to do more than 
assign to it in (cs)oaa.  Instead, use a local version of the variable.  
This will make things easier when I get some time to refactor the code 
to remove all.probabilities later and deal with other label type issues 
later.

-John

vowpalwabbit/oaa.cc
martinpopel: In the final resume, I want to report both zero-one loss and multiclass log loss. So I need a separate field to store both.

BTW: I tried to understand the existing accounting mechanism, but it was quite difficult as it is scattered over several files. One of the purposes of this TODO note was to check also the *existing* mechanism for possible bugs (e.g. with regards to `all.sd->weighted_unlabeled_examples` handling and unknown labels).

JohnLangford: I think we need a more general 'random variable' object to store 
statistics like this.

I think the best thing that you can do is not record any loss when there 
is no label.  Ideally, you also record the denominator so that later on 
you can observe the 0/0 situation and correctly handle it.

-John

martinpopel: I would like to implement handling of unknown labels in `--probabilities` consistently with the rest of VW, but this is surrently impossible because VW is inconsistent in this aspect -- see #847.

vowpalwabbit/example.h
martinpopel: This is just a pointer. In oaa.cc, there is both `data.probs = calloc_or_throw<float>(data.k)` and `data.probs = calloc_or_throw<float>(data.k)`. So I hope there are no memory leaks.

Why not using `v_array`? I don't need dynamically growing array. The size is fixed to k (number of classes).

Why not allocating and freeing `probs` within each example? First, I don't see a use case for this (where I would need to access `probs` of all the `ring_size` examples **simultaneously**). Second, I think allocating new memory for each example is slower than reusing one array for all the examples.

JohnLangford: If things are to work under reduction (we want them to!), you need to have
allocated memory associated with the example rather than the reduction.
Otherwise, a reduction which calls oaa twice will get corrupted results.

-John

martinpopel: OK, I'll change it.

martinpopel:
`ec.pred.probs` are now allocated within each example. `all.probabilities` are now moved to oaa.cc and csoaa.cc. As for the loss of examples without labels, I am waiting for #847 to be resolved (and I would say that's a low-priority issue). Is there anything else I should fix?

vowpalwabbit/oaa.cc
JohnLangford: Seems fair.

JohnLangford:
This is merged in now.  Thanks, and thanks for your persistence.

-John

JohnLangford:
Huh, so I thought there was a memory leak, but it turns out that _perl_ has
a memory leak.  I tweaked the test to remove perl.

-John

martinpopel:
> Huh, so I thought there was a memory leak, but it turns out that _perl_ has a memory leak.

Out of curiosity: How have you detected the memory leak?

(Memory leaks in Perl are common when using cyclic references without weaken, but this is surely not the case of `perl -pE 's/^0/2/;s/^-1/4/'`. Even so, the leaked memory should be freed after perl exists. I would be very surprised if such simple perl command resulted in memory leaks. Have you tried sed instead of perl?)

> I tweaked the test to remove perl.

For reference: it was done in 5fe5cbfacafb44655b1c1ee83eafa25a4b6db616
I have nothing against this change (my original motivation was to prevent adding a new test file to the repository), but I think VW reading from stdin should be also tested somewhere.

JohnLangford:
On 10/30/2015 11:21 PM, Martin Popel wrote:
>
>     Huh, so I thought there was a memory leak, but it turns out that
>     /perl/ has a memory leak.
>
> Out of curiosity: How have you detected the memory leak?
>

By prepending valgrind to the command.

> (Memory leaks in Perl are common when using cyclic references without 
> weaken, but this is surely not the case of |perl -pE 
> 's/^0/2/;s/^-1/4/'|. Even so, the leaked memory should be freed after 
> perl exists. I would be very surprised if such simple perl command 
> resulted in memory leaks. Have you tried sed instead of perl?)
>
>     I tweaked the test to remove perl.
>
> For reference: it was done in 5fe5cbf 
> <https://github.com/JohnLangford/vowpal_wabbit/commit/5fe5cbfacafb44655b1c1ee83eafa25a4b6db616>
> I have nothing against this change (my original motivation was to 
> prevent adding a new test file to the repository), but I think VW 
> reading from stdin should be also tested somewhere.
>

That's a good idea.  Maybe add a test that uses echo?

-John

Issue #840, opened 2015-10-27
early stopping when using --ftrl


zachmayer:
It seems that ftrl always uses all of the passes, regardless of whether or not `--holdout_off` is set.  It'd be nice to be able to use early stopping via the holdout in the same way I do with regular VW.

JohnLangford:
Agreed.  The code for this is here:

https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/gd.cc#L146

Can you send me a pull request to make it happen in ftrl.cc?

zachmayer:
See #843

PR #843, opened 2015-10-28
add ftrl early stopping

zachmayer:
Test pass for me locally and the `--early_stopping` flag for `--ftrl` seems to work, but this is the first piece of C++ code I've ~~written~~ copy/pasted in 15 years, so review carefully:
image


zachmayer:
See #840

JohnLangford:
Seems to be complaining that it can't automerge?

zachmayer:
You added test 104 while I was writing a test 104, which created a merge conflict.  I fixed that, it should be good to go now.  We'll see if the new tests pass!

JohnLangford:
This is becoming a common problem. Ariel, have you thought about how to 
make it so new tests don't collide with each other?

-John

JohnLangford:
Merged, thanks.

arielf:
As a good practice, I always `git pull jl master` before pushing any of my own changes.  This way I catch conflicts locally rather than having someone else hit them.  Even then there could be a short window of possible collision.  That's why git merging is there.

We could place tests as separate files in a test-data dir, instead of in the RunTests script, but this may not avoid people trying to create the same "test.099" file with different contents... Unless I misunderstand the problem and perceived solution.

JohnLangford:
If there was a directory of tests and people had meaningful names for 
tests, perhaps that would work better?

Historically, the window of collisions is not so short because I often 
am busy for several days before I can deal with pull requests.

-John

zachmayer:
@arielf I thought I had pulled master =D but John wrote a new test in the half hour window of divergence.  I should have pulled right before pushing, which I will do in the future.

@JohnLangford Personally, I don't mind resolving merge conflicts.  It's just a single file and a few lines of code so it's usually pretty straightforward to resolve.  Separate test files with meaningful names would work too.

JohnLangford:
It's certainly no biggy so far, but it is a bit of friction.

-John

arielf:
Switching to a directory of tests, or even a hierarchy of tests is not hard.

The question is how to impose execution order on these.  The order is important because some tests assume prior tests before them (e.g. model creation followed by a prediction test), plus some developers may have gotten used to certain familiar numbers as their tests.

We could do something like number prefixing to ensure ordering:
      tests/0099-somename.vwtest
or even two level:
      tests/0099-test-group-name/99-individual-test-name.vwtest

This means some additional complexity to grok/explain.

I don't mind much if that's what John prefers (can add to my TODO).
Personally, I feel the additional complexity may be greater than the benefit. Should we go there?

JohnLangford:
I hadn't appreciated the execution order issue.   Maybe a set of 
directories where each directory may have one or more tests with order 
imposed by numbers?

I agree, this seems a relatively low order concern.  The problem from 
naming collisions is only a minor nuisance.

-John

PR #844, opened 2015-10-29
fixed memory leaks

eisber:
Conflicts:
	vowpalwabbit/parse_args.cc


JohnLangford:
The try catch throw stuff is pretty unpleasant to the flow of the code.  Using proper constructor/destructors is becoming more appealing.

PR #845, opened 2015-10-29
Redoing JNI interface to be more elegant

jmorra:
After trying to integrate the new JNI layer into Aloha we ran into some problems.  This PR address these issues.  I have verified that everything works and it's MUCH better than before because you can't do something unsafe anymore.  I have solved all of my problems and feel much better about this version.


java/src/main/c++/vw_learner_VWFactory.cc
```diff
@@ -0,0 +1,39 @@
...
+  }
+  return vwPtr;
+}
+
+JNIEXPORT jobject JNICALL Java_vw_learner_VWFactory_getReturnType(JNIEnv *env, jobject obj, jlong vwPtr)
+{ jclass clVWReturnType = env->FindClass("vw/learner/VWReturnType");
+    jfieldID field = env->GetStaticFieldID(clVWReturnType , "Unknown", "Lvw/learner/VWReturnType;");
+    try
+    { vw* vwInstance = (vw*)vwPtr;
+        if (vwInstance->p->lp.parse_label == simple_label.parse_label) {
```
jmorra: @JohnLangford please check these if statements.  I'm not sure if they are correct and the one on line 30 is really bothering me.

java/src/main/c++/vw_learner_VWFactory.cc
```diff
@@ -0,0 +1,52 @@
...
+{ try
+  { VW::finish(*((vw*)vwPtr));
+  }
+  catch(...)
+  { rethrow_cpp_exception_as_java_exception(env);
+  }
+}
+
+JNIEXPORT jobject JNICALL Java_vw_learner_VWFactory_getReturnType(JNIEnv *env, jobject obj, jlong vwPtr)
+{
```
jmorra: @JohnLangford this really needs to be checked and updated.

java/src/main/c++/vw_learner_VWFactory.cc
```diff
@@ -0,0 +1,52 @@
...
+    vw* vwInstance = (vw*)vwPtr;
+    if (vwInstance->p->lp.parse_label == simple_label.parse_label) {
+        if (vwInstance->lda > 0)
+            field = env->GetStaticFieldID(clVWReturnType , "VWFloatArrayType", returnTypeInstance);
+        else
+            field = env->GetStaticFieldID(clVWReturnType , "VWFloatType", returnTypeInstance);
+    }
+    else if (vwInstance->p->lp.parse_label == MULTILABEL::multilabel.parse_label)
+        field = env->GetStaticFieldID(clVWReturnType , "VWIntArrayType", returnTypeInstance);
+    // This is really bad and we need to figure out why this is not working
```
jmorra: @JohnLangford please help me fix this.

java/src/main/java/vw/learner/VWFactory.java
```diff
@@ -0,0 +1,124 @@
...
+     *      VWIntLearner vw = VWFactory.getVWLearner("--cb 4");
+     *  }
+     * </pre>
+     * @param command The VW initialization command.
+     * @param <T> The type of learner expected.  Note that this type implicitly specifies the output type of the learner.
+     * @throws ClassCastException If the specified type T is not a super type of the returned learner given the command.
+     * @return A VW Learner
+     */
+    @SuppressWarnings("unchecked")
+    public static <T extends VWLearner> T getVWLeaner(final String command) {
```
jmorra: This is super magic sauce that allows the return type to be specified at object creation time AND it'll fail if the supplied command does not return the correct type.

java/src/main/c++/vw_learner_VWFactory.cc
```diff
@@ -0,0 +1,51 @@
...
+    if (vwInstance->p->lp.parse_label == simple_label.parse_label) {
+        if (vwInstance->lda > 0)
+            field = env->GetStaticFieldID(clVWReturnType , "VWFloatArrayType", RETURN_TYPE_INSTANCE);
+        else
+            field = env->GetStaticFieldID(clVWReturnType , "VWFloatType", RETURN_TYPE_INSTANCE);
+    }
+    else if (vwInstance->p->lp.parse_label == MULTILABEL::multilabel.parse_label)
+        field = env->GetStaticFieldID(clVWReturnType , "VWIntArrayType", RETURN_TYPE_INSTANCE);
+    else if (vwInstance->p->lp.parse_label == MULTICLASS::mc_label.parse_label ||
+             vwInstance->p->lp.parse_label == CB::cb_label.parse_label ||
```
jmorra: One minor note here is that I'm not sure what the difference between CB and CB_EVAL is.

jmorra:
@JohnLangford at this point I think this is ready to be merged in and I can do a maven release off of that commit.

deaktator:
I have a thought to make this API more elegant.  Please hold off merging.  I'll provide an update later today to tell if I could get my idea to work.

deaktator:
@JohnLangford @jmorra Prototyped the idea and it works.  It is simpler and automatically closes leaked JNI memory that wasn't previously closed.

see [https://github.com/deaktator/vowpal_wabbit-jmorra/blob/master/java/src/main/java/vw/learner/VWLearners.java]

jmorra:
So after thinking about this a LOT I went back to the way it was before.  This still has the potential to cause a memory leak if the user is very sloppy but there are copious comments in `VWLearners.create` telling the user about it and how to fix the leak.  @deaktator @JohnLangford any opinions?

deaktator:
I agree.  I like it better.  I have one suggestion.  I would make the base class override equals and hashcode. I think equals should be based on class type equality and nativePointer equality. I think hashcode should be entirely based nativePointer.  The reason I suggest this is that we can use these functions in an effort to automatically close instances that have gone out of scope without close being called.

We could change these later but it would be nice to have them stay consistent going forward.

Thoughts @jmorra?

java/src/main/c++/vw_learner_VWFactory.cc
JohnLangford: CB_EVAL has an augmented (by an action) label type, and it evaluates the 
policy implied by the sequence of actions rather than learning a policy.

-John

JohnLangford:
I looked through the .cc side---seems fine.  I skimmed the Java side.   
Seems fine from what I know. I'm ready to merge this in, but will wait 
until you two settle.

-John

jmorra:
@deaktator I've added equals and hashcode.  I think this is ready to be merged, do you?

jmorra:
@JohnLangford this build seems to have failed cause of some test issues on the C side, any ideas?

deaktator:
Go for it!  "Merge. Merge!"

jmorra:
OK, we're considering this done and it'll be released as Vowpal Wabbit version 8.1.1 to Maven once this is merged in and the build is passing again.

JohnLangford:
I fixed a test bug.  Bump again?

-John

jmorra:
I don't think it matters as 8.1.1 was never released anywhere.  My release script is designed to snag the version from configure.ac, so whatever version is in there will be the released version.  You might want to release 8.1.1 on your side too.

PR #848, opened 2015-10-30
fix buggy exp(-inf) on Windows

danmelamed:
https://github.com/JohnLangford/vowpal_wabbit/commit/35ccb4179582f69199702728bcb2d9fdf70e6919


JohnLangford:
What is the dmtest stuff?

JohnLangford:
Why are there changes to vw.sln?

JohnLangford:
Also is this still somehow inducing segfaults?  I don't see anything odd, except that you might want to replace isinf() with infpattern().   I believe the latter is more cross-platform.

danmelamed:
vw.sln changed to include the new header file.

dmtest was a separate commit that was not supposed to be part of this pull request; not sure why it was included.  Sorry, still figuring out the interface.  dmtest is just my own miniproject that I use for testing small code stubs.

JohnLangford:
Generally, you want to make a branch with just the stuff for the pull request.

JohnLangford:
I'd suggest editing vw.sln in a text editor---it seems that much more stuff is being changed than just adding a line for the header file.

danmelamed:
isinf() is a standard function in C++11, and a standard macro in C99.  infpattern() seems to rely on particular encodings, which I imagine different platforms could implement differently....

JohnLangford:
Either isinf() or isnan() turned out to not be cross-platform (I forget 
which).   Since this is only for windows, it doesn't matter much as long 
as it works on windows.

-John

arielf:
I wonder if it is possible to leave `exp()` alone in all files where it appears, and (possibly) define it to whatever-else on windows only.  It seems cleaner than changing all `exp()` calls to an uglier and less intuitive/familiar form because of a specific bug on a specific platform.  Generally, we should try to leave simple/original call alone.  I once suggested `platform.h` for this.

See:  http://pastebin.com/thHtb7vr   (not tested, but someone has to start somewhere :sweat_smile: )

JohnLangford:
I tried some #defines like this when I was first dealing with porting to
windows.   It did not seem to work well with visual studio unfortunately,
although I forget what the problem was exactly.

-John

vowpalwabbit/vw.sln
```diff
@@ -63,6 +63,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		cs_unittest.testsettings = cs_unittest.testsettings
 	EndProjectSection
 EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "dmTest", "..\dmTest\dmTest.vcxproj", "{DFD56302-EC7E-4480-9A44-A0467887ED20}"
```
JohnLangford: Can you remove these lines?

vowpalwabbit/vw.sln
```diff
@@ -207,6 +209,17 @@ Global
 		{E4E962AE-7056-4EB0-A8C5-8DC824A4B068}.Release|Win32.ActiveCfg = Release|x64
 		{E4E962AE-7056-4EB0-A8C5-8DC824A4B068}.Release|x64.ActiveCfg = Release|x64
 		{E4E962AE-7056-4EB0-A8C5-8DC824A4B068}.Release|x64.Build.0 = Release|x64
+		{DFD56302-EC7E-4480-9A44-A0467887ED20}.Debug|Mixed Platforms.ActiveCfg = Debug|Win32
```
JohnLangford: What does all of this do?

danmelamed: It just sets x64 as the build target.  I thought I excluded my vw.sln from any commits.  Maybe I failed to include it when I reverted the dmTest project.  Sorry… will try again.

D

PR #858, opened 2015-11-18
Support for new docker on mac during build

jmorra:
This will support the new docker installed by brew on mac.  This will allow the Java interface to be built quicker.


JohnLangford:
Merged in, thanks.

-John

PR #860, opened 2015-11-23
Add support for pkg-config

willnewton:
Add support for installing a pkg-config metadata file for VW libraries.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -66,5 +66,8 @@ noinst_HEADERS = vowpalwabbit/accumulate.h \
 
 ACLOCAL_AMFLAGS = -I acinclude.d
 
+pkgconfigdir = $(libdir)/pkgconfig
+pkgconfig_DATA = libvw.pc libvw_c_wrapper.pc
+
 test: all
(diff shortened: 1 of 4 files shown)


JohnLangford:
What is the use case for this?

JohnLangford:
Restated, if I break something, how will I know it?

willnewton:
The use case is various bits of Linux software packaging machinery (and supposedly Mac and Windows but I haven't tested that):

http://www.freedesktop.org/wiki/Software/pkg-config/

It makes it much easier if you are building a program or library that depends on VW to just add PKG_CHECK_MODULES to your configure.ac and it will set the CFLAGS and LIBS correctly.

In terms of what the risks are of breaking things - if a new library dependency was added then the pkg-config file might need updating to take account of that, if it wasn't peoples builds would fail, hopefully they would report it and it would be fixed. It should be fairly low maintenance unless you anticipate large numbers of changes of that type.

JohnLangford:
Ok, merged.   Library structure should not be changing too often.

-John

PR #862, opened 2015-11-23
install all required headers so that VW can be used as a library

ales-t:
When installed (``make install``), the header file ``example.h`` is copied into the ``include`` directory but its dependencies aren't. Also, the header ``ezexample.h`` which provides a simpler interface is not installed. This pull request simply adds these files in ``Makefile.am``.


willnewton:
Sorry, I missed this PR when I submitted #866

JohnLangford:
Thanks.

PR #863, opened 2015-11-23
C# refactoring, memory leak fixes, general goodness,...

eisber:
fixed Runtime library mismatch between zlib, libvw, vw.exe, VowpalWabbitCore.dll (CLR),... by using zlib/boost nuget provided msbuild targets
included Visual Leak Detector for memory leak detection on windows
refactored C# API to allow users to dynamically constructor serializers based on alternate descriptions (not just on static annotations)
string marshalling is compatible to command line (either escaping or splitting)
schema based pre-hashing: if hash can be determine from schema it's only generated once and re-used for each example.
added type extension API for marshalling
allow user to generate native and string examples in parallel in both debug and release
keep marshalling expression tree for debugging
refactored marshalling expression tree generation to improve readability
added sweeping helper
improved C# label parsing extensibility
added assembly signing
fixed memory leaks in C# usage of VW
fixed model hashing/reload interaction
fixed handling of empty line examples within set of action dependent features
fixed order issue when predicting ADF examples containing empty action dependent features
fixed default namespace incompatibility (space vs. 0)
improved RunTests to C# test wrapping (detects inter-test dependencies and input files)
unit tests are run in test/ folder, thus no need copy all input files
added user-supplied model id support


cs/Serializer/VowpalWabbitDefaultMarshallerExt.tt
```diff
@@ -0,0 +1,194 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
```
JohnLangford: I'm assuming the corresponding 2970 lines .cs file is automatically generated.  Right?  Can that be left out?

cs_unittest/TestAll.cs
```diff
@@ -0,0 +1,1014 @@
...
+        [TestMethod]
+        [Description("")]
+        [TestCategory("Command line")]
+        // [DeploymentItem(@"train-sets/0001.dat", @"train-sets")]
+        // [DeploymentItem(@"train-sets/ref/0097.stderr", @"train-sets\ref")]
+
+        public void CommandLine_Test97()
+        {
+
+            using (var vw = new VowpalWabbit("-d train-sets/0001.dat -f models/0097.model --save_resume"))
```
JohnLangford: There is quite a bit of repeat stuff here.  Why can't the test lines and references in RunTests be used directly?

JohnLangford: Cut and paste seems ... inelegant.

JohnLangford:
I looked through things up to the point where the bug starts.  It would help to have someone who understands C# look through the C# parts in more detail.

cs/Serializer/VowpalWabbitDefaultMarshallerExt.tt
eisber: Yes, I already have .tt compilation in place. Will remove it.

cs_unittest/TestAll.cs
eisber: It actually does use RunTests. There is a separate project (cs_testhelp) that parses RunTests and generates the tests.
I opted to actually have individual test methods to allow for simple debugging.

eisber: It’s generated by cs_testhelp, but I don’t run it at build time as it actually requires the tests to run (otherwise I’m not able to extract all command line arguments, remember the input file discussion?)

JohnLangford: So, this is another example of generated code?  Can you get rid of it?

-John

JohnLangford: Why do we need to checkin generated code?  It seems undesirable.

-John

JohnLangford:
Can you update this to work against the current master?

eisber:
Done. 1x Test (105) is failing with floating point differences… The rest passed.

JohnLangford:
It does not look like a test 105 issue---looks like a permissions problem on RunTests?

eisber:
Bogus merge… looking into it…

eisber:
RunTests is a separate issue (chmod +x was lost).

On Windows 106 fails with:

Assert.AreEqual failed. Expected a difference no greater than <1E-05> between expected value <0> and actual value <0.242642432451248>.

arielf:
The <0> looks like a failure to update to latest  `test-sets/ref/0001_ftrl_holdout_106.stderr`
The correct reference-file, shouldn't have an expected value of zero (used to have before this was fixed).

eisber:
It turns out that the stderr parsing code on Windows was missing the fact that average loss can be postfixed with “ h” for holdout.
It still fails due to differences. Is this legit?

Failed   CommandLine_Test105
Error Message:
   Assert.AreEqual failed. Expected a difference no greater than <1E-05> between expected value <0.246914> and actual value <0.242642432451248>. AverageLoss

arielf:
Hi Markus,

First a little background:
The perl RunTests fuzzy compare has two cases:  absolute & relative error.   When two numbers are being compared, first there's an absolute difference. If it is smaller than epsilon, we are done (success).  If the delta is too big, we do a second check and look for the difference of the ratio vs 1.0. This is to avoid failures on large numbers:

     if (abs(v1/v2 - 1.0) < Epsilon)
         success
     else
         failure

Unfortunately, in this case, even the fall-back (ratio) error is about 0.0176 according to the above numbers so the failure persists.  On both my system and the travis autobuild, there's no difference so the test succeeds.   I think this is a too fragile test.  Early stopping makes it do only two passes out of 10, so we have a case of not enough convergence in some environments (like appveyor).

Should we try to make the test converge a bit more?  e.g. by changing ftrl parameters?  Other ideas?

e.g. when I use:
```
  ../vowpalwabbit/vw -k -d train-sets/0001.dat -f models/0001_ftrl.model --passes 10 --ftrl --ftrl_alpha 1 --ftrl_beta 0 --l1 0.9
```
The average loss goes down from ~0.246 to ~0.107 which may make this test less sensitive to FP difference.  Might not be enough, I fear, but this is the direction I suggest trying.

JohnLangford:
Changing the test to have better convergence seems hard to object to.  

We could also create a -mswin version of the outputs if necessary.

cs_test/NIPS.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using VW;
+using VW.Labels;
+using VW.Serializer.Attributes;
+
+namespace cs_test
+{
+    public class NIPS
+    {
```
lhoang29: Is NIPS here referring to the conference? If so can you rename it to something else more relevant

cs_unittest/Rcv1Test.cs
```diff
@@ -11,29 +11,27 @@
...
-            using (var gz = new GZipStream(File.OpenRead(@"D:\Data\rcv1.train.vw.gz"), CompressionMode.Decompress))
-            using (var reader = new StreamReader(gz))
-            {
-                MyListener listener;
-                using (var vw = new VowpalWabbit<Data>("-k -f rcv1.model -c rcv1.cache"))
-                {
-                    listener = new MyListener(vw);
-                    string line;
+        //[Ignore]
+        //public void Rcv1Test()
```
lhoang29: delete?

cs_unittest/RunTests.tt
```diff
@@ -0,0 +1,218 @@
...
+    if (line.Trim().Length == 0)
+    {
+        if (skipTest)
+        {
+            Reset();
+            continue;
+        }
+
+		// resolve dependencies
+		if (testcase.FinalRegressor != null)
```
lhoang29: space/tab inconsistency

cs_test/NIPS.cs
eisber: Add link to slides.

cs_unittest/Rcv1Test.cs
eisber: done

cs/Serializer/VowpalWabbitDefaultMarshaller.cs
```diff
@@ -0,0 +1,358 @@
...
+            }
+
+            foreach (var kvp in value)
+            {
+                // TODO: not sure if negative numbers will work
+                context.AppendStringExample(
+                    feature.Dictify,
+                    " {0}:" + (context.VW.Settings.EnableStringFloatCompact ? "{1}" : "{1:E20}"),
+                    Convert.ToString(kvp.Key),
+                    (float)Convert.ToDouble(kvp.Value, CultureInfo.InvariantCulture));
```
lhoang29: Convert.ToSingle? Did you also resolve your TODO? Seems like negative numbers should work here

cs/Serializer/VowpalWabbitDefaultMarshaller.cs
```diff
@@ -0,0 +1,358 @@
...
+                context.NamespaceBuilder = context.ExampleBuilder.AddNamespace(ns.FeatureGroup);
+
+                var position = 0;
+                var stringExample = context.StringExample;
+                if (context.StringExample != null)
+                {
+                    position = stringExample.Append(ns.NamespaceString).Length;
+                }
+
+                featureVisits();
```
lhoang29: does this visit & marshal all features or just check if any feature is present?

eisber: It visit all the features, but if none of the visits generated a feature (in native terms), the namespace must not be added.

cs/Serializer/VowpalWabbitSerializer.cs
```diff
@@ -136,16 +176,37 @@ public VowpalWabbitExample Serialize(VowpalWabbit vw, TExample example, ILabel l
             }
             else
             {
-                result = new CacheEntry 
+                // TODO: catch exception and dispose!
```
lhoang29: remove todo. (I wonder if anything else in native VW needs to be freed in exception state, but with VLD integrated the unit tests are probably enough)

cs/Serializer/Intermediate/Namespace.cs
```diff
@@ -5,23 +5,63 @@
...
+
+            if (featureGroup == null && !string.IsNullOrWhiteSpace(name))
+            {
+                throw new ArgumentException("If Namespace is provided, FeatureGroup must be set too");
+            }
+
+            // compute shared namespace hash
+            this.NamespaceHash = name == null ?
+                vw.HashSpace(this.FeatureGroup.ToString()) :
+                vw.HashSpace(this.FeatureGroup + this.Name);
```
lhoang29: don't you need FeatureGroup.ToString() here?

eisber: HashSpace takes a string. FeatureGroup is a char.

lhoang29:
you'd want to sync up your fork since I still see some native code that were removed in the main repo

eisber:
Done

cs_unittest/Test1and2.cs
```diff
@@ -97,9 +100,9 @@ public void Test1and2()
             }
         }
 
-        //[TestMethod]
+        //// [TestMethod] // TODO
```
lhoang29: delete

JohnLangford:
Why is gd_mf_weights changed?

JohnLangford:
Why are attributes on mini_vw.py changed?

JohnLangford:
why are attributes changed on test/model-sets/7.10.2_corrupted.model?
On utl/vw-experiment ?

JohnLangford:
Why is accumulate.cc a giant diff that does nothing?

JohnLangford:
Why is struct bfgs reformatted?

JohnLangford:
It looks like the normalization code just isn't applied to bfgs.cc?

eisber:
Newline weirdness.. Looking into it.

JohnLangford:
It looks like there are many spurious changes to parse_regressor.cc also?  Can you canonicalize everything?

JohnLangford:
What is raii.h?

JohnLangford:
What is the purpose of vw_common?

eisber:
I re-ran the formatting scripts and seemed to go away – but the new lines are super weird here. It depends on the platform that you checkout. Also the diff I ran apparently runs against the repository version.
Can you check if it’s fixed in the way you run diffs?

Markus

JohnLangford:
gd_mf_weights appears no longer changed now.

eisber:
Can you send me your diff command, so I can verify myself too.

@ mini_vw.py: Visual Studio strips +x as soon as you open the file (happens to me everytime I look at RunTests).

eisber:
Fixed. VS messes with +x

JohnLangford:
In the branch for your code on my local machine I run:

git pull https://github.com/eisber/vowpal_wabbit.git master
git diff master | less

-John

eisber:
Resource Acquisition Is Initialization sample code. I removed it and made notes were it is…

eisber:
This is code that’s easier writeable/maintainable in C#, but I got a dependency on it from the C++/CLI component thus I’m unable to put it into cs/*.
Should I rename the project to cs_common?

Markus

eisber:
done

eisber:
I just reran Louies formatting commands and still see lots of change. Can this be right?

JohnLangford:
I'm a bit worried that we are developing too many top-level 
directories.  Having one per language binding seems reasonable.   This 
could be ok, but if you can think of a plausible way to not have it be 
top-level, maybe that is better?

-John

JohnLangford:
It should not be?  I'm pretty sure the bfgs.cc code is unchanged since 
the last version release.

-John

eisber:
That’s my local diff to upstream/master after running “pretty”. The indentation looks right on my version I’d say.

Markus


@@ -145,10 +145,10 @@ bool test_example(example& ec)
 { return ec.l.simple.label == FLT_MAX;
 }

-  float bfgs_predict(vw& all, example& ec)
+float bfgs_predict(vw& all, example& ec)
 { ec.partial_prediction = GD::inline_predict(all,ec);
-    return GD::finalize_prediction(all.sd, ec.partial_prediction);
-  }
+  return GD::finalize_prediction(all.sd, ec.partial_prediction);
+}

eisber:
I like the idea. Do you want me know to push everything into

cs/
cs/cli
cs/common
cs/unittest
…

?

JohnLangford:
That's much better than the diff I saw (around 1K lines).

-John

JohnLangford:
Sounds helpful.

-John

eisber:
Ok, I’ll do this as part of this PR.

Markus

eisber:
Can you run

find -regextype posix-extended -regex '.*\.(cc|h|cpp)' | xargs eclint fix
find -regextype posix-extended -regex '.*\.(cc|h|cpp)' | grep -v vw_clr | xargs astyle -s2 --style=horstmann --lineend=linux --keep-one-line-blocks --keep-one-line-statements

on vowpal_wabbit/vowpalwabbit/* and push?

Otherwise it’s super hard to see if I miss any real changes…

Markus

JohnLangford:
Done.

-John

eisber:
Pushed.

JohnLangford:
It says "


        This branch has conflicts that must be resolved"

Fix?

-John

JohnLangford:
Can all the cs_<stuff>/ go into cs/?

eisber:
That’s what I’m doing right now. Need to go through all the cs_proj files.

JohnLangford:
Cool.

eisber:
My tests are crashing after upstream merge…

JohnLangford:
Unclear what's up?  The master branch is in the same state as before as 
far as I know?

-John

eisber:
I believe there are more changes on the master compared to when I last sync’d. I moved all C# into cs/, but my VS Test environment crashes and I can’t properly debug.
Also changes in parse_regressor.cc (buff/buff2 allocation) introduced a memory leak – fixed, but it’s the usual ugliness (double free in catch(…) and outside)

Markus

JohnLangford:
What disables debugging?  For any file, you can get the changes between 
two checkins using: git diff <checkin1> <checkin2> <filename> .

-John

eisber:
Already looked at the changes and didn’t see anything suspicious. Investigating…

eisber:
My diff output below. From what I can find on the web, git only stores files as 644 or 755. I actually don’t understand why it should have been something else… I think 644 is the right mode – non-executable as it should be for a model file.

mwtds@ubuntu:~/vw.eisber/vowpal_wabbit/test/model-sets$ git diff upstream/master 7.10.2_corrupted.model
diff --git a/test/model-sets/7.10.2_corrupted.model b/test/model-sets/7.10.2_corrupted.model
new file mode 100644
index 0000000..05e9f4d

JohnLangford:
644 does seem like the right mode.  I'm just unclear on why it is 
changing at all.

-John

eisber:
From the diff it looks like there was no mode set all. I’m not sure if one can return to that state…

eisber:
Found 2 issues. Visual Leak Detector somehow sneaked into Debug builds.

I found the offending commit: 923bbe6b from Will Newton.
%z is not supported on Windows.

Changing to %llu http://stackoverflow.com/questions/6655410/why-doesnt-zd-printf-format-work-in-vs2010

Still one to go…

JohnLangford:
Cool.

-John

eisber:
Pushed.

Found the 2nd offender too. The extension for dynamic memory when reading model introduced for the Java API didn’t care for corrupted models.
The Diff got quite a big larger, as I moved all projects into cs/.

I’ll have to wait for Visual Leak Detector nuget to be build, minor change to make sure there isn’t an unwanted dependency onto VLD in Debug.

Markus

JohnLangford:
Checks failing still?

eisber:
Cs/ move around aftermath…

eisber:
fixed

JohnLangford:
Merged at last!

All, this is a major update to the C# .net library interface, plus many 
minor robustness and memory leak fixes.

-John

PR #865, opened 2015-11-24
make only vw target by default

danmelamed:


arielf:
This fails because there are some tests that test the library.
Need to make the target `library_example` as well by default.

danmelamed:
Fails where/how?  The Travis build passed….

arielf:
Sorry for not being clear!

Explanation: It passes travis because travis does `make all` rather than `make` (default) or `make vw`.

But, if I start from a clean tree and run:
```
     make
     make test
```
It works, but after this patch, it doesn't.

Then the 2nd step fails test 59 because `library/ezexample_predict` doesn't exist when building only `vw`

I think testing the library example (C/C++ only, no other language bindings) is important not to leave out.  And `make test` should succeed by default.

So the default in the `Makefile` should be:
```
default: vw  library_example
```

Rather than just:
```
default: vw
```

Sorry again for not being clear!

danmelamed:
library_example is already a dependency of the 'test' target.  So even though it won't be built by 'make', it will be built by 'make test', before it runs any tests.  Indeed this works for me from a fresh tree.  Are you seeing something different?

arielf:
Sorry again!

You're right.  It was my fix, actually.  I didn't realize I made the fix to `test` rather than `default` target.

So this patch is good to go. 

Thanks for your patience and sorry for my confusion.

PR #866, opened 2015-11-24
Makefile.am: Install more headers

willnewton:
Try and install a set of headers that is totally self-contained.


JohnLangford:
I merged the earlier one, so probably not this one?

Issue #859, opened 2015-11-21
Model is not saved when --passes exceeds 2


kirillfish:
Recently I asked a question [here](http://stackoverflow.com/questions/33800919/vowpal-wabbit-doesnt-save-the-model-despite-f-flag-is-present) on StackOverflow.

Actually, there appeared to be a much more minimalistic way to describe the problem. Here is the training set:

-1 |D a
1 |D b
-1 |D a
-1 |D c
1 |D d
-1 |D a
1 |D b
1 |D a
-1 |D d

After running this command:
`vw toy_example -f heythere.model -c --loss_function logistic --passes 3`
no model is saved, as well as for greater --passes values. 

But these commands:
`vw toy_example -f heythere.model -c --loss_function logistic --passes 2`
`vw toy_example -f heythere.model -c --loss_function logistic --passes 1`
both work as expected.

arielf:
I can 100% reproduce this.
`--loss_function logistic` isn't necessary to reproduce.

The bug has been there for a long time (I went back to 2013 and it is still there)

It seems to be related to holdout and hold-out period.

The bug disappears if any of the following hold true:
    * Add `--holdout_off` to the options, no hold-out, no issues.
    * Increase passes to 10 or more (strangely: 2 passes work byt 3 to 9 don't, 10 passes work again)
   * Add more examples: 10 or more examples make it work again

 Here's a script that can easily show the bug
```
$ cat 3pass-no-model-save.sh
#!/bin/bash
#
# https://github.com/JohnLangford/vowpal_wabbit/issues/859
# -f model - model doesn't get saved with 3-9 passes
#
# Broken:
# VW=~/bin/vw-8.20151121
# VW=~/bin/vw-8.20151101
# VW=~/bin/vw-8.20151019
# VW=~/bin/vw-8.20151013
# VW=~/bin/vw-8.20151005
# VW=~/bin/vw-8.20151002
# VW=~/bin/vw-8.20150927
# VW=~/bin/vw-8.20150902
# VW=~/bin/vw-7.20150623
# VW=~/bin/vw-7.20150609
# VW=~/bin/vw-7.20150608
# VW=~/bin/vw-7.20150527
# VW=~/bin/vw-7.20140824

VW=`which vw`

MyName=`basename -s .sh $0`

TrainSet="$MyName.train"
Model="$MyName.model"

test_it() {
    passes=$1

    pass_opt=""
    case "$1" in
[...]

PR #867, opened 2015-11-24
Multidimensional vw hyperparameter optimization with hyperopt

kirillfish:
All functionality (and even a bit more) and almost all syntax is in the line with @martinpopel 's suggestion here: https://github.com/martinpopel/vowpal_wabbit/wiki/vw-hyperopt-plans.
See also SO questions: 
http://stackoverflow.com/questions/33262598/get-holdout-loss-in-vowpal-wabbit
http://stackoverflow.com/questions/33242742/multidimensional-hyperparameter-search-with-vw-hypersearch-in-vowpal-wabbit

Thank you @martinpopel @arielf for inspiration! I did it!

Unlike `vw-hypersearch`, `vw-hyperopt.py` can be multidimensional. It implements Tree of Parzen Estimators and Random search algorithms from `hyperopt`. TPE uses adaptive sampling strategy and addresses explore-exploit dilemma. [There is some research](http://papers.nips.cc/paper/4443-algorithms-for-hyper-parameter-optimization.pdf) showing that TPE may return better hyperparameter configuration than manual expert choice or grid search. 

Here is an example of using `vw-hyperopt`:

```bash
python vw-hyperopt.py --train train.dat --holdout holdout.dat --max_evals 200 --outer_loss_function logistic --vw_space '--algorithms=[ftrl,sgd] --l2=[1e-8..1e-4]LO --l1=[1e-8..1e-4]LO -l=[0.01..10]L --ftrl_alpha=[5e-5..5e-1]L --ftrl_beta=[0.01..1] --passes=[1..10]I -q=[SE+SZ+DR,SE]O --ignore=[T]O --loss_function=[logistic] -b=[29]'
```

The quoted part after `--vw_space` flag means literally the same thing as @martinpopel described. It is converted to `hyperopt` tree-like search space. There is even additional functionality: you can list different combinations of quadratic features, just separating namespace combinations by "+" symbol (see an example). 

Another new functionality is that you can optimize hyperparameters with respect to a custom metric, such as ROC-AUC, that can be different from inner vw loss function. This can be useful sometimes. The flag corresponding to this is `--outer_loss_function`. By now, only `logistic` (default) and `roc-auc` are implemented, but this list can be easily expanded. 

Yet another additional feature is that you can specify several algorithms at once. They will be converted to `hyperopt.hp.choice()` object. Currently SGD and FTRL-Proximal are supported. If there are prohibited flags, such as `--ftrl_alpha` for SGD, they will simply be excluded from `hyperopt` search space for this particular algorithm. 

You can also specify the maximum number of hyperparameter combinations to explore with `--max_evals` (default=100). 

The possible (non-critical, I hope) problems and ways to improve my module, as I see it now:

1) Some difficulties with exploiting natural VW holdout error-handling functionality (see [this issue](https://github.com/JohnLangford/vowpal_wabbit/issues/859)) motivated me to use *separate train and holdout sets*. But, if I understood correctly, the same way was proposed [here](https://github.com/martinpopel/vowpal_wabbit/wiki/vw-hyperopt-plans). So the dataset should be manually split before using `vw-hyperopt`. 
2) All the training is now done with `--holdout_off` flag in order to make use all the training data, and also to make sure that model is always saved (see the mentioned issue).
3) In light of (1), supplementary files "current.model" and "holdout.pred" are created in the course of optimization (and deleted at the end). The validation metrics are computed with `scikit-learn` methods using these files.
4) Regression problems are not yet supported. In the nearest future I'm going to add everything needed for them.
5) Need to write "python" in the beginning of a command and to quote a search space part. I hope this doesn't complicate things too much.
6) If the code needs to be better documented, I can do it.
7) Plotting and visualization are still to be done. I'm going to make it with matplotlib+seaborn.
8) Maybe there is better format for the final output. Now it writes to stdout (and to a log file) something of this kind:

```
2015-11-25 00:06:20,853 INFO     [root/vw-hyperopt:290]:

A FULL TRAINING COMMAND WITH THE BEST HYPERPARAMETERS:
vw -d train.dat -f current.model --holdout_off -c  --passes 5 --ftrl_alpha 0.418819407315 -l 1.43698068436 -b 29 -q SE -q SZ -q DR --loss_function logistic --ftrl_beta 0.428875395523 --ftrl

2015-11-25 00:06:20,853 INFO     [root/vw-hyperopt:291]:

THE BEST LOSS VALUE:
0.0207917909491
```

### DIFF
--- a/utl/vw-hyperopt.py
+++ b/utl/vw-hyperopt.py
@@ -1,0 +1,307 @@
+# coding: utf-8
+
+"""
+Github version of hyperparameter optimization for Vowpal Wabbit via hyperopt
+"""
+
+__author__ = 'kurtosis'
+
+from hyperopt import hp, fmin, tpe, rand, Trials, STATUS_OK
+from sklearn.metrics import roc_curve, auc, log_loss, precision_recall_curve
+import numpy as np
+from datetime import datetime as dt
+import subprocess, shlex
+from math import exp, log
+import argparse
+import re
+import logging
+
+
+def read_arguments():
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument('--searcher', type=str, default='tpe', choices=['tpe', 'rand'])
+    parser.add_argument('--max_evals', type=int, default=100)
+    parser.add_argument('--train', type=str, required=True, help="training set")
+    parser.add_argument('--holdout', type=str, required=True, help="holdout set")
+    parser.add_argument('--vw_space', type=str, required=True, help="hyperparameter search space (must be 'quoted')")
+    #parser.add_argument('--loss_function', default='logistic',
+    #                    choices=['logistic', 'squared', 'hinge', 'quantile', 'classic'])
+    parser.add_argument('--outer_loss_function', default='logistic', choices=['logistic', 'roc-auc'])  # TODO: implement squared, hinge, quantile, PR-auc
+    parser.add_argument('--regression', action='store_true', default=False, help="""regression (continuous class labels)
+                                                                        or classification (-1 or 1, default value).""")
+
+    args = parser.parse_args()
+    return args
+
+
+class HyperoptSpaceConstructor(object):
+    """
+    Takes command-line input and transforms it into hyperopt search space
+    """
+
+    def __init__(self, command=("--algorithms=[ftrl,sgd] --l2=[1e-8..1e-4]LO "
+                                "-l=[0.01..10]L --ftrl_beta=[0.01..1] "
+                                "--passes=[1..10]I -q=[SE+SZ+DR,SE]O")):
+        self.command = command
+        self.space = None
+        self.algorithm_metadata = {
+            'ftrl': {'arg': '--ftrl', 'prohibited_flags': set()},
+            'sgd': {'arg': '', 'prohibited_flags': {'--ftrl_alpha', '--ftrl_beta'}}
+        }
+
+        self.range_pattern = re.compile("(?<=\[).+(?=\])")
+        self.distr_pattern = re.compile("(?<=\])[IOL]*")
+        self.only_continuous = re.compile("(?<=\])[IL]*")
+
+    def _process_vw_argument(self, arg, value, algorithm):
+        distr_part = self.distr_pattern.findall(value)[0]
+        range_part = self.range_pattern.findall(value)[0]
+        is_continuous = '..' in range_part
+
+        if not is_continuous and self.only_continuous.findall(value)[0]!='':
+            raise ValueError(("Need a range instead of a list of discrete values to define "
+                              "uniform or log-uniform distribution. "
+                              "Please, use [min..max]%s form") % (distr_part))
+
+        if is_continuous and arg == '-q':
+            raise ValueError(("You must directly specify namespaces for quadratic features "
+                              "as a list of values, not as a parametric distribution"))
+
+        hp_choice_name = "_".join([algorithm, arg.replace('-','')])
+
+        try_omit_zero = 'O' in distr_part
+        distr_part = distr_part.replace('O','')
+
+        if is_continuous:
+            vmin, vmax = [float(i) for i in range_part.split('..')]
+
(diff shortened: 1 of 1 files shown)


JohnLangford:
This seems like an obviously good idea and it's entirely isolated so I
merged it.  Thanks much.

Ariel has a number of good suggestions.  I'm sure working through them
would help improve usability.

-John

P.S. If you have happen to be going to NIPS, tell me and we'll squeeze you
into the VW tutorial (http://hunch.net/).

kirillfish:
Hi Ariel, thank you for the response! Here are the answers to your comments:

* Yes, I'll make it executable and commit again, thanks.
* The main reason for logging is that typically hyperparameter tuning takes at least dozens of iterations and is rather lengthy process, even for VW (when it comes to millions of observations). It cat take hours and even days. So I thought it may be better if messages are stored somewhere on disc so that user could easier access them later, in addition to being displayed on the screen (now they are displayed, too). I can remove this and leave only printing, if it's needed.
* Ok, I'll make notifications lowercase.
* As yet, too much information is printed to stdout to also display dots or progress bar. If all the messages are not suppressed, or `--quiet` vw flag is not added by default, then progress dots will be lost in the clutter.
* What do you mean by default values? For the time being, only `--max_evals`, `--searcher` and `--outer_loss_function` have some values by default (100, 'tpe' and 'logistic', respectively). Maybe it's worth to make them match `vw` core default values (that is `squared` loss)? Search space has no default ranges now.
* Initially, I quoted the `--vw_space` value because it inevitably contains spaces, and quoting is the only way to make python `argparse` module treat it as a single argument. The syntax with square brackets is taken from @martinpopel 's forked Vowpal Wabbit repository: https://github.com/martinpopel/vowpal_wabbit/wiki/vw-hyperopt-plans
The meaning of [] is that it separates the range of possible values and distribution options (such as L, I, O). For instance, this: `-q=[SE+SZ+DR,SE]O` translates as "take SE quadratic feature, or SE and SZ and DR quadratic features, and *try also omit* this flag at all". L and I letters denote log-uniform and integer-valued uniform prior distributions, respectively (if there is .. of course). An absence of letters means real-valued uniform prior distribution. If there were no "]", the letters would fuse to the range of values and become difficult to parse. Maybe I'd better try another separators such as tilde, so ranges look like `-l=0.01..10~L ` ?
* `--algorithms` flag is really parsed differently. The reason is that I wanted to make completely separate search spaces for different algorithms. What would happen if `--algorithms` is counted as ordinary discrete-valued hyperparameter? The following thing. On the one hand, discrete values are treated as completely non-correlated by `hyperopt`, so in principle an independent sub-space should be created by `hyperopt` for each such value. On the other hand, the points from the *one single pool of hyperparameters* will be drawn for each iteration, regardless of the algorithm. So for the given algorithm there can be "prohibited" dimensions in search space. `hyperopt` will be mislead and will try to elicit `--ftrl_beta` contribution to SGD performance, when deciding what point to sample next. This may reduce efficiency and increase the number of steps required to discover some good configuration. That's why I treated `--algorithm` option the way I did.

Nevertheless, it seems like I can unify the syntax of the command-line. Intuitively, this expression: `--algorithms=,--ftrl` should be written as `algorithms=[--ftrl]O` following the logic just described, that is "try --ftrl or default algorithm". But now it is written as `--algorithms=[ftrl,sgd]` (`sgd` converts not to `--sgd`, but to default algorithm with AdaGrad etc.). Omitting `--algorithms` is equivalent to writing `--algorithms=[sgd]`.

So, should I: (a) change separator of range and distribution options? (b) make `--algorithms` usage syntax the same as for other flags? (c) make something else?

On top of this, `hyperopt` allows to save any useful information in the course of optimization. For instance , it may be useful to make diagnostic plots by default, such as Iteration number vs. The best score obtained by this iteration. Also train loss can be tracked and plotted to prevent overfitting. Should I add such functionality?

kirillfish:
@JohnLangford @arielf Thank you!

PR #868, opened 2015-11-25
adding scikit-learn interface

gramhagen:
adding python interface for using vowpal wabbit with scikit-learn
- allows calling vowpal wabbit as a sklearn classifier or regressor and using sklearn functionality (cross-validation hyper-parameter tuning)
- includes a utility (tovw) to convert sklearn data formats to vowpal wabbit input format
- added example ipython notebook to demonstrate capabilities
- unit tests using pylint

the interface itself requires scipy, numpy, and scikit-learn


JohnLangford:
Merged in.   I know many people use scikit-learn, so this could be handy to
many people.

-John

PR #870, opened 2015-11-27
Reduce warnings and enable strict aliasing

willnewton:
These changes reduce the levels of warnings that can be generated from the headers. These are only seen with --enable-nitpick or if the headers are included in a source base that has those some of those additional warnings enabled.

In the process it fixes the aliasing warnings and removes -fno-strict-aliasing. The latter is a behavioural change which should provide potentially better optimization of the code but may cause bugs if subtle alias issues are present in the code. The tests still pass and I went back into the git history but couldn't find the precise reason why the flag was added in the first place.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,6 +4,7 @@ nobase_include_HEADERS = vowpalwabbit/allreduce.h \
 	vowpalwabbit/comp_io.h \
 	vowpalwabbit/config.h \
 	vowpalwabbit/example.h \
+	vowpalwabbit/floatbits.h \
 	vowpalwabbit/global_data.h \
 	vowpalwabbit/hash.h \
 	vowpalwabbit/io_buf.h \
--- a/vowpalwabbit/Makefile.am
+++ b/vowpalwabbit/Makefile.am
@@ -25,7 +25,7 @@ endif
 if VWBUG
 CXXOPTIMIZE += -g -O1
 else
-CXXOPTIMIZE += -O3 -fomit-frame-pointer -fno-strict-aliasing -DNDEBUG
+CXXOPTIMIZE += -O3 -fomit-frame-pointer -DNDEBUG
 endif
 
 if NITPICK
--- a/vowpalwabbit/Makefile.am
+++ b/vowpalwabbit/Makefile.am
@@ -13,7 +13,7 @@ libvw_la_LIBADD = liballreduce.la
 
 ACLOCAL_AMFLAGS = -I acinclude.d
 
-AM_CXXFLAGS = ${BOOST_CPPFLAGS} ${ZLIB_CPPFLAGS} ${PTHREAD_CFLAGS} -Wall -Wno-unused-local-typedef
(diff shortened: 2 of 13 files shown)


JohnLangford:
This seems good, except that quake_inv_sqrt is performance sensitive and
hence the memcpy is pretty undesirable.  It looks like the correct and
perfomant method for handling this is via type-punning using a union type.
Can you create inline functions doing that instead?  After that, I'll merge.

I have no recollection where no-strict-aliasing came in or why.  Let's turn
it off and see if anything breaks.

-John

JohnLangford:
I agree the memcpy looks like a pretty terrible idea for performance but it
actually works out ok. Essentially the compiler can see that what we are
doing is:

memcpy(4 byte aligned ptr, 4 byte aligned ptr, 4 bytes)

and optimize accordingly. For example on x86_64 at -O2 gcc 5.1 generates:

float_to_bits:
        movd    %xmm0, %eax
        ret
bits_to_float:
        movl    %edi, -4(%rsp)
        movss   -4(%rsp), %xmm0
        ret

And similarly on ARM gcc 4.9 at -O2:

float_to_bits:
        fmrs    r0, s0  @ int
        bx      lr
bits_to_float:
        fmsr    s0, r0
        bx      lr

It would be possible to rewrite with the union trick but my understanding
is whilst the union trick is widely supported in compilers it is strictly
undefined behaviour by the C++ standard to use a union in that fashion.

JohnLangford:
Wikipedia suggests otherwise:
https://en.wikipedia.org/wiki/Type_punning
in the sense that using type punning is explicitly allowed in some ways
(you just need to be careful).  Type punning seems better than the assembly
above since all the 'mov' operations are dropped.

-John

willnewton:
Wikipedia is a little unclear and I am not convinced it is authoritative on
this topic (although I am not sure who is!). Type-punning via a union
explicitly works with gcc so it seems fine to use that approach, although
it does appear that, for example, Sun CC will miscompile this type of code:

http://blog.qt.io/blog/2011/06/10/type-punning-and-strict-aliasing/

I believe the movs are in order to get the data into the right place for
the calling convention and will be optimized out if the function is
inlined. For the compilers I have the union type punning and memcpy
approaches produce exactly the same code but I have respun the branch with
the type punning approach.

JohnLangford:
Do you have an objection to using reinterpret_cast?

-John

willnewton:
No, I think that would work too. It also seems to compile to the same thing.

willnewton:
Actually using reinterpret_cast seems to trigger strict aliasing warnings:

floatbits.h: In function ‘uint32_t float_to_bits(float)’:
floatbits.h:11:42: warning: dereferencing type-punned pointer will break
strict-aliasing rules [-Wstrict-aliasing]
   return *reinterpret_cast<uint32_t *>(&a);
                                          ^
floatbits.h: In function ‘float bits_to_float(uint32_t)’:
floatbits.h:16:39: warning: dereferencing type-punned pointer will break
strict-aliasing rules [-Wstrict-aliasing]
   return *reinterpret_cast<float *>(&a);

JohnLangford:
The new one is merged now, thanks.  Closing this one.

PR #875, opened 2015-12-04
Fix test 105 reference stderr (was incorrect): train-sets/ref/0001_ft…

arielf:
…rl_holdout.stderr


arielf:
Looks like fixing this is insufficient for fixing `make test`.

Now the test after it fails.

The two tests (105 and 106) have the same reference stderr which is incorrect since their output is different.  Looking into this.

arielf:
The updated PR fixes the stderr of test 106 but there's a regression in the predict output:
`test/pred-sets/ref/0001_ftrl_holdout.predict` so I'm reluctant to override this reference.

JohnLangford:
Working on it.

JohnLangford:
The fix to holdout for ftrl seems to have revealed a bug in the testing framework?  

The cascading effect was just due to test 106 loading a regressor from test 105.

PR #876, opened 2015-12-06
dynamic buffer for file_options

### LINKED ISSUE #831: Suggestion: avoid fixed buffer usage while storing options in model file
Hi,
< Some unnecesary details>
Today I've trained my model with `--cubic :::`. Then demangled its hashes (I'm going to contribute this module in 16Q1) and got millions of records `<"namespace^feature1 * namespace^feature2 * namespace^feature3", hash, weight>`. Then I summarized absolute values of weights by namespaces and sort them. Thus I've got ~50k namespaces' interactions ordered by their importance in regressor. So I took first 100 of them and spoof `--cubic :::` with explicit calls of necessary cubic  interactions.
I trained the new model and at the test step I've faced with fact that my model file is corrupted because the `all.file_options` contained more than 512 characters (list of 100 --cubic XXX parameters) and it was truncated by buf_size const [here](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parse_regressor.cc#L320).   
</ Some unnecesary details>  

My suggestion is to avoid usage of fixed buffer while storing `file_options`. Let it allocates its own memory chunk big enough to take the whole string.

trufanov-nok:
resolves #831 

To reproduce the problem you may use following commands:
```
$ ./vw ../test/train-sets/rcv1_small.dat --leave_duplicate_interactions -f temp.model  -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff -q ff

$./vw -t ../test/train-sets/rcv1_small.dat -i temp.model
vw (io_buf.h:239): bad model format!

```

Perhaps the proposed solution is a bit excessively for such a small problem. I was intended to replace all `memcpy` and `sprintf_s` calls with their `safe_` analogues. But as default buffer size can't be really overflowed anywhere except `all->file_options` by the nature of serialized data I decided to make less changes as possible. This also minimizes possible performance hits. And I also decided to use 2 calls of `bin_read_fixed` instead of adding some functions to io_buf that let user to peek and test object size before calling `bin_text_read_write_validated`.
Well, so it doesn't look like elegant solution after all. Let me know if you prefer to remove `safe_` functions and  just encode this logic into `save_load_header` body.

### DIFF
--- a/vowpalwabbit/parse_regressor.cc
+++ b/vowpalwabbit/parse_regressor.cc
@@ -338,7 +338,8 @@ void save_load_header(vw& all, io_buf& model_file, bool read, bool text)
     {
         uint32_t len;
         size_t ret = bin_read_fixed(model_file, (char*)&len, sizeof(len), "");
-        if (len > 512 || ret < sizeof(uint32_t)) THROW("bad model format!");
+        if (len > 104857600 /*sanity check: 100 Mb*/ || ret < sizeof(uint32_t))
+            THROW("bad model format!");
         resize_buf_if_needed(buff2, buf2_size, len);
         bytes_read_write += bin_read_fixed(model_file, buff2, len, "") + ret;
         all.file_options->str(buff2);
--- a/vowpalwabbit/parse_regressor.cc
+++ b/vowpalwabbit/parse_regressor.cc
@@ -10,11 +10,13 @@ using namespace std;
 #ifndef _WIN32
 #include <unistd.h>
 #define sprintf_s snprintf
+#define vsprintf_s vsnprintf
 #endif
 #include <stdlib.h>
 #include <stdint.h>
 #include <math.h>
 #include <algorithm>
+#include <stdarg.h>
 
 #include "rand48.h"
 #include "global_data.h"
@@ -44,11 +46,49 @@ void initialize_regressor(vw& all)
       all.reg.weight_vector[j << all.reg.stride_shift] = (float)(frand48() - 0.5);
 }
 
-const size_t buf_size = 512;
+const size_t default_buf_size = 512;
+
+bool resize_buf_if_needed(char *& __dest, size_t& __dest_size, const size_t __n)
+{
+    if (__dest_size < __n)
+    {
+        if ( (__dest = (char*) realloc(__dest, __n)) == NULL)
+            THROW("Can't realloc enough memory.");
+        __dest_size = __n;
+        return true;
+    }
+    return false;
+}
+
+int32_t safe_sprintf_s(char *& buf, size_t& buf_size, const char * fmt, ...)
+{
+    va_list args;
+    va_start(args,fmt);
+    int32_t len = vsprintf_s(buf, buf_size, fmt, args);
+    va_end(args);
+    if (len < 0) THROW("Encoding error.");
+    if (resize_buf_if_needed(buf, buf_size, len+1))
+    {
+        va_start(args,fmt);
+        vsprintf_s(buf, buf_size, fmt, args);
+        va_end(args);
+    }
+
+
+    return len;
+}
+
+inline void safe_memcpy(char *& __dest, size_t& __dest_size, const void *__src, size_t __n)
+{
+  resize_buf_if_needed(__dest, __dest_size, __n);
+  memcpy(__dest, __src, __n);
+}
 
 void save_load_header(vw& all, io_buf& model_file, bool read, bool text)
-{ char buff[buf_size];
-  char buff2[buf_size];
+{ char* buff = (char*) malloc(default_buf_size);
+  size_t buf_size = default_buf_size;
+  char* buff2 = (char*) malloc(default_buf_size);
+  size_t buf2_size = default_buf_size;
   uint32_t text_len;
 
   if (model_file.files.size() > 0)
@@ -58,7 +98,7 @@ void save_load_header(vw& all, io_buf& model_file, bool read, bool text)
     text_len = sprintf_s(buff, buf_size, "Version %s\n", version.to_string().c_str());
     memcpy(buff2, version.to_string().c_str(), min(v_length, buf_size));
     if (read)
-    { v_length = buf_size;
+    { v_length = buf2_size;
     }
     bytes_read_write += bin_text_read_write_validated(model_file, buff2, v_length,
                         "", read,
@@ -112,7 +152,7 @@ void save_load_header(vw& all, io_buf& model_file, bool read, bool text)
     VW::validate_num_bits(all);
 
     if (all.model_file_ver < VERSION_FILE_WITH_INTERACTIONS_IN_FO)
-    { // -q, --cubic and --interactions are saved in vw::file_options
+    { // -q, --cubic and --interactions are not saved in vw::file_options
       uint32_t pair_len = (uint32_t)all.pairs.size();
       text_len = sprintf_s(buff, buf_size, "%d pairs: ", (int)pair_len);
       bytes_read_write += bin_text_read_write_fixed_validated(model_file, (char *)&pair_len, sizeof(pair_len),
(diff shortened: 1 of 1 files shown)


JohnLangford:
Something is barfing in the java test:
https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/95160808#L1899

But it may be a transient. @jmorra any opinions?

jmorra:
I don't see anything obvious here that would cause the problem.  There is clearly an issue on the C side because of the error reported.  I would start by trying to get the generated `hs_err_pidxxxx.log` file.  @JohnLangford, is there anyway to get the file mentioned in the Travis logs?  If there is not, can this be reproduced on your machine?  If so send me a copy of the `hs_err_pidxxxx.log` file and we can look at it.  Without seeing this file I can't really help much as this will list the stack trace on the C side.

JohnLangford:
I don't think travis can produce output other than via cat.

trufanov-nok:
I'll try to bisect the problem a bit later by playing with PR's code.

trufanov-nok:
Ok, it was surprisingly hard to debug. But I found the problem.
1. I've replaced a call to `bin_text_read_write_validated()` by 2 calls to `bin_read_fixed()` in read mode. In fact I've reimplemented `bin_read()`. This allowed me to check object size and increase dynamic buffer if needed before the object is read. Unfortunately, I hadn't copied `bin_read()` exactly. I omitted the [lines](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/io_buf.h#L238) that throws exemption `"bad model format!"` as it's not so important when you have a dynamic buffer.  

2. It looks like one of Java tests literally expects  `"bad model format!"` exception to be thrown. If you comment this check in my [code](https://github.com/trufanov-nok/vowpal_wabbit/commit/84c75aaaf1a9221551df7b2e23b2fd2efa7b77a9#diff-6e06dc02b6816202f8189cbcb5c08434L341) the test fails.

3. Worst thing that VW throws "bad model format!" in case object length is greater than default buffer (512). This means that even if you use dynamic buffer and you are not limited with 512 bytes anymore you still must throw this exemption if you want to pass Java tests. And this makes dynamic buffer useless.
You can take a look in my latest [commit](https://github.com/trufanov-nok/vowpal_wabbit/commit/84c75aaaf1a9221551df7b2e23b2fd2efa7b77a9#diff-6e06dc02b6816202f8189cbcb5c08434L341) - it just adds `len > 512 ||` in exception check and makes things work.
----
Obviously, we must change Java tests to resolve this. I don't think we need a Java test that checks if VW throws "bad format!" anyway. Perhaps this test is just broken.

@jmorra Could you take a look into this and confirm that one of Java tests fails if VW result isn't a "bad format!"? Could you fix it or point me to a right piece of code?

jmorra:
You are correct, one of the tests does look for this string.  The test is called testOldModel in [VWLearnersTest.java](https://github.com/JohnLangford/vowpal_wabbit/blob/master/java/src/test/java/vw/learner/VWLearnersTest.java).  In this class is also another test called testBadModel that tests a malformed VW model.  The reason that these tests were written to detect specific strings is because there was no way to distinguish the Exception by type that was being thrown by VW.  If you have a look at [jni_base_learner.cc](https://github.com/JohnLangford/vowpal_wabbit/blob/master/java/src/main/c%2B%2B/jni_base_learner.cc) the function called rethrow_cpp_exception_as_java_exception shows how exceptions are propagated from the C side to the Java side.  I really didn't want to switch on a string in the main code, hence if the exception type could not be distinguished I just returned a Java Exception.  However, in the test I wanted a way to ensure that the test was performing properly hence I checked the error string directly.  In my opinion the proper way to solve this would be by updating rethrow_cpp_exception_as_java_exception to throw different Exception types.  I would actually proposed making two new Java classes, BadModelFormat and OldModelFormat that extend exception and are thrown as needed.  This would mean the Java tests would change to detect these specific exception types and the string test could go away and the same guarantees would hold as currently are in the tests.

trufanov-nok:
Ok, so the problem was in testOldModel. VW tries to read file_options from chunk of binary data which gives an object of huge size. My problem is that I hadn't made any sanity checks on size of object that could be read. Now I've replaces old 512 bites limit with 100 Mb. It allows to pass the test and still benefit from dynamically allocated buffer.

As for proposed changes to Java exception handling - I wrote my last piece of java 10 years ago and barely can suggest anything that has sense here.

P.S. I'm quite surprised that VW isn't compatible with 7.8 models. At least with one that is used for java checks. @JohnLangford , is this correct?

vowpalwabbit/parse_regressor.cc
```diff
@@ -294,21 +334,27 @@ void save_load_header(vw& all, io_buf& model_file, bool read, bool text)
...
-
     if (read)
-    { len = buf_size;
+    {
+        uint32_t len;
+        size_t ret = bin_read_fixed(model_file, (char*)&len, sizeof(len), "");
+        if (len > 104857600 /*sanity check: 100 Mb*/ || ret < sizeof(uint32_t))
+            THROW("bad model format!");
+        resize_buf_if_needed(buff2, buf2_size, len);
+        bytes_read_write += bin_read_fixed(model_file, buff2, len, "") + ret;
```
JohnLangford: I'm worried that the output of this call is not checked as in _validated().  The _validated() functions were added to prevent corrupt model failure modes.   Can you add the same checks in?

JohnLangford:
W.r.t. compatibility: it should be backwards compatible.  Is there evidence against this?

The Java code explicitly tests that the right thing happens when a corrupted model is loaded so as to prevent corrupted models from damaging front-end runtime environments.  (A bunch of changes to model loading were added to make this quite robust.)

i think this patch is read to go with the extra validation check.

trufanov-nok:
It doesn't look that there is any additional check that _validated function performs. Just look at it:
```
{ size_t nbytes = bin_text_read_write(io, data, len, read_message, read, text_data, text_len, text);
  if (read && len > 0)
  { if (nbytes == 0)
    { THROW("Unexpected end of file encountered.");
    }
  }
  return nbytes;
}
```
That THROW is used only in read mode and won't be ever executed as `bin_text_read_write` calls `bin_read`. And `bin_read` has this one inside:
```
  size_t ret = bin_read_fixed(i,(char*)&obj_len,sizeof(obj_len),"");
  if (obj_len > len || ret < sizeof(uint32_t))
    THROW("bad model format!");
  ret += bin_read_fixed(i,data,obj_len,read_message);
```
So "bad model format!" will be always thrown before "Unexpected end of file encountered." due to ` ret < sizeof(uint32_t)` unless `bin_read_fixed` is able to return `-4`.


*I would also grumble about confusing inconsistency in `bin_text_write` and `bin_write`. They store 0-length objects in a different way. 1st one write nothing while 2nd one write 4 zero bytes (length). 2nd approach would be better for saving empty strings (no need to store trailing 0 byte), but changing this may make models binary incompatible.*

JohnLangford:
Your logic seems sound.  @lhoang29 should we get rid of the _validate() functions?

JohnLangford:
The complaint about bin_text_write() vs bin_write() also seems sound.  This is fixable though---bin_write() is only used in bin_text_write() in the entire codebase.  Can you send me a pull request?

lhoang29:
there are 2 versions of the _validate function: `bin_text_read_write_validated` and `bin_text_read_write_fixed_validated`. We can remove the first one but the latter is still needed because it doesn't follow the same code path above.

JohnLangford:
Cool.  Please send me a patch removing the first.

-John

PR #877, opened 2015-12-06
ftrl tests 105 & 106 change hyper params for better convergence

arielf:


arielf:
appveyor failure is unrelated to this change.
The change looks good.

PR #879, opened 2015-12-07
Changed syntax, added plotting, small usability improvements in vw-hyperopt

kirillfish:
This is a follow-up to my [previous pull request](https://github.com/JohnLangford/vowpal_wabbit/pull/867). 

@arielf I worked through your suggestions, but your comment for some reason disappeared from that conversation :disappointed: Maybe it's because I re-forked the repository instead of re-syncing it.

Anyways, I made the following improvements:
* New syntax for hyperopt search space definition: one `~` symbol instead of square brackets. Now a full command looks like this:
```bash
./vw-hyperopt.py --train train_set --holdout holdout_set --max_evals 200 --outer_loss_function logistic --vw_space '--algorithms=ftrl,sgd --l2=1e-8..1e-4~LO --l1=1e-8..1e-4~LO -l=0.01..10~L --ftrl_alpha=5e-5..5e-1~L --ftrl_beta=0.01..1 --passes=1..10~I -q=SE+SZ+DR,SE~O --ignore=T~O --loss_function=logistic -b=29'
```
* Added diagnostic plotting utility. Now after searching is completed, 2 plots are created: (i)  `--outer_loss_function` value vs iteration number and (ii) the loss of the best configuration discovered so far vs iteration number. The tool used for visualization is matplotlib + seaborn. Plots look like this (this one is not from real data):
hyperopt_tracking

Here there are 2 lines (orange one denotes just random sampling from the initial distribution), but actually there will be only 1 line that denotes `tpe` algorithm. I can attach also some pictures from real data, or you can try it out by yourself.

* Results of all trials are now stored into `trials.json` so they can be analyzed afterwards.
* Added shebang line and wrote chmod 755 hyperopt.py, so that one can run it as a shell script.
* Made notifications lowercase.
* Removed default hyperparameter values and ranges from `HyperoptSpaceConstructor`
* The iteration number is now displayed at the beginning of each step.

### DIFF
--- a/utl/vw-hyperopt.py
+++ b/utl/vw-hyperopt.py
@@ -2,7 +2,7 @@
 # coding: utf-8
 
 """
-Hyperparameter optimization for Vowpal Wabbit via hyperopt
+Github version of hyperparameter optimization for Vowpal Wabbit via hyperopt
 """
 
 __author__ = 'kurtosis'
@@ -131,16 +131,21 @@ class HyperoptSpaceConstructor(object):
                 break
 
         self.space = {algo: {'type': algo, 'argument': self.algorithm_metadata[algo]['arg']} for algo in algorithms}
+        #print '\n\nALGOS:', self.space, '\n\n'
         for algo in algorithms:
             for arg in line:
                 arg, value = arg.split('=')
                 if arg == '--algorithms':
                     continue
-                distrib = self._process_vw_argument(arg, value, algo)
                 if arg not in self.algorithm_metadata[algo]['prohibited_flags']:
+                    #print "\n NOT PROHIBITED: ", arg, self.algorithm_metadata[algo]['prohibited_flags']
+                    distrib = self._process_vw_argument(arg, value, algo)
                     self.space[algo][arg] = distrib
+                else:
+                    pass #print "\nPROHIBITED: ", arg, self.algorithm_metadata[algo]['prohibited_flags']
+        #print self.space
         self.space = hp.choice('algorithm', self.space.values())
-
+        #print self.space
 
 class HyperOptimizer(object):
     def __init__(self, train_set, holdout_set, command, max_evals=100,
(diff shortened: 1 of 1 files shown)


JohnLangford:
Are the plotting tools required or optional?  The first seems undesirable because fewer people will be able to use it.

kirillfish:
@JohnLangford Thanks! Now it is optional.

arielf:
Kyryll,

Thanks for the additions.
Question: does it support the empty option among alternatives?
E.g: algorithm `--ftrl` vs the default (3-way enhanced SGD):  `--algorithms=ftrl,`

kirillfish:
@arielf, now actually `algorithms=sgd` means the default 3-way enhanced SGD (it transforms to no flags in shell command). So, the semantics are the following:

`algorithms=sgd` or ` ` (no argument option at all) means the default algorithm;
`algorithms=ftrl` means only FTRL;
`algorithms=sgd,ftrl` or `algorithms=ftrl,sgd` means FTRL and the default algorithm.

To get vanilla SGD with no enhancements is not easy now. Unfortunately, I still haven't added binary flags, which take no values, such as `--adaptive=~B`. Now one could, in principle, write `--adaptive=\ ~O`, but this is not intuitive . I will put it right. Also, now `--adaptive=\ ~O` will fail for sampled configurations with `--ftrl` (I should add `--adaptive`, `--normalized`, `--invariant` to the list of prohibited options for `--ftrl` for this to work correctly).

kirillfish:
Now commited a change that allows to treat `--algorithms` like any other option. So, `--algorithms=ftrl~O` is now valid and equivalent to `--algorithms=ftrl,sgd`

JohnLangford:
Looks good, thanks.  All, the vw-hyperopt tool looks pretty handy to me.

Kirill, you might consider using the '--args ' command. Essentially, you 
can pass a file with one set of commandline arguments per line, and VW 
will parallelize the learning (but not the parsing) across the lines.

-John

PR #881, opened 2015-12-08
removed use of std::to_string

lhoang29:
Removing all references to std::to_string to avoid build issues with other platforms (such as Cygwin). I didn't hear back from cygwin on whether there's a workaround for the "to_string not declared" error so am rewriting it using sprintf_s, which is actually faster. Linux tests ran ok, and cygwin no longer complains about this.


lhoang29:
This also includes a fix for Windows build break.

JohnLangford:
Merged.  The windows build is however still breaking?

lhoang29:
It should be fixed as of this PR. There were some static_cast calls that didn't specify types.

PR #883, opened 2015-12-09
multilabel hamming loss bugfix

pmineiro:
Bugfix for failing to penalize for false positives.


JohnLangford:
Why is travis barfing?

-John

pmineiro:
I don't know I haven't looked (is some test comparing with exact output?)

p

JohnLangford:
Unclear.  Merged.  I'll deal with fallout.

PR #884, opened 2015-12-11
Minor build fixes

willnewton:
This is a series of patches that fixes build issues with building on 32bit platforms and out of tree builds.

### DIFF
--- a/explore/static/utility.h
+++ b/explore/static/utility.h
@@ -228,7 +228,7 @@ public:
 
   class StringUtils
   { public:
-    static inline string to_string(size_t& n)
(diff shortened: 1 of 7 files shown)


explore/static/utility.h
```diff
@@ -228,7 +228,7 @@ MWT_NAMESPACE
 
   class StringUtils
   { public:
-    static inline string to_string(size_t& n)
+    static inline string to_string(u64& n)
```
JohnLangford: Why is u64 preferred here?

JohnLangford: Ah, got it from the individual checkin.

PR #886, opened 2015-12-16
Sensitivity fix

tzukuoh:
Fixed bugs in sensitivity calculations so that states don't get modified when stateless is true.


JohnLangford:
Merged in, thanks.

All, this is an important bugfix for the active learning/confidence code.

-John

PR #833, opened 2015-10-20
preconditioner update in bfgs and reverting weight computation for incremental BFGS

tzukuoh:
Updated bfgs.cc 
(1) Preconditioner is recomputed every pass
(2) Added reverting weight computation in test-only mode with input_feature_regularizer. The formula are based on approximate reverting weight for incremental BFGS.
(3) Fixed a bug in output_feature_regularizer computation.

PR #887, opened 2015-12-16
preconditioner update in bfgs and reverting weight computation for incremental BFGS

tzukuoh:
The previous pull request #833 is replaced by this one, which is in a separate branch.


JohnLangford:
Ok, I think this patch is about ready to go in.  Can you sync to the master branch?

tzukuoh:
ok, it's synced to the master branch. I took out the reverting weight computation because it doesn't seem super useful and uses ec.revert_weight which has been removed.

PR #889, opened 2015-12-19
Refactoring exploration library as subtree

lhoang29:
Removed the existing code for exploration library and added a subtree pointing to that under MultiWorldTesting account.

In the future, this subtree can be updated as follows (but I will update if necessary):

First add a remote pointing to the mwt repo:
`git remote add explore-cpp https://github.com/multiworldtesting/explore-cpp.git`

Then use `subtree` command to pull from master branch of explore-cpp remote, and place it under explore/cpp directory:
`git subtree pull --prefix=explore/cpp/ explore-cpp master`
or alternatively:
`git pull -X subtree=explore/cpp explore-cpp master`

Note that the `subtree` command may not be installed by default.


JohnLangford:
This is not merging at present.  Can you update?

lhoang29:
should work now

lhoang29:
oops, don't merge this yet, I realized some C# unit tests are failing so I'll need to merge in the C# library as well. Will let you know once it's ready

lhoang29:
ok, good to go

JohnLangford:
Why are ExploreClock.cs and ExploreOnlySample.cs in cs/test rather than explore/csharp?

lhoang29:
Those were just remnants from the previous code, removed.

JohnLangford:
It seems like we should have one repository for all versions of the exploration library, with blackbox testing setup and automated.  

Instead, we apparently have a different repository for every language and for the test suite with no automation?  Is there a good reason for this?  (I'm hesitating in taking a dependency on a subtree which is not properly tested).

Also, why csharp and c++?   If you want to be minimal, c++ should suffiice, and if you want to be complete, why leave out java?

lhoang29:
Right, unlike in VW where the various language interface are simply wrappers around a C++ core, the different versions of exploration library are independent of each other so the main reason is to keep them modular and lightweight so that users only have to work with one. This follows the same pattern done for other popular libraries such as  https://github.com/reactjs and https://github.com/Azure?page=2.

Testing: every repository is covered by internal unit tests and external black box tests. The latter is set up so it can be more easily deployed to the build server. However, there's no build automation yet like you said, the current code is locally tested. I was thinking of boarding Jenkins/Docker for this if they let us. On the other hand, subtree does not automatically sync with its remote so we won't have to worry about unwanted changes creeping in.

I initially wanted to include C++ only but there was dependency on C# version as well. However, the dependency was just sample code so I'll remove it (unless you prefer otherwise).

JohnLangford:
All the tests should be lightweight, so a travis approach seems best for 
testing?  That way we don't have do deal with server support.

Removing the C# stuff seems like a good idea.  It also seems like a good 
idea to include a README in root directory for each library containing 
pointers to the other libraries.

-John

lhoang29:
That's a good suggestion, I'll add some README. And yes, maybe Travis + AppVeyor would be enough, am going to look at it this week.

JohnLangford:
Try syncing with master again?

lhoang29:
yup

lhoang29:
Black box tests are now automated in the build server: https://github.com/multiworldtesting/explore-csharp, and https://github.com/multiworldtesting/explore-cpp

JohnLangford:
Excellent!

-John

JohnLangford:
What about Java?

-John

lhoang29:
Yeah I wanted to do it but am not too familiar with the build env enough so that will take some more time

Issue #895, opened 2015-12-22
clang format warnings


arielf:
These warnings started to appear after the recent changes when building with clang 3.x on Ubuntu:

```
search.cc:435:35: warning: format specifies type 'unsigned long long' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
    fprintf(stderr, " %15llusec", num_sec);
                      ~~~~~~      ^~~~~~~
                      %15zu

search.cc:461:47: warning: format specifies type 'unsigned long long' but the argument has type 'unsigned long' [-Wformat]
    int num = sprintf(a.feature, "fid=%llu_", (idx & mask) >> ss);
                                      ~~~~    ^~~~~~~~~~~~~~~~~~
                                      %lu

gd.cc:581:60: warning: format specifies type 'unsigned long long' but the argument has type 'unsigned long' [-Wformat]
        text_len = sprintf_s(buff, buf_size, ":%llu:%f\n", it->second, *v);
                                               ~~~~ ^~~~~~~~~~
                                               %lu

log_multi.cc:373:56: warning: format specifies type 'unsigned long long' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
    text_len = sprintf(buff, "max_predictors = %llud ",b.max_predictors);
                                               ~~~~ ^~~~~~~~~~~~~~~~
                                               %zu
log_multi.cc:376:57: warning: format specifies type 'unsigned long long' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
    text_len = sprintf(buff, "predictors_used = %llud ",b.predictors_used);
                                                ~~~~ ^~~~~~~~~~~~~~~~~
                                                %zu
```

arielf:
It is easy for me to fix but will a fix break windows because we define types differently in Windows?

JohnLangford:
Dealt with, thanks.

PR #896, opened 2015-12-22
Fix clang format %llu vs arg (unsigned long or size_t) type warnings:…

arielf:
… #895

Affected files:
    - vowpalwabbit/gd.cc
    - vowpalwabbit/log_multi.cc
    - vowpalwabbit/search.cc

### DIFF
--- a/vowpalwabbit/gd.cc
+++ b/vowpalwabbit/gd.cc
@@ -578,7 +578,7 @@ void save_load_regressor(vw& all, io_buf& model_file, bool read, bool text)
         brw = bin_text_write_fixed(model_file, (char*)it->first.c_str(), sizeof(*it->first.c_str()),
                                    buff, text_len, true);
 
-        text_len = sprintf_s(buff, buf_size, ":%llu:%f\n", it->second, *v);
+        text_len = sprintf_s(buff, buf_size, ":%lu:%f\n", it->second, *v);
         brw += bin_text_write_fixed(model_file, (char *)v, sizeof(*v),
                                     buff, text_len, true);
       }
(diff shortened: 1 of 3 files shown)


JohnLangford:
I'm planning to fix this a different way.  In particular, I believe stringstream is more standardized across platforms.

JohnLangford:
Ok, this is (hopefully) permanently fixed in the master now.   In 
essence, I replaced a massive number of printf's with stringstream's.  
All the tests are checking out, but I'd like Alex to take a look at this 
line:

https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parse_regressor.cc#L245

and tell me if it is correct.

-John

trufanov-nok:
If I recall right the `%.*s` is a [string of a known size, and its size is passed via parameter too](http://stackoverflow.com/questions/13289315/sprintf-usage-with-s).  

And `<< all.interactions[i].begin;` expecting a simple null terminated string. 
So there are two problems:  
1. Technically, there might be interactions with null namespace.
2. The all.interactions[i].end might not point to null. I think even when memory is allocated with calloc there might be a cases when you can get non-null bytes after real end of array. 

I think you should convert content of v_array to std::string with constructor `string (const char* s, size_t n);` ([string can contain embedded nulls](http://stackoverflow.com/questions/2845769/can-a-stdstring-contain-embedded-nulls)) and then write string to stream. I [expect it](http://stackoverflow.com/questions/19575173/how-to-copy-a-string-with-multiple-null-characters-to-a-stringstream) to be correctly written to stream in this case.

JohnLangford:
I guess this is fixed now but for the record the correct specifier for
size_t types is %z which should work for all sizes of size_t. I'm not 100%
sure which version of MSVC is required to support it but it should be part
of C++11 support (and C99).

JohnLangford:
Thanks.

This is the elegant fix, I think.  Yes?

https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parse_regressor.cc#L247

-John

trufanov-nok:
Yes, this is a better solution.

PR #899, opened 2015-12-30
enabled setting of OPTIM_FLAGS on the command line, to enable debuggi…

danmelamed:
…ng without editing the Makefile


JohnLangford:
Looks handy, thanks.

PR #900, opened 2016-01-03
Bugfix of compilation error

wbt:
Was previously throwing a compilation error that stderr was not defined in the call to fputs(...) on what's now line 15 and was [introduced in](https://github.com/wbt/vowpal_wabbit/blob/d1697f80fe2f174b7acd55ec64f2415632bc97df/vowpalwabbit/memory.h) version d1697f8.


JohnLangford:
Thanks, merged.

-John

PR #901, opened 2016-01-05
Added a Doxyfile for use with doxygen.

danmelamed:
This can help new VW users to understand the code.


JohnLangford:
How do you envision doxygen being used with the wiki?

danmelamed:
Since code docs are only useful to the extent that they are up to date, they should ideally be updated automatically and associated with branches, rather than with static and manually updated info like the wiki.  Eventually, we can make doxygen run as part of the build process (at least some of the time) on a build server, and make its output browsable directly on the build server.  For now, devs can run it (`make doc`) and browse its output locally on their own machine.

JohnLangford:
My understanding is that this is presently a noop, because no one is taking advantage of doxygen?  I'm hesitant to add things in until it actually does something useful.

danmelamed:
Why do you think that nobody is taking advantage of doxygen?  I certainly am.

JohnLangford:
I don't see any documentation in the path.  How does doxygen work?

-John

danmelamed:
doxygen creates the documentation from the source code.  As with other resources that are generated by `make`, these docs are not checked into the repo.

To use, install doxygen and dot/graphviz on your machine, checkout this branch, type `make doc`, and then point your browser at doc/html/index.html .

Dan

JohnLangford:
I see.  Hyperlinked code is pretty handy.  How do you make the start page for doxygen point to the wiki as an additional source of information?

JohnLangford:
Maybe also add the Doxyfile to the doc directory?

danmelamed:
both done

JohnLangford:
Merged.  I'd like to setup hunch.net/~vw to have this also.  I'll look into it.

JohnLangford:
Merged, thanks.  Also, now on http://hunch.net/~vw

-John

PR #902, opened 2016-01-05
fixed windows build

lhoang29:
Fixing build for the C++/CLI layer to work with new feature representation.


JohnLangford:
Why is AppVeyor still barfing?

lhoang29:
some unit tests are failing. I've been looking at it

lhoang29:
fixed now

JohnLangford:
Can you change "features.h" to be "feature_group.h"?

Also, is it possible to compile the cli stuff in linux?  (And automate that via Makefile? so that these sorts of changes are more automatic.)

JohnLangford:
Merged.  I did the features.h -> feature_group.h change, but would still like to know the answer to the second question.

lhoang29:
oh, I just checked in the rename as well, but maybe auto merge took care of it. 

As for the second question, I don't believe so. C++/CLI is a .NET specific language which requires the .NET CLR runtime and a compatible compiler. It seems there is a distribution of .NET on Linux (although I'm not sure what support it has). However the compiler is non-existent as far as I know.

JohnLangford:
That's a little bit frustrating---it's very close to C++.

-John

PR #904, opened 2016-01-07
java NativeUtils: lsb_release to get linux distro and version

### LINKED ISSUE #894: Docker Java Native Library `java.lang.UnsatisfiedLinkError: no vw_jni in java.library.path`
Hi,

We have a java project that uses VW, and we love it!  However, when the java code runs in an Ubuntu 14.04 docker container, it can't load the native bindings.   I figured out why this happens and will propose a solution at the bottom of this issue description.

**The Error:**  The following traceback occurs when we run our java program inside a docker container that inherits from Ubuntu:14.04.
```
[info] java.lang.UnsatisfiedLinkError: no vw_jni in java.library.path
[info]  at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1864)
[info]  at java.lang.Runtime.loadLibrary0(Runtime.java:870)
[info]  at java.lang.System.loadLibrary(System.java:1122)
[info]  at vw.jni.NativeUtils.loadOSDependentLibrary(NativeUtils.java:116)
[info]  at vw.learner.VWLearners.loadNativeLibrary(VWLearners.java:85)
[info]  at vw.learner.VWLearners.initializeVWJni(VWLearners.java:73)
[info]  at vw.learner.VWLearners.create(VWLearners.java:44)
```

**Why this happens:**
Looking at VW source code for `java/src/main/java/vw/jni/NativeUtils.java`, a function named `getDistroName` directly queries `/proc/version` ([see source](https://github.com/JohnLangford/vowpal_wabbit/blob/9e5831a72d5b0a124c845dcaec75879f498b355f/java/src/main/java/vw/jni/NativeUtils.java#L29)).  In docker, this returns the host's kernel version, which may not be Ubuntu.  (If you use docker-machine, /proc/version returns information about the docker-machine VM that the container runs in).  What we actually want it to do, as far as I understand, is query the container OS version, which is Ubuntu 14.04.

**Possible Solution:**
For better compatibility with VW in docker containers, I suggest we don't query /proc/version.  What if, instead, we replace the linux-specific contents of `getDistroName` with `lsb_release -i`?  Since we already use `lsb_release -r` for `getLinuxVersion`, this seems to make quite a bit of sense to me, and I believe it will work in a docker container.

I'd love to get your feedback and would be happy to submit a PR if necessary.

adgaudio:
closes #894 
related:  #807 #780 #822

a change to java NativeUtils class to attempt more standardized access to linux distro and version
fixes a couple bugs in build.sh
updates readme

@jmorra Build.sh passes up to the last 3 lines, where I have issues (unrelated to this PR) compiling the C code on my mac.  After some bug fixes to build.sh, VW appears to compile correctly in the linux docker images.   Could you check out this branch and try to compile it on your mac?  Or if you happen to know what's wrong in the traceback below, that would be helpful.  I wouldn't think this change should prevent VW java from working on mac, as the changes relate to linux issues, but it probably makes sense to check.

---

Some error msgs (unrelated to this PR) that I get when trying to manually compile VW on my mac (10.11.2):
```
./../explore/cpp/utility.h:241:24: note: candidate function not viable: no known conversion from
      'size_t' (aka 'unsigned long') to 'u32 &' (aka 'unsigned int &') for 1st argument
  static inline string to_string(u32& n)
                       ^
cbify.cc:277:78: error: no matching function for call to 'to_string'
  ...= data.mwt_explorer->Choose_Action(*data.generic_explorer, StringUtils::to_string(ec.exa...
              ```

### DIFF
--- a/java/src/main/bin/build.sh
+++ b/java/src/main/bin/build.sh
@@ -29,7 +29,7 @@ $make_base
 mv java/target/vw_jni.lib java/target/vw_jni.Ubuntu.14.amd64.lib"
 
 early_red_hat="yum update -q -y
-yum install -q -y wget which zlib-devel java-1.7.0-openjdk-devel perl;
+yum install -q -y wget which zlib-devel java-1.7.0-openjdk-devel perl redhat-lsb-core;
 cd /etc/yum.repos.d;
 wget http://people.centos.org/tru/devtools-2/devtools-2.repo;
 yum clean all;
@@ -55,7 +55,7 @@ make -f Makefile.permissive;
 rm -f Makefile.permissive;
 mv java/target/vw_jni.lib java/target/vw_jni.Red_Hat.6.amd64.lib"
 
-red_hat_7="yum install -q -y gcc-c++ make boost-devel zlib-devel java-1.7.0-openjdk-devel perl;
+red_hat_7="yum install -q -y gcc-c++ make boost-devel zlib-devel java-1.7.0-openjdk-devel perl redhat-lsb-core;
 export JAVA_HOME=/usr/lib/jvm/java-1.7.0-openjdk;
 $make_base
 mv java/target/vw_jni.lib java/target/vw_jni.Red_Hat.7.amd64.lib"
--- a/java/src/main/bin/build.sh
+++ b/java/src/main/bin/build.sh
@@ -11,7 +11,7 @@ make clean;
 make vw java;"
 
 ubuntu_base="apt-get update -qq;
-apt-get install -qq software-properties-common g++ make libboost-all-dev default-jdk;"
+apt-get install -qq software-properties-common g++ make libboost-program-options-dev default-jdk;"
 
 ubuntu_12="$ubuntu_base
 export JAVA_HOME=/usr/lib/jvm/java-6-openjdk-amd64;
@@ -55,7 +55,8 @@ make -f Makefile.permissive vw java;
 rm -f Makefile.permissive;
 mv java/target/vw_jni.lib java/target/vw_jni.Red_Hat.6.amd64.lib"
 
-red_hat_7="yum install -q -y gcc-c++ make boost-devel zlib-devel java-1.7.0-openjdk-devel perl clang redhat-lsb-core;
+red_hat_7="yum update -q -y;
+yum install -q -y gcc-c++ make boost-devel zlib-devel java-1.7.0-openjdk-devel perl clang redhat-lsb-core;
 export JAVA_HOME=/usr/lib/jvm/java-1.7.0-openjdk;
 $make_base
 mv java/target/vw_jni.lib java/target/vw_jni.Red_Hat.7.amd64.lib"
@@ -145,6 +146,7 @@ install_brew_app "docker-machine"
 install_brew_app "docker"
 
 docker-machine create --driver virtualbox default
+docker-machine start default
 eval "$(docker-machine env default)"
 
 set -e
@@ -165,5 +167,5 @@ run_docker "centos:6" "$red_hat_6"
 run_docker "centos:7" "$red_hat_7"
 
 make clean
-make
+make vw java
 mv java/target/vw_jni.lib java/target/vw_jni.$(uname -s).$(uname -m).lib
--- a/java/src/main/bin/build.sh
+++ b/java/src/main/bin/build.sh
@@ -8,7 +8,7 @@ __brew_not_installed=2
 
 make_base="cd /vowpal_wabbit;
 make clean;
-make;"
+make vw java;"
 
 ubuntu_base="apt-get update -qq;
 apt-get install -qq software-properties-common g++ make libboost-all-dev default-jdk;"
@@ -44,18 +44,18 @@ yum install -q -y make epel-release;
 yum install -q -y boost141-devel;
 make clean;
 cat Makefile.permissive | sed 's/BOOST_LIBRARY = -L \/usr\/lib/BOOST_LIBRARY = -L \/usr\/lib64\/boost141/g' | sed 's/BOOST_INCLUDE = -I \/usr\/include/BOOST_INCLUDE = -I \/usr\/include\/boost141/g' > Makefile.permissive.boost141;
-make -f Makefile.permissive.boost141;
+make -f Makefile.permissive.boost141 vw java;
 rm -f Makefile.permissive Makefile.permissive.boost141;
 mv java/target/vw_jni.lib java/target/vw_jni.Red_Hat.5.amd64.lib"
 
 red_hat_6="$early_red_hat
 yum install -q -y boost-devel;
 make clean;
-make -f Makefile.permissive;
+make -f Makefile.permissive vw java;
 rm -f Makefile.permissive;
 mv java/target/vw_jni.lib java/target/vw_jni.Red_Hat.6.amd64.lib"
 
-red_hat_7="yum install -q -y gcc-c++ make boost-devel zlib-devel java-1.7.0-openjdk-devel perl redhat-lsb-core;
+red_hat_7="yum install -q -y gcc-c++ make boost-devel zlib-devel java-1.7.0-openjdk-devel perl clang redhat-lsb-core;
 export JAVA_HOME=/usr/lib/jvm/java-1.7.0-openjdk;
 $make_base
 mv java/target/vw_jni.lib java/target/vw_jni.Red_Hat.7.amd64.lib"
@@ -131,7 +131,7 @@ install_cask_app() {
 run_docker() {
   local machine=$1
   local script=$2
-  docker run -v $(pwd):/vowpal_wabbit $machine /bin/bash -c "$script"
+  docker run --rm -v $(pwd):/vowpal_wabbit $machine /bin/bash -c "$script"
 }
 
 # =============================================================================
@@ -147,6 +147,9 @@ install_brew_app "docker"
 docker-machine create --driver virtualbox default
 eval "$(docker-machine env default)"
 
+set -e
+set -u
+set -x
 run_docker "ubuntu:12.04" "$ubuntu_12"
 run_docker "32bit/ubuntu:14.04" "$ubuntu_14_32"
 run_docker "ubuntu:14.04" "$ubuntu_14"
--- a/java/src/main/bin/build.sh
+++ b/java/src/main/bin/build.sh
@@ -11,7 +11,7 @@ make clean;
 make vw java;"
 
 ubuntu_base="apt-get update -qq;
-apt-get install -qq software-properties-common g++ make libboost-program-options-dev default-jdk;"
+apt-get install -qq software-properties-common g++ make libboost-program-options-dev zlib1g-dev default-jdk;"
 
 ubuntu_12="$ubuntu_base
 export JAVA_HOME=/usr/lib/jvm/java-6-openjdk-amd64;
--- a/java/src/main/java/vw/jni/NativeUtils.java
+++ b/java/src/main/java/vw/jni/NativeUtils.java
@@ -20,39 +20,17 @@ public class NativeUtils {
     }
 
     /**
-     * This will read from /proc/version to attempt to find the Linux distribution.
-     * @return The Linux distribution or null if the version cannot be found.
-     * @throws IOException If an I/O error occurs
+     * lsb_args = "-i", regexp = "Distributor ID: *(.*)$"
      */
-    public static String getDistroName() throws IOException {
-        Pattern distroRegex = Pattern.compile("[^(]+\\([^(]+\\([^(]+\\(([A-Za-z\\s]+).*");
-        BufferedReader reader = new BufferedReader(new FileReader("/proc/version"));
-        String distro;
-        try {
-            Matcher line = distroRegex.matcher(reader.readLine());
-            distro = line.matches() ? line.group(1) : null;
-        }
-        finally {
-            reader.close();
-        }
-        return distro;
-    }
-
-    /**
-     * This will attempt to find the Linux version by making use of {@code lsb_release -r}
-     * @return The Linux version or null if the version cannot be determined.
-     * @throws IOException If an I/O error occurs
-     */
-    public static String getLinuxVersion() throws IOException {
+    public static String lsb_release(String lsb_args, Pattern regexp) throws IOException {
         BufferedReader reader = null;
         try {
-            Process process = Runtime.getRuntime().exec("lsb_release -r");
+            Process process = Runtime.getRuntime().exec("lsb_release " + lsb_args);
             reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
             String line;
-            Pattern releasePattern = Pattern.compile("Release:\\s*(\\d+).*");
             Matcher matcher;
             while ((line = reader.readLine()) != null) {
-                matcher = releasePattern.matcher(line);
+                matcher = regexp.matcher(line);
                 if (matcher.matches()) {
                     return matcher.group(1);
                 }
@@ -62,6 +40,24 @@ public class NativeUtils {
             reader.close();
         }
         return null;
+    }
+
+    /**
+     * Attempt to find the Linux distribution ID.
+     * @return The Linux distribution or null if the version cannot be found.
+     * @throws IOException If an I/O error occurs
+     */
+    public static String getDistroName() throws IOException {
+        return lsb_release("-i", Pattern.compile("Distributor ID:\\s*(.*)\\s*$"));
+    }
+
+    /**
+     * This will attempt to find the Linux version by making use of {@code lsb_release -r}
+     * @return The Linux version or null if the version cannot be determined.
+     * @throws IOException If an I/O error occurs
+     */
+    public static String getLinuxVersion() throws IOException {
+        return lsb_release("-r", Pattern.compile("Release:\\s*(.*)\\s*$"));
     }
 
     /**
@@ -88,7 +84,7 @@ public class NativeUtils {
             if (version == null) {
                 throw new UnsupportedOperationException("Cannot determine linux version");
             }
-            return distro.trim().replaceAll(" ", "_") + "." + version;
+            return String.format("%s.%s", distro, version.split("\\.")[0]);
         }
         throw new IllegalStateException("Unsupported operating system " + osName);
     }
--- a/java/src/main/java/vw/jni/NativeUtils.java
+++ b/java/src/main/java/vw/jni/NativeUtils.java
@@ -20,7 +20,10 @@ public class NativeUtils {
     }
 
     /**
-     * lsb_args = "-i", regexp = "Distributor ID: *(.*)$"
+     * Shell call to lsb_release to get OS info
+     * @param lsb_args parameters to lsb_release.  EX: "-i" or "-a"
+     * @param regexp a regular expression pattern used to parse the response from lsb_release
+     *   For example:  lsb_args = "-i", regexp = "Distributor ID: *(.*)$"
      */
     public static String lsbRelease(String lsb_args, Pattern regexp) throws IOException {
         BufferedReader reader = null;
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by JohnLangford** [comment]
@lhoang29 I thought we got rid of the to_string madness?

**Comment by jmorra** [comment]
I just checked this out and tried to build it and I see the same errors

```
cbify.cc:150:74: error: no matching function for call to 'to_string'
  uint32_t action = data.mwt_explorer->Choose_Action(*data.tau_explorer, StringUtils::to_string(ec.example_counter), vwc);
                                                                         ^~~~~~~~~~~~~~~~~~~~~~
./../explore/cpp/utility.h:235:24: note: candidate function not viable: no known conversion from 'size_t' (aka 'unsigned long') to 'u64 &' (aka 'unsigned long long &') for 1st argument
  static inline string to_string(u64& n)
                       ^
./../explore/cpp/utility.h:241:24: note: candidate function not viable: no known conversion from 'size_t' (aka 'unsigned long') to 'u32 &' (aka 'unsigned int &') for 1st argument
  static inline string to_string(u32& n)
```

**Comment by jmorra** [comment]
It should be noted that the same thing happens on master as well.  I think that once this is fixed on master, then master can be pulled into this PR and I can test it again.

**Comment by JohnLangford** [comment]
This is a longstanding issue related to _slightly_ different prototypes 
in libraries.   Louie had a fix but maybe it was lost?  Not sure.  Let's 
allow him to take a look.

-John

**Comment by lhoang29** [comment]
Hmm, strange. @jmorra are you compiling just the c++ core on master or the java wrapper? I'm taking a look now

**Comment by jmorra** [comment]
This is just the c++ core.

**Comment by lhoang29** [comment]
I sent a PR that could likely fix this: https://github.com/JohnLangford/vowpal_wabbit/pull/905. @jmorra and @adgaudio would one of you mind trying the patch to see if it works?

**Comment by JohnLangford** [comment]
Louie's pull request is merged into the master.  Try again?

**Comment by adgaudio** [comment]
I can confirm that `make` on osx passes for me.  `make` in all the docker containers is also passing for me!  At this point, I would guess that the PR is safe to merge.  (though someone more familiar should make this call, of course!).  Also, did anyone actually review the code changes?

**Comment by jmorra** [comment]
Would it be possible to pull current master into this PR so I can just check out this PR?

File: java/src/main/bin/build.sh [hunk 1]
```diff
@@ -44,18 +44,18 @@ yum install -q -y make epel-release;
...
 red_hat_6="$early_red_hat
 yum install -q -y boost-devel;
 make clean;
-make -f Makefile.permissive;
+make -f Makefile.permissive vw java;
 rm -f Makefile.permissive;
 mv java/target/vw_jni.lib java/target/vw_jni.Red_Hat.6.amd64.lib"
 
-red_hat_7="yum install -q -y gcc-c++ make boost-devel zlib-devel java-1.7.0-openjdk-devel perl;
+red_hat_7="yum install -q -y gcc-c++ make boost-devel zlib-devel java-1.7.0-openjdk-devel perl clang redhat-lsb-core;
```
jmorra: Why do we need clang here?

java/src/main/bin/build.sh
```diff
@@ -131,7 +131,7 @@ install_cask_app() {
 run_docker() {
   local machine=$1
   local script=$2
-  docker run -v $(pwd):/vowpal_wabbit $machine /bin/bash -c "$script"
+  docker run --rm -v $(pwd):/vowpal_wabbit $machine /bin/bash -c "$script"
```
jmorra: What does the `--rm` flag mean?

java/src/main/java/vw/jni/NativeUtils.java
```diff
@@ -20,39 +20,17 @@ private NativeUtils() {
...
-    }
-
-    /**
-     * This will attempt to find the Linux version by making use of {@code lsb_release -r}
-     * @return The Linux version or null if the version cannot be determined.
-     * @throws IOException If an I/O error occurs
+     * lsb_args = "-i", regexp = "Distributor ID: *(.*)$"
      */
-    public static String getLinuxVersion() throws IOException {
+    public static String lsb_release(String lsb_args, Pattern regexp) throws IOException {
```
jmorra: In Java, method names should be camel cased.  Please rename to `lsbRelease`.

java/src/main/java/vw/jni/NativeUtils.java
```diff
@@ -88,7 +84,7 @@ else if (osName.toLowerCase().contains("linux")) {
             if (version == null) {
```
jmorra: I think we should add a check for the period as well.  So the whole thing should read

```
int separatorIndex = version.indexOf(".")
if (separatorIndex < 0) {
    throw new UnsupportedOperationException("Cannot find separator of . in linux version of " + version);
}
return distro + version.subString(0, separatorIndex + 1);

**Comment by jmorra** [comment]
I've added some comments to consider.

File: java/src/main/bin/build.sh (hunk 1 above)
Comment (adgaudio) [inline java/src/main/bin/build.sh]: We need clang because otherwise the redhat 7 build otherwise fails.  Maybe it wasn't caught before because set -e and set -u weren't set in build.sh?  Or am I misunderstanding something?

File: java/src/main/bin/build.sh (hunk 2 above)
Comment (adgaudio) [inline java/src/main/bin/build.sh]: ```  --rm=false                      Automatically remove the container when it exits```

File: java/src/main/java/vw/jni/NativeUtils.java (hunk 3 above)
Comment (adgaudio) [inline java/src/main/java/vw/jni/NativeUtils.java]: ah thank you!

File: java/src/main/java/vw/jni/NativeUtils.java (hunk 4 above)
Comment (adgaudio) [inline java/src/main/java/vw/jni/NativeUtils.java]: some linux distros don't have a period in the version, such as debian:jessie.  If we throw an error, we won't be able to one day use this code in Debian.  Are you sure you want to add it?

File: java/src/main/bin/build.sh (hunk 1 above)
Comment (jmorra) [inline java/src/main/bin/build.sh]: I'll check it out when I build it.

File: java/src/main/java/vw/jni/NativeUtils.java (hunk 4 above)
Comment (jmorra) [inline java/src/main/java/vw/jni/NativeUtils.java]: Hmm, ok, in that case let's remove the error and just add the following

```
return distro + "." + version.split("\\.")[0];
```

In general you don't see `String.format(...)` much in Java code.

**Comment by adgaudio** [comment]
Thank you!  I made some comments to your comments and renamed lsb_release to lsbRelease.  

I'm unsure whether I should change anything related the comments you made on the below topics, but I posted questions/comments as relevant.
  - redhat 7 clang
  - docker --rm flag
  - version containing a period.

**Comment by adgaudio** [comment]
Thanks for the java tips.  Only thing remaining now seems to be the redhat 7 clang thing.  I'm assume you'll take a look at that one and I will watch for new comments.  Thanks again!

**Comment by jmorra** [comment]
The code here seems good.  Does this now have the fix from @lhoang29 merged in?  If so I'll try and build it.

**Comment by adgaudio** [comment]
If you mean #905, yes, it does.

File: java/src/main/java/vw/jni/NativeUtils.java [hunk 5]
```diff
@@ -20,39 +20,17 @@ private NativeUtils() {
...
-            distro = line.matches() ? line.group(1) : null;
-        }
-        finally {
-            reader.close();
-        }
-        return distro;
-    }
-
-    /**
-     * This will attempt to find the Linux version by making use of {@code lsb_release -r}
```
jmorra: JavaDoc requires \@param for the arguments here.

java/src/main/bin/build.sh
jmorra: For some reason this now requires the following line be run first in Red Hat 7

```
yum update -q -y;
```

java/src/main/bin/build.sh
```diff
@@ -8,7 +8,7 @@ __brew_not_installed=2
 
 make_base="cd /vowpal_wabbit;
 make clean;
-make;"
+make vw java;"
 
 ubuntu_base="apt-get update -qq;
 apt-get install -qq software-properties-common g++ make libboost-all-dev default-jdk;"
```
jmorra: This can be made slightly faster by changing `libboost-all-dev` to `libboost-program-options-dev`.

java/src/main/bin/build.sh
```diff
@@ -147,6 +147,9 @@ install_brew_app "docker"
 docker-machine create --driver virtualbox default
```
jmorra: I forgot a required line.  Please add the following

```
docker-machine start default
```

java/src/main/bin/build.sh
```diff
@@ -147,6 +147,9 @@ install_brew_app "docker"
...
 eval "$(docker-machine env default)"
 
+set -e
+set -u
+set -x
 run_docker "ubuntu:12.04" "$ubuntu_12"
 run_docker "32bit/ubuntu:14.04" "$ubuntu_14_32"
 run_docker "ubuntu:14.04" "$ubuntu_14"
```
jmorra: We can make the mac build happen faster by changing the last `make` to `make vw java`.

jmorra:
I finally did get compilation to work on my machine with the build script.  I forgot a line when I first committed this file that I mentioned in the above comments.  I also suggested some build script updates that would make the build script run faster.

There is still a problem though, that this doesn't run properly on Scientific Linux.  We make heavy use of Scientific Linux here and the RedHat binaries work fine.  Hence when on an SL box we see the following

```
[jmorra@lp-prod2 ~]$ lsb_release -i
Distributor ID:	Scientific
```

This fails to match any of the prebuilt libraries and the jar doesn't work.  The main reason I was reading from /proc/version is because on SL it returns RedHat

```
[jmorra@lp-prod2 ~]$ cat /proc/version
Linux version 2.6.32-431.el6.x86_64 (mockbuild@sl6.fnal.gov) (gcc version 4.4.7 20120313 (Red Hat 4.4.7-3) (GCC) ) #1 SMP Thu Nov 21 13:35:52 CST 2013
```

I'm all open to suggestions on how to coalesce these two ideas, but it has to work on SL for me.  Maybe reading `/proc/version` as a backup if `lsb_release` doesn't return a match.

java/src/main/bin/build.sh
adgaudio: We could also consider removing the docker-machine related lines, as it should be safe to assume that people can install docker the way they wish.  For instance, on my linux computer, I don't use docker-machine at all.  On my mac, I wouldn't want people's scripts possibly messing with my "default" machine.  I'll add the line, but make a note if you'd like me to remove all docker-machine related commands.

jmorra: That's a good point, and I think worthy of commenting.  Honestly this file is only intended to be run on a Mac.  I would love a more general script, but since I can't figure out a way to build on OSX unless the native OS is OSX we're stuck with this setup.

adgaudio:
Ok - a couple comments here:

I believe I addressed all your comments in the most recent commit except the one related to scientific linux.  

On scientific linux (I used ringo/scientific:6.5), I can see what you see, and this is the correct and intended behavior.  That is, 

```
bash-4.1# lsb_release -i
Distributor ID: Scientific
bash-4.1# lsb_release -r
Release:        6.5
```

Ironically, though, the problem you have is the same one that inspired this PR.  The way VW compiles the vw_jni.jar is OS dependent in the jar's filename.  The quick solution for you is to `mv java/target/vw_jni.lib java/target/vw_jni.Scientific.6.amd64.lib`.  If you look at the last line of the docker builds, that `mv` command is there.  I'm guessing you got this error because you compiled VW in CentOS and then mounted the prebuilt jar file into the scientific distro?  Let me know if that solves the problem.  Otherwise, I'd need more info to figure out another solution.  I currently think a /proc fallback is not the way to go.

java/src/main/java/vw/jni/NativeUtils.java
```diff
@@ -20,39 +20,20 @@ private NativeUtils() {
...
-     * @return The Linux version or null if the version cannot be determined.
-     * @throws IOException If an I/O error occurs
+     * Shell call to lsb_release to get OS info
+     * @param lsb_args parameters to lsb_release.  EX: "-i" or "-a"
+     * @param regexp a regular expression pattern used to parse the response from lsb_release
```
jmorra: JavaDoc will complain here as well because you need an @throws and @return.  Sorry for being a stickler here, but once we move to Java 8 these JavaDoc warnings become compiler errors.

jmorra:
I completely agree with everything you said above.  I wonder then if the best solution is to just add another step in the build script that just does a copy from Red_Hat to Scientific along with a big fat comment saying these two OS's are compatible (at least for the versions we current support).  The only downside to that is the resultant jar will be bigger.  If we don't want to go this way then we need to add some code in NativeUtils that states that Scientific maps to Red Hat.

One on hand the first is a cleaner solution but adds unnecessary baggage to the jar and the second makes the jar slimmer but makes the code messier.

Because I like smaller binaries I would put in a slight vote for option 2, but could be convinced by an impassioned plea for option 1.  @JohnLangford do you have an opinion on this?

adgaudio:
I wonder if we can symlink the lib file like this:  

`cd java/target ; ln -s vw_jni.Red_Hat.7.amd64.lib vw_jni.Scientific.6.amd64.lib`

@jmorra Could you try that out?  I think it would be much easier for you to try in your Scientific image than me.

Also, just figured I'd explicitly ask, so we've covered our bases a bit:   Can do away with the "Red_Hat" or "Ubuntu" in the filename?  Is there a good reason why the OS distribution is included at all?  Are we dependent on OS-specific features in the binary not in the linux kernel?

jmorra:
I'm confused on what your suggesting.  The point of the function `getOsFamily()` is to find a path to lookup a file within the jar.  Symlinking here doesn't make sense, unless you're advocating for doing that in the build script and packaging both jars which I think is redundant.  What I was advocating for is something like this.  This would allow a general framework where other identical OS families could be added later.

```
private static final Map<String, List<String>> identicalOSs = new HashMap<String, List<String>>();

    static {
        List<String> redHatAlternatives = new ArrayList<String>();
        redHatAlternatives.add("Scientific");
        identicalOSs.put("Red_Hat", redHatAlternatives);
    }
...
   /**
     * Returns the system dependent OS family.  In the case of a Linux OS it will combine
     * {@link NativeUtils#getDistroName()} and {@link NativeUtils#getLinuxVersion()}.
     * @return A list system dependent string identifying the OS.  This is a list because there are some OSs which are
     * functionally identical to the one returned by lsb_release.
     * @throws UnsupportedEncodingException If an error occurs while determining the Linux specific
     *          information.
     * @throws IOException If an I/O error occurs
     * @throws IllegalStateException If the os.name property returns an unsupported OS.
     */
    public static List<String> getOsFamilies() throws IOException {
        final String osName = System.getProperty("os.name");
        final String primaryName;
        if (osName.toLowerCase().contains("mac")) {
            primaryName = "Darwin";
        }
        else if (osName.toLowerCase().contains("linux")) {
            String distro = getDistroName();
            if (distro == null) {
                throw new UnsupportedEncodingException("Cannot determine linux distribution");
            }
            String version = getLinuxVersion();
            if (version == null) {
                throw new UnsupportedOperationException("Cannot determine linux version");
            }
            // get the major version.
            // don't expect a period because linux version might not have one
            primaryName = distro + "." + version.split("\\.")[0];
        }
        else {
            throw new IllegalStateException("Unsupported operating system " + osName);
        }

        List<String> viableFamilies = new ArrayList<String>();
        viableFamilies.add(primaryName);
        if (identicalOSs.containsKey(primaryName)) {
            viableFamilies.addAll(identicalOSs.get(primaryName));
        }
        return viableFamilies;
    }

    /**
     * Loads a library from current JAR archive by looking up platform dependent name.
     * @param path The filename inside JAR as absolute path (beginning with '/'), e.g. /package/File.ext
     * @param suffix The suffix to be appended to the name
     * @throws UnsupportedEncodingException If an error occurs while determining the Linux specific
     *          information.
     * @throws IOException If temporary file creation or read/write operation fails
     */
        List<String> osFamilies = getOsFamilies();
        String osDependentLib = null;
        for (String osFamily : osFamilies) {
            String currentOsDependentLib = path + "." + osFamily + "." + System.getProperty("os.arch") + suffix;
            if (NativeUtils.class.getResource(currentOsDependentLib) != null) {
                osDependentLib = currentOsDependentLib;
                break;
            }
        }
        if (osDependentLib != null) {
            loadLibraryFromJar(osDependentLib);
        }
        else {
            try {
                loadLibraryFromJar(path + suffix);
            }
            catch (FileNotFoundException e) {
                // If we cannot find an OS dependent library then try and load a library in a system independent fashion.
                System.loadLibrary(path.replaceFirst("/", ""));
            }
        }
    }
```

adgaudio:
Sorry - I was confusing.  What I wanted to try was to add a symlink to the jar like this:

ln -s vw_jni.Redhat.7.amd64.lib vw_jni.Scientific.6.amd64.lib
jar uf vw-jni*.jar vw_jni.Scientific.6.amd64.lib

But after looking into it, it appears that this is [not possible](http://www.linuxquestions.org/questions/linux-general-1/how-to-create-jar-files-with-symbolic-links-639381/).

A variation of above would be to fatten the jar, as I think you mentioned earlier, though it's uglier. 

```
cd /.ivy2/cache/com.github.johnlangford/vw-jni/jars/
jar xf vw-jni*.jar vw_jni.Redhat.7.amd64.lib
mv vw_jni.Redhat.7.amd64.lib vw_jni.Scientific.6.amd64.lib
uf vw-jni*.jar vw_jni.Scientific.6.amd64.lib "
```

I think your approach is probably the best way to go, but I don't maintain VW so you guys should do the decision making.

jmorra:
I think we're on the same page.  Thanks for doing the research on the symlinks in JAR files.  I assumed you couldn't, but it's good to see validation.  Unless @JohnLangford has an exception I would incorporate the changes I suggested into this PR.  After that I'll try it one more time, and it should work.  Let me know once it's updated.

JohnLangford:
Cool.  I'll wait for one last round of changes before integrating.

-John

JohnLangford:
Status?

jmorra:
At this point I think @adgaudio has to decide what to do with the changes I suggested.

adgaudio:
sorry guys - I have been working on a couple different projects.  I plan to update this PR today or tomorrow.

adgaudio:
Ok @jmorra  I added your changes.  I believe this is working, though you should test on your Scientific Linux image.

jmorra:
This worked great, with one minor change.  I'll mention it inline.  Once that's done I think this is ready to go.  @adgaudio it is probably worthwhile to do a build yourself and test on your Red Hat image to make sure everything works properly for you as well.

java/src/main/bin/build.sh
```diff
@@ -8,10 +8,10 @@ __brew_not_installed=2
 
 make_base="cd /vowpal_wabbit;
 make clean;
-make;"
+make vw java;"
 
 ubuntu_base="apt-get update -qq;
-apt-get install -qq software-properties-common g++ make libboost-all-dev default-jdk;"
+apt-get install -qq software-properties-common g++ make libboost-program-options-dev default-jdk;"
```
jmorra: I forgot to mention that once we just use `libboost-program-options-dev` we also need to include `zlib1g-dev`.  The whole line should read

```
apt-get install -qq software-properties-common g++ make libboost-program-options-dev zlib1g-dev default-jdk;"
```

adgaudio:
ok great! - I added the dependency.  build.sh passed for me.  I'm  not on a mac, so didn't test the build on mac.  At this point, I think this is good to consider merging.

jmorra:
I agree, this looks good to go.

adgaudio:
I just rebased against master - let me know if there's anything else for me to do.  Thanks @jmorra for your comments and help!  It was fun working with you and I'm happy to have contributed!

JohnLangford:
This is merged finally.  All, this is better support for autodetecting 
architecture in the Java interface.

-John

PR #907, opened 2016-01-12
more informative memory allocation failure message

danmelamed:
helpful for people who might not know the connection between the -b switch and their RAM size

### DIFF
--- a/vowpalwabbit/memory.h
+++ b/vowpalwabbit/memory.h
@@ -10,7 +10,7 @@ T* calloc_or_throw(size_t nmemb)
 
   void* data = calloc(nmemb, sizeof(T));
   if (data == nullptr)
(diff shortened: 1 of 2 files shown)


vowpalwabbit/memory.h
```diff
@@ -10,7 +10,7 @@ T* calloc_or_throw(size_t nmemb)
 
   void* data = calloc(nmemb, sizeof(T));
   if (data == nullptr)
-  { const char* msg = "internal error: memory allocation failed; dying!";
+  { const char* msg = "internal error: memory allocation failed; you probably ran out of memory -- try a lower -b setting; dying!\n";
     // use low-level function since we're already out of memory.
```
JohnLangford: This point in the code is very generic.   Can you catch the throw at the point of allocation in parse_regressor.cc instead?

danmelamed:
done

Issue #603, opened 2015-04-28
--invert_hash should report when it fails to invert a hash


sam-s:
as the output of 
```
vw -i my-model -t --invert_hash readable < /dev/null
```
indicates, `vw` can fail to invert some (or, in the above case, _all_! hashes) and the user will never know.
I suggest two modifications:
1. report the percentage of hashes that have been inverted (so that when it is not 100%, the user will know that there was a problem)
2. when not saving predictions (as in the above code), stop as soon as all the hashes have been inverted.

JohnLangford:
I don't follow (2)---how can the program know when there are no hashes left to invert?

JohnLangford:
I also don't follow (1).  How can the program know how many hashes there are to invert?  Without that, a percentage seems ill-defined.

sam-s:
depending on the `-b N` arg, the model is a vector `M` of length `2^N`
initially:
```
int num_hashes=0;
for i from 0 to 1<<N
    if M[i] != 0:
        num_hashes ++;
```
now `num_hashes` is the number of hashes to be inverted.

JohnLangford:
But what if M[hash[string]] = 0?

-John

martinpopel:
`--invert_hash` reports only the features seen in the input data (`-d`). So it is obvious that if the input is /dev/null then no features are inverted.
Note that `--invert_hash` reports also hash collisions. So the early stopping suggested in (2) would result in inconsistent results (just changing the order of input examples could increase the number of lines of invert_hash.model).
I agree that (1) may be sometimes useful, even if it will be only an approximation because of the cases M[hash[string]] = 0. Can it be implemented without substantial slow down and increased memory requirements? If no then it should be optional. I am not sure if it is worth the work.
A simple workaround is to use `--invert_hash=invert.model --readable_model=readable.model` and then `wc -l {invert,readable}.model`.

sam-s:
such features do not affect the model behavior, right?

On Wed, May 6, 2015 at 2:13 PM, John <notifications@github.com> wrote:

> But what if M[hash[string]] = 0?
>
> -John
>
> On Wed, May 6, 2015 at 1:59 PM, Sam Steingold <notifications@github.com>
> wrote:
>
> > depending on the -b N arg, the model is a vector M of length 2^N
> > initially:
> >
> > int num_hashes=0;
> > for i from 0 to 1<<N
> > if M[i] != 0:
> > num_hashes ++;
> >
[...]

PR #923, opened 2016-01-26
Audit regressor

### LINKED ISSUE #510: Another long feedback

trufanov-nok:
resolves #510 
Perhaps it may re-sol-ve #603 ?

This is an audit mode which is alternative to `--invert_hash` that I mentioned year ago.
The main idea is not to store audit data in memory till regressor values are finalized but process existing regressor file. This is needed bcs audit data may consume much more memory than regressor itself and make `--invert_hash` mode not feasible for datasets with thousands of different features. 

The proposed mode requires processing both regressor file and dataset that it was trained on (ideally). Technically it may be any dataset that contains same features that were observed during regressor training. VW reads example from dataset, generate features, stores their audit data, then check regressor values. If there is a value corresponding to feature in regressor and it's not marked as already seen, then feature's audit data is printed out to the audit file and regressor's value is marked as already seen. That allows to print each feature audit data only once. Technically marking regressor values is implemented by setting value [back to 0](https://github.com/JohnLangford/vowpal_wabbit/blob/8cd2bcbe1a14f18b673d444ae43235ec6950eaf8/vowpalwabbit/audit_regressor.cc#L92). Audit data is printed out in same format as `--invert_hash`. In case of hash collisions only the first observed feature's audit data will be printed out.

So the steps are:
1. Train your model. Save it with `-f`
2. Load your model with `-i` and specify same dataset as on step 1. Additionally `--audit_regressor <OUTPUT_FILE>` must be specified.

In comparison with `--invert_hash` the order of audit records will be different.

For example, old mode (modified test case # 30):

```
$ ../vowpalwabbit/vw -d train-sets/0001.dat -f models/mask.model --invert_hash mask.predict --l1 0.01 --examples 1
$ cat mask.predict 
Version 8.1.1
Id
Min label:0
Max label:1
bits:18
lda:0
0 ngram:
0 skip:
options:
Checksum: 3773470323
:0
Constant:116060:0.0176908
features^1028:14715:0.39391
features^1122:14809:0.152542
features^1160:14847:0.144763
features^1189:14876:0.265019
features^13:13700:0.479143
features^140:13827:0.587675
features^1530:15217:0.204277
features^156:13843:0.183806
features^1664:15351:0.291036
features^16893:30580:0.0950749
features^175:13862:0.276892
features^1865:15552:0.220691
features^188:13875:0.670377
features^229:13916:0.253146
features^230:13917:0.206882
features^234:13921:0.350215
features^236:13923:0.423537
features^238:13925:0.355351
features^24:13711:0.54651
features^24036:37723:0.0569528
features^24303:37990:0.0827366
features^2524:16211:0.114783
features^2525:16212:0.163968
features^2526:16213:0.194823
features^2536:16223:0.031825
features^2543:16230:0.125855
features^2848:16535:0.181052
features^3370:17057:0.204835
features^387:14074:0.134174
features^394:14081:0.25229
features^3960:17647:0.121147
features^433:14120:0.170035
features^434:14121:0.150606
features^438:14125:0.289733
features^465:14152:0.0827924
features^468:14155:0.22043
features^518:14205:0.185221
features^534:14221:0.201002
features^613:14300:0.267029
features^646:14333:0.215927
features^660:14347:0.262302
features^69:13756:0.410287
features^7052:20739:0.149508
features^709:14396:0.208818
features^752:14439:0.178744
features^757:14444:0.279088
features^812:14499:0.0826436
features^85:13772:0.306742
features^932:14619:0.277823
features^const:121009:1.90428
```
`--examples 1` was set to shorten results file.

New mode (using same model file):
```
$ ../vowpalwabbit/vw -d train-sets/0001.dat -i models/mask.model --audit_regressor new.reg --l1 0.01 --examples 1
$ cat new.reg
features^13:13700:0.479143
features^24:13711:0.54651
features^69:13756:0.410287
features^85:13772:0.306742
features^140:13827:0.587675
features^156:13843:0.183806
features^175:13862:0.276892
features^188:13875:0.670377
features^229:13916:0.253146
features^230:13917:0.206882
features^234:13921:0.350215
features^236:13923:0.423537
features^238:13925:0.355351
features^387:14074:0.134174
features^394:14081:0.25229
features^433:14120:0.170035
features^434:14121:0.150606
features^438:14125:0.289733
features^465:14152:0.0827924
features^468:14155:0.22043
features^518:14205:0.185221
features^534:14221:0.201002
features^613:14300:0.267029
features^646:14333:0.215927
features^660:14347:0.262302
features^709:14396:0.208818
features^752:14439:0.178744
features^757:14444:0.279088
features^812:14499:0.0826436
features^932:14619:0.277823
features^1028:14715:0.39391
features^1122:14809:0.152542
features^1160:14847:0.144763
features^1189:14876:0.265019
features^1530:15217:0.204277
features^1664:15351:0.291036
features^1865:15552:0.220691
features^2524:16211:0.114783
features^2525:16212:0.163968
features^2526:16213:0.194823
features^2536:16223:0.031825
features^2543:16230:0.125855
features^2848:16535:0.181052
features^3370:17057:0.204835
features^3960:17647:0.121147
features^7052:20739:0.149508
features^16893:30580:0.0950749
features^24036:37723:0.0569528
features^24303:37990:0.0827366
features^const:121009:1.90428
Constant:116060:0.0176908
```

The regressor's file header info is not printed out to the results file in 2nd case.

There is also one optimization that I made that require some changes in `VW` class. Now it [stores number](https://github.com/JohnLangford/vowpal_wabbit/blob/8cd2bcbe1a14f18b673d444ae43235ec6950eaf8/vowpalwabbit/gd.cc#L615) of values that were actually loaded from regressor file. That allows to [early terminate](https://github.com/JohnLangford/vowpal_wabbit/blob/8cd2bcbe1a14f18b673d444ae43235ec6950eaf8/vowpalwabbit/audit_regressor.cc#L169) audit when all regressor values are audited but dataset still has more examples (with already observed features). I found out this very useful. Sometimes all possible features could be found in first 5% of dataset. To track this progress the VW stdout was changed. For ex. for previous call it will output:
```
Regressor contains 51 values
example      values         total   
counter      audited        progress
1                   51          100%
```

Or for same train dataset without limit on examples number it will be:
```
Regressor contains 220 values
example      values         total   
counter      audited        progress
1                   21            9%
2                   46           20%
4                   73           33%
8                  105           47%
16                 140           63%
32                 166           75%
64                 196           89%
128                217           98%
155                220          100%
```
Or the same with `-q :: --cubic ::: ` and without regularization
```
Regressor contains 262144 values
example      values         total   
counter      audited        progress
1                22432            8%
2               146916           56%
4               241489           92%
8               261651           99%
14              262144          100%
```

*only 14 features needed to find all 262144 = 2^18 features which is max possible number. Most of features weren't saved in regressor due to hash collisions.*

The mode is implemented as a reduction. It is supposed to support multiclass `--oaa, --ect and --csoaa`. It [reads](https://github.com/JohnLangford/vowpal_wabbit/blob/8cd2bcbe1a14f18b673d444ae43235ec6950eaf8/vowpalwabbit/audit_regressor.cc#L190) their arguments to find correct number of classes and [iterate](https://github.com/JohnLangford/vowpal_wabbit/blob/8cd2bcbe1a14f18b673d444ae43235ec6950eaf8/vowpalwabbit/audit_regressor.cc#L105) `ec.ft_offset`.
As for other algorithms, they shall be supported till they use `INTERACTIONS::generate_interactions()` to iterate features. If some use own hardcoded features - they wont be found. This may be fixed later.


trufanov-nok:
I'm still going to add few more examples and test cases to RunTests for this PR.

JohnLangford:
A couple questions:

1) I believe there is a script in the utilities directory which does 
something like this.  How much better than the script is this?

2) What happens precisely with a hash collision?  You just print the 
first value?

-John

trufanov-nok:
> 1) I believe there is a script in the utilities directory which does something like this. How much better than the script is this?

Are you sure? Can't find it. If you could specify it I'll make a comparison.

> What happens precisely with a hash collision? You just print the first value?

Exactly. Alike the rest VW code it doesn't distinguish different features with same hash at all. This mode is intended to audit data that would blow up RAM usage in regular audit modes. So it doesn't keep any info about features that were just processed. And the regressor memory space itself is used to mark its values as already seen. So it has limited RAM overheads and thus limited possibilities to handle hash collisions as you need to remember something to compare with a new feature to detect their hash collision.

JohnLangford:
I was thinking of vw-varinfo.

-John

trufanov-nok:
Adding @arielf 
> I was thinking of vw-varinfo.

I would say that vw-varinfo operates on a higher level than proposed mode. It utilizes vw's audit mode and its algo is described as:
```
# Algorithm:
#   1)  Collect all variables and their ranges from training-set
#   2)  Train with VW to determine regressor weights
#   3)  Build a test-set with a single example including all variables
#   4)  run VW with --audit on 3) to map variable names to hash values
#       and weights.
#   5)  Output collected information about the input variables.
```
Due to 1 and 4 it should be impossible to use on very big datasets for same reason as --invert_hash - it should overflow RAM.
It also lacks `--interactions` and `--redefine` support. And I think it supports only `-q` but not `--cubic`

I wonder if it really need to generate one big example on step 4.. Could it use N examples with one feature in each? It would eliminate RAM bottleneck on step 4, but step 1 still may hit the RAM limit (as it keeps collected data in RAM while performing steps 2 and 4).

It would be also feasible to reimplement vw-varinfo on top of `--audit_regressor` but the last one can't handle (described in my prev comment) feature hash collisions as vw-varinfo needs.

It's up to you to decide if this functionality is feasible to include in VW. I'm fine to close this PR if it'll make vw too complicated.

P.S. audit_regressor is also not finalized as it still has few multiclass reduction support problems due to inconsistency of usage base_learner::inrement/base_learner::weights among them. I'm just too lazy to fix it still.

arielf:
Yes, unfortunately, all limitations listed by Alex are correct.

There is a "semi" solution though: rewrite `vw-varinfo` to run only one pass (with `--audit`) while collecting the inventory and min/max of all features, on the fly, just like `--invert_hash` does.  At the end it would simply summarize all seen features.  This way there's no need to add special handling for `--cubic`, `--redefine` or even `-q` like today.

JohnLangford:
Can you update this to the current master?  Various checks are failing and I want to stare at the code a bit to think about the complexity impact.

trufanov-nok:
Done. It still fail appveyor test for some reason but builds and tests fine for linux. That shall be enough to check the approach complexity.

eisber:
Can you schedule another AppVeyor build? This looks like an intermittent issue…

trufanov-nok:
@ eisber It's not. I hope it's a problem that I described here #926 but can't reproduce on my linux machine anymore. Will check it later.

JohnLangford:
It looks like this has good modularity, except for a few elements, which look fixable.

(1) Instead of having all.loaded_regressor_values in global_data.h and used in all base learning algorithm save/loads, can you just have it operate in the audit_regressor reduction directly?  Just have a flag in predict/learn which does a dense sweep over the values checking to see how many are nonzero.

(2) Instead of having all.audit_regressor, just have the equivalent printout happen in a finish() hook in the audit_regressor reduction?

-John

JohnLangford:
(I'm assuming you are busy, but this still looks like a good candidate for merge with fixes.)

trufanov-nok:
Yeah, I didn't forget about this and still keen to complete this PR. Just need to find a big enough chunk of free time to make required adjustments. I hope you'll get next merge candidate this weekend.

trufanov-nok:
@JohnLangford 
Ok, I pushed 2 commits to do both. Some notes:
As for 1st one - reduction technically doesn't know how other reduction stores weights in regressor memory space. That's why I wanted to infiltrate `loaded_regressor_vales++` in their save_load. Now its correctness depends on the way reduction use `leaner::increment` and `leaner::weights` values. There are some inconsistency in that usage among reductions. Still most used one will be supported.
As for second - I still need `all.vm.count("audit_regressor")` in `main.cc` to suppress print of gd's summary and its header.

JohnLangford:
Perhaps it is better to have audit_regressor much lower in the reduction stack?  The advantage is that it can report parameter values for features perturbed by a learning reduction.

trufanov-nok:
> Perhaps it is better to have audit_regressor much lower in the reduction stack? The advantage is that it can report parameter values for features perturbed by a learning reduction.

I'm afraid that reductions own output couldn't be prevented in this case.

JohnLangford:
Ok.  @eisber Do you know why appveyor fails here?

eisber:
See #951

trufanov-nok:
Ok, after two-empty-lines problem was fixed the appveyor behavior seems to be more predictable. But it seems there is still one serious problem.

As for 
> As Windows tests currently ignore the second line of RunTests (e.g. train-sets/ref/ftrl.audit_regr) the test passes if I add if (dd.loaded_regressor_values > 0)

It still fails with

```
At line:4 char:4
+    vstest.console /Platform:x64 /inIsolation vowpalwabbit\x64\Release\cs_unittes ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
   Assert.Fail failed. 0 vs 1: delta=1 > Epsilon=0.001: WeightedExampleSum
 
 
Stack Trace:
 
 
   at cs_unittest.VWTestHelper.FuzzyEqual(Double expected, Double actual, Double epsilon, String message) in c:\vowpal_wabbit\cs\unittest\VWTestHelper.cs:line 139
   at cs_unittest.VWTestHelper.AssertEqual(VowpalWabbitPerformanceStatistics expected, Vowpa
lWabbitPerformanceStatistics actual) in c:\vowpal_wabbit\cs\unittest\VWTestHelper.cs:line 151
   at cs_unittest.RunTests.CommandLine_Test116() in c:\vowpal_wabbit\cs\unittest\RunTests.cs:line 1255
```

which is totally fine for me. Even if c# tests only first file of the reference files list this change must affect vw's stderr output and cause this failure. So it's expectable.

The real problem is why `loaded_regressor_values` has 0 and cause `System.DivideByZeroException: Attempted to divide by zero.` if you don't add this `if`. It just can't be 0 because I have 
```
    if (dat.loaded_regressor_values == 0)
        THROW("regressor has no non-zero weights. Nothing to audit.");
```
In `audit_regressor.cc::init_driver()`.

I suspect that `init_driver` functions of reduction assigned to learner class with `set_init_driver` are not calledbacked. It seems that init_driver callback is rarely used. Only bfgs reduction has it and there is only one operand in it. C# performs only one of two tests in Rusntest for bfgs (15 and ignores 16). So the problem might be unnoticed.
Another proof of `init_driver` not called is that if I put `THROW("something")` in them - the C# Runtests results are unaffected (it fails with divide by zero on testcase 216). Even if I put it into bfgs reduction. On other hand if I put such throw into any reduction's setup function the c# test flaw is completely halted without much debug information
```
Starting test execution, please wait...
vstest.console : Error: The active Test Run was aborted because the execution process exited unexpectedly. The test execution process crashed while running the tests. To investigate further, open 
file:///C:/Users/appveyor/AppData/Local/CrashDumps/TE.ProcessHost.Managed.exe.3052.dmp file in Visual Studio and choose "Debug in mixed mode".
At line:4 char:4
+    vstest.console /Platform:x64 /inIsolation vowpalwabbit\x64\Release\cs_unittes ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (Error: The acti...in mixed mode".:String) [], RemoteException
 
 
 
 
Command executed with exception: 
```

So I really suspect that `init_driver()` callbacks are not executed in CS mode. But I can't find out how this even possible. Bcs there s no way learner::generic_driver() function ignores reduction's `init_driver` call. It also seems that init_driver's value can't be overridden with `nope` function  one more time after `set_init_driver` is called.
Any ideas?

trufanov-nok:
Hey, now most of appveyor problems are gone. But there is still one more to discuss. 
Currently testing fails with `Assert.Fail failed. 0 vs 1: delta=1 > Epsilon=0.001: WeightedExampleSum`. That's because audit_regressor mode doesn't print out standard vw footer text to stderr. There is no weighted example sum printed. 
In C build toolchain stderr files are compared just line by line with some automatic double values detection and given accuracy for them. In C# some fields are implicitly looked in output and then they are compared with vw statistics. It would be great if such comparison would be suppressed for data that's not present in stderr.

Perhaps `FindDoubleEntry` could be modified to return NaN for `candidate == null` instead of 0.0. In this case NaN value could be used to bypass `FuzzyEqual` check.

trufanov-nok:
Hey, now most of appveyor problems are gone. But there is still one more to discuss. 
Currently testing fails with `Assert.Fail failed. 0 vs 1: delta=1 > Epsilon=0.001: WeightedExampleSum`. That's because audit_regressor mode doesn't print out standard vw footer text to stderr. There is no weighted example sum printed. 
In C build toolchain stderr files are compared just line by line with some automatic double values detection and given accuracy for them. In C# some fields are implicitly looked in output and then they are compared with vw statistics. It would be great if such comparison would be suppressed for data that's not present in stderr.

Perhaps `FindDoubleEntry` could be modified to return NaN for `candidate == null` instead of 0.0. In this case NaN value could be used to bypass `FuzzyEqual` check.

eisber:
I’m returning double? (Nullable<double>) to signal missing values in the stderr. Again you can pull from #954

There is yet another issue though, that only triggers in Debug on Windows. You’re using vector<string> in audit_regressor_data. Since the data structure is calloc’d the constructor for vector<string> is never called.
This in turn triggers a debug check for compatibility of iterators. I temporarily changed inline void audit_regressor_feature(audit_regressor_data& dat, const float /*ft_weight*/, const uint64_t ft_idx) to
    ostringstream tempstream;
    vector<string> foo = dat.ns_pre;
    for (const string& s : foo)
      tempstream << s;
    tempstream << ':' << (index >> stride_shift) << ':' << weights[index] << '\n';


    string temp = tempstream.str();

But this is not a permanent solution, as it really means you can’t use dat.ns_pre without copying. 2 options:

-          You switch to v_array

-          We move to “new”

Markus

JohnLangford:
You can also do a placement new (similar to here: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/v_array.h#L90 ) when mixing the C++ and calloc.

trufanov-nok:
I've replaced `audit_regressor_data::vector<string> ns_pre` with `audit_regressor_data::vector<string>* ns_pre` and new\delete it manually. LIke I did with `io_buf* out_file`.
This should fix the issue in Debug on Windows.

JohnLangford:
Looks good, merged.

JohnLangford:
(And thanks)

PR #925, opened 2016-01-28
Changes to support sklearn piplines, grid search, and estimator validation

jconwell:
I'm gonna be pretty detailed here so you know why I did what I did.

Line 24: removed pyvw.vw as a base class of sklearn_vw.VW for two reasons. First when initializing a pipeline you have to pass in a list of estimator instances, which would actually create an instance of pyvw.vw with its arguments.  Any secondary estimators that need initial regressors will fail to initialize because the prior estimators haven't run and generated their final regressors.  So I move the pyvw.vw access into a lazy factory method that everything else calls when they need it.  Second, sklearn's grid search can run the search in parallel, but it requires the estimator be pickleable.  The VW estimator can't be pickled because pylibvw doesn't include Boost.Python pickle support. Removing pyvw.vw as a base class solves this issue.

I moved convert_to_vw argument out of fit() and predict() and into the class init params.  This wasn't necessary for the VW class, but the VWClassifier inherits from LinearClassifierMixin which has a predict function with only an X arg
`def predict(self, X):`
This means you can't pass in convert_to_vw into VWClassifier.predict().  This is important if we're operating on pre-formatted vw string records and don't need to convert to vw format

I changed `locals().iteritems()` to `dict(locals())` because a weird thing in python/PyCharm when debugging, stepping over `locals().iteritems()` can cause the collection to change during iteration and throw an error.  This doesn't happen when not debugging / stepping over the code, but it doesn't hurt to fix it so it never throws the exception.

Returned self from fit() so pipeline can call transform() on the returned object `estimator.fit().transform()`

Added a transform() function. Any estimator in a pipeline that isn't the last estimator in the pipeline must have a transform() function.  This one doesn't do any transformation on X, but it does call vw.finish() to make sure the final regressor is persisted before the next estimator in the pipeline needs it.  

In predict() changed how num_samples gets initialized.  Not sure if this was a bug or not.  If X has only one dimension, I init num_samples to len(X).  Also if X is a list, it'll fail the call to shape(), so the except also sets num_samples to len(X). This will happen if you are passing in a list of pre-formatted vw string records.
 
Added a class ThresholdingLinearClassifierMixin, that lets you set the score threshold for pos/neg class separation.  LinearClassifierMixin is hardcoded to use 0.

Added args to support bfgs.  At some point we may want to add **kwargs to the end of VW.__init__ args to allow users to pass in other vw arguments that aren't hard coded parameters.  Maybe the same way pyvw.vw accepts its list of parameters.

Made some small changes to the tests to get them to pass.

Added a test that calls sklearn's check_estimator() on both VW and VWClassifier.  There were some small changes to sklearn_vw.VW to its class level attributes.  Ending them with an underscore, and deleting them on initialization.   This was because check_estimator() runs a test that clones estimators and makes sure all parameters are set to None on the cloned instances.  For some reason it was important that the self attributes have different names than the __init__ input parameters, otherwise the input parameters would get the values from the like named self attributes and the test would fail.

All that might be moot though, because check_estimator() also tries to invoke the estimator with its default parameters, in this case just "--quiet" and pass in some test data to its fit and predict functions.  This caused the python process to die with the error: "Process finished with exit code 139".  For now I just commented out the calls to check_estimator() in the test.  But I do think this would be a good thing to get working at some point.  I've added more details about the test failure in the comments for the test.

### DIFF
--- a/python/sklearn_vw.py
+++ b/python/sklearn_vw.py
@@ -21,7 +21,7 @@ CONSTANT_HASH = 116060
 INVALID_CHARS = re.compile(r"[\|: \n]+")
 
 
-class VW(BaseEstimator, vw):
+class VW(BaseEstimator):
     """ Vowpal Wabbit Scikit-learn Base Estimator wrapper
 
         Attributes
@@ -33,11 +33,14 @@ class VW(BaseEstimator, vw):
     """
 
     params = dict()
-    passes = 1
 
     def __init__(self,
+                 probabilities=None,
+                 bfgs=None,
+                 mem=None,
                  random_seed=None,
                  ring_size=None,
+                 convert_to_vw=None,
                  learning_rate=None,
                  l=None,
                  power_t=None,
@@ -100,6 +103,7 @@ class VW(BaseEstimator, vw):
 
         Parameters
         ----------
+        convert_to_vw (bool): flag to convert X input to vw format
         random_seed (int): seed random number generator
         ring_size (int): size of example ring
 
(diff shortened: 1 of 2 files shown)


arielf:
@jconwell
Thanks so much for all this work!

jconwell:
Absolutely

JohnLangford:
@hal3 have you had a chance to take a look at this?

PR #927, opened 2016-01-28
parser_done() inconsistency fix

### LINKED ISSUE #926: Parser::parser_done() inconsistency
While testing #923 I found out that resulting small audit files randomly appears empty. That happens bcs I call close() for output file in `end_examples()` and it wasn't called. We already had this problem in #726 (III) when it was never called after parser::set_done(). It seems it wasn't fixed for 100% and now it's randomly not invoked.

The problem is that `parser_done()` check may return `false` [here](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/learner.cc#L58) after [set_done()](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parser.cc#L598) was called.
That's because it has additional check `p->used_index != p->begin_parsed_examples`.
It looks like it was designed to not return true until all loaded examples are processed. But this function is only used twice [here](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/learner.cc#L45) in loop over `all.early_terminate` and in `if` clause with `early_terminate`. It seems it's not used anywhere else at current point of time. Also `set_done()` sets both `early_terminate` and `p->done()` to true.

I believe that  `p->used_index != p->begin_parsed_examples` shouldn't be checked at least when `all.early_terminate` is true. Is it's not possible to access `all` from this function I'd suggest replacing line 49
` else if (parser_done(all.p))` with
`  else if (all.early_terminate || parser_done(all.p))`
and line 58
`if (parser_done(all.p))`
with just
`if (all.p->done)`
in learner.cc

Perhaps parser_done() is never invoked in situation when p->done is true and all.early_terminate is false. In this case this function may be just thrown out and replaced with all.p->done. Perhaps I just misunderstand something but the soft fix proposed above seems to be safe.

trufanov-nok:
closes #926


arielf:
Sorry, you need to merge from master and repush.

trufanov-nok:
> Sorry, you need to merge from master and repush.
done

JohnLangford:
Reading through this code gave me a headache, so I rewrote it to be more obviously correct.  

Proof of correctness:
(1) set_done() makes all.early_terminate true, and hence no more examples are dispatched to the learner here: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/learner.cc#L47  Furthermore, when early_terminate=true, examples are drained as before.  Then, in all cases, end_examples() is called.  
(2) w.r.t. the parser thread, set_done also turns off the parser here: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parser.cc#L901 as before.

As a bonus, we get to nuke parser_done().   I believe this fully addresses the issue with a simpler solution.  Please reopen if not so.

PR #929, opened 2016-01-29
C++ 11 for (auto& elem : range)

eisber:
this also includes the JSON PR

### DIFF
--- a/cs/cs/Serializer/VowpalWabbitJsonSerializer.cs
+++ b/cs/cs/Serializer/VowpalWabbitJsonSerializer.cs
@@ -68,11 +68,11 @@ namespace VW.Serializer
         /// </summary>
         /// <param name="json">The example to parse.</param>
         /// <returns>The VowpalWabbit native example.</returns>
-        public VowpalWabbitExample Parse(string json)
+        public VowpalWabbitExample Parse(string json, ILabel label = null)
         {
             using (var textReader = new JsonTextReader(new StringReader(json)))
             {
-                return this.Parse(textReader);
+                return this.Parse(textReader, label);
             }
         }
 
@@ -81,13 +81,16 @@ namespace VW.Serializer
         /// </summary>
         /// <param name="reader">The example to parse.</param>
         /// <returns>The VowpalWabbit native example.</returns>
-        public VowpalWabbitExample Parse(JsonReader reader)
+        public VowpalWabbitExample Parse(JsonReader reader, ILabel label = null)
         {
             // avoid parameter passing for the sake of non-reentrantness
             this.reader = reader;
             using (VowpalWabbitMarshalContext context = new VowpalWabbitMarshalContext(this.vw))
             using (VowpalWabbitMarshalContext defaultNamespaceContext = new VowpalWabbitMarshalContext(this.vw, context.ExampleBuilder))
             {
+                if (label != null)
+                    this.defaultMarshaller.MarshalLabel(context, label);
+
                 if (!reader.Read())
                     throw new ArgumentException("json");
 
@@ -109,7 +112,9 @@ namespace VW.Serializer
                                     switch (propertyName)
                                     {
                                         case "_label":
-                                            this.ParseLabel(context);
+                                            // passed in label has precedence
+                                            if (label == null)
+                                                this.ParseLabel(context);
                                             break;
                                         //case "_shared":
                                         //    break;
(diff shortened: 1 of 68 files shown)


cs/cs/Serializer/VowpalWabbitJsonSerializer.cs
```diff
@@ -0,0 +1,304 @@
...
+using VW.Serializer.Intermediate;
+
+namespace VW.Serializer
+{
+    /// <summary>
+    /// A deserializer from JSON to Vowpal Wabbit native examples.
+    /// </summary>
+    public sealed class VowpalWabbitJsonSerializer
+    {
+        internal const string FeatureIgnorePrefix = "_";
```
lhoang29: why not put this in VowpalWabbitConstants and perhaps under a namespace/class called Serialization? Then you have all of these constant strings defined in the same place.

vowpalwabbit/interactions.h
```diff
@@ -128,7 +124,7 @@ template<class R>
   const uint64_t  weight_mask   = all.reg.weight_mask;
   weight* weight_vector = all.reg.weight_vector;
 
-  inner_data<R> id ={dat, offset, weight_mask, weight_vector};
+  // inner_data<R> id ={dat, offset, weight_mask, weight_vector};
```
lhoang29: nit: remove commented code

vowpalwabbit/feature_group.h
```diff
@@ -151,13 +322,15 @@ struct features { // the core definition of a set of features.
     slice.delete_v();
     return true;
   }
+
+  features& operator=(const features& rhs)
```
lhoang29: nice change

lhoang29:
Looks like I've already reviewed most of the C# side previously and the rest involves mainly C++ 11 refactoring. My main comment (as expected) is the pervasive use of 'auto'. IMHO using explicit types where needed makes code inspection and readability much easier because:

1. Via non-IDE text editors or browsers, locating code without knowing the type can be very difficult.
2. Although some IDEs like Visual Studio allow you to navigate to type definition on "auto" or "var" keywords, it's possible that other packages may not have the same support (for C# it's not a big issue because almost everyone uses VS for development).
3. Even in the cases where we don't care what the type is (i.e. as long as it implements some interfaces) which seems to be the main argument for 'auto', for code debugging we'd still need to find the actual resolved type. Without a step-through debugger this would again be difficult to do.

For an open-source project with many contributors where the code is rapidly changing, I'd vote for using auto only when necessary (e.g. lambda types) or if the type is clear from a glance (e.g. auto b = new Foo()) .

JohnLangford:
How simple can the full type be made?

-John

lhoang29:
You'd just change 'auto' to the actual type of the object/variable. The rest can stay the same (including the c++ 11 foreach syntax). e.g: `for (int& val : val_vector) ...`

eisber:
I think there are 2 more reasons to use auto:


-          If the types are awkward to write and not sure if anybody actually cares (e.g. vector<string>::iterator)

-          Eases refactoring. (e.g. if one wants to change the type of feature indices)

Markus

JohnLangford:
Hence my question: How simple can the full type be?

-John

lhoang29:
Hmm, I tend to agree with the first one for well known built-in types. The second seems like necessary plumbing work, plus the needs for refactoring is probably less than that for inspection & debugging.

lhoang29:
If the full type is too long but non-trivial, we can also use typedef

eisber:
I like Louie’s argument on debugging and non-IDE editors.

I’ll go through the code…

cs/cs/Serializer/VowpalWabbitJsonSerializer.cs
eisber: The constants class is already in VW.Serializer. Also the constant is only used in VowpalWabbitJsonSerializer, thus I’d rather keep it there.

vowpalwabbit/gd.cc
```diff
@@ -25,6 +25,7 @@ license as described in the file LICENSE.
 #include "reductions.h"
 #include "vw.h"
 #include "floatbits.h"
+#include <numeric>
```
JohnLangford: Can you shift <numeric> to come first in include order?

JohnLangford: That is 'numeric'.  Also, why is numeric needed?

JohnLangford:
This says conflicts must be resolved?   

The iterator stuff looks pretty good, but let's try to discuss the json stuff today.

vowpalwabbit/gd.cc
eisber: I can’t remember why I added it, but the code is gone. I removed it.

PR #932, opened 2016-02-03
JSON & ASP.NET

eisber:
Fixed DLL loading issues in ASP.NET
Refactored JSON serializer to enable example modification (e.g. add a label)
Minor fix for JSON serializer to be usable from within a JsonConverter

@lhoang29 please review

### DIFF
--- a/cs/cs/Serializer/VowpalWabbitJsonSerializer.cs
+++ b/cs/cs/Serializer/VowpalWabbitJsonSerializer.cs
@@ -24,6 +24,8 @@ namespace VW.Serializer
     /// </summary>
     public sealed class VowpalWabbitJsonSerializer
     {
+        internal const string FeatureIgnorePrefix = "_";
+
         /// <summary>
         /// Mapping from properties to types for labels.
         /// </summary>
@@ -67,6 +69,10 @@ namespace VW.Serializer
         /// Parses the example.
         /// </summary>
         /// <param name="json">The example to parse.</param>
+        /// <param name="label">
+        /// Optional label, taking precedence over "_label" property found in <paramref name="json"/>.
+        /// If null, <paramref name="json"/> will be inspected and the "_label" property used as label.
+        /// </param>
         /// <returns>The VowpalWabbit native example.</returns>
         public VowpalWabbitExample Parse(string json, ILabel label = null)
         {
@@ -80,6 +86,10 @@ namespace VW.Serializer
         /// Parses the example.
         /// </summary>
         /// <param name="reader">The example to parse.</param>
+        /// <param name="label">
+        /// Optional label, taking precedence over "_label" property found in <paramref name="reader"/>.
+        /// If null, <paramref name="reader"/> will be inspected and the "_label" property used as label.
+        /// </param>
         /// <returns>The VowpalWabbit native example.</returns>
         public VowpalWabbitExample Parse(JsonReader reader, ILabel label = null)
         {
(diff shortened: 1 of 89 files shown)


cs/cs/Serializer/VowpalWabbitJsonSerializer.cs
```diff
@@ -91,82 +103,130 @@ public VowpalWabbitExample Parse(string json, ILabel label = null)
...
+        }
 
-                var vwExample = context.ExampleBuilder.CreateExample();
+        /// <summary>
+        /// Creates the managed example representation.
+        /// </summary>
+        /// <returns>Returns the managed example.</returns>
+        public VowpalWabbitExample CreateExample()
+        {
+            try
```
lhoang29: how about 2 using statements for DefaultNSContext and Context? You can set them to null outside the using statements & before return

cs/cs/VowpalWabbitJson.cs
```diff
@@ -209,9 +227,13 @@ public TPrediction Predict<TPrediction>(string json, IVowpalWabbitPredictionFact
         /// </param>
         public TPrediction Predict<TPrediction>(JsonReader reader, IVowpalWabbitPredictionFactory<TPrediction> predictionFactory, ILabel label = null)
         {
-            using (var example = this.serializer.Parse(reader, label))
+            using (var serializer = new VowpalWabbitJsonSerializer(vw))
             {
-                return this.vw.Predict(example, predictionFactory);
+                serializer.Parse(reader, label);
+                using (var example = serializer.CreateExample())
```
lhoang29: seems like you could use another overload which does Parse & CreateExample in one shot

cs/cs/VowpalWabbitJsonThreadedPrediction.cs
```diff
@@ -0,0 +1,34 @@
...
+        public VowpalWabbitJsonThreadedPrediction(VowpalWabbitModel model)
+            : base(model)
+        {
+        }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="VowpalWabbit{TExample}"/>.
+        /// </summary>
+        /// <param name="vw">The wrapped vw instance.</param>
+        sealed protected override VowpalWabbitJson InternalCreate(VowpalWabbit vw)
```
lhoang29: "sealed protected" seems self-contradictory? :) maybe I'm not understanding what it means. The class is already sealed as well.

eisber: The base class defines a protected method, thus the protected. The sealed here makes sure nobody can override the method, but since the class is already sealed it’s not needed.

Done.

cs/cs/VowpalWabbitJson.cs
eisber: I was hesitating due to the 2 overloads (string, jsonreader). I’ve added it.

cs/cs/Serializer/VowpalWabbitJsonSerializer.cs
eisber: You need initialize in the using(…), and the contexts are initialized in the ctor.

lhoang29: I think you can do using() with an already initialized object

eisber: True, but I have to wrap the using statement with another try-finally to ensure the members are set to null (exception thrown vs. not thrown).

PR #934, opened 2016-02-05
Multiline JSON support

eisber:
@lhoang29 review time :)

I split VowpalWabbitJsonSerializer into VowpalWabbitJsonSerializer  and VowpalWabbitJsonBuilder

### DIFF
--- a/cs/cs/Serializer/VowpalWabbitJsonBuilder.cs
+++ b/cs/cs/Serializer/VowpalWabbitJsonBuilder.cs
@@ -112,6 +112,7 @@ namespace VW.Serializer
         /// Optional label, taking precedence over "_label" property found in <paramref name="reader"/>.
         /// If null, <paramref name="reader"/> will be inspected and the "_label" property used as label.
         /// </param>
+        /// <param name="specialPropertyAction">Action to be executed when sepcial properties are discovered.</param>
         /// <returns>The VowpalWabbit native example.</returns>
         public void Parse(JsonReader reader, ILabel label = null, Func<string, bool> specialPropertyAction = null)
         {
@@ -173,7 +174,7 @@ namespace VW.Serializer
             // special fields
             switch (propertyName)
             {
-                case "_label":
+                case VowpalWabbitConstants.LabelProperty:
                     // passed in label has precedence
                     if (label == null)
                         this.ParseLabel();
--- a/cs/cs/Serializer/VowpalWabbitJsonBuilder.cs
+++ b/cs/cs/Serializer/VowpalWabbitJsonBuilder.cs
@@ -1,22 +1,61 @@
-﻿using Newtonsoft.Json;
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="VowpalWabbitJsonSerializer.cs">
+//   Copyright (c) by respective owners including Yahoo!, Microsoft, and
+//   individual contributors. All rights reserved.  Released under a BSD
+//   license as described in the file LICENSE.
+// </copyright>
+// --------------------------------------------------------------------------------------------------------------------
+
+using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
+using System.Globalization;
 using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using VW.Interfaces;
+using VW.Labels;
 using VW.Serializer.Intermediate;
 
 namespace VW.Serializer
 {
+    /// <summary>
+    /// Build <see cref="VowpalWabbitExample"/> from JSON following https://github.com/JohnLangford/vowpal_wabbit/wiki/JSON
+    /// </summary>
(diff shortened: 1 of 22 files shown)


cs/cs/Serializer/VowpalWabbitJsonBuilder.cs
```diff
@@ -0,0 +1,389 @@
...
+
+                                if (reader.TokenType == JsonToken.StartObject)
+                                    this.ParseNamespaceAndFeatures(this.Context, propertyName);
+                                else
+                                    this.ParseFeature(this.DefaultNamespaceContext, defaultNamespace, propertyName);
+
+                            }
+                            break;
+                        case JsonToken.EndObject:
+                            goto done;
```
lhoang29: wouldn't break be enough here?

cs/cs/Serializer/VowpalWabbitJsonBuilder.cs
```diff
@@ -0,0 +1,389 @@
...
+                }
+            }
+        }
+
+        private void ParseSpecialProperty(VowpalWabbitMarshalContext context, Namespace ns, string propertyName)
+        {
+            // special fields
+            switch (propertyName)
+            {
+                case "_label":
```
lhoang29: didn't you already define this string as constant somewhere?

cs/cs/Serializer/VowpalWabbitJsonBuilder.cs
```diff
@@ -0,0 +1,389 @@
...
+        private void ParseFeatureArray(VowpalWabbitMarshalContext context, Namespace ns)
+        {
+            ulong index = 0;
+
+            while (reader.Read())
+            {
+                switch (reader.TokenType)
+                {
+                    case JsonToken.Integer:
+                        MarshalFloatFeature(context, ns, index, (float)(long)reader.Value);
```
lhoang29: why double cast?

cs/cs/Serializer/VowpalWabbitJsonBuilder.cs
```diff
@@ -0,0 +1,389 @@
...
+                        break;
+                    case JsonToken.Float:
+                        MarshalFloatFeature(context, ns, index, (float)(double)reader.Value);
```
lhoang29: same here

cs/cs/Serializer/VowpalWabbitJsonSerializer.cs
```diff
@@ -24,33 +24,270 @@ namespace VW.Serializer
...
+            /// </summary>
+            public abstract void Dispose();
+        }
 
         /// <summary>
-        /// Mapping from properties to types for labels.
+        /// Result for a single example.
         /// </summary>
-        private static readonly Dictionary<string, Type> labelPropertyMapping;
+        public sealed class SingleVowpalWabbitExample : VowpalWabbitExamples
```
lhoang29: perhaps rename this to SingleLine... to be in parity with Multiline

cs/cs/Serializer/VowpalWabbitJsonSerializer.cs
```diff
@@ -24,33 +24,270 @@ namespace VW.Serializer
     /// </summary>
     public sealed class VowpalWabbitJsonSerializer : IDisposable
     {
-        internal const string FeatureIgnorePrefix = "_";
+        /// <summary>
+        /// Base class for JSON deserialization result.
+        /// </summary>
+        public abstract class VowpalWabbitExamples : IDisposable
```
lhoang29: hmm, this class names seems slightly confusing given that there's already one called VowpalWabbitExample. I don't have a better name in mind though

lhoang29:
looks good in general

cs/cs/Serializer/VowpalWabbitJsonBuilder.cs
eisber: We’re inside a switch. Break only breaks the switch, not the loop.

cs/cs/Serializer/VowpalWabbitJsonBuilder.cs
eisber: No and done.

cs/cs/Serializer/VowpalWabbitJsonBuilder.cs
eisber: reader.Value is of type object, the actual object is a long. Cast directly from object to float fails. The first cast gets us from object to long, the second actually does a conversion.

cs/cs/Serializer/VowpalWabbitJsonSerializer.cs
eisber: I was unhappy too. Renamed to VowpalWabbitExampleCollection and moved to separate files.

cs/cs/Serializer/VowpalWabbitJsonBuilder.cs
lhoang29: ah, duh

Issue #935, opened 2016-02-09
scan-build static analysis multiple bugs


arielf:
I ran clang's static analyzer (`scan-build`) on latest master and it uncovered several issues that look like bona-fide bugs not covered by the test suite.

Here's a link to the report:

[scan-buid report on vowpal wabbit 2016-02-08](http://yendor.com/scan-build-2016-02-08/)

Summary of issues uncovered:

   ```
    Bug Group     Bug Type                                 File             Line
    Logic error   Branch condition evaluates to garbage    kernel_svm.cc     579
    Dead store    Dead assignment                          multiclass.cc      50
    Dead store    Dead increment                           gd.cc             590
    Dead store    Dead initialization                      search.cc         885
    Dead store    Dead initialization                      search.cc         936
    Logic error   Dereference of null pointer              cb_algs.cc         63
```

@trufanov-nok may be up your alley?

arielf:
Went through these.  All the 'Dead ...' are simply unused values so not really bugs but questionable assignments.

The last one in `cb_algs.cc` I'm not sure because of C++ opaqueness.

I'm fixing `kernel_svm.cc` which is a bug (potential use of uninitialized value).

eisber:
I had a look and I’m not getting it.

https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/cb_algs.cc#L63

does call begin()  and end() https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/v_array.h#L35 and https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/v_array.h#L36

ld is by ref, so that can’t be null. And ld.cost is not a pointer either.
ld.cost is initializes through the regular calloc.

According to http://en.cppreference.com/w/cpp/language/range-for the loop is transformed into

auto && __range = range_expression ;
for (auto __begin = begin_expr,
__end = end_expr;
__begin != __end; ++__begin) {
range_declaration = *__begin;
loop_statement
}

begin_expr being ld.cost.begin() and end_expr being ld.cost.end()

Markus

arielf:
Hi Markus, 
[...]

PR #936, opened 2016-02-09
Initialize fec to nullptr before use

arielf:
Partial fix for https://github.com/JohnLangford/vowpal_wabbit/issues/935


lhoang29:
can you fix other uninitialized variable declarations in the codebase as well? these could lead to nasty bugs

PR #938, opened 2016-02-10
Trace improvements

eisber:
Added VowpalWabbitString property for single and multiline example
Added missing file to VS proj

@lhoang29 review should be quick...


lhoang29:
lg

PR #939, opened 2016-02-12
search leak fix

### LINKED ISSUE #937: sequence tagging memory usage too high
I've ran an example provided in [this tutorial](https://github.com/JohnLangford/vowpal_wabbit/wiki/2015_tutorial.pdf).

Example snippet:
```
wget http://bilbo.cs.illinois.edu/~kchang10/tmp/wsj.vw.zip
unzip wsj.vw.zip
vw -b 24 -d wsj.train.vw -c --search_task sequence --search 45 --search_neighbor_features -1:w,1:w --affix -1w,+1w -f wsj.weights
```

bug is present in current HEAD 7eebf3a
Memory usage explodes (more than 10GB) and training is significantly slow, gets killed before training finishes.

training finished successfully (not the first commit where this happens): c78f026
memory usage is constant here, around 260MB (I think 256MB is allocated once for the longest possible sequence or weight array). Although, training took 10 minutes, not 10 seconds as reported in the tutorial.

Built it from source on osx i5 and ubuntu i7 - on both the memory explodes.

vjeranc:
this seems to resolve #937

### DIFF
--- a/vowpalwabbit/search.cc
+++ b/vowpalwabbit/search.cc
@@ -99,6 +99,12 @@ struct action_repr
   repr.space_names = _repr.space_names;
  }
   action_repr(action _a) : a(_a), repr() {}
+  void operator=(const action_repr & ar)
+  { repr.values = ar.repr.values;
+    repr.indicies = ar.repr.indicies;
+    repr.space_names = ar.repr.space_names;
+    a = ar.a;
+  }
 };
 
 struct action_cache
@@ -1953,8 +1959,6 @@ void search_finish(search& sch)
   if (priv.cb_learner) priv.gte_label.cb.costs.delete_v();
   else                 priv.gte_label.cs.costs.delete_v();
 
-  for (Search::action_repr& ar : priv.condition_on_actions)
-    ar.repr.delete_v();
   priv.condition_on_actions.delete_v();
   priv.learn_allowed_actions.delete_v();
   priv.ldf_test_label.costs.delete_v();
@@ -1986,8 +1990,6 @@ void search_finish(search& sch)
   }
   priv.learn_condition_on_names.delete_v();
   priv.learn_condition_on.delete_v();
-  for (Search::action_repr& ar : priv.learn_condition_on_act)
-    ar.repr.delete_v();
   priv.learn_condition_on_act.delete_v();
 
   if (priv.task->finish) priv.task->finish(sch);
--- a/vowpalwabbit/search.cc
+++ b/vowpalwabbit/search.cc
@@ -94,9 +94,9 @@ struct action_repr
 { action a;
   features repr;
   action_repr(action _a, features& _repr) : a(_a), repr()
-{ copy_array(repr.values, _repr.values);
-  copy_array(repr.indicies, _repr.indicies);
-  copy_array(repr.space_names, _repr.space_names);
+{ repr.values = _repr.values;
+  repr.indicies = _repr.indicies;
+  repr.space_names = _repr.space_names;
  }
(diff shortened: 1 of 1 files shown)


eisber:
Is deep_copy the right thing to do?

vjeranc:
There seems to be a need for deep copy in `push_at`, the compiler could make a move since `push_at` is called with a temporary `action_repr` but move semantics aren't implemented for features.

Deep copy is happening every time. The problem was that `push_at` calls had a temporary deeply copied object that was then assigned and copied deeply over object stored in `v_array<action_repr>`. The temporary object never freed its memory.

Now, instead of making a deep copy in the constructor, it's made at `push_at`. It seems no other code uses `action_repr` constructor so the semantics stay the same.

eisber:
Implementing move semantics in featuers sounds fine with me.

vjeranc:
If it's done for `v_array` then without overloading any operator it would automatically be done for `features`. I'm not sure how much code relies on `features` being deallocated on call (`delete_v()`, and similar), given my lack of familiarity with code base, I can't do it currently.

`action_repr` copy constructor was put there (at least that's how it looks to me) for easy deep construction of that temporary object. The assignment in `push_at` relied accidentally on shallow copy which turned out to be deep at one point - now the copy construction is shallow but assignment is deep. 

So, this fix is in that spirit :smile: 

Did you mean if deep copy at `push_at` is the right thing to do, or rephrased is deep copy necessary at all for `action_repr`?

Given the commit history deep copy of the features array was added during the initial implementation.

I added `operator=` for `action_repr` which is shallow always, and lack of deep_copy doesn't seem to influence the error values, have no idea why.

eisber:
I’m fine to do the change, but I’d like to discuss with John first. Here are the options:

  1.  I revert the operator=() overload and replace it with copy_array()
     *   copy_array() = deep_copy
     *   Default implementation of operator=() … shallow copy
  2.  We make the operation explicit, but forbidding assignment (private operator=())
     *   Rename copy_array() = to deep_copy()
     *   Introduce shallow_copy()
  3.  Enhance v_array in a way that we lazily copy the array, but it’s not obvious on how to do this. That way operator=() could server both shallow and deep copy.
I’d vote for (2).

This also implies that change has trickle down from v_array to features.

vjeranc:
this fix isn't good, [there's a deep copy](https://github.com/vjeranc/vowpal_wabbit/blob/leak_fix/vowpalwabbit/search.cc#L1417) which was shallow before (before `features` `operator=`).


Replacing `action_repr` with things below and removing my additions in `search_finish` seems to remove the leaks completely. This means deep_copy isn't used at all. (the average loss is surprisingly the same)

I'll push the change but given that implementation regressed a magnitude from the reported runtime in tutorial, I guess there's more problems with it.

```
struct action_repr
{ action a;
  features repr;
  action_repr(action _a, features& _repr) : a(_a), repr()
{ repr.values = _repr.values;
  repr.indicies = _repr.indicies;
  repr.space_names = _repr.space_names;
 }
  action_repr(action _a) : a(_a), repr() {}
  void operator=(const action_repr & ar)
  { repr.values = ar.repr.values;
    repr.indicies = ar.repr.indicies;
    repr.space_names = ar.repr.space_names;
    a = ar.a;
  }
};
```

JohnLangford:
My understanding of the bug is that '=' meant shallow copy, but Markus 
assumed it was meant as deep copy, yielding giant memory leaks.

Does '=' mean deep or shallow copy to people by default?   (For me, I 
think of it as shallow copy until proven otherwise.)

-John

PR #941, opened 2016-02-12
Enable marking weight vector to be sharable using KSM

smartkiwi:
This would mark weight vector as KSM sharable:
Doing so allows to save memory if you run multiple instances of the same model in read only mode.
See more https://www.kernel.org/doc/Documentation/vm/ksm.txt
To make use of this feature you'll need to have Linux kernel >= 2.6.32 and KSM enabled
To check is KSM enabled run the command

    $ grep KSM /boot/config-`uname -r`

if KSM is enabled you should see:

    >> CONFIG_KSM=y

You can enable KSM with command:

     $ sudo "echo 1 > /sys/kernel/mm/ksm/run"

To monitor shared pages use command:

    $ grep [0123456789] /sys/kernel/mm/ksm/*

Here is an example of shared pages - two VW models (24 bits each) are shared between 14 processes. 32770 (4Kb each) unique pages_shared, pages_sharing - 425998 pages.

    /sys/kernel/mm/ksm/full_scans:251
    /sys/kernel/mm/ksm/merge_across_nodes:1
    /sys/kernel/mm/ksm/pages_shared:32770
    /sys/kernel/mm/ksm/pages_sharing:425998
    /sys/kernel/mm/ksm/pages_to_scan:100
    /sys/kernel/mm/ksm/pages_unshared:0
    /sys/kernel/mm/ksm/pages_volatile:0
    /sys/kernel/mm/ksm/run:1
    /sys/kernel/mm/ksm/sleep_millisecs:20

### DIFF
--- a/vowpalwabbit/memory.h
+++ b/vowpalwabbit/memory.h
@@ -44,7 +44,10 @@ T* calloc_mergable_or_throw(size_t nmemb)
 // it allows to save memory if you run multiple instances of the same model
 // see more https://www.kernel.org/doc/Documentation/vm/ksm.txt
 // you need to have Linux kernel >= 2.6.32 and KSM enabled
-// to check is KSM enabled: `grep KSM /boot/config-`uname -r` CONFIG_KSM=y`
+// to check is KSM enabled run the command
+// $ grep KSM /boot/config-`uname -r`
+// if KSM is enabled you should see:
+// >> CONFIG_KSM=y
 // you can enable ksmd with sudo "echo 1 > /sys/kernel/mm/ksm/run"
 // mark address space as a candidate for merging
 
--- a/vowpalwabbit/memory.h
+++ b/vowpalwabbit/memory.h
@@ -55,19 +55,8 @@ T* calloc_mergable_or_throw(size_t nmemb)
     const char* msg = "internal warning: marking memory as ksm mergeable failed!\n";
     fputs(msg, stderr);
   }
-
-  void* data = calloc(nmemb, sizeof(T));
-  if (data == nullptr)
-  { const char* msg = "internal error: memory allocation failed!\n";
-    // use low-level function since we're already out of memory.
-    fputs(msg, stderr);
-    THROW(msg);
-  }
-
   return (T*)data;
 }
-template<class T> T& calloc_mergable_or_throw()
-{ return *calloc_mergable_or_throw<T>(1); }
 #else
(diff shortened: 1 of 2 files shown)


smartkiwi:
cc @dcrosta

arielf:
@smartkiwi 

Thanks for this contribution.  Some workloads (Linux-only + Large `-b` model + multiple separate processes sharing the model + no-learning) can greatly reduce memory footprint by sharing the weight vector with this change.

I'll wait a bit to let other people comment before merge to mainline.

JohnLangford:
Where is MADV_MERGEABLE defined?  Why don't you call calloc_or_throw() in the #else case?

vowpalwabbit/memory.h
```diff
@@ -21,4 +21,54 @@ T* calloc_or_throw(size_t nmemb)
 template<class T> T& calloc_or_throw()
 { return *calloc_or_throw<T>(1); }
 
+template<class T>
```
dcrosta: You might consider wrapping the whole thing in the `#ifdef`, and alias it to `calloc_or_throw` (with `#define`) if the constant isn't defined.

arielf:
@JohnLangford 
Looks like the patch is like this by design:  `MADV_MERGEABLE` is only defined if it is supported by the current headers on the machine this is built on (comees from `/usr/include/bits/mman-linux.h`).  If not, the patch just degenerates into whatever `calloc_or_throw()` was before the patch.

Even when `MADV_MERGEABLE` is defined, it is not sufficient to get the effect of the patch by calling `calloc_mergable_or_throw()`. The new merging identical pages functionality, needs also to be enabled on the machine by `sudo sh -c "echo 1 > /sys/kernel/mm/ksm/run"` if this second condition is true as well, the mapping of model-files into in-memory weight-vectors makes them sharable across processes.

smartkiwi:
@dcrosta Thank you for suggestions - I've replaced duplicate code with alias.
@arielf @JohnLangford please suggest what else need to be done for the pull request to be accepted/\

arielf:
The latest changes make this patch cleaner.

The Appveyor build faiiled because of tests 62,63, and 64 having small FP differences (failing the 1e-5 allowed epsilon assertion).

These failures don't look in any way related to this PR.

If @JohnLangford has no objection, I think we should accept this.

JohnLangford:
Looks good to me now.  Thanks much.

PR #942, opened 2016-02-15
fix documentation typo

jongbinjung:


JohnLangford:
I'm merging this on the theory that it helps, although I don't use R personally.

PR #946, opened 2016-02-18
Fixing #937 

eisber:
Talked with John on shallow/deep copy usage in search.cc. 

I can confirm that I'm getting 30% speed improvement & equivalent results when switching from deep to shallow copy for action_repr. According to John he purposefully introduced deep copy to fix a but. 
My conclusion: the parameters we use for testing and the data set doesn't require the deep copy...


vjeranc:
This means that for some tasks or set of parameters search is being slowed down without any particular reason.

What is the reason for deep copying  in the `action_repr`?

Custom conditioning features? Accidentally freeing freed memory?

JohnLangford:
@hal3 @KaiWeiChang 

A while ago I introduced a deep copy to fix a bug in the l2s code.  Apparently, this has a substantive negative impact on speed.  Can you take a look and see if there is a better fix?

PR #948, opened 2016-02-20
AppVeyor problems demo

trufanov-nok:
This isn't a real PR but a demo how fragile is C# Runtests check. It breaks AppVeyor's test whenever I add any new testcase for several reasons.


trufanov-nok:
The error description is unclear:
```
C:\vowpal_wabbit\cs\unittest\RunTests.tt : error : An expression block evaluated as Null [C:\vowpal_wabbit\cs\unittest\cs_unittest.csproj]
  System.ArgumentNullException: Value cannot be null.
  Parameter name: objectToConvert
     at Microsoft.VisualStudio.TextTemplating.ToStringHelper.ToStringWithCulture(Object objectToConvert)
     at Microsoft.VisualStudio.TextTemplating5734fd2532cd42bf942b72a8a56a871e.GeneratedTextTransformation.TransformText()
  executing transform_all from C:\vowpal_wabbit\cs\unittest\
  the following T4 templates will be transformed:
  C:\vowpal_wabbit\cs\unittest\RunTests.tt
  C:\vowpal_wabbit\cs\unittest\TestMarshalNumeric.tt
    \--> C:\vowpal_wabbit\cs\unittest\RunTests.cs    
    \--> C:\vowpal_wabbit\cs\unittest\TestMarshalNumeric.cs    
  transformation complete
```
I believe that TransformText crashes when RunTest contains 2 empty lines one after another. With this little commit I have added 2nd empty line to the end of RunTest. GitHub shows only one line for some reason, but there are 2.

The problem seems to happen in RunTest.tt in this [loop](https://github.com/JohnLangford/vowpal_wabbit/blob/master/cs/unittest/RunTests.tt#L34) and caused by addition of an empty `testcase` object to array.

This is not the only the only problem I faced with RunTest.tt, but only one I was able to localize. 
Unfortunately I can't efficiently debug them.
Would be great if someone will help me with other AppVeyor/C# specific  bugs that could be discussed at #923

trufanov-nok:
Inviting @eisber to the thread.

eisber:
I agree that the parsing of RunTests isn’t optimal. I can harden the parsing a bit. When Dan started working on the bigtests I was asking for a more parser friendly format…

Anyway, going forward we have 2 options:

-          Hardening the parsing of the existing perl file

-          Introducing a new format that can be easily parsed and contains more meta information (e.g. dependency information between tests). I’d vote for JSON, XML or a like. Something not bound to a particular testing framework, but parseable by everybody.

Markus

JohnLangford:
Hardening the parser sounds radically easier.  I'd suggest doing that in
the short term.

-John

trufanov-nok:
Yeah, hardening the parsing of the existing perl file seems to be enough.
I'll close this PR if you could make a new one that fixes two empty lines problem.

And then I would like to ask your opinion about other bugs I faced.
I've added 2 new test cases to Runtests in PR #923 
They cause Appveyor to fail. I couldn't find the reason. Then I tried to mark them as ignored with `# SkipC#` and alternatively by adding their numbers in `skipList`. In both cases Appveyor still fails. I'm not sure if skipping isn't working or there is other problem. ..

PR #949, opened 2016-02-22
Additional fixes to support sklearn grid search. 

jconwell:
Additional fixes to support sklearn grid search.  Added score method (required by grid search), and fixed get/set_params for how sklearn updates estimator parameters during grid search operations.

NOTE: As of this point _**parallel**_ grid search doesn't work when using a pipeline.  Pipelined estimators take their initial regressors from a file path set in the estimator parameters.  So if you have a grid search running across 4 jobs, each estimator will be reading its initial regressor from the same path. More than likely we'd have to subclass gridsearch to override initial and final regressor paths based on the job number that is running.  Or something like that.  

Pull request details:

__init__()
Added a few more vw parameters.  Unfortunately sklearn wants an explicitly named parameter for each estimator parameter, vs passing in a dictionary of key value pairs.  

transform() 
double check if vw is finished 

predict()
adding support for oaa models

score()
the score method is required for sklearn's grid search

get_params()
returns estimator's parameters exactly as needed for __init__ to be called with that copy of the parameters.

set_params()
updated to handle passes and convert_to_vw


JohnLangford:
Looks good from what I can tell, merging.

PR #950, opened 2016-02-22
Windows Installer & JSON command line tool

eisber:
added Windows installer
added JSON command line tool (vw_json.exe)
added VS style command prompt shortcut

@lhoang29 Review time... this included changes from JSON direct multi. Probably best to do this one first.


cs/setup/Product.wxs
```diff
@@ -0,0 +1,49 @@
...
+		</Feature>
+	</Product>
+
+	<Fragment>
+		<Directory Id="TARGETDIR" Name="SourceDir">
+			<Directory Id="ProgramFiles64Folder">
+        <Directory Id="INSTALLDIR" Name="VowpalWabbit" >
+          <Component Id="MainExecutable" Guid="02824B33-C335-4C2E-8B24-C4963ADEF294">
+            <File Id="SetEnvBat" Source="$(var.cs_console.ProjectDir)setenv.bat" KeyPath="yes"/>
+            <File Name="vw.exe" Source="$(var.vw.TargetDir)vw.exe"  />
```
lhoang29: is this the .exe from cs_console? would this require .net 4.5 to install as well?

cs/cs/Serializer/VowpalWabbitMultiExampleSerializerCompiler.cs
```diff
@@ -0,0 +1,91 @@
...
+                    !settings.EnableStringExampleGeneration);
+
+                var exampleParameter = Expression.Parameter(typeof(TExample), "example");
+
+                // CODE condition1 && condition2 && condition3 ...
+                var condition = multiFeature.ValueValidExpressionFactories
+                    .Skip(1)
+                    .Aggregate(
+                        multiFeature.ValueValidExpressionFactories.First()(exampleParameter),
+                        (cond, factory) => Expression.AndAlso(cond, factory(exampleParameter)));
```
lhoang29: what are these conditions? is this checking if the feature value is valid?

cs/setup/Product.wxs
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
+	<Product Id="*" Name="setup" Language="1033" Version="1.0.0.0" Manufacturer="Microsoft Research"
+           UpgradeCode="ddebf366-a8ca-4192-8552-dbf1d3603ae9">
```
lhoang29: where is the .exe hosted & does it auto-detect updates?

cs/unittest/TestJsonDirect.cs
```diff
@@ -149,7 +233,7 @@ public class JsonText
     [JsonObject(MemberSerialization = MemberSerialization.OptOut)]
     public class JsonContextArray
     {
-        [JsonProperty("_label")]
+        [JsonIgnore]
         public ILabel Label { get; set; }
```
lhoang29: looks like this is now auto-inferred in the serializer, correct?

cs/cs/Serializer/VowpalWabbitMultiExampleSerializerCompiler.cs
eisber: It’s checking the no object along the access path is null.

Imagine

Example ex;
Ex.namespace1.feature1

If (ex.namespace1 != null)



In the general serializer the access path can be arbitrary long.

cs/unittest/TestJsonDirect.cs
eisber: Actually “embedded” label support for JSON direct is not done yet.

You can pass labels directly with calls to Learn(), but extraction out of the object is not there yet…

Note: ingesting from string JSON, embedded labels are supported using _label.

cs/setup/Product.wxs
eisber: We’ll put it in the release section on Github. No auto-updating. Not sure when to check for this… I’m just installing the vw.exe, the vw_json.exe and create a shortcut to a command prompt with updated PATH variable.

You think it’s needed?

lhoang29: not necessarily, was curious mainly

JohnLangford:
What is the status on this?

eisber:
I really liked @lhoang29 comment on .NET 4.5 auto-detection and install. There are examples on how to do so in WIX, maybe I get time next week...

eisber:
@lhoang29  final look?

lhoang29:
looks good

PR #952, opened 2016-02-25
Fixed init_driver call

eisber:
It turns out that init_driver never got called from the Windows side. I walked through it with John and we also simplified the code a bit. Note: generic_driver is not used when going through the C# part, as the parser is externalized there.

@trufanov-nok can you merge that into #923 and try again?


trufanov-nok:
Ok, let's try it.

PR #953, opened 2016-02-26
Adding uncertainty scores (confidence estimate)

friday0013:
It generates uncertainty scores for ftrl regression model with --ftrl_confidence argument in command args.
The outputs look like below: 

prediction uncertainty score

    -0.917181 1.581583
    -1.040545 1.582132
    -1.098954 1.591752
    -1.280777 1.592833
    -1.290698 1.625208
    -1.404583 1.701759
    -1.545548 1.482271
    -1.432665 1.761744
    -1.563036 1.777928
    -1.548311 1.766415
    -1.598989 1.834791
    -1.815084 1.873315
    -1.646653 1.848350
    -1.742782 1.922865
    -2.340056 1.510852
    -1.816267 1.979960
     ...       ...

The score is calculated based on the following research paper, section (6): https://www.eecs.tufts.edu/~dsculley/papers/ad-click-prediction.pdf

Our validating efforts show very similar results as the one shown in the paper. Our works can be seen at: 
http://nativoplatform.github.io/vowpal_wabbit/uncertainty.html


JohnLangford:
Can the sensitivity interface be used for this?

https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/gd.cc#L975

Or is something fundamentally different being calculated?

vowpalwabbit/ftrl.cc
```diff
@@ -33,11 +34,79 @@ struct ftrl
...
+{ float pred;
+  float ub;
+  ftrl& b;
+  etas(ftrl& ftrlb) : b(ftrlb)
+  { pred = 0;
+  	ub = 0;
+  }
+};
+
+inline float sign(float w) { if (w < 0.) return -1.; else  return 1.;}
```
dmaust: Is this unused?

vowpalwabbit/ftrl.cc
```diff
@@ -33,11 +34,79 @@ struct ftrl
...
+
+inline void pred_confidence(etas& d, const float fx, float& fw)
+{ float* w = &fw;
+  d.pred += w[W_XT] * fx;
+  float sqrtf_ng2 = sqrtf(w[W_G2]);
+  float eta = ( (d.b.data.ftrl_beta+sqrtf_ng2)/d.b.data.ftrl_alpha +d.b.data.l2_lambda);
+  if(fx < 0)
+    d.ub += (1/eta)*fx*(-1);
```
dmaust: I think this is where you meant to use `sign()`

vowpalwabbit/ftrl.cc
```diff
@@ -33,11 +34,79 @@ struct ftrl
...
+  else
+    d.ub += (1/eta)*fx;
+}
```
dmaust: I think you meant to use `sign()` for this. Right now `sign()` is unused.

dmaust:
@friday0013 and I have been working together on this. I don't believe the sensitivity interface is applicable for what we are implementing, but we'll take a closer look on Monday.

JohnLangford:
Cool.  The sensitivity stuff measures how far an update changes the 
prediction.  It's used in confidence.cc and the active learning code.  
If that interface works, it means that FTRL can be used for active 
learning (for example).

-John

friday0013:
Hi John,

Thanks for your feedback.

We've taken a closer look at what confidence.cc does and how it's tied to other pieces. I think the outputs with --confidence are not correct when the --lrq option is turned on. In console, It's basically printing out two outputs for one example because of the loop in the method called predict_or_learn within lrq.cc. In addition, The outputs we'd get by using -p option, the predicted value is the first one in the loop as opposed to the confidence which is the value calculated in the second round of the loop.

Please take a look at the following code from lrq.cc. With --lrq, what’s being printed out must be the value from the first loop, not the second. 

    //Restore example
    if (iter == 0)
      { first_prediction = ec.pred.scalar;
        first_loss = ec.loss;
      }
    else
      { ec.pred.scalar = first_prediction;
        ec.loss = first_loss;
      }

example

     ./vowpalwabbit/vw -d features_with_samplelabels_original.gz --loss_function logistic --confidence --lrq ap10 -p prediction_with_uncertainty.pred
    confidence = 0 pred = 0 threshold = 0 sensitivity = 4.35838
    confidence = 0.18106 pred = -1.3807 threshold = 0 sensitivity = 7.62564
    0.693147 0.693147            1            1.0  -1.0000   0.0000        9
    confidence = 0.0429607 pred = -0.200151 threshold = 0 sensitivity = 4.65893
    confidence = 0.181702 pred = -1.51792 threshold = 0 sensitivity = 8.35394
    0.645609 0.598071            2            2.0  -1.0000  -0.2002        9
    confidence = 0.0666896 pred = -0.329055 threshold = 0 sensitivity = 4.93413
    confidence = 0.180085 pred = -1.61665 threshold = 0 sensitivity = 8.97714
    confidence = 0.112286 pred = -0.627078 threshold = 0 sensitivity = 5.58464
    confidence = 0.174897 pred = -1.83276 threshold = 0 sensitivity = 10.4791

in -p file

    0.000000 0.181060
    -0.200151 0.181702
    -0.329055 0.180085
    -0.627078 0.174897

As you can see, the first predicted value is paired with the second confidence value.

To correct the outputs in -p file, we can modify the above code to the following:

    float first_uncertainty = 0;
    //Restore example
    if (iter == 0)
      { first_prediction = ec.pred.scalar;
        first_loss = ec.loss;
        first_uncertainty = ec.confidence;
      } 
    else
      { ec.pred.scalar = first_prediction;
        ec.loss = first_loss;
        ec.confidence = first_uncertainty;
      }

JohnLangford:
Right.  lrq.cc does not call set_sensitivity() in the setup() function.
Most of the reductions do not do this---the base.sensitivity interface was
recently added.

Modifying any learner to support this interface would be great.  And if
there interface cannot work, I'd like to understand why.

-John

friday0013:
Hi John,

We are trying to compute uncertainty score (what we before referred to as a confidence), it's not fundamentally different to the sensitivity score in terms that they both are trying to answer the same question (how confidence are you with your predicted probability). But what's different is that it's specifically designed for ftrl. So I think we should add a function in ftrl.cc, which specifically compute the uncertainty score for the ftrl. Plus, we can edit confidence.cc so that it can point to a correct function with --confidence option depending on the user's choice of base algorithm (ftrl vs gd)

However, one question is that we aren't sure if our uncertainty score can be used as a replacement of the confidence in active.cc. If it's not, ftrl uncertainty score should only be used independently from active learning. 

Do you happen to have some references as to how confidence is calculated and used in active learning
? That'd be helpful. 

Please let us know what you think.

JohnLangford:
I'd prefer no edits to confidence.cc if possible (via using 
set_sensitivity in ftrl.cc).  If that doesn't work, let's discuss.

This notion of uncertainty comes from active learning analysis. See page 
6 here: http://arxiv.org/pdf/1011.1576v4.pdf .

-John

friday0013:
Thank you John. Sounds good to me. I'm working on it.

JohnLangford:
Status?

friday0013:
Hi John,

Sorry for the delay.
We've been trying hard to push this change(uncertainty score) to our production and it took much longer than I thought.
I'm still going to have to modify our current version a little bit to meet the vw's coding conventions. Plan is to update the pull request this week. I'll let you know if there's anything needs to be discussed.
Thank you,

Ryan

JohnLangford:
No worries.

-John

friday0013:
Hi John,

Just updated the ticket. Most of the changes were made to the ftrl.cc so that we can conveniently leverage the existing confidence.cc interface. It sets up the sensitivity in ftrl_setup method, return and print its output through confidence.cc.

I also made change to the lrq.cc by adding uncertainty variable so that it works properly with lrq option. 

One important thing to note is that, for our purpose, uncertainty score or sensitivity needs to be computed before the example gets trained whereas currently it gets trained before the confidence was calculated. Basically we want to know what the model would have predicted prior to the given training example being seen, so that we can optimize exploration. Looks like how the confidence is currently used better serve the purpose of active learning more closely than others. 

Example is below:

    template <bool is_learn>
    void predict_or_learn_with_confidence(confidence& c, base_learner& base, example& ec)
    { if (is_learn)
         base.learn(ec);
      else
        base.predict(ec);
      float threshold = 0.f;

      ec.confidence = fabsf(ec.pred.scalar - threshold) / base.sensitivity(ec);
      cout << "confidence = " << ec.confidence << " pred = " << ec.pred.scalar << " threshold = " << threshold << " sensitivity = " << base.sensitivity(ec) << endl;
    }

for our purpose it needs to be,

    template <bool is_learn>
    void predict_or_learn_with_confidence(confidence& c, base_learner& base, example& ec)
    { 
      ec.confidence = fabsf(ec.pred.scalar - threshold) / base.sensitivity(ec);
      if (is_learn)
         base.learn(ec);
      else
        base.predict(ec);
        float threshold = 0.f;
      
      cout << "confidence = " << ec.confidence << " pred = " << ec.pred.scalar << " threshold = " << threshold << " sensitivity = " << base.sensitivity(ec) << endl;
    }

So both confidence values before and after training are needed, would it make sense to add another member to the example class to store the new confidence/uncertainty value?

JohnLangford:
In active learning, sensitivity also comes before learning ( https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/active.cc#L49 ).   Given this, I'd suggest just changing confidence.cc to default to the behavior you want (confidence before training) and add a flag to get the present order instead.

JohnLangford:
I don't quite follow what you are trying to do in lrq.cc?  It looks like you assign a variable but never read from it?

friday0013:
Hi John,

I updated the ticket now to reflect your requested changes. 

In addition, I edited a RunTests because now confidence is computed by default. I also made changes to the confidence.preds and confidence_after_training.preds files accordingly. 

In lrq, it runs in a loop twice (#L72) and saves the first output to the dummy variables such as first_prediction and first_loss, then assigns those values to the ec variables in the second iteration so that those outputs can be the ones being exported in prediction file. I simply did the same thing for the confidence output.

https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/lrq.cc#L72
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/lrq.cc#L137-L144 

It looks like testing fails because of the changes made to the RunTests, but it works fine for me.

JohnLangford:
There is a nan in confidence.preds.  Can you fix?

friday0013:
fixed.
let me know.

JohnLangford:
(1) confidence values are always zero.   Seems like a bug?
(2) There are cout statements that should be removed.
(3) Why is appveyor complaining?

friday0013:
Well... looks like my outputs are identical to the outputs that I see from ec.confidence in the active.cc (See below and the link). Since computing confidence takes precedence over learning in active.cc, these outputs should be the same as the ones in the confidence.cc and they are. https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/active.cc#L49

    base.sensitivity(ec) -nan
    ec.pred.scalar 0
    ec.confidence -nan
    base.sensitivity(ec) inf   
    ec.pred.scalar -0.112613
    ec.confidence 0
    base.sensitivity(ec) inf
    ec.pred.scalar -0.0697124
    ec.confidence 0
    base.sensitivity(ec) inf
    ec.pred.scalar -0.170091
    ec.confidence 0
    base.sensitivity(ec) inf
    ec.pred.scalar -0.14779

You can also see all the confidence in the above output are equal to zeros. For the purpose of better testing, I suggest that we use dataset '/test/train-sets/rcv1_small.dat' which is used for testing active learning. This dataset contains more examples, and will let us see the outputs other than just zeros. Plus, adding '--initial_t 0.1' flag will resolve the issue of seeing nan value as well. But nan was generated from the gd.cc when initial_t is set by default.

friday0013:
Failed   CommandLine_Test121
    At line:4 char:4
    +    vstest.console /Platform:x64 /inIsolation vowpalwabbit\x64\Release\cs_unittes ...
    +       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~    ~
 
      Assert.Fail failed. 0.148571 vs 0.130575978518464: delta=0.0179950214815363 > Epsilon=0.001: AverageLoss
 
 
       Stack Trace:
 
 
      at cs_unittest.VWTestHelper.FuzzyEqual(Nullable`1 expected, Double actual, Double epsilon, String message) in c:\vowpal_wabbit\cs\unittest\VWTestHelper.cs:line 142
      at cs_unittest.VWTestHelper.AssertEqual(VowpalWabbitStdErrPerformanceStatistics expec
    ted, VowpalWabbitPerformanceStatistics actual) in c:\vowpal_wabbit\cs\unittest\VWTestHelper.cs:line 183
      at cs_unittest.RunTests.CommandLine_Test121() in c:\vowpal_wabbit\cs\unittest\RunTests.cs:line 1310

I'm getting the error above from the AppVeyor on the test #121, which was recently added. However, it passed all the tests from Travis as well as from my linux machine. It says the avg losses do not agree. In the RunTests.tt, I see skipList, and am wondering if this test should've been added to that list for some reasons. 

    var skipList = new[] { 9, 10, 13, 14, 18, 25, 26, 33, 16, 17, 19, 20, 24, 31, 32, 34, 43, 44, 45, 46, 47, 48, 49,
    50, 51, 52, 53, 54, 55, 56, 57, 58, 59, 60, 61, 65, 66, 67, 68, 69, 70,
    42, 71, 77, 84, 85, 86, 87, 92, 93, 96, 98, 99, 100, 101, 102, 103, 104, 105, 106,
    110, 111, 112, 114};

eisber:
John and I hunted this down. John/Alekh needs to fix something…

friday0013:
Hi there,
Hope you guys had a great weekend.
If you have any updates, please let me know so I can run the test again.

friday0013:
Hi John, 
Do you have any update on this? 
Please let me know if there's any thing I can help.
Thanks.

JohnLangford:
I've been very overloaded---will look at this soon.

-John

JohnLangford:
Can you tweak the test 113 to use rcv1_micro.dat instead?  (I shortened that one because rcv1_small was taking to long with valgrind.)

JohnLangford:
@eisber can you remind me what the test issue was here?

JohnLangford:
I think the test fail issue is the only thing fundamentally holding up a merge.

JohnLangford:
The bug for test 121 is fixed in the master.  Can you sync one last time please?

friday0013:
sync is done. looks like it passed AppVeyor test this time. 
any idea about coverage?

JohnLangford:
Merged at last!   Thanks much for sticking with it.

JohnLangford:
All, this patch has gone in, so FTRL now supports a sensitivity analysis 
useful for things like per-example confidence intervals and active 
learning.

Thanks much.

-John

friday0013:
Thank you John for helping us make this happen!

dmaust:
Thanks as well. Our wabbit is happy now!
wabbit-cropped

PR #958, opened 2016-03-03
fixed empty id being overridden in seeded model with @eisber

lhoang29:
When --id is not specified, we still add it to vw args as --id "", eventually this empty string is removed during the args copy in seeded model, causing the next flag to be used as --id, which is bad.


eisber:
great work :)

PR #961, opened 2016-03-17
merged changes with exploration library to add log-by-default for tau-first

lhoang29:


lhoang29:
@JohnLangford can you review?

JohnLangford:
Overloaded today, but I will when I have time.

JohnLangford:
Do we need the log_action variable any more?

Can you sync to current master?

lhoang29:
No we don't need it, but since it doesn't affect anything and could be useful in the future I decided to leave it (for example I could see us trying to optimize logging).

JohnLangford:
I think we should remove it on the theory that leaving dead code around is undesirable.

JohnLangford:
Remove the bool?

lhoang29:
Should we still log things when exploration is turned off? Or should we allow an option to log / not log? It might be also useful in this case to indicate whether the event was chosen in a deterministic fashion.

JohnLangford:
The right view is that we always explore, and hence determinism should not
prevent logging.

Turning off logging as an option does seem reasonable.  Is there an
interface to do so?

-John

lhoang29:
Yes there is already an interface to control logging. That's what I wanted to keep around. I can delete the extra local bools that do nothing (and always pass true into the logging interface).

JohnLangford:
Deleting the extra but keeping the interface functional seems like a 
good idea.

-John

PR #962, opened 2016-03-18
Cbify

eisber:
Alekh's refactoring.


JohnLangford:
This seems like it's unquestionably an improvement. 

@lhoang29 @alekh 
However, I have a concern related to the cbify<->cb_explore interface.  Should the random draw happen in cbify (via a generic explorer) instead of cb_explore?  That would allow a lower variance estimate of the loss.   Or, maybe we should bite the bullet and change the exploration library interface to report the probability distribution over actions?  We've discussed this with the view towards reporting confidence intervals.  Then, exploration can happen at the same point as present but the interface would again change to report the distribution over actions as well.

alekh:
You mean the random draw in the call to exploration library? This has to
happen in cb_explore for it to function correctly with code other than
cbify. But maybe I misunderstood you.

Alekh

JohnLangford:
I think it's more correct to say that the random call must either happen in cb_explore or we must export the probability vector and have it occur in every piece of code calling cb_explore.  It's not obvious to me which it should be, because we have both several elements plausibly calling cb_explore and several calls to the exploration library in cb_explore.   

Regardless of _where_ we do the exploration though, it seems like we'll end up wanting the vector of probabilities in the interface---for example for confidence interval calculation.

lhoang29:
I'm a bit confused, isn't the probability vector something that is given to the exploration library via the generic explorer interface? That means it already exists on the caller side, so I don't see why we need to return it back?

alekh:
Yes, I do agree that returning the probability vector is needed in various
places. The loss compassion in cbify is certainly higher variance than what
I had in mind initially due to this gap.

One thing which is a little funny is the interplay of this decision with a
generic explorer. All the call to generic explorer did was sample from a
specified distribution (after some validity checks). If a vector of
probabilities is the return type, doesn't it make the generic explorer
obsolete? Perhaps we want to return a tuple?

Alekh

alekh:
Only the generic explorer has it, not the others. But we want a consistent
API for all explorers.

Alekh

JohnLangford:
There is another problem.  --cb_explore appears unexposable on the
commandline (see below).  It looks like the callback inside the exploration
library is segfaulting because of something unexpected.   Why?

 1:36PM work-80: echo "1:1:0.5 |" | valgrind ../vowpalwabbit/vw
--cb_explore
2
~/programs/vowpal_wabbit/test [jcl/ttypts/0]
==8796== Memcheck, a memory error detector
==8796== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==8796== Using Valgrind-3.10.0.SVN and LibVEX; rerun with -h for copyright
info
==8796== Command: ../vowpalwabbit/vw --cb_explore 2
==8796==
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile =
num sources = 1
==8796== Invalid read of size 4
==8796==    at 0x577DCE: observed_cost(CB::cb_class*) (gen_cs_example.cc:20)
==8796==    by 0x577A14: get_observed_cost(CB::label&)
(gen_cs_example.cc:26)
==8796==    by 0x53D64D: void predict_or_learn<false>(cb&,
LEARNER::learner<char>&, example&) (cb_algs.cc:52)
==8796==    by 0x42F0FB: LEARNER::learner<char>::predict(example&, unsigned
long) (learner.h:106)
==8796==    by 0x504CF2: vw_policy::Choose_Action(vw_context&)
(cb_explore.cc:35)
==8796==    by 0x50B1F8:
MultiWorldTesting::SingleAction::EpsilonGreedyExplorer<vw_context>::Choose_Action(unsigned
long, vw_context&) (MWTExplorer.h:431)
==8796==    by 0x50847F:
MultiWorldTesting::SingleAction::MwtExplorer<vw_context>::Choose_Action(MultiWorldTesting::SingleAction::IExplorer<vw_context>&,
std::string, vw_context&) (MWTExplorer.h:139)
==8796==    by 0x507885: void predict_or_learn_greedy<true>(cb_explore&,
LEARNER::learner<char>&, example&) (cb_explore.cc:165)
==8796==    by 0x42388B: LEARNER::learner<char>::learn(example&, unsigned
long) (learner.h:101)
==8796==    by 0x491653: dispatch_example(vw&, example&) (learner.cc:9)
==8796==    by 0x4918C8: LEARNER::process_example(vw&, example*)
(learner.cc:38)
==8796==    by 0x491E92: void LEARNER::generic_driver<vw&,
&LEARNER::process_example>(vw&, vw&) (learner.cc:46)
==8796==  Address 0x3f80000000000001 is not stack'd, malloc'd or (recently)
free'd
==8796==
==8796==
==8796== Process terminating with default action of signal 11 (SIGSEGV)
==8796==  General Protection Fault
==8796==    at 0x577DCE: observed_cost(CB::cb_class*) (gen_cs_example.cc:20)
==8796==    by 0x577A14: get_observed_cost(CB::label&)
(gen_cs_example.cc:26)
==8796==    by 0x53D64D: void predict_or_learn<false>(cb&,
LEARNER::learner<char>&, example&) (cb_algs.cc:52)
==8796==    by 0x42F0FB: LEARNER::learner<char>::predict(example&, unsigned
long) (learner.h:106)
==8796==    by 0x504CF2: vw_policy::Choose_Action(vw_context&)
(cb_explore.cc:35)
==8796==    by 0x50B1F8:
MultiWorldTesting::SingleAction::EpsilonGreedyExplorer<vw_context>::Choose_Action(unsigned
long, vw_context&) (MWTExplorer.h:431)
==8796==    by 0x50847F:
MultiWorldTesting::SingleAction::MwtExplorer<vw_context>::Choose_Action(MultiWorldTesting::SingleAction::IExplorer<vw_context>&,
std::string, vw_context&) (MWTExplorer.h:139)
==8796==    by 0x507885: void predict_or_learn_greedy<true>(cb_explore&,
LEARNER::learner<char>&, example&) (cb_explore.cc:165)
==8796==    by 0x42388B: LEARNER::learner<char>::learn(example&, unsigned
long) (learner.h:101)
==8796==    by 0x491653: dispatch_example(vw&, example&) (learner.cc:9)
==8796==    by 0x4918C8: LEARNER::process_example(vw&, example*)
(learner.cc:38)
==8796==    by 0x491E92: void LEARNER::generic_driver<vw&,
&LEARNER::process_example>(vw&, vw&) (learner.cc:46)
==8796==
==8796== HEAP SUMMARY:
==8796==     in use at exit: 11,210,958 bytes in 1,143 blocks
==8796==   total heap usage: 7,151 allocs, 6,008 frees, 11,592,601 bytes
allocated
==8796==
==8796== LEAK SUMMARY:
==8796==    definitely lost: 0 bytes in 0 blocks
==8796==    indirectly lost: 0 bytes in 0 blocks
==8796==      possibly lost: 17,187 bytes in 316 blocks
==8796==    still reachable: 11,193,771 bytes in 827 blocks
==8796==         suppressed: 0 bytes in 0 blocks
==8796== Rerun with --leak-check=full to see details of leaked memory
==8796==
==8796== For counts of detected and suppressed errors, rerun with: -v
==8796== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)

alekh:
My best guess from looking at this is that ec.l is pointing to not the cb
but some other label type in the union. Hence the reading of ec.l.costs
results in an invalid access. I am stumped as to why that should be the
case though, or why this behavior only happens on access through the
command line.

A more plausible worry might have been that objects such as all.scorer and
all.cs are properly initialized and setup when invoked via cbify but not
via command line. That should have manifested as a different error if it
were the case though.

Alekh

JohnLangford:
It isn't clear this is _only_ a commandline issue.  It might also happen by
other interfaces, and this might be a bug due to some failed expected
coupling with cbify.cc.  Does that help?

JohnLangford:
Ah, I think I've got it. You are calling init_multiclass_learner which sets
the wrong label parser.  You need to call init_learner as in cb_algs.cc.

-John

JohnLangford:
Confirmed, fixed, added a test, pushed.

Shockingly, something sane happens when you use -p: it outputs the
predicted action.  This should make everyone feel squeaby because it is
misaccessing a union type.

JohnLangford:
Ok, fixed that issue.   I think cb_explore is ready for wider use now
except for the interface issue.

lhoang29:
Regarding the interface of the library, this means we are returning the prob vector at every ChooseAction? So for eps greedy, you'd compute probability for each action and return them all? This does make the generic explorer sort of obsolete as Alekh suggested, but I think it's ok.

JohnLangford:
Not at present, but that's the proposal.

Why does it make generic explorer obsolete?   Generic explorer could be
used on the probability vector to create the current interface.

-John

lhoang29:
It's just because if one uses generic explorer as the exploration algorithm, then the only benefit it provides is the sampling from the distribution. So if I understand correctly, you're proposing that all explorers return the prob vector, then generic explorer can be used as a way to sample from it?

JohnLangford:
There might be a confusion.  There are two 'returns' from exploration 
library: the return to the calling function and the return to the log.

The return to the calling function should not change.  The return to the 
log needs to change in order to enable confidence interval calculations.

In between that, there is a question about how we represent thing 
internally.  At the moment, I'm not feeling fond of every explorer 
having it's own type.  Instead, it seems like there should be one type 
which can be configured to use different explorers.  I'm also thinking 
that maybe the explorers should all backend on the generic explorer so 
randomization and logging only happen at one place.

What do you think?

-John

lhoang29:
Oh I see, things are much clearer now.

Regarding type, I think it's nice to have the specifics of each explorer be wrapped in a standalone type. The ChooseAction logic can differ quite a bit so if we collapse things to a single type it might get messy.  

The generic explorer backend is interesting, having a single point of randomization seems useful. I don't see any issues with it at the moment. What do others think?

alekh:
Thanks for your help with fixing this John. I had wondered while
refactoring whether init_multiclass made sense, but figured it was all good
since the code did work correctly with cbify.

Also, thanks Louie and Markus for your help as always.

Alekh

JohnLangford:
On 03/21/2016 06:50 PM, Luong Hoang wrote:
>
> Regarding type, I think it's nice to have the specifics of each 
> explorer be wrapped in a standalone type. The ChooseAction logic can 
> differ quite a bit so if we collapse things to a single type it might 
> get messy.
>

It's a question of "where is the mess?"  Right now, the cb_explore code 
is kind of messy, because the type of the explorer is externalized.  It 
seems better (as in: easier interface to use) if the mess is internal to 
the exploration library.


-John

lhoang29:
That's fair. This discussion actually reminds me that our very first attempt at the API was as you mentioned. We have a single explorer type MwtExplorer, then we initialized it with the exploration algorithm: explorer.InitializeWithEpsilonGreedy() and the rest follows. For some reasons (which I can't recall), we decided to redesign it. @sidsen do you remember? I tried to look through the thread but there were too many of them.

sidsen:
I like John's idea of putting the randomization/logging in one place (e.g.,
generic explorer) and having all other exploration algorithms invoke it. I
am lacking context here (haven't looked at the code) so I don't know how
feasible this is in cb_explore.

Whether we should use a single type that supports multiple exploration
algorithms is a more debatable. I believe we moved away from this because
each exploration algorithm had different state and a different way of
deciding which action to return. The latter will be alleviated if they all
invoke the generic explorer and pass in different distributions, but the
former won't go away. For things like epsilon greedy, tau first, and soft
max the state is minimal, though even then you have three different
parameters in the same class. For more complicated exploration algorithms
that actually maintain bags of policies, it could get messier.

Sid

JohnLangford:
This is not a question for cb_explore, but rather for the exploration 
library (which is used by cb_explore).

The state of different exploration algorithms definitely does differ.  
The argument for putting it in the library is that it minimizes the 
complexity for the user of the library.  Why should changing the 
exploration algorithm require retyping things?  It should 
not---conceptually a different exploration algorithm should have little 
impact on a program's type structure.

I realize union is out of favor these days, but it entirely addresses 
the 'extra state' issue.

-John

lhoang29:
I think there's a nice middle ground here. Since our goal is to simplify user interface, we can provide a single-typed presentation without having to combine the internal classes. Essentially the current usage would be moved inside the library. Externally one can do something similar to the below:

MwtExplorer explorer(app_id, recorder, exploration_type); // can also pass in the 
explorer.choose_action(context, ...)

where exploration_type can be just the parameters of the desired algorithm, for example: epsilon, tau, etc... 

What do you think?

sidsen:
I agree with John that changing exploration algorithms need not change a program's type structure, at least conceptually. I'm also willing to add messiness inside our library if it simplifies the user's experience -- I'd prefer separate internal structs/inner classes to a union type, though.

Note that the user will still have to pass in the parameters etc. specific to the exploration algorithm, either in the MwtExplorer constructor as Louie suggests above, or to some method that allows you to "SetExplorationAlgorithm" or something. The user doesn't need to create a new type, but still has to write some code. One nice thing about our current interface is you can pass whatever explorer class you want into ChooseAction(...); perhaps we can get the same effect by extending Louie's proposal above to include a "SetExplorationAlgorithm" method or still taking it in as a parameter in ChooseAction, in which case it would override the default specified in the constructor.

The problem with the above is that some exploration algorithms have state, like tau first. If we want to support changing/switching exploration algorithms, I think I prefer our current interface where the exploration alg is encapsulated in its own class. Otherwise maybe we should just disallow changing/switching and force them to specify the exploration alg in the constructor, and that's it. (If you want to change, create a new MwtExplorer and start again.)

PR #967, opened 2016-03-23
deleted dead bool

lhoang29:
no more dead


lhoang29:
@JohnLangford good to merge?

PR #972, opened 2016-03-27
Fix for Issue 960

gramhagen:
cleaning up predict call in sklearn_vw to use new predict method in pyvw
removing finish() there which causes segmentation fault when run predict, or score multiple times
added test to catch this case

### DIFF
--- a/python/pylibvw.cc
+++ b/python/pylibvw.cc
@@ -122,17 +122,13 @@ float my_learn_string(vw_ptr all, char*str)
 }
 
 float my_predict(vw_ptr all, example_ptr ec)
-{ bool old_test_only = ec->test_only;
-  ec->test_only = true;
-  all->l->learn(*ec);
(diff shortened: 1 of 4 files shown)


JohnLangford:
This seems to kill of the possibility of using the python interface for multiclass predictions?

JohnLangford:
This seems like an improvement on what's their so I'm willing to incorporate it.  

However, the label type detection is still kindof buggy and prone to needing future fixes.    Do you think something like this: https://github.com/JohnLangford/vowpal_wabbit/blob/master/java/src/main/c%2B%2B/vw_learner_VWLearners.cc#L41 is feasible instead?

gramhagen:
I can take a look and clean it up a bit.

scott

gramhagen:
should be good now, take a look

JohnLangford:
Merged.  Thanks for working through it!

PR #977, opened 2016-04-08
Update README.md

nikete:


arielf:
@nikete 

Thanks for the patch!

Trying to think minimalist: do we actually need the `-dev` version of zlib?
If (unlike boost) it is not written in C++, woudn't `zlib1g` be sufficient?

Also: I'm surprised this is actually needed at all.  Have you actually got a build error without it on Ubuntu? `zlib` is so universal (used by so many programs), that I believe on Ubuntu it always already be installed anyway. No?

Thanks again

nikete:
you need to the header to build, hence the -dev, so I think yes.

it is not in many of the default ubuntu EC2 images, I had it come up twice in the wild in a couple of days, it seems to be mostly in new instances or very inexperienced users; since I am helping write some DNA analysis software (hhga) a lot of biologists probably dont have them.

arielf:
Thanks @nikete 

@JohnLangford for some reason my Merge pull request button is greyed-out so I can't merge right now.

I think this should go in.  It seems definitely safe.

JohnLangford:
I merged by hand.  Thanks!

Ariel, I think the button based merge fails when the patch is not uptodate against the master head.  Since the head is a moving target, this happens.

arielf:
John, thanks for the greyed-out button explanation. Makes sense.

PR #978, opened 2016-04-15
Makefile.am: Install some missing header dependencies

willnewton:
These headers are now required to build against the library so install them.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,6 +4,10 @@ nobase_include_HEADERS = vowpalwabbit/allreduce.h \
 	vowpalwabbit/comp_io.h \
 	vowpalwabbit/config.h \
 	vowpalwabbit/example.h \
+	vowpalwabbit/feature_group.h \
+	vowpalwabbit/cb_explore.h \
+	vowpalwabbit/crossplat_compat.h \
+	vowpalwabbit/parse_example.h \
 	vowpalwabbit/floatbits.h \
 	vowpalwabbit/global_data.h \
 	vowpalwabbit/hash.h \


JohnLangford:
Travis barfed?  Why?

willnewton:
I'm not 100% sure of the precise reason, the cause is a testsuite failure which seems rather unrelated. Possibly could be a transient error and restarting the build would fix it?

It looks like the handling of autotools output is a little bit unusual in that the Makefile.in is not checked in (this is fairly common and requires developers to run autogen.sh and have autotools installed to build) but Makefile is checked in. Traditionally Makefile is generated from Makefile.in by configure and not checked in as configure will write a different Makefile for different users systems. 

In that respect I think the "right" thing to do is:
 1. Remove Makefiles from the repository
 1. Add back Makefile.in to the repository generated by appropriate automake
 1. Make Travis call configure before make

However I can't really say if that would have any effect on the testsuite failure seen here. If you agree I can put together a PR to implement the above.

JohnLangford:
I like the custom Makefile, because it makes compiling easier/simpler in 
many situations and avoids the need for autoconf steps.

The fail is here: 
https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/123382260#L2479 
This looks to me like an unrelated failure due to a heisenbug we need to 
track down.  Ariel, can you excite the bug reliably?

-John

JohnLangford:
Merged, thanks.

arielf:
I tried to reproduce this failure in test 20.
Couldn't reproduce.

Indeed looks like a new kind of heisenbug.

Losses are zero during training progress, and in that travis run they were non zero which is very weird:
```
    --- diff -u --minimal pred-sets/ref/0022.stderr 0022.stderr
    --- pred-sets/ref/0022.stderr   2016-04-15 16:26:34.534762162 +0000
    +++ 0022.stderr 2016-04-15 16:31:29.046762162 +0000
    @@ -9,17 +9,17 @@
    num sources = 1
    average   since         example        example  current  current current
    loss      last          counter         weight    label  predict features
    -0.000000 0.000000            1            1.0  -1.0000  -1.0000 4
    -0.000000 0.000000            2            2.0  -1.0000  -1.0000 4
    -0.000000 0.000000            4            4.0  -1.0000  -1.0000 4
    -0.000000 0.000000            8            8.0   1.0000   1.0000 4
    +1.786114 1.786114            1            1.0  -1.0000   0.3365 4
    +2.855245 3.924376            2            2.0  -1.0000   0.9810 4
    +1.587504 0.319763            4            4.0  -1.0000  -0.2003 4
    +0.848866 0.110229            8            8.0   1.0000   1.0000 4

    finished run
    number of examples per pass = 8
    passes used = 1
    weighted example sum = 8.000000
    weighted label sum = 0.000000
    -average loss = 0.000000
    +average loss = 0.848866
    best constant = 0.000000
    best constant's loss = 1.000000
    total feature number = 32
    RunTests: test 20: FAILED: ref(pred-sets/ref/0022.stderr) != stderr(0022.stderr)
```

Unrelated: I see we also print predictions as {1, -1} now whereas in the reference file they still
show in the old format: {1.00000, -1.00000} (and succeed because of the (slow) fuzzy-compare).
I suggest changing the ref-file
    `test/pred-sets/ref/0022.predict`
to use integer predicts, like the current code is producing.

JohnLangford:
I'm almost certain that this problem is due to the change we made in the 
learner.cc driver structure.

-John

PR #986, opened 2016-04-29
Thread-safe CLR label parsing

eisber:
- added thread-safe label parsing in CLR
- minor fixes for dynamic object marshalling
- added thread-safe example pooling


JohnLangford:
@lhoang29 can you do a code check?

cs/common/Bag.cs
```diff
@@ -0,0 +1,92 @@
...
+        void Add(T item);
+
+        T Remove();
+
+        IEnumerable<T> RemoveAll();
+    }
+
+    public static class Bag
+    {
+        public static IBag<T> Synchronized<T>(IBag<T> bag)
```
lhoang29: Is ConcurrentBag<> not what you want here?

cs/cs/Serializer/VowpalWabbitDefaultMarshallerExt.tt
```diff
@@ -88,13 +101,16 @@ namespace VW.Serializer
...
             {
-                context.NamespaceBuilder.AddFeaturesUnchecked((ulong)(ns.NamespaceHash + i), begin, begin + value.Length);
+                var temp = begin + offset;
+                context.NamespaceBuilder.AddFeaturesUnchecked((ulong)(ns.NamespaceHash + i), temp, temp + length);
             }
 
             <# } else { #>
 
-            foreach (var v in value)
+            for (var j = offset;j<length;j++)
```
lhoang29: nit: spacing issues

lhoang29:
looks good overall, I went over the cache resolver and didn't see any issues.

cs/common/Bag.cs
eisber: Looked at ConcurrentBag. Doesn’t have exactly the interface I’m looking for. Also there is no base class/interface that I could switch in/out if thread safety isn’t needed…

PR #987, opened 2016-04-29
bugfix - namespace now stores 16 instead of just 2 affixes

### LINKED ISSUE #983: search learning model performance regression
It seems that the learner (at least the one used for `--search`) has regressed (not a speed related issue).
The [7.10](https://github.com/JohnLangford/vowpal_wabbit/tree/c1184888abc799433d4071ca1e74dbbf8894aec9) seems to be significantly outperforming latest HEAD b879ee2222b9bb647b4646b608fdde9592294be4.

It might also be that for my particular case the behaviour changed significantly. I couldn't get similar performance on the same data no matter how hard I tried :smile: At first I thought Croatian is difficult, or that the dataset isn't sufficient but it isn't true.

I've attached a small Croatian corpus ([data.zip](https://github.com/JohnLangford/vowpal_wabbit/files/238461/data.zip)) for part-of-speech and MSD tagging. (MSD includes part-of-speech and various other attributes, which implies that accuracy will be lower than POS)

I use the following settings on both 7.10 and current head.
```sh
$VW --cache \
    --bit_precision 28 \
    --data $TRAIN \
    --kill_cache \
    --search_task sequence \
    --search $poscount \
    --passes 3 \
    --search_neighbor_features -2:w,-1:w,1:w,2:w \
    --affix -5w,-4w,-3w,-2w,-1w,1w,+2w,+3w,+4w,+5w \
    --search_alpha 1e-8 \
    --holdout_off \
    --search_rollout=none \
    --final_regressor $MODEL \
    --ring_size 256
```
Accuracy and vw reported loss on test set by commit:

| Commit                                   | POS    | MSD    | Avg. loss |
|------------------------------------------|--------|--------|-----------|
| c1184888abc799433d4071ca1e74dbbf8894aec9 | 0.9738 | 0.8830 | 1.4323    |
| b879ee2222b9bb647b4646b608fdde9592294be4 | 0.9464 | 0.8088 | 2.19382   |

The performance with the current HEAD can be better if `--affix` features are removed and `--search_neighbor_features -1:w,1:w` used instead but still not near the 7.10 numbers. I'm not sure why that is the case. (this might not work on the current permutation of the dataset but works in general)

I also couldn't get satisfying results on dependency parsing for Croatian and other languages (couldn't reproduce http://arxiv.org/abs/1503.05615), I assumed I just didn't play long enough but it seems that every search task might be affected.

It's also not the case that only the current head has the problem. Similar bad performance happens few commits before.

I'll do a `git bisect` to see where the regression point is for the POS task.

`--search 13` for POS
`--search 686` for MSD

I use bash below for measuring accuracy where files do not have empty lines:
```sh
a1=`diff .outputtrue .output | grep "^>" | wc -l`
a2=`wc -l < .outputtrue`
echo "Acc:" `calc "1-$a1/$a2"`
```

This is the testing call:
```sh
$VW -t -d $TEST -i $MODEL --ring_size=256 -p .outputtest # --quiet
```

vjeranc:
fixes #983 
previously with char there were only 2 affix values possible (4 bits) per namespace
`uint32_t` could hold 8 (changed code comment)

currently 16 are available which is, I guess, more than enough.

### DIFF
--- a/vowpalwabbit/global_data.h
+++ b/vowpalwabbit/global_data.h
@@ -475,7 +475,7 @@ struct vw
   uint32_t skips[256];//skips in ngrams.
   std::vector<std::string> limit_strings; // descriptor of feature limits
   uint32_t limit[256];//count to limit features by
-  char affix_features[256]; // affixes to generate (up to 8 per namespace)
+  uint64_t affix_features[256]; // affixes to generate (up to 16 per namespace - 4 bits per affix)
   bool     spelling_features[256]; // generate spelling features for which namespace
   vector<string> dictionary_path;  // where to look for dictionaries
   vector<feature_dict*> namespace_dictionaries[256]; // each namespace has a list of dictionaries attached to it
--- a/vowpalwabbit/parse_example.cc
+++ b/vowpalwabbit/parse_example.cc
@@ -33,7 +33,7 @@ public:
   unsigned char (*redefine)[256];
   parser* p;
   example* ae;
-  char* affix_features;
+  uint64_t* affix_features;
   bool* spelling_features;
   v_array<char> spelling;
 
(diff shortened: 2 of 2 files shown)


vjeranc:
Hmm. 1af8a55fd52852e668a1e12eaa8c0a9ba3f1619d fails on test 17. My commit too.

JohnLangford:
This looks good but the test fail is concerning.  Can you bump the branch so that it retests?  (There is evidence of a heisenbug related to thread locking that we saw recently.)

vjeranc:
Bumped.

JohnLangford:
Yep, looks like the heisenbug.

JohnLangford:
Thanks much for nailing this.

PR #989, opened 2016-05-03
working except print formatting

alekh:


JohnLangford:
This is merged in now, thanks Alekh.

All, the exploration distribution over actions created in --cb_explore is
now fully exposed, implying that the randomization over the action can now
be external to VW.

-John

Issue #965, opened 2016-03-19
unrecognized command line option ‘-stdlib=libc++’ when making with autogen on On Mac OS X 10.11.3


kamal94:
``` git clone https://github.com/JohnLangford/vowpal_wabbit ```
``` cd vowpal_wabbit ```
``` sh autogen.sh --enable-libc++ ```
``` make ```

OUTPUT:

git clone https://github.com/JohnLangford/vowpal_wabbit
Cloning into 'vowpal_wabbit'...
remote: Counting objects: 38776, done.
remote: Total 38776 (delta 0), reused 0 (delta 0), pack-reused 38776
Receiving objects: 100% (38776/38776), 97.58 MiB | 3.22 MiB/s, done.
Resolving deltas: 100% (29241/29241), done.
Checking connectivity... done.

 18:36:23  ~/Repositories 
$ cd vowpal_wabbit 

 18:36:37  ~/Repositories/vowpal_wabbit    master ✔ 
$ sh autogen.sh --enable-libc++
glibtoolize: putting auxiliary files in '.'.
glibtoolize: copying file './ltmain.sh'
glibtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'acinclude.d'.
glibtoolize: copying file 'acinclude.d/libtool.m4'
glibtoolize: copying file 'acinclude.d/ltoptions.m4'
glibtoolize: copying file 'acinclude.d/ltsugar.m4'
glibtoolize: copying file 'acinclude.d/ltversion.m4'
glibtoolize: copying file 'acinclude.d/lt~obsolete.m4'
configure.ac:15: installing './compile'
configure.ac:13: installing './config.guess'
configure.ac:13: installing './config.sub'
configure.ac:3: installing './install-sh'
configure.ac:3: installing './missing'
cluster/Makefile.am: installing './depcomp'
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... ./install-sh -c -d
checking for gawk... no
checking for mawk... no
checking for nawk... no
checking for awk... awk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether the C++ compiler works... yes
checking for C++ compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C++ compiler... yes
checking whether g++ accepts -g... yes
[...]

PR #990, opened 2016-05-03
Update README.md

nikete:
aded a fix suggested by @kamal94 in https://github.com/JohnLangford/vowpal_wabbit/issues/965 (which I just tested and works, while the the previous instructions did not on OS X 10.10 )


nikete:
sorry, i just did this from a clean pull of master and it failed

PR #992, opened 2016-05-04
--cb: fix one eligible action for test sample

pierce1987:
John,
This is a fix for the small bug for --cb we discussed, where if only one action is provided without cost or prob, it should be treated as only one action is eligible (and prediction should be this action). 

The problem was, when generating cost-sensitive example, when ld.costs.size() == 1, all actions are generated. This is valid for sth like 3:1:0.5 | features, but not consistent for example with test label, such as 3 | features.

So the fix is to generate all actions when (ld.costs.size() == 1 && !is_test_label(ld)). This is added for all 3 cb algorithms. 

after fix:
 echo "3:1:0.5 | three
3 | three" | ./vowpalwabbit/vw --cb 5 --cb_type ips
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = 
num sources = 1
0.000000 0.000000            1            1.0    known        1        2
0.000000 0.000000            2            2.0  unknown       **3**       2

finished run
number of examples per pass = 2
passes used = 1
weighted label sum = 0.000000
average loss = 0.000000

Please let me know if there are any concerns.

### DIFF
--- a/vowpalwabbit/gen_cs_example.cc
+++ b/vowpalwabbit/gen_cs_example.cc
@@ -40,7 +40,7 @@ void gen_cs_example_ips(cb_to_cs& c, CB::label& ld, COST_SENSITIVE::label& cs_ld
 { //this implements the inverse propensity score method, where cost are importance weighted by the probability of the chosen action
   //generate cost-sensitive example
   cs_ld.costs.erase();
-  if (ld.costs.size() == 1)   //this is a typical example where we can perform all actions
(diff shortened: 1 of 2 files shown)


JohnLangford:
appveyor appears to be failing.  Do you know why?

JohnLangford:
This looks like a good fix to me.

JohnLangford:
Merged in.  We'll deal with whatever appveyor fallout there is.

PR #994, opened 2016-05-06
Changes to support Python 3

irustandi:
Tested on Anaconda Python running Python 2.7 and Python 3.5 running on Ubuntu. In python/Makefile, PYTHON_VERSION needs to be modified to the appropriate version. Learning_to_Search.ipynb not converted yet (when run on Python 2.7 and 3.5, it crashes the kernel).


python/sklearn_vw.py
```diff
@@ -222,8 +221,8 @@ def __init__(self,
 
         # assign all valid args to params dict
         args = dict(locals())
-        for k, v in args.iteritems():
```
gramhagen: you can use the six compatibility library here:
```
import six
...
for k, v in six.iteritems(args):
```

python/sklearn_vw.py
```diff
@@ -222,8 +221,8 @@ def __init__(self,
...
-            if k != 'self' and v is not None:
+        for k, v in args.items():
```
gramhagen: you can use the six compatibility library here:
```
import six
...
for k, v in six.iteritems(args):
```

python/pyvw.py
```diff
@@ -61,9 +61,8 @@ def format(key,val):
             s = ('-'+key) if len(key) == 1 else ('--'+key)
             if type(val) is not bool or val != True: s += ' ' + str(val)
             return s
-        l = [format(k,v) for k,v in kw.iteritems()]
+        l = [format(k,v) for k,v in kw.items()]
```
gramhagen: same comment as above about using six.iteritems()

gramhagen:
I get this error trying to build:

irustandi:
Check and see if you're linking against libboost-python.so built for Python 3.5 or for other versions of Python. Very likely you're linking against libboost-python.so built for Python 2.7.

python/Makefile
```diff
@@ -8,7 +8,8 @@
 #  $(error Please run 'make' at the top level only)
 #endif
 
-PYTHON_VERSION = 2.7
+# change this to 2.7 for Python 2.7
+PYTHON_VERSION = 3
 PYTHON_INCLUDE = $(shell python$(PYTHON_VERSION)-config --includes)
 PYTHON_LDFLAGS = $(shell python$(PYTHON_VERSION)-config --ldflags)
```
gramhagen: I had to change this line to get it to build appropriately for 3 or 2.7
```
ifeq "$(PYTHON_VERSION)" "3"
	PYTHON_LIBS = -l boost_python3
else
	PYTHON_LIBS = -l boost_python
endif
```

gramhagen:
yes that was it, thanks

gramhagen:
also for posterity, if we're going to switch to python3 as the default in the Makefile when people run into this error:
```
ImportError: dynamic module does not define module export function (PyInit_pylibvw)
```
it means you are trying to run the pylibvw compiled for python3 with python2

python/Makefile
irustandi: The "-l boost_python3" part cannot be counted on to work in all systems. In my Ubuntu system using Boost from the APT package libboost-all-dev, there's no libboost_python3.so. libboost_python.so is just a symbolic link that originally points to Boost Python library for Python 2.7, I manually changed this symbolic link to point to Boost Python library for Python 3.5.

gramhagen: oh, that's unfortunate.  I wonder if there's another way to find the correct library automatically?

JohnLangford:
@hal3 Can you take a close look at this?

JohnLangford:
I want to have a broader discussion on the list about this one.  My
understanding is that there are presently two drawbacks:

1) Python 2.7 interface will no longer compile by default, although you can
change a flag here:
https://github.com/JohnLangford/vowpal_wabbit/pull/994/files#diff-0c67e320859c05045e7211eb3d0e3cf6R12
to make it work.
2) This patch disables the learning to search stuff.

W.r.t. 1, can the right version of python be autodetected rather than
defined?

What is needed to make problems with 2 go away?  I realize that a python
callback to a C++ interface from python is a bit hairy, but it's also
pretty cool and something that I'd like to see not break.

More generally, there is a question about which version of python we should
be targeting.  I don't know what the prevalent usage pattern is so advice
is desirable.

-John

JohnLangford:
My impression is that most people still use 2.7 and not 3. It would be 
great to support 3, but I think 2.7 is the priority. (I could be wrong.)

For me, at least, it's very important not to lose l2s.

gramhagen:
wrt 1)
I think the python community is approaching the tipping point for migration to python3.  Many greenfield projects are starting with python3, but there is still a lot of institutional code on 2.7.  Official support for python 2.7 will expire in 2020.  IMO to avoid breaking changes with current users it would be safer to default to 2.7 but offer instructions on how to compile for 3.
You can certainly detect which version of python is the default in the system from the Makefile, but the problem lies in resolving the appropriate boost dependencies across different system types / installations.
It might more effective to be able to install a python vowpal_wabbit package using pip or conda, which could manage the versions and dependencies more appropriately.  I have not had much time to purse setting either of those up though.

irustandi:
Compared to say a couple of years ago, support for Python 3 (3.4 onwards) in the popular numerical packages (e.g. numpy, scipy, sklearn) has definitely matured that there are no major obstacles that I'm aware of using those packages in Python 3. This page also adds some context on the recent (as of August 2015) status of Python 2 vs 3:

[http://ncoghlan-devs-python-notes.readthedocs.io/en/latest/python3/questions_and_answers.html](url)

The ideal case is to support both Python 2.7 and 3.x equally so that there are no discernible differences when using either Python version, I think this is now the case in numpy, scipy, and sklearn. I also think it is a good idea to manage the installation of the Python package of VW through setuptools, which is typically used in most Python packages these days. Time permitting, I can look into this.

JohnLangford:
Please do look into Python 2.7+3 support.   That sounds quite good if we can get it.

gramhagen:
I just added some improvements to get the wrapper to work with setuptools and ultimately be installable using pip via PyPI.  https://github.com/JohnLangford/vowpal_wabbit/pull/1012

Will need to merge changes made here into that if it is pulled in.  

Also, regardless of the change to the Makefile we should merge in the python3 updates since they are still compatible with Python 2.  At least we can move forward there, and then we can work through how to resolve building pylibvw.so appropriately for Python 2-3 as a separate issue.

JohnLangford:
Closing as this one has been incorporated in #1052 .

PR #997, opened 2016-05-23
Recall tree clean

pmineiro:
cleaned up code plus aloi demo


JohnLangford:
Can you tweak the windows build to compile?

JohnLangford:
Instead of a demo, is it possible to do bigtests?

pmineiro:
windows build: probably not, I'd be guessing, I don't have a working
compilation environment for that.

bigtests: I can keep the demos as-is but add a "gauss.c" style dataset as a
test case so that the coverage is mitigated.

p

JohnLangford:
A little test could go in RunTests (and that seems desirable).  But what is
the benefit of having something in demo but not bigtests?

-John

pmineiro:
I think of demos as guided tours of the software explaining how to apply to
your own problem via examples.  Not really a test per se: certainly not
intended to be run ballistically.  Ideally it would be a python notebook to
encourage modification for learning.

Also, I want to get the complete paper repro scripts in demo/ so we can put
the github url into the final version of the paper.

p

pmineiro:
The remaining windows build issues are not recall tree specific.

JohnLangford:
What's the status of this?  Have you tried valgrind on the linux version? 

W.r.t. demo vs. bigtests, I think it's valuable to have a large number of learning problems whose performance can be triggered via 'make' so that it is easy to test the learning primitive for the purpose of improving it.

pmineiro:
The status is.... I live on the west coast and I went to sleep.

😀

p

Skickat från min Nexus 6p

JohnLangford:
Sounds good :-)

-John

pmineiro:
So far so good ... let me try the "empty output with no reference" fix next.

pmineiro:
YAHOOOOO! You're all clear, kid. Now let's blow this thing and go home.

JohnLangford:
Merged in finally.   Thanks much.

All, the recall tree that Paul checked in has sometimes yielded 
prediction performance superior to one-against-all.

-John

PR #998, opened 2016-05-24
Pr997

eisber:
Windows fix


JohnLangford:
My understanding is that this is best done as a pull request against Paul's branch?

PR #1000, opened 2016-05-26
Add formatting to code

jeffwidman:
I wasn't 100% sure of the code formatting convention used here, it looks like there were several variants used throughout this readme. I can update if needed.

PR #1001, opened 2016-05-26
Adding support for poisson regression. 

sharatsc:
Adding support for Possion regression. Does not support --invariant mode yet. Haven't figured out the closed form for the update. Currently works with --normalized as well as --bfgs

Usage documented in a python notebook:
https://github.com/sharatsc/vowpal_wabbit/blob/c1e95f98038c628d0ff3995fb3d85bf7c35eefa3/python/poisson_regression.ipynb


sharatsc:
Coverage is off by rounding error: -0.003%

Appveyor is failing on the same location as an unrelated pull request 
https://github.com/JohnLangford/vowpal_wabbit/pull/1000

This could be a cs code issue (or CLR version difference b/w Travis and the CS wrapper). Can @eisber somebody take a look?

JohnLangford:
This seems good except that getUpdate is not safe.  In particular, a very large learning rate/importance weight will result in the update going crazy.  Have you considered designing a safe update along the lines of this paper: http://arxiv.org/abs/1011.1576 ?  This has been a clear win experimentally.

sharatsc:
Table 1 does list a closed form solution for Poisson. I'll try to work it out. In the mean time, it does work with --normalized and --bfgs.

JohnLangford:
Nikos says he has it worked out.

fest:
Hi folks,

Yes the paper does not list Poisson regression but it can be derived by solving the ODE. Then you have to translate between paper and VW which I believe can be summarized in the following changes:
- VW computes -s() and the table lists s()
- VW replaces (learning rate * importance) by update_scale
- VW replaces norm(x)^2 by pred_per_update

With these in mind and some help from alpha
https://www.wolframalpha.com/input/?i=solve+y%27(t)%3Da*(exp(b-y(t)*c)-d),+y(0)%3D0
my recommendation is
```
float getUpdate(float prediction, float label, float update_scale, float pred_per_update) {
    if (label > 0) {
      return label * update_scale - log1p(exp(prediction)*expm1(label * update_scale * pred_per_update)/label)/pred_per_update;
    }
    else {
       return - log1p(exp(prediction) * update_scale * pred_per_update)/pred_per_update;
    }
  }
```

I believe I have done this carefully enough but I have not tested it. Sharat can you verify that this update is at least on par with yours?

JohnLangford:
Sharat have you had a chance to look at Nikos's update?

sharatsc:
Looking at it now.

eisber:
You need to use # Test 133: and not # Train 133:

sharatsc:
Nikos,
 I was able to derive the same. Derivation outlined here: https://dl.dropboxusercontent.com/u/1897767/Implementing%20poisson%20loss%20in%20VW.pdf

The expression in notation used in the original paper (http://bit.ly/1Us0oVe) vs. your derivation (http://bit.ly/25ZG6Gp)

JohnLangford:
Can you implement today? I want to drop a new version soon.

sharatsc:
safe update changes incorporated along fixing with @eisber's comments. The windows build seems to be broken (also on master) 

Windows build failing with unrelated error
Assert.AreEqual failed.
 Expected:<--no_stdin --max_prediction 1 --bit_precision 18 --cb_adf --cb_type ips --rank_all --csoaa_ldf multiline --interact ud --csoaa_rank>. 
Actual:<--no_stdin --max_prediction 1 --bit_precision 18 --cb_adf --cb_type ips --rank_all --csoaa_ldf multiline --csoaa_rank --interact ud>.

eisber:
My PR has fixes for this… essentially replace the expected with the actual string in TestArguments.cs

sharatsc:
Windows build has gotten quite flaky. We should consider spinning off cs wrapper as a separate project that consumes the vw nuget? Build time is taken up with cs and cs-tests

JohnLangford:
This looks good to go to me.  Not sure you should worry about the windows-side tests as @eisber is working on that.

sharatsc:
@JohnLangford this is ready to go. Fixed windows errors as well.

JohnLangford:
This is merged in, thanks.   All, we now have Poisson loss working thanks
to Sharat and Nikos.

@eisber there are some minor changes to make the windows test work.

-John


On Wed, Jun 15, 2016 at 10:33 AM, Sharat Chikkerur <notifications@github.com

trufanov-nok:
Hello,
Does poisson loss supposed to support negative labels? Like {-1,1}.
According to `poisson_loss::getUpdate()` it does as it [has](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/loss_functions.cc#L306) a special case for negative labels.
But `poisson_loss::getLoss()` returns 0 for -1 labels and I'm getting no weights update. Looks like bcs `logf(label + 1e-6)` is `NaN` [here](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/loss_functions.cc#L300). So it doesn't support labels that is < 0 ...

fest:
It should not support labels <0 as the prediction is exp(w*x). The
assumption is that labels are >=0 so the else part is to avoid division by
0.

Nikos

trufanov-nok:
Ok, then it may have sense to put a label check and a warning in its `getLoss()` as it's done for `logloss`...

PR #1004, opened 2016-06-03
Action/score fix

eisber:


cs/cli/vw_prediction.cpp
```diff
@@ -93,21 +93,38 @@ namespace VW
...
+      auto length = ex->pred.a_s.size();
+      auto values = gcnew cli::array<ActionScore>(length);
+
+      auto index = 0;
+      for (auto& as : ex->pred.a_s)
+      {
+          values[index].Action = as.action;
+          values[index].Score = as.score;
+          index++;
+      }
```
lhoang29: if you used 'for each' here it should be ok since you're not modifying 'as', right?

cs/cs/VowpalWabbitMultiLine.cs
```diff
@@ -424,7 +424,7 @@ public static class VowpalWabbitMultiLine
             int i = 0;
             foreach (var index in values)
             {
-                result[i++] = validActionDependentFeatures[index];
+                result[i++] = validActionDependentFeatures[(int)index.Action];
```
lhoang29: index.Action is 1-based right? are you using 1-based index here too?

cs/cli/vw_prediction.cpp
eisber: Actually not. So the weird thing is that ‘for each’ modifies the ex->pred.a_s._begin. Not sure why. I think one is only supposed to use it for IEnumerable. That’s the reason I referred to it as magic operator ☺

cs/cs/VowpalWabbitMultiLine.cs
eisber: This is the new action/score feature John implemented. This is 0-based. Only exploration library has 1-based actions. VW prediction output was always 0-based.

Markus

lhoang29:
other changes look good

PR #1006, opened 2016-06-06
cb_explore_adf 

alekh:


vowpalwabbit/cb_adf.cc
```diff
@@ -208,14 +210,14 @@ void call_predict_or_learn(cb_adf& mydata, base_learner& base, v_array<example*>
...
-      examples[0]->pred.multiclass = action;
-  }
+  // if (!mydata.rank_all)
+  //   { uint32_t action = 0;
+  //     for (size_t i = 0; i < examples.size(); i++)
+  // 	if (!CB::ec_is_example_header(*examples[i]) && !example_is_newline_not_header(*examples[i]))
+  // 	  if (examples[i]->pred.multiclass != 0)
+  // 	    action = examples[i]->pred.multiclass;
+  //     examples[0]->pred.multiclass = action;
+  // }
```
lhoang29: do you want to keep these comments?

vowpalwabbit/cb_adf.cc
```diff
@@ -269,14 +271,14 @@ void gen_cs_example_MTR(cb_adf& data, v_array<example*>& ec_seq, v_array<example
...
   if (predict) //first get the prediction to return
   { gen_cs_example_ips(examples, mydata.cs_labels);
     call_predict_or_learn<false>(mydata, base, examples, mydata.cb_labels, mydata.cs_labels);
-    if (!mydata.rank_all) //preserve prediction
-      action = examples[0]->pred.multiclass;
-    else
-      std::swap(examples[0]->pred.a_s, mydata.a_s);
+    //    if (!mydata.rank_all) //preserve prediction
+    //      action = examples[0]->pred.multiclass;
+    //    else
```
lhoang29: same as above

vowpalwabbit/cb_adf.cc
```diff
@@ -367,7 +369,7 @@ void output_example(vw& all, cb_adf& c, example& ec, v_array<example*>* ec_seq)
 
   float loss = 0.;
 
-  uint32_t action = ec.pred.multiclass;
+  uint32_t action = ec.pred.a_s[0].action;
```
lhoang29: lacking a bit of context on pred.a_s, but is it always guaranteed to contain at least 1 item?

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -0,0 +1,484 @@
...
+      cerr<<"Received predictions of wrong size from CB base learner\n";
+
+    if(!is_learn || !data.learn_only) {
+      float prob = data.epsilon/(float)num_actions;
+      for(int i = 0;i < num_actions;i++) {
+	action_score a_s;
+	a_s.action = preds[i].action;
+	a_s.score = prob;
+	data.id_probs.push_back(a_s);
+      }
```
lhoang29: nit: there are some mixed use of tabs and spaces here so it's hard to read, can you run the styling scripts found in utl/new_version?

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -0,0 +1,484 @@
...
+      float prob = 1.0/(float)data.bag_size;
+      data.id_probs[0].score += prob;
+      for(size_t i = 1;i < data.bag_size;i++) {
+	for(example* ec : examples)
+	  base.predict(*ec, i);
+	uint32_t chosen = preds[0].action;
+	for(size_t i = 0;i < num_actions;i++)
+	  if(data.id_probs[i].action == chosen)
+	    data.id_probs[i].score += prob;
+      }
```
lhoang29: In the exploration library, we randomly choose a bag and take that bag's action as the chosen action. Here it seems that the first bag is used instead, which I think is ok but is that intended? Additionally, the prob for the first entry is added in line 170, wouldn't this end up double counting as in 177? I'm probably missing something here, so clarifications would be appreciated.

lhoang29:
is it possible to use exploration library here instead of having to rewrite the logic (e.g. choosing action, computing probability)?

vowpalwabbit/cb_explore_adf.cc
alekh: Hi Louie,

What is happening here is that each bagged predictor returns an
action_score vector. Each such vector scores all actions in sorted order.
So the action in the position 0 is the predicted action.

Basically, you call predict with index i to obtain the prediction of the
bagged predictor i. The 0th entry of action_score is then the predicted
action we want, and we then create a uniform probability distribution over
these actions.

Does that make sense?

Also, the reason we cannot use the exploration library here is because it
does not return a distribution over the actions which we would like to
return from here. So the idea is that this would be used in the exploration
library rather than this using the exploration library.

Alekh

vowpalwabbit/cb_adf.cc
alekh: It should, unless there are no actions for some bizarre reason.

Alekh

vowpalwabbit/cb_adf.cc
alekh: These comments were mostly left for John to see since they involved
modifying the return type from cb_adf when --rank_all is not specified.
They can be removed.

Alekh

vowpalwabbit/cb_explore_adf.cc
lhoang29: re: bag: Ah that makes sense, I missed the sorted order assumption. Also, the double counting is clearer upon closer look as i starts from 1. 

re: exploration library: Right, I think we're just talking about where the logic to compute prob should live (in exploration lib or vw). Let me try to find you offline to clarify.

PR #1009, opened 2016-06-08
Implementation using base C++ array internally

aartibagul:


vowpalwabbit/vw.sln
```diff
@@ -7,12 +7,6 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "libvw", "libvw.vcxproj", "{
 		{1E205806-7F80-47DD-A38D-FC08083F3592} = {1E205806-7F80-47DD-A38D-FC08083F3592}
 	EndProjectSection
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "cs_test", "..\cs\test\cs_test.csproj", "{D5E462FC-3DD6-4B65-A9E9-DA0B0B11D254}"
```
akshaya-a: let's work with @eisber to get the VS build working without removing the clr-based projects from the sln

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,65 @@
+#pragma once
+#include <stdio.h>
+#include <iostream>
+#include <stdlib.h>
+#include <string.h>
+
+using namespace std;
+typedef float weight;
+
+class array_parameters
```
eisber: alternative name dense_parameters?

vowpalwabbit/gd.h
```diff
@@ -27,17 +28,17 @@ struct multipredict_info { size_t count; size_t step; polyprediction* pred; regr
...
 
   fi &= mask;
   uint64_t top = fi + (uint64_t)((mp.count-1) * mp.step);
   if (top <= mask)
-  { weight* last = w + top;
+  { array_parameters last = w + top;
     w += fi;
     for (; w <= last; w += mp.step, ++p)
-      p->scalar += fx **w;
+      p->scalar += fx * w.value();
```
eisber: implement operator*()

vowpalwabbit/parser.cc
```diff
@@ -455,9 +455,9 @@ void enable_sources(vw& all, bool quiet, size_t passes)
                      PROT_READ|PROT_WRITE, MAP_SHARED|MAP_ANONYMOUS, -1, 0);
 
       size_t float_count = all.length() << all.reg.stride_shift;
-      weight* dest = shared_weights;
-      memcpy(dest, all.reg.weight_vector, float_count*sizeof(float));
-      free(all.reg.weight_vector);
+      array_parameters dest(shared_weights);
+      memcpy(dest.begin, all.reg.weight_vector.begin, float_count*sizeof(float));
```
eisber: why isn't this a method?

vowpalwabbit/search_graph.cc
```diff
@@ -83,10 +85,14 @@ struct task_data
   vector<size_t>   pred;  // predictions
   example*cur_node;       // pointer to the current node for add_edge_features_fn
   float* neighbor_predictions;  // prediction on this neighbor for add_edge_features_fn
-  weight* weight_vector;
+  array_parameters weight_vector;
   uint32_t* confusion_matrix;
   float* true_counts;
   float true_counts_total;
+
+  task_data() :weight_vector(nullptr)
```
eisber: in general I'd not assume constructors being called.
most of the structs are allocated using calloc()

aartibagul: The constructor is being called at line 101 in the same file.
`task_data * D = new task_data();`

cs/cli/vw_clr.vcxproj
```diff
@@ -310,9 +310,6 @@
     <ProjectReference Include="$(SolutionDir)\vw_dynamic.vcxproj">
       <Project>{1e205806-7f80-47dd-a38d-fc08083f3593}</Project>
     </ProjectReference>
-    <ProjectReference Include="..\common\vw_common.csproj">
-      <Project>{e621e022-c1f8-433f-905a-ab9a3de072b7}</Project>
-    </ProjectReference>
```
JohnLangford: This should not be changing?

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,65 @@
...
+{
+private:
+	weight* _begin;
+
```
JohnLangford: You probably want all the elements here:
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/global_data.h#L117 folded into this class.

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,65 @@
...
+public:
+	array_parameters(weight* begin)
```
JohnLangford: Maybe this constructor should not exist?  It seems dangerous to allow external allocation.  I'd rather be absolutely positive that the mask matches the length of the array.

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,65 @@
...
+		: _begin(begin)
+	{ }
+
+	array_parameters(const array_parameters& other)
+		: _begin(other._begin)
+	{ }
+	
+	weight& operator[](size_t i) const { return _begin[i]; }
```
JohnLangford: Inline?

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,65 @@
...
+
+	inline weight*& begin() { return _begin; }
+
+	template<class T> array_parameters operator+(T index) { return array_parameters(_begin + index); }
```
JohnLangford: I'm confused.  What is this one for?

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,65 @@
...
+
+	array_parameters& operator++()
+	{
+		_begin++;
+		return *this;
```
JohnLangford: This operator seems dangerous to me.  You either do a copy/increment/discard or craziness happens when _begin increments.   I think we want an iterator which is best done using stride and with a different type.

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,65 @@
...
+		return *_begin;
+	}
+	
+	template<class T> array_parameters& operator+=(T index)
+	{
+		_begin += index;
+		return *this;
+	}
+
+	bool operator<(const array_parameters& other)
```
JohnLangford: Overloading the type of the array parameters with the type of an element seems semantically awkward.  Can you just have the return type be a weight* or weight&?

vowpalwabbit/accumulate.cc
```diff
@@ -19,18 +19,22 @@ using namespace std;
...
-  float* local_grad = new float[length];
-  weight* weights = reg.weight_vector;
-  for(uint32_t i = 0; i < length; i++)
-    local_grad[i] = weights[stride*i+o];
-
-  all_reduce<float, add_float>(all, local_grad, length);
-  for (uint32_t i = 0; i < length; i++)
-    weights[stride*i+o] = local_grad[i];
-  delete[] local_grad;
+  weight_vector local_grad(length);
```
eisber: +1 for proper exception handling

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,161 @@
...
+	bool operator==(const weights_iterator& rhs) { return _current == rhs._current; }
+	bool operator!=(const weights_iterator& rhs) { return _current != rhs._current; }
+
+	//to iterate within a bucket
+	w_iter begin() { return w_iter(_current); }
+	w_iter end(size_t offset) { return w_iter(_current + offset); }
+
+	~weights_iterator(){
+		if (_current != nullptr)
+			free(_current);
```
eisber: bug

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,161 @@
...
+	
+	bool isNull() const { return _begin == nullptr; }
+
+	//weight_vector(const weight_vector& other)
+	//	: _begin(other._begin)
+	//{ }
+
+	inline weight& operator[](size_t i) const { return _begin[i & _weight_mask]; }
+
+	uint64_t getMask()
```
eisber: uint64_t mask()

stride()
stride(int foo)

vowpalwabbit/audit_regressor.cc
```diff
@@ -46,28 +46,25 @@ inline void audit_regressor_interaction(audit_regressor_data& dat, const audit_s
 
 inline void audit_regressor_feature(audit_regressor_data& dat, const float /*ft_weight*/, const uint64_t ft_idx)
 {
-    size_t index = ft_idx & dat.all->reg.weight_mask;
-    weight* weights = dat.all->reg.weight_vector;
+    weight_vector weights = dat.all->wv;
```
eisber: copy of vector

vowpalwabbit/bfgs.cc
```diff
@@ -106,20 +106,16 @@ const char* curv_message = "Zero or negative curvature detected.\n"
...
 void zero_preconditioner(vw& all)
 { //set derivative to 0.
-  uint32_t length = 1 << all.num_bits;
-  size_t stride_shift = all.reg.stride_shift;
-  weight* weights = all.reg.weight_vector;
-  for(uint32_t i = 0; i < length; i++)
-    weights[(i << stride_shift)+W_COND] = 0;
+  weight_vector weights = all.wv;
+  for (weight_vector::iterator i = weights.begin(W_COND); i != weights.end(W_COND); ++i)
+	  *i = 0;
```
eisber: for (auto& i : all.wv.offset(_COND))
   *i = 0;

vowpalwabbit/bfgs.cc
```diff
@@ -197,48 +193,51 @@ double regularizer_direction_magnitude(vw& all, bfgs& b, float regularizer)
...
 
-  uint32_t length = 1 << all.num_bits;
-  size_t stride_shift = all.reg.stride_shift;
-  weight* weights = all.reg.weight_vector;
+  weight_vector weights = all.wv;
+  
   if (b.regularizers == nullptr)
-    for(uint32_t i = 0; i < length; i++)
-      ret += regularizer*weights[(i << stride_shift)+W_DIR]*weights[(i << stride_shift)+W_DIR];
+	  for (weight_vector::iterator i = weights.begin(W_DIR); i != weights.end(); ++i)
```
eisber: bug: weights.end(W_DIR)

vowpalwabbit/parse_args.cc
```diff
@@ -1477,8 +1477,8 @@ void finish(vw& all, bool delete_all)
   { all.l->finish();
     free_it(all.l);
   }
-  if (all.reg.weight_vector != nullptr && !all.seeded) // don't free weight vector if it is shared with another instance
-    free(all.reg.weight_vector);
+  if (!all.wv.isNull() && !all.seeded) // don't free weight vector if it is shared with another instance
```
eisber: remove destructor call

vowpalwabbit/accumulate.cc
```diff
@@ -19,18 +19,21 @@ using namespace std;
```
JohnLangford: @alekh accumulate isn't factored right to support sparse models.  Can you think about how to do it right?

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,181 @@
...
+#ifndef _WIN32
+#include <sys/mman.h>
+#endif
+
+using namespace std;
+typedef float weight;
+
+class weight_vector;
+
+class weights_iterator_iterator
```
JohnLangford: This class seems a bit opaque.  Why do we need it?

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,181 @@
...
+
+	weights_iterator(weight* current, uint32_t stride_shift)
+		: _current(current), _stride_shift(stride_shift)
+	{ }
+
+	weight& operator*() { return *_current; }
+
+	weights_iterator& operator++()
+	{
+		_current += _stride_shift;
```
JohnLangford: stride_shift seems to be used as a stride so it should be renamed.

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,181 @@
...
+
+};
+
+class weight_vector //different name? (previously array_parameters)
+{
+private:
+	weight* _begin;
+	uint64_t _weight_mask;  // (stride*(1 << num_bits) -1)
+	uint32_t _stride_shift;
+	size_t _size;
```
JohnLangford: Is _size redundant with _weight_mask?

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,181 @@
...
+	bool _seeded;
+
+public:
+	typedef weights_iterator iterator;
+
+	weight_vector()
+		: _begin(nullptr), _weight_mask((uint64_t)LONG_MAX), _size(0), _stride_shift(0),_seeded(false)
+	{ }
+
```
JohnLangford: Why do we need a non-constructor like this?

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,181 @@
...
+};
+
+class weight_vector //different name? (previously array_parameters)
+{
+private:
+	weight* _begin;
+	uint64_t _weight_mask;  // (stride*(1 << num_bits) -1)
+	uint32_t _stride_shift;
+	size_t _size;
+	bool _seeded;
```
JohnLangford: What is _seeded supposed to mean?

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,181 @@
...
+	//{ }
+
+	inline weight& operator[](size_t i) const { return _begin[i & _weight_mask]; }
+
+	uint64_t mask()
+	{ return _weight_mask;
+	}
+
+	uint32_t stride()
+	{ return _stride_shift;
```
JohnLangford: stride / _stride_shift confusion again.

vowpalwabbit/bfgs.cc
```diff
@@ -106,20 +106,16 @@ const char* curv_message = "Zero or negative curvature detected.\n"
...
 void zero_derivative(vw& all)
 { //set derivative to 0.
-  uint32_t length = 1 << all.num_bits;
-  size_t stride_shift = all.reg.stride_shift;
-  weight* weights = all.reg.weight_vector;
-  for(uint32_t i = 0; i < length; i++)
-    weights[(i << stride_shift) +W_GT] = 0;
+  weight_vector& weights = all.wv;
+  for (weight_vector::iterator i = weights.begin(W_GT); i != weights.end(W_GT); ++i)
+	  *i = 0;
```
JohnLangford: You know something is going right when the number of lines of code decreases :-)

vowpalwabbit/array_parameters.h
aartibagul: I used _seeded in the destructor to make sure that the weight vector is not freed if it is shared with another instance

JohnLangford:
Overall, this is looking much closer to what we need so the current issue is stabilizing the code.

vowpalwabbit/array_parameters.h
JohnLangford: Ok, makes sense.

-John

vowpalwabbit/accumulate.cc
alekh: How are sparse models represented?

Alekh

vowpalwabbit/array_parameters.h
aartibagul: I used the class when there is an outer for loop that moves with a stride and the inner loop moves within the bucket -
```
for (uint64_t j = 0; j < stride * length; j += stride)
 for (size_t k = 0; k < all->lda; k++)
   all->reg.weight_vector[j + k] = (float)(-log(frand48()) + 1.0f);  
```
(Line 634-642 in lda_core.cc)

vowpalwabbit/accumulate.cc
JohnLangford: Probably by a hashtable.

JohnLangford:
Why is this still failing the travis build?

aartibagul:
I hadn't modifed python\pylibvw.cc. It also gives an error on Test16-
```
RunTests: test 16: '/usr/bin/timeout 30 ../vowpalwabbit/vw -k -c -d train-sets/rcv1_small.dat --loss_function=logistic -b 20 --bfgs --mem 7 --passes 20 --termination 0.001 --l2 1.0 --holdout_off' timed-out (exitcode=124)
RunTests: test 16: you may increase the imposed time-out: $TimeOutSec=30
RunTests: exiting with status=124
make: *** [test_gcov] Error 124
```

JohnLangford:
If you sync to the master, can you reproduce the problem on your local machine?

JohnLangford:
Removing --always-make seems bad because it ends up reporting coverage 0%.

JohnLangford:
This is passing tests now.  Is it ready to go in your opinion?

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
+#pragma once
```
TomasPuverle: Hi Aarti,

I have been looking at your pull request, and it's a really nice piece of work.

I have a fair bit of experience writing C++ but am only getting started working with VW.  As such, I would like to offer some suggestions regarding design, coding style, and a few other observations.  They will range from from (hopefully!) useful, to nitpicky.

As such, please feel free to take on as much or as little of what I write.  From now on, I will keep my comments very short, but please don't mistake my brevity for criticism.  

My first comment will be a bit of a nitpick.  #pragma once is not standard C++, and as such, it isn't guaranteed to work on all compilers.  It turns out that it works well on Visual C++, but has historically not worked well on gcc.

It looks like the rest of the project is using it, so you are perhaps right to stick to the style.  At the same time, when writing completely new code, it may be worthwhile evolving the current style.  :D

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+#include <stdio.h>
+#include <iostream>
+#include <stdlib.h>
+#include <string.h>
```
TomasPuverle: I don't think any of these headers are really needed, with the possible exception of string.h (for memcpy below).  Having said that, the correct header names for C++ are actually <cstdio>, <stdlib> and <cstring>.

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+
+#ifndef _WIN32
+#include <sys/mman.h>
+#endif
+
+using namespace std;
```
TomasPuverle: You should never, ever, do "using namespace std" in a header file.  There are plenty of articles written about why, so I won't  go into it here.

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+typedef float weight;
```
TomasPuverle: I like the typedef.  On second thought, have you considered making your weight_vector class a template?

vowpalwabbit/array_parameters.h
TomasPuverle: Regarding your iterator classes:

You should really look at boost::iterator and boost::range libraries.  Your iterators are not complete in the sense of C++ iterators (for example, missing the -- operators) but also not having their iterator traits defined correctly.  That means they they may not inter-operate with the C++ standard algorithms properly.  Many of the algorithms in the STL will dispatch to different implementations depending on the capabilities of iterators, meaning that your iterators below could  lead to sub-optimal code (or they may not be able to be used at all).

Also, in this case, the iterator_iterator may not really be necessary at all (see my comments further on).

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+public:
+	weights_iterator_iterator(weight* cur)
+		: _cur(cur)
+	{ }
+	
+	weight& operator*() { return *_cur; }
+
+	weights_iterator_iterator& operator++()
+	{
+		_cur++;
```
TomasPuverle: Nitpick: This should really be ++_cur;

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+		return *this;
+	}
+
+	template<typename T>
+	weights_iterator_iterator operator+(T index) { return weights_iterator_iterator(_cur + index); }
+
+	template<typename T>
+	weights_iterator_iterator& operator+=(T index)
```
TomasPuverle: Why are you using a template here?  I am wondering if there's some problem you ran into which you were trying to address, but it shouldn't be necessary.

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+	{
+		_cur += index;
+		return *this;
+	}
+
+	weights_iterator_iterator& operator=(const weights_iterator_iterator& other)
```
TomasPuverle: Why did you define an operator=, but not a copy or move constructor?  For one, the assignment operator is not necessary, as it will do the same thing as the default generated one.
Also, in general, C++ has the rule of 3/5 - if you write a copy/move/assignment/move assignment/destructor, you should consider the writing the other functions, too!

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+	{
+		_cur = other._cur;
+		return *this;
+	}
+
+	bool operator==(const weights_iterator_iterator& rhs) { return _cur == rhs._cur; }
+	bool operator!=(const weights_iterator_iterator& rhs) { return _cur != rhs._cur; }
+
+};
+class weights_iterator
```
TomasPuverle: Many of my coding nitpicks apply to this weights_iterator class, too, so I won't repeat myself.   Honestly, use boost.iterator.  It will save you from having to write (and test!) all this iterator boilerplate.  It will also save others having to read an maintain it, which is always a big win.

However, let's discuss some design.  Specifically, your intention is to make it possible to iterate ranges of weights in a bucket.  As such, I would consider returning a *range* as the result of your iterator dereference - effectively, your signature would be e.g. boost::iterator_range<weight *, weight*> operator*().  This will allow you to write very clean code along the lines of:

    for (auto ww: *weight_bucket)
    {
       //
    }

There's also a downside - if you expect people to mostly access only the first element of the weight vector, you may have to write `(*iter)[0]` to access the first weight.  However, I would be cleaner to actually return a different iterator class from your `weights_vector::begin(offset)` function.  See more below.

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+{
+private:
+	weight* _current;
+	uint32_t _stride;
```
TomasPuverle: Nitpick: Strictly speaking, this should be std::ptrdiff_t, not uint_32.
Nitpick 2: Unless there's a good reason for it, you should always prefer to use abstract types such as int, long, or std::size_t instead of the bitness specific ones.
Nitpick 3: All of your uintX_t should really be std::uintX_t (at least in header files)

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+	//to iterate within a bucket
+	w_iter begin() { return w_iter(_current); }
+	w_iter end(size_t offset) { return w_iter(_current + offset); }
+
+};
+
+class weight_vector //different name? (previously array_parameters)
+{
+private:
+	weight* _begin;
```
TomasPuverle: In modern C++, you should really never use raw pointers.  There are so many better alternatives.  First and foremost, because you didn't follow the rule of 5, it means that that copying/asisgning your class will lead to memory problems.  You should either:
a) Use a more specialized type for your weights
b) Implement/disable copy/move constructors and operators=.

For example, you could just write

weight_vector(const weight_vector &) = delete; //disable copying.

But even better, use a standard component.  If you prefer not to use std::vector(because you want to do memory mapping below), at least consider using std::unique_ptr.  You won't have to worry about writing a destructor and your copy/assignment will also suddetly become correct.

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+	uint64_t _weight_mask;  // (stride*(1 << num_bits) -1)
+	uint32_t _stride_shift;
+	bool _seeded;
+
+public:
+	typedef weights_iterator iterator;
+
+	weight_vector()
+		: _begin(nullptr), _weight_mask((uint64_t)LONG_MAX), _stride_shift(0),_seeded(false)
```
TomasPuverle: It is best to avoid having default constructors like this which don't really do anything.  But, if you feel you need it, you may want to consider an alternative: In C++ 11, you can also just set the member variables in the class, e.g.
bool _seeded = false;

Then you won't even need this.

Second, the correct C++ way of saying `_weight_mask((uint64_t)LONG_MAX)` is `_weight_mask(std::numeric_limits<type>::max())`.

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+	{ }
+
+	weight_vector(size_t length, uint32_t stride_shift=0)
+		: _begin(calloc_mergable_or_throw<weight>(length << stride_shift)),
+		_weight_mask((length << stride_shift) - 1),	
+		_stride_shift(stride_shift),
+		_seeded(false)
+	{ }
+
+	inline weight* first() { return _begin; } //TODO: Temporary fix for lines like (&w - all.reg.weight_vector). Needs to change for sparse.
```
TomasPuverle: Here, you don't need the `inline`, as it's already implied by the in-class definition.  Second, it C++ this function is usually called `.data()` (see, for example, http://en.cppreference.com/w/cpp/container/vector/data)

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+	iterator begin() { return iterator(_begin, 1); }
+	iterator end() { return iterator(_begin + _weight_mask + 1, 1); }
+
+	//iterator with stride and offset
+	iterator begin(size_t offset) { return iterator(_begin + offset, (1<<_stride_shift)); }
+	iterator end(size_t offset) { return iterator(_begin + _weight_mask + 1 + offset, (1 << _stride_shift)); }
+
```
TomasPuverle: I have a number of comments regarding the iterator interfaces.  First of all, you should consider adding a const_iterator const versions of these functions, so programmers can express their desire to only read the data.   Also, consider providing a range interface for your `begin` and `end` (especially with offsets), so it's possible to write something like `for (auto val: vw.read(offset)) { }`.  They say that one shouldn't repeat themselves in code, and having to call `begin(offset)` and `end(offset)` certainly qualifies as that.

Finally, one more thing: Your rewrite of vw using your weight_vector class has brought up an interesting observation:  There are a lot of instances of loops which only iterate over a specific offset.  Now that you have this beautiful abstraction, you may want to consider the cache behaviour of your code.  If you're always iterating a specific offset in a bucket, it would be much more efficient to bucket the offsets together, i.e. instead of

w11 w12 w13 w21 w22 w23 ...

you could do

w11 w21 ... w12 w22 ... w13 w23...

As far as your clients are concerned, nothing changes because you are still presenting the same interface, but the computation could be a lot faster.

From my cursory review of VW, it seems that the layout may need to be algorithm dependent, but I still thought it's worth mentioning.

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+	inline weight* first() { return _begin; } //TODO: Temporary fix for lines like (&w - all.reg.weight_vector). Needs to change for sparse.
+	iterator begin() { return iterator(_begin, 1); }
+	iterator end() { return iterator(_begin + _weight_mask + 1, 1); }
+
+	//iterator with stride and offset
+	iterator begin(size_t offset) { return iterator(_begin + offset, (1<<_stride_shift)); }
+	iterator end(size_t offset) { return iterator(_begin + _weight_mask + 1 + offset, (1 << _stride_shift)); }
+	
+	bool isNull() const { return _begin == nullptr; }
```
TomasPuverle: Typically, functions such as this are an indication that something is not quite right in your design.  Also, the only places I see these used in the code could be better served by having a properly written move constuctor and move assignment operator (in parse_regressor.cc).

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+
+	inline weight& operator[](size_t i) const { return _begin[i & _weight_mask]; }
```
TomasPuverle: Nitpick: std::size_t.
Also, try to avoid using single letter variable names.  They are IMPOSSIBLE to find in your code.  Try searching for 'i' in this code.  ;)

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+	
+	void mask(uint64_t weight_mask)
+	{ _weight_mask = weight_mask;
+	}
+	
+	bool seeded()
+	{ return _seeded;
+	}
+	
+	void seeded(bool seeded)
```
TomasPuverle: It seems that one of the primary goals of `seeded` is to tell your class that the weight_vector is shared, (so that it doesn't get deleted in the destructor).  But in that case, would it not be better to let the share function to set it?

Second, even though your class does a good job of trying to clean up memory, it just leaks it in the case of the mmap() and `_seeded` being true.

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+	{ _seeded = seeded;
+	}
+
+	void share(size_t length)
```
TomasPuverle: I think the `#ifdef _WIN32_` should be around the entire function, not just the code in the body.  Imagine you are a user of your class:  You look at the interface and then write some code using share().  It works on UNIX and seems to compile on Windows, and you basic tests pass.  But then your users come back and start telling you some features don't work.  Why doesn't sharing work?  It would be better if the error was discovered at compile time, not at runtime.

There are other system related issues I can think of here but I'm no so familiar with the shared state functionality right now, so I won't say anything else.

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+	{
+	  #ifndef _WIN32
+	  float* shared_weights = (float*)mmap(0, (length << _stride_shift) * sizeof(float),
+			                  PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
+          size_t float_count = length << _stride_shift;
+      	  weight* dest = shared_weights;
+	  memcpy(dest, _begin, float_count*sizeof(float));
```
TomasPuverle: What happens here if the `mmap` failed?

vowpalwabbit/array_parameters.h
```diff
@@ -0,0 +1,172 @@
...
+	}
+	~weight_vector(){
+		if (_begin != nullptr && !_seeded)
+		{
+			free(_begin);
+			_begin = nullptr;
+		}
+	}
+	friend class weights_iterator;
+};
```
TomasPuverle: Another general observation.  There is no reason for this entire class to live in the header file.  I don't see any of the functions are being performance critical.  Not only are you coupling the implementation to the interface, forcing people to recompile basically the whole project if you change a small implementation detail, this will also lead to code bloat if things are inlined unnecessarily.

Even if the class was to be turned into a template (like I suggested), many of the functions could still be written to be type agnostic (e.g. share) and live in a .cc file.

cs/cli/vw_example.cpp
```diff
@@ -139,7 +139,7 @@ namespace VW
 
     System::String^ FormatFeature(vw* vw, feature_value& f1, feature_index& i1)
     {
-        uint64_t masked_weight_index1 = i1 & vw->reg.weight_mask;
+		uint64_t masked_weight_index1 = i1 & vw->wv.mask();
```
TomasPuverle: Since code like this is all over the place, have you considered adding an inline "index" function to your weigh_vector class?

cs/cli/vw_example.cpp
```diff
@@ -183,8 +183,8 @@ namespace VW
         vector<size_t> fa_missing;
         for (size_t ia = 0, ib = 0; ia < fa.values.size(); ia++)
         {
-            auto masked_weight_index = fa.indicies[ia] & vw->reg.weight_mask;
-            auto other_masked_weight_index = fb.indicies[ib] & vw->reg.weight_mask;
+            auto masked_weight_index = fa.indicies[ia] & vw->wv.mask();
```
TomasPuverle: Again, perhaps `wv.index(fa.indicies[ia])`?  This is the last time I will mention this.

vowpalwabbit/OjaNewton.cc
```diff
@@ -52,51 +52,53 @@ struct OjaNewton {
...
-        if(normalize) { // initialize normalization part
-	    for (uint32_t i = 0; i < length; i++)
-	        weights[(i << stride_shift) + NORM2] = 0.1;
-        }
-
+	void initialize_Z()
+	{  weight_vector& weights = all->wv;
+       uint32_t length = 1 << all->num_bits;
+	   if (normalize) { // initialize normalization part
+		 for (weight_vector::iterator i = weights.begin(NORM2); i != weights.end(NORM2); ++i)
```
TomasPuverle: Here's an example of the interface suggestion I had for using a range.  

    for (auto & ww: weights.mutate(NORM2))
    {
      ww = 0.1;
    }

Seems a bit nicer, no?

In any case, your should avoid writing loops like these.  The above is really just 

`std::fill(weights.begin(NORM2), weights.end(NORM2), 0.1)`.

Or, even better, if you decide to use boost.range library, along with the range-based interface I suggested:

`boost::fill(weights.mutate(NORM2), 0.1);

vowpalwabbit/array_parameters.h
TomasPuverle: I should also mention that it may be interesting for you to look at a matrix/vector library such as EIGEN.  This way, instead of returning ranges, you could try and return objects (vectors or matrices), which would then allow you to write code such as `wv.as_vector() = 0.1`, so you won't have to write any loops at all, as the classes handle striding automatically.

Second, such libraries also handle numeric operations, so a lot of the computations in loops could be just rewritten using basic arithmetic operations on vectors/matrices.  This would make the code more readable, and EIGEN uses some sophisticated C++ template methods to compile to very efficient code.  It even does automatic vectorization!

vowpalwabbit/OjaNewton.cc
TomasPuverle: Or with eigen, as I mentioned, `weights.as_vector(NORM2) = 0.1`.

vowpalwabbit/OjaNewton.cc
```diff
@@ -52,51 +52,53 @@ struct OjaNewton {
...
         if(!random_init) {
             // simple initialization
-            for (int i = 1; i <= m; i++)
-                weights[(i << stride_shift) + i] = 1;
+		  weight_vector::iterator iter = weights.begin(0) + 1;
+		  for (int i = 1; i <= m; ++i, ++iter)
+		  { weight_vector::iterator::w_iter j = iter.begin() + i;
+			*j = 1;
+		  }
         }
```
TomasPuverle: If you rewrote the code for your iterator as I suggested, again, you wouldn't need the loop with the (single letter name variable! :D) i.  This would simply end up being a std::fill or `boost::fill(*iter, 1)`.

vowpalwabbit/accumulate.cc
```diff
@@ -39,20 +42,22 @@ float accumulate_scalar(vw& all, float local_sum)
...
-  for (uint32_t i = 0; i < length; i++)
-    weights[stride*i+o] = local_grad[i]/numnodes;
-  delete[] local_grad;
+  weight_vector local_grad(length);
+
+  weight_vector::iterator i = weights.begin(o);
+  weight_vector::iterator j = local_grad.begin();
+  for (; i != weights.end(); ++i, ++j)
+	  *j = *i;
+
```
TomasPuverle: std::fill

vowpalwabbit/accumulate.cc
```diff
@@ -39,20 +42,22 @@ float accumulate_scalar(vw& all, float local_sum)
...
+  all_reduce<float, add_float>(all, local_grad.first(), length); //need to be modified to not use first()
+ 
+  i = weights.begin(o);
+  j = local_grad.begin();
+  for (; i != weights.end(); ++i, ++j)
+	  *i = *j/numnodes;
 }
```
TomasPuverle: This can be written as

`std::transform(local_grad.begin(), local_grad.end(), weights.begin(o), [numnodes](float weight) { return weight/numnodes; });`

vowpalwabbit/bfgs.cc
```diff
@@ -197,48 +193,49 @@ double regularizer_direction_magnitude(vw& all, bfgs& b, float regularizer)
...
-  uint32_t length = 1 << all.num_bits;
-  size_t stride_shift = all.reg.stride_shift;
-  weight* weights = all.reg.weight_vector;
+  weight_vector& weights = all.wv;
+  
   if (b.regularizers == nullptr)
-    for(uint32_t i = 0; i < length; i++)
-      ret += regularizer*weights[(i << stride_shift)+W_DIR]*weights[(i << stride_shift)+W_DIR];
+	  for (weight_vector::iterator i = weights.begin(W_DIR); i != weights.end(W_DIR); ++i)
+	  ret += regularizer* (*i) * (*i);
```
TomasPuverle: std::accumulate

vowpalwabbit/bfgs.cc
```diff
@@ -197,48 +193,49 @@ double regularizer_direction_magnitude(vw& all, bfgs& b, float regularizer)
...
+  weight_vector& weights = all.wv;
   double ret = 0.;
-  uint32_t length = 1 << all.num_bits;
-  size_t stride_shift = all.reg.stride_shift;
-  weight* weights = all.reg.weight_vector;
-  for(uint32_t i = 0; i < length; i++)
-    ret += weights[(i << stride_shift)+W_DIR]*weights[(i << stride_shift)+W_DIR];
-
+  for (weight_vector::iterator i = weights.begin(W_DIR); i != weights.end(W_DIR); ++i)
+	  ret += (*i) * (*i);
```
TomasPuverle: std::accumulate

vowpalwabbit/lda_core.cc
```diff
@@ -627,48 +626,47 @@ size_t next_pow2(size_t x)
...
-        { all->reg.weight_vector[j + k] = (float)(-log(frand48() + 1e-6) + 1.0f);
-          all->reg.weight_vector[j + k] *= (float)(l.lda_D / all->lda / all->length() * 200);
-        }
-      }
-      all->reg.weight_vector[j + all->lda] = all->initial_t;
-    }
+	weight_vector::iterator j = weights.begin(0);
+	weight_vector::iterator w_lda = weights.begin(all->lda);
+	for (; j != weights.end(); ++j, ++w_lda)
+	{ for (weight_vector::iterator::w_iter k = j.begin(); k != j.end(all->lda); ++k)
```
TomasPuverle: A couple of things:

1) Why are the begin iterators almost always defined outside of the for() loop preamble?  I can't see why you just don't do:
`for(iterator jj = weights.begin, kk = weights.begin(all->lda), end = weights.end(all->lda); ...) {}`

Also note that I moved the `end(all->lda)` into the initialization section.  The condition `k != ...` needs to be evaluated on each iteration and it's dubious whether or not the compiler will notice that it's not changing.  So unless the contents of your iterator range is changing during the iteration, it's better to evaluate it just once.  This happens in a lot of places.

TomasPuverle:
Ok, I will stop here for now.  Thanks for your work on this - I learnt a lot about VW by studying your changes.  Please let me know if my comments don't make sense; I wrote these late at night and might now have been as lucid in my explanations as I should be.

vowpalwabbit/array_parameters.h
arielf: Thank you both @aartibagul and @TomasPuverle 

Many (John once commented than not all could, due to some Visual Studio issues(?)) of the ugly `#ifndef _WIN32`) should be defined and isolated in the platform header `vowpalwabbit/crossplat_compat.h`

Also see this extended/suggested old attempt at [ http://pastebin.com/thHtb7vr ]( http://pastebin.com/thHtb7vr )

The code would look so much nicer without this `#ifdef`s all over.

With you two,  we might have the contributors who are up to this challenge now. I'm sorry I don't have the C++ and Windows fu to make this happen.   Grateful for any cleanup any of you could do on this front.

vowpalwabbit/array_parameters.h
JohnLangford: Minimizing headers does seem like a good idea.

vowpalwabbit/array_parameters.h
JohnLangford: I hadn't thought about this, but it makes sense due to uncontrolled namespace pollution.  Other parts of the VW codebase were committing this sin, so I fixed it.  It's a drag though needing to open std in many .cc files.

aartibagul:
Hi Tomas,

Thank you for your comments! They are very useful and I will try to incorporate them into my code over multiple commits.

JohnLangford:
Aarti, is the encapsulation complete?

aartibagul:
I think so.

cs/cli/vw_example.cpp
```diff
@@ -139,7 +139,7 @@ namespace VW
 
     System::String^ FormatFeature(vw* vw, feature_value& f1, feature_index& i1)
     {
-        uint64_t masked_weight_index1 = i1 & vw->reg.weight_mask;
+		uint64_t masked_weight_index1 = i1 & vw->wv->mask();
```
JohnLangford: Maybe we should have a different name than 'wv'?  First, it's easy to transpose with 'vw' and second I'm not sure that 'vector' is a good descriptor for the general concept you are encapsulating.  Maybe 'parameters'?

python/pylibvw.cc
```diff
@@ -330,7 +330,7 @@ void unsetup_example(vw_ptr vwP, example_ptr ae)
     }
   }
 
-  uint32_t multiplier = all.wpp << all.reg.stride_shift;
+  uint32_t multiplier = all.wpp << all.stride_shift;
```
JohnLangford: Why define stride_shift independent of the parameters?

vowpalwabbit/OjaNewton.cc
```diff
@@ -52,51 +52,53 @@ struct OjaNewton {
...
-	    for (uint32_t i = 0; i < length; i++)
-	        weights[(i << stride_shift) + NORM2] = 0.1;
-        }
-
+	void initialize_Z()
+	{  weight_vector& weights = *(all->wv);
+       uint32_t length = 1 << all->num_bits;
+	   if (normalize) { // initialize normalization part
+		 for (weight_vector::iterator i = weights.begin(NORM2); i != weights.end(NORM2); ++i)
+		   *i = 0.1;
```
JohnLangford: I'm afraid of this operation because it will blow up with the number of parameters as I understand it. Maybe it's better to have a set_default() interface which for dense parameters does this but for the hashmap operates in a lazy fashion?

vowpalwabbit/OjaNewton.cc
```diff
@@ -52,51 +52,53 @@ struct OjaNewton {
...
+	    }
         if(!random_init) {
             // simple initialization
-            for (int i = 1; i <= m; i++)
-                weights[(i << stride_shift) + i] = 1;
+		  weight_vector::iterator iter = weights.begin(0) + 1;
+		  for (int i = 1; i <= m; ++i, ++iter)
+		  { weight_vector::iterator::w_iter j = iter.begin() + i;
+			*j = 1;
```
JohnLangford: Initialization issue again.

cs/cli/vw_cbutil.cpp
```diff
@@ -13,6 +13,6 @@ namespace VW
   {
     CB::cb_class observation = { cost, actionObservered, probability };
 
-    return get_unbiased_cost(&observation, actionTaken);
+    return CB_ALGS::get_unbiased_cost(&observation, actionTaken);
```
JohnLangford: Fixed in the master---sorry about that.

vowpalwabbit/OjaNewton.cc
```diff
@@ -52,51 +52,53 @@ struct OjaNewton {
...
-	           double r2 = frand48();
-	           weights[(i << stride_shift) + j] = sqrt(-2 * log(r1)) * cos(2 * PI * r2);
-	        }
-	    }
-
+		for (weight_vector::iterator i = weights.begin(0); i != weights.end(); ++i)
+		  for (weight_vector::iterator::w_iter j = i.begin() + 1; j != i.end(m + 1); ++j)
+		  {  double r1 = frand48();
+		     double r2 = frand48();
+			 *j = sqrt(-2 * log(r1)) * cos(2 * PI * r2);
```
JohnLangford: This is box-meuller (i.e. gaussian random variable) initialization.  Making this work in a reproducible sparse fashion is a little bit tricky but possible.  You can seed the PRG for an entry with the entry's index.

vowpalwabbit/OjaNewton.cc
```diff
@@ -268,27 +270,30 @@ struct OjaNewton {
...
-        for (uint32_t i = 0; i < length; i++)
-            for (int j = 1; j <= m; j++)
-            weights[(i << stride_shift)] += weights[(i << stride_shift) + j] * b[j] * D[j];
-
+        //size_t stride_shift = all->stride_shift;
+        weight_vector& weights = *(all->wv);
+		for (weight_vector::iterator i = weights.begin(0); i != weights.end(); ++i)
+		{  weight_vector::iterator::w_iter w_j = i.begin() + 1;
+		   for (int j = 1; j <= m; ++j, ++w_j)
+		     *i += *w_j * b[j] * D[j];
```
JohnLangford: I like the way stride_shift stuff goes away.   Much cleaner.

JohnLangford:
Reading through, I think my primary concern is the parameter initialization routines.  It seems like these need to be in the parameter class or else when you plug in -b 64 various terrible things will fall apart.  Doable?

python/pylibvw.cc
aartibagul: When I had it as a variable in weight_parameters, I initialized the weights in parse_regressor as 
```all.weights = *(new weight_parameters(length, all.weights.stride_shift())); ```
which is very hacky.  This is because the stride_shift is set in parse_modules() whereas weights is initialized in parse_sources() (line 1360,1361 in parse_args.cc). So I could either propagate the stride_shift back up and pass it to parse_sources or I could just put it in the vw struct. I think we decided on doing the latter (for now) on our last Skype call.

**Comment by aartibagul** [comment]
I think this is done.

File: vowpalwabbit/stagewise_poly.cc [hunk 63]
```diff
@@ -269,8 +269,8 @@ void sort_data_update_support(stagewise_poly &poly)
   for (uint64_t i = 0; i != poly.all->length(); ++i)
   { uint64_t wid = stride_shift(poly, i);
     if (!parent_get(poly, wid) && wid != constant_feat_masked(poly))
-    { float wval = (fabsf(poly.all->reg.weight_vector[wid])
-                    * poly.all->reg.weight_vector[poly.all->normalized_idx + (wid)])
+    { float weightsal = (fabsf(poly.all->weights->operator[](wid))
```
JohnLangford: I'm writing some notes as I work through the code for reference (don't act on them, I'm acting).  
The use of operator[] is bleck.  The basic problem is that we have weight_parameters* weights instead of weight_parameters weights.  I don't see a good reason for a pointer dereference, so I'm going to remove it.

vowpalwabbit/gd.h
```diff
@@ -23,40 +24,44 @@ void print_audit_features(vw&, example& ec);
...
+		  p->scalar += fx * (*iter);
   }
   else    // TODO: this could be faster by unrolling into two loops
-    for (size_t c=0; c<mp.count; ++c, fi += (uint64_t)mp.step, ++p)
-    { fi &= mask;
-      p->scalar += fx * w[fi];
+  { iter += fi;
+    for (size_t c=0; c<mp.count; ++c, iter += (uint64_t)mp.step, ++p)
+    { 
+      p->scalar += fx * (*iter);
```
JohnLangford: There is a bug at this point because iter can run over the end of the array.  I've fixed.

vowpalwabbit/parse_regressor.cc
```diff
@@ -491,8 +504,9 @@ void parse_mask_regressor_args(vw& all)
       io_temp.close_file();
 
       // Re-zero the weights, in case weights of initial regressor use different indices
-      for (size_t j = 0; j < length; j++)
-      { all.reg.weight_vector[j << all.reg.stride_shift] = 0.;
+	  weight_parameters& w = all.weights;
+      for (weight_parameters::iterator j = w.begin(0); j != w.begin(); ++j)
```
JohnLangford: Is the w.begin() here a bug?

vowpalwabbit/lda_core.cc
```diff
@@ -30,7 +30,7 @@ license as described in the file LICENSE.
 #include "gd.h"
```
JohnLangford: @eisber can you verify no negative impact on computation in lda?

vowpalwabbit/kernel_svm.cc
```diff
@@ -872,8 +872,8 @@ LEARNER::base_learner* kernel_svm_setup(vw &all)
   else
     params.kernel_type = SVM_KER_LIN;
 
-  params.all->reg.weight_mask = (uint64_t)LONG_MAX;
-  params.all->reg.stride_shift = 0;
+  //params.all->weights->mask((uint64_t)LONG_MAX);
```
JohnLangford: Are you sure this is ok?

vowpalwabbit/global_data.h
```diff
@@ -537,7 +532,9 @@ struct vw
   time_t init_time;
 
   std::string final_regressor_name;
-  regressor reg;
+
+  uint32_t stride_shift;
```
JohnLangford: This loss of modularity of stride_shift is annoying.  I'd like to hide this entirely inside of the parameters, but that requires more work than seems reasonable at this point.

vowpalwabbit/gd.cc
```diff
@@ -731,82 +735,95 @@ void save_load_online_state(vw& all, io_buf& model_file, bool read, bool text, g
   }
 
   uint64_t length = (uint64_t)1 << all.num_bits;
-  uint64_t stride = (uint64_t)1 << all.reg.stride_shift;
 
-  int c = 0;
-  uint64_t i = 0;
+  weight_parameters& weights = all.weights;
+  uint32_t i = 0;
```
JohnLangford: Why the change to uint32_t?

vowpalwabbit/gd.cc
```diff
@@ -731,82 +735,95 @@ void save_load_online_state(vw& all, io_buf& model_file, bool read, bool text, g
...
   size_t brw = 1;
   do
   { brw = 1;
-    weight* v;
+    weight_parameters::iterator v = weights.begin(0);
+    weight_parameters::iterator v1 = weights.begin(1);
```
JohnLangford: This seems unfortunate.  You can replace *v1 with *(&(*v)+1).  It looks like line noise... but probably better than keeping extra iterators around.

JohnLangford: (Fyi, done.)

vowpalwabbit/example.cc
```diff
@@ -128,8 +128,11 @@ flat_example* flatten_example(vw& all, example *ec)
   fec.num_features = ec->num_features;
 
   full_features_and_source ffs;
-  ffs.stride_shift = all.reg.stride_shift;
-  ffs.mask = (uint64_t)all.reg.weight_mask >> all.reg.stride_shift;
+  ffs.stride_shift = all.stride_shift;
+  if (all.weights.not_null())  //TODO:temporary fix. all.weights is not initialized at this point in some cases.
+    ffs.mask = (uint64_t)all.weights.mask() >> all.stride_shift;
+  else
+    ffs.mask = (uint64_t)LONG_MAX >> all.stride_shift;
```
JohnLangford: This is for the kernel code?

vowpalwabbit/gd.cc
JohnLangford: Or better: (&(*v))[1]  
(it looks like formatting was messed up in the first suggestion)

vowpalwabbit/bfgs.cc
```diff
@@ -420,102 +426,101 @@ double add_regularization(vw& all, bfgs& b, float regularization)
...
 
 void update_weight(vw& all, float step_size)
-{ uint32_t length = 1 << all.num_bits;
-  size_t stride = 1 << all.reg.stride_shift;
-  weight* w = all.reg.weight_vector;
-
-  for(uint32_t i = 0; i < length; i++, w+=stride)
-    w[W_XT] += step_size * w[W_DIR];
+{ weight_parameters& w = all.weights;
+  weight_parameters::iterator w_xt = w.begin(W_XT);
```
JohnLangford: I've eliminated all the double iterator stuff in bfgs.cc

vowpalwabbit/accumulate.cc
```diff
@@ -69,38 +74,42 @@ float min_elem(float* arr, int length)
...
   uint32_t length = 1 << all.num_bits; //This is the number of parameters
-  size_t stride = 1 << all.reg.stride_shift;
-  weight* weights = reg.weight_vector;
-  float* local_weights = new float[length];
+  weight_parameters local_weights(length);
 
-  for(uint32_t i = 0; i < length; i++)
-    local_weights[i] = weights[stride*i+1];
+  weight_parameters::iterator i = weights.begin(1);
+  weight_parameters::iterator j = local_weights.begin();
```
JohnLangford: It's not clear to me that local_weights should be a weight_parameters because it has a rather different meaning.  This is temporary storage just for the purpose of communication.

test/test-sets/ref/ml100k_small.stderr
```diff
@@ -9,17 +9,17 @@ Reading datafile = test-sets/ml100k_small_test
...
+1.469532 0.557971            8            8.0   4.0000   3.2111       23
+1.509373 1.549214           16           16.0   5.0000   3.2134       23
 
 finished run
 number of examples per pass = 30
 passes used = 1
 weighted example sum = 30.000000
 weighted label sum = 110.000000
-average loss = 1.471805
+average loss = 1.457823
```
JohnLangford: It's rather odd that the new way of doing random initialization seems to have helped in every test case.

vowpalwabbit/parse_regressor.cc
aartibagul: Yes. Sorry about that. I'll fix it.

vowpalwabbit/global_data.h
aartibagul: It won't take too long but we'll end up with something like - 
```new(&all.weights) weight_parameters(length, all.weights.stride_shift()); ``` in parse_regressor to get the stride_shift. Is that okay?

vowpalwabbit/gd.cc
aartibagul: That was not intentional. Will change it.

vowpalwabbit/global_data.h
JohnLangford: That seems fine.  Much better modularity-wise.

JohnLangford:
Overall, I think it's looking ok but I'd like to make a couple more changes before merging in (or perhaps you can make them)?

1) I don't see a reason why the iterators need iterators.  The base type is a float, so there is no need to abstract iteration over floats.  Instead, you can just use float* base = (&(*iter)); and then a for loop.  This seems simpler, because it will cut work out of both abstractions (sparse and dense) while causing no extra work where it is used.
2) I don't see a fundamental reason why we need begin() which iterates over everything yet.
3) I don't see why we need a begin(...) which takes an argument.  Why can't we just use a begin() that uses stride to shift?
4) It would be nice to have stride_shift in one place instead of two.  Two places causes bugs...

JohnLangford:
@eisber do you know what's up with the appveyor here?  (Is something broken in the C# layer?)  Valgrind reports no errors.

JohnLangford:
I thought about this a little bit more.  I think leaving (1) in is ok.  The advantage of an iterator over an array deref is that you are sure to not mess up the end.

eisber:
Having begin(), end() on containers enables C++ 11 iterators. I don’t like the begin(…) because it breaks C++ 11 iterators. The pattern to follow container.produceSubContainerTemporary(…).begin()

vowpalwabbit/kernel_svm.cc
aartibagul: The mask is calculated using the length and the stride_shift so I don't think there should be a setter for mask.

vowpalwabbit/lda_core.cc
eisber: Just figured, LDA got broken. Fixed in next PR.

PR #1010, opened 2016-06-08
fixed indentation for cb explore adf

lhoang29:


lhoang29:
going to merge since this is just indentation fix.

PR #1011, opened 2016-06-10
Cbify

alekh:


JohnLangford:
@eisber why is the test failing here?

eisber:
I’m checking if I can extract the arguments correctly again.
I merged Alekhs patches into my branch and updated a couple other things along. Can we discard this PR and I’ll create one?

JohnLangford:
Merging from your branch is scary.   We can certainly do this, but only if
it does not create a large delay.

-John

eisber:
Let me know how scary this one is: https://github.com/JohnLangford/vowpal_wabbit/pull/1013

Markus

JohnLangford:
You didn't break github?  Clearly you are getting soft.

JohnLangford:
I've merged this now because it has an additional bug fix.  I expect that git will do the right thing because your patch incorporates this one.

PR #1015, opened 2016-06-18
I few warnings for cache file processing problems

trufanov-nok:
Just faced with a problem when VW ignores specified cache file and tries to make another one instead. It took some time to realize that this happens bcs I forgot to specify the right `-b` value during cache creation. Frankly, I didn't realized this till I debug. I think would be nice to add a few warnings for such cases if anyone else forgets how VW processes cache files.


JohnLangford:
The change to learner.cc seems to have created a heisenbug.
https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/138553277#L1982

JohnLangford:
Merged, thanks.

I've seen the heisenbug on travis twice now and (anecdotally) a few times on my laptop.

trufanov-nok:
Yeah, a bit weird.

PR #1022, opened 2016-06-30
Adding default limits to prediction.

sharatsc:
The max and min are updated using labels and cause problems with non identity link functions


JohnLangford:
Merged, thanks.

PR #1025, opened 2016-07-06
Getting Java compilation to work

jmorra:
While trying to make a release version of VW 8.2 I noticed that the Java build script was broken.  This commit fixes that issue.

### DIFF
--- a/java/src/main/bin/build.sh
+++ b/java/src/main/bin/build.sh
@@ -28,7 +28,7 @@ export JAVA_HOME=/usr/lib/jvm/java-7-openjdk-amd64;
 $make_base
 mv java/target/vw_jni.lib java/target/vw_jni.Ubuntu.14.amd64.lib"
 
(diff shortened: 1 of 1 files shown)


jmorra:
@deaktator could you have a look at this?

Thanks

deaktator:
This looks good to me.  It seems like the `\s*` in the *sed* expressions in `build.sh` is much more robust than the spaces previously used.  I'm not sure what the semicolon does, but that's OK.  It looks like the point of the sed expression is to include the proper include path in case it's not `/usr/local/lib`.  Is this correct?  Maybe put a comment above line 46.

jmorra:
@deaktator, you are correct the BOOST_LIBRARY needed to be more general to continue to work.  The other changes were simply stylistic.  @JohnLangford, do you think this is good to merge in?

JohnLangford:
Yep.   Thanks.

PR #1027, opened 2016-07-07
Azure Cloud Service Tracing Improvements

eisber:


JohnLangford:
@lhoang29 can you take a look at this?

lhoang29:
looks good

JohnLangford:
Travis is failing on this one.

eisber:
hmm some of my unfinished LDA improvements made into this. can you hold of a bit?

JohnLangford:
Status on this?

lhoang29:
Do I need to review more?

eisber:
fixed Dictionary<string, float> support
fixed Test 77

JohnLangford:
@lhoang29 Can you take a look again?

cs/azure/Trainer/Operations/InitializeOperation.cs
```diff
@@ -75,7 +75,11 @@ internal void FreshStart(OnlineTrainerState state = null)
             // start from scratch
             this.state = state;
 
-            this.InitializeVowpalWabbit(new VowpalWabbitSettings(this.settings.Metadata.TrainArguments));
+            var settings = model == null ?
+                new VowpalWabbitSettings(this.settings.Metadata.TrainArguments) :
+                new VowpalWabbitSettings(string.Empty) { ModelStream = new MemoryStream(model) };
```
lhoang29: This is probably wrapped in a try-catch under the PUT method right? Also, we'll likely see many user errors w.r.t. using warm start model so perhaps in the logged message, we could have a link to some wiki page for diagnosing it? By wiki page I'm thinking of general guidelines on what kind of models would be compatible etc...

cs/azure/Worker/OnlineTrainerSettingsWatcher.cs
```diff
@@ -152,7 +152,7 @@ private ICheckpointPolicy ParseCheckpointPolicy()
                 if (checkpointString.Contains(":"))
                 {
                     TimeSpan interval;
-                    if (!TimeSpan.TryParse(checkpointString, CultureInfo.InvariantCulture, out interval))
+                    if (TimeSpan.TryParse(checkpointString, CultureInfo.InvariantCulture, out interval))
```
lhoang29: this seems useful :)

cs/azure/Worker/ResetController.cs
```diff
@@ -57,5 +66,29 @@ public ResetController(LearnEventProcessorHost trainProcessorFactory)
...
+
+            try
+            {
+                var model = await Request.Content.ReadAsByteArrayAsync();
+
+                await this.trainProcessorHost.ResetModelAsync(model: model);
+
+                return this.Request.CreateResponse(HttpStatusCode.OK);
+            }
+            catch (Exception ex)
```
lhoang29: ah this answers part of my other question

explore/cpp/utility.h
```diff
@@ -232,7 +232,7 @@ class NumberUtils
 class StringUtils
 {
 public:
-  static inline string to_string(size_t& n)
+  static inline string to_string(size_t n)
```
lhoang29: why is this needed?

eisber: In cbify.cc I’m calling to_string(<temp result>) and one can’t take &<temp result>

Also I don’t see any benefit of using the reference here…

vowpalwabbit/lda_core.cc
```diff
@@ -809,12 +858,227 @@ void learn(lda &l, LEARNER::base_learner &, example &ec)
...
+struct feature_pair
+{
+	uint64_t f1;
+	uint64_t f2;
+
+	feature_pair(uint64_t _f1, uint64_t _f2) : f1(_f1), f2(_f2)
+	{}
+};
+
+void compute_metrics(lda &l)
```
lhoang29: Perhaps add comments / clean up this part to make it more readable. Additionally, should this be named something like compute_coherence?

explore/cpp/utility.h
lhoang29: Doesn't look like this is on the hot path at the moment so not a real issue. Just seemed odd to pass by value but I see this avoids an awkward new variable.

vowpalwabbit/cbify.cc
```diff
@@ -40,6 +40,8 @@ struct cbify
   MwtExplorer<example>* mwt_explorer;
   vw_recorder* recorder;
   v_array<action_score> a_s;
+  // used as the seed
+  size_t example_counter;
```
lhoang29: Is this being initialized somewhere? I see it's just being used as a seed so bogus starting values are still ok, although someone else could use it for its name.

lhoang29:
done with another pass

vowpalwabbit/cbify.cc
eisber: I joined the dark side by relying on calloc. John laughed pretty hard on me when he saw me doing it…

cs/azure/Trainer/Operations/InitializeOperation.cs
eisber: done

JohnLangford:
Status on this?

eisber:
Ready to pull…

JohnLangford:
Merged, thanks.

PR #1032, opened 2016-07-10
vw-hyperopt extract label from input

gwulfs:
adds support for optional arguments in order to prevent the following error:

`ValueError: invalid literal for float(): -5.123953 6171_1851165462_2016-05-23_02:00:00[0.0]`


JohnLangford:
@gramhagen @hal3 can you take a look at this?

gramhagen:
looks good to me, I'm assuming the extra stuff is the example's tag that's coming in with the prediction?

JohnLangford:
Merged, thanks.

Issue #1028, opened 2016-07-08
[Feature Request] Rectified linear unit option for the neural net


zachmayer:
The neural net model in VW currently uses sigmoid units.  I was thinking it might be useful to have some other activation functions, most notably rectified linear units.

Thanks!

ekg:
I am also interested in this. Would it be a straightforward change?

zachmayer:
I think the activation happens here: https://github.com/JohnLangford/vowpal_wabbit/blob/4faae05eddce3bcde15e0626d52b04b19c8bc24c/vowpalwabbit/nn.cc#L212

It looks like you could replace `fasttanh` with a function for calculating the rectified linear activation, but I'm not sure.

JohnLangford:
Let me 'cc Paul Mineiro.  You need to get the gradient right also.

-John

JohnLangford:
Yes I think letting the user select the activation function would be easy, you just need two functions which do the forward and backward, and then make (fast)tanh the default.


If you send John a pull request cc-me on it and I'll take a look.


p


Sent from Outlook<http://aka.ms/weboutlook>

________________________________

joshloyal:
I'd be happy to submit a pull request for this. Since we're only changing the activation, it should be as simple as abstracting away the activation to a general nonlinearity that has a basic call and a `first_derivative` and plugging them in were necessary as Paul said.

JohnLangford:
Go ahead.

-John

JohnLangford:
Closing, as we can discuss in the pull request.

PR #1033, opened 2016-07-11
Rectified linear unit option for neural net

joshloyal:
This pull request addresses issue #1028 (adding different activation functions to the neural net). 

Specifically it adds a `nonlinearity` member to the `nn` struct, which defines a `fire` function for the forward pass and a `first_derivative` for the backward pass. Adding additional activation functions can now be down by creating additional sub-classes that override these methods. My c++ is a bit rusty and I'm not too familiar with the vowpal wabbit code base, so criticism is definitely welcome.

However as it is written now, the relu neural net is not learning.. I believe it is because the hidden weights are currently initialized to zero, which for the relu is a point of zero gradient so there is no chance of learning to occur. Looking through the code I couldn't find the best place to set a different initialization for the weights (only `calloc`), so if you could point me to where I could make that change that would be great. Of course, if you see another reason why the relu layer is not learning, let me know and I will make the change.

Once I fix the learning issue I will go ahead and add a test. Thanks!

cc: @pmineiro @JohnLangford

### DIFF
--- a/test/pred-sets/ref/0022.stderr
+++ b/test/pred-sets/ref/0022.stderr
@@ -1,5 +1,6 @@
 only testing
 predictions = 0022.predict
+Use activation 'tanh' for neural network testing
 Num weight bits = 16
 learning rate = 0.5
 initial_t = 0
--- a/test/train-sets/ref/nn-1-noquiet.stderr
+++ b/test/train-sets/ref/nn-1-noquiet.stderr
@@ -1,3 +1,4 @@
+Use activation 'tanh' for neural network training
 Num weight bits = 18
 learning rate = 0.5
 initial_t = 0
--- a/test/train-sets/ref/search_dep_parser_one_learner.stderr
+++ b/test/train-sets/ref/search_dep_parser_one_learner.stderr
@@ -1,3 +1,4 @@
+Use activation 'tanh' for neural network training
 Enabling FTRL based optimization
 Algorithm used: Proximal-FTRL
 ftrl_alpha = 0.005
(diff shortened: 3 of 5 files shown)


pmineiro:
John is a fan of initializing from zero :) so can you try putting the hinge in the relu somewhere else, e.g., -0.5f?

This is like initializing the bias to be positive.

pmineiro:
(Bonus points: there's this thing Patrice really likes, "paired RELUs", where (in pairs) they have the same input->hidden weights but one computes max(x,0) and one computes min(x,0), so that somebody is always on; and the hidden->output weights are distinct.  In that case, there's always some path for the gradient.)

JohnLangford:
A couple things here:

(1) I don't see a good reason to stick the definition of activation_function into the header.  My understanding is that this is used nowhere else in the codebase.

(2) I generally prefer using template to virtual as an interfacing mechanism, because you will pay an indirection every time you use it.  This usually will  not matter here because activation functions are amortized by the dot-product, but if there are not many features it will matter.  If you look here https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/scorer.cc#L59 you'll see several examples.  

(3) W.r.t. weight initialization, the default in VW is zeros as Paul says.  I'm a little bit hazy though on why this was working with sigmoid because the need to do something nonrandom exists there.  It looks like random weights are being used here: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/nn.cc#L226 to randomize the label?

pmineiro:
Re: 3), iirc, this was to break symmetry between the hidden units.  If
there was only one hidden unit, my recollection is, it wasn't necessary.
Whereas, the problem noted with relus being stuck at zero would effect a
single hidden unit architecture.

So I'm thinking if we just shift the hinge of the relu to minus 1 instead
of zero, then relu and sigmoid should work (or fail) similarly.

p

JohnLangford:
I'm closing this for now---please reopen if interested.

PR #1041, opened 2016-07-16
Fix crash when running with --audit.

TomasPuverle:
When running with --audit, if an anonymous feature is present, VW will segv.

### DIFF
--- a/vowpalwabbit/gd.cc
+++ b/vowpalwabbit/gd.cc
@@ -183,8 +183,14 @@ struct audit_results
 
 
 inline void audit_interaction(audit_results& dat, const audit_strings* f)
-{ if (f == nullptr)
-  { dat.ns_pre.pop_back();
+{ 
+  if (f == nullptr)
+  {
+    if (!dat.ns_pre.empty())
+    { 
+        dat.ns_pre.pop_back();
+    }
+
     return;
   }
 
--- a/vowpalwabbit/gd.cc
+++ b/vowpalwabbit/gd.cc
@@ -199,15 +199,20 @@ inline void audit_interaction(audit_results& dat, const audit_strings* f)
     ns_pre += '*';
 
   if (f->first != "" && ((f->first) != " "))
-    {
-      ns_pre.append(f->first);
-      ns_pre += '^';
-    }
+  {
+    ns_pre.append(f->first);
+    ns_pre += '^';
+  }
+  
   if (f->second != "")
-    {
-      ns_pre.append(f->second);
-      dat.ns_pre.push_back(ns_pre);
-    }
+  {
+    ns_pre.append(f->second);
+  }
+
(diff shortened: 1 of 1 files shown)


TomasPuverle:
This is easily reproducible in current master by the following example:

1 | one
2 | two
3 |ns three
4 | :1.3

vw -d <file above> --audit.  The fourth example will crash.

The main question I have is regarding the audit output. The fix causes the output to look like this:

vowpalwabbit/vw -d multiclass --audit --noconstant
<snip>
        :115744:1.3:0@0
7.500000 12.500000            4            4.0   4.0000   0.0000        1
<snip>

Note that there is no namespace (or otherwise) information preceding the : (colon), and I think that may be desirable.  How should this be handled?  Should the output look like e.g.

ns^:115744:1.3:0@0

or even

ns^<anon>:115744:1.3:0@0?

arielf:
This looks like a recent regression.  Older versions of vw work fine, for example:

vw-8.20151204 (Dec 2015):
```
$ echo "4 | :1.3" | vw -a
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile =
num sources = 1
0
        :0:1.3:0@0      Constant:116060:1:0@0
16.000000 16.000000            1            1.0   4.0000   0.0000        2

finished run
number of examples per pass = 1
passes used = 1
weighted label sum = 4.000000
average loss = 16.000000
best constant = 4.000000
best constant's loss = 0.000000
total feature number = 2
```

While latest does:
```
$ echo "4 | :1.3" | vowpalwabbit/vw -a
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile =
num sources = 1
0
Segmentation fault (core dumped)
```

Could you run a `git bisect run ...` and find the change that broke it?  Thanks!

TomasPuverle:
Hi,

Thanks for pointing that out.

The problem was introduced in the refactor here:

commit badeedb2c11dfb56f5c2314088de2d8e3eb88e4f
Date:   Mon Jan 4 20:07:06 2016 -0500
fix compile warnings in VS

Ironically (besides having a hash of "bad" :D), it also broke the linking on Linux, so I can't actually build that specific commit (the linking issue was resolved a couple of commits later in eb84b13ca73005c02fe5bfa9df60b5bcf53d79b1)

In any case, looking at the code, it's clear what's going on - before, the code distinguished between the absence of a name and and empty name, which is no longer the case.  It also answers my question from above what the output should look like.  Unless there are any objections, I will update my code to replicate the old output which includes the namespace names.

arielf:
@TomasPuverle  Thanks so much for the bisect and for figuring the causes so quickly.

TomasPuverle:
Thanks for pointing out that it used to work, and linking the related issue!

PR #1052, opened 2016-07-24
Python updates

gramhagen:
I merged in the changes from #994 (thanks @irustandi!) and updated the packaging so we now supports python3 (should pick the correct boost lib automatically based on version being used during installation)

Also, I added a couple of functions (save and load) to support pickling the models built using the sklearn interface and fixed a bug when scoring VWRegressor type models


JohnLangford:
@hal3 can you take a look?

It looks like this drops support for earlier python?  Or am I misreading?

gramhagen:
I should have been more explicit, this extends support for both python2 and
3, and resolves the appropriate boost python version based on the default
version of python available when the python package installation occurs

JohnLangford:
Sounds great.  I'm going to merge.

JohnLangford:
(Thanks!)

PR #1053, opened 2016-07-25
Test cases for --audit output.

TomasPuverle:
As requested, here's a test case for the --audit regression.

Also, while looking at this, I realized that the --audit output actually goes to stdout, not stderr.  Not sure if that's intended or not.

One of the issues I see is that the test case will fail if the hash function used by VW is ever changed.  I don't see a good way to template it in the current test framework.

### DIFF
--- a/test/RunTests
+++ b/test/RunTests
@@ -1485,3 +1485,9 @@ printf '3 |f a b c |e x y z\n2 |f a y c |e x\n' | \
 {VW} -d train-sets/cb_adf_crash_2.data -i models/cb_adf_crash.model -t
     train-sets/ref/cb_adf_crash2.stderr
 
+# Test 137: Fix for regression introduced by badeedb.  
+# Ensure audit output continues to work correctly in the presence of anon features.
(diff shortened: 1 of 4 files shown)


TomasPuverle:
Oh, I did that too.  I actually generated the output from badeedb^1 and then added the tests to master.  Sorry, perhaps I should have been clearer and could have saved you some work.

In this case, and for similar small issues, it would be nice to have an actual unit-testing framework, as opposed to the full integration test suite.

TomasPuverle:
Oh - one more thing.  Is the output to stdout intended?

arielf:
Yes, if we ever change the hash function, the reference output would have to change.  No a biggie considering that the current hash function (murmurhash32 v3.0) has been pretty good and stable for the past couple of years.

`--audit` goes to stdout:  yes, it has been like this since forever.  The progress output and informational messages go to stderr, and everything else goes to stdout.  Not sure if this is strictly "right" but I would assume changing it would break some existing scripts which rely on this de-facto "API", so I would leave this as it is.

TomasPuverle:
That's what I thought, too.  Thanks.

Issue #758, opened 2015-08-21
Python module: get_tag() returns empty string


KristianHolsheimer:
Hi there,

I just noticed an issue with the `pyvw.vw.example.get_tag()` method in `python/pyvw.py`. Namely, `get_tag()` fails to parse the example tag and returns an empty string.

The minimal way to reproduce this is to run (in source root):

    $ python python/test.py | grep 'tag ='
    tag = 
    tag = 

which should in fact return

    tag = foo
    tag = foo

I tried to figure out where things go wrong, but without success..

Would this be general issue with example tags or could it be a `pylibvw` issue?

JohnLangford:
I'm assuming pylibvw.

JohnLangford:
Hal, can you look into this one?

-John

JohnLangford:
@hal3 Can you look into this?

JohnLangford:
@gramhagen are tags working in python?

gramhagen:
I'll take a look

gramhagen:
looks like this is still a problem, I'll dig in further

JohnLangford:
Closing.  Thanks @gramhagen !

KristianHolsheimer:
Thanks @gramhagen and @JohnLangford 
I added a test for tags in pyvw in `python/tests/test_pyvw.py`, see https://github.com/JohnLangford/vowpal_wabbit/pull/1076

gramhagen:
looks good! thanks @KristianHolsheimer

PR #1055, opened 2016-07-26
retaining example tag in python lib

gramhagen:
this is to fix issue #758 

there was some hacky stuff in pylibvw which erased the tag and used it to store the label type because that was needed to deallocate memory when the example was destroyed.

I couldn't find an easy way to retain the size_t labelType variable on it's own, and since example_counter was also being overridden to store this info I kept it there and used example_counter in the dealloc_example function.

this is still a bit hacky, but I didn't want to modify the base example class for this case, maybe someone has a better implementation for dealloc_example that can resolve this issue, but for now this works to get the tag.  The example counter information is still not valid, but that seems less relevant for the python library.

### DIFF
--- a/python/examples/test.py
+++ b/python/examples/test.py
@@ -80,7 +80,6 @@ for i in range(2):
     ex = vw.example("1 foo| a b")
     ex.learn()
     print('tag =', ex.get_tag())
-    print('counter =', ex.get_example_counter())
     print('partial pred =', ex.get_partial_prediction())
     print('loss =', ex.get_loss())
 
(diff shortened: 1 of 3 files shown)


JohnLangford:
Thanks.  

This seems unpleasant.  I think we need to have explicit typing for labels and predictions.

pommedeterresautee:
Is it supposed to improve Python package performance?

gramhagen:
not really, it's just some metadata that might be helpful if people want to extract information about an example, but I agree we should just use explicit typing to make example creation and destruction more straightforward to implement.

PR #1056, opened 2016-07-28
Make namespace ignored when used singularly, but not when used as int…

andraztori:
When using namespaces in interactions, even when explicitely ignoring namespace in command line, it gets pulled back in as part of interaction.
This changes the behavior, so "ignore" really means ignoring it and using the namespace in an interaction does not by default also uses it in singular form.

### DIFF
--- a/vowpalwabbit/feature_group.h
+++ b/vowpalwabbit/feature_group.h
@@ -213,7 +213,7 @@ struct features {
   typedef features_value_index_iterator iterator;
   typedef features_value_iterator iterator_value;
   typedef features_value_index_audit_iterator iterator_all;
-
+  char ignored;
   /// defines a "range" usable by C++ 11 for loops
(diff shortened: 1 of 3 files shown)


JohnLangford:
This patch is scary at present because there are templates exposed in gd.h which loop over the features in a namespace as well as the features in an example.  These seem to be used in mwt.cc and search.cc for namespace manipulation purposes.  

Instead of conditionally invoking foreach_feature(weight*...), maybe it's better to always invoke foreach_feature(weight*...) and have that function check to see if the namespace should be ignored?  This would imply adding a bool to foreach_feature(weight*...) which controls whether ignore is overridden (as in INTERACTIONS) or not.

JohnLangford:
I'm closing this for now.  Please reopen if you are interested in it.

PR #1060, opened 2016-08-04
Update instruction for Python 2

danqing:
At least on my machine (el capitan) `brew install boost --with-python` doesn't work. `pip install vowpalwabbit` after running that still complains about missing boost library. On the other hand, `brew install boost-python` worked.


danqing:
Hmm not sure why changing README breaks test...

JohnLangford:
The windows build bug is getting fixed (and unrelated).  Merging, thanks.

PR #1064, opened 2016-08-07
fix --ect in case --link=logistic is used

trufanov-nok:
resolves https://github.com/JohnLangford/vowpal_wabbit/issues/726
Just adjust boundary of prediction-to-class interpretation. It's 0.5 for --link logistic as predictions are scaled to [0;1] and 0. for all other link functions.


JohnLangford:
Merged, thanks.

PR #1066, opened 2016-08-09
Accept multiline examples in the JNI interface

jmorra:
This is a pretty major refactoring of the JNI layer.  The impetus for this refactoring was the ability to accept multiline examples but it lead to a much larger change.  The biggest change is the decoupling of the return type and the prediction function.  This proved necessary to support all the different ways to extract data with `--cb_explore`.

I am going to go over this in detail with @deaktator offline, so @JohnLangford let's hold off on merging this for now.  If anyone else has any comments at this time they are certainly welcome.


java/src/main/c++/vw_learner_VWLearners.cc
```diff
@@ -48,3 +61,74 @@ JNIEXPORT jobject JNICALL Java_vw_learner_VWLearners_getReturnType(JNIEnv *env,
   return env->GetStaticObjectField(clVWReturnType, field);
 }
 
+// Here are all the prediction functions we might want to use.  Add to this as more predictions are required.
+jint multiclass_predictor(example* vec, JNIEnv *env) { return vec->pred.multiclass; }
```
jmorra: @JohnLangford this needs VERY careful review.  Because I've separated the return type from the function that creates it, I now need to create all the prediction functions in one place.  This is that place.  Basically I need to be able to tie the VW arguments with the return function.  Can you please review this and the function `Java_vw_learner_VWLearners_getPredictionFunction` to make sure that all combination of input arguments are appropriately mapped to their output function.

java/src/main/java/vw/learner/VWBase.java
```diff
@@ -69,12 +71,15 @@ public boolean equals(Object o) {
 
         VWBase vwBase = (VWBase) o;
 
-        return nativePointer == vwBase.nativePointer;
+        if (nativePointer != vwBase.nativePointer) return false;
```
sahil-goyal: Why not `return nativePointer == vwBase.nativePointer && predictionFunctionPointer == vwBase.predictionFunctionPointer;`? Readability?

java/src/test/java/vw/learner/VWFloatArrayLearnerTest.java
```diff
@@ -109,4 +117,64 @@ private void writeVwModelToDisk() {
...
+            "3:1.5:0.7 | a d"
+        };
+
+        VWFloatArrayLearner vw = VWLearners.create("--quiet --cb_explore 4");
+        float[][] trainPreds = new float[cbTrain.length][];
+        for (int i=0; i<cbTrain.length; ++i) {
+            trainPreds[i] = vw.learn(cbTrain[i]);
+        }
+        float[][] expectedTrainPreds = new float[][]{
+            new float[]{0.962500f, 0.012500f, 0.012500f, 0.012500f},
```
sahil-goyal: How do you obtain the expected predictions?

java/src/main/java/vw/learner/VWBase.java
jmorra: Both `equals` and `hashCode` are methods generated by my Java editor, IntelliJ.  Because of this I'd rather not touch them.  They had to be regenerated because a new member variable was added, `predictionFunctionPointer`.

java/src/test/java/vw/learner/VWFloatArrayLearnerTest.java
jmorra: When performing either `--cb_explore` or `--cb_explore_adf` the user is requesting a score over actions.  On the Java side this is represented by a `float[]` where the index is the action and the value is the score.  The user would feed this array into another function that makes a decision, like the [Decision Service](https://mwtds.azurewebsites.net/).  This does bring up an interesting point though and that is that I assume that on the C side the number of actions is the same as the size of the `ACTION_SCORE::action_scores` `v_array` and that each index is filled in with a score.  @JohnLangford can you please comment as to whether or not this is correct.

java/src/main/c++/vw_learner_VWLearners.cc
JohnLangford: @eisber is looking into creating a tagged union for label and prediction types, effectively allowing dynamic typing. Plausibly, this will be very helpful here.  (More tomorrow, traveling today.)

jmorra: Thanks, I think this would be very helpful here.

JohnLangford:
Yes, that is correct.  

I've also eliminated 'probs' from the prediction types now.

.gitignore
```diff
@@ -31,6 +31,16 @@ test/*.cmp
 t_*
 bin
 vw
+!java/
```
jmorra: These are all necessary as the JNI classes start with VW and that is excluded by the above line.  If anyone knows of a prettier and more robust way to do this that would be great.

vowpalwabbit/lda_core.cc
```diff
@@ -1234,7 +1234,7 @@ LEARNER::base_learner *lda_setup(vw &all)
 
   ld.decay_levels.push_back(0.f);
 
-  LEARNER::learner<lda> &l = init_learner(&ld, ld.compute_coherence_metrics ? learn_with_metrics : learn, 1 << all.weights.stride_shift());
+  LEARNER::learner<lda> &l = init_learner(&ld, ld.compute_coherence_metrics ? learn_with_metrics : learn, 1 << all.weights.stride_shift(), prediction_type::scalars);
```
jmorra: The LDA module was returning the wrong prediction type because of this.  @JohnLangford do you think it's a good idea to force all callers of `init_learner` to supply the appropriate `pred_type`?  This would force this problem to come up at compile time instead of relying on the fact that the base learner's type is the same as `this`'s type (which is clearly not true for all cases).

vowpalwabbit/rand48.cc
```diff
@@ -20,6 +20,12 @@ float merand48(uint64_t& initial) {
   return bits_to_float(temp) - 1;
 }
 
+float merand48(unsigned long& initial) {
```
jmorra: OSX complained at compile time unless this was added.

java/src/main/java/vw/responses/ActionScores.java
```diff
@@ -0,0 +1,36 @@
+package vw.responses;
+
+import java.util.Arrays;
+
+/**
+ * Created by jmorra on 8/12/16.
+ */
+public class ActionScores {
+
+    private final ActionScore[] actionScores;
```
deaktator: Is it weird that you have a mutable "*collection *" that is private but then you return it directly in the getter method?  Should this be a `java.util.List` that is backed by `Collections.unmodifiableList`?  Maybe it will just suffice to make a javadoc comment that says that the backing array is returned directly and let the users make their own mistakes if they so choose.

jmorra: I agree.  Writing the JNI for a `Collections.unmodifiableList` would be very cumbersome and I'd rather just add a comment.

java/src/main/java/vw/responses/Multilabels.java
```diff
@@ -0,0 +1,35 @@
+package vw.responses;
+
+import java.util.Arrays;
+
+/**
+ * Created by jmorra on 8/12/16.
+ */
+public class Multilabels {
+
+    private final int[] labels;
```
jmorra: Comment added.

deaktator:
If I haven't told you @jmorra, these changes look great.

jmorra:
@JohnLangford I feel pretty good about the current state of the branch and after a final look from you I'm good merging it in.

vowpalwabbit/lda_core.cc
JohnLangford: This is fixed now, right? (With the elimination of topic_predictions.)

vowpalwabbit/rand48.cc
```diff
@@ -20,6 +20,14 @@ float merand48(uint64_t& initial) {
   return bits_to_float(temp) - 1;
 }
 
+#ifdef __APPLE__
```
JohnLangford: Why do we need an extra merand48?

java/src/main/c++/vw_learner_VWScalarsLearner.h
```diff
@@ -0,0 +1,30 @@
+/* DO NOT EDIT THIS FILE - it is machine generated */
```
JohnLangford: Can we keep this out of the repository and machine generate on demand?  (What is the generation mechanism?)

.gitignore
JohnLangford: Maybe change filenames instead? 'Vowpal' instead of VW?

vowpalwabbit/lda_core.cc
jmorra: This seems to be working in this case.  My concern was in other reductions.  Since the `pred_type` doesn't need to be specified, this might actually be broken in some other spots.

vowpalwabbit/rand48.cc
jmorra: Without this VW doesn't build on my machine.  Here is the output

```
/usr/bin/g++ -std=c++0x   -D__extern_always_inline=inline -Wall -pedantic -DNDEBUG -O3 -fomit-frame-pointer -fno-strict-aliasing  -D_FILE_OFFSET_BITS=64 -I /opt/local/include  -fPIC  -c oaa.cc -o oaa.o
parse_regressor.cc:40:24: error: no matching function for call to 'merand48'
{*iter = (float)(0.1 * merand48(ind));
                       ^~~~~~~~
./rand48.h:5:7: note: candidate function not viable: no known conversion from 'size_t' (aka 'unsigned long') to 'uint64_t &' (aka 'unsigned long long &') for 1st argument
float merand48(uint64_t& initial);
      ^
parse_regressor.cc:44:18: error: no matching function for call to 'merand48'
{*iter = (float)(merand48(ind) - 0.5);
                 ^~~~~~~~
./rand48.h:5:7: note: candidate function not viable: no known conversion from 'size_t' (aka 'unsigned long') to 'uint64_t &' (aka 'unsigned long long &') for 1st argument
float merand48(uint64_t& initial);
      ^
2 errors generated.
make[1]: *** [parse_regressor.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make: *** [vw] Error 2
```

I found the fix [here](http://stackoverflow.com/questions/11603818/why-is-there-ambiguity-between-uint32-t-and-uint64-t-when-using-size-t-on-mac-os).

java/src/main/c++/vw_learner_VWScalarsLearner.h
jmorra: We really shouldn't.  This is generated by [javah](http://docs.oracle.com/javase/7/docs/technotes/tools/windows/javah.html).  It just generates the headers.  I think the repo would look pretty weird if it just contained the .cc files.  It might be doable but I don't see the point because if the signature changed on the JNI side that would cause these files to be generated differently then the .cc files would have to change anyway.

jmorra:
@JohnLangford I've renamed a whole bunch of classes on both sides of the Java layer to revert back my gitignore changes.  @deaktator do you think these class names are ok?

jmorra:
@JohnLangford the build appears to be broken on my machine again.  We really need to consider adding a travis build for Mac.  I'm going to try and fix it now.  Here are my errors.

```parse_regressor.cc:64:16: error: no matching member function for call to 'set_default'
          all.weights.set_default<random_positive>();
          ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
./array_parameters.h:140:14: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'T'
        inline void set_default()
                    ^
./array_parameters.h:147:14: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'T'
        inline void set_default()
                    ^
./array_parameters.h:155:14: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'T'
        inline void set_default()
                    ^
./array_parameters.h:163:7: note: candidate function template not viable: requires single argument 't', but no arguments were provided
        void set_default(T t)
             ^
parse_regressor.cc:66:16: error: no matching member function for call to 'set_default'
          all.weights.set_default<random_weights>();
          ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
./array_parameters.h:140:14: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'T'
        inline void set_default()
                    ^
./array_parameters.h:147:14: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'T'
        inline void set_default()
                    ^
./array_parameters.h:155:14: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'T'
        inline void set_default()
                    ^
./array_parameters.h:163:7: note: candidate function template not viable: requires single argument 't', but no arguments were provided
        void set_default(T t)
             ^
2 errors generated.
make[1]: *** [parse_regressor.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make: *** [vw] Error 2
```

vowpalwabbit/lda_core.cc
JohnLangford: pred_type inherits from the base learner: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/learner.h#L260  
just like predictions inherit from the base learner. 

We could imagine enforcing pred_type specification everywhere, but this seems awkward because there are some reductions which do not touch the prediction.

vowpalwabbit/rand48.cc
JohnLangford: There was an issue posted that should be fixed in the master now.  Fix = convert size_t to uint64_t elsewhere.  This seems better because it promotes cross-platform compatibility and less code.

JohnLangford:
Try current master---I just merged a pull request that fixed some size_t/uint64_t issues for macs.  

I'm all for a travis-on-macs

jmorra:
I was able to get everything to build again and Travis is now passing.  I think this is ready to be merged in.

JohnLangford:
Status on this?

jmorra:
I think this is ready to be merged in.  Everything is building and the changes outside the java folder have been minimized.

JohnLangford:
Merged in. Thanks!

I'm still not keen on the machine generated files.  If you can think of a way to remove and autogenerate them, that's cool.

PR #1069, opened 2016-08-10
ADF offline eval / quiet;-t improvements

eisber:
@lhoang29 can you review?

@hmmroger can you review?


cs/azure/Trainer/Operations/EvalOperation.cs
```diff
@@ -138,17 +139,23 @@ private IEnumerable<EvalData> OfflineEvaluateInternal(TrainerResult trainerResul
...
+
                 yield return new EvalData
                 {
-                    PolicyName = $"Constant Policy {action}",
+                    PolicyName = $"Constant Policy {tag}",
                     JSON = JsonConvert.SerializeObject(
                     new
                     {
-                        name = $"Constant Policy {action}",
+                        name = $"Constant Policy {tag}",
```
lhoang29: does name = PolicyName work?

cs/cli/AssemblyInfo.cpp
```diff
@@ -12,5 +12,5 @@ using namespace System::Security::Permissions;
 [assembly:AssemblyCopyrightAttribute("Copyright (C) 2014")];
 [assembly:AssemblyTrademarkAttribute("Copyright (C) Microsoft Corp 2012-2016, Yahoo! Inc. 2007-2012, and many individual contributors. All rights reserved")];
 [assembly:AssemblyCultureAttribute("")];
-[assembly:AssemblyVersionAttribute("8.2.0.11")];
-[assembly:AssemblyFileVersion("8.2.0.11")];
+[assembly:AssemblyVersionAttribute("8.2.0.10")];
+[assembly:AssemblyFileVersion("8.2.0.10")];
```
lhoang29: hmm this seems odd, downgrade expected?

cs/cli/vw_base.cpp
```diff
@@ -49,15 +49,17 @@ namespace VW
                 if (settings->Model != nullptr)
                 {
                     m_model = settings->Model;
-                    m_vw = VW::seed_vw_model(m_model->m_vw, string);
+					if (!settings->Verbose && !settings->Arguments->Contains("--quiet") && !m_model->Arguments->CommandLine->Contains("--quiet"))
+						string.append(" --quiet");
+					m_vw = VW::seed_vw_model(m_model->m_vw, string);
```
lhoang29: style nazi: indentation is off

lhoang29:
looks good otherwise

cs/azure/Trainer/Operations/EvalOperation.cs
eisber: ?

cs/cli/AssemblyInfo.cpp
eisber: I’ll upgrade ones you’re ok…

cs/azure/Trainer/Operations/EvalOperation.cs
lhoang29: Just wanted to reuse the string that was set to PolicyName here, so something like:

PolicyName = $". . .",
. . .
new
{
    name = PolicyName,
. . .

eisber: Doesn’t work. Introduced a variable…

cs/cli/vw_base.cpp
eisber: Replace \t with space

eisber:
@jmorra this might help with predictions.
@JohnLangford can you review too?

jmorra:
This looks beautiful for the JNI layer.  @JohnLangford let's hold off on merging my branch until this is merged in.  Based on this work I think I can refactor the current branch to be even better.  @deaktator let's discuss this offline as well to make sure we're on the same page.

vowpalwabbit/oaa.cc
```diff
@@ -89,19 +91,23 @@ void predict_or_learn(oaa& o, LEARNER::base_learner& base, example& ec)
 
   if (is_probabilities)
   { float sum_prob = 0;
-    ec.pred.probs = calloc_or_throw<float>(o.k);
+    float* probs = calloc_or_throw<float>(o.k);
```
JohnLangford: Can we avoid a spurious allocation?

eisber: This is mainly to avoid: ec.pred.probs. You rather have it spread all over (ec.pred.probs[i] = prob;)?

float* probs = calloc_or_throw<float>(o.k);
    for (uint32_t i=0; i<o.k; i++)
    { // probability of class (i+1) = logistic_link_function(raw_prediction)
      float prob = 1.f / (1.f + exp(- o.pred[i].scalar));
      probs[i] = prob;
      sum_prob += prob;
    }
    // make sure that the probabilities sum up (exactly) to one
    for (uint32_t i=0; i<o.k; i++)
      probs[i] /= sum_prob;

       ec.pred.probs = probs;

JohnLangford:
The current approach has the property that it can be done wrong since you must explicitly type the prediction.  Maybe we should instead make it so that you cannot do it wrong?  

One approach is to add an overloaded set_prediction() function which tags the appropriate type to the example struct.  This would also eliminate many lines of code since you no longer explicitly type the predictions.

I'm also hesitating a little bit between typing the prediction type of the example and typing the prediction type of the learner.  Doing it for the learner has the advantage that it adds zero overhead to the learning and will (eventually) allow us to check compatibility of any sequence of reductions.  Why do we want to do it in the example rather than in the learner?

jmorra:
I actually agree with @JohnLangford about typing the learner as once that's established it fundamentally cannot change for each prediction.  One more question, how can we know in general the size of the `probs` array?  It seems to vary depending on the learning type.  Could we make `probs` a `v_array` or have the length easily knowable in general?

deaktator:
@JohnLangford, I agree with @jmorra. Setting this `prediction_type::prediction_type_t prediction_type` value (in `example.h`) is a huge help and we've talked before about how this would simplify things from the *JNI* perspective.  We can perform a prediction at model construction time and determine the output type from the enum.  This eliminates guessing.  Assuming this return type is consistent, we can set up the return value extraction algorithm once and future calls on the predict method should be fast.

eisber:
I actually played quite a bit with make this automatic yesterday, replacing the union and adding “some” auto tracking when the field (e.g. scalar vs a_s) is accessed. I ended up reverting it, due to the overhead it introduced.
Throughout the lifetime of an example the prediction type is actually changed (e.g. a reduction calling into the base).

Markus

jmorra:
While I understand that the prediction type may change as it goes through a series of reductions, isn't a VW model nothing more than an ordered list of reductions which is known at instantiation time?  Assuming this is true then the type for that specified order of reductions can also be known at example time.  If this this is not true, please correct me.

eisber:
Maybe this boils down to different question: do we want prediction type support only on the surface C++ <-> Java/C# or also for intra-reduction communication.
If this is only intended for the surface, setting it per learner seems fine.

Markus

jmorra:
@eisber your point is well taken.  Maybe the best approach would be to do both?  @JohnLangford what do you think?

JohnLangford:
Why doesn't typing the learners (aka reductions) address intra-reduction
communication?

-John

eisber:
I see… Maybe we should look at this together next week. I’m not entirely sure where to place the flag. Should we also move the label field then?

JohnLangford:
The label field itself needs the present communication pattern so I don't
think that can move to the learner.  We might be able to strongly type it
however.

-John

jmorra:
One more point to consider is that the LDA model does not set one of these types, but instead sets the topic_predictions variable.  How should this fold into this new design?

eisber:
Good observation… didn’t know what to do about it either…

JohnLangford:
"What topic_predictions"?  I converted everything to use pred.scalars :-)

jmorra:
@JohnLangford, now come on, my point was valid when I asked it ;).

jmorra:
I now have this working on the JNI side assuming the following code (or something like it) will exist in this branch

```
switch (vwInstance->prediction_type_t)
  { case prediction_type::action_scores:
      field = env->GetStaticFieldID(clVWReturnType , "ActionScores", RETURN_TYPE_INSTANCE);
      break;
    case prediction_type::multiclass:
      field = env->GetStaticFieldID(clVWReturnType , "Multiclass", RETURN_TYPE_INSTANCE);
      break;
    case prediction_type::multilabels:
      field = env->GetStaticFieldID(clVWReturnType , "Multilabels", RETURN_TYPE_INSTANCE);
      break;
    case prediction_type::prob:
      field = env->GetStaticFieldID(clVWReturnType , "Prob", RETURN_TYPE_INSTANCE);
      break;
    case prediction_type::probs:
      field = env->GetStaticFieldID(clVWReturnType , "Probs", RETURN_TYPE_INSTANCE);
      break;
    case prediction_type::scalar:
      field = env->GetStaticFieldID(clVWReturnType , "Scalar", RETURN_TYPE_INSTANCE);
      break;
    case prediction_type::scalars:
      field = env->GetStaticFieldID(clVWReturnType , "Scalars", RETURN_TYPE_INSTANCE);
      break;
  }
```

I am able to test it fully and everything works except for the probs type.  I still do not have a way to get the size of the `float*` array given just the `example`.  Once this is solved we should be good to go within JNI.

eisber:
@JohnLangford this includes the learner::prediction_type change
@jmorra I think this should suite you well...

jmorra:
I think this looks great.  My only concern is the `float*` for the `probs` value.  Could this be fixed here?  If not, @JohnLangford where could we handle this?

vowpalwabbit/multilabel_oaa.cc
```diff
@@ -57,10 +57,10 @@ LEARNER::base_learner* multilabel_oaa_setup(vw& all)
   data.k = all.vm["multilabel_oaa"].as<size_t>();
 
   LEARNER::learner<multi_oaa>& l = LEARNER::init_learner(&data, setup_base(all), predict_or_learn<true>,
-                                   predict_or_learn<false>, data.k);
+                                   predict_or_learn<false>, data.k, prediction_type::multilabels);
   l.set_finish_example(finish_example);
   all.p->lp = MULTILABEL::multilabel;
-  all.delete_prediction = MULTILABEL::multilabel.delete_label;
```
JohnLangford: Why is this line getting removed?  Looks like a bug?

JohnLangford:
I fixed the plausible bug and merged this.

JohnLangford:
I've merged this in.  The major change for everyone writing a library
interface is that the prediction of the top level learner in the reduction
stack is now dynamically typed here:

https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/learner.h#L113


This should be super helpful.

-John

PR #1077, opened 2016-08-16
Fix expected file for big_test 4a

danmelamed:
also make big_tests/Makefile more compatible with systems like Cygwin, which don't have lscpu by default


JohnLangford:
Merged, thanks.

PR #1079, opened 2016-08-17
Ensure `rotl32` will be done in a single instruction

yazevnul:


arielf:
@yazevnul 

Thanks! this could be a significant optimization since the hash function is a significant part of the reading/parsing thread.  Have you checked the effect on speed?  Also: this seems IA64 only, what about other architectures? I think the `ifdef` should be extended to Intel/AMD only.

JohnLangford:
Was gcc failing to effectively use rotl?

yazevnul:
@JohnLangford it seems like GCC was able to optimize it into `rotl`, see https://godbolt.org/g/cvuwzr

In codebase I'm working with `rotl32` was implemented a little different and compiler failed to optimize it into a single instruction.

Since compiler is so good here, should we keep going with this PR? Or should I discard it?

arielf:
This (and many others) should be a pretty standard peephole-optimization in any modern (recent ~10 years) compiler.  Moreover: native compiler code-generators often know a very large number of per-architecture and model tricks to replace same-block instruction sequences by the fastest equivalent for the target processor (e.g. shifts and adds instead of multiply by a constant, if the hardware multiplier is relatively slow). Lastly, since `rotl32` is defined `inline` the opportunities for optimizations are even greater when combined with surrounding code (and global register allocation schemes).

`asm` statements essentially tell the compiler "I know better then you" and disable optimizations so they should only be used as a last resort.  That's why I asked if a speed-up is actually observed here.  Otherwise I would be uncomfortable applying this.  John asked an even better question: is the compiler not doing this optimization (or an even better one) by itself?

Adding `-S` and `-fverbose-asm` to the `g++` command line allows looking at the assembly code.

yazevnul:
Since you decided such a long reply. Hire is an implementation that compiler wasn't able to optimize into a single instruction:
```
# include <stdint.h>
# include <assert.h>

uint32_t rotl32(uint32_t value, const uint8_t shift) noexcept {
    constexpr uint8_t bits = sizeof(uint32_t) * 8;
    constexpr uint8_t mask = bits - 1;
    assert(shift <= mask);

    return (value >> shift) | (value << ((-shift) & mask));
}
```

Clang 3.7.1 asm (succeeded): [link](https://gcc.godbolt.org/#compilers:!((compiler:clang371,options:'-O2+-std%3Dc%2B%2B11+-Wall',source:'%23+include+%3Cstdint.h%3E%0A%23+include+%3Cassert.h%3E%0A%0Auint32_t+rotl32(uint32_t+value,+const+uint8_t+shift)+noexcept+%7B%0A++++constexpr+uint8_t+bits+%3D+sizeof(uint32_t)+*+8%3B%0A++++constexpr+uint8_t+mask+%3D+bits+-+1%3B%0A++++assert(shift+%3C%3D+mask)%3B%0A%0A++++return+(value+%3E%3E+shift)+%7C+(value+%3C%3C+((-shift)+%26+mask))%3B%0A%7D')),filterAsm:(commentOnly:!t,directives:!t,labels:!t),version:3)
GCC 4.7.3 asm (failed): [link](https://gcc.godbolt.org/#compilers:!((compiler:g473,options:'-O2+-std%3Dc%2B%2B11+-Wall',source:'%23+include+%3Cstdint.h%3E%0A%23+include+%3Cassert.h%3E%0A%0Auint32_t+rotl32(uint32_t+value,+const+uint8_t+shift)+noexcept+%7B%0A++++constexpr+uint8_t+bits+%3D+sizeof(uint32_t)+*+8%3B%0A++++constexpr+uint8_t+mask+%3D+bits+-+1%3B%0A++++assert(shift+%3C%3D+mask)%3B%0A%0A++++return+(value+%3E%3E+shift)+%7C+(value+%3C%3C+((-shift)+%26+mask))%3B%0A%7D')),filterAsm:(commentOnly:!t,directives:!t,labels:!t),version:3)
GCC 4.8.2 asm (failed): [link](https://gcc.godbolt.org/#compilers:!((compiler:g482,options:'-O2+-std%3Dc%2B%2B11+-Wall',source:'%23+include+%3Cstdint.h%3E%0A%23+include+%3Cassert.h%3E%0A%0Auint32_t+rotl32(uint32_t+value,+const+uint8_t+shift)+noexcept+%7B%0A++++constexpr+uint8_t+bits+%3D+sizeof(uint32_t)+*+8%3B%0A++++constexpr+uint8_t+mask+%3D+bits+-+1%3B%0A++++assert(shift+%3C%3D+mask)%3B%0A%0A++++return+(value+%3E%3E+shift)+%7C+(value+%3C%3C+((-shift)+%26+mask))%3B%0A%7D')),filterAsm:(commentOnly:!t,directives:!t,labels:!t),version:3)
GCC 4.9.2 asm (succeeded): [link](https://gcc.godbolt.org/#compilers:!((compiler:g492,options:'-O2+-std%3Dc%2B%2B11+-Wall',source:'%23+include+%3Cstdint.h%3E%0A%23+include+%3Cassert.h%3E%0A%0Auint32_t+rotl32(uint32_t+value,+const+uint8_t+shift)+noexcept+%7B%0A++++constexpr+uint8_t+bits+%3D+sizeof(uint32_t)+*+8%3B%0A++++constexpr+uint8_t+mask+%3D+bits+-+1%3B%0A++++assert(shift+%3C%3D+mask)%3B%0A%0A++++return+(value+%3E%3E+shift)+%7C+(value+%3C%3C+((-shift)+%26+mask))%3B%0A%7D')),filterAsm:(commentOnly:!t,directives:!t,labels:!t),version:3)

arielf:
Great work, thanks.
Ubuntu 16.04 has gcc 5.4.0 and clang 3.8.0 which are also a success.
Obviously not all versions succeed here.  Good to know.
But `asm` statements have (long-term) costs and in this particular case, the `inline` effect (a separate optimization) may be lost when defining the function as `asm`.

arielf:
BTW: it is important to compile with the same optimization options we use when we compile `vw`:

     `-O3 -fomit-frame-pointer -DNDEBUG -std=c++0x -Wall`

Which gives me (success):
```
rotl32(unsigned int, unsigned char):
        movl    %edi, %eax
        movl    %esi, %ecx
        rorl    %cl, %eax
        ret
```

yazevnul:
Yep, but `-O3` is not always an option. In some codebases `-O3` is deprecated as it's exploits undefined behaviour too much.

PR #1087, opened 2016-08-21
fix mac build

### LINKED ISSUE #1086: Build fails on Mac OS X after commit 96c3756
The current master branch (commit 08bb806) fails to build on Mac OS X with the following error message:

```
parse_regressor.cc:40:24: error: no matching function for call to 'merand48'
{*iter = (float)(0.1 * merand48(ind));
./rand48.h:3:7: note: candidate function not viable: no known conversion from 'size_t' (aka 'unsigned long') to 'uint64_t &' (aka 'unsigned long long &') for 1st argument
float merand48(uint64_t& initial);
```

I noticed that a _size_t_ is passed on to _merand48_ instead of _uint64_t_, should be an easy fix. I think this maybe a compiler dependant conversion issue. Build is successful for commit (c593f3a)

bartleusink:
There were a few more places where `size_t` had to be replaced with `uint64_t` to fix #1086

### DIFF
--- a/vowpalwabbit/array_parameters.h
+++ b/vowpalwabbit/array_parameters.h
@@ -137,7 +137,7 @@ public:
 	      T(iter);
 	  }
 	
-	template<void(*T)(iterator&, size_t)> //for random initialization of weights (with stride) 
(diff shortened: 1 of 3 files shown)


JohnLangford:
Merged, thanks.

PR #1093, opened 2016-08-25
CB ADF return type

jmorra:
While working with the recent changes within the JNI layer a problem arose.  We wanted to feed the output of a CB model to the [generic decision service](https://github.com/Microsoft/mwt-ds-explore-java/blob/master/src/main/java/com/mwt/explorers/GenericExplorer.java) within Java.  We noticed that the return type of a CB model was `VWMulticlassLearner` just as we would expect and the output of CB explore and CB ADF explore were both `VWActionScoresLearner` as we would expect.  The problem is that the output of CB ADF is also `VWActionScoresLearner`.  This poses a problem because in the case of CB ADF we only want the one best action (unless `--rank_all` is supplied).

This pull request addresses this issue by making the `prediction_type` `multiclass` for CB ADF models and keeps is `action_scores` for CB ADF with `--rank_all` and both explore learners.

@JohnLangford can you please review the C side to make sure I didn't break anything.  @deaktator can you please review the Java side to make sure the tests look good.


vowpalwabbit/cb_adf.cc
```diff
@@ -188,6 +188,8 @@ void output_example(vw& all, cb_adf& c, example& ec, v_array<example*>* ec_seq)
   float loss = 0.;
 
   uint32_t action = ec.pred.a_s[0].action;
+  ec.pred.a_s.delete_v();
```
jmorra: @JohnLangford is this OK to do here?  I think it is cause we don't need the `a_s` anymore in this case just the `action`.

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -551,7 +551,7 @@ base_learner* cb_explore_adf_setup(vw& all)
   all.p->lp = CB::cb_label;
   all.label_type = label_type::cb;
 
-  learner<cb_explore_adf>& l = init_learner(&data, base, CB_EXPLORE_ADF::predict_or_learn<true>, CB_EXPLORE_ADF::predict_or_learn<false>, problem_multiplier);
```
deaktator: Seems like reduction authors should have to specify this type explicitly.  Weird that there's a default type.

vowpalwabbit/cb_adf.cc
JohnLangford: It's not.  There are two issues:
(1) VW generally follows a strategy of allocation at the beginning of the computation and deallocation at the end of the computation. This violates that, and a deallocation here implies that you must insert an allocation somewhere else.  
(2) pred is a union so extra care must be taken.  Here, you are calling delete_v on a piece of memory interpreted as a v_array<action_score>, then using the same memory as a uint32_t.  This will cause problems...

JohnLangford:
I'm not quite understanding what the issue is.  

We ended up tweaking cb_adf to always use prediction_type::action_scores because it was negligible overhead compared to multiclass and made all the interfaces more uniform.  

Why not use pred.a_s[0] if you want the multiclass prediction?  This seems simpler and more minimal.

jmorra:
The problem with the current interface is that the return type alone does not say what to do with the returned value.  In my case we are wrapping the JNI code in [Aloha](https://github.com/eHarmony/aloha) and this poses a big design problem.  If a user creates a model which returns a `VWActionScoresLearner` then the return type itself does not say whether we should just look up the first action and return it with probability 1 or take the entire array and feed it to decision service to choose which action to use.  Without this change we'll be relying on the user to submit an additional flag which states whether or not to use decision service which would expose the existence of decision service to the users of our models, something I'd rather not do.

vowpalwabbit/cb_adf.cc
jmorra: Yea, I figured I was doing it wrong, it didn't feel right when I was writing it.  Can you give me some pointers about how I could do this better?

JohnLangford:
If I understand correctly, you would rather have a flag for VW which
controls whether the return type is multiclass or action_scores then a flag
in Aloha for doing the conversion there?

If you want to do it in VW, I think the cleanest solution is to create a
reduction which does the conversion.  This could then be applied to
action_score producing predictor.

-John

jmorra:
I don't know if a flag is necessary, I would think it can be inferred by the current parameters.  Do you think a flag supplied at train time makes more sense?

JohnLangford:
Not really.   Why can't you have a java side function which looks like this:

get_multiclass_prediction()
{
if (prediction_type == action_scores)
  return pred.a_s[0].action;
else if (prediction_type == multiclass)
  return pred.multiclass;
}

?

-John

deaktator:
## Problem
@JohnLangford: after being walked through the code, I think I understand what @jmorra is saying.  The issue is that @jmorra wants a specific concrete type to be returned by each type of model.  Currently it seems that for `cb-adf` there is a disparity between the class of model (one that returns action scores) and the type the model outputs, a single action.

## Details

If I'm not mistaken, `cb-adf` places `ACTION_SCORE::action_scores` in the `example` but just outputs the best action to standard out in `output_example` in `cb_adf.cc`.  This seems to cause an issue because there seems to be a disparity between the class/type of model and the type of value outputted.  This seems OK on the *C* side because the value is being printed out and subsequently cleaned up.  By printing and returning a `void` return type, this issue doesn't come up.  However, if the `output_example_seq` procedure was a function that returned a prediction value of a specific, non-union type, the problem would arise because you'd expect back a single value (like in the *C* print statement) but you'd get back a collection of action scores.  

If you wanted to extract the best action and return that value, analogous to the print statement seems to do, you wouldn't know under what circumstances to do it. Since the `prediction_type` is `action_scores` (*line 414*, `cb_adf.cc`) and not `multiclass`, you don't know to extract the just the best value.

I think the only way to solve this treating the `C` code as a black box seems to be interrogating the command line string.  I think this is what @jmorra was talking about in the comment

> I don't know if a flag is necessary, I would think it can be inferred by the current parameters. 
> Do you think a flag supplied at train time makes more sense?

Instead of doing that, I think that @jmorra wanted to make the return type of `cb-adf` more transparent and also more precise. I think this is exactly the issue this PR is trying to solve.  Does that make sense?

## Program trace:

In `cb_adf.cc`:
* 415:  `l.set_finish_example(CB_ADF::finish_multiline_example);`
* 294:  `finish_multiline_example` calls `{ output_example_seq(all, data);`
* 274:  `output_example_seq ` calls `{ output_example(all, data, **(data.ec_seq.begin()), &(data.ec_seq));`
* 190: `output_example` sets `uint32_t action = ec.pred.a_s[0].action;`
* 207: `output_example` calls `all.print(sink, (float)action, 0, ec.tag);`

JohnLangford:
So the problem is the inconsistency between cb_adf returning a single action by -p but an action_scores by the library interface? 

What we could do is a modify cb_adf to return action_scores always (via -p or the library interface) and add a new reduction which converts this into a uint32_t returning only the uint32_t (via -p or the library interface). 

Is that good?  Having different return types dependent on flags for a single reduction seems somewhat messy and I could imagine this "just give me the best" reduction being useful elsewhere.

jmorra:
I think that seems reasonable.  That reduction would have the type multiclass which would solve this issue.  I would assume that `--rank_all` would not go to this new type of reduction, is that correct?

I've never written a reduction before, so if you could give me some places in the code to look for examples that'd be really helpful.

JohnLangford:
It's going to be a bit messy, because the multiline reductions are not very
clean yet (we need to standardize some boilerplate).   The functionalities
you want are in cbify.cc (which converts to uint32_t) and explore_eval.cc
(which is the simplest example of dealing with the multiline stuff).

-John

jmorra:
Given that this is going to be refactored anyway should I just wait on merging this in until after that is done?  I'm honestly really struggling with how to do this without a LOT of code duplication.

JohnLangford:
Yeah, it's ugly.

I'm on vacation this week, but will look into things next week.

-John

jmorra:
I've taken the recent work that @JohnLangford has done and incorporated it in here.  I think this will solve our issues.  Looking at the new tests we can differentiate between when a decision needs to be made probabilistically vs deterministically just by observing the type.  If a type `ActionScores` is returned then the caller should ALWAYS choose the first action in the list and ignore the score (unless performing softmax).  If a type `ActionProbs` is returned then the caller should ALWAYS use an exploration library to choose the action.

@deaktator please let me know what you think
@JohnLangford please verify the above statements are correct

If everyone agrees with what is done then this is ready to be merged back in.

java/src/main/java/vowpalWabbit/responses/ActionProb.java
```diff
@@ -0,0 +1,42 @@
+package vowpalWabbit.responses;
+
+/**
+ * Created by jmorra on 8/12/16.
+ */
+public class ActionProb {
+
+    private final int action;
```
deaktator: Any harm in making these public?

deaktator: I guess you probably want to "*program to interfaces*" and whatever.

java/src/main/java/vowpalWabbit/responses/ActionProbs.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Arrays;
+
+/**
+ * Created by jmorra on 8/12/16.
+ */
+public class ActionProbs {
+
+    // Although this is modifiable it is not intended to be updated by the user.  This data structure mimics the
+    // C data structure.
+    private final ActionProb[] actionProbs;
```
deaktator: `java.util.List` *vs* `Array`?  Is there a reason to prefer one to another?

java/src/test/java/vowpalWabbit/learner/VWActionProbsLearnerTest.java
```diff
@@ -0,0 +1,136 @@
...
+            "2:1:0.3 | b c",
+            "3:1.5:0.7 | a d"
+        };
+
+        VWActionProbsLearner vw = VWLearners.create("--quiet --cb_explore 4 -f " + model);
+        ActionProbs[] trainPreds = new ActionProbs[cbTrain.length];
+        for (int i=0; i<cbTrain.length; ++i) {
+            trainPreds[i] = vw.learn(cbTrain[i]);
+        }
+        ActionProbs[] expectedTrainPreds = new ActionProbs[]{
```
deaktator: Stupid thought, but do you think this would be more readable as just:

```java
ActionProbs[] expectedTrainPreds = new ActionProbs[]{
  actionProbs(0.9625f, 0.0125f, 0.0125f, 0.0125f),
  actionProbs(0.0125f, 0.9625f, 0.0125f, 0.0125f),
  ...
};
```

Then you could have a function, `actionProbs` that takes a `float ...` *varargs* and returns a `ActionProbs` by zipping the probability with the index in the *varargs* array.  This seems like it would be more readable and more terse.

java/src/test/java/vowpalWabbit/learner/VWActionScoresLearnerTest.java
```diff
@@ -18,71 +18,33 @@
...
 
     @Test
     public void testCBADFWithRank() throws IOException {
-        VWActionScoresLearner vw = VWLearners.create("--quiet --cb_adf --rank_all");
+        testCBADF(true);
+    }
+
+    private void testCBADF(boolean withRank) throws IOException {
+        String[][] cbADFTrain = new String[][]{
+            new String[]{"| a:1 b:0.5","0:0.1:0.75 | a:0.5 b:1 c:2"},
```
deaktator: I would replace the `,` with `\n,` so that you can see the multiline input better.

java/src/main/java/vowpalWabbit/responses/ActionProb.java
jmorra: Yea, I'm gonna leave it like this cause it's more Java-y (although it's super verbose).

java/src/main/java/vowpalWabbit/responses/ActionProbs.java
jmorra: I'm emulating the C++ side here.  Plus array is a little faster.

java/src/test/java/vowpalWabbit/learner/VWActionProbsLearnerTest.java
jmorra: Yup, will fix.

java/src/test/java/vowpalWabbit/learner/VWActionScoresLearnerTest.java
jmorra: '\n' is not important for multiline input through the library so I'm not sure what you're referring to.

java/src/test/java/vowpalWabbit/learner/VWActionProbsLearnerTest.java
jmorra: I can't do all the stuff you wanted because for some of the cases the index is NOT the action.  Either way I think this is more readable.

JohnLangford:
Status on this?

jmorra:
I think it's good.  I made some of the changes @deaktator requested and responded to all of them.

JohnLangford:
Merged, thanks.

PR #1101, opened 2016-09-09
consider precision

lhoang29:
The policy probabilities can sometimes be > 1 with very small precision leading to unjustified violations


JohnLangford:
Merged, thanks.

PR #1102, opened 2016-09-12
Bug fixes

eisber:
Fixed NRE on empty hashes
Skip model load/initialize when seeding from in-memory model
Fixed progressive validation in Azure trainer
Includes mixed JSON string and JSON direct support
Includes native C++ JSON parsing


eisber:
@jmorra Java build seems broken. Seems like on master too. Are you looking into this?
I'm trying to build the java part on Ubuntu (in Windows) and am getting 

make[1]: *** No rule to make target `pom.xml', needed by `pom_version'.  Stop 

any hints?

jmorra:
I'm not in the office now. I will check once I get in.

jmorra:
I have an open PR that addresses since Java issue. Before I look at this I think it makes sense to merge that back in. And have it pulled into this branch to make sure the issue is still here. Does that make sense?

jmorra:
OK, my latest Java changes have been merged into master.  Once it's pulled in here, if this build is still failing I'll look into it.

lhoang29:
Github view is too sluggish so let's do this here:

parse_example_json.cc:L250 It'd be safer if 7, 6 and 2 are all computed on actual string length, but these seem self contained so it's not too bad currently either

parse_example_json.cc:L329 I suppose if there are multiple spaces this will split multiple times, is that desired?

parse_example_json.cc:L364 Minor: compute length of "shared" instead of 6

parse_example.cc:L62 What's the effect of empty examples when not in cb/cb_adf mode? Does VW just ignore it?

rapidjson* Do you need the files under \bin and \doc? Or do you need to include the entire library for license purposes?

lhoang29:
I primarily went over the json parsing part, looks good in general

eisber:
parse_example_json.cc:L250
parse_example_json.cc:L364
this is an optimization. Tried to follow http://stackoverflow.com/questions/25890784/computing-length-of-a-c-string-at-compile-time-is-this-really-a-constexpr but seems to be another portability adventure. Keeping as is.

parse_example_json.cc:L329 very good catch!

parse_example.cc:L62 this is reduction dependent… from what I’ve seen its ignored.

rapidjson* there’s a nuget package on Windows (although some older version), but there is no maintained one on linux. I talked with John and this seemed like the least pain for everybody from a build setup perspective.

JohnLangford:
In looking through the public diff, I can't see the parts which are impacting core VW.  What is impacted and how can I see it?

JohnLangford:
The diff seems to be reverting a bugfix in bs.cc?

JohnLangford:
Why this?

--- a/vowpalwabbit/cb_explore_adf.cc
+++ b/vowpalwabbit/cb_explore_adf.cc
@@ -152,8 +152,9 @@ namespace CB_EXPLORE_ADF{
       num_actions--;
 
     data.action_probs.resize(num_actions);
+    data.action_probs.erase();
     for (uint32_t i = 0; i < num_actions; i++)
-      data.action_probs[i] = {i,0.};
+      data.action_probs.push_back({ i,0. });

JohnLangford:
Diff reverts a warning/bugfix in confidence.cc

JohnLangford:
diff reverts a bugfix in gd.cc

JohnLangford:
It's nice that you are nuking some void*'s.  Thanks.

JohnLangford:
I'm not quite following this comment:
+  // note: the json parser does single pass parsing and cannot determine if a shared example is needed.
+  // since the communication between the parsing thread the main learner expects examples to be requested in order (as they're layed out in memory)
+  // there is no way to determine upfront if a shared example exists
+  // thus even if there are no features for the shared example, still an empty example is returned.

Are you saying that there is always a (possibly zero feature) shared example?

JohnLangford:
Why does parse_example_json.h need to have so much stuff in it?  Can't you embed it into parse_example_json.cc?

JohnLangford:
+example* get_unused_example(vw* all)

Why change the signature to have a pointer instead of a reference?


Also, I'm not quite following the flow control here:
   while(>p->done)
-  { example* ae = get_unused_example(*all);
-    if (!all->do_reset_source && example_number != all->pass_length && all->max_examples > example_number
-        && VW::parse_atomic_example(*all, ae) )
-    { VW::setup_example(*all, ae);
-      example_number++;
+  { examples.push_back(VW::get_unused_example(all)); // need at least 1 example 
+       if (!all->do_reset_source && example_number != all->pass_length && all->max_examples > example_number
+        && all->p->reader(all, examples) > 0)
+    { VW::setup_examples(*all, examples);
+      example_number+=examples.size();
+      examples_available=examples.size();
     }

What happens when you need two examples but only have one?

eisber:
In multiline mode there is always a possible zero feature shared example. The underlying problem is that I can’t determine ahead of parsing if I have features for a shared example or not. I need to parse the complete JSON to figure that out, but by then I already allocated examples for each action.

Markus

eisber:
I need to allocate json_parser in parser.cc:575 and don’t like the idea of void* + factory methods (apart from the fact that I templated the complete structure at 1-point in time) …

I haven’t convinced you yet that .h + class + private provide modularity that *most* C++ devs expect?
Also, the move of items into .cc files didn’t stop others in VW codebase to make things public and just peek into data structures…

Markus

eisber:
I didn’t sync upstream.

eisber:
I didn’t sync upstream.

eisber:
@Luong Hoang<mailto:lhoang@microsoft.com> that’s a bug fix Louie introduced for bag implementation.

lhoang29:
yep if I recall correctly, the action_probs array was initialized incorrectly so that size() would return 0. This reduces safety() to a noop and so you wouldn't be able to enforce minimum prob values

JohnLangford:
You only need the pointer, so you should be able to forward declare the 
type, right?

I realize I'm a dinosaur here, but why share when it's unneeded? It 
seems unpleasant to take something that logically fits all in one file 
and split it across 2, one of which will be shared with the rest of the 
codebase.   W.r.t. others, I'm not claiming this is a perfect solution, 
but discouraging antimodular design seems good?

-John

JohnLangford:
Do soon?  I'm flying tomorrow afternoon.

-John

JohnLangford:
Good, thanks.

-John

JohnLangford:
Overall, I think this is about ready to go after you sync to the master.  The other issues are all style vs. content questions.  Please think them through.

JohnLangford:
appveyor fails?

PR #1104, opened 2016-09-13
Fixing compilation warnings in Java

jmorra:
After releasing the last round of Java updates I noticed a few warnings that shouldn't be there.  This fixes those warnings.  In addition I noticed that we were not using the custom exception that was in the Java side, so that is now deleted.

@deaktator, do you want to have a look at this.  It's very basic.


deaktator:
Looks good. 👍

PR #1108, opened 2016-09-23
Some dubug code updated

trufanov-nok:
Looks like the code that is inside `DEBUG_EVAL_COUNT_OF_GEN_FT` got outdated a bit when vw switched to 64bit address space.
These blocks are not compiled by default so it was unnoticed.
The code itself is a selfcheck for `eval_count_of_generated_ft` function. This function estimates number of features simple permutations according to some combinatorics math which is faster than just iterating throw them. And this debugging block additionally iterates throw all interactions and simple counts them and then compare with evaluated result.


JohnLangford:
Merged, thanks.

PR #1114, opened 2016-09-29
Expose multiclass probabilities in c# wrapper

Wirack:
Implemented a factory for the c# wrapper to expose all probabilities for multiclass predictions, which I found was missing with the current implementation.

By examining the source code I found that the probabilities for multiclass predictions are stored in the scalars array, which I thus used to expose the probabilities.

I don't know what is the common practice regarding unit tests for prediction factories of the c# wrapper, but can look into it if necessary.


eisber:
nice work. can you add a unit test to cs_unittest?

Wirack:
Yes I will probably take a look at it this weekend.
Two quick questions:

1.  Do you have some pointers of your acceptance criteria for unit tests?
2. I am not that used to pull requests. Should I make a new pull request with above changes + the added unit tests?

eisber:
1.      In Visual Studio make sure to Test \ Test Settings \ Default Processor Arch \ x64: run all tests

2.      Just push to the fork:branch that you created the PR from, it will automatically show up.

Wirack:
Thanks. Any specific .cs file in cs_unittest I should expand with tests for the new factory, or should I create a new test file?

Wirack:
Have now added some unit tests.
The Travis build fail seems to be related to python and not connected to the c# wrapper modification proposed here.

Let me know if further adjustments have to be performed for the pull request to be accepted.

JohnLangford:
@gramhagen do you know what the python issue is here?

More generally, why do we need a new type here?  We already have an interface (action_probs) for returning a probability distribution over a set.

gramhagen:
It looks like the expected error changed type, I'm pulling in a miniconda image to load the python scikit learn library. I suspect the version of sklearn changed to induce this problem.  I can update the test to be a little more robust.

eisber:
@Wirack can you sync with master?

Wirack:
Done. That resolved the python issue.

PR #1115, opened 2016-10-06
Sklearn0.19

eisber:
fixed exception import

@gramhagen is this fine? I did this with @hal3 and he mentioned users will have to upgrade sklearn to 0.19


gramhagen:
yeah that's fine, but true this will require an upgrade for end users, i'm surprised this changed in sklearn, there's probably a more robust way to address this that is not version dependent.  if i can come up with something I'll send a pr, until now this works. thank!

eisber:
I talked with Andy from sci-learn today (happened to visit our lab). They’re are going to fix this on their end (reverse the inheritance hierarchy for the exception), but it will take a few weeks.
Going forward we should revert my change and we’d be compatible with 0.18 again.

Issue #1107, opened 2016-09-19
latest 8.2.1 release fails to build due to some linking error


david-geiger:
Hi,

On Mageia we tried to build latest vowpal_wabbit 8.2.1 but we get some linking error (undefined reference to `adjust_used_index(vw&)'), see the full compile log:

http://pastebin.com/0LLvdK90

So we have found a workaround to fix temporally this issue to build without "-Wl,--no-undefined" LD flags enabled adding in the spec file this line: 

"%define _disable_ld_no_undefined 1"

Compiled with GCC 5.4.0 and boost 1.60.0

Best regards,
David

david-geiger:
Any comment about this issue?

JohnLangford:
This may be fixed on the master branch.   Try updating?  I can certainly
compile with gcc 5.4.0 there.

-John

david-geiger:
I just tried with the current master branch and I get the same underlinking issue.

Have you tried to compile with?

export LDFLAGS='-Wl,--no-undefined -Wl'

Our build system use as LDFLAGS:

LDFLAGS=' -Wl,--as-needed -Wl,--no-undefined -Wl,-z,relro -Wl,-O1 -Wl,--build-id -Wl,--enable-new-dtags'

JohnLangford:
Can you switch flags around to figure out something reproducible?

I use the default flags to compile.

-John

arielf:
Neither I can reproduce with standard building: i.e just `make`

The pastebin log shows it was run through autogen/configure with several non-standard building options, some security related.  Subsequently, the libtool chain seems to have broken it massively (a lot of symbols defined in libvw.a appear as undefined).

david-geiger:
to figure out something reproducible, you can add this following line before configure/make:

export LDFLAGS=' -Wl,--as-needed -Wl,--no-undefined -Wl,-z,relro -Wl,-O1 -Wl,--build-id -Wl,--enable-new-dtags'

i.e in a rpm spec file:
%build
export LDFLAGS=' -Wl,--as-needed -Wl,--no-undefined -Wl,-z,relro -Wl,-O1 -Wl,--build-id -Wl,--enable-new-dtags'
./configure
make

david-geiger:
So I come with a very good new!

One of our Mageia packager has found what is going wrong and has create a patch:

http://svnweb.mageia.org/packages?view=revision&revision=1057904
[...]

PR #1120, opened 2016-10-24
Fix to https://github.com/JohnLangford/vowpal_wabbit/issues/1107

arielf:
Contributed by david-geiger:
    Modified file:  vowpalwabbit/Makefile.am
    Based on Mageia patch:
        http://svnweb.mageia.org/packages/cauldron/vowpal-wabbit/current/SOURCES/vowpal_wabbit-8.2.1-linking.patch?view=markup&pathrev=1057904

Verified by me on Ubuntu.
I get a smooth build and no errors after applying the patch:

```
   autogen.sh
   make
   make test
```

Many thanks to David (and the anonymous Mageia packager) for the fix.


JohnLangford:
Yep, thanks.

PR #1126, opened 2016-11-01
fix JSON label bug & improve performance

eisber:


JohnLangford:
Where is _stricmp() documented?  

Also, can you update the test case to use >256 examples?

eisber:
https://msdn.microsoft.com/en-us/library/k59z8dwe.aspx

yep, will do.

JohnLangford:
Merged, thanks.

PR #1127, opened 2016-11-03
Update file paths in README.windows.txt

ukaratay:


JohnLangford:
@eisber can you check this?

PR #1130, opened 2016-11-15
min_prob fix for CB::safety

eisber:


JohnLangford:
What are the changes to the .cs files in this patch?

eisber:
I'm recording trainer start time to able to reproduce offline

JohnLangford:
Merged, thanks :-)

PR #1133, opened 2016-11-21
Enable direct method as cb_type for cb_adf reduction

adith387:
Merging a simple update to "--cb_adf --cb_type dr" to enable training with --cb_type dm (the non-ADF reduction already supports dm).


arielf:
@adith387 thanks for this.
Could you also add a small test to `tests/RunTests` to cover this change and document usage+result?

adith387:
Probably true.

In the cb_to_cs case (e.g. see gen_cs_example.h::gen_cs_example_dm(cb_to_cs ...) ), after generating the cs example, the current example's prediction is immediately updated with the argmin from predicted scores. I bet the same can be done in the cb_to_cs_adf case safely.

I can't say I understand the gen_cs_example code though. This patch was the smallest edit to enable dm in cb_adf that I could assert correctness without understanding other parts of the code.

alekh:
Right, but the prediction type is different in adf. So you have to do something a bit more than argmin. Basically, you have to have a sorted list of (action, score) pairs, which are present in the first example of the examples v_array. The current implementation is probably not quite 2x slower, but somewhere thereabouts in runtime and certainly 2x in memory.

adith387:
Can someone (who understands the code better than me) implement that?
I've found "--cb_adf --cb_type dm" useful as a baseline to benchmark against.

JohnLangford:
I checked in a more efficient direct method.  Adith, please take a look and tell us if that is not what you want.

adith387:
Works. Thanks.

Issue #1096, opened 2016-09-06
Test 42 Failed on Mac


adah1972:
Environment: OS X Lion 10.7.5 + MacPorts (clang-3.7 & boost+clang37+no_single+no_static+openmpi+python27)

Procedure:

```
git clone https://github.com/JohnLangford/vowpal_wabbit.git
cd vowpal_wabbit
./autogen.sh CC=clang-mp-3.7 CXX=clang++-mp-3.7 --enable-libc++
./configure CC=clang-mp-3.7 CXX=clang++-mp-3.7 --enable-libc++ --with-boost-program-options=boost_program_options-mt
make
make test
```

I got one failure:

```
…
RunTests: test 41: stderr OK
RunTests: test 42: FAILED: ref(train-sets/ref/lda-2pass-hang.stderr) != stderr(lda-2pass-hang.stderr)
    cmd: ../vowpalwabbit/vw -k -d train-sets/lda-2pass-hang.dat --lda 10 -c --passes 2 --holdout_off
RunTests: test 43: stderr OK
RunTests: test 44: stderr OK
RunTests: test 44: sequence_data.nonldf.test.predict OK
…
```

Any idea whether this is significant or not?

arielf:
Thanks for the report.

Doesn't reproduce on Ubuntu 16.04.1 with clang 3.8.0-2ubuntu4 (tags/RELEASE_380/final)

Can you rerun the failed command and share the `diff` between the `lda-2pass-hang.stderr` and reference?  You may pass `-d` to `RunTests` to get the diff by default.

Looks like this `-d` is missing from `Makefile.am` (which is a bug, I'll fix shortly).

adah1972:
The diff:

```diff
--- train-sets/ref/lda-2pass-hang.stderr	2016-09-06 17:35:21.000000000 +0800
+++ lda-2pass-hang.stderr	2016-09-07 16:58:22.000000000 +0800
@@ -8,20 +8,20 @@
 num sources = 1
 average  since         example        example  current  current  current
 loss     last          counter         weight    label  predict features
-12.813775 12.813775            1            1.0  unknown   0.0000      201
-12.946867 13.079960            2            2.0  unknown   0.0000      220
-13.718190 14.489512            4            4.0  unknown   0.0000      136
-14.787531 15.856872            8            8.0  unknown   0.0000      371
-15.934851 17.082171           16           16.0  unknown   0.0000      138
[...]

PR #1135, opened 2016-11-25
call return_example in lda_core.cc

bartleusink:
Hi John,

I might be missing something but it looks the new `return_example` method in lda_core.cc (bea9ce4) isn't called anywhere.
This PR replaces the calls to `return_simple_example` with `return_example` in lda_core.cc, which makes your fix for #1096 work beautifully :-)


JohnLangford:
Wow, that's a head scratcher.  Thanks much.

PR #1132, opened 2016-11-21
Sparse weight vector

aartibagul:

PR #1136, opened 2016-11-27
Sparse weights

JohnLangford:
This is a version of #1132 which deals with merge conflicts and compiles (at least on linux).  It is still causing test failures.


JohnLangford:
@aartibagul @eisber I merged aarti's branch into a new branch and eliminated change_begin().  @aartibagul, if you have time to work on the compile issues, I'd suggest merging this branch into yours.

vowpalwabbit/array_parameters.h
```diff
@@ -1,214 +1,368 @@
...
+
+	//const iterator
+	const_iterator cbegin() { weight_map::iterator i = _map.begin(); return const_iterator(i,  _stride); }
+	const_iterator cend() { weight_map::iterator i = _map.begin(); return const_iterator(i, _stride); }
+	
+	inline weight& operator[](size_t i)
+	{   uint64_t index = i & _weight_mask;
+		weight_map::iterator iter = _map.find(index);
+		if (iter == _map.end()) 
+		  {     _map.insert(std::make_pair(index, calloc_mergable_or_throw<weight>(_stride)));
```
JohnLangford: @eisber does mergable make sense here?  In particular, how would sharing models work when the model consists of a large number of pieces of memory?

JohnLangford:
This is now working for all tests with --sparse_weights.  LDA and OjaNewton will not actually scale.

@arielf Is there a way to add --sparse_weights to every VW test command?  My crude attempt failed on tests 92 and 96.

JohnLangford:
I've confirmed that the dense case appears to run with no overhead (due to template use).  Using sparse features during training seems to have a cost of ~ x15 in the worst case.  (My expectation is that sparsity will generally be used in testing.)

We need a testing approach.

arielf:
I just send a PR https://github.com/JohnLangford/vowpal_wabbit/pull/1161
which adds a new option `-O ...` to RunTests

You may call `RunTests -O "<additional forced options> ..." ...`
e.g:     `RunTests ... -O --sparse_weights ...`
will force adding `--sparse_weights` to each vw test.

Hope that this is what's requested here.

JohnLangford:
That seems exactly it--thanks Ariel.

cs/cli/vowpalwabbit.cpp
```diff
@@ -727,33 +727,43 @@ cli::array<List<VowpalWabbitFeature^>^>^ VowpalWabbit::GetTopicAllocation(int to
...
+	auto K = (int)m_vw->lda;
+	auto allocation = gcnew cli::array<cli::array<float>^>(K);
+	for (int k = 0; k < K; k++)
+		allocation[k] = gcnew cli::array<float>((int)length);
+
+	// TODO: better way of peaking into lda?
+	auto lda_rho = m_vw->vm["lda_rho"].as<float>();
+
+	// over weights
+	if (m_vw->weights.sparse)
```
eisber: I refactored the code into a single method.

vowpalwabbit/accumulate.cc
```diff
@@ -19,20 +19,26 @@ using namespace std;
...
-{ uint32_t length = 1 << all.num_bits; //This is size of gradient
+void accumulate(vw& all, parameters& weights, size_t offset)
+{ uint64_t length = 1 << all.num_bits; //This is size of gradient
   float* local_grad = new float[length];
 
-  weight_parameters::iterator iter = weights.begin();
-  for (uint32_t i = 0; iter != weights.end(); ++i, ++iter)
-    local_grad[i] = (&(*iter))[offset];
-
+  if (weights.sparse)
```
eisber: why can't this be done on a higher level?

vowpalwabbit/accumulate.cc
```diff
@@ -19,20 +19,26 @@ using namespace std;
...
+    for (uint64_t i = 0; i < length; i++)
+      local_grad[i] = (&(weights.sparse_weights[i << weights.sparse_weights.stride_shift()]))[offset];
```
eisber: sure that i << weights.sparse_weights.stride_shift() this gets optimized out?

again using template function this can be simplified

vowpalwabbit/audit_regressor.cc
```diff
@@ -167,7 +174,12 @@ void finish(audit_regressor_data& dat)
          dat.values_audited << " of " << dat.loaded_regressor_values << " found)." << endl;
 }
 
-
+template<class T>
+void regressor_values(audit_regressor_data& dat, T& w)
+{  for (typename T::iterator iter = w.begin(); iter != w.end(); ++iter)
+		for (weight_iterator_iterator it = iter.begin(); it != iter.end(); ++it)
+			if (*it != 0) dat.loaded_regressor_values++;
```
eisber: I propose to simply use C++ lambda to simplify the code.

vowpalwabbit/bfgs.cc
```diff
@@ -189,341 +187,435 @@ double regularizer_direction_magnitude(vw& all, bfgs& b, float regularizer)
...
+		  (&(*w))[W_COND] += b.regularizers[2 * (w.index()>> weights.stride_shift())];
+			if ((&(*w))[W_COND] > max_hessian)
+				max_hessian = (&(*w))[W_COND];
+			if ((&(*w))[W_COND] > 0)
+				(&(*w))[W_COND] = 1.f / (&(*w))[W_COND];
+		}
+
+	float max_precond = (max_hessian == 0.f) ? 0.f : max_precond_ratio / max_hessian;
+
+	for (typename T::iterator w = weights.begin(); w != weights.end(); ++w)
```
eisber: just use for (auto w = ...

vowpalwabbit/bfgs.cc
```diff
@@ -189,341 +187,435 @@ double regularizer_direction_magnitude(vw& all, bfgs& b, float regularizer)
...
+		return add_regularization(all, b, regularization, all.weights.dense_weights);
 }
 
+template <class T>
+void finalize_preconditioner(vw& all, bfgs& b, float regularization, T& weights)
+{
+	float max_hessian = 0.f;
+
+	if (b.regularizers == nullptr)
+		for (typename T::iterator w = weights.begin(); w != weights.end(); ++w)
```
eisber: just use for (auto w =  ...

Issue #1113, opened 2016-09-28
[Python] Probabilities flag is not working


montenegrodr:
This snippet does not return the expected result:

```
from vowpalwabbit.sklearn_vw import VW

X = ['1 | feature1:2.5',
     '2 | feature1:0.11 feature2:-0.0741',
     '3 | feature3:2.33 feature4:0.8 feature5:-3.1',
     '1 | feature2:-0.028 feature1:4.43',
     '2 | feature5:1.532 feature6:-3.2']

model = VW(
    convert_to_vw=False, loss_function='logistic', oaa=3, probabilities=True
)
model.fit(X)
print model.predict([' | feature1:0.11 feature2:-0.0741'])
```

That results in something like:
**`[ 28036672.]`**

It shouldn't be probabilities like:
**`[0.3, 0.6, 0.1]`**

Thanks.

JohnLangford:
@gramhagen @hal3 I suspect this is a type conversion error.  This should be much more easily avoided now that the prediction type is explicit in the top level learner.

gramhagen:
there's a couple things going on. right now the file exposing methods to python (pylibvw.cc) doesn't have support for arrays of probabilities. I'm not super familiar with how to convert c++ arrays to python objects, so my initial attempt failed.

```
v_array<float> ex_get_multiclass_probabilities(example_ptr ec) { return ec->pred.scalars; }
```

I'm sure there's probably a way around that, but it will take a little manipulating. Accessing the prediction still seems to give the large integer value you noticed, so there's still something funky with how it's accessing the label here:

```
uint32_t ex_get_multiclass_label(example_ptr ec) { return ec->l.multi.label; }
```

I'm a little out of my depth on the c++ side, though. so I'm not completely sure what we should be accessing if not l.multi.label.

JohnLangford:
(I'm similarly ill-informed about the python side.)

To return a vector of probabilities, you need to add --probabilities to the
commandline---otherwise if defaults to returning a single value.  Also
v_array<float> is a type that python is very unlikely to recognize.

I'm a little bit confused about why you mention the label when you want to
predict.  To get the prediction, you want something like ec->pred.scalar
Could that be the issue?

-John

gramhagen:
ok, np I was a bit confused about the labels (not realizing that they have to be encoded as ints)

[...]

PR #1139, opened 2016-11-28
adding support for probabilities flag in pyvw and sklearn_vw

gramhagen:
addressing problem identified in #1113 

pyvw.vw can provide a list of probabilities if initialized with --oaa and --probabilities
- python/tests/test_pyvw::test_oaa_probs()

similarly sklearn_vw.VW can provide a 2D numpy array of probabilities
- python/tests/test_sklearn_vw::test_oaa_probs()


JohnLangford:
This is unquestionably an improvement, but it would be better if the prediction was controlled by the prediction type rather than the label type since these are not always in correspondence and that lack of correspondence will drive constant "oops, need to fix this" types of issues.  For example, I just added "scores" to oaa.cc which also returns a vector.

The prediction type is recorded in the top level predictor here: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/learner.h#L114 and can be accessed via all->l.pred_type.

gramhagen:
Yep makes sense. There are a few issues circling this area. I'll rework it
to use the prediction type.

gramhagen:
@JohnLangford, wanted to make sure you saw the updates. this should be all set now. I based the changes on the approach used in the jni wrapper to leverage the prediction type

JohnLangford:
Merged, thank you.

PR #1149, opened 2016-12-11
fix --initial_weight to initialize weights

denik:
It is ignored since 2f5bc2b075, which seems to be a typo.

See https://github.com/JohnLangford/vowpal_wabbit/commit/2f5bc2b075152c22ae4d2#diff-6e06dc02b6816202f8189cbcb5c08434L43


JohnLangford:
Merged, thank you.

PR #1150, opened 2016-12-14
Azure refactoring / python on windows

eisber:


python/setup.py
```diff
@@ -92,22 +92,30 @@ def run(self):
...
-        env['PYTHON_LIBS'] = '-l {}'.format(find_boost())
-        subprocess.check_call(['make', 'python'], cwd=path.join(here, 'src'), env=env)
         target_dir = path.dirname(self.get_ext_fullpath(ext.name))
         if not path.isdir(target_dir):
             makedirs(target_dir)
-        ext_suffix = 'so' if not system == 'Cygwin' else 'dll'
-        copy(path.join(here, 'src', 'python', '{name}.{suffix}'.format(name=ext.name, suffix=ext_suffix)),
-             self.get_ext_fullpath(ext.name))
-
+        if system == 'Windows':
```
gramhagen: not sure if this is necessary here, the pyvw.dll copy should be handled by line 111, and the zlib.dll shouldn't be necessary if this is being pre-built and the lib is installed via InstallLib
Can you just extract lines 106-109 in the system != 'Windows' case?
Will need to extend the check for Cygwin or Windows at 110

python/setup.py
```diff
@@ -92,22 +92,30 @@ def run(self):
...
+        else:
+            env = environ
+            env['PYTHON_VERSION'] = '{v[0]}.{v[1]}'.format(v=sys.version_info)
+            env['PYTHON_LIBS'] = '-l {}'.format(find_boost())
+            subprocess.check_call(['make', 'python'], cwd=path.join(here, 'src'), env=env)
+            ext_suffix = 'so' if not system == 'Cygwin' else 'dll'
+            copy(path.join(here, 'src', 'python', '{name}.{suffix}'.format(name=ext.name, suffix=ext_suffix)),
+                 self.get_ext_fullpath(ext.name))
+
+class InstallLib(_install_lib):
```
gramhagen: this is great, I had no idea about install_lib

python/setup.py
eisber: I removed the duplicate zlib copy and are now supporting both python2.7 and 3.5

Not sure what you're referring to exactly wrt to 106-109

gramhagen: just meant the final copy of the pyvw artifact is the same, so we could do something like:

```
if system == 'Windows':
  ext_dir = path.join(here, 'bin')
  ext_suffix = 'dll'
else:
   env = environ
   env['PYTHON_VERSION'] = '{v[0]}.{v[1]}'.format(v=sys.version_info)
   env['PYTHON_LIBS'] = '-l {}'.format(find_boost())
   subprocess.check_call(['make', 'python'], cwd=path.join(here, 'src'), env=env)
   ext_dir = path.join(here, 'src', 'python')
   ext_suffix = 'so' if not system == 'Cygwin' else 'dll'

copy(path.join(ext_path, '{name}.{suffix}'.format(name=ext.name, suffix=ext_suffix)), self.get_ext_fullpath(ext.name))
```

but it's not really a big deal

JohnLangford:
What's the status on this?

eisber:
AppVeyor build needed adaptation – fixed now

This patch includes:


-          Azure refactoring to support Service Fabric hosting

-          C++ JSON audit support

-          Various nuget package dependency upgrades

-          A fix wrt to example count restoration (that’s the reason all the CB test cases are updated)

-          Python build on Windows for 2.7 and 3.5

Markus

python/setup.py
JohnLangford: @eisber Can you respond?  I want to merge this.

vowpalwabbit/OjaNewton.cc
```diff
@@ -72,8 +72,8 @@ struct OjaNewton {
 	    const double PI2 = 2.f * 3.1415927f;
 		for (weight_parameters::iterator i = weights.begin(); i != weights.end(); ++i)
 		  for (weights_iterator_iterator<weight> j = i.begin() + 1; j != i.end(m + 1); ++j)
-		  {  float r1 = frand48();
-		     float r2 = frand48();
+		  {  float r1 = merand48(all->random_state);
+		     float r2 = merand48(all->random_state);
```
JohnLangford: Why this?

vowpalwabbit/parse_example_json.h
```diff
@@ -14,153 +14,740 @@ license as described in the file LICENSE.
 //#define RAPIDJSON_SSE42
```
JohnLangford: Why the shift to a header file here?

vowpalwabbit/parser.cc
```diff
@@ -571,8 +572,18 @@ void enable_sources(vw& all, bool quiet, size_t passes)
...
 
       if (all.vm.count("json"))
-      { all.p->reader = read_features_json;
-        all.p->jsonp = new json_parser;
+      { 
+		  // TODO: change to class with virtual method
+		  if (all.audit)
+		  {
+			  all.p->reader = &read_features_json<true>;
+			  all.p->jsonp = new json_parser<true>;
```
JohnLangford: I see why the header file...

vowpalwabbit/parser.h
```diff
@@ -56,7 +55,7 @@ struct parser
 
   label_parser lp;  // moved from vw
 
-  json_parser* jsonp;
+  void* jsonp;
```
JohnLangford: void* ?

JohnLangford:
I have two questions after review.
(1) Can you respond to @gramhagen about the python?  He had one suggested tweak.
(2) Why put the PRG into a the global datastructure?  (Predictability in a multiple threads scenario?)

Other than this, the patch looks good to me from what I understand.

eisber:
(1)    I think I already addressed his tweaks. Need to recheck.

(2)    Predictability in multi-threading is one issue, but it’s also an issue if you instantiate multiple VW instances in serial as the global PRG state is kept around.

eisber:
json_parser is templated using audit. I could introduce a base class + virtual method…

python/setup.py
eisber: just merge, I'll add it to my todo list

JohnLangford:
astyle created a massive conflict here :-(  I think we need to revisit formatting.

I'll settle for anything that looks reasonable in github.

JohnLangford:
I've merged this after dealing with the merge fails.... Hopefully everything works.

Issue #919, opened 2016-01-25
L2 regularization shrinks the intercept as well.


sharatsc:
This causes some bias to be introduced on large volume datasets since intercept cannot capture the overall bias. As far as I know, there isn't an option to exempt intercept from regularization.

JohnLangford:
We do have an implementation of separating updates on constant and non-constant features in FTRL, which is not pushed back yet since it's an unclean and less generalizable solution. (https://github.com/sinaad/vowpal_wabbit/commit/01cea80c330c31c456357f14b7527dc318af4d36).

And I think a better way to handle this is to move the bias term out of feature/example. It's a global parameter and we do not need to set it up in the feature array.

Thanks,
-Gang

________________________________
发件人: vowpal_wabbit@yahoogroups.com [vowpal_wabbit@yahoogroups.com]
发送时间: 2016年1月28日 8:43
收件人: JohnLangford/vowpal_wabbit; vowpal_wabbit@yahoogroups.com
主题: Re: [vowpal_wabbit] L2 regularization shrinks the intercept as well. (#919)



This is correct. There is a hack though---you could make your own
constant feature which has a large value. Since regularization is not
unit invariant the bias can be made negligible.

I don't see a good way to do something else, computationally. Do you?

Having a test on every feature update which triggers at most once per
example seems... kind of unclean. Having a different update rule for
the constant feature seems worse.

-John

On 01/25/2016 02:53 PM, Sharat Chikkerur wrote:
>
> This causes some bias to be introduced on large volume datasets since
> intercept cannot capture the overall bias. As far as I know, there
> isn't an option to exempt intercept from regularization.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/JohnLangford/vowpal_wabbit/issues/919>.
>


__._,_.___
________________________________
Posted by: John Langford <jl@hunch.net>
________________________________
[...]

PR #1151, opened 2016-12-22
Add option to not regularize the intercept term for BFGS

arirang-pust:
This is a partial fix to #919.


JohnLangford:
This seems reasonable to do, but having a branch on the inner loop is not very pleasant.  Can you instead record weights[constant_feature] externally and then subtract it off?

JohnLangford:
This looks good now, except that it's violating modularity.  Can you put the parsing of arguments into bfgs.cc near the bottom and the storage into the bfgs datastructure?

mikeizbicki:
The current implementation only works for bfgs, but this is something that in theory any optimizer could support.  If someone later adds this functionality to another optimizer, they'd have to undo the "modularization" and moving everything back where it currently is.  So I think it makes sense to leave these parts of the code as-is.

JohnLangford:
Ok, I can live with that. 

I had to change multiple things to make this sensible.  Can you create a test case which confirms this option is working?

PR #1161, opened 2017-01-02
remove legacy code from RunTests + Add -O ... option to add vw args to all tests

arielf:


arielf:
So a few days passed, and this is a good change (mostly because of removal of long dead code) and no errors in Travis, I decided to merge this in. Hope it is ok :)

Issue #924, opened 2016-01-26
bfgs can't use sgd final regressor as initial regressor in pyvw.py


jconwell:
Using pyvw python API I can't use a sgd regressor as an initial regressor to bfgs.

I've got a full example below that demonstrates this issue (below).  In STEP 1, it creates a regressor with a couple demo examples.  At the end of the step, it calls vw.finish() to persist the final regressor.

STEP 2 creates a new vw instance that takes as initial regressor the regressor saved from STEP 1.  This works great and gives me predictions.

STEP 3 tries to create a new vw instance that uses bfgs, and takes as initial regressor the regressor saved from STEP 1.  At this point the python process does with an exit code of 139.

If I try to run step 3 from the command line, using the same initial regressor:

```
vw --data <full_path_to_data_file>/tfidf.training.vw --bfgs --hash all --mem 18 --learning_rate 1.0 --power_t 1.0 --l2 0.005 --loss_function squared --l1 0.01 --initial_regressor <full_path_to_model_file>/vw.sgd.model --final_regressor <full_path_to_model_file>/vw.bfgs.model --bit_precision 18 --passes 5 --cache
```

VW outputs: "vw (./io_buf.h:231): file is not a model file"

If I run both the SGD step and the BFGS step from the command line, everything works great.

```
vw --data <full_path_to_data_file>/tfidf.training.vw --hash all --learning_rate 0.5 --quiet --power_t 1.0 --loss_function squared --final_regressor <full_path_to_model_file>/vw.sgd.model --bit_precision 18

vw --data <full_path_to_data_file>/tfidf.training.vw --bfgs --hash all --mem 18 --learning_rate 1.0 --power_t 1.0 --l2 0.005 --loss_function squared --l1 0.01 --initial_regressor <full_path_to_model_file>/vw.sgd.model --final_regressor <full_path_to_model_file>/vw.bfgs.model --bit_precision 18 --passes 5 --cache -p predictions.txt
```

This works great, as expected. 

Here is the python code that reproduces the issue I'm seeing

```
import pyvw

# STEP 1
# create vw instance
vw = pyvw.vw("--hash all --learning_rate 0.5 --quiet --power_t 1.0 --loss_function squared --final_regressor <full_path_to_model_file>/vw.sgd.model --bit_precision 18")

# learn a couple examples
ex = vw.example('1 |a two features |b more features here')
ex.set_label_string('1.0')
vw.learn(ex)
ex.finish()
[...]

PR #1163, opened 2017-01-07
Python wrapper improvements for handling external data files and model i/o

gramhagen:
For #924 

Adding ability to processes external data file with python wrapper, and save model in vw format, and support bfgs in sklearn wrapper


JohnLangford:
Merged, thanks :-)

PR #1168, opened 2017-01-18
Exposing tracing via API

eisber:
+ minor refactoring/cleanup for sparse/dense parameters


JohnLangford:
What is the status of the appveyor fail?

JohnLangford:
The void* and the not working on windows seem like the primary concerns here.

eisber:
void* for the json parser? Looking into the Windows build.

Markus

JohnLangford:
This is still failing in the appveyor build.  I'm going to merge on the theory that it is better than where we are right now, but fixing the appveyor build seems important.

Issue #1105, opened 2016-09-15
--boosting doesn't work with --link and more


arielf:
I tried to use `--boosting N` and noticed that model accuracy is worse than without it, which is unexpected.

For example, consider this pretty minimal train-set:

```
-1 1|  a b c
1 2|  c d f
```

Put it in `train.vw` and train without boosting:

```
$ vw --quiet --loss_function logistic --link glf1 -f model -d train.vw
```

now that we have a model, predict on the same data:

```
$ vw --quiet -i model -d train.vw  -p /dev/stdout
-0.231024 1
0.059729 2
```

Which is a reasonable prediction (1st example negative, 2nd example positive)

Now try to do the same with `--boosting 2`:

```
$ vw --loss_function logistic --link glf1 --quiet -f model -d train.vw --boosting 2
Number of weak learners = 2
Gamma = 0.1

$ vw --loss_function logistic --quiet -i model -d train.vw -p /dev/stdout
Number of weak learners = 2
Gamma = 0.1
-1 1
-1 2
```

I see a few problems here:
1) `--boosting` breaks --link glf1 (which is stored in the model:  it outputs binary {-1, 1} instead of probabilities
2) `--boosting` ignores the `--quiet` flag and remains chatty
3) `--boosting` produces a result that is worse than without it (both examples predicted negative)

JohnLangford:
@haipengl any thoughts?  

(2) Is surely easily fixed.  

W.r.t. (3), it's very difficult to guarantee monotonic performance in general for ML algorithms, although I agree that it is desirable.

haipengl:
The boosting algorithms implemented are the online version. I guess you would see improvement if you look at progressive error and try larger data sets...

JohnLangford:
@beygel @haipengl can you handle (1) and (2) maybe?

JohnLangford:
Reviewing this, (2) is fixed.  

I'm not sure what to do for (1).  The problem is that boosting: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parse_args.cc#L1122 is applied after link functions: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parse_args.cc#L1114
We could move boosting earlier but boosting uses the raw score rather than some threshold on the score.  As a consequence, maybe we should just forbid glf1 with boosting?

For (3), there is no fix for this---it's just a different algorithm with different performance characteristics.

[...]

PR #1171, opened 2017-01-19
Make --boosting obey --quiet

bartleusink:
Hi!

This PR should fix problem (2) of #1105

Best,

Bart


arielf:
Thanks!
The AppVeyor failure can be ignored (unrelated issues).
I would have merged this (looks good) but the merge button is greyed-out because of the above, so will leave this to John :)

JohnLangford:
This looks good.  Thanks.

PR #1172, opened 2017-01-21
Marginal

alekh:
I am using a modified potential that Haipeng suggested for scale issues. This computes \Phi(R,C) as \exp(R_+^2/3C), with C being sum of squared regrets instead of absolute values. The weights are set to the continuous rather than discrete derivative of \Phi wrt R. There are a couple of places where we have some design choices, and I have left comments accordingly.


JohnLangford:
It looks like there is a massive chmod over all files.  Can you undo?  We don't want a diff affecting 787 files.

alekh:
Ugh. This was painful. Some quirk of the Ubuntu subsystem on Windows.

Fixed now.

Alekh

vowpalwabbit/marginal.cc
```diff
@@ -43,11 +77,31 @@ void predict_or_learn(data& sm, LEARNER::base_learner& base, example& ec)
...
 		f.space_names.push_back(sm.temp[n].space_names[2*(f.size()-1)]);
+
+	      if(sm.compete) { // compute the prediction from the marginals using the weights
+		float weight = sm.expert_state[key].first.weight;
+		average_pred += weight*marginal_pred;
+		net_weight += weight;
+		net_feature_weight += sm.expert_state[key].second.weight;
+		if(is_learn)
+		  alg_loss += weight*(marginal_pred - label)*(marginal_pred - label);
+	      }
```
JohnLangford: Why assume squared loss here?

vowpalwabbit/marginal.cc
```diff
@@ -43,11 +77,31 @@ void predict_or_learn(data& sm, LEARNER::base_learner& base, example& ec)
...
 	      uint64_t key = second_index + ec.ft_offset;
-	      if (sm.marginals.find(key) == sm.marginals.end())//need to initialize things.
+	      if (sm.marginals.find(key) == sm.marginals.end()) {//need to initialize things.
 		sm.marginals.insert(make_pair(key,make_pair(sm.initial_numerator, sm.initial_denominator)));
-	      f.push_back((float)(sm.marginals[key].first / sm.marginals[key].second), first_index);
+		if(sm.compete) {
+		  expert e;
+		  e.regret = 0;
+		  e.abs_regret = 0;
+		  e.weight = 1.;
```
JohnLangford: It would be cleaner to use:
expert e = {0,0,1.};

vowpalwabbit/marginal.cc
```diff
@@ -56,6 +110,29 @@ void predict_or_learn(data& sm, LEARNER::base_learner& base, example& ec)
...
+  if(sm.compete) { //add in the feature-based expert and normalize,
+		   //assign to example
+    //net_feature_weight = 0.; useful in case we want to try pure marginals
+    if(net_weight + net_feature_weight > 0.) 
+      average_pred += net_feature_weight * feature_pred;      
+    else {
+      net_feature_weight = 1.;
+      average_pred = feature_pred;
+    }
+    average_pred /= (net_weight + net_feature_weight);
```
JohnLangford: Instead of dividing, can you precompute 1. / (net_weight + net_feature_weight) then multiply?  multiplication is typically much faster than division.

vowpalwabbit/marginal.cc
```diff
@@ -67,8 +144,22 @@ void predict_or_learn(data& sm, LEARNER::base_learner& base, example& ec)
           uint64_t second_index = j.index() & mask;
           uint64_t key = second_index + ec.ft_offset;
           marginal& m = sm.marginals[key];
+
+	  if(sm.compete) { //now update weights, before updating marginals
+	    expert_pair& e = sm.expert_state[key];
+	    float regret1 = alg_loss - (m.first/m.second - label)*(m.first/m.second - label);
```
JohnLangford: The comment about loss seems to apply here as well.  Why adopt squared loss only?

vowpalwabbit/marginal.cc
```diff
@@ -129,22 +271,28 @@ LEARNER::base_learner* marginal_setup(vw& all)
...
+    d.compete = true;
   d.all = &all;
   string s = (string)all.vm["marginal"].as<string>();
 
   for (size_t u = 0; u < 256; u++)
     if (s.find((char)u) != string::npos)
       d.id_features[u] = true;
   new(&d.marginals)unordered_map<uint64_t,marginal>();
+  if(d.compete)
+    new(&d.expert_state)unordered_map<uint64_t,expert_pair>();
```
JohnLangford: Have you run under valgrind to check for leaks?

alekh:
The main reason for picking squared loss was that the marginals are
computing an average, which is the right thing to do if we want to use a
squared loss. If not, then we would want to adapt marginals to do something
else as well in order to compute the best scalar prediction.

Alekh

alekh:
And 1/x is also much faster than a/x for some other quantity a?

Alekh

alekh:
No, I can do that later. I will have to be on my linux machine for running
valgrind, since it does not appear to work in my Windows Ubuntu shell.

There is only one data structure created though, which is destroyed at the
end. So pretty confident that I should not be causing any memory leaks
which did not already exist :)

Alekh

JohnLangford:
Status on tweaking the loss function?

It's not the 1/x is faster than a/x, but rather that 1/x is faster than a/x done many times.

alekh:
Haven't had a chance, sorry.

Alekh

alekh:
I will look into this now. Just so you know, I actually prefer the more
explicit declaration, since in the "cleaner" version I have to always find
the struct definition to figure out what is being assigned which value.

Alekh

alekh:
Okay, I have addressed all the issues, and added a test case.

Alekh

JohnLangford:
@alekh Can you update this patch?  It's in a confusing state.  (We need to handle patches faster...)

alekh:
Later today, or Friday. Yeah, this one has been pending for a while. Might
be some fun bringing it up to the master branch :)

Alekh

alekh:
Done.

Alekh

JohnLangford:
This is merged now, although I had to tweak a few things.

PR #1180, opened 2017-01-27
autogen.sh overwrites multiple Makefiles + how to restore

arielf:
1) Fix: autogen.sh overwrites multiple Makefile's (not just the top one).
2) Document the clean way to restore original Makefiles.


JohnLangford:
Merged, thanks.

PR #1186, opened 2017-01-30
update to python packaging to make it easier to automate process

gramhagen:
updating manifest to match new windows build files and package instructions for using twine to upload python packages


JohnLangford:
Merged, thanks :-)

PR #1189, opened 2017-01-30
Enable the C wrapper to extract model to memory and initialize model from memory

jrmarkle:
This workaround allows library users to load and store the model directly without using the file system.


JohnLangford:
@eisber I thought you had looked into RAM-based model storage in the past? What's the status?

eisber:
You’ll have to implement something like https://github.com/JohnLangford/vowpal_wabbit/blob/master/cs/cli/clr_io_memory.h

save_load_header takes in an io_buf: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parse_regressor.h#L19


-Markus

jrmarkle:
Yes, the PR includes `memory_io_buf` for that purpose.

JohnLangford:
I'm hazy on an important detail here.  @eisber is this functionality already there?  If not, why is there an interface?  
(The confidence stuff seems desirable.)

eisber:
@JohnLangford it's not there yet. It's only in the C# side. I think this is good.

JohnLangford:
Merged in, thanks.

PR #1190, opened 2017-01-31
Fix issue where invert_hash doesn't print parameters due to save_resume and passes > 0

### LINKED ISSUE #1185: --invert_hash doesn't work with --save_resume when passes != 0
The condition at [this line](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/gd.cc#L228) i.e.

```C++
if (dat.all.current_pass == 0 && dat.all.hash_inv)
```

doesn't allow the human readable feature names and their weights to be printed when using `--save_resume` if the `passes != 0`

As --invert_hash isn't compatible with --passes flag, it is recommended to use the [following way](http://stackoverflow.com/questions/24437152/vowpal-wabbit-inverted-hash-option-produces-empty-output-but-why) to produce the inverted_hash model file:

```Bash
vw -d mytrain.data -c --passes 4 -f model.binary
vw -d mytrain.data -t -i model.binary --invert_hash model.humanreadable
```

However with the additional save_resume flag, the binary model file preserves the number of passes and thus the condition mentioned fails.

This issue is similar to this one. #337

apoorvreddy:
This commit fixes #1185 and allows the invert_hash to be printed in testing mode when passes > 0 and save_resume option has been used.

### DIFF
--- a/vowpalwabbit/gd.cc
+++ b/vowpalwabbit/gd.cc
@@ -225,7 +225,7 @@ inline void audit_feature(audit_results& dat, const float ft_weight, const uint6
       dat.results.push_back(sv);
     }
   
-  if (dat.all.current_pass == 0 && dat.all.hash_inv)
+  if ((dat.all.current_pass == 0 || dat.all.training == false) && dat.all.hash_inv)
     { //for invert_hash
       
       if (dat.offset != 0)


apoorvreddy:
@JohnLangford can we merge this ?

JohnLangford:
The patch looks fine to merge.   Can you add a test case to test/RunTests which exhibits this?

PR #1199, opened 2017-02-14
workaround GCC 6.x compilation issue

### LINKED ISSUE #1183: parse_regressor.cc can't be compiled with g++
It's been at least 6 month I haven't look into VW code and it changed a lot since when. I was working with my old fork, but today decided to compare it with master branch. And faced with the fact that I can't build master branch with g++ which is default. With libtool or without it.
I'm getting following errors:
```
libtool: compile:  g++ -DHAVE_CONFIG_H -I. -I/usr/include -I/include -Wall -Wno-unused-local-typedefs -pedantic -I ../rapidjson/include -O3 -fomit-frame-pointer -DNDEBUG -MT parse_regressor.lo -MD -MP -MF .deps/parse_regressor.Tpo -c parse_regressor.cc  -fPIC -DPIC -o .libs/parse_regressor.o
parse_regressor.cc: In function ‘void initialize_regressor(vw&, T&)’:
parse_regressor.cc:55:22: error: parse error in template argument list
     weights.template set_default<float,set_initial<T> >(all.initial_weight);
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
parse_regressor.cc:55:22: error: expected identifier
parse_regressor.cc:57:22: error: parse error in template argument list
     weights.template set_default<random_positive<T> >();
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
parse_regressor.cc:57:22: error: expected identifier
parse_regressor.cc:59:22: error: parse error in template argument list
     weights.template set_default<random_weights<T> >();
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
parse_regressor.cc:59:22: error: expected identifier
```
It seems the code is too complicated for g++. My knowledge of templates was far not enough to bypass this. So I switched to clang++ and successfully build the project and passed `make test`.

My compilers are:

```
gcc version 6.2.0 20161005 (Ubuntu 6.2.0-5ubuntu12)
clang version 3.8.1-12ubuntu1 (tags/RELEASE_381/final)
```
Kubuntu 16.10 , 4.8.0-37-generic GNU/Linux
Could someone reproduce this problem? It's most probably g++ version dependent .

trufanov-nok:
resolves #1183 
The problem with GCC 6.x is described here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79508
It's a regression. In short - GCC since 5.4.1 understands only template classes in nested template list arguments while VW tries to push function pointer there. So I've just wrapped all functions in classes and make them static.

### DIFF
--- a/vowpalwabbit/array_parameters.h
+++ b/vowpalwabbit/array_parameters.h
@@ -137,18 +137,18 @@ private:
 
 	inline weight& strided_index(size_t index){ return operator[](index << _stride_shift);}
 
-	template<class R, void(*T)(iterator&, R&)> void set_default(R& info)
+	template<class R, class T> void set_default(R& info)
 	{
 	  iterator iter = begin();
 	  for (size_t i = 0; iter != end(); ++iter, i += stride())
-	    T(iter, info);
+	    T::func(iter, info);
 	}
 
-	template<void(*T)(iterator&)> void set_default()
+	template<class T> void set_default()
 	{ 
 	  iterator iter = begin();
 	  for (size_t i = 0; iter != end(); ++iter, i += stride())
-	    T(iter);
+	    T::func(iter);
 	}
 
 	void set_zero(size_t offset)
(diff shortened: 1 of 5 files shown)


JohnLangford:
I'm merging this for now.  When gcc gets fixed, maybe we can pull it back out.

JohnLangford:
(Thanks for the thorough investigation.)

PR #1201, opened 2017-02-15
fixing typo in python makefile 

gramhagen:
needed to get correct python version includes / ldflags


JohnLangford:
Merged, thanks!

PR #1212, opened 2017-03-12
fixing a small bug

trufanov-nok:
My compiler is barking:
```
parse_args.cc: In function ‘const char* VW::are_features_compatible(vw&, vw&)’:
parse_args.cc:450:24: warning: self-comparison always evaluates to false [-Wtautological-compare]
   if (vw1.permutations != vw1.permutations)
```


JohnLangford:
Merged, thanks!

PR #1219, opened 2017-04-01
In daemon mode, do not run in the background

seifert:
In daemon mode, the service is forked an run in the background. However, it is not possible to run vw in foreground. It is useful if I want to check the service by some supervisor - systemd, supervisord, daemontools, ... There is simple patch, which adds --foreground command line argument, which forces keeping the process in the foreground.


arielf:
The enhancement looks good to me.  Thanks!
Note that coverage decreased though - could you add a test of the new option to `RunTests` too?

seifert:
I extended daemon-test.sh and added new test into RunTests. It checks PID of the daemon - if it is run in the background, PID of the runnig process must be different than PID of the executed process. In the foreground PID must be the same.

However, test which is not related to enhancement is broken.

arielf:
Thanks for the addition!

You're right.

There are two unrelated travis failures:
1) make test: Test 4, heisenbug (we've seen it before, some rarely occurring diff due to race condition)
2) make gcov: Test 19 (3-parity) fails with SEGV (exit 139 = 128 + 11)

Both are not reproducible.  The travis env sometimes seems a bit unreliable.
You may try to trigger a new build by a bogus push (e.g. white space or comment change).
Sometimes these new builds succeed.  Sorry about the troubles.

test/daemon-test.sh
```diff
@@ -46,25 +60,35 @@ fi
...
-DaemonCmd="$VW -t -i $MODEL --daemon --num_children 1 --quiet --port $PORT"
+DaemonCmd="$VW -t -i $MODEL --daemon $Foreground --num_children 1 --quiet --port $PORT"
 # libtool may wrap vw with '.libs/lt-vw' so we need to be flexible
 # on the exact process pattern we try to kill.
 DaemonPat=`echo $DaemonCmd | sed 's/^[^ ]*vw /.*vw /'`
 
 stop_daemon() {
     # Make sure we are not running. May ignore 'error' that we're not
-    $PKILL -9 -f "$DaemonPat" 2>&1 | grep -q 'no process found'
+    $PKILL -15 -f "$DaemonPat" 2>&1 | grep -q 'no process found'
```
arielf: Yes, SIGTERM is cleaner/more elegant but unfortunately experience showed that SIGKILL is a more sure thing in some test environments.  It this test there's no concern of leaving anything corrupted so SIGKILL is ok.

arielf:
@seifert - thanks for the contribution!

matanster:
This option is still absent in the docs I believe
https://github.com/JohnLangford/vowpal_wabbit/wiki/Command-line-arguments

PR #1223, opened 2017-04-07
do not use deprecated sklearn module

iezepov:
sklearn.cross_validation module was deprecated in version 0.18 in favor of the model_selection module into which all the refactored classes and functions are moved. This module will be removed in 0.20

### DIFF
--- a/python/README.rst
+++ b/python/README.rst
@@ -60,7 +60,7 @@ Or you can use the included scikit-learn interface like this:
 
     >>> import numpy as np
     >>> from sklearn import datasets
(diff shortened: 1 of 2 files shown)


iezepov:
This is a simple documentation fix, have no idea why tests failed.

Scott-Graham-Bose:
i think the failures are unrelated, but i would bump the scikit-learn version to >= 0.18 in the python/requirements.txt to enforce compatibility with newer versions

JohnLangford:
Thanks!

Yeah, unrelated test fail.  Something is going on with test 140 that I can't reproduce locally.

PR #1226, opened 2017-04-10
Add prerequisites: zlib-devel (or zlib1g-dev)

arielf:


JohnLangford:
Something is wrong with test 140, but I can't seem to reproduce on my local box, even with valgrind.  In the meantime, thanks.

PR #1227, opened 2017-04-16
Add saveModel method to Java interface

stephendavidmarsh:
This adds a `saveModel` method to the JNI interface, allowing VW instances used through the JNI interface to save their state to disk.


JohnLangford:
@jmorra can you take a look?

java/src/main/c++/vowpalWabbit_learner_VWLearners.cc
```diff
@@ -26,6 +26,20 @@ JNIEXPORT void JNICALL Java_vowpalWabbit_learner_VWLearners_closeInstance(JNIEnv
...
 }
 
+JNIEXPORT void JNICALL Java_vowpalWabbit_learner_VWLearners_saveModel(JNIEnv *env, jclass obj, jlong vwPtr, jstring filename)
+{ try
+  {
+    const char* utf_string = env->GetStringUTFChars(filename, NULL);
+    std::string filenameCpp(utf_string);
+    VW::save_predictor(*(vw*)vwPtr, filenameCpp);
+    env->ReleaseStringUTFChars(filename, utf_string);
+    env->DeleteLocalRef(filename);
```
jmorra: Can we release the memory outside of the try block (preferably in a finally block).

java/src/main/java/vowpalWabbit/learner/VWLearner.java
```diff
@@ -10,4 +10,6 @@
 
     // This tells the implementations that an IOException cannot be thrown.
     @Override void close();
+
+    void saveModel(String filename);
```
jmorra: I think the save interface should accept a `java.io.File` to show that it's a local File being saved.

java/src/main/java/vowpalWabbit/learner/VWLearners.java
```diff
@@ -104,4 +104,6 @@ private static void loadNativeLibrary() {
 
     // Closing needs to be done here when initialization fails and by VWBase
     static native void closeInstance(long nativePointer);
+
+    static native void saveModel(long nativePointer, String filename);
```
jmorra: I think the native method should be private the one exposed should be `java.io.File.`

java/src/main/c++/vowpalWabbit_learner_VWLearners.cc
stephendavidmarsh: C++ doesn't have finally blocks, unfortunately. I believe the `std::string` copies the characters inside of itself, so we should be able to do the memory releases after it is constructed and before `save_predictor` is called. This will prevent leaks from errors while saving. The issue with releasing inside of the exception-converting try block is done throughout the library. I took the pattern used here from `read_example` in jni_base_learner.cc, which isn't doing anything to prevent leaks.

jmorra: Good point.  It's been a while since I've looked at this code.

java/src/main/java/vowpalWabbit/learner/VWLearners.java
stephendavidmarsh: This needs to be callable from `VWBase`, so it can't be private. `closeInstance` above is the same. They are package-private though, so users of the library can't call them directly.

java/src/main/java/vowpalWabbit/learner/VWLearner.java
stephendavidmarsh: Done.

java/src/main/c++/vowpalWabbit_learner_VWLearners.cc
stephendavidmarsh: I made the change to release the memory before calling `save_predictor`.

java/src/main/c++/vowpalWabbit_learner_VWLearners.cc
```diff
@@ -26,6 +26,20 @@ JNIEXPORT void JNICALL Java_vowpalWabbit_learner_VWLearners_closeInstance(JNIEnv
   }
 }
 
+JNIEXPORT void JNICALL Java_vowpalWabbit_learner_VWLearners_saveModel(JNIEnv *env, jclass obj, jlong vwPtr, jstring filename)
+{ try
+  {
+    const char* utf_string = env->GetStringUTFChars(filename, NULL);
```
jmorra: This looks better.  The way it's written now the JNI lines (everything other than `save_predictor`) doesn't even need to be in the `try` block.

java/src/main/java/vowpalWabbit/learner/VWBase.java
```diff
@@ -63,6 +64,23 @@ final boolean isOpen() {
...
     }
 
+    /**
+     * Save the model in the VW instance.
+     */
+    public void saveModel(File filename) {
+        lock.lock();
+        try {
+            if (isOpen()) {
+                VWLearners.saveModel(nativePointer, filename.getPath());
```
jmorra: This should be `file.getAbsolutePath` in my opinion to handle relative paths.

java/src/main/c++/vowpalWabbit_learner_VWLearners.cc
jmorra: Good point.

java/src/test/java/vowpalWabbit/learner/VWLearnersTest.java
```diff
@@ -67,6 +68,15 @@ public void testBadModel() {
     }
 
     @Test
+    public void testSaveModel() {
+        VWLearner vw = VWLearners.create("--quiet");
+        File file = new File("/tmp/saved_test_model");
```
jmorra: Just noticed this will leave a file on the user's file system after the test is run.  Tests should leave the machine in the same state as before they were run.  Consider using a [TemporaryFolder](http://blog.codeleak.pl/2015/01/testing-with-files-and-directories-in.html) rule to keep the state of the machine the same.

java/src/main/java/vowpalWabbit/learner/VWBase.java
stephendavidmarsh: It can handle relative paths in its current form actually - the underlying C++ interprets them correctly.

java/src/test/java/vowpalWabbit/learner/VWLearnersTest.java
stephendavidmarsh: I didn't know about TemporaryFolder, it definitely is convenient. I've made the change.

stephendavidmarsh:
Anything left on my end for this?

JohnLangford:
There is a merge conflict.  Can you update to the current branch please?

stephendavidmarsh:
I've merged in the latest master and resolved the merge conflict.

JohnLangford:
@jmorra do you know what this error is exactly?
https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/243804143#L2064

It's about this line: https://github.com/JohnLangford/vowpal_wabbit/blob/master/java/src/main/java/vowpalWabbit/learner/VWMulticlassLearner.java#L6

and is replicated for each of the different learner types.  The master branch is currently working and I don't see what in the patch could cause this.  Any ideas?

JohnLangford:
What is this error?
https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/251685719#L2064
[ERROR] /home/travis/build/JohnLangford/vowpal_wabbit/java/src/main/java/vowpalWabbit/learner/VWMulticlassLearner.java:[6,14] vowpalWabbit.learner.VWMulticlassLearner is not abstract and does not override abstract method getNativePointer() in vowpalWabbit.learner.VWLearner

java/src/main/java/vowpalWabbit/learner/VWLearner.java
```diff
@@ -10,4 +11,9 @@
 
     // This tells the implementations that an IOException cannot be thrown.
     @Override void close();
+
+    void saveModel(File filename);
+    
+    // Get the JNI pointer for the learner
+    public long getNativePointer();
```
jon-morra-zefr: Why is this needed?  In response to @JohnLangford's question.  The reason the build is failing is that the function `long getNativePointer()` is defined in this interface (which every typed learner uses) and then not implemented anywhere.  If this is removed it should compile fine.

ryan-deak-zefr: @jon-morra-zefr @JohnLangford @stephendavidmarsh isn't exposing the `getNativePointer()` method in the interface a good bad idea.  Seems like there is little practical use for this and it will likely only be used in nefarious ways.

jon-morra-zefr: I agree

JohnLangford:
I removed the native pointer and merged.  Thanks :-)

java/src/main/java/vowpalWabbit/learner/VWLearner.java
stephendavidmarsh: That line got brought in when I merged master in to resolve the previous merge conflict. I have no idea why it was added. And of course, it then got removed from master, creating yet another merge conflict.

JohnLangford:
This may be due to a java patch that was merged then pulled out because it
was causing test failures.

-John

PR #1228, opened 2017-04-20
fix sizeof bug

Webern:
A file written with vowpal_wabbit on a 64-bit machine cannot be opened by a 32-bit version.  This is because sizeof(size_t) is being used as a stride length during file operations.  By changing the size_t to uint64_t I believe files formats will no longer be different between architectures.


Webern:
I guess the change doesn't pass tests, but this is the change that we needed to make in order to use vowpal_wabbit in our project.  Our situation is that we are running on 64-bit and 32-bit machines (iPads) using the same training file (which was written by a 64-bit machine).  When reading that file on a 32-bit machine vowpal_wabbit would throw `THROW("Model content is corrupted, weight vector index " << i << " must be less than total vector length " << length); (gd.cc)` because the stride length had gotten off and a junky number had been read into i.

JohnLangford:
That's good detective work.

The error seems unrelated but we must be careful here to preserve backwards compatibility.  That means a bit more work testing the model version and reading one way or the other.

arielf:
Thanks @Webern !
@JohnLangford : I just verified that this change is safe in my 64-bit env (Ubuntu 16.04).
All tests pass, no issue.
:+1:

JohnLangford:
Merged finally, with a backwards compatibility check :-)  Thank you for spotting this.

Issue #1208, opened 2017-03-06
Failed to install vowpal_wabbit on Python 3


acepor:
Hi,

I followed the installation instruction to install vowpal_wabbit on MacOS 10.12.3 with Python 3.6.0/3.5.2, and neither succeeded. However, when I installed it with Python 2.7.13, everything was smooth.

Can anyone offer any help? Thanks you so much!

> Collecting vowpalwabbit

>   Using cached vowpalwabbit-8.3.2.tar.gz
> Installing collected packages: vowpalwabbit
>   Running setup.py install for vowpalwabbit ... error
>     Complete output from command /Library/Frameworks/Python.framework/Versions/3.5/bin/python3.5 -u -c "import setuptools, tokenize;__file__='/private/var/folders/0y/9l17xzzs43j4syrjschyyssw0000gn/T/pip-build-6mumnwv2/vowpalwabbit/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /var/folders/0y/9l17xzzs43j4syrjschyyssw0000gn/T/pip-tmbcjegf-record/install-record.txt --single-version-externally-managed --compile:
>     running install
>     running build
>     running build_py
>     creating build
>     creating build/lib.macosx-10.6-intel-3.5
>     creating build/lib.macosx-10.6-intel-3.5/vowpalwabbit
>     copying vowpalwabbit/__init__.py -> build/lib.macosx-10.6-intel-3.5/vowpalwabbit
>     copying vowpalwabbit/pyvw.py -> build/lib.macosx-10.6-intel-3.5/vowpalwabbit
>     copying vowpalwabbit/sklearn_vw.py -> build/lib.macosx-10.6-intel-3.5/vowpalwabbit
>     running egg_info
>     writing dependency_links to vowpalwabbit.egg-info/dependency_links.txt
>     writing vowpalwabbit.egg-info/PKG-INFO
>     writing top-level names to vowpalwabbit.egg-info/top_level.txt
>     warning: manifest_maker: standard file '-c' not found
> 
>     reading manifest file 'vowpalwabbit.egg-info/SOURCES.txt'
>     reading manifest template 'MANIFEST.in'
>     warning: no files found matching '*.dll' under directory 'bin'
>     warning: no previously-included files matching '*.o' found anywhere in distribution
[...]

PR #1230, opened 2017-04-25
python 3: updated python/setup.py to detect the installed boost_python library

florinb:
The changes, python/setup.py and python/Makefile, are related to #1208.

I used them to have the python wrapper build and install on Ubuntu 14.04 and 16.04 with the latest Anaconda 4.31 (the Python 3.6 version 64-bit installer, with ~anaconda3/bin added to PATH) as  mentioned earlier in the comments for #1208.


JohnLangford:
@gramhagen have you had a chance to take a look at this?

Scott-Graham-Bose:
yeah, we should pull this in, there's still more work to be done here to support mac, but this is moving in the right direction

JohnLangford:
Merged, thanks :-)

PR #1238, opened 2017-05-04
ignore linear

pmineiro:
Support for --ignore_linear


JohnLangford:
Can you add a test case?

pmineiro:
Ok I added a test case.

p

PR #1239, opened 2017-05-05
Add --cubic regression test

arielf:
+ Left over commit I had in my master: small addition to the `vw-convergence` plotting utility: `-s` option to change point size.


arielf:
All failures unrelated to my changes.

Test 140, marginal is a new heisenbug, exitcode 139 is SEGV.
I was able to reproduce only once in my env.

appveyor and java/maven build failures are also unrelated.

JohnLangford:
yeah, the build fails need to be dealt with, but I'm out of time at the moment.

JohnLangford:
There are conflicts though.  Can you fix that?

arielf:
:) Conflict is because both Paul and I added a test to `RunTests` and he was merged first.
Let me remerge...

arielf:
Should be mergable now  (failures still unrelated)

PR #1241, opened 2017-05-11
Fix --audit_regressor for save_resume models

trufanov-nok:
There is a mistake made in `audit_regressor.cc::regressor_values` (i think this happens during sparse/dense weights change). This iterator should count non-zero weights to let audit_regressor stop its work when all values are found and track the progress but currently it counts all non-zero **floats** in regressor (weights + stuff inside their stride shift). It wasn't caught by RunTests as stuff inside stride is initialized with 0s for regular model file. But not for model that saved with --save_resume. For --save_resume model it counts up to 3 time more weights in regressor than it really has (depends on gd options). This doesn't let `--audit_regressor` stop its work till the end of data file (as it never can find all weights from it) and wrongly prints a warning for user. Warning like:  
`Note: for some reason audit couldn't find all regressor values in dataset (4287 of 12861 found).`

I've added one more test case to check `--audit_regressor` result for model that was stored with `--save_resume`


JohnLangford:
Thank you :-)

Issue #1233, opened 2017-04-27
ADF Hello World example


tilayealemu:
Featurized actions [1] can be very useful for dynamic number of actions but documentation is a bit unclear. Tried to make a very simple hello-world example to see it work.

Let's say we have two features, `orange` and `apple`. I designed an example where action 0 would be the best action when `orange` is observed, and action 1 when `apple` is observed.

action | feature | cost
------ | ------- | ----
|0|orange|low|
|0|apple|high|
|1|orange|high|
|1|apple|low|

Here is the vw formatted training data. I encoded `low cost` as 0 and `high cost` as 1.

```
0:0:0.5 | orange
| orange

0:1:0.5 | apple
| apple

| orange
0:1:0.5 | orange

| apple
0:0:0.5 | apple
```

Test data for when `apple` feature is observed,

```
0:0:0 | apple
| apple
```

Test for `orange`,

```
0:0:0 | orange
| orange
```

I trained with `--cb_adf` option. Java source code attached as well.

Problem is, I get action 0 as the predicted action in both test cases. Tried giving it more training data by duplicating the examples and result is the same.

Why doesn't vw predict action 1 for `orange`? All pointers appreciated!

[1] https://github.com/JohnLangford/vowpal_wabbit/wiki/Contextual-Bandit-algorithms

[AdfHelloWorld.zip](https://github.com/JohnLangford/vowpal_wabbit/files/962386/AdfHelloWorld.zip)

JohnLangford:
You seem to have 2-action examples (i.e. two lines) and the features of
these actions are the same.   Hence, the internal predictions for each
examples will be the same, and then ties break towards the first action.

-John

tilayealemu:
Correct. The features are the same. In my scenario, 1) all actions have the same features and 2) number of actions can change through time. Is it achievable?

Imagine for example a learner system that guesses the type of fruit in a picture. All actions will see the same features. And at times, we may want to add new actions. So if the system sees something that looks like an orange, it would give us actions 0,1. And later on if we add a third action for mandarin, it will explore the new action for some time and then correctly predict 0,2,1 when it is presented with an orange-like feature.

JohnLangford:
Actions must be distinct at the feature level for VW learning algorithms to
do something.   Stated another way, there is no implicit identifier for an
action (maybe there should be?).

-John

tilayealemu:
[...]

Issue #1118, opened 2016-10-20
Retrieving raw predictions in c# wrapper


Wirack:
I am currently working with oaa, and can expose scores with use of logistic loss.
However, when I use another loss function I am not able to expose the scores.

I found that --raw_predictions can be used to extract raw predictions in vowpal wabbit.

How can I extract raw predictions in the c# wrapper, apart from outputting to a file?

And if that is not available currently, where should I modify the wrapper (and maybe vw?) to be able to expose those raw predictions?

eisber:
You should modify the prediction factory, but I suppose we need some advice from John on how we should report those back through the example data structure. I don’t think it’s include yet.

Wirack:
I have begun implementing that, but stumbled into the same problem with raw predictions not being exposed in the example data structure.

I am wondering if oaa.cc should be adjusted to expose this in the example data structure, or if it is possible to implement something more general?

eisber:
Without taking a too close look, we could introduce a commonly used field used by each instance.

Markus

Wirack:
Yes, I think that is the right approach, and is also discussed [here](https://github.com/JohnLangford/vowpal_wabbit/issues/918).
I also think the '--scores' flag would be  a good way to go.

mttdbrd:
I wrote a method in the Java wrapper which shows how to extract raw predictions without writing to a file. You can see it here: [https://github.com/JohnLangford/vowpal_wabbit/compare/master...getservice:master](url)

Wirack:
Thanks, will take a look when I have some free time.

JohnLangford:
You can now pass --scores to oaa.cc and it will expose all predictions via pred.scalars.

Wirack:
Thanks, that's a nice clean solution, 
And since VowpalWabbitScalarsPredictionFactory already exposes pred.scalars I would say this issue is resolved.

Wirack:
It seems VowpalWabbitScalarsPredictionFactory does not allow for multiclass scores, but VowpalWabbitPredictionType.MultiClassProbabilities can be used instead.

marcocaccin:
I'm using this feature with `hinge` loss and I'm noticing that all scores are capped in the [-1, 1] range. Is this intended, and what's the reason?
(I am intuitively assuming that a score with hinge loss is the margin of a linear SVM, correct me if I'm wrong!)

Wirack:
[...]

Issue #1082, opened 2016-08-18
How to generate prediction files with probabilities instead of chosen label for cost sensitive multi class regression


dhruvghulati:
I am using: https://github.com/JohnLangford/vowpal_wabbit/wiki/Cost-Sensitive-One-Against-All-%28csoaa%29-multi-class-example

With training of (example line):

```
1:1 2:1 3:0 4:1 5:1 6:1 7:1 8:1 9:1 10:1 11:1 12:1 13:1 14:1 15:1 16:1 0| cat mat sat
```

And test of (example line):

```
0| the mat is red
```

Currently using:

```
vw --csoaa --loss_function=logistic 16 open_cost_1.dat -f csoaa.model
vw -t -i --link=logistic csoaa.model test.dat -p open_cost_1.predict
```

But this is outputting a single prediction of a label from 1 to 16 for each test instance, instead of something like (test instance 1):

```
1:0.53 2:0.21 3:0.121 4:0.98 5:0.12 6:0.78 7:0.11 8:0.89 9:0.34 10:0.03 11:0.09 12:0.08 13:0.89 14:0.13 15:0.034 16:0.078 0
```

Which is the softmax probabilities of each label for that instance (apologies if doesn't sum up to 1).

martinpopel:
Have you tried `--probabilities`?

dhruvghulati:
I am doing cost sensitive classification for multi class sentence classification using logistic regression, and find there is very low recall (as with normal logistic regression). However if I threshold the probabilities (so anything below 5% probability is an unknown class and so always is wrong), I can use the threshold as a hyperparameter and up my recall to an acceptable rate. I did this with the same problem in scikit learn using the `LogisticRegression.predict_proba()` method and it works well.

dhruvghulati:
Also, small point but I really wish someone put `--probabilities` in the documentation. If you google "Vowpal Probabilities instead of Predictions", this `--link=logistic` parameter appears, and never clearly where in the command line it should appear in a command.

dhruvghulati:
@martinpopel what would my updated original command be with `--csoaa_ldf=mc`?

martinpopel:
> could you put --probabilities in the documentation?

OK, I did now: https://github.com/JohnLangford/vowpal_wabbit/wiki/Predicting-probabilities
(It's wiki, so you/anyone can improve it.)

martinpopel:
> what would my updated original command be with --csoaa_ldf=mc?

[...]

PR #1244, opened 2017-05-15
Expose raw predictions in java wrapper (#1118 #1233)

tilayealemu:


JohnLangford:
Sorry, can you resolve conflicts on this?  

We're dealing with a bit of a mess because I had to back out the concurrent java interface patch.

tilayealemu:
@JohnLangford Resolved.

JohnLangford:
@jmorra Can you take a look at this? It's short.

java/src/main/c++/vowpalWabbit_learner_VWMulticlassLearner.cc
```diff
@@ -11,3 +11,17 @@ JNIEXPORT jint JNICALL Java_vowpalWabbit_learner_VWMulticlassLearner_predict(JNI
 JNIEXPORT jint JNICALL Java_vowpalWabbit_learner_VWMulticlassLearner_predictMultiline(JNIEnv *env, jobject obj, jobjectArray example_strings, jboolean learn, jlong vwPtr)
 { return base_predict<jint>(env, example_strings, learn, vwPtr, multiclass_predictor);
 }
+
+jfloatArray multiclass_raw_predictor(example* vec, JNIEnv *env){
+  size_t num_values = vec->l.cs.costs.size();
+  jfloatArray j_labels = env->NewFloatArray(num_values);
+  for (int i=0 ; i<num_values; i++) {
+    jfloat f[] = { vec->l.cs.costs[i].partial_prediction };
+    env->SetFloatArrayRegion(j_labels, i, 1, (float*)f);
```
jon-morra-zefr: I may be a little lost here, but how are you sure that the length of `f` is 1?

java/src/main/java/vowpalWabbit/learner/VWMulticlassLearner.java
```diff
@@ -13,4 +13,24 @@
...
     protected native int predictMultiline(String[] example, boolean learn, long nativePointer);
+
+    protected native float[] rawPredict(String example, boolean learn, long nativePointer);
+
+    /**
+     * Get raw prediction output.
+     *
+     * @param example a single vw example string
+     * @return Raw prediction
+     */
```
jon-morra-zefr: I don't think this code should go here.  From what I understand about VW, the action of getting a raw prediction is applicable to a wider variety of cases than just multi class prediction.  @JohnLangford can you comment on this.  If this is the case then we'd want to put this code higher up in the chain and allow more learners to take advantage of it.  If not, then this seems fine.  @deaktator what do you think?

JohnLangford:
I'm somewhat uncomfortable about the patch because it's using some prediction information squirreled away in the _label_ rather than the prediction data.  This was a hack Hal introduced some time ago which is certainly useful, but since then we've update our prediction types.  https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/example.h#L44 
The existing code already exposes several prediction types: 
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/csoaa.cc#L808  
Do any of these work for what you are trying to do?

jmorra:
I wonder if you can use `VWProbLearner` to get the probabilities from a csoaa model?  If so then maybe we don't need this PR.  Is this possible?

JohnLangford:
With the right flags, yes.   The action-score representation is also available.

@tilayealemu are you using --csoaa or --csoaa_adf?

java/src/main/c++/vowpalWabbit_learner_VWMulticlassLearner.cc
tilayealemu: Expected a double here instead of an array because there should be one value. I ran tests to verify array sizes are all 1 in this case. My understanding is there could be cases where array sizes can be >1 (adf?). I can investigate further if we decide to go with the PR.

tilayealemu:
@JohnLangford @jmorra 

I'm using --csoaa.

My goal is to rank all actions from best to worst. I have tried both scores and probabilities. Neither of them worked (discussed under #1233). Raw predictions were understandable and usable. Others have raised raw predictions as well [1, 2]. But if there is a way to get raw without this PR then I'm all for that. 

[1] https://groups.yahoo.com/neo/groups/vowpal_wabbit/conversations/topics/5958
[2] https://github.com/JohnLangford/vowpal_wabbit/issues/1082

jon-morra-zefr:
After looking at this for a while now I think the more appropriate change to make is to add the ability for the CSOAA learner to return action scores.  On the LDF front there exists a flag called `--csoaa_rank` that (now) correctly returns the action scores and it works with the JNI layer.  I see no reason why this same functionality cannot be ported over to the base csoaa learner.  I don't have the expertise to do this.  @JohnLangford do you think you can take a stab at this?  If this is done this PR is not necessary and no changes to the JNI layer are needed as the JNI layer will return the appropriate type (`ActionScores`) when the `--csoaa_rank` flag is set for traditional CSOAA learning.

JohnLangford:
I think Jon Morra is right here---it's better (and more uniform) to have the system --csoaa have the same return types as --csoaa_ldf.  This implies some c++ coding.  Are you up for that?

jon-morra-zefr:
@JohnLangford this is out of my area of expertise.  I've been struggling with it for a number of hours now and I'm not really sure how to do this.  Is there any chance you could take a stab at it?

tilayealemu:
Saw the c++ code too, and it was beyond me. Happy to help if we decide to update any of the java code though. I know it better now.

JohnLangford:
I'll look into it, but there are some higher priorities at the moment
(build is breaking...).

-John

jon-morra-zefr:
@JohnLangford I might need this feature pretty soon, I was wondering if you had any ETA on when you might be able to look at this.

JohnLangford:
I'm working towards a release before ICML.  Hopefully something addressing this can go in.

JohnLangford:
I've decided to not deal with this right now, although a pull request remains welcome.
(1) There is a refactoring brewing which will deal with this in the future.  In essence, I want to refactor things so multiline examples are first class within VW and then refactor csoaa (amongst many others) so we do not have different versions for multline and singleline representations.
(2) You can already get the desired functionality through csoaa_ldf.  It's slightly more curmudgeonly since you must use a multiline example but that should work fine.

Issue #980, opened 2016-04-20
"error, failed to write example" problem - c# wrapper


Wirack:
I am utilising the c# vowpal wabbit wrapper version 8.0.0.46.

I have 90 instances of vowpal wabbit running, each of which has its own assigned cachefile name (replace X in line below with the specific instance number.)

" --cache_file mycachefilenameX.cache  --oaa 7 --loss_function logistic --sgd --passes 150"

With these amount of vowpal wabbit intances running I have experienced the following error:

> error, failed to write example

 which I have located to: [io_buf.h](https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/io_buf.h) line 186

That is, it seems like a problem related to the flushing of the caches.
In cases where caches (and thus passes) are not used, I do not receive the error.
I also do not receive the error if I have a much lower amount of instances (I tried with 9 instances)

I do, however, have no idea if this is an error on my part or from vowpal wabbit.

eisber:
This might be related to https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/io_buf.cc#L103

We had some discussion on how to fix that, but haven’t come to an agreement. My proposal was to switch to boost io. I believe it’s not save to use io_buf with more than 32 instances in the same process on Windows. The errors I experienced were spurious.
Out of curiosity, how do you pass example data into VW? Strings, JSON or C# objects?

Markus

Wirack:
Thanks for the reply. That is certainly worth looking into.
As far as passing example data into VW, I currently pass it as strings, but have also tried C# Objects with the same error.

eisber:
Oh sorry for the confusion. The question on passing example data in is completely unrelated to you cache file issue.

Wirack:
Thanks, so your suggestion would be to switch from using io_buf.h to boost io, and then recompile vowpal wabbit, such that it is adjusted to those changes?

Also, is it an issue you are discussing how to handle at this point in time?

JohnLangford:
If I recall right, the problem is that windows only supports a 
relatively limited set of file descriptors.   Can you reproduce the bug 
with 33 processes but not 32?

Boost IO work is not presently happening (too busy with other stuff).  
There is almost certainly an easier fix to this problem related to more 
explicitly representing things?

-John

Wirack:
Hi.
Yes that idd seems to be the limit. I just tested and reprocuded the bug with 33 processes but not with 32 processes.

[...]

PR #1249, opened 2017-05-22
io_buf.cc: Increase the amount of files supported on windows to half of _getmaxstdio()

Wirack:
Use [`_getmaxstdio`](https://msdn.microsoft.com/en-us/library/xt874334.aspx) to set the nhandle limit.

This adjustment address the issue [here](https://github.com/JohnLangford/vowpal_wabbit/issues/980).

[`_setmaxstdio`](https://msdn.microsoft.com/en-us/library/6e3b887c.aspx) could be used to increase the file descriptor limit (possibly as a command-line argument).


JohnLangford:
Merged, thanks.

PR #1263, opened 2017-06-21
Updating readme for python installation using anaconda

gramhagen:
capturing information from comments in issue #1208


rth:
The readme in the python folder https://github.com/JohnLangford/vowpal_wabbit/tree/master/python#troubleshooting advises not to use Anaconda though, which is confusing..

PR #1270, opened 2017-06-30
Bi-annual PR

eisber:


cs/azure/Join/QueueDictionary.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
```
JohnLangford: Why/how is this datastructure used?

vowpalwabbit/cntk.cc
```diff
@@ -0,0 +1,294 @@
+/*
```
JohnLangford: CNTK is not ready for the master branch.

vowpalwabbit/comp_io.cc
```diff
@@ -1,76 +1,148 @@
-#include "zlib.h"
 #include "comp_io.h"
```
JohnLangford: Have you computationally measured the impact of this alternative access to zlib?

vowpalwabbit/libvw.vcxproj
```diff
@@ -242,17 +242,91 @@
...
+    <Import Project="packages\boost_chrono-vc140.1.63.0.0\build\native\boost_chrono-vc140.targets" Condition="Exists('packages\boost_chrono-vc140.1.63.0.0\build\native\boost_chrono-vc140.targets')" />
+    <Import Project="packages\boost_container-vc140.1.63.0.0\build\native\boost_container-vc140.targets" Condition="Exists('packages\boost_container-vc140.1.63.0.0\build\native\boost_container-vc140.targets')" />
+    <Import Project="packages\boost_context-vc140.1.63.0.0\build\native\boost_context-vc140.targets" Condition="Exists('packages\boost_context-vc140.1.63.0.0\build\native\boost_context-vc140.targets')" />
+    <Import Project="packages\boost_coroutine-vc140.1.63.0.0\build\native\boost_coroutine-vc140.targets" Condition="Exists('packages\boost_coroutine-vc140.1.63.0.0\build\native\boost_coroutine-vc140.targets')" />
+    <Import Project="packages\boost_date_time-vc140.1.63.0.0\build\native\boost_date_time-vc140.targets" Condition="Exists('packages\boost_date_time-vc140.1.63.0.0\build\native\boost_date_time-vc140.targets')" />
```
JohnLangford: Why are there are large number of boost dependencies?

vowpalwabbit/marginal.cc
```diff
@@ -221,9 +221,11 @@ void predict_or_learn(data& sm, LEARNER::base_learner& base, example& ec)
   
 void finish(data& sm)
 { sm.marginals.~unordered_map();
-  sm.expert_state.~unordered_map();
-  for (size_t i =0; i < 256; i++)
-    sm.temp[i].delete_v();
+  if (sm.compete) {
+    sm.expert_state.~unordered_map();
```
JohnLangford: I believe you have a memory leak here.

vowpalwabbit/search_multiclasstask.cc
```diff
@@ -41,7 +41,7 @@ void run(Search::search& sch, vector<example*>& ec)
   size_t learner_id = 0;
 
   for(size_t i=0; i<my_task_data->num_level; i++)
-  { size_t mask = 1<<(my_task_data->num_level-i-1);
+  { size_t mask = UINT64_ONE <<(my_task_data->num_level-i-1);
```
JohnLangford: What is the point of UNIT64_ONE?

JohnLangford:
There are several comments in the code---please answer.  

Something is changing/breaking here: https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/248842440#L2187  Why?

In addition, it seems like there is quite a bit of JSON parser stuff on the C# side.  Why?  I thought we were nuking that in favor of the C++ parser?

vowpalwabbit/comp_io.cc
eisber: no

vowpalwabbit/libvw.vcxproj
eisber: boost split up all the packages. I have a dependency on boost and that pulls in the rest.

vowpalwabbit/marginal.cc
eisber: I changed to 

void finish(data& sm)
{ sm.marginals.~unordered_map();
  if (sm.compete) 
    sm.expert_state.~unordered_map();
 for (size_t i =0; i < 256; i++)
    sm.temp[i].delete_v();
}

vowpalwabbit/search_multiclasstask.cc
eisber: compiler differences:

const uint64_t UINT64_ONE = 1ULL;
#else
const uint64_t UINT64_ONE = 1i64;

cs/azure/Join/QueueDictionary.cs
eisber: removed unused class

eisber:
The DS front end still needs the C# JSON parser. Also for MSN the C++ parser is not a replacement as they have the data in C# objects and serializing directly will still be faster.
The JSON C++ parser is for parsing JSON strings, not JSON serializable C# objects.

vowpalwabbit/comp_io.cc
JohnLangford: Ok, we'll need to do that.

vowpalwabbit/libvw.vcxproj
JohnLangford: The one dependency pulls in all the others?  Does that bloat the binary?

vowpalwabbit/marginal.cc
JohnLangford: expert_state seems to be allocated unconditionally?

JohnLangford:
I'm somewhat uncomfortable with having two JSON parsers.  What could be done to avoid that?

eisber:
I don't believe this can be avoided if we want that performance optimization. The C++ JSON parser takes strings as input, the C# "JSON direct parser" takes C# objects annotated using JSON.NET attributes as input.

JohnLangford:
Perhaps it is missemantics to think about the C# side as a JSON parser.

-John

JohnLangford:
My understanding is that none of your changes should be causing this?
https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/253622315#L2669
Yet somehow it's happening.  Have you tried valgrind or the windows equivalent?

vowpalwabbit/libvw.vcxproj
eisber: it does not. it just sets up the library directories.

JohnLangford:
Merged, thanks. 

It looks like you might be able to remove some boost library dependencies on the windows side?

Issue #1271, opened 2017-06-30
./daemon-test.sh w/ foreground (test 152) fails in the travis env


arielf:
https://travis-ci.org/JohnLangford/vowpal_wabbit#L3089

Exit status 124 from 'timeout' means one thing only:

    This test is timing out (waiting for some input?)
    in the travis (non interactive) env.

'man timeout' says:
    If  the  command times out, and --preserve-status is not set,
    then exit with status 124.  Otherwise, exit with the status of
    COMMAND.   If no signal is specified, send the TERM signal upon
    timeout.  The TERM signal kills any process that does not block
    or catch that signal.  It may be necessary to use the KILL (9)
    signal, since this signal cannot be caught, in which case the exit
    status is 128+9 rather than 124.

The ../vowpalwabbit/vw: free(): invalid pointer: 0x00002ae11a9dd000 ***
error seems to be side-effect of killing the deamon by the 'timeout'
wrapper.

@seifert could you take a look?

BTW (but seems unrelated to problem): it is cleaner to pass `--` to end `grep` options rather than try to escape the searched pattern: `\-\-foreground`

seifert:
I will see to it.

JohnLangford:
This one seems pretty mysterious to me as the code change is pretty simple.  I haven't been able to reproduce on my local box.  Hopefully you can?  Without it, we'll need to disable things, as a failing test is to disruptive.

seifert:
Unfortunately, I haven't been able to reproduce problem on my environment. How often has the test been failing? Or it was first and only one fail?

I have theory why daemon-test.sh can block itself:

* on line 117 daemon starts
* then, on lines 119 - 139, PID is tested. The test was successful on these lines, so the daemon probably started successfully.
* then, on line 148, prediction request is sent
* finally, on line 150, there is waiting for response

I think that the daemon hadn't been ready when the prediction request was sent. Maybe vw process had been started, but didn't listen on port yet. So netcat had stopped immediately with exit status 1 (without any stdout) and infinite loop on line 150 was aborted by timeout.

I tried refactor daemon-test.sh, see https://github.com/seifert/vowpal_wabbit/commit/f601857581f9886cfcb4d9d508c43f2770ce608a. Is it ok and shall I create merge request?

JohnLangford:
It's failing >50% of the time on Travis.

Go ahead and send a pull request---let's see what Travis says.

-John

PR #1275, opened 2017-07-04
daemon-test.sh - check status code of the netcat

seifert:
daemon-test.sh refactor, trying to fix issue https://github.com/JohnLangford/vowpal_wabbit/issues/1271

### DIFF
--- a/test/daemon-test.sh
+++ b/test/daemon-test.sh
@@ -11,6 +11,7 @@ MODEL=$NAME.model
 TRAINSET=$NAME.train
 PREDREF=$NAME.predref
 PREDOUT=$NAME.predict
+NETCAT_STATUS=$NAME.netcat-status
 PORT=54248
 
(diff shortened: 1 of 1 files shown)


JohnLangford:
Looks good, thanks.

arielf:
I see travis succeeded after all with this change...
I'm surprised it did (don't understand how), but glad.
Thanks again @seifert !

seifert:
That's fantastic! Maybe `sleep 0.1`? I think, problem was that `vw` hadn't listened on port when `netcat` sent the request. So the test script waited inside the `until` infinite loop. Finally it was killed by `timeout`. More complex `until` loop now tests this situation.

arielf:
Yes, infinite loop in the `until` is the best explanation.
The changes in `sleep`/`wait` probably weren't and only made the test slower.
Thanks again.

PR #1278, opened 2017-07-06
JNI memory leak fix + CentOS6 build manual

thenobody:
Hi guys,

I've updated the JNI bindings with a fix that addresses a memory leak when creating new VWLearner instances: every `env->GetStringUTFChars` should be paired with a corresponding `env->ReleaseStringUTFChars` otherwise the string arrays never get cleaned up. (docs: [JNIEnv.GetStringUTFChars](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#GetStringUTFChars)).

Additionally, I've extended the README with CentOS 6 build and install steps.

Looking forward to your feedback.

Thanks,

Anton

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,24 +239,10 @@ git clone https://github.com/JohnLangford/vowpal_wabbit.git
 cd vowpal_wabbit
 git checkout <VERSION_TAG>
 
-CFLAGS='-fPIC' CXXFLAGS='-fPIC' ./autogen.sh
-sudo make install
+sudo make JAVA_HOME='/usr/lib/jvm/java-openjdk/' install
 ```
 
-### Build JNI bindings
-```bash
-sudo yum install -y java-1.8.0-openjdk java-1.8.0-openjdk-devel
-make FLAGS='-fPIC -I../rapidjson/include -std=c++11' JAVA_HOME='/usr/lib/jvm/java-openjdk/' UNAME=`uname` -C java things
-```
-
-### (Optional) additional OS tools
-##### Install `lsb_release`
-The tool is needed for Java JNI runtime to find the correct version of `vw_jni...` binding file in the classpath.
-Install this if you see `NullPointerException` from `vowpalWabbit.jni.NativeUtils.lsbRelease`.
-
-```
-sudo yum install -y redhat-lsb
-```
+This also builds the JNI bindings at `java/target/libvw_jni.so`.
 
 ## Code Documentation
 To browse the code more easily, do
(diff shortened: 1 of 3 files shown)


thenobody:
The JNI build is failing on the `libvw.a` path. Might be due to differences in `make` config between Ubuntu a CentOS. Investigating

JohnLangford:
@jmorra can you review?

jon-morra-zefr:
These changes look reasonable except for the README changes.  With the new way of doing the build I'm now longer packaging the compiled libraries with the jar.  These instructions are valid for older releases of the library though.  The new way requires manual building of the shared library through make and then either manual loading of the jar with `System.load(...)` or putting the shared library on the `java.library.path`.

thenobody:
@jon-morra-zefr Thanks for looking into this.
My understanding is that the only the last section of the README update (`additional OS tools`) is related to automatic loading of JNI binaries (which is still present in the Java code up to `v8.3.1`).
Since the Java code in `master` loads the lib explicitly as you mentioned would it make sense to move the`lsb_release` README bit into a wiki subpage? It would still be useful for people who are using older versions of VW.
Apart from this, I haven't had much luck so far with the JNI `make` failure on `libvw.a` path (`vowpal_wabbit/libvw.a` vs `vowpal_wabbit/.libs/libvw.a`). It seems that the `.libs/` dir is added automatically on centos `make`. Would you know of a way how to explicitly direct `make` to compile into `vowpal_wabbit/`?

jon-morra-zefr:
I'm not sure about the make failures, but you are correct about the README.  The boost version is ONLY important for the old version of the library.  Now since you're building it from scratch it'll properly link with whatever boost version you have on your system.

arielf:
`vowpal_wabbit/.libs/libvw.a` is a result of runing `autogen.sh` which pulls in GNU `autoconf`/`configure` which pulls in `libtool` and which overwrites the original `Makefile`s.   I generally recommend trying a simple `make` before going the GNU autoconf route.    Try `git checkout Makefile */Makefile` and run `make`. If this doesn't work on CentOS, I think it would be better/cleaner/simpler to fix that directly rather than resorting to `autogen.sh`.

jon-morra-zefr:
Yes, none of the autogen stuff works with Java.

thenobody:
@arielf indeed, that was the reason. Thanks a bunch!
I've updated the readme and paths in `java/Makefile`. The build should now be OK.

arielf:
@thenobody thanks!  Good to know things build smoothly on CentOS.

PR #1290, opened 2017-07-20
Making all VW responses serializable

jon-morra-zefr:
This is a simple PR that makes all the responses Serializable on the Java side.


ryan-deak-zefr:
Yeah.  Looks fine.  I looked through and all are objects with arrays or primitives or nested.  Should be fine.

JohnLangford:
Thanks!

PR #1291, opened 2017-07-21
Java with multipass

jon-morra-zefr:
Previously the JNI interface did not support that ability to do multiple passes over the data.  This PR fixes that.  In addition this PR exposes the base Java abstract class `VWLearnerBase` so we can avoid having a typed VW learned when just doing learning.


java/src/main/java/vowpalWabbit/learner/VWLearnerBase.java
```diff
@@ -17,7 +17,7 @@
  *
  * @author deak
  */
-abstract class VWLearnerBase<T> extends VWBase implements VWTypedLearner<T> {
+public abstract class VWLearnerBase<T> extends VWBase implements VWTypedLearner<T> {
```
ryan-deak-zefr: This is a bad idea I think.  `VWTypedLearner<T>` has everything we need: `predict` and `learn` methods and `close`.  When calling `VWLearners.create`,  just do `VWLearners.create<Object>` or something.

java/src/main/java/vowpalWabbit/learner/VWLearners.java
```diff
@@ -1,17 +1,12 @@
...
-import java.util.concurrent.locks.ReentrantLock;
-
 /**
  * This is the only entrance point to create a VWLearner.  It is the responsibility of the user to supply the type they want
  * given the VW command.  If that type is incorrect a {@link java.lang.ClassCastException} is thrown.  Refer to
  * {@link #create(String)} for more information.
  * @author jmorra
  */
 final public class VWLearners {
-    private static final Lock STATIC_LOCK = new ReentrantLock();
```
ryan-deak-zefr: Yup.  Should be removed.

java/src/main/java/vowpalWabbit/learner/VWLearnerBase.java
JohnLangford: @jon-morra-zefr what is the use case for this?

jon-morra-zefr: We don't need it, it turns out that @ryan-deak-zefr is correct.

JohnLangford:
Merged, thanks :-)

PR #1292, opened 2017-07-21
Removing spurious "cout" statements

usptact:
When running in --search mode, there are *lots* of printed lines that contain
- "del_top"
- "del_example_cond"


JohnLangford:
That's embarrasing. Thanks.

PR #1295, opened 2017-07-24
In Java JNI, added the ability to asynchronous close the underlying VW model

deaktator:
## Problem

With @jon-morra-zefr's introduction of support for `--passes` flag, the `close()` method might now take a long time when `--passes` is provided.  Blocking on `close()` may be unacceptable in certain situations.

## Solution

To accommodate the situation where blocking on `close()` is unacceptable, I added a method, `closer()` that returns a 

```java
java.util.concurrent.Callable<Boolean>
```

The result of `closer()` can be submitted to a `java.util.concurrent.ExecutorService` to be executed asynchronously so that closing does not block.  VW learns still adhere to the `java.io.Closeable` and this method still blocks.

## Validation
[`vowpalWabbit.ClosingTest`](https://github.com/deaktator/vowpal_wabbit/blob/async-close/java/src/test/java/vowpalWabbit/ClosingTest.java) in `java/src/test/java` shows how this new code may be used and demonstrates that `java.io.Closeable` is still observed.


java/pom.xml
```diff
@@ -64,6 +65,18 @@
             <version>4.11</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
deaktator: Only test dependencies add.  No new compile dependencies.

java/src/main/c++/vowpalWabbit_learner_VWLearners.cc
```diff
@@ -17,7 +17,7 @@ JNIEXPORT jlong JNICALL Java_vowpalWabbit_learner_VWLearners_initialize(JNIEnv *
   return vwPtr;
 }
 
-JNIEXPORT void JNICALL Java_vowpalWabbit_learner_VWLearners_closeInstance(JNIEnv *env, jclass obj, jlong vwPtr)
+JNIEXPORT void JNICALL Java_vowpalWabbit_learner_VWLearners_performRemainingPasses(JNIEnv *env, jclass obj, jlong vwPtr)
```
deaktator: Separated performing remaining passes from the close methods.  This is useful from a code transparency perspective.

java/src/main/java/vowpalWabbit/learner/VWBase.java
```diff
@@ -16,15 +17,15 @@
  *
  *
  */
-abstract class VWBase implements Closeable {
+abstract class VWBase implements VWLearner {
```
deaktator: Now extends `VWLearner` to expose the `closer()` method.

java/src/main/java/vowpalWabbit/learner/VWBase.java
```diff
@@ -16,15 +17,15 @@
...
     private boolean isOpen;
 
     /**
      * Load tests have shown that a Lock is faster than synchronized (this).
      * It was originally hypothesized that {@link java.util.concurrent.locks.ReadWriteLock} would be a better
      * alternative, but at this time this is not possible cause of <a href="https://mail.google.com/mail/u/0/?ui=2&ik=cdb4bef19b&view=lg&msg=14dfe18a4f82a199#14dfe18a4f82a199_5a">this</a>.
      */
-    protected final Lock lock;
+    final Lock lock;
```
deaktator: Increased privacy level.

java/src/main/java/vowpalWabbit/learner/VWBase.java
```diff
@@ -36,30 +37,34 @@
      * 3.  Call either {@link System#load(String)} or {@link System#loadLibrary(String)}<br>
      * If a user wishes to use the prepackaged JNI libraries (which is encouraged) then no additional steps need to be taken.
      */
-    protected VWBase(final long nativePointer) {
+    VWBase(final long nativePointer) {
```
deaktator: Increased privacy level to indicate that only classes  in the package can extend this class.

java/src/main/java/vowpalWabbit/learner/VWBase.java
```diff
@@ -36,30 +37,34 @@
...
         try {
-            if (isOpen) {
-                isOpen = false;
-                VWLearners.closeInstance(nativePointer);
-            }
+            closer().call();
         }
-        finally {
-            lock.unlock();
+        catch (Exception e) {
```
deaktator: Wrap to adhere to `Closeable` interface.

java/src/main/java/vowpalWabbit/learner/VWBase.java
```diff
@@ -96,4 +101,38 @@ public boolean equals(Object o) {
...
+         * May throw an unchecked exception if the underlying native code throws an Exception.
+         * </p>
+         * @return <code>true</code> if model was open and an attempt is made to close the model.
+         *         A return value of <code>false</code> indicates an attempt to close the model was
+         *         made previously.  This is consistent with other languages where a <code>true</code>
+         *         return value indicates success.
+         */
+        @Override
+        public Boolean call() {
+            lock.lock();
```
deaktator: All closing related behavior occurs after locking.

java/src/main/java/vowpalWabbit/learner/VWLearner.java
```diff
@@ -2,15 +2,50 @@
...
+     * <b>NOTE</b>: If the Callable is interrupted once started, there's a good chance that both the underlying
+     * VW model could corrupted and it may not be closable in the future (which could result in a memory leak).
+     * </p>
+     *
+     * @return a Callable responsible for closing the model.  The Callable should return <code>true</code> if
+     *         model was open and an attempt is made to close the model.  A return value of <code>false</code>
+     *         indicates an attempt to close the model was made previously.  This is consistent with other
+     *         languages where a <code>true</code> return value indicates success.
+     */
+    Callable<Boolean> closer();
```
deaktator: **This is the main addition.**

java/src/test/java/vowpalWabbit/ClosingTest.java
```diff
@@ -0,0 +1,214 @@
...
+    public void testCloseAsyncIsNearImmediate() throws Exception {
+        VWScalarLearner vw = model(PASSES);
+
+        long t1 = System.currentTimeMillis();
+        Callable<Boolean> closer = vw.closer();
+        long t2 = System.currentTimeMillis();
+        closer.call();
+        long t3 = System.currentTimeMillis();
+
+        assertEquals("Should only take a few milliseconds to get future", 0, t2 - t1, DELTA_TIME_MS);
```
deaktator: Shows the creation of the `Closer` is quick.

java/src/test/java/vowpalWabbit/ClosingTest.java
```diff
@@ -0,0 +1,214 @@
...
+        assertNotEquals("Should take longer to finish synchronous closing", 0, t3 - t1, DELTA_TIME_MS);
```
deaktator: Shows the actual closing can be slow.

java/src/test/java/vowpalWabbit/ClosingTest.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    @Test
+    public void closeMultipleTimes() throws IOException {
+        try {
+            final VWScalarLearner model = model(1);
+            model.close();
```
deaktator: Subsequent calls to close after the first call should have no effect according to the `Closeable` interface.

java/src/test/java/vowpalWabbit/ClosingTest.java
```diff
@@ -0,0 +1,214 @@
...
+        catch (Exception e) {
+            final PrintWriter str = new PrintWriter(new StringWriter());
+            e.printStackTrace(str);
+            fail("Shouldn't throw exception but threw " + str.toString());
+        }
+    }
+
+    @Test
+    public void multipleAsynchronousCloserCallsReturnTrueOnlyOnce() throws Exception {
+        final int numClosers = 10;   // Should be greater than 2 to avoid confusion.
```
deaktator: Shows asynchronous closing multiple times.  This (probabilistically) demonstrates thread-safety and shows only one close has an effect.

java/pom.xml
```diff
@@ -155,8 +168,12 @@
                 </executions>
                 <configuration>
                     <redirectTestOutputToFile>true</redirectTestOutputToFile>
-                    <forkMode>once</forkMode>
+
+                    <!-- It's important to run tests serially because we are doing timing tests in some of the tests. -->
```
deaktator: **This is important!**  *The tests can fail when not run serially.*  This may indicate a thread-safety issue in the Java *OR* C code.  This might be worth investigating.

deaktator:
Hey @JohnLangford, I was trying to get this merged to @jon-morra-zefr's master prior to your acceptance of his PR to `JohnLangford/master`.  This should play nicely with his changes but provides a mechanism to avoid blocking when closing the model, all the while maintaining the current level of thread-safety.

java/src/main/java/vowpalWabbit/learner/VWLearner.java
```diff
@@ -2,15 +2,50 @@
...
 /**
  * This is the super type of all different typed VW learners.  This type exists to ensure that the
  * {@link VWLearners#create(String)} method has a super type.
  */
 public interface VWLearner extends Closeable {
 
-    // This tells the implementations that an IOException cannot be thrown.
-    @Override void close();
+    /**
+     * This will close the underlying VW model after performing any
```
jon-morra-zefr: Missing words at the end of the first line.

JohnLangford:
Markus, this is the second time I've seen Appveyor fail on this test,
although it does not always happen.   Could we have a race condition?

https://ci.appveyor.com/project/JohnLangford/vowpal-wabbit/build/1.0.2122#L3615

-John

ryan-deak-zefr:
@JohnLangford I was just about to comment on the **appveyor** failure.  It's timing out for some reason.

java/pom.xml
JohnLangford: I'm interested in this comment.  What is the threading setup under use here?

deaktator: You can check it out at the maven [surefire plugin docs](https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html).  Search for "**Forked Test Execution**".  

> The parameter forkCount defines the maximum number of JVM processes that maven-surefire-plugin will spawn concurrently to execute the tests. It supports the same syntax as `-T` in maven-core: if you terminate the value with a 'C', that value will be multiplied with the number of available CPU cores in your system. For example `forkCount=2.5C` on a Quad-Core system will result in forking up to ten concurrent JVM processes that execute tests.
>
> The parameter reuseForks is used to define whether to terminate the spawned process after one test class and to create a new process for the next test in line (`reuseForks=false`), or whether to reuse the processes to execute the next tests (`reuseForks=true`).

deaktator: So, currently, it should spawn one thread and reuse for all tests as far as I can tell.  This works fine.  What was hanging intermittently was the use of `<forkMode>once</forkMode>`.  My suspicion is that it has something to do with calling `VWLearners.create` many times concurrently.  That's just a guess though.

JohnLangford:
There should be zero shared state between multiple created VW objects.  Is
that what it's doing?  (Creating multiple distinct VW objects?)

-John

deaktator:
@JohnLangford.  It looks like just one `VW` object.  Each Java call does the following on the C side:

```c
vw* vwInstance = VW::initialize(env->GetStringUTFChars(command, NULL);
```

JohnLangford:
A single VW object can not be operated on in multiple threads because the
code inside VW is not thread safe.  If you want to have a model which is
shared by multiple threads, you set this up more explicitly by initializing
a new VW object with an existing model.

-John

deaktator:
Hey @JohnLangford.  We take care of multi-threaded access to VW by locking anywhere that requires access to the C code.  The thread-safety issues I encountered before were on an incomplete version of the code that locked in the wrong place.  When I run the tests in parallel, they seem to work just fine now.  I ran them a bunch of times with forking in the tests and didn't see any issues.

JohnLangford:
Merged, thank you.

JohnLangford:
Maybe that the same test was run concurrently during another build? It's using Azure resources, so maybe there was overlap?

Markus

Issue #1276, opened 2017-07-04
Internal error (using l2s)


grafael:
I was trying to reproduce the example described in the document: [Joint Prediction](https://github.com/JohnLangford/vowpal_wabbit/wiki/l2s.pdf) 
That says:
```shell
vw -b 24 -d wsj.train.vw -c --search_task sequence  \
--search 45 --search_neighbor_features -1:w,1:w \
--affix -1w,_1w -f wsj.weights
```

I just changed the `wsj.train` to be my own file. My file contains the following data:
```
1|This
2|is
1|a
3|sequence

1|Another
2|sequence
```

Also, i changed the param `--search` to be equal **3** _(number of labels)_
After running the command I got the error:
```c
vw (search.cc:473): internal error (bug): expecting top namespace to be '131' but it was 128l
```

**Is there anything I'm missing? Or indeed it is a bug?**

JohnLangford:
@hal3 @KaiWeiChang any ideas?

JohnLangford:
This looks like a bug.  I can reproduce via:
echo "1| This" | vw -b 24 -d data -c --search_task sequence --search 2 --search_neighbor_features 1:w

JohnLangford:
Or even just: 
echo "1| f" | vw data --search_task sequence --search 2 --search_neighbor_features 1:w

usptact:
I use L2S to train a NER model. Is there a documentation for all L2S related options? In particular, here are the options I use:

```
vw --data $DATA \
     --learning_rate 0.05 \
     --cache --kill_cache \
     --threads \
     --passes 10 \
     --search_task sequence \
     --search $NUM_LABELS \
     --search_rollin=policy \
     --search_rollout=none \
     --named_labels $LABELS \
     -b 28 \
     --l1=1e-7 \
     -f $MODEL
```

Is this a reasonable setup? The two options that are not very clear to me are the "--search_rollin" and "--search_rollout".

Thanks

usptact:
I created a PR to remove the cout statements that print these lines #1292

hal3:
The bug seems to be: if you ask for neighbor features on a namespace that doesn't exist, bad things happen. In keeping with general VW style, I think the preferred behavior is for this to essentially be ignored. I'll fix this way shortly.

hal3:
I have a fix. One minor comment is that your data example above won't work anyway. You need a space after the "|" in, eg:

`1|This`

JohnLangford:
[...]

Issue #1042, opened 2016-07-19
[Python] Decreases of L2S perf when oracle=0 (needs PR)


pommedeterresautee:
I try to build a POS tagger for French with L2S based on French UD dataset.
I use the Python wrapper.

When I don't use `allowed=` parameter in `predict()` function I get a 0.08 hamming loss (state of the art is 0.03).

When I used `allowed=` to inform VW that I know deterministically some POS tag (for punctuation and det) and I provided it to VW through `allowed=` the hamming loss goes up to 0.11/0.12 (whatever the policies I tried for roll_in/roll_out).

It doesn't make sense to me as I am reducing the search space.
I have checked that the POS tag I deterministically guess are 100% matching the real POS tag.

I have also noticed that VW needs more call to `_run()` function to finish, with less info to guess...

Can someone help to understand why the perf are decreasing and may be what I can change to improve perf?

KaiWeiChang:
Could you provide the _run function?

pommedeterresautee:
Thank you @KaiWeiChang for your anwer.

Please find below my whole class:
To make the code more clear (and not make ref outside of the class), I just forced the correct answer for pos tag == 4 which is punctuation.
Results are of course the same than the one described in my previous message.

```python
class SequenceLabelerNoLDF(pyvw.SearchTask):
    def __init__(self, vw, sch, num_actions):
        pyvw.SearchTask.__init__(self, vw, sch, num_actions)
        self.counter = 0
        self.cache_ex = {}
        sch.set_options(sch.AUTO_HAMMING_LOSS | sch.AUTO_CONDITION_FEATURES)

    def _run(self, sentence):
        output = []
        for n, item in enumerate(sentence):
            pos, word, word_lower, word_description = item
            if word in self.cache_ex:
                ex = self.cache_ex[word]
            else:
                ex = self.vw.example({'w': ["w_" + word, "wl_" + word_lower],
                                      'p': ["p3_" + word_lower[:3], "s3_" + word_lower[-3:],
                                            "p1_" + word_lower[:1], "s1_" + word_lower[-1:]],
                                      't': ["upper_" + str(word[0].isupper()), "digit_" + str(word[0].isdigit())]})

                self.cache_ex[word] = ex

            pos_possible = [4] if pos == 4 else None

            pred = self.sch.predict(examples=ex, my_tag=n + 1, oracle=pos,
                                    condition=[(n, 3, 'x')], allowed=pos_possible)
[...]

PR #1296, opened 2017-07-24
fixed bugs #1276 and #1042

hal3:


JohnLangford:
This is merged, thanks :-)  The travis fail has to do with a rare race condition on termination that I haven't tracked down yet.

Issue #1300, opened 2017-07-26
JNI Layer throws Exceptions when close method is called in parallel ON DIFFERENT MODELS


ryan-deak-zefr:
## Problem

In the JNI layer, when multiple passes are enabled (`> 1`) and an attempt is made to close separate models in parallel, exceptions can be thrown.  This is true even though each model has it's own lock to guard all accesses to the native code paths.  Only a **global** lock around calls to the model `close` methods seems to avoid this issue.

I'm looking for help on identifying if any critical sections of the C code that can be guarded by a lock to avoid thread-safety issues.  I'm not asking for the C code to lock.  I just want help trying to figure out where to put the locks in the Java code that wraps the C code.

### Scope

This seems to be in PR #1291 but was not fixed with PR #1295

## Discussion

From empirical testing, it appears one of these lines seems to be the problem.  I am wondering if any of these use global state.  I am trying to figure out if we can lock only over a short critical section to avoid thread-safety issues.

1. ` adjust_used_index(*vwInstance);`
1. `vwInstance->do_reset_source = true;`
1. `VW::start_parser(*vwInstance);`
1. `LEARNER::generic_driver(*vwInstance);`
1. `VW::end_parser(*vwInstance);`

## Previous Conversation
In PR #1295 there was the following conversation:

@JohnLangford 

>There should be zero shared state between multiple created VW objects.  Is
that what it's doing?  (Creating multiple distinct VW objects?)

@deaktator 

> @JohnLangford. It looks like just one VW object. Each Java call does the following on the C side:
>
> `vw* vwInstance = VW::initialize(env->GetStringUTFChars(command, NULL);`

@JohnLangford 

>A single VW object can not be operated on in multiple threads because the code inside VW is not thread safe.  If you want to have a model which is shared by multiple threads, you set this up more explicitly by initializing a new VW object with an existing model.

@deaktator

> Hey @JohnLangford. We take care of multi-threaded access to VW by locking anywhere that requires access to the C code. The thread-safety issues I encountered before were on an incomplete version of the code that locked in the wrong place. When I run the tests in parallel, they seem to work just fine now. I ran them a bunch of times with forking in the tests and didn't see any issues.

## Tracking Down What's Happening

It appears @jon-morra-zefr pretty much copied the C# code for multiple passes, so this seems like it might apply to C# as well.  Both C# and JNI C++ code appear below as well as the calling code that blows up.  

[...]

PR #1303, opened 2017-07-31
VWJNI safe multiple passes

jon-morra-zefr:
After testing we have discovered that multiple passes within the JNI layer is not thread safe even within different VW instances.  Please refer to #1300 for more details.  While that issues is being discussed this PR will at least make the JNI layer thread safe again.


matanster:
@jon-morra-zefr Could you kindly explain what is meant here by "multiple passes"?

PR #1307, opened 2017-08-04
added two more weight initialization flavors - normal & truncated normal

rcherukuri12:
Adding two more flavors for weight initialization :  
a) normal_weights and 
b) truncated_normal_weights 
added the code in : 
a) parse_regressor.cc ( actual code )
b) parse_args.cc ( adding input params to boost_options )
c) global_data.h ( for declaring boolean params )
d) global_data.cc ( for initializing the boolean params to default 'false' )


vowpalwabbit/parse_regressor.cc
```diff
@@ -18,6 +18,8 @@ using namespace std;
 #include <math.h>
 #include <algorithm>
 #include <stdarg.h>
+#include <numeric>
```
arielf: Does this add new C++ prerequisite libraries to vw?  If yes we will have to update build docs as well.

vowpalwabbit/parse_args.cc
```diff
@@ -1201,6 +1201,8 @@ vw& parse_args(int argc, char *argv[], trace_message_t trace_listener, void* tra
     ("initial_regressor,i", po::value< vector<string> >(), "Initial regressor(s)")
     ("initial_weight", po::value<float>(&(all.initial_weight)), "Set all weights to an initial value of arg.")
     ("random_weights", po::value<bool>(&(all.random_weights)), "make initial weights random")
+    ("normal_weights", po::value<bool>(&(all.normal_weights)), "make initial weights normal")
+    ("truncated_normal_weights", po::value<bool>(&(all.tnormal_weights)), "make initial weights truncated normal")
```
arielf: Does truncated normal means exactly at +/- 1 SD?  What if someone wants more control?
Maybe make the interface `--normal_weights <mean>,<std>` with zero/omitted SD meaning some default.

This will need some documentation update too (in the wiki of command line args).

vowpalwabbit/parse_regressor.cc
arielf: Also needs a test case added to test/RunTests

rcherukuri12: No need to add any libs to vw build files.
I will add test case to RunTests.

vowpalwabbit/parse_args.cc
rcherukuri12: correct for now it does default mean=0, sd=1.
Implemented the way it is defined in tensorflow docs for default.
(https://www.tensorflow.org/api_docs/python/tf/truncated_normal)
will add user control of "mean" and "sd" values in the next iteration.

vowpalwabbit/parse_regressor.cc
```diff
@@ -18,6 +18,8 @@ using namespace std;
...
+#include <random>
```
JohnLangford: Instead of using <random> (which is not cross-platform), can you use the builtin pseudorandom numbers in rand48.h?  That's our knock-off which is cross platform.

JohnLangford: (It looks like you would need Box-Mueller, but that's easy enough.)

vowpalwabbit/parse_regressor.cc
```diff
@@ -69,6 +102,12 @@ template<class T> void initialize_regressor(vw& all, T& weights)
     weights.template set_default<random_positive_wrapper<T> >();
   else if (all.random_weights)
     weights.template set_default<random_weights_wrapper<T> >();
+  else if (all.normal_weights)
+    weights.template set_default<normal_weights_wrapper<T> >();
+  else if (all.tnormal_weights){
+    weights.template set_default<normal_weights_wrapper<T> >();
+    truncate(all,weights);
```
JohnLangford: This is going to barf if the weights are sparse.  Ideas?

vowpalwabbit/parse_regressor.cc
JohnLangford: Instead of using "random" (which apparently I can't put in quotes :-)

rcherukuri12: yes..I will switch to rand48.h and add the box-muller function to get normal dist. instead of <random>

JohnLangford:
Raja is planning to update to use the internal RNG.

rcherukuri12:
Added polar implementation of box-muller for normal distribution.

vowpalwabbit/parse_regressor.cc
```diff
@@ -50,9 +50,51 @@ template <class T> class random_weights_wrapper
       *iter = (float)(merand48(index) - 0.5);
     }
 };
+// box-muller polar implementation
+template <class T> class polar_normal_weights_wrapper
+{
+public:
+    static void func(typename T::iterator& iter)
```
JohnLangford: This isn't the Box-Muller I'm familiar with?  This one: https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform#Basic_form doesn't need the Do{...} While

rcherukuri12:
I did a polar implementation where there is a do/while .
Here are scipy normality test results:
NormaltestResult(statistic=0.039988554829643842, pvalue=0.98020428259320458) 
Want me to just change implementation to basic form ?

JohnLangford:
This went out of sync so I hand resolved and merged.   

Raja, some test cases would be welcome, illustrate the initializations, and prevent errors from propagating into this code.

PR #1312, opened 2017-08-12
reset current_pass to 0 rather than 1 in test_only

denik:
Otherwise "vw -t" reports halved 'number of examples per pass'.

without this patch

$ vw test/train-sets/rcv1_micro.dat -f tmp.model --save_resume --quiet && vw test/train-sets/rcv1_micro.dat -i tmp.model -t 2>&1 | egrep 'number of|passes'
number of examples per pass = 5
passes used = 2

with this patch:

$ vw test/train-sets/rcv1_micro.dat -f tmp.model --save_resume --quiet && vw test/train-sets/rcv1_micro.dat -i tmp.model -t 2>&1 | egrep 'number of|passes'
number of examples per pass = 10
passes used = 1

### DIFF
--- a/test/test-sets/ref/0099.stderr
+++ b/test/test-sets/ref/0099.stderr
@@ -8,7 +8,7 @@ Reading datafile = train-sets/0001.dat
 num sources = 1
 average  since         example        example  current  current  current
 loss     last          counter         weight    label  predict features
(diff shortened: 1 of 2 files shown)


JohnLangford:
Thanks :-)

PR #1314, opened 2017-08-16
Java Maven clean works

jon-morra-zefr:
After the refactor of the JNI layer I forgot to update the pom to the new library name.  This fixes that.


ryan-deak-zefr:
I see.  This was an issue because the library name was changed after the POM was authored so the library was clobber during a `mvn clean`.  Yeah.  This looks good.

jmorra:
@JohnLangford I cannot merge this in because the appveyor build is failing.  It appears to be failing on master too so I don't think this PR is the reason.  I'd like this merged in ASAP so I can do a public Java release.

PR #1315, opened 2017-08-16
Adding Java README

jon-morra-zefr:
I've added a Java README to correspond to the new Java version released (8.4.1).

### DIFF
--- a/java/README.md
+++ b/java/README.md
@@ -4,14 +4,20 @@ The Java layer for Vowpal Wabbit uses the [Java Native Interface](https://en.wik
 ```java
 class Foo {
     static {
-        VWScalarLearner learner = VWLearners.create("-f model.vw"); // Create the model.  The model type is checked here
-        learner.learn("0 | price:.23 sqft:.25 age:.05 2006"); // Learn some data
+        // Create the model.  The model type is checked here
+        VWScalarLearner learner = VWLearners.create("-f model.vw");
+        
+        // Learn some data
+        learner.learn("0 | price:.23 sqft:.25 age:.05 2006");
         learner.learn("1 2 'second_house | price:.18 sqft:.15 age:.35 1976");
         learner.learn("0 1 0.5 'third_house | price:.53 sqft:.32 age:.87 1924");
-        learner.close(); // Closing finalizes the model and frees up the native memory
+        
+        // Closing finalizes the model and frees up the native memory
+        learner.close();
         
         VWScalarLearner learner = VWLearners.create("-i model.vw -t --quiet");
-        float prediction = learner.predict("| price:0.23 sqft:0.25 age:0.05 2006"); // Get a prediction.
+        // Get a prediction.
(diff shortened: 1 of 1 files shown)


java/README.md
```diff
@@ -0,0 +1,44 @@
+# Summary
+The Java layer for Vowpal Wabbit uses the [Java Native Interface](https://en.wikipedia.org/wiki/Java_Native_Interface) to communicate between Java and the native Vowpal Wabbit C code.  The philosophy of this layer is that we want to expose a simple type safe, thread safe interface so a Java programmer can use Vowpal Wabbit for either learning or scoring.  An example usage is shown below
+
+```java
+class Foo {
+    static {
+        VWScalarLearner learner = VWLearners.create("-f model.vw"); // Create the model.  The model type is checked here
```
ryan-deak-zefr: I would put the comments on top of the code instead of to the right and add double new lines between 

* model creation
* learning
* closing

java/README.md
```diff
@@ -0,0 +1,44 @@
...
+final public class VWLearners {
+    static {
+        System.loadLibrary("vw_jni");
+    }
+}
+```
+
+The implication here is that platform specific library must be on your `java.library.path` with the [appropriate name](https://stackoverflow.com/questions/37203247/while-loading-jni-library-how-the-mapping-happens-with-the-actual-library-name).  Creating this file is already part of the standard build process of VW, therefore you only have to execute `make java` to create the shared library.  After executing this a file will be created under `java/target/*vw_jni*` where the `*` are platform specific.  This file then needs to be put somewhere on the `java.library.path` of the machine you wish to use the JNI library from.
+
+It should also be noted that Vowpal Wabbit makes all attempts at comparability between versions, but the only true way to guarantee comparability to use the exact same git hash that was used when creating the jar.  I will try and keep this list up to date as new versions are released
```
ryan-deak-zefr: compatibility?

java/README.md
```diff
@@ -0,0 +1,44 @@
...
+        float prediction = learner.predict("| price:0.23 sqft:0.25 age:0.05 2006"); // Get a prediction.
+        learner.close();
+    }
+}
+```
+
+More examples can be found in the [Java tests](src/test/java/vowpalWabbit/learner).
+
+## Important Notes
+    1.  Most standard Vowpal Wabbit command line options are supported when calling create.  If you find some that work from the regular command line and not within the JNI please file a bug.
```
ryan-deak-zefr: These numbers make a weird code block or something when I view on Github.  Maybe do:

```
1. point 1
1. point 2
1. point 2
```

which should result in:

1. point 1
1. point 2
1. point 2

ryan-deak-zefr:
Any point putting Boost-related nightmares (eh, issues) in the `README.md`?

jon-morra-zefr:
@ryan-deak-zefr I've addressed your comments, let me know what you think.

JohnLangford:
Merged, thank you.

PR #1318, opened 2017-08-25
cbify: support adf reduction

albietz:
Add support for the adf contextual bandit reduction from cbify: for each (single line) multiclass example, this generates multiline examples (one per action, then one empty example) by shifting feature indicies for each action, and reduces to cb_explore_adf instead of cb_explore.

### DIFF
--- a/vowpalwabbit/cbify.cc
+++ b/vowpalwabbit/cbify.cc
@@ -28,6 +28,12 @@ struct vw_recorder : public IRecorder<example>
   { }
 };
 
+struct cbify_adf_data
+{ example* ecs;
+  example* empty_example;
+  size_t num_actions;
+};
+
 struct cbify
 { CB::label cb_label;
   GenericExplorer<example>* generic_explorer;
@@ -38,6 +44,9 @@ struct cbify
   v_array<action_score> a_s;
   // used as the seed
   size_t example_counter;
+  vw* all;
+  bool use_adf; // if true, reduce to cb_explore_adf instead of cb_explore
+  cbify_adf_data adf_data;
 };
 
 vector<float> vw_scorer::Score_Actions(example& ctx)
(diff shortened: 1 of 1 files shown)


arielf:
I think this PR is good.  The AppVeyor failure seems unrelated.

vowpalwabbit/cbify.cc
```diff
@@ -64,6 +73,42 @@ void finish(cbify& data)
   delete_it(data.mwt_explorer);
   delete_it(data.recorder);
   data.a_s.delete_v();
+  if (data.use_adf)
+  { for (size_t a = 0; a < data.adf_data.num_actions; ++a)
+    { VW::dealloc_example(CB::cb_label.delete_label, data.adf_data.ecs[a]);
```
JohnLangford: It's computationally inelegant to dealloc a large number of examples often.   Can you tweak the code to just reuse the examples instead?  That's what the rest of VW does.

vowpalwabbit/cbify.cc
```diff
@@ -107,14 +201,20 @@ base_learner* cbify_setup(vw& all)
...
+  data.all = &all;
 
-  if (count(all.args.begin(), all.args.end(),"--cb_explore") == 0)
+  if (data.use_adf)
+  { init_adf_data(data, num_actions);
+  }
+
+  if (count(all.args.begin(), all.args.end(),"--cb_explore") == 0 && !data.use_adf)
   { all.args.push_back("--cb_explore");
     stringstream ss;
```
JohnLangford: I expected a pushback of --cb_explore_adf here?

vowpalwabbit/cbify.cc
albietz: Doesn't the `finish` function here just get called on termination (through `set_finish`)? I am allocating the examples once on setup (see `init_adf_data`), reusing them on every example, and this call deallocs them.

PR #1319, opened 2017-08-27
fix --binary when used together with bootstrap

denik:
When --binary and --bootstrap are used together, the binary's logic is currently applied
before bootstrap's causing undesirable effect: the finaly loss reported is no longer binary,
the predictions are no longer -1 / 1.

This commit puts binary's reduction last which fixes the behavior in this case: --bootstrap is no
longer affected by whether binary is present, but final predictions and loss are correctly binarized.

Note, I haven't tested or tried to understand how that affects interations of binary and other
reductions that change order because of this commit.

### DIFF
--- a/test/RunTests
+++ b/test/RunTests
@@ -1573,3 +1573,11 @@ echo "1 | feature:1" | {VW} -a --initial_weight 0.1 --initial_t 0.3
 {VW} -d train-sets/decisionservice.json --dsjson --cb_explore_adf --epsilon 0.2 --quadratic GT
     train-sets/ref/decisionservice.stderr
 
+# Test 159: test --bootstrap & --binary interaction
+{VW} train-sets/rcv1_mini.dat --bootstrap 5 --binary -c -k --passes 2
+    train-sets/ref/bootstrap_and_binary.stderr
+
+# Test 160: test --bootstrap & --oaa interaction
+# # (Also adds -q :: and -P1 to get & verify perfect predictions in 2nd pass)
+{VW} train-sets/multiclass --bootstrap 4 --oaa 10 -q :: --leave_duplicate_interactions  -c -k --passes 2 --holdout_off -P1
+    train-sets/ref/bootstrap_and_oaa.stderr
--- a/test/train-sets/ref/bootstrap_and_binary.stderr
+++ b/test/train-sets/ref/bootstrap_and_binary.stderr
@@ -1,0 +1,29 @@
+Num weight bits = 18
+learning rate = 0.5
+initial_t = 0
+power_t = 0.5
+decay_learning_rate = 1
+creating cache_file = train-sets/rcv1_mini.dat.cache
+Reading datafile = train-sets/rcv1_mini.dat
+num sources = 1
+average  since         example        example  current  current  current
+loss     last          counter         weight    label  predict features
+0.000000 0.000000            1            1.0  -1.0000  -1.0000      128
+0.000000 0.000000            2            2.0  -1.0000  -1.0000       44
+0.250000 0.500000            4            4.0  -1.0000  -1.0000      190
+0.250000 0.250000            8            8.0   1.0000   1.0000       34
+0.312500 0.375000           16           16.0   1.0000  -1.0000      193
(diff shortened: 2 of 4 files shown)


arielf:
Thanks so much for the catch and fix, @denik !

This change looks good/simple and it doesn't break any test so I think it should be merged in.

Could you also add a test to `test/RunTests` combining `--binary` and `--bootstrap` on one of the existing datasets?

From experience, combination of options, especially when reductions are involved, are the most brittle as this issue demonstrates.  I believe `--interactions` and `--bootstrap` combo is also still broken.

For even better coverage, I would also add `-c --passes 2` so the code-paths of (end+start of pass) logic are also covered, and work as expected.

JohnLangford:
I'm wondering if the reverse (moving bagging earlier) is more desirable?  In the changed position, there is no reasonable expectation of binary composing with (for example) --search.

denik:
Thanks @arielf and @JohnLangford for your comments. There is a test now.

I'm not sure how binary and search interact, so I've put both bs & binary above it per John's suggestion.

Using --binary and multiclass reductions together does not seem to make sense, does it? So it should not matter what's the order there.

Regarding multiclass (I'm looking at oaa code for reference) and bootstrap, it does not work currently, but it seems like it can be made work in both ways:

- by putting bootstrap above oaa. It works in a sense that it predicts something but I'm not sure what can be said about quality of those predictions.
- by keeping existing order (oaa then bootstrap) and fixing '--bs_type vote' to select the class with the most votes. Even better, '--bs_type vote' should be enabled by default when labels are multiclass.

denik:
> As John mentioned, this fixes the specific interaction between --binary and --bootstrap but it doesn't fix other --bootstrap interactions which are still broken. In particular, with multiclass (--oaa).
> Rather than moving --binary down, I just tried moving --bootstrap (bs_setup) up to the top of the reductions (line 1118 in parse_args.cc) like this:

Thanks for your comments, @arielf. I've applied the suggested changes.

I just noticed that --binary has an effect on --boosting as well. I wonder if these two were intended to be composable? The binary documentation is very simple ("--binary report loss as binary classification on -1,1" but the effect of it is far from simple when reductions after it are enabled).

Also, I think --binary breaks --oaa as well (however, they don't make sense together, so this maybe not relevant).

>  In the changed position, there is no reasonable expectation of binary composing with (for example) --search.

@JohnLangford, could you please elaborate a little bit? How do the two compose now? What else is --binary supposed to compose with?

arielf:
@JohnLangford can we have this important fix merged into mainline?
It allows the use of `--bootstrap` together with other options which were crashing before.

travis is passing all tests, including the new ones.

AppVeyor complains about something (unrelated) that:

- Is of its own making (0001.model being used by two tests, because it runs in parallel, I'm guessing)
- Has nothing to do with this PR (which doesn't use this 0001.model model at all). Looks like tests 94 and 143 which are also writing models/0001.model should simply use their own (new) names for models.

JohnLangford:
Yes, I expect to have time to digest next week.   There's still something
not-quite-right that I want to tweak before incorporating.

-John

JohnLangford:
We need to nail down the appveyor fail.  It does seem odd.

arielf:
I was trying to solve this warning/error:
[https://ci.appveyor.com/project/JohnLangford/vowpal-wabbit/build/1.0.2266#L3512](https://ci.appveyor.com/project/JohnLangford/vowpal-wabbit/build/1.0.2266#L3512)
```
Failed to delete model directory: The process cannot access the file '0001.model' because it is being used by another process.
```

arielf:
Hi John,

Any chance of merging this (or the alternative/near-equivalent https://github.com/JohnLangford/vowpal_wabbit/pull/1345) with the 2 additional tests, soon?

This seems like a pretty important/desirable fix (without this `--bootstrap` and or `--binary` cannot be combined with most other reductions).

Note that by now it causes conflicts at the tail of `RunTests` because of new tests added by to other merged commits.  Very simple to resolve though.

Thanks!

JohnLangford:
This is merged in a separate pull request.

PR #1323, opened 2017-09-07
Update README.md

andyreagan:
Couple typos

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ The very latest version is always available via 'github' by invoking one of the 
 ## For the traditional ssh-based Git interaction:
 $ git clone git://github.com/JohnLangford/vowpal_wabbit.git
 
-## If that hangs (it does on OSX), try the following SSH URL:
+## You can also try the following SSH URL:
 $ git clone git@github.com:JohnLangford/vowpal_wabbit.git
 
 ## For HTTP-based Git interaction
--- a/README.md
+++ b/README.md
@@ -33,7 +33,10 @@ The very latest version is always available via 'github' by invoking one of the 
 
 ```
 ## For the traditional ssh-based Git interaction:
-$ git clone git://github.com:JohnLangford/vowpal_wabbit.git
+$ git clone git://github.com/JohnLangford/vowpal_wabbit.git
+
+## If that hangs (it does on OSX), try the following SSH URL:
+$ git clone git@github.com:JohnLangford/vowpal_wabbit.git
 
 ## For HTTP-based Git interaction
 $ git clone https://github.com/JohnLangford/vowpal_wabbit.git


README.md
```diff
@@ -33,7 +33,7 @@ The very latest version is always available via 'github' by invoking one of the
 
 ```
 ## For the traditional ssh-based Git interaction:
-$ git clone git://github.com/JohnLangford/vowpal_wabbit.git
+$ git clone git://github.com:JohnLangford/vowpal_wabbit.git
```
arielf: @andyreagan  - Can you undo this line change?
(trying this on Ubuntu) It breaks the cloning process from perfectly working to a failure:

```
tmp$ git clone git://github.com:JohnLangford/vowpal_wabbit.git
Cloning into 'vowpal_wabbit'...
fatal: unable to connect to github.com:JohnLangford:
github.com:JohnLangford: Name or service not known
```

andyreagan:
Ah, well my change made it work for OSX. Maybe we should add a note for OSX rather than change the line?

arielf:
Are you saying that this illegal URL works on OSX?
Can you retry/verify?

Check `git help clone`, under the `GIT URLS` section.

According to this official documentation, the only case where a `:` after a hostname is allowed in the URL is to specify an optional non standard port:
```
          git://host.xz[:port]/path/to/repo.git/
```
`JohnLangford` is not a port so I don't understand how this could work (in any environment).

andyreagan:
The line, as is, just hangs on OSX. This hangs:

`git clone git://github.com/JohnLangford/vowpal_wabbit.git`

I proposed changing it to the line give by github in the "clone or download" dropdown, which does work for me on OSX.
The one that works for me, as copied right from github, is:

`git clone git@github.com:JohnLangford/vowpal_wabbit.git`

andyreagan:
I'm looking at the code I submitted, and I "halfway" changed it between the two, and you're right that it shouldn't work as my proposed change. I'll fix this.

andyreagan:
Should be better now...

arielf:
@andyreagan  - Thanks for the extra checks!  The suggested cloning line works fine for me.

However, I believe you have a specific/localized configuration/env issue (personal firewall?) that causes the hang in your env when using the original `git clone git://...` line.  It is unlikely to be an OS-X general issue since `git` is standard across platforms (same source code, same protocols used).

Also: there's already a line showing how to clone via `https` in the line below so that one may work for you too.

In any case, I'm not convinced the `README.md` should be changed to cover localized/personal one-off issues unlikely to be hit by other OS-X users.

As a compromise, how about adding this line just as yet another way to clone (just remove the "it will hang on OS-X" claim) ?

Thanks!

andyreagan:
Yeah it could be just a firewall issue.

My apologies for wasting so many of your brain cycles on this. Should be good to go!

JohnLangford:
Merged, thanks :-)

PR #1327, opened 2017-09-15
Fix --dsjson crash

eisber:


vowpalwabbit/gd.cc
```diff
@@ -108,17 +108,17 @@ inline void update_feature(float& update, float x, float& fw)
 
 //this deals with few nonzero features vs. all nonzero features issues.
   template<bool sqrt_rate, size_t adaptive, size_t normalized>
-  float average_update(float total_weight, float normalized_sum_norm_x, float neg_norm_power)
+  float average_update(double total_weight, double normalized_sum_norm_x, float neg_norm_power)
```
JohnLangford: Why change to double?

eisber: fix VS warning

JohnLangford:
I'd rather fix this way:
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/gd.cc#L491
since the downcast to float is happening just after it ceases to be an
accumulator and starts being used for general math ops.

-John

JohnLangford:
Status on this?

eisber:
It’s changed in the MSFT branch

-Markus

PR #1336, opened 2017-10-05
baseline reduction: separate learning of additive regression baseline

albietz:
This reduction allows a regression learner to separately learn an additive baseline prediction from only "constant" features (taken from the constant_namespace), and the residual on top of that. This seems to make it faster to learn a possibly large constant offset in practice.

cc @JohnLangford


JohnLangford:
The windows barf here: https://ci.appveyor.com/project/JohnLangford/vowpal-wabbit/build/1.0.2255#L2307 is presumably because the windows build doesn't include the new file.

vowpalwabbit/baseline.cc
```diff
@@ -0,0 +1,70 @@
...
+
+struct baseline
+{ example* ec;
+};
+
+template <bool is_learn>
+void predict_or_learn(baseline& data, base_learner& base, example& ec)
+{ if (is_learn)
+  { // do a full prediction, for safety in accurate predictive validation
+    base.predict(ec);
```
JohnLangford: You can factor base.predict() out of the if/else for simplicity.

JohnLangford:
This looks good to go other than the minor refactoring and fixing the windows build about here: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/vw_dynamic.vcxproj#L436 .  Can you tweak?

albietz:
Some comments:
* I added a learning rate multiplier based on the largest label magnitude seen so far. It seems like occasionally these get really large (e.g. I was getting values extremely large values at some point when using doubly robust estimates, even though labels were smaller than 10), hence the cap at 1000. It might be useful to allow to explicitly add the multiplier as a flag instead.
* I added an option for using a separate example with a single global feature for the baseline (assuming the examples don't have that feature), which seems easier than fiddling with feature values if an example has other constant features other than the global. Perhaps I can use a separate namespace instead to avoid conflicts?

JohnLangford:
Merged in, thanks.

PR #1345, opened 2017-11-01
Incorporate Denis Bilenko's changes + fix AppVeyor errors

arielf:
Hopefully this passes all tests so we can have this change merged in.

Thanks to Denis Bilenko for the actual fixes. He deserved most of the credit.


arielf:
Most of these changes are already incorporated in JohnLangford/master (and are simply pulls from JL master). I only added the last 5 commits above to try and help Denis' changes to go in.

JohnLangford:
@eisber do you know what's going wrong here? https://ci.appveyor.com/project/JohnLangford/vowpal-wabbit/build/1.0.2262#L3522

test/RunTests
```diff
@@ -1503,7 +1503,7 @@ printf '3 |f a b c |e x y z\n2 |f a y c |e x\n' | \
 
 # Test 143: Test 1 using JSON
 {VW} -k -l 20 --initial_t 128000 --power_t 1 -d train-sets/0001.json --json \
-    -f models/0001.model -c --passes 8 --invariant \
+    -c --passes 8 --invariant \
```
JohnLangford: Why modify the test here?

arielf: Hi John,

This is because I saw the Appveyor test fail when the same file is written to by different tests.
(perhaps because it supports running tests in parallel)
The file model name `models/0001.model` appears in 3 different tests.

Since to produce the output it is not critical to write a model to disk, and since the model is not used later, I simply removed the `-f ...`. Another possible fix is to give the model a unique name for each separate test.

eisber:
Since it is  the output filename, it should just be unique.

arielf:
I debated with myself whether to unique the file names or remove them altogether.

I liked removing the `-f modelfile` in the later two tests better because:

- The models produced aren't used anywhere so there's no guarantee they are even correct
- Having many unused/temp models unnecessarily pollutes the environment
- Without storing to disk at the end, the tests runs significantly faster
- The `-f file` store model to disk code path is tested many times in the test suite already, so it is pretty redundant

JohnLangford:
I've started fooling with this myself in #1360 .  Something funky is going on in the windows test suite that we need to fix.

JohnLangford:
Merged finally!  @rajan-chari figured out the windows build issue.

PR #1348, opened 2017-11-06
add --classweight

denik:
$ vw --classweight LABEL1:WEIGHT1[,LABEL2:WEIGHT2...]

Multiplies the weight of the examples with label equal to LABEL1 by WEIGHT1, examples with label equal to LABEL2 by WEIGHT2 and so on.

This is already possible to achieve by rewriting the input file, however, making this built-in makes it much faster to experiment with different class weights.

Note, that this does affect the 'average loss' reported, the same way as modifying the examples' weight would.


JohnLangford:
This seems like a good feature idea but the implementation is antimodular.  Could I ask you to refactor things so that classweights is a reduction instead of a modification to the parser?  Then you could stick the reduction about here: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parse_args.cc#L1128

denik:
I've moved all the code to a reduction. Can you take a look?

Note, that I had to replace usage of ec.l.multi.weight in oaa.cc and recall_tree.cc with ec.weight. It seems that ec.weight should be always the same, but maybe I'm wrong about it.

vowpalwabbit/label_parser.h
```diff
@@ -16,4 +16,5 @@ struct label_parser
   float (*get_weight)(void*);
   void (*copy_label)(void*,void*); // copy_label(dst,src) performs a DEEP copy of src into dst (dst is allocated correctly).  if this function is nullptr, then we assume that a memcpy of size label_size is sufficient, so you need only specify this function if your label constains, for instance, pointers (otherwise you'll get double-free errors)
   size_t label_size;
+  int (*get_int_label)(void*);
```
JohnLangford: Why is this needed?

vowpalwabbit/oaa.cc
```diff
@@ -133,7 +135,7 @@ void finish_example_scores(vw& all, oaa& o, example& ec)
   { if (ec.l.multi.label <= o.k) // prevent segmentation fault if labeĺ==(uint32_t)-1
       correct_class_prob = ec.pred.scalars[ec.l.multi.label-1];
     if (correct_class_prob > 0)
-      multiclass_log_loss = -log(correct_class_prob) * ec.l.multi.weight;
+      multiclass_log_loss = -log(correct_class_prob) * ec.weight;
```
JohnLangford: This is fixing a bug.

vowpalwabbit/label_parser.h
denik: In order to apply class weight, we need to know example's label. This is a way to get the label regardless of example type (simple_label and multiclass store label differently).

vowpalwabbit/multiclass.cc
```diff
@@ -86,7 +91,8 @@ label_parser mc_label = {default_label, parse_label,
                          cache_label, read_cached_label,
                          delete_label, weight,
                          nullptr,
-                         sizeof(label_t)
+                         sizeof(label_t),
+                         get_int_label,
```
JohnLangford: Removable?

vowpalwabbit/parse_args.cc
```diff
@@ -1144,6 +1145,7 @@ void parse_reductions(vw& all)
   all.reduction_stack.push_back(bs_setup);
 
   all.reduction_stack.push_back(audit_regressor_setup);
+  all.reduction_stack.push_back(classweight_setup);
```
JohnLangford: This is to late in the reduction stack.   Can you shift to the spot I pointed out?

vowpalwabbit/simple_label.cc
```diff
@@ -38,6 +38,11 @@ float get_weight(void* v)
   return ld->weight;
 }
 
+static int get_int_label(void* v)
```
JohnLangford: Removable?

vowpalwabbit/multiclass.cc
denik: Not sure how to get example label otherwise for different label types. They all stored in a union, that makes it a bit complicated.

vowpalwabbit/oaa.cc
denik: should I commit it separately?

JohnLangford:
This seems much closer, but it looks like there is still some abstraction leakage.  Removable?

JohnLangford:
The appveyor build can presumably be fixed by tweaking the windows build for the new file just above and below here: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/vw_dynamic.vcxproj#L436

vowpalwabbit/label_parser.h
JohnLangford: The traditional method would be get the datatype in the arguments, but we actually have some type information here:
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/learner.h#L113 
now.  

You should be able to create the base learner and then observe it's type.  It's also important that the reduction set it's type appropriately with the init() call.

vowpalwabbit/oaa.cc
JohnLangford: No, it's fine.

denik:
I've pushed an update, please take a look.

vowpalwabbit/label_parser.h
denik: ah, do you mean add a new templater parameter to predict_or_learn that specifies how to get the label? so far I've done it at runtime. I can update to templated one.

JohnLangford:
The code looks good now, thanks :-)  I'll merge after automated tests succeed.

vowpalwabbit/label_parser.h
JohnLangford: I'd template myself---it should never vary once the reduction stack is instantiated.

denik:
I've templated the switch on pred_type in https://github.com/JohnLangford/vowpal_wabbit/pull/1348/commits/28eb92c92f27141cb7209808b94c1fcd1ee8bec6

JohnLangford:
Merged, thank you.

PR #1353, opened 2017-11-15
Baseline + cb changes

albietz:
* parameter for lr multiplier, added automatically in cbify
* mechanism to only use baseline when manually specified in a feature. Used in DR to only use baseline for loss estimates.

cc @JohnLangford

### DIFF
--- a/vowpalwabbit/baseline.cc
+++ b/vowpalwabbit/baseline.cc
@@ -149,19 +149,21 @@ base_learner* baseline_setup(vw& all)
     data.lr_multiplier = all.vm["lr_multiplier"].as<float>();
   data.check_enabled = all.vm.count("check_enabled") > 0;
   data.global_only = all.vm.count("global_only") > 0;
-  if (data.global_only)
-  { // use a separate global
-    data.ec->indices.push_back(constant_namespace);
-    // shifted index to avoid conflicts
-    data.ec->feature_space[constant_namespace].push_back(1, constant - 17);
-    data.ec->total_sum_feat_sq++;
-    data.ec->num_features++;
-  }
 
   base_learner* base = setup_base(all);
   learner<baseline>& l = init_learner(&data, base, predict_or_learn<true>, predict_or_learn<false>);
 
   l.set_finish(finish);
+
+  if (data.global_only)
+  { // use a separate example for global constant
+    data.ec->indices.push_back(constant_namespace);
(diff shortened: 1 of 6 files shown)


vowpalwabbit/baseline.cc
```diff
@@ -94,10 +145,17 @@ base_learner* baseline_setup(vw& all)
...
+  if (all.vm.count("lr_multiplier"))
+    data.lr_multiplier = all.vm["lr_multiplier"].as<float>();
+  data.check_enabled = all.vm.count("check_enabled") > 0;
   data.global_only = all.vm.count("global_only") > 0;
   if (data.global_only)
   { // use a separate global
-    VW::add_constant_feature(all, data.ec);
+    data.ec->indices.push_back(constant_namespace);
+    // shifted index to avoid conflicts
+    data.ec->feature_space[constant_namespace].push_back(1, constant - 17);
```
JohnLangford: I'm worried about the correctness of this line.   You can't generally subtract an arbitrary constant and 'hit' a parameter.  Instead, you need to obey the stride shift.   See for example this line: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/stagewise_poly.cc#L197 which creates an index for a feature in a polynomial learning reduction.

albietz: iirc (from when I was working on adf in cbify), I think the stride shift and mask aren't set up correctly yet when the setup function here is called. But perhaps I can just have this constant be in a different namespace (maybe a new line in `example.h`) to avoid the arbitrary index altogether here?

JohnLangford:
I'm agnostic about which index is used, but any index used needs to
essentially go through this operation:
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/stagewise_poly.cc#L68


stride_shift is defined here:
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/gd.cc#L1048
so after you call 'setup_base()' it should be readable.  (Yes, this isn't
as modular as it should be...)

-John

albietz:
I see - I thought the problem was with subtracting the constant 17, but actually the issue you're describing was already there before even with the `add_constant_feature`.
I will keep current code with the subtraction and do the shifting after the setup_base call.

JohnLangford:
Merged, thanks :-)

PR #1356, opened 2017-11-16
Fix issues detected by Secure Development Life check.  (SDL-7.0)

rajan-chari:
Fix code around SDL warnings:
1) cast to uint64_t when needed
2) handle Winsock errors
3) handle file rename() errors
4) handle realloc() failure
5) revert range based for loops to iterator based for loops


JohnLangford:
There's a barf in the linux compile here: 
https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/302803208#L1710

rajan-chari:
Checked in my change.  Will it run the check again automagically or do I need to do something?

c_test/c_test.vcxproj
```diff
@@ -217,4 +220,5 @@
     </PropertyGroup>
     <Error Condition="!Exists('$(SolutionDir)\.nuget\NuGet.targets')" Text="$([System.String]::Format('$(ErrorText)', '$(SolutionDir)\.nuget\NuGet.targets'))" />
   </Target>
+  <Import Project="..\sdl\SDL-7.0-NativeAnalysis.targets" />
```
JohnLangford: Do we want to do this for all windows builds?

test/RunTests
```diff
@@ -821,12 +821,12 @@ exit $ErrorCount;
 __DATA__
 # Test 1:
 {VW} -k -l 20 --initial_t 128000 --power_t 1 -d train-sets/0001.dat \
-    -f models/0001.model -c --passes 8 --invariant \
+    -f models/0001_1.model -c --passes 8 --invariant \
```
JohnLangford: Why the name change?

vowpalwabbit/array_parameters.h
```diff
@@ -262,8 +262,8 @@ class sparse_parameters
 		// TODO: this is level-1 copy (weight* are stilled shared)
 		if (!_seeded)
 		{
-			for (auto& pair : _map)
-				free(pair.second);
+		  for (auto iter = _map.begin(); iter != _map.end(); ++iter)
```
JohnLangford: Why use an explicit iterator?

vowpalwabbit/bs.cc
```diff
@@ -41,9 +41,10 @@ void bs_predict_vote(example& ec, vector<double> &pred_vec)
   // float sum_labels = 0; // uncomment for: "avg on votes" and getLoss()
   bool majority_found = false;
   bool multivote_detected = false; // distinct(votes)>2: used to skip part of the algorithm
-  int* pred_vec_int = new int[pred_vec.size()];
+  auto pred_vec_sz = pred_vec.size();
+  int* pred_vec_int = new int[pred_vec_sz];
```
JohnLangford: This isn't your fault, but generally speaking reductions should not use 'new' in the middle---we want to externalize memory allocations to reduce overhead.

c_test/c_test.vcxproj
rajan-chari: Yes.  It degrades gracefully.  It will be ignored if SDL tools are not present.

test/RunTests
rajan-chari: I wanted to keep it consistent with other name changes.  I append "_<test#>" when I notice an issue.

JohnLangford:
Look here: https://github.com/JohnLangford/vowpal_wabbit/pull/1360

-John

arielf:
If these models aren't actually used, why store them in the 1st place.
I would recommend removing the `-f <modelname>` from these tests instead.

JohnLangford:
Merged, thanks :-)  

Ariel, feel free to tweak tests.

PR #1357, opened 2017-11-17
cs_active

akshaykr:
Merging in TK's and Hal's CS active code. 

One thing I didn't see how to resolve was the use of example_t for sum of importance weights so far, which no longer exists (apparently due to some race condition?). We were using this for the radius in the active learning algorithm. Now I'm using priv.ec_seq[0]->example_counter, but I am not sure if this is correct. This is only in the l2s part.


JohnLangford:
It looks like the label printout somehow changed for cost sensitive classification?  https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/303338539#L3332  Can you fix?

akshaykr:
I eliminated some dead code but I do not see why/how the real label is getting printed in the main branch. It's happening in cost_sensitive::print_update, but in both versions (before/after pull) we are setting label_buf explicitly to known/unknown. 

My current hypothesis is that it is a namespace issue, since that is basically the only diff... But I don't really see what it is!

akshaykr:
Oh I was confused. Fixed but there are some issues with l2s still -- looking into it.

JohnLangford:
The windows build is presumably barfing because the new file needs to be included above and below here: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/vw_dynamic.vcxproj#L437  Fixable?

(And thanks for your work here!)

akshaykr:
I think it is fixed now.

vowpalwabbit/cost_sensitive.h
```diff
@@ -17,6 +17,13 @@ struct wclass
   float partial_prediction;  // a partial prediction: new!
   float wap_value;  // used for wap to store values derived from costs
   bool operator==(wclass j) {return class_index == j.class_index;}
+
```
JohnLangford: This is a bit of a mess since it's contaminating a label data structure with a bunch of nonlabel stuff.

vowpalwabbit/gd.cc
```diff
@@ -466,7 +466,8 @@ inline void pred_per_update_feature(norm_data& nd, float x, float& fw)
         }
         w[normalized] = x_abs;
       }
-      nd.norm_x += x2 / (w[normalized] * w[normalized]);
+      if (w[normalized] != 0)
```
JohnLangford: Do we know where this is coming from?  This is deep in a white-hot inner loop so it compromises computational integrity.

vowpalwabbit/gd.cc
```diff
@@ -479,7 +480,11 @@ float get_pred_per_update(gd& g, example& ec)
 { //We must traverse the features in _precisely_ the same order as during training.
   label_data& ld = ec.l.simple;
   vw& all = *g.all;
-  float grad_squared = all.loss->getSquareGrad(ec.pred.scalar, ld.label) * ec.weight;
+
+  float grad_squared = ec.weight;
+  if (count(all.args.begin(), all.args.end(),"--cs_active") == 0)
```
JohnLangford: Another strange test here.  Why does the estimate need to change for cs_active?

vowpalwabbit/gd.cc
```diff
@@ -498,6 +503,8 @@ float get_pred_per_update(gd& g, example& ec)
       }
     nd.pred_per_update *= g.update_multiplier;
   }
+  if (nanpattern(nd.pred_per_update) || infpattern(nd.pred_per_update))
```
JohnLangford: Another strange test.  Is this for debugging purposes?  This seems like the kind of thing which covers up a bug that should be addressed more directly.

vowpalwabbit/global_data.h
```diff
@@ -199,6 +199,13 @@ class namedlabels
 
 struct shared_data
 { size_t queries;
+  size_t overlapped_and_range_small;
```
JohnLangford: Why do we need to squirrel away information in this data structure?

vowpalwabbit/parse_args.cc
```diff
@@ -1529,7 +1531,23 @@ void finish(vw& all, bool delete_all)
 
     all.trace_message << endl << "total feature number = " << all.sd->total_features;
     if (all.sd->queries > 0)
-      all.trace_message << endl << "total queries = " << all.sd->queries << endl;
+      all.trace_message << endl << "total queries = " << all.sd->queries;
+    if (all.sd->overlapped_and_range_small > 0)
```
JohnLangford: All this codes seems like it should live in a finish() call in the reduction itself rather than being globally invoked.

vowpalwabbit/gd.cc
alekh: I believe I know the answer to this. We decided that for cs_active, it is better to use sum of x^2 rather than gradient squared as the metric for updating the weights and confidence bounds. The use of gradient squared driven metric in computing sensitivities was causing a lot of issues. Once your learner becomes good, it makes few mistakes in prediction so that the gradients are small, and hence the squared gradients do not grow. That ends up meaning that your confidence plateaus and you maintain a reasonable chance of querying things. This does not happen when things are based on squared x values, which also corresponds better to previous works with linear representations specifically.

JohnLangford: I'm not quite understanding.  getSquareGrad gets the squared gradient, right?

alekh: Yes, but you only need the squared gradient as a multiplier for adagrad, which we do not want in cs_active.

JohnLangford:
If we don't want a squared gradient multiplier, then why not just turn it
off directly via --normalized --invariant?

alekh:
Because we want x^2, instead of g^2x^2.

Alekh

JohnLangford:
--normalized --invariant has no g^2 dependence.

-John

alekh:
It does not have x^2 either?

Alekh

JohnLangford:
Ah, I see.

In this case, it should be an option to gd.cc and exposed as a template
parameter at this point rather than such an expensive operation.

-John

alekh:
--adanograd? :)

Alekh

JohnLangford:
maybe --adax to keep things positive?

akshaykr:
I added --adax as an option but haven't made it a template parameter. There is still an if statement but no array lookups, etc. Is there more to be gained by exposing as a template?

vowpalwabbit/cost_sensitive.h
akshaykr: I can move most of the stuff to cs_active but query_needed is set in cs_active but used in search. I'm not sure of a better place to have it?

JohnLangford:
The code is looking much better now.  Making adax be a template is straightforward.  Getting rid of query_needed is probably the trickier bit so concentrate effort there.

I'm planning to deal with other things related to a release first then I'll look into this further.

JohnLangford:
Status on this?  It looks like everything is dealt with except templatization of adax?  Should I do that?  (I have some time tonight.)

akshaykr:
Yeah that would be awesome. I’m out of commission tonight :/

JohnLangford:
Ok, I'll take it on.  You've done great work wrestling this into shape.

JohnLangford:
Merged! thank you.

akshaykr:
Awesome! I guess cs_active should be working but I am not sure about
search+active since I didn't get to make a test case. We will need to work
with Hal on it. search.cc is intense.

hal3:
It looks like not all of the cs_active stuff got merged? See eg https://github.com/hal3/vowpal_wabbit/blob/cs_active2/vowpalwabbit/search.cc#L956, which contains the cs_active flag in search.cc, to do the controlling.

JohnLangford:
Send a pull request?

-John

PR #1360, opened 2017-11-25
Checkin

JohnLangford:


JohnLangford:
@eisber, this test fail is confusing me: https://ci.appveyor.com/project/JohnLangford/vowpal-wabbit/build/1.0.2297#L3521  How can test 159 fail to delete 0001.model when only tests 1&2 use it (https://github.com/JohnLangford/vowpal_wabbit/blob/80bdc048bb2dbb7363a54c31dada092be311a42f/test/RunTests#L823 ) ?

eisber:
The test base class tries to delete the complete directory: https://github.com/JohnLangford/vowpal_wabbit/blob/master/cs/unittest/TestBase.cs
Maybe we should create unique directories per test case run.

JohnLangford:
Why is this happening for test 159 and not all the other tests which do not have a model?  And how is the fail message getting 0001.model which was last used in test 2?  It seems like something very nonlocal is happening.

eisber:
Not all tests inherit from TestBase. The tests share the same model directory (=nonlocal). Localizing this would mean to create a unique executing directory per run. Doing so shouldn’t be hard, one has to think a bit how to do cleanup.

JohnLangford:
Why doesn't test 158 have the same problem?

-John

PR #1367, opened 2017-12-01
Fix error message on incompatible types when call python vw.predict and vw.learn

Spikhalskiy:
Now if call the python wrapper in a way:
```
vw = pyvw.vw(quiet=True)
vw.learn(u'1 | a b c')
vw.predict(ex)
```
the result would be an exception like:
```
ArgumentError: Python argument types in
    vw.learn(vw, unicode)
did not match C++ signature:
    learn(boost::shared_ptr<vw>, boost::shared_ptr<example>)
```
Which is a bit unclear and breaks abstraction with exposing c++ code details.

Proposed solution: add a strict parameters check and print a message like it's done in ```example#__init__```


JohnLangford:
Tests are barfing: https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/310271396#L4166   Fixable?

JohnLangford:
@gramhagen Can you examine and comment?

gramhagen:
seems to be an erroneous injection of text into pyvw.py on the part of Travis

```
>           if hasattr(ec, 'setup_done') and not erappec.setup_done:
E           NameError: global name 'erappec' is not defined
```

the code in the change does not have that error:
https://github.com/JohnLangford/vowpal_wabbit/pull/1367/files#diff-9346bed6ca280f68056c096edb091d62L100
```
if hasattr(ec, 'setup_done') and not ec.setup_done:
```

I'd suggest bumping it to rerun Travis and see if it passes

Spikhalskiy:
@gramhagen no, it was really my typo, but I had just fixed it already when you checked.
@JohnLangford thanks for pointing out, it's fixed and tests are passed.

JohnLangford:
@gramhagen I'm assuming the patch itself is fine (looks good to me).

gramhagen:
yep, lgtm too

PR #1372, opened 2017-12-03
Add save_resume_test.py to automated testing

denik:
The test that fail are marked as 'known_failure' and will not report an error.

'make test' will now run this script as well.


JohnLangford:
(fyi, this test is showing things that I want fixed, so I'm leaving it unmerged at the moment as a reminder)

matanster:
I'm posting this here rather than a new issue as it's the same functional area. I'm seeing a decline in average sliding window costs when loading a model saved with `save-resume` working off a simulation harness I've built. My simulation harness is written in Scala (using the Scala Build Tool and is sliding window based in the way that it tracks model performance). It makes it very easy to test actual model performance in the online serving mode. May I suggest merging some variant of my Scala test harness allowing for such higher-level tests?

JohnLangford:
Yes, --save_resume is not a sticky parameter by default.  

Working through the test issues in this patch is on my todo list.  

W.r.t. a scala test infrastructure, how much overhead is this and what does it provide beyond the python?

matanster:
Thanks a lot for confirming my empirical trials ― I was not entirely sure what was going on, because after restarting a saved model, the performance would not start climbing up from the start point that it would in initial training, but only from some point "midway".

As to my test infrastructure ― I think it differentiates in being a real online test suite. It allows one to write tests that simulate an environment (through supplying arbitrarily flexible functions correlating context feature values to cost maps), and get logging of a model's performance per epoch, with regard to three metrics on a convenient sliding window basis: 

+ choice of a best-cost action (0/1)
+ aggregate accumulated cost
+ a "regret" against the simulated _known_ least cost per epoch.

It may also count the number of turns needed to arrive at a performance goal (for example, with simple simulations with epsilon greedy, I observe how many epochs has it taken for it to reach 1-epsilon sliding window performance.

Of course, in epoch above, I mean recommendation cycles, user turns, etc. not anything from an offline learning scenario.

JohnLangford:
Maybe this is appropriate for the demo directory?   Generally, I stick larger scale (i.e. those to expensive for the Travis time budget) there.

JohnLangford:
FYI, I've diagnosed several of the failures.
--bootstrap failed because the PRG seed was not saved (and the update rule is dependent on it).  That's fixed now.
--l1 and --l2 fail because of rounding issues.  In essence, VW does regularization in two ways---implicitly so computation remains sparse and explicitly via a sweep through the parameters.  The parameter sweep currently happens just before saving implying rounding point differences creep in whenever saves occur.
The other 3 test failures (pistol, ftrl, ksvm) remain undiagnosed.

I'm planning to merge a slightly modified version of the patch here: https://github.com/JohnLangford/vowpal_wabbit/pull/1514  Thank you for this patch---I expect it will be very useful going forward.

PR #1373, opened 2017-12-03
add --hash_seed option

denik:
When collisions are present, setting this will change the groups of features that collide.
If VW is used as a base model for ensemble classifier, setting different hash_seeds will reduce correlations.

I cleaned up hash_base static variable since I did not want there to be two places where hash seed can be set.


JohnLangford:
This seems well done.  Can you add a test example?

denik:
Added.

PR #1376, opened 2017-12-12
Add ability to pass a list in a python wrapper parameter, which will be translated into multiple keys

Spikhalskiy:
Currently vowpal-wabbit in cmd accepts multiple instances of the same key, for example, quadratic features could be configured thru "-q ab -q ac", on the same time it's impossible to make ```pyvw.vw(q="ab", q="ac")``` by python syntax limitation and to make ```pyvw.vw(q=["ab", "ac"])``` by the current implementation. And the only one way to do it now - mixing separate parameters with plain ```arg_str='-q ab -q ac'```.

A goal of this pull request is to make the second way of passing multi-valued keys acceptable.

### DIFF
--- a/python/tests/test_pyvw.py
+++ b/python/tests/test_pyvw.py
@@ -138,3 +138,12 @@ def test_regressor_args():
     # clean up
     os.remove('{}.cache'.format(data_file))
     os.remove('tmp.model')
+
+
+def test_keys_with_list_of_values():
+    # No exception in creating and executing model with a key/list pair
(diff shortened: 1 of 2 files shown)


JohnLangford:
This looks good to me.  @Scott-Graham-Bose any comment?

Scott-Graham-Bose:
This seems like a good idea and well implemented 👍

JohnLangford:
Merged, thanks :-)

PR #1377, opened 2017-12-12
Linux python3 setup

gramhagen:
updating setup.py to support additional python3 installations (recent releases of ubuntu in particular)

also updating python packaging instructions because pypi has changed their servers


JohnLangford:
Thanks :-)

PR #1378, opened 2017-12-12
Simple multi-class classification example adapted from fastText

kedarbellare:
## Summary

This PR adds a simple end-to-end example for multi-class classification that was adapted from fastText (https://github.com/facebookresearch/fastText). 

## Testing

Ran: `make && cd demo/dbpedia && ./classification-example.sh` on my macbook

### DIFF
--- a/demo/dbpedia/classification-example.sh
+++ b/demo/dbpedia/classification-example.sh
@@ -28,10 +28,12 @@ then
 fi
 
 VW_EXEC=../../vowpalwabbit/vw
+NUM_CLASSES=`cat ${DATADIR}/dbpedia_csv/classes.txt | wc -l`
 
 ${VW_EXEC} -d "${DATADIR}/dbpedia.train" \
   --cache_file "${RESULTDIR}/dbpedia.cache" -f "${RESULTDIR}/dbpedia.bin" \
-  --oaa 14 --passes 5 --ngram 2 --loss_function hinge \
+  --oaa ${NUM_CLASSES} --passes 5 --ngram 2 --skips 2 \
+  --loss_function hinge --bit_precision 25 --l2 1e-5 \
   -k --threads
 
 ${VW_EXEC} -t -d "${DATADIR}/dbpedia.test" \
--- a/demo/dbpedia/classification-example.sh
+++ b/demo/dbpedia/classification-example.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+#
+# Adapted from fastText/classification-example.sh
+
+myshuf() {
+  perl -MList::Util=shuffle -e 'print shuffle(<>);' "$@";
+}
+
+normalize_text() {
+  tr '[:upper:]' '[:lower:]' | sed -e 's/^\([0-9][0-9]*\),/\1 | /g' | \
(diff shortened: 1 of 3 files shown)


demo/dbpedia/classification-example.sh
```diff
@@ -0,0 +1,43 @@
...
+  cat "${DATADIR}/dbpedia_csv/train.csv" | normalize_text > "${DATADIR}/dbpedia.train"
+  cat "${DATADIR}/dbpedia_csv/test.csv" | normalize_text > "${DATADIR}/dbpedia.test"
+fi
+
+VW_EXEC=../../vowpalwabbit/vw
+
+${VW_EXEC} -d "${DATADIR}/dbpedia.train" \
+  --cache_file "${RESULTDIR}/dbpedia.cache" -f "${RESULTDIR}/dbpedia.bin" \
+  --oaa 14 --passes 5 --ngram 2 --loss_function hinge \
+  -k --threads
```
denik: No --bit_precision? It's likely to make a big difference there.

kedarbellare: I played around with the bit precision but it didn't change the result by much. the values i tried were 16 and 32. i tried 8 but that increases the error rate. is there some other value you think i should try?

one extra option available in fastText is the ability to specify a minimum count when building the vocabulary. that really doesn't fit with the way `vw` operates (i.e. using feature hashing) but i'm not sure if it helps fastText in terms of generalization or not.

arielf: Try anything above 18 (the default) e.g. `-b 20`, you may go as high as 31, I think but at really high values things may get significantly slower when storing the model.

kedarbellare: Updated based on your comments (I also played around with regularization and skips) PTAL.

Scott-Graham-Bose:
this is pretty cool, I've been curious about evaluating the different multiclass options vw has: log_multi, recall_tree, etc, and this would be a nice benchmark.

I put together a quick python script to generate metrics using sklearn
```
from __future__ import print_function
import argparse
from sklearn.metrics import classification_report, confusion_matrix


if __name__ == '__main__':
    parser = argparse.ArgumentParser(description='Evaluate Classification Results')
    parser.add_argument('--truth', dest='truth', type=str, required=True, help='path to truth file')
    parser.add_argument('--pred', dest='pred', type=str, required=True, help='path to prediction file')

    args = parser.parse_args()

    y_true = []
    with open(args.truth, 'rb') as t:
        for line in t:
            y_true.append(int(line.split(' ', 1)[0]))
    
    y_pred = []
    with open(args.pred, 'rb') as p:
        for line in p:
            y_pred.append(int(line.strip()))

    print(confusion_matrix(y_true, y_pred))
    print()
    print(classification_report(y_true, y_pred))
```

you can call it like so:
```
python eval.py --truth data/dbpedia.test --pred result/dbpedia.test.predict
```

kedarbellare:
@Scott-Graham-Bose thanks for the eval script!! do you mind if I incorporate your script into this PR or do you want to create a new PR yourself once this gets merged?

JohnLangford:
This looks good---will merge shortly.   Thanks :-)

PR #1380, opened 2018-01-03
Don't clobber --with-boost and --with-boost-libdir

ben-ng:
This is necessary to specify a different boost on CentOS.


JohnLangford:
Merged, thanks.

PR #1381, opened 2018-01-03
New options

JohnLangford:


JohnLangford:
@gramhagen can you take a look at this: https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/329162486#L4170 ?

Something is barfing in py.test tests but I haven't been able to reproduce locally after an hour or two of fooling around.  On my local machine, 'source activate test-python27' seems to not work and then py.test tests barfs with: 
ImportError: /usr/local/lib/python2.7/dist-packages/vowpalwabbit-8.5.0-py2.7-linux-x86_64.egg/pylibvw.so: undefined symbol: _ZN2vwC1ERKS_

I'm rather unclear on how python can fail at all.  Although this is a giant patch, all changes should be internal to VW.  In essence, this is a rewrite of the way that arguments work.

JohnLangford:
Never mind---bug is nailed.... now onto the windows tests.

PR #1389, opened 2018-01-15
Fixed compile issues with visual studio projects/sources 

rajan-chari:
1) added parser_helper.cc to vw_dynamic project 2) migrated cli files to use vw->opt_n_args.vm instead of vw->vm 3) switched libvw win build to MultiThreadedDebugDLL to keep it compatible with other projects


JohnLangford:
Merged, thanks.

PR #1390, opened 2018-01-16
boost cast error causing test failure

rajan-chari:
boost throws error because underlying type in options data structure is unsigned int.

### DIFF
--- a/cs/cli/vw_arguments.h
+++ b/cs/cli/vw_arguments.h
@@ -56,7 +56,7 @@ internal:
     m_commandLine = sb->ToString()->TrimEnd();
 
     if (vw->opts_n_args.vm.count("cb"))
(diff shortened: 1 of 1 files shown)


arielf:
It would be better to change this type (everywhere) to a C99 standard explicit size integer: 
```
#include <stdint.h>
```
Use:  `uint32_t`  instead of `unsigned int`  everywhere `unsigned int` is currently used.
This will ensure portability and stability for a long time.

rajan-chari:
Makes sense.  The registered variable is typed uint32_t (see below)

      arg.critical("cb", data->cbcs.num_actions, "Use contextual bandit learning with <k> costs")
      .
      .
      struct cb_to_cs {
              uint32_t num_actions;

arielf:
Thanks @rajan-chari !

PR #1403, opened 2018-02-07
Add --onethread option to disable parsing thread

denik:
Enabling this option fixes races conditions, like the one in #1383 

I've also modified 'make test' to run all the tests with --onethread enabled.

Another benefit of --onethread is that when running multiple vw instances on a multi-core machine it makes it easier to manage resources (the optimal number of vw processes is number of cores) whereas if vw uses two threads this is no longer true.


JohnLangford:
I reconciled this one and modified it in a few ways at #1422.

PR #1404, opened 2018-02-07
check for endLine when parsing

MelanieFreed:
When running in daemon mode, example parsing sometimes fails because the parser reads past the end of the example line.  These edits prevent that from happening.


JohnLangford:
Thank you :-)

PR #1417, opened 2018-03-02
Update boost-python installation info

kagemusha:
`--with-python3` flag isn't supported anymore for brew's `boost-python` https://stackoverflow.com/questions/48994346/installing-boost-python-for-python3-using-brew

Using flag caused me errors, but was able to install `vowpalwabbit` python package using `brew boost-python3` instead


JohnLangford:
@Scott-Graham-Bose take a look?

Scott-Graham-Bose:
looks good

PR #1421, opened 2018-03-06
Multiline example

rajan-chari:
Preliminary pull request.

Still need to fix:

a) Python binding tests
b) Cleanup
      1) formatting/stlyle cleanup
      2) check for any missed base.predict() base.learn() calls


rajan-chari:
csoa_ldf now takes complete multiline examples.  So single example learn and predict functionality no longer works.  This test invokes single line prediction functionality.  I will check with Hal and John to see how to support this test.

cs/cs/VowpalWabbitMultiLineExampleCollection.cs
```diff
@@ -10,6 +10,7 @@
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Linq;
+using System.Web.UI.WebControls;
```
eisber: why did you add this?

cs/cs/VowpalWabbitMultiLineExampleCollection.cs
```diff
@@ -90,7 +93,9 @@ private TPrediction Execute<TPrediction>(VowpalWabbit vw, Action<VowpalWabbitExa
                 // signal end-of-block
                 empty = vw.GetOrCreateNativeExample();
                 empty.MakeEmpty(vw);
-                predictOrLearn(empty);
+                ec_col.Add(empty);
```
eisber: is the newline example still needed at the end of the collection?

cs/test/RunTestsHelper.cs
```diff
@@ -0,0 +1,129 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using VW;
+
+namespace cs_test
```
eisber: did you move this? wasn't this always there?

vowpalwabbit/cb_adf.cc
```diff
@@ -280,54 +278,48 @@ void output_rank_example(vw& all, cb_adf& c, example& ec, v_array<example*>* ec_
...
+    global_print_newline(all);
   }
+  clear_seq_and_finish_examples(all, ec_seq);
 }
 
 void end_examples(cb_adf& data)
 {
-  if (data.need_to_clear)
-    data.ec_seq.erase();
+  // noop
```
eisber: can't you remove this completely and then simply not register the callback?

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -293,13 +279,11 @@ void predict_or_learn_softmax(cb_explore_adf& data, base_learner& base, v_array<
 
 void end_examples(cb_explore_adf& data)
 {
-  if (data.need_to_clear)
-    data.ec_seq.erase();
+  // noop
```
eisber: same as above, remove, don't register callback.

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -575,10 +525,17 @@ base_learner* cb_explore_adf_setup(arguments& arg)
       arg.trace_message << "warning: cb_type must be in {'ips','dr'}; resetting to ips." << std::endl;
   }
 
-  learner<cb_explore_adf>& l = init_learner(data, base, CB_EXPLORE_ADF::predict_or_learn<true>, CB_EXPLORE_ADF::predict_or_learn<false>, problem_multiplier, prediction_type::action_probs);
+  //learner<cb_explore_adf>& l = init_learner(data, base, CB_EXPLORE_ADF::predict_or_learn<true>, CB_EXPLORE_ADF::predict_or_learn<false>, problem_multiplier, prediction_type::action_probs);
```
eisber: remove this?

vowpalwabbit/csoaa.h
```diff
@@ -11,6 +11,25 @@ namespace CSOAA {
...
+    {
+      return ec_seq.pop();
+    }
+    return nullptr;
+  }
+
+  class restore_last_example
+  {
+  public:
+    restore_last_example(multi_ex& coll, example* ex) :_coll(coll), _ex(ex) {};
```
eisber: shouldn't the constructor do the opposite of the destructor?

vowpalwabbit/explore_eval.cc
```diff
@@ -37,30 +35,13 @@ struct explore_eval
...
-      base.predict(*ec, id);
-    ec->ft_offset = old_offset;
-  }
-}
-
 void end_examples(explore_eval& data)
 {
-  if (data.need_to_clear)
-    data.ec_seq.erase();
+  // noop
```
eisber: remove and don't register

vowpalwabbit/global_data.h
```diff
@@ -440,6 +440,7 @@ struct vw
   LEARNER::base_learner* cost_sensitive;//a cost sensitive learning algorithm.
 
   void learn(example*);
+  void learn(multi_ex&);
```
eisber: why & and not *? seems inconsistent with example*

vowpalwabbit/learner.cc
```diff
@@ -32,32 +32,136 @@ const char* to_string(prediction_type_t prediction_type)
...
+  ec_seq.push_back(ec);
+  return false;
+}
 
-    if ((ec->tag).size() >= 6 && (ec->tag)[4] == '_')
-      final_regressor_name = string(ec->tag.begin()+5, (ec->tag).size()-5);
+template <class T, void(* f)(T, multi_ex&)>
+void dispatch_multi_ex(vw& all, T& context, example* ec, multi_ex& ec_seq)
+{
+  if (complete_multi_ex(ec, ec_seq, all)) {
```
eisber: seems like a good place for guard? or is finish_example just updating counts?

vowpalwabbit/learner.h
```diff
@@ -104,23 +121,51 @@ typedef void (*tend_example)(vw& all, void* d, example& ec);
...
     learn_fd.learn_f(learn_fd.data, *learn_fd.base, ec);
     ec.ft_offset -= (uint32_t)(increment*i);
   }
-  inline void predict(example& ec, size_t i=0)
-  { ec.ft_offset += (uint32_t)(increment*i);
+  inline void learn(multi_ex& ec_seq, size_t i=0) 
+  {
+    if (ec_seq.size() > 0)
+    {
+      for(auto ec : ec_seq)
```
eisber: add guard or at least comment?

vowpalwabbit/learner.h
```diff
@@ -326,4 +460,48 @@ typedef void (*tend_example)(vw& all, void* d, example& ec);
...
+   }
+ }
+
+ template<bool is_learn>
+ void base_learn_or_predict(base_learner& base, example* ec, uint32_t id = 0)
+ {
+   // if base can handle multiline example, it cannot deal with single
+   // example.  so call it using a multiline example
+
+   if (base.accepts_multi_ex()) {
```
eisber: any chance of making this a template parameter?

vowpalwabbit/example.h
```diff
@@ -147,3 +147,6 @@ inline void add_passthrough_feature_magic(example& ec, uint64_t magic, uint64_t
 }
 
 #define add_passthrough_feature(ec, i, x) add_passthrough_feature_magic(ec, __FILE__[0]*483901+__FILE__[1]*3417+__FILE__[2]*8490177, i, x);
+
+typedef v_array<example*> multi_ex;
+typedef std::unique_ptr<multi_ex> mult_ex_ptr;
```
eisber: is this used anywhere? maybe rename to multi_ex_ptr?

cs/cs/VowpalWabbitMultiLineExampleCollection.cs
JohnLangford: I'd hope not---we shouldn't need two delimiters.

library/libsearch.h
```diff
@@ -91,11 +91,11 @@ template<class INPUT, class OUTPUT> class SearchTask
 };
 
 
-class BuiltInTask : public SearchTask< vector<example*>, vector<uint32_t> >
+class BuiltInTask : public SearchTask< v_array<example*>, vector<uint32_t> >
```
JohnLangford: Why is v_array preferred over vector?

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -109,13 +95,13 @@ example* test_adf_sequence(v_array<example*>& ec_seq)
...
 
 template <bool is_learn>
-void predict_or_learn_first(cb_explore_adf& data, base_learner& base, v_array<example*>& examples)
+void predict_or_learn_first(cb_explore_adf& data, base_learner& base, multi_ex& examples)
 {
   //Explore tau times, then act according to optimal.
-  if (is_learn && data.gen_cs.known_cost.probability < 1 && test_adf_sequence(data.ec_seq) != nullptr)
-    multiline_learn_or_predict<true>(base, examples, data.offset);
+  if (is_learn && data.gen_cs.known_cost.probability < 1 && test_adf_sequence(examples) != nullptr)
+    base_learn_or_predict<true>(base, examples, data.offset);
```
JohnLangford: Why isn't this just base.learn(examples)?

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -315,18 +299,21 @@ void finish(cb_explore_adf& data)
 //are specified. We print the first action and probability, based on
 //ordering by scores in the final output.
 
-void output_example(vw& all, cb_explore_adf& c, example& ec, v_array<example*>* ec_seq)
+void output_example(vw& all, cb_explore_adf& c, multi_ex& ec_seq)
 {
+  if (ec_seq.size() <= 0) return;
+
+  auto& ec = *ec_seq[0];
   if (CB_ALGS::example_is_newline_not_header(ec)) return;
```
JohnLangford: We should not need this line now, right?

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -358,48 +345,43 @@ void output_example(vw& all, cb_explore_adf& c, example& ec, v_array<example*>*
...
   }
 }
 
-
-void clear_seq_and_finish_examples(vw& all, cb_explore_adf& data)
+void clear_seq_and_finish_examples(vw& all, multi_ex& ec_seq)
 {
-  if (data.ec_seq.size() > 0)
-    for (example* ecc : data.ec_seq)
+  if (ec_seq.size() > 0)
```
JohnLangford: Can this function be shared across all multiline reductions?

vowpalwabbit/csoaa.cc
```diff
@@ -398,51 +404,111 @@ void do_actual_learning_oaa(ldf& data, base_learner& base, size_t start_K)
...
+template <bool is_learn>
+void do_actual_learning(ldf& data, base_learner& base, multi_ex& ec_seq_all)
+{
+  if (ec_seq_all.size() == 0) return;  // nothing to do
+
+  data.ft_offset = ec_seq_all.last()->ft_offset;
+
+  // Remove last element if empty.  This reduction expects no empty examples at end
+  // how ever other reductions need empty example at the end.  So driver adds empty example
+  example* empty_ex = remove_empty_last_example(ec_seq_all);
```
JohnLangford: This should be removable, right?

vowpalwabbit/csoaa.cc
```diff
@@ -725,120 +791,57 @@ void output_example_seq(vw& all, ldf& data)
   }
 }
 
-void clear_seq_and_finish_examples(vw& all, ldf& data)
+void clear_seq_and_finish_examples(vw& all, ldf& data, multi_ex& ec_seq)
 {
-  if (data.ec_seq.size() > 0)
-    for (auto ec : data.ec_seq)
+  if (ec_seq.size() > 0)
```
JohnLangford: Again, it seems like this should be a shared function

vowpalwabbit/csoaa.cc
```diff
@@ -725,120 +791,57 @@ void output_example_seq(vw& all, ldf& data)
...
+    global_print_newline(all);
   }
+  clear_seq_and_finish_examples(all, data, ec_seq);
 }
 
 void end_examples(ldf& data)
 {
-  if (data.need_to_clear)
-    data.ec_seq.erase();
+// noop
```
JohnLangford: Delete entirely.

vowpalwabbit/csoaa.h
```diff
@@ -11,6 +11,25 @@ namespace CSOAA {
...
   void finish_example(vw& all, csoaa&, example& ec);
+
+  inline example* remove_empty_last_example(multi_ex& ec_seq)
+  {
+    if (ec_seq.size() > 1 && example_is_newline(*ec_seq.last()))
+    {
+      return ec_seq.pop();
+    }
+    return nullptr;
+  }
```
JohnLangford: It seems like we should not need this because it's double punctuation to have both a vector and a newline example.

vowpalwabbit/example.h
```diff
@@ -147,3 +147,6 @@ inline void add_passthrough_feature_magic(example& ec, uint64_t magic, uint64_t
 }
 
 #define add_passthrough_feature(ec, i, x) add_passthrough_feature_magic(ec, __FILE__[0]*483901+__FILE__[1]*3417+__FILE__[2]*8490177, i, x);
+
+typedef v_array<example*> multi_ex;
```
JohnLangford: Why v_array instead of vector?

vowpalwabbit/explore_eval.cc
```diff
@@ -121,58 +102,54 @@ void output_example(vw& all, explore_eval& c, example& ec, v_array<example*>* ec
...
   }
 }
 
 
-void clear_seq_and_finish_examples(vw& all, explore_eval& data)
+void clear_seq_and_finish_examples(vw& all, multi_ex& ec_seq)
 {
-  if (data.ec_seq.size() > 0)
-    for (example* ecc : data.ec_seq)
+  if (ec_seq.size() > 0)
```
JohnLangford: The same function again.   consolidate?

vowpalwabbit/explore_eval.cc
```diff
@@ -272,10 +222,13 @@ base_learner* explore_eval_setup(arguments& arg)
...
 
-  learner<explore_eval>& l = init_learner(data, base, predict_or_learn<true>, predict_or_learn<false>, 1, prediction_type::action_probs);
+  learner<explore_eval>& l = init_learner(data, base, 
+    do_actual_learning<true>, do_actual_learning<false>, 
+    1, prediction_type::action_probs);
 
   l.set_finish_example(finish_multiline_example);
   l.set_finish(finish);
   l.set_end_examples(end_examples);
+  l.set_test_example(CB::example_is_test);
```
JohnLangford: This feels like it's a property of the label type rather than a property of the reduction.

vowpalwabbit/gen_cs_example.h
```diff
@@ -166,16 +166,16 @@ void gen_cs_example(cb_to_cs& c, example& ec, CB::label& ld, COST_SENSITIVE::lab
   }
 }
 
-void gen_cs_test_example(v_array<example*> examples, COST_SENSITIVE::label& cs_labels);
+void gen_cs_test_example(multi_ex examples, COST_SENSITIVE::label& cs_labels);
```
JohnLangford: Should these by pass by reference instead?  Or maybe const?

vowpalwabbit/global_data.h
JohnLangford: Personally, I'd prefer & since it avoids allowing operations like ++ which don't make sense.  I agree this is inconsistent though.  How viable is it to change learn(example*)?

vowpalwabbit/learner.cc
```diff
@@ -32,32 +32,136 @@ const char* to_string(prediction_type_t prediction_type)
...
+
+bool example_is_test(example& ec, vw& all)
+{
+  return all.l->is_test_example(ec);
+}
+
+/* example headers have the word "shared" */
+bool ec_is_example_header(example& ec)  
+{
+  v_array<CB::cb_class> costs = ec.l.cb.costs;
```
JohnLangford: Why does learner.cc need to know the label type to drive things with multiline examples but not singleline examples?

vowpalwabbit/learner.cc
```diff
@@ -32,32 +32,136 @@ const char* to_string(prediction_type_t prediction_type)
...
+  const bool need_to_break = VW::is_ring_example(all, ec) && (ec_seq.size() >= all.p->ring_size - 2);
+
+  if ((example_is_newline_not_header(*ec) && is_test_ec) 
+      || need_to_break)
   {
-    all.l->end_pass();
-    VW::finish_example(all, ec);
+    ec_seq.push_back(ec);
+    if (ec_seq.size() == 1) {
+      cout << "Something is wrong---an example with no choice.  Do you have all 0 features? Or multiple empty lines?" << endl;
```
JohnLangford: cout is the wrong thing to use here in general due to behavior in a library.  Generally, we want to throw an exception.

vowpalwabbit/learner.cc
```diff
@@ -32,32 +32,136 @@ const char* to_string(prediction_type_t prediction_type)
...
+{
+  const bool is_test_ec = example_is_test(*ec,all);
+  const bool need_to_break = VW::is_ring_example(all, ec) && (ec_seq.size() >= all.p->ring_size - 2);
+
+  if ((example_is_newline_not_header(*ec) && is_test_ec) 
+      || need_to_break)
   {
-    all.l->end_pass();
-    VW::finish_example(all, ec);
+    ec_seq.push_back(ec);
```
JohnLangford: Why add the newline example?  It seems like asking for trouble to double delimit the end of a multiline example?

vowpalwabbit/learner.cc
```diff
@@ -32,32 +32,136 @@ const char* to_string(prediction_type_t prediction_type)
...
+    VW::finish_example(all, ec_seq);  // clean up 
+  }
+}
 
-    if (!all.quiet)
-      all.opts_n_args.trace_message << "saving regressor to " << final_regressor_name << endl;
-    save_predictor(all, final_regressor_name, 0);
+template <class T, void(*f)(T, multi_ex&)>
+void multi_ex_generic_driver(vw& all, T context)
+{
```
JohnLangford: Is it possible to template this code so as to avoid a proliferation of drivers?

vowpalwabbit/learner.h
```diff
@@ -104,23 +121,54 @@ typedef void (*tend_example)(vw& all, void* d, example& ec);
   func_data end_pass_fd;
   func_data end_examples_fd;
   func_data finisher_fd;
+  test_example_data test_example_fd;
 
 public:
+  bool inline multiline_learn() const { return learn_fd.learn_multiline_f != nullptr; }
```
JohnLangford: We should be able to assume that things are either multiline for both learn & predict or neither learn & predict.

vowpalwabbit/learner.h
```diff
@@ -104,23 +121,54 @@ typedef void (*tend_example)(vw& all, void* d, example& ec);
...
+  inline void learn(example& ec, size_t i=0) { 
+    ec.ft_offset += (uint32_t)(increment*i);
     learn_fd.learn_f(learn_fd.data, *learn_fd.base, ec);
     ec.ft_offset -= (uint32_t)(increment*i);
   }
-  inline void predict(example& ec, size_t i=0)
-  { ec.ft_offset += (uint32_t)(increment*i);
+  inline void learn(multi_ex& ec_seq, size_t i=0) 
+  {
+    if (ec_seq.size() > 0)
```
JohnLangford: This seems like it should be an assert rather than a guard---anything producing zeroline multline examples should generate observable errors.

vowpalwabbit/learner.h
```diff
@@ -209,18 +257,46 @@ typedef void (*tend_example)(vw& all, void* d, example& ec);
...
+    return test_example_fd.test_example_f(ec);
+  }
+  void set_test_example(bool(*f)(example&))
+  {
+    test_example_fd.test_example_f = (ttest_example)f;
+  }
 
-  friend learner<T>& init_learner<>(T*, base_learner*, void (*l)(T&, base_learner&, example&),
-                                      void (*pred)(T&, base_learner&, example&), size_t, prediction_type::prediction_type_t);
+  friend learner<T>& init_learner<>(T*, base_learner*,
```
JohnLangford: Why should we need a learner that supports both single and multline examples?  It seems like this is cuttable?

vowpalwabbit/learner.h
```diff
@@ -311,7 +441,10 @@ typedef void (*tend_example)(vw& all, void* d, example& ec);
  template<class T> learner<T>& init_learner(base_learner* base,
                                             void(*learn)(T&, base_learner&, example&),
                                             void(*predict)(T&, base_learner&, example&))
-   { return init_learner<T>(nullptr, base, learn, predict, 1, base->pred_type); }
+ {
+   typedef void(*fptr)(T&, base_learner&, multi_ex&);  // compile hint to resolve the right template
```
JohnLangford: It seems like we should avoid requiring a multi_ex function pointer to eliminate all these repeated typedefs?

vowpalwabbit/learner.h
```diff
@@ -326,4 +459,48 @@ typedef void (*tend_example)(vw& all, void* d, example& ec);
    }
 
  template<class T> base_learner* make_base(learner<T>& base) { return (base_learner*)&base; }
+
+ template<bool is_learn>
+ void base_learn_or_predict(base_learner& base, multi_ex& ec_seq, uint64_t offset, uint32_t id = 0)
```
JohnLangford: These should be cuttable?

vowpalwabbit/search.cc
```diff
@@ -1162,15 +1153,18 @@ action single_prediction_LDF(search_private& priv, example* ecs, size_t ec_cnt,
...
     uint64_t old_offset = ecs[a].ft_offset;
     ecs[a].ft_offset = priv.offset;
-    priv.base_learner->predict(ecs[a], policy);
-    ecs[a].ft_offset = old_offset;
-
+    tmp.push_back(&ecs[a]);
     priv.empty_example->in_use = true;
     priv.empty_example->ft_offset = priv.offset;
-    priv.base_learner->predict(*priv.empty_example);
+    tmp.push_back(priv.empty_example);
```
JohnLangford: This should be avoidable?

vowpalwabbit/search.cc
```diff
@@ -1339,7 +1333,7 @@ void generate_training_example(search_private& priv, polylabel& losses, float we
       int learner = select_learner(priv, priv.current_policy, priv.learn_learner_id, true, is_local > 0);
       ec.in_use = true;
       cdbg << "BEGIN base_learner->learn(ec, " << learner << ")" << endl;
-      priv.base_learner->learn(ec, learner);
+      base_learn_or_predict<true>(*priv.base_learner, &ec, learner);
```
JohnLangford: Use base_learner->learn() instead?

vowpalwabbit/search.cc
```diff
@@ -1376,17 +1373,32 @@ void generate_training_example(search_private& priv, polylabel& losses, float we
         lab.costs[0].x = losses.cs.costs[a-start_K].x;
         //cerr << "cost[" << a << "] = " << losses[a] << " - " << min_loss << " = " << lab.costs[0].x << endl;
         ec.in_use = true;
-        uint64_t old_offset = ec.ft_offset;
+        
+        // store the offset to restore it later
+        tmp_offsets.push_back(ec.ft_offset);
```
JohnLangford: There is a vectorization of things that previously were not vectors here that I don't understand yet.

vowpalwabbit/search.cc
```diff
@@ -2088,29 +2100,56 @@ void train_single_example(search& sch, bool is_test_ex, bool is_holdout_ex)
...
 
   bool is_test_ex = false;
   bool is_holdout_ex = false;
-  for (size_t i=0; i<priv.ec_seq.size(); i++)
+
+  search_private& priv = *sch.priv;
+  priv.offset = ec_seq.last()->ft_offset;
+  priv.base_learner = &base;
+
+  example* empty_ex = CSOAA::remove_empty_last_example(ec_seq);
```
JohnLangford: This is the kind of issues that double punctuation creates...

JohnLangford:
@arielf I discussed singleline with @hal3 and he's under the impression this code path is unused in practice.  It's an odd one: multiline examples at training time but singleline examples at test time. 

I finished my first pass reviewing things.  Overall, this seems like good progress (thank @rajan-chari ), but there is some additional code consolidation which is desirable if you can manage it.  For this refactoring, I was hoping that the number of lines of code would actually decrease and I suspect that's possible as the calling pattern is further consolidated to be the same regardless of whether multline or single line examples are in use.

cs/cs/VowpalWabbitMultiLineExampleCollection.cs
rajan-chari: Empty line was used to trigger multiline prediction in reductions before.  Not needed to trigger multiline learn/predict in reductions anymore.  However some reductions used to add the emptyline before processing.  I left this here to be backward compatible.

cs/test/RunTestsHelper.cs
rajan-chari: Not needed.  I added a copy to isolate and test some functionality.

cs/cs/VowpalWabbitMultiLineExampleCollection.cs
rajan-chari: Added by resharper.  Not needed.

vowpalwabbit/cb_adf.cc
rajan-chari: Yes. removed.

vowpalwabbit/cb_explore_adf.cc
rajan-chari: Yes. removed.

vowpalwabbit/csoaa.cc
rajan-chari: Done.

vowpalwabbit/example.h
rajan-chari: I experimented with both during development.  

There are pros for v_array.  I can create a sub array without any allocation.   Copy and construction are really cheap.  There are parts of code that pass example collections by value instead of by reference.  This is inexpensive for v_array.  From a performance perspective this is great.

vector<> has other pros.  for example, it has exception safety on modify operations.  Conservative allocation / de-allocation.  Authors of reductions who don't know VW might be less familiar with v_array and therefore more prone to making allocation/de-allocation errors.    

I don't have a clear winner for which is better.

I tried to future proof it behind a somewhat leaky abstraction (typedef) that we can update in the future.  We may want to another type entirely.  For example, we may want to abstract the underlying type to support shared features more directly.

vowpalwabbit/learner.cc
rajan-chari: Make sense to me.  I thought about doing that here but my worry here is that we might break existing VW users by throwing exception here.  Happy to change the existing behavior so let me know.

MelanieFreed:
@rajan-chari Will these changes make it possible to use the --cb_explore_adf mode via the python wrapper?  (...from what I can tell, this doesn't seem to be possible at the moment from the master branch ...is that right?)

JohnLangford:
I believe the answer is 'yes'.  Rajan's patch is making multiline examples
first-class citizens internally.

-John

MelanieFreed:
Oh, that's amazing, thanks @JohnLangford for letting me know, can't wait for the update!  :)

vowpalwabbit/explore_eval.cc
rajan-chari: done

library/libsearch.h
rajan-chari: Response further down

vowpalwabbit/cb_explore_adf.cc
rajan-chari: yep.  took it out

vowpalwabbit/cb_explore_adf.cc
rajan-chari: Changed.

vowpalwabbit/gen_cs_example.h
rajan-chari: Changed to pass by reference

vowpalwabbit/cb_explore_adf.cc
rajan-chari: done

cs/cli/vowpalwabbit.h
```diff
@@ -40,6 +40,10 @@ public ref class VowpalWabbit : VowpalWabbitBase, IVowpalWabbitExamplePool
   template<typename T>
   cli::array<cli::array<float>^>^ FillTopicAllocation(T& weights);
 
+  // Cached empty example.  
+  // This is used to emit empty lines to cache while handling multiline examples
+  void CacheEmptyLine();
```
eisber: should this be private? Not I'm even clear on when to call this and it seems Learn/Predict(multi line) takes care of it.

Also please use c++/cli comment style

cs/cs/VowpalWabbitMultiLine.cs
```diff
@@ -274,10 +273,8 @@ public static class VowpalWabbitMultiLine
                 actionDependentFeatures,
                 (examples, _, __) =>
                 {
-                    foreach (var ex in examples)
-                    {
-                        vw.Learn(ex);
-                    }
+                    var ec_col = examples.ToList();
```
eisber: either collapse to single line or use camel case variable names.

test/train-sets/ref/cb_adf_crash2.stderr
```diff
@@ -8,11 +8,11 @@ Reading datafile = train-sets/cb_adf_crash_2.data
...
-    n.a.     n.a.            1            1.0  unknown        0:1...        3
+    n.a.     n.a.            1            1.0  unknown        0:1...        2
 
 finished run
 number of examples = 1
 weighted example sum = 1.000000
 weighted label sum = 0.000000
 average loss = n.a.
-total feature number = 3
+total feature number = 2
```
eisber: why is the number of features changing? I assume you removed an empty line?

vowpalwabbit/cb_explore.cc
```diff
@@ -353,10 +353,10 @@ base_learner* cb_explore_setup(arguments& arg)
   base_learner* base = setup_base(arg);
   data->cbcs.scorer = arg.all->scorer;
 
-  learner<cb_explore>* l;
+  learner<cb_explore,example>* l;
   if (arg.vm.count("cover"))
   {
-    data->cs = arg.all->cost_sensitive;
+    data->cs = decltype(data->cs)(arg.all->cost_sensitive);
```
eisber: why do you need the decltype?

java/src/main/c++/jni_base_learner.h
```diff
@@ -64,18 +64,29 @@ T base_predict(
   int example_count = env->GetArrayLength(example_strings);
 
   // When doing multiline prediction the final result is stored in the FIRST example parsed.
+  multi_ex ex_coll = v_init<example*>();
```
JohnLangford: Is this a memory leak?

java/src/main/c++/jni_base_learner.h
```diff
@@ -26,14 +26,14 @@ T base_predict(
 { T result = 0;
   try
   { if (learn)
-      vwInstance->l->learn(*ex);
+      as_singleline(vwInstance->l)->learn(*ex);
```
JohnLangford: Why do we need to leak type information?   Couldn't the top-level learn() know the type?

python/pylibvw.cc
```diff
@@ -168,29 +168,51 @@ void my_finish_example(vw_ptr all, example_ptr ec)
...
 }
 
 float my_predict(vw_ptr all, example_ptr ec)
-{ all->l->predict(*ec);
+{ as_singleline(all->l)->predict(*ec);
   return ec->partial_prediction;
 }
 
+template<bool learn>
+void predict_or_learn(vw_ptr& all, py::list& ec)
```
JohnLangford: This is nice.

python/tests/test_pyvw.py
```diff
@@ -81,18 +81,22 @@ def test_multiclass_prediction_type():
 
 def test_prob_prediction_type():
     model = vw(loss_function='logistic', csoaa_ldf='mc', probabilities=True, quiet=True)
-    model.learn('1 | a b c')
+    multi_ex = [model.example('1:0.2 | a b c'), model.example('2:0.8  | a b c')]
```
JohnLangford: But, we don't want to lose the ability to feed in single line examples?

test/train-sets/ref/cb_adf_crash1.stderr
```diff
@@ -8,11 +8,11 @@ Reading datafile = train-sets/cb_adf_crash_1.data
 num sources = 1
 average  since         example        example  current  current  current
 loss     last          counter         weight    label  predict features
--14.624999 -14.624999            1            1.0    known        0:0.975...        5
+-14.624999 -14.624999            1            1.0    known        0:0.975...        4
```
JohnLangford: What's the explanation for the disparity?

vowpalwabbit/OjaNewton.cc
```diff
@@ -578,7 +578,7 @@ base_learner* OjaNewton_setup(arguments& arg)
 
   arg.all->weights.stride_shift((uint32_t)ceil(log2(ON->m + 2)));
 
-  learner<OjaNewton>& l = init_learner(ON, learn, predict, arg.all->weights.stride());
+  learner<OjaNewton, example>& l = init_learner(ON, learn, predict, arg.all->weights.stride());
```
JohnLangford: Does carrying the type information have value after init_learner?  Basically, I'm wondering if type erasure should happen inside of init_learner().

vowpalwabbit/active_cover.cc
```diff
@@ -92,7 +92,7 @@ float query_decision(active_cover& a, base_learner& l, example& ec, float predic
 
   for(size_t i = 0; i < a.cover_size; i++)
   {
-    l.predict(ec,i+1);
+    l.predict(ec,(int32_t)i+1);
```
JohnLangford: negative numbers don't make sense so this should be a positive type.

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -358,48 +343,34 @@ void output_example(vw& all, cb_explore_adf& c, example& ec, v_array<example*>*
...
-{
-  if (data.ec_seq.size() > 0)
-    for (example* ecc : data.ec_seq)
-      if (ecc->in_use)
-        VW::finish_example(all, ecc);
-  data.ec_seq.erase();
-}
-
-void finish_multiline_example(vw& all, cb_explore_adf& data, example& ec)
+void finish_multiline_example(vw& all, cb_explore_adf& data, multi_ex& ec_seq)
```
JohnLangford: In the long run, it would be good to unify finish_multiline_example and finish_example.   That seems beyond the scope at the moment.

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -575,10 +514,14 @@ base_learner* cb_explore_adf_setup(arguments& arg)
...
+  learner<cb_explore_adf,multi_ex>& l = init_learner(data, base,
+    CB_EXPLORE_ADF::do_actual_learning<true>,
+    CB_EXPLORE_ADF::do_actual_learning<false>,
+    problem_multiplier, 
+    prediction_type::action_probs);
 
   l.set_finish_example(CB_EXPLORE_ADF::finish_multiline_example);
   l.set_finish(CB_EXPLORE_ADF::finish);
-  l.set_end_examples(CB_EXPLORE_ADF::end_examples);
+  l.set_test_example(CB::example_is_test);
```
JohnLangford: Why is this needed?

vowpalwabbit/csoaa.cc
```diff
@@ -12,6 +12,9 @@ license as described in the file LICENSE.
 #include "vw.h"
 #include "gd.h" // GD::foreach_feature() needed in subtract_example()
 #include "vw_exception.h"
+#include <algorithm>
+#include "csoaa.h"
+#include "gen_cs_example.h"
```
JohnLangford: This dependency is odd.   It's defined for the cb reductions.

vowpalwabbit/csoaa.cc
```diff
@@ -398,51 +400,60 @@ void do_actual_learning_oaa(ldf& data, base_learner& base, size_t start_K)
...
-  if (data.ec_seq.size() <= 0) return;  // nothing to do
-  /////////////////////// handle label definitions
+  if (ec_seq_all.size() == 0) return;  // nothing to do
 
-  if (ec_seq_is_label_definition(data.ec_seq))
-  {
-    for (size_t i=0; i<data.ec_seq.size(); i++)
-    {
-      features new_fs = data.ec_seq[i]->feature_space[data.ec_seq[i]->indices[0]];
+  data.ft_offset = ec_seq_all.last()->ft_offset;
```
JohnLangford: The first element is canonical for the prediction, so maybe we should make that the rule?

vowpalwabbit/csoaa.cc
```diff
@@ -859,24 +791,21 @@ base_learner* csldf_setup(arguments& arg)
...
     {
-      ld->treat_as_classifier = false;
-      ld->is_singleline = true;
+      THROW("ldf requires either m/multiline or mc/multiline-classifier.  s/sc/singleline/singleline-classifier is no longer supported");
     }
     else if (ldf_arg.compare("singleline-classifier") == 0 || ldf_arg.compare("sc") == 0)
     {
-      ld->treat_as_classifier = true;
-      ld->is_singleline = true;
+      THROW("ldf requires either m/multiline or mc/multiline-classifier.  s/sc/singleline/singleline-classifier is no longer supported");
```
JohnLangford: Can be joined with previous.

vowpalwabbit/csoaa.cc
```diff
@@ -904,18 +833,61 @@ base_learner* csldf_setup(arguments& arg)
...
-  if (bare->is_singleline)
-    l.set_finish_example(finish_singleline_example);
-  else
-    l.set_finish_example(finish_multiline_example);
+  learner<ldf,multi_ex>& l = init_learner(ld, setup_base(arg), do_actual_learning<true>, do_actual_learning<false>, 1, pred_type);
+  l.set_finish_example(finish_multiline_example);
   l.set_finish(finish);
-  l.set_end_examples(end_examples);
   l.set_end_pass(end_pass);
+  l.set_test_example(COST_SENSITIVE::example_is_test);
```
JohnLangford: I think I know where this is coming from, and there is a better solution in associating this function with the label type.

vowpalwabbit/csoaa.cc
```diff
@@ -904,18 +833,61 @@ base_learner* csldf_setup(arguments& arg)
...
   arg.all->cost_sensitive = make_base(l);
   return arg.all->cost_sensitive;
 }
 
+/*
```
JohnLangford: Shift up so setup_example() is at bottom of file (in canonical position).

vowpalwabbit/example.h
JohnLangford: My feeling here is that we should generally reserve v_array for instances where it is useful beyond vector since vector is more standard.

vowpalwabbit/expreplay.h
```diff
@@ -92,8 +92,8 @@ LEARNER::base_learner* expreplay_setup(arguments& arg)
   if (! arg.all->quiet)
     std::cerr << "experience replay level=" << er_level << ", buffer=" << er->N << ", replay count=" << er->replay_count << std::endl;
 
-  er->base = setup_base(arg);
-  LEARNER::learner<expreplay>* l = &init_learner(er, er->base, predict_or_learn<true,lp>, predict_or_learn<false,lp>);
+  er->base = LEARNER::as_singleline(setup_base(arg));
```
JohnLangford: Maybe just overload setup_base instead?

vowpalwabbit/gen_cs_example.h
```diff
@@ -245,33 +245,31 @@ void call_cs_ldf(LEARNER::base_learner& base, v_array<example*>& examples, v_arr
   }
 
   // 1st: save cb_label (into mydata) and store cs_label for each example, which will be passed into base.learn.
+  // also save offsets
+  std::vector<uint64_t> offsets(examples.size());
```
JohnLangford: Why are we doing allocation on the inner loop?

vowpalwabbit/global_data.cc
```diff
@@ -138,12 +138,32 @@ void set_mm(shared_data* sd, float label)
...
 
-void vw::learn(example* ec)
+void vw::learn(example& ec)
 {
-  if (ec->test_only || !training)
-    l->predict(*ec);
-  else
-    l->learn(*ec);
+  if (ec.test_only || !training){
+    if (l->is_multiline)
```
JohnLangford: Test once.

vowpalwabbit/learner.cc
```diff
@@ -33,33 +33,140 @@ const char* to_string(prediction_type_t prediction_type)
 
 namespace LEARNER
 {
+
+void dispatch_end_pass(vw& all, example* ec)
+{
```
JohnLangford: Ref instead of pointer?

java/src/main/c++/jni_base_learner.h
JohnLangford: Todo: Overload VW::learn to take either multiline or singleline examples and check type agreement dynamically, then replace with a call to VW::learn.

java/src/main/c++/jni_base_learner.h
JohnLangford: Todo: fix

python/tests/test_pyvw.py
JohnLangford: Not an issue.

test/train-sets/ref/cb_adf_crash1.stderr
JohnLangford: Change in multiline semantics to not have an emptyline internally.

test/train-sets/ref/cb_adf_crash2.stderr
JohnLangford: As above.

vowpalwabbit/cb_explore.cc
JohnLangford: arg.all->cost_sensitive erases multilabel vs. singlelabel type information so data->cs needs to recover it.  

TODO: Use type addition function which inverts setup_base to safely add type information.

vowpalwabbit/cb_explore_adf.cc
JohnLangford: Todo: sort out so is_test is a part of the label type.

vowpalwabbit/csoaa.cc
JohnLangford: Todo: remove?

vowpalwabbit/csoaa.cc
JohnLangford: Todo: shift to first.

vowpalwabbit/csoaa.cc
JohnLangford: Todo: minor syntax.

vowpalwabbit/csoaa.cc
JohnLangford: Todo: minor

vowpalwabbit/expreplay.h
JohnLangford: Todo: typecheck the base type against the single/multiline type expected by predict_or_learn.

vowpalwabbit/gen_cs_example.h
JohnLangford: Todo: just use first offset

vowpalwabbit/global_data.cc
JohnLangford: todo: minor syntax

vowpalwabbit/learner.cc
JohnLangford: todo: minor syntax

vowpalwabbit/learner.cc
JohnLangford: Todo: review

vowpalwabbit/learner.h
```diff
@@ -80,21 +93,18 @@ void generic_driver_onethread(vw& all);
 
 inline void noop_sl(void*, io_buf&, bool, bool) {}
 inline void noop(void*) {}
-inline float noop_sensitivity(void*, base_learner&, example&) { return 0.; }
+inline float noop_sensitivity(void*, base_learner&, void*) { return 0.; }
```
JohnLangford: The void* is kind of scary here.

vowpalwabbit/learner.h
```diff
@@ -80,21 +93,18 @@ void generic_driver_onethread(vw& all);
...
 
-typedef void (*tlearn)(void* d, base_learner& base, example& ec);
-typedef float (*tsensitivity)(void* d, base_learner& base, example& ec);
-typedef void (*tmultipredict)(void* d, base_learner& base, example& ec, size_t, size_t, polyprediction*, bool);
-typedef void (*tsl)(void* d, io_buf& io, bool read, bool text);
-typedef void (*tfunc)(void*d);
-typedef void (*tend_example)(vw& all, void* d, example& ec);
+inline void adjust_offset(example& ex, const size_t increment, const int32_t i)
```
JohnLangford: int32_t is the wrong type?

vowpalwabbit/learner.h
```diff
@@ -105,79 +115,98 @@ typedef void (*tend_example)(vw& all, void* d, example& ec);
...
-  inline void learn(example& ec, size_t i=0)
-  { ec.ft_offset += (uint32_t)(increment*i);
-    learn_fd.learn_f(learn_fd.data, *learn_fd.base, ec);
-    ec.ft_offset -= (uint32_t)(increment*i);
+  inline void learn(E& ec, int32_t i = 0)
+  { assert((is_multiline && std::is_same<multi_ex, E>::value) ||
+      (!is_multiline && std::is_same<example, E>::value));  // sanity check under debug compile
+    adjust_offset(ec, increment, i);
+    learn_fd.learn_f(learn_fd.data, *learn_fd.base, (void*)&ec);
+    adjust_offset(ec, increment, -i);
```
JohnLangford: This is why you are doing int32_t.  It seems ok to use internally, but not in the external types.

vowpalwabbit/learner.h
```diff
@@ -191,139 +220,280 @@ typedef void (*tend_example)(vw& all, void* d, example& ec);
...
+  { finish_example_fd.finish_example_f(all, finish_example_fd.data, (void*)&ec);}
+  //called after learn example for each example.  Explicitly not recursive.
+  void set_finish_example(void (*f)(vw& all, T&, E&))
   { finish_example_fd.data = learn_fd.data;
-    finish_example_fd.finish_example_f = (tend_example)f;
+    finish_example_fd.finish_example_f = (end_fptr_type)(f);
+  }
+  
+  //called to check if an example is a test example.  Used in creating multiline example
+  inline bool is_test_example(example& ec)
```
JohnLangford: Again, this belongs to the label type not the learner.

cs/cli/vowpalwabbit.h
rajan-chari: This is used to emit empty lines to cache while handling multiline examples.  Used internally by Learn(IEnumerable<String> lines)

rajan-chari: Changed comment style

vowpalwabbit/search.cc
```diff
@@ -1027,7 +1016,8 @@ action single_prediction_notLDF(search_private& priv, example& ec, int policy, c
 
   cdbg << "allowed_actions_cnt=" << allowed_actions_cnt << ", ec.l = ["; for (size_t i=0; i<ec.l.cs.costs.size(); i++) cdbg << ' ' << ec.l.cs.costs[i].class_index << ':' << ec.l.cs.costs[i].x; cdbg << " ]" << endl;
 
-  priv.base_learner->predict(ec, policy);
+  as_singleline(priv.base_learner)->predict(ec, policy);
```
JohnLangford: Why can't we store whether or not something is singleline in the base_learner?

cs/cs/VowpalWabbitMultiLine.cs
rajan-chari: Collapsed to single line

cs/cs/VowpalWabbitMultiLineExampleCollection.cs
```diff
@@ -70,17 +70,19 @@ private TPrediction Execute<TPrediction>(VowpalWabbit vw, Action<VowpalWabbitExa
             VowpalWabbitExample empty = null;
             try
             {
+                var ec_col = new List<VowpalWabbitExample>();
```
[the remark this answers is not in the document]
rajan-chari: Changed.

vowpalwabbit/search.cc
```diff
@@ -1363,30 +1352,36 @@ void generate_training_example(search_private& priv, polylabel& losses, float we
     {
       int learner = select_learner(priv, priv.current_policy, priv.learn_learner_id, true, is_local > 0);
 
-      for (action a= (uint32_t)start_K; a<priv.learn_ec_ref_cnt; a++)
-      {
-        example& ec = priv.learn_ec_ref[a];
+      // create an example collection for 
 
+      multi_ex tmp = v_init<example*>();
+      vector<uint64_t> tmp_offsets;
```
JohnLangford: Likely eliminatable.

test/train-sets/ref/cb_adf_crash2.stderr
rajan-chari: Yes.  This is coming from the empty line.

vowpalwabbit/search.cc
JohnLangford: todo:  also deal with memory leak.

vowpalwabbit/search.cc
```diff
@@ -1162,15 +1152,14 @@ action single_prediction_LDF(search_private& priv, example* ecs, size_t ec_cnt,
 
     polylabel old_label = ecs[a].l;
     ecs[a].l.cs = priv.ldf_test_label;
+
+    multi_ex tmp = v_init<example*>();
```
JohnLangford: memory leak?

eisber:
Can you add a comment there?

vowpalwabbit/search.cc
JohnLangford: todo: shift earlier.

vowpalwabbit/learner.h
```diff
@@ -191,139 +220,280 @@ typedef void (*tend_example)(vw& all, void* d, example& ec);
...
+    dat.release();
+    l.set_finish_example(MULTICLASS::finish_example<T>);
+    p->lp = MULTICLASS::mc_label;
+    return l;
+  }
+
+  template<class T, class E> base_learner* make_base(learner<T,E>& base) { return (base_learner*)(&base); }
+
+  template<class T, class E> 
+  multi_learner* as_multiline(learner<T, E>* l)
```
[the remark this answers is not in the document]
JohnLangford: Todo

vowpalwabbit/learner.h
```diff
@@ -191,139 +220,280 @@ typedef void (*tend_example)(vw& all, void* d, example& ec);
...
+                                                                  void(*predict)(T&, L&, E&))
+  { return learner<T, E>::init_learner(
+                                          nullptr, base, 
+                                          (typename learner<T, E>::fn) learn,
+                                          (typename learner<T, E>::fn) predict,
+                                          1, base->pred_type);
+  }
+
+  //multiclass reduction
+  template<class T, class E, class L> learner<T, E>& init_multiclass_learner(
```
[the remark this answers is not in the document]
JohnLangford: Do we have a minimal set of init_learner?

vowpalwabbit/learner.h
JohnLangford: ok to use internally

vowpalwabbit/learner.h
JohnLangford: todo: change type?

java/src/main/c++/jni_base_learner.h
rajan-chari: Done.

java/src/main/c++/jni_base_learner.h
rajan-chari: Fixed

vowpalwabbit/csoaa.cc
rajan-chari: Done

vowpalwabbit/csoaa.cc
rajan-chari: Done

vowpalwabbit/csoaa.cc
rajan-chari: Defer to John's upcoming change.

vowpalwabbit/csoaa.cc
rajan-chari: Not sure what this means.  I don't see a setup_example() in the code.

vowpalwabbit/cb_explore_adf.cc
rajan-chari: Need to understand this more.  Can work on this in future refactoring.

vowpalwabbit/explore_eval.cc
rajan-chari: Defered to John's upcoming change.

vowpalwabbit/example.h
rajan-chari: Makes sense.  Will keep that in mind for future refactoring.

vowpalwabbit/cb_explore_adf.cc
rajan-chari: Defered to John's upcoming change

vowpalwabbit/expreplay.h
rajan-chari: 1) Added typecheck for base type against single/multiline type expected by predict_or_learn.
2) Cannot overload setup_base because it only differs by return type.

vowpalwabbit/learner.h
rajan-chari: Defered to John's upcoming change.

vowpalwabbit/search.cc
rajan-chari: In search priv.base_learner can be either a single-line or a multi-line depending on command line arguments.

vowpalwabbit/search.cc
rajan-chari: 1) eliminated tmp_offsets v_array
2) fixed memory leak

vowpalwabbit/search.cc
rajan-chari: fixed.

vowpalwabbit/OjaNewton.cc
rajan-chari: Fixed:  set_* functions use E from class template when example type is needed.

vowpalwabbit/learner.cc
rajan-chari: In the past multiline reductions branch on the following test to accumulate or do_actual_learning()

inline bool example_is_newline_not_header(example& ec) { return (example_is_newline(ec) && !ec_is_example_header(ec));}

For example: https://github.com/JohnLangford/vowpal_wabbit/blob/abd3561a1759eed9ab5f18337d98e6844e2799ae/vowpalwabbit/cb_adf.cc#L350

I use the same test in driver to determine end of multiline example.

vowpalwabbit/global_data.h
rajan-chari: Done

vowpalwabbit/cb_explore.cc
rajan-chari: Changed.
1) Safe cast with type check to singleline learner.
2) explicit type instead of decltype for clarity.

vowpalwabbit/csoaa.cc
rajan-chari: Done.

vowpalwabbit/learner.h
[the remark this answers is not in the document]
rajan-chari: Removed 2 out of 9 init_learners() that are no longer used.

JohnLangford:
Closing for #1467

PR #1422, opened 2018-03-07
Denik remove parse thread

JohnLangford:


JohnLangford:
This was originally from @denik .   I like the idea of being able to run without threads, but I'm a little bit hesitant with this patch at present because it repeats of a bunch of lines of code from parser.cc.   Refactoring things into a template seems like the plausibly-right approach here.

JohnLangford:
@rajan-chari maybe take a look?

JohnLangford:
I've refactored---this is in near-final form now.

PR #1425, opened 2018-03-09
Refactoring for slim scoring

eisber:


vowpalwabbit/example_predict.h
```diff
@@ -0,0 +1,55 @@
...
+  v_array<namespace_index> indices;
+  features feature_space[256]; //Groups of feature values.
+  uint64_t ft_offset;//An offset for all feature values.
+
+  iterator begin() { return iterator(feature_space, indices.begin()); }
+  iterator end() { return iterator(feature_space, indices.end()); }
+};
+
+// make sure we have an exception safe version of example_predict
+class ezexample_predict : public example_predict
```
JohnLangford: ezexample is a name already taken by some of Hal's code.  For confusion avoidance reasons, we should probably pick another name...

JohnLangford:
This looks good.  It's a little bit more fine-grained than I was expecting but that's not an issue.   Let's address the comments and merge.

vowpalwabbit/example_predict.h
eisber: safe_example_predict? it's just adding ctor/destructor to make sure there are no memory leaks.

JohnLangford:
Sounds fine.

PR #1428, opened 2018-03-16
improve model portability by fixing version field size to int32_t

eisber:


JohnLangford:
These don't actually affect portability because the model file format is a string.  Nevertheless, I'm inclined to accept just because it makes things more precise.

PR #1432, opened 2018-03-21
Sampling and exploration refactoring

eisber:


vowpalwabbit/Makefile
```diff
@@ -1,4 +1,4 @@
-BINARIES = vw active_interactor
+BINARIES = vw vwpredict active_interactor
```
JohnLangford: Namewise, maybe vw_slim_predict?  We want to be clear that this does not replace vw -t in general.

Organizationally, maybe we should have the specialized parts in a subdirectory?

eisber: that far it's only vw_predict.* files. should I move to ../vw_slim_predict folder?

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -146,11 +150,13 @@ void predict_or_learn_greedy(cb_explore_adf& data, base_learner& base, v_array<e
...
 
   v_array<action_score>& preds = examples[0]->pred.a_s;
-  uint32_t num_actions = (uint32_t)preds.size();
-  float prob = data.epsilon/(float)num_actions;
-  for (size_t i = 0; i < num_actions; i++)
-    preds[i].score = prob;
-  preds[0].score += 1.f - data.epsilon;
+
+  // generate distribution over actions
+  vector<float> pdf(preds.size());
```
JohnLangford: Move this variable into cb_explore_adf?  This avoids both redeclaration across multiple reductions and stack allocation in the reduction.

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -171,7 +177,8 @@ void predict_or_learn_bag(cb_explore_adf& data, base_learner& base, v_array<exam
   data.action_probs.erase();
   for (uint32_t i = 0; i < num_actions; i++)
     data.action_probs.push_back({ i,0. });
-  float prob = 1.f / (float)data.bag_size;
+  vector<uint32_t> top_actions(num_actions);
```
JohnLangford: declare in cb_explore_adf instead?

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -185,12 +192,19 @@ void predict_or_learn_bag(cb_explore_adf& data, base_learner& base, v_array<exam
...
         multiline_learn_or_predict<true>(base, examples, data.offset, i);
   }
 
+  // generate distribution over actions
+  vector<float> pdf(num_actions);
+  bag(&top_actions[0], &pdf[0], num_actions);
+  for (uint32_t i = 0; i < num_actions; i++)
+    data.action_probs[i].score = pdf[i];
+
+  // TODO: use exploration::safety
```
JohnLangford: Should the TODO be done?

vowpalwabbit/cbify.cc
```diff
@@ -75,11 +51,6 @@ template<class T> inline void delete_it(T* p) { if (p != nullptr) delete p; }
 void finish(cbify& data)
 {
   CB::cb_label.delete_label(&data.cb_label);
-  //data.probs.delete_v();
-  delete_it(data.scorer);
-  delete_it(data.generic_explorer);
-  delete_it(data.mwt_explorer);
-  delete_it(data.recorder);
```
JohnLangford: The deleting of all this stuff is great.  It says we have a better interface now.

vowpalwabbit/cbify.cc
```diff
@@ -126,6 +97,14 @@ void copy_example_to_adf(cbify& data, example& ec)
   }
 }
 
+vector<float> a_s_get_scores(ACTION_SCORE::action_scores& a_s)
+{
+  vector<float> scores(a_s.size());
```
JohnLangford: Can we shift the allocation into cb_explore_adf?

vowpalwabbit/cbify.cc
```diff
@@ -139,11 +118,13 @@ void predict_or_learn(cbify& data, base_learner& base, example& ec)
...
-  uint32_t action = data.mwt_explorer->Choose_Action(*data.generic_explorer, StringUtils::to_string(data.example_counter++), ec);
+  vector<float> pdf = a_s_get_scores(ec.pred.a_s);
+  sample s = sample_from_pdf(data.app_seed + data.example_counter++, &pdf[0], (uint32_t)pdf.size());
 
   CB::cb_class cl;
-  cl.action = action;
-  cl.probability = ec.pred.a_s[action-1].score;
+  cl.action = s.index + 1;
+  // TODO: if pdf is not normalized, this prob is off. probably better to use s.probability
+  cl.probability = ec.pred.a_s[s.index].score;
```
JohnLangford: Yes, we should use the normalized version of probability.

vowpalwabbit/exploration.cc
```diff
@@ -0,0 +1,121 @@
+#include "exploration.h"
+
+#include <stdexcept>
+#include <algorithm>
+#include <numeric>
+#include <cstring>
+
+using namespace std;
+
+namespace exploration
```
JohnLangford: Structurally, maybe this should be in a .h file?  That makes it easier to use in C++.  

Then, when we want to bind across languages, we can just have a dummy .cc file.

vowpalwabbit/exploration.cc
```diff
@@ -0,0 +1,121 @@
...
+{
+  void epsilon_greedy(float epsilon, uint32_t top_action, float* probability_distribution, uint32_t num_actions)
```
JohnLangford: We wanted to create a modestly more convenient C++ interface on top of the C interface like:
void epsilon_greedy(float epsilon, uint32_t top_action, vector<float> probability_distribution);
Can you add that?  (the above calling the C interface)

vowpalwabbit/exploration.cc
```diff
@@ -0,0 +1,121 @@
...
+    for (float *d = probability_distribution; d != prob_end; ++d, ++s)
+    {
+      float prob = exp(lambda*(*s - max_score));
+      norm += prob;
+
+      *d = prob;
+    }
+
+    // normalize
+    if (norm > 0)
```
JohnLangford: This if seems strange, because there is no 'else'?  

Is the 'if' even required?  exp necessarily returns a number >= 0 and one number (the max_score number) is guaranteed to be > 0.

vowpalwabbit/exploration.cc
```diff
@@ -0,0 +1,121 @@
...
+      for (float* d = probability_distribution; d != prob_end; ++d)
+        *d /= norm;
+  }
+
+  void bag(const uint32_t* top_actions, float* probability_distribution, uint32_t num_actions)
```
JohnLangford: Add the C++ style interface.

vowpalwabbit/exploration.cc
```diff
@@ -0,0 +1,121 @@
...
+  {
+    if (num_actions == 0)
+      return;
+
+    uint32_t num_models = std::accumulate(top_actions, top_actions + num_actions, 0);
```
JohnLangford: The use of stl primitives is nice.

vowpalwabbit/exploration.cc
```diff
@@ -0,0 +1,121 @@
...
+      probability_distribution[i] = top_actions[i] * prob;
+  }
+
+  void enforce_minimum_probability(float min_prob, float* probability_distribution, uint32_t num_actions)
+  {
+    float* prob_end = probability_distribution + num_actions;
+    bool zeros = false;
+
+    //input: a probability distribution
+    //output: a probability distribution with all events having probability > min_prob.  This includes events with probability 0 if zeros = true
```
JohnLangford: This comment seems obsolete.  

Also, is this code used in vw?

vowpalwabbit/interactions_predict.h
```diff
@@ -91,8 +91,8 @@ namespace INTERACTIONS
   // this templated function generates new features for given example and set of interactions
   // and passes each of them to given function T()
   // it must be in header file to avoid compilation problems
-  template <class R, class S, void(*T)(R&, float, S), bool audit, void(*audit_func)(R&, const audit_strings*), class W> // nullptr func can't be used as template param in old compilers
-  inline void generate_interactions(v_array<v_string>& interactions, bool permutations, example_predict& ec, R& dat, W& weights) // default value removed to eliminate ambiguity in old complers
+  template <class R, class S, void(*T)(R&, float, S), bool audit, void(*audit_func)(R&, const audit_strings*), class W, class I> // nullptr func can't be used as template param in old compilers
+  inline void generate_interactions(I& interactions, bool permutations, example_predict& ec, R& dat, W& weights) // default value removed to eliminate ambiguity in old complers
```
JohnLangford: Why do we need a class I?  It seems to be always a v_array<v_string>?

vowpalwabbit/sampling.cc
```diff
@@ -0,0 +1,96 @@
+#include "hash.h"
```
JohnLangford: It seems like we would need to shift this into a header for a header-only version as well.

Are there any changes in the code below from the old library?

JohnLangford: Ah, it is changed below.

vowpalwabbit/sampling.cc
```diff
@@ -0,0 +1,96 @@
...
+  // uniform random between 0 and 1
+  float uniform_random_merand48(uint64_t initial)
+  {
+    initial = a * initial + c;
+    int_float temp;
+    temp.i = ((initial >> 25) & 0x7FFFFF) | bias;
+    return temp.f - 1;
+  }
+
+  sample sample_from_pdf(const char* seed, const float* pdf, uint32_t len)
```
JohnLangford: It would be nice to support a C++ style interface as well (for this and those below).

vowpalwabbit/sampling.cc
```diff
@@ -0,0 +1,96 @@
...
+        action_index = i;
+        action_probability = pdf[i] / total;
+        break;
+      }
+    }
+
+    return { action_probability, action_index };
+  }
+
+  void sample_from_pdf(const char* seed, const float* pdf, const float* scores, uint32_t len, ranked_sample& result)
```
JohnLangford: This interface seems confused because we are using c-style and C++ style in the same interface leaving us something which is both c++-awkward and does not easily cross language boundaries.  What's the fix?

vowpalwabbit/vw_predict.cc
```diff
@@ -0,0 +1,137 @@
+#include "vw_predict.h"
```
JohnLangford: Perhaps a name change inline with the binary.

vowpalwabbit/vw_predict.cc
```diff
@@ -0,0 +1,137 @@
...
+  mp.skip(3 * ngram_len);
+
+  uint32_t skips_len = mp.read<uint32_t>("skips_len");
+  mp.skip(3 * skips_len);
+
+  mp.read_string("file_options");
+
+  if (mp.read<uint32_t>("check_sum_len") != 4)
+    throw std::invalid_argument("check_sum_len must be 4");
+  mp.read<uint32_t>("check_sum");
```
JohnLangford: Why are we ignoring the checksum?

vowpalwabbit/vw_predict.cc
```diff
@@ -0,0 +1,137 @@
...
+  else
+    mp.read_dense_weights<uint64_t>(_weights, weight_length);
+
+  // check that permutations is not enabled (or parse it)
+}
+
+void vw_predict::score(example_predict* shared, std::vector<example_predict*> actions, std::vector<float>& out_scores)
+{
+  safe_example_predict ex;
+  GD::inline_predict<dense_parameters, vector<string>>(*_weights, false, _ignore_linear, _interactions, /* permutations */ false, ex);
```
JohnLangford: This is where the I parameter varies.  

Have you considered instead changing VW to use vector<string>?

vowpalwabbit/Makefile
JohnLangford: Let's do that, along with the make details.

JohnLangford:
Thinking through things more deeply, I think we want:
(a) A slim_predict subdirectory.
(b) A exploration (sub?)directory.
(c) To nuke the existing exploration library from the codebase.

This leaves things with a clean conceptual organization.

vowpalwabbit/interactions_predict.h
eisber: not for the slim predict. I'm passing vectors to get allocation/deallocation safety in case of exceptions

vowpalwabbit/exploration.cc
```diff
@@ -0,0 +1,121 @@
...
+      throw out_of_range("must supply at least one top_action from a model");
+
+    // determine probability per model
+    float prob = 1.f / (float)num_models;
+
+    for (size_t i = 0;i < num_actions;i++)
+      probability_distribution[i] = top_actions[i] * prob;
+  }
+
+  void enforce_minimum_probability(float min_prob, float* probability_distribution, uint32_t num_actions)
```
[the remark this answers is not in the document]
eisber: this CB*::safety

vowpalwabbit/sampling.cc
eisber: the problem is how to return the ranking. maybe switch to caller allocated as well?

vowpalwabbit/vw_predict.cc
eisber: if you change VW to use vector you loose the v_array efficient memory mgmt (e.g. re-alloc every 1024th time)

vowpalwabbit/exploration.cc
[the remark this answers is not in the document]
JohnLangford: The flags are different though?  

Can we eliminate safety?

vowpalwabbit/sampling.cc
JohnLangford: Perhaps an interface which is just C-style with caller-allocated and the cleaner C++ style with vector& ?  The C-style interface can be used across languages and the C++ style will be more convenient in C++.

vowpalwabbit/vw_predict.cc
JohnLangford: My understanding is that this use of v_array can be easily replaced with vector in vw?  (Trying to change every instance of v_array is a giant mess far beyond the scope of what we want to do here.)

vowpalwabbit/cb_explore_adf.cc
eisber: vector doesn't support the lazy shrink like v_array does. You want me to switch to v_array or just move the vector and reallocate every time (the later is slightly more annoying as I need to deal with explicit ctor/dtor calls)

vowpalwabbit/cb_explore_adf.cc
eisber: same question as above

vowpalwabbit/cbify.cc
eisber: done

vowpalwabbit/exploration.cc
eisber: can move to .cc, why do we need a dummy .cc file?

eisber: done. TODO: dummy .cc unclear

vowpalwabbit/exploration.cc
eisber: done

vowpalwabbit/exploration.cc
eisber: done. I think this was left over from previous code that didn't check if there are any scores...

vowpalwabbit/exploration.cc
eisber: done

vowpalwabbit/vw_predict.cc
eisber: I'd have to re-implement the header hashing too.

vowpalwabbit/cb_explore_adf.cc
JohnLangford: I don't believe we need a lazy shrink here, so you can move the vector to the reduction and just reuse it without reallocation.

vowpalwabbit/exploration.cc
JohnLangford: (Let's deal with binding later)

vowpalwabbit/interactions_predict.h
JohnLangford: So, make it vector everyhwere?

vowpalwabbit/vw_predict.cc
JohnLangford: That's scary----it means that we can't verify the correctness.  

Do you have an alternative to verify correctness?

JohnLangford:
Another detail: it seems important to blow away the existing exploration library and show that this one replicates decisions on the same tests.

eisber:
How do I know the number of actions in the setup function when I need to initialize? How do I deal with changing number of actions?

eisber:
So I should replace v_array in example with vector?

vowpalwabbit/cb_explore_adf.cc
eisber: I just realized that we'll have allocation regardless due to the pdf returned. should I switch to the C-style interface to avoid allocation?

JohnLangford:
(We dicussed: allocation can be avoided by using pass-by-reference or
STL-style iterators.)

vowpalwabbit/cbify.cc
eisber: done

vowpalwabbit/sampling.cc
eisber: switched to pure c-style

vowpalwabbit/vw_predict.cc
eisber: done

vowpalwabbit/Makefile.am
```diff
@@ -4,7 +4,7 @@ liballreduce_la_SOURCES = allreduce_sockets.cc allreduce_threads.cc vw_exception
...
+libvw_la_SOURCES = hash.cc parser_helper.cc global_data.cc io_buf.cc parse_regressor.cc parse_primitives.cc unique_sort.cc cache.cc rand48.cc simple_label.cc multiclass.cc oaa.cc multilabel_oaa.cc boosting.cc ect.cc marginal.cc autolink.cc binary.cc lrq.cc cost_sensitive.cc multilabel.cc label_dictionary.cc csoaa.cc cb.cc cb_adf.cc cb_algs.cc search.cc search_meta.cc search_sequencetask.cc search_dep_parser.cc search_hooktask.cc search_multiclasstask.cc search_entityrelationtask.cc search_graph.cc parse_example.cc scorer.cc network.cc parse_args.cc accumulate.cc gd.cc learner.cc mwt.cc lda_core.cc gd_mf.cc mf.cc bfgs.cc noop.cc print.cc example.cc parser.cc loss_functions.cc sender.cc nn.cc confidence.cc bs.cc cbify.cc explore_eval.cc topk.cc stagewise_poly.cc log_multi.cc recall_tree.cc active.cc active_cover.cc cs_active.cc kernel_svm.cc best_constant.cc ftrl.cc svrg.cc lrqfa.cc interact.cc comp_io.cc interactions.cc vw_exception.cc vw_validate.cc audit_regressor.cc gen_cs_example.cc cb_explore.cc action_score.cc cb_explore_adf.cc OjaNewton.cc parse_example_json.cc baseline.cc classweight.cc sampling.cc exploration.cc
```
JohnLangford: Should we have sampling.cc and exploration.cc?

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -185,12 +192,16 @@ void predict_or_learn_bag(cb_explore_adf& data, base_learner& base, v_array<exam
...
+    top_actions[preds[0].action]++;
     if (is_learn && !test_sequence)
       for (uint32_t j = 1; j < count; j++)
         multiline_learn_or_predict<true>(base, examples, data.offset, i);
   }
 
+  // generate distribution over actions
+  bag(top_actions.begin(), top_actions.end(), begin_scores(data.action_probs), end_scores(data.action_probs));
+
+  // TODO: use exploration::safety
```
JohnLangford: Why do we have two safety functions?

vowpalwabbit/cbify.cc
```diff
@@ -3,32 +3,20 @@
 #include "cb_algs.h"
 #include "rand48.h"
 #include "bs.h"
-#include "../explore/cpp/MWTExplorer.h"
+#include "sampling.h"
 #include "vw.h"
+#include "hash.h"
+#include "sampling.h"
```
JohnLangford: A double include?

vowpalwabbit/cbify.cc
JohnLangford: Does this function need to be cleaned up?

vowpalwabbit/cbify.cc
```diff
@@ -139,11 +118,12 @@ void predict_or_learn(cbify& data, base_learner& base, example& ec)
   base.predict(ec);
   //data.probs = ec.pred.scalars;
 
-  uint32_t action = data.mwt_explorer->Choose_Action(*data.generic_explorer, StringUtils::to_string(data.example_counter++), ec);
+  vector<float> pdf = a_s_get_scores(ec.pred.a_s);
```
JohnLangford: It looks like we still have critical path memory allocation going on.  Remove?

vowpalwabbit/exploration.cc
JohnLangford: Nuke?

vowpalwabbit/exploration.h
```diff
@@ -0,0 +1,123 @@
...
+#include <stdint.h>
+#include <stdexcept>
+#include <algorithm>
+#include <numeric>
+#include <cstring>
+
+namespace exploration
+{
+  template<typename OutputIt>
+  void epsilon_greedy(float epsilon, uint32_t top_action, OutputIt pdf_first, OutputIt pdf_last)
```
JohnLangford: I'm still not fully clear on the portability of this interface.

vowpalwabbit/exploration.h
```diff
@@ -0,0 +1,123 @@
...
+
+    *(pdf_first + top_action) += 1.f - epsilon;
+  }
+
+  template<typename InputIt, typename OutputIt>
+  void softmax(float lambda, InputIt scores_begin, InputIt scores_last, OutputIt pdf_first, OutputIt pdf_last)
+  {
+    if (scores_begin == scores_last)
+      return;
+
```
JohnLangford: What is the correct behavior with pdf has a different length from scores?

vowpalwabbit/exploration.h
```diff
@@ -0,0 +1,123 @@
...
+  template<typename InputIt, typename OutputIt>
+  void bag(InputIt top_actions_begin, InputIt top_actions_last, OutputIt pdf_first, OutputIt pdf_last)
+  {
+    if (pdf_first == pdf_last)
+      return;
+
+    uint32_t num_models = std::accumulate(top_actions_begin, top_actions_last, 0);
+    if (num_models == 0)
+      throw std::out_of_range("must supply at least one top_action from a model");
```
JohnLangford: As an alternative, we could impose a uniform distribution.

vowpalwabbit/exploration.h
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    // normalize
+    for (OutputIt d = pdf_first; d != pdf_last; ++d)
+      *d /= norm;
+  }
+
+  template<typename InputIt, typename OutputIt>
+  void bag(InputIt top_actions_begin, InputIt top_actions_last, OutputIt pdf_first, OutputIt pdf_last)
+  {
```
JohnLangford: We should probably check that top_actions length is pdf length.

vowpalwabbit/exploration.h
```diff
@@ -0,0 +1,123 @@
...
+    // determine probability per model
+    float prob = 1.f / (float)num_models;
+
+    InputIt t_a = top_actions_begin;
+    for (OutputIt d = pdf_first; d != pdf_last && t_a != top_actions_last; ++d, ++t_a)
+      *d = *t_a * prob;
+  }
+
+  // Note: must be inline to compile on Windows VS2017
+  inline void enforce_minimum_probability(float min_prob, float* probability_distribution, uint32_t num_actions)
```
JohnLangford: Is this code getting used?  It's in C-style rather than STL-style.  We should probably stick with one style.

vowpalwabbit/gd.h
```diff
@@ -81,8 +81,8 @@ inline void foreach_feature(vw& all, example& ec, R& dat)
 inline float inline_predict(vw& all, example& ec)
 {
   return all.weights.sparse ?
-    inline_predict<sparse_parameters>(all.weights.sparse_weights, all.ignore_some_linear, all.ignore_linear, all.interactions, all.permutations, ec, ec.l.simple.initial) :
-    inline_predict<dense_parameters>(all.weights.dense_weights, all.ignore_some_linear, all.ignore_linear, all.interactions, all.permutations, ec, ec.l.simple.initial);
+    inline_predict<sparse_parameters, v_array<v_string>>(all.weights.sparse_weights, all.ignore_some_linear, all.ignore_linear, all.interactions, all.permutations, ec, ec.l.simple.initial) :
```
JohnLangford: I'd like to get rid of the extra type parameterization by coalescing between the vw_slim_predict and vw.  Can you modify vw to use vector instead here?

vowpalwabbit/sampling.cc
```diff
@@ -0,0 +1,95 @@
...
+    return temp.f - 1;
+  }
+
+  sample sample_from_pdf(const char* seed, const float* pdf, uint32_t len)
+  {
+    uint64_t seed_hash = uniform_hash(seed, strlen(seed), 0);
+    return sample_from_pdf(seed_hash, pdf, len);
+  }
+
+  sample sample_from_pdf(uint64_t seed, const float* pdf, uint32_t len)
```
JohnLangford: If we want to be consistent, should this also have an STL-style iterator interface?

vowpalwabbit/sampling.cc
```diff
@@ -0,0 +1,95 @@
...
+        action_index = i;
+        action_probability = pdf[i] / total;
+        break;
+      }
+    }
+
+    return { action_probability, action_index };
+  }
+
+  void sample_from_pdf(const char* seed, const float* pdf, const float* scores, uint32_t len, uint32_t* out_ranking, float* out_probability)
```
JohnLangford: The semantics of this is unclear.   Perhaps a comment.   Also, if we are doing return reference, perhaps it's better to use a reference?

vowpalwabbit/sampling.h
```diff
@@ -0,0 +1,26 @@
...
+{
+  struct sample
+  {
+    // if the pdf is not normalized, this is the normalized probability
+    float probability;
+
+    uint32_t index;
+  };
+
+  sample sample_from_pdf(const char* seed, const float* pdf, uint32_t len);
```
JohnLangford: Is it considered bad these days to use self-delimiting character strings?

vowpalwabbit/exploration.cc
eisber: comes back to the other question if we favor header only, with potential include/namespace pollution or .h/.cc files.

vowpalwabbit/exploration.h
eisber: not sure what you mean by portability. Exposing it to other languages through bindings? If yes, neither this or the previous version fit. Based on the path one chooses (e.g. using swig bound vectors in python or plain c-style pointers) you'll have to write slim adapters in C/C++. Those adapters will have an easily live calling this function w/o memory copy.

vowpalwabbit/sampling.cc
eisber: I'd favro the STL-style interface

vowpalwabbit/sampling.cc
eisber: the assumption is that out_* points to pre-allocated memory. What kind of reference are you proposing. we can do float& out_probability. Not clear what you want to do for out_ranking.

vowpalwabbit/sampling.h
eisber: well, a lot of the functions ask for length, but if you don't have it, it can be real pain (as you need to call strlen...). not sure.

vowpalwabbit/exploration.h
```diff
@@ -0,0 +1,123 @@
...
+    for (OutputIt d = pdf_first; d != pdf_last && s != scores_last; ++d, ++s)
+    {
+      float prob = exp(lambda*(*s - max_score));
+      norm += prob;
+
+      *d = prob;
+    }
+
+    // normalize
+    for (OutputIt d = pdf_first; d != pdf_last; ++d)
```
JohnLangford: Prescompute inverse norm.

vowpalwabbit/Makefile.am
eisber: decision was to remove and move to header only

vowpalwabbit/cbify.cc
eisber: done

vowpalwabbit/cbify.cc
eisber: done

vowpalwabbit/interactions_predict.h
eisber: done

vowpalwabbit/sampling.cc
eisber: done

vowpalwabbit/cbify.cc
eisber: done

vowpalwabbit/gd.h
eisber: done

vowpalwabbit/sampling.cc
eisber: done

vowpalwabbit/exploration.h
eisber: reverted to division for numerical stability

vowpalwabbit/cbify.cc
eisber: done

vowpalwabbit/cbify.cc
eisber: done

vowpalwabbit/exploration.h
eisber: not sure. we can have 10 actions, but only 5 models?

vowpalwabbit/gd.h
eisber: done

vowpalwabbit/interactions_predict.h
eisber: done

vowpalwabbit/sampling.cc
eisber: done

explore/exploration.h
```diff
@@ -0,0 +1,152 @@
...
+  void generate_bag(InputIt top_actions_begin, InputIt top_actions_last, OutputIt pdf_first, OutputIt pdf_last)
+  {
+    typedef typename std::iterator_traits<InputIt>::iterator_category top_actions_category;
+    typedef typename std::iterator_traits<OutputIt>::iterator_category pdf_category;
+
+    generate_bag(top_actions_begin, top_actions_last, top_actions_category(), pdf_first, pdf_last, pdf_category());
+  }
+
+  // Note: must be inline to compile on Windows VS2017
+  inline void enforce_minimum_probability(float min_prob, float* probability_distribution, uint32_t num_actions)
```
JohnLangford: This was going to be removed or changed to a STL-style interface, right?

explore/exploration.h
```diff
@@ -0,0 +1,152 @@
...
+#include <stdint.h>
+#include <stdexcept>
+#include <algorithm>
+#include <numeric>
+#include <cstring>
+
+namespace exploration
+{
+  template<typename OutputIt>
+  void generate_epsilon_greedy(float epsilon, uint32_t top_action, OutputIt pdf_first, OutputIt pdf_last, std::output_iterator_tag pdf_tag)
```
JohnLangford: Why is there a pdf_tag argument which is never used?

explore/sampling_internal.h
```diff
@@ -0,0 +1,86 @@
...
+  inline float uniform_random_merand48(uint64_t initial)
+  {
+    initial = a * initial + c;
+    int_float temp;
+    temp.i = ((initial >> 25) & 0x7FFFFF) | bias;
+    return temp.f - 1;
+  }
+
+  template<typename InputIt>
+  uint32_t sample_from_pdf(uint64_t seed, InputIt pdf_first, InputIt pdf_last, std::input_iterator_tag pdf_category)
```
JohnLangford: Why have a pdf_category which is unused?

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -185,12 +192,16 @@ void predict_or_learn_bag(cb_explore_adf& data, base_learner& base, v_array<exam
...
     assert(preds.size() == num_actions);
-    data.action_probs[preds[0].action].score += prob;
+    top_actions[preds[0].action]++;
     if (is_learn && !test_sequence)
       for (uint32_t j = 1; j < count; j++)
         multiline_learn_or_predict<true>(base, examples, data.offset, i);
   }
 
+  // generate distribution over actions
+  generate_bag(begin(top_actions), end(top_actions), begin_scores(data.action_probs), end_scores(data.action_probs));
```
JohnLangford: I didn't know that you can do begin(top_actions).  Is that defined in STL?  Why is this preferred to top_actions.begin()? Or is it just a different style?

explore/exploration.h
```diff
@@ -0,0 +1,180 @@
...
+  template<typename It>
+  void generate_epsilon_greedy(float epsilon, uint32_t top_action, It pdf_first, It pdf_last, std::random_access_iterator_tag pdf_tag)
+  {
+    size_t num_actions = pdf_last - pdf_first;
+    if (num_actions == 0)
+      return;
+
+	if (top_action >= num_actions)
+	#ifdef EXPLORE_NOEXCEPT
+  	  top_action = (uint32_t)num_actions - 1;
```
JohnLangford: Why don't we always do this?

explore/exploration.h
```diff
@@ -0,0 +1,180 @@
...
+  void generate_softmax(float lambda, InputIt scores_begin, InputIt scores_last, std::input_iterator_tag scores_tag, OutputIt pdf_first, OutputIt pdf_last, std::random_access_iterator_tag pdf_tag)
+  {
+    size_t num_actions_scores = scores_last - scores_begin;
+    size_t num_actions_pdf = pdf_last - pdf_first;
+
+	if (num_actions_scores != num_actions_pdf)
+	{
+	#ifdef EXPLORE_NOEXCEPT
+		// fallback to the minimum
+		scores_last = scores_begin + std::min(num_actions_scores, num_actions_pdf);
```
JohnLangford: Why not always do this?

explore/exploration.h
```diff
@@ -0,0 +1,180 @@
...
+  void generate_epsilon_greedy(float epsilon, uint32_t top_action, It pdf_first, It pdf_last)
+  {
+    typedef typename std::iterator_traits<It>::iterator_category pdf_category;
+    generate_epsilon_greedy(epsilon, top_action, pdf_first, pdf_last, pdf_category());
+  }
+
+  template<typename InputIt, typename OutputIt>
+  void generate_softmax(float lambda, InputIt scores_begin, InputIt scores_last, std::input_iterator_tag scores_tag, OutputIt pdf_first, OutputIt pdf_last, std::random_access_iterator_tag pdf_tag)
+  {
+    size_t num_actions_scores = scores_last - scores_begin;
```
JohnLangford: What happens if scores_last < scores_begin?

explore/exploration.h
```diff
@@ -0,0 +1,180 @@
...
+  void generate_bag(InputIt top_actions_begin, InputIt top_actions_last, std::input_iterator_tag top_actions_tag, OutputIt pdf_first, OutputIt pdf_last, std::random_access_iterator_tag pdf_tag)
+  {
+    if (pdf_first == pdf_last)
+      return;
+
+    uint32_t num_models = std::accumulate(top_actions_begin, top_actions_last, 0);
+	if (num_models == 0)
+	{
+	#ifdef EXPLORE_NOEXCEPT
+		*pdf_first = 1;
```
JohnLangford: Maybe always do this?

explore/exploration.h
```diff
@@ -0,0 +1,180 @@
...
+        touched_mass += min_prob;
+        prob = min_prob;
+      }
+      else
+        untouched_mass += prob;
+    }
+
+    if (touched_mass > 0.)
+    {
+	#ifndef EXPLORE_NOEXCEPT
```
JohnLangford: Maybe define an alternate semantics instead?

explore/explore.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+namespace exploration {
```
JohnLangford: This is good.  Maybe add some comments about the meaning of arguments?

eisber:
Not getting an error is kind of weird. This is mainly to handle embedded/IoT scenarios. But I’m fine either way.

eisber:
Do we have standard comments system? Doxygen?

explore/sampling_internal.h
```diff
@@ -0,0 +1,100 @@
...
+  {
+    // Create a discrete_distribution based on the returned weights. This class handles the
+    // case where the sum of the weights is < or > 1, by normalizing agains the sum.
+    float total = 0.f;
+    for (InputIt pdf = pdf_first; pdf != pdf_last; ++pdf)
+    {
+#ifdef EXPLORE_NOEXCEPT
```
JohnLangford: The variable semsantics seems iffy here.  Can we always do the same thing?

explore/exploration.h
```diff
@@ -0,0 +1,180 @@
...
+  {
+    typedef typename std::iterator_traits<InputIt>::iterator_category top_actions_category;
+    typedef typename std::iterator_traits<OutputIt>::iterator_category pdf_category;
+
+    generate_bag(top_actions_begin, top_actions_last, top_actions_category(), pdf_first, pdf_last, pdf_category());
+  }
+
+  // Note: must be inline to compile on Windows VS2017
+  template<typename It>
+  void enforce_minimum_probability(float min_prob, bool update_zero_elements, It pdf_first, It pdf_last, std::random_access_iterator_tag pdf_tag)
```
JohnLangford: I believe this is still not being used in VW.  Why not?  I'd like to either have safety() or enforce_minimum_probability() (but not both) in the codebase.

JohnLangford:
I agree that not getting an error is odd, but two different semantics seems worse.

I was just imagining actual C++ comments.  Nothing long---just enough to give the names enough meaning to be used on reading.

explore/exploration.h
eisber: Since scores_begin is an InputIterator it's not possible to check for scores_last <= scores_begin.

eisber:
If I understand the code correctly, this means that the PDF is not properly normalized. Should everything be re-normalized?
It feels a bit weird that we only re-normalize if just the elements that would be adjusted sum to more than 1.

JohnLangford:
Sorry, I don't know which code this refers to.  If you comment at the line of code on github, that will make it clear what you are referring to.

explore/exploration.h
eisber: If I understand the code correctly, this means that the PDF is not properly normalized. Should everything be re-normalized?
It feels a bit weird that we only re-normalize if just the elements that would be adjusted sum to more than 1.

explore/explore.h
eisber: done

explore/exploration.h
JohnLangford: touched_mass large implies that basically everything is not meeting the minimum probability so the function is being asked to make the lower bound to large.  A reasonable alternate semantics in this case is to simply create a uniform distribution over each entry generating touched_mass.

explore/exploration.h
```diff
@@ -0,0 +1,181 @@
+#pragma once
+
+#include <stdint.h>
+#include <stdexcept>
+#include <algorithm>
+#include <numeric>
+#include <cstring>
+
+namespace exploration
+{
```
JohnLangford: I believe we should have a nonzero return value for 'bad range' and always guarantee that if the return value is zero then the distribution is normalized.

explore/exploration.h
```diff
@@ -0,0 +1,181 @@
...
+  template<typename It>
+  void generate_epsilon_greedy(float epsilon, uint32_t top_action, It pdf_first, It pdf_last, std::random_access_iterator_tag pdf_tag)
+  {
+    if (pdf_last < pdf_first)
+      return;
+
+    size_t num_actions = pdf_last - pdf_first;
+    if (num_actions == 0)
+      return;
```
JohnLangford: The distribution needs to be created in this case.

explore/exploration.h
```diff
@@ -0,0 +1,181 @@
...
+
+    size_t num_actions_scores = scores_last - scores_begin;
+    size_t num_actions_pdf = pdf_last - pdf_first;
+
+    if (num_actions_scores != num_actions_pdf)
+    {
+      // fallback to the minimum
+      scores_last = scores_begin + ((std::min)(num_actions_scores, num_actions_pdf));
+      pdf_last = pdf_first + ((std::min)(num_actions_scores, num_actions_pdf));
+    }
```
JohnLangford: Should zero out distribution beyond the minimum.

explore/exploration.h
```diff
@@ -0,0 +1,181 @@
...
+
+    if (num_actions_scores == 0)
+      return;
```
JohnLangford: Should create normalized distribution.

explore/exploration.h
```diff
@@ -0,0 +1,181 @@
...
+  void generate_bag(InputIt top_actions_begin, InputIt top_actions_last, std::input_iterator_tag top_actions_tag, OutputIt pdf_first, OutputIt pdf_last, std::random_access_iterator_tag pdf_tag)
+  {
+    if (pdf_first == pdf_last || pdf_last < pdf_first)
+      return;
+
+    uint32_t num_models = std::accumulate(top_actions_begin, top_actions_last, 0);
+    if (num_models == 0)
+    {
+      *pdf_first = 1;
+      return;
```
JohnLangford: Should also zero elements.

explore/exploration.h
```diff
@@ -0,0 +1,181 @@
...
+
+    generate_bag(top_actions_begin, top_actions_last, top_actions_category(), pdf_first, pdf_last, pdf_category());
+  }
+
+  template<typename It>
+  void enforce_minimum_probability(float min_prob, bool update_zero_elements, It pdf_first, It pdf_last, std::random_access_iterator_tag pdf_tag)
+  {
+	  size_t num_actions = pdf_last - pdf_first;
+
+    if (min_prob > 0.999) // uniform exploration
```
JohnLangford: The wrong guard.  You want min_prob > 1/num_actions

explore/exploration.h
```diff
@@ -0,0 +1,181 @@
...
+    typedef typename std::iterator_traits<InputIt>::iterator_category top_actions_category;
+    typedef typename std::iterator_traits<OutputIt>::iterator_category pdf_category;
+
+    generate_bag(top_actions_begin, top_actions_last, top_actions_category(), pdf_first, pdf_last, pdf_category());
+  }
+
+  template<typename It>
+  void enforce_minimum_probability(float min_prob, bool update_zero_elements, It pdf_first, It pdf_last, std::random_access_iterator_tag pdf_tag)
+  {
+	  size_t num_actions = pdf_last - pdf_first;
```
JohnLangford: Should check range.

explore/exploration.h
JohnLangford: Make that min_prob > 1 / (number of actions to be updated).

explore/explore.h
JohnLangford: The comments look good here.

explore/sampling_internal.h
```diff
@@ -0,0 +1,87 @@
...
+  template<typename InputIt>
+  uint32_t sample_from_pdf(uint64_t seed, InputIt pdf_first, InputIt pdf_last, std::input_iterator_tag pdf_category)
+  {
+    // Create a discrete_distribution based on the returned weights. This class handles the
+    // case where the sum of the weights is < or > 1, by normalizing agains the sum.
+    float total = 0.f;
+    for (InputIt pdf = pdf_first; pdf != pdf_last; ++pdf)
+    {
+      if (*pdf < 0)
+        *pdf = 0;
```
JohnLangford: Is it valid to do an assignment on an input iterator?

explore/sampling_internal.h
```diff
@@ -0,0 +1,87 @@
...
+      total += *pdf;
+    }
+    if (total == 0)
+    {
+      // assume the first is the best
+      *pdf_first = 1.f;
```
JohnLangford: We should check for a valid range.

explore/exploration.h
```diff
@@ -0,0 +1,181 @@
...
+  {
+    if (pdf_last < pdf_first)
+      return;
+
+    size_t num_actions = pdf_last - pdf_first;
+    if (num_actions == 0)
+      return;
+
+	if (top_action >= num_actions)
+  	  top_action = (uint32_t)num_actions - 1;
```
eisber: should this use the num_actions - 1 fallback or return E_EXPLORATION_BAD_RANGE

explore/exploration.h
eisber: I'm not sure what you mean.

explore/exploration.h
eisber: done

explore/exploration.h
eisber: is 0-sized pdf valid or E_EXPLORATION_BAD_RANGE?

explore/exploration.h
eisber: how should we deal with the return values in cb_explore_adf/cb_explore. Throw a VW exception?

explore/exploration.h
eisber: not sure what you mean?

explore/exploration.h
eisber: done

explore/exploration.h
```diff
@@ -0,0 +1,207 @@
...
+{
+  template<typename It>
+  int generate_epsilon_greedy(float epsilon, uint32_t top_action, It pdf_first, It pdf_last, std::random_access_iterator_tag pdf_tag)
+  {
+    if (pdf_last < pdf_first)
+      return E_EXPLORATION_BAD_RANGE;
+
+    size_t num_actions = pdf_last - pdf_first;
+    if (num_actions == 0)
+      return S_EXPLORATION_OK;
```
JohnLangford: This doesn't seem ok, because you aren't returning a normalized distribution.

explore/exploration.h
```diff
@@ -0,0 +1,207 @@
...
+
+      // zero out pdf
+      for (OutputIt d = pdf_new_last; d != pdf_last; ++d)
+        *d = 0;
+
+        pdf_last = pdf_new_last;
+    }
+
+    if (num_actions_scores == 0)
+      return S_EXPLORATION_OK;
```
JohnLangford: Return is not normalized in this case.

explore/exploration.h
```diff
@@ -0,0 +1,207 @@
...
+    typedef typename std::iterator_traits<OutputIt>::iterator_category pdf_category;
+
+    return generate_softmax(lambda, scores_begin, scores_last, scores_category(), pdf_first, pdf_last, pdf_category());
+  }
+
+  template<typename InputIt, typename OutputIt>
+  int generate_bag(InputIt top_actions_begin, InputIt top_actions_last, std::input_iterator_tag top_actions_tag, OutputIt pdf_first, OutputIt pdf_last, std::random_access_iterator_tag pdf_tag)
+  {
+    if (pdf_first == pdf_last || pdf_last < pdf_first)
+      return E_EXPLORATION_BAD_RANGE;
```
JohnLangford: Here, pdf_first == pdf_last is E_EXPLORATION_BAD_RANGE while up above it is not.  This is inconsistent.

explore/exploration.h
```diff
@@ -0,0 +1,207 @@
...
+
+    uint32_t num_models = std::accumulate(top_actions_begin, top_actions_last, 0);
+    if (num_models == 0)
+    {
+      if (pdf_last - pdf_first > 0)
```
JohnLangford: This 'if' is redundant?

explore/exploration.h
```diff
@@ -0,0 +1,207 @@
...
+
+  template<typename It>
+  int enforce_minimum_probability(float min_prob, bool update_zero_elements, It pdf_first, It pdf_last, std::random_access_iterator_tag pdf_tag)
+  {
+    if (pdf_last < pdf_first)
+      return E_EXPLORATION_BAD_RANGE;
+
+	  size_t num_actions = pdf_last - pdf_first;
+
+    if (min_prob > 0.999) // uniform exploration
```
JohnLangford: This is still the wrong guard.   You want 1 / support_size.

JohnLangford: Ah, I see. This isn't the minimum probability per action.  It's actions * minimum probability per action.

explore/sampling_internal.h
JohnLangford: It seems odd and unintuitive to have sampling modify the distribution input.  Naively, I would have even imagined that an input iterator disallows that.

explore/sampling.h
```diff
@@ -0,0 +1,41 @@
...
+  template<typename InputIt>
+  uint32_t sample_from_pdf(const char* seed, InputIt pdf_first, InputIt pdf_last)
+  {
+    typedef typename std::iterator_traits<InputIt>::iterator_category pdf_category;
+    return sample_from_pdf(seed, pdf_first, pdf_last, pdf_category());
+  }
+
+  // sample ranking using top-slot swap from pdf using string seed
+  template<typename InputPdfIt, typename InputScoreIt, typename OutputIt>
+  void sample_from_pdf(const char* seed, InputPdfIt pdf_begin, InputPdfIt pdf_end, InputScoreIt scores_begin, InputScoreIt scores_end, OutputIt ranking_begin, OutputIt ranking_end)
```
JohnLangford: Why is the return type inconsistent?

JohnLangford:
For the test case added, did you verify the same printout under the old exploration library?

explore/exploration.h
eisber: done

explore/exploration.h
eisber: done

explore/sampling.h
eisber: what return type do you want?

explore/exploration.h
eisber: done

eisber:
Yes I did.

-Markus

eisber:
done

explore/explore.h
```diff
@@ -0,0 +1,148 @@
...
+  int enforce_minimum_probability(float minimum_uniform, bool update_zero_elements, It pdf_first, It pdf_last);
+
+  /**
+   * @brief Sample an index from the provided pdf.
+   * 
+   * @tparam InputIt Iterator type of the pdf. Must be an InputIterator.
+   * @param seed The seed for the pseudo-random generator.
+   * @param pdf_first Iterator pointing to the beginning of the pdf.
+   * @param pdf_last Iterator pointing to the end of the pdf.
+   * @param chosen_index returns the chosen a index.
```
JohnLangford: Typo "chosen a index"

explore/explore.h
```diff
@@ -0,0 +1,152 @@
...
+   * @param seed The seed for the pseudo-random generator.
+   * @param pdf_first Iterator pointing to the beginning of the pdf.
+   * @param pdf_last Iterator pointing to the end of the pdf.
+   * @param chosen_index returns the chosen index.
+   * @param pdf_updated set to true if the pdf required normalization and was updated in-place.
+   * @return int returns 0 on success, otherwise an error code as defined by E_EXPLORATION_*. 
+   */
+  template<typename InputIt>
+  int sample_after_normalizing(uint64_t seed, InputIt pdf_first, InputIt pdf_last, uint32_t& chosen_index, bool* pdf_updated = nullptr)
+  {
```
JohnLangford: Can we shift the function definition to the internal file?

explore/explore.h
```diff
@@ -0,0 +1,152 @@
...
+   * 
+   * @tparam InputIt Iterator type of the pdf. Must be an InputIterator.
+   * @param seed The seed for the pseudo-random generator.
+   * @param pdf_first Iterator pointing to the beginning of the pdf.
+   * @param pdf_last Iterator pointing to the end of the pdf.
+   * @param chosen_index returns the chosen index.
+   * @param pdf_updated set to true if the pdf required normalization and was updated in-place.
+   * @return int returns 0 on success, otherwise an error code as defined by E_EXPLORATION_*. 
+   */
+  template<typename InputIt>
```
JohnLangford: I believe this should be an It not a InputIt.

JohnLangford:
Please put all the C# stuff into a separate pull request.

explore/explore.h
eisber: done

explore/explore.h
eisber: done

explore/explore_internal.h
```diff
@@ -0,0 +1,322 @@
...
+  }
+  
+  template<typename InputIt>
+  int sample_after_normalizing(uint64_t seed, InputIt pdf_first, InputIt pdf_last, uint32_t& chosen_index, std::input_iterator_tag pdf_category, bool* pdf_updated = nullptr)
+  {
+    if (pdf_first == pdf_last || pdf_last < pdf_first)
+      return E_EXPLORATION_BAD_RANGE;
+
+    if (pdf_updated)
+      *pdf_updated = false;
```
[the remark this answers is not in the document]
eisber: if (pdf_update) <=> if (pdf_update != nullptr) <=> if(pdf_update != NULL)

PR #1437, opened 2018-03-24
Improve some parsing in vw-hyperopt.py

Plantaquatix:
In this utility script, this part reads the holdout file and it seems like weights and tags are converted to float along with the label.
For the input `-1 0.2 1|`, this is trying to do `float("-1 0.2 1")` and thus raising a ValueError exception.
I think it's fair to consider the label only up to the first space.


JohnLangford:
@pmineiro can you take a look?

pmineiro:
Looks reasonable.

JohnLangford:
Merged, thanks :-)

PR #1444, opened 2018-04-04
Support ppc64le

vibkulkarni:
Added check for ppc64le and removed the -msse2 option as its not supported on power. Changes have been tested.


JohnLangford:
Merged, thanks.

PR #1447, opened 2018-04-07
[quick fix] turn off exception throwing

yannstad:
Turn off exception throwing as long as the sum of probabilities is within 1e-4 of 0 and 1


yannstad:
not sure if it actually fixes the issue

PR #1451, opened 2018-04-17
Correctly exit loop searching for Boost libs on Linux

jaidevd:
It looks like line 33 in this file is supposed to stop iterating through candidates whenever `find_library` returns a non-empty result. However, `exit` is not what does that. `exit` is just a Python function (see https://docs.python.org/3/library/constants.html?highlight=exit#exit) which is simply sitting there, without being called. Even if it was called, it would exit the whole process, not the loop. Please see this for details: https://gist.github.com/jaidevd/81b870dacae045faabba587b73185386

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -30,7 +30,7 @@ def find_boost():
             for candidate in ['-py36', '-py35', '-py34', '3']:
                 boost_lib = 'boost_python{}'.format(candidate)
                 if find_library(boost_lib):
-                    exit
+                    break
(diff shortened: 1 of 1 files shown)


JohnLangford:
@Scott-Graham-Bose can you take a look?

Scott-Graham-Bose:
yes, good catch! thank you

JohnLangford:
Merged, thanks :-)

jaidevd:
@JohnLangford You're welcome :)

PR #1457, opened 2018-04-20
Sampling & Exploration refactoring

eisber:


explore/explore_internal.h
```diff
@@ -0,0 +1,347 @@
...
+      if (*pdf < 0)
+        *pdf = 0;
+
+      total += *pdf;
+    }
+
+    // assume the first is the best
+    if (total == 0)
+    {
+      chosen_index = 0;
```
JohnLangford: In this case, the returned distribution is wrong.   Make *pdf_first = 1?

explore/explore_internal.h
```diff
@@ -0,0 +1,347 @@
...
+  int sample_after_normalizing(const char* seed, It pdf_first, It pdf_last, uint32_t& chosen_index)
+  {
+	  typedef typename std::iterator_traits<It>::iterator_category pdf_category;
+    return sample_after_normalizing(seed, pdf_first, pdf_last, chosen_index, pdf_category());
+  }
+
+  template<typename PdfIt, typename InputScoreIt, typename OutputIt>
+  int sample_after_normalizing(uint64_t seed,
+      PdfIt pdf_begin, PdfIt pdf_end, std::random_access_iterator_tag pdf_category,
+      InputScoreIt scores_begin, InputScoreIt scores_last, std::random_access_iterator_tag scores_category,
```
JohnLangford: Why are we using begin/end, begin/last and first/last?  It seems like we should normalize throughout the interfaces?

explore/explore_internal.h
eisber: done

explore/explore_internal.h
eisber: done

explore/explore.h
```diff
@@ -0,0 +1,130 @@
...
+   * @param pdf_first Iterator pointing to the beginning of the pdf.
+   * @param pdf_last Iterator pointing to the end of the pdf.
+   * @param scores_begin Iterator pointing to the beginning of the scores.
+   * @param scores_last Iterator pointing to the end of the scores.
+   * @param ranking_begin Iterator pointing to the pre-allocated beginning of the output ranking.
+   * @param ranking_last Iterator pointing to the pre-allocated end of the output ranking.
+   * @return int returns 0 on success, otherwise an error code as defined by E_EXPLORATION_*. 
+   */
+  template<typename PdfIt, typename InputScoreIt, typename OutputIt>
+  int sample_after_normalizing(const char* seed, PdfIt pdf_first, PdfIt pdf_last, InputScoreIt scores_begin, InputScoreIt scores_last, OutputIt ranking_begin, OutputIt ranking_last);
```
JohnLangford: Still a begin/last pair.

JohnLangford: (Multiple instances---fix all.)

eisber: done

PR #1466, opened 2018-04-23
Update Link to determine where command line tools can be found.

sharatsc:
Command line tool location is specific to VS installation. Providing a generic link to determine it.


JohnLangford:
Merged, thanks.

PR #1478, opened 2018-05-07
README: Mention Graphviz dependency for doc generation

jinnovation:
Doxyfile has `HAVE_DOT` hard-coded to `YES`, which causes `command not found`
warnings when attempting to generate documentation on systems that don't have
Graphviz currently installed.

We add a clause in the corresponding section of the README that makes note of
this dependency.


JohnLangford:
Merged, thanks :-)

PR #1479, opened 2018-05-08
RL-clientlib PR: Interface, Logging, Error handling, Unit tests

rajan-chari:
From @rajan-chari on behalf of @yannstad

**In scope for this PR:**
a) Interface for decision service.
b) Logging events to event hub in background thread
c) Error handling
d) Unit tests

**Not in scope this PR**
a) Factory pattern and dependency injection for logging
b) Model management
c) Predict implementation
e) Linux build
f) Packaging for distribution

?) Single thread for all background processing


decision_service/ds.sln
```diff
@@ -0,0 +1,38 @@
...
+		Debug|x64 = Debug|x64
+		Debug|x86 = Debug|x86
+		Release|x64 = Release|x64
+		Release|x86 = Release|x86
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{C40E3263-B6FF-46D5-92F7-D67E6F9FCEDD}.Debug|x64.ActiveCfg = Debug|x64
+		{C40E3263-B6FF-46D5-92F7-D67E6F9FCEDD}.Debug|x64.Build.0 = Debug|x64
+		{C40E3263-B6FF-46D5-92F7-D67E6F9FCEDD}.Debug|x86.ActiveCfg = Debug|Win32
+		{C40E3263-B6FF-46D5-92F7-D67E6F9FCEDD}.Debug|x86.Build.0 = Debug|Win32
```
eisber: >|x86.Build.0 = Debug|Win32 [](start = 46, length = 26)

it's supported?

eisber:
> CodeFlow for GitHub does not download executable files. If you trust this file and would like to run it, please download it yourself through GitHub.

remove binary?

---
Refers to: decision_service/dsclientlib/dsclientlib:1 in f2590c2. [](commit_id = f2590c207896cf26b695fa937ef7731628b89436, deletion_comment = False)

decision_service/dsclientlib/dsclientlib.vcxproj.filters
```diff
@@ -0,0 +1,105 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
+    <Filter Include="Source Files">
+      <UniqueIdentifier>{4FC737F1-C7A5-4376-A066-2A32D752A2FF}</UniqueIdentifier>
+      <Extensions>cpp;c;cc;cxx;def;odl;idl;hpj;bat;asm;asmx</Extensions>
+    </Filter>
+    <Filter Include="Header Files">
+      <UniqueIdentifier>{93995380-89BD-4b04-88EB-625FBE52EBFB}</UniqueIdentifier>
+      <Extensions>h;hh;hpp;hxx;hm;inl;inc;xsd</Extensions>
```
eisber: do we need this file?

decision_service/dsclientlib/ds_api_status.cc
```diff
@@ -0,0 +1,41 @@
...
+		_error_code = new_code;
+	}
+
+	const std::string& api_status::get_error_msg() const {
+		return _error_msg;
+	}
+
+	void api_status::set_error_msg(const std::string& new_msg) {
+		_error_msg = new_msg;
+	}
```
eisber: why is this changable after creation?

decision_service/dsclientlib/ds_api_status.cc
```diff
@@ -0,0 +1,41 @@
...
+
+	void api_status::set_error_code(int new_code) {
+		_error_code = new_code;
+	}
+
+	const std::string& api_status::get_error_msg() const {
+		return _error_msg;
+	}
+
+	void api_status::set_error_msg(const std::string& new_msg) {
```
eisber: >void api_status::set_error_msg(const std::string& new_msg) { [](start = 1, length = 60)

shuold be private?

decision_service/include/ds_api_status.h
```diff
@@ -0,0 +1,36 @@
...
+		const int eventhub_http_bad_status_code = 4;
+	}
+
+
+	class api_status {
+	public:
+		int get_error_code() const;
+		void set_error_code(int);
+		const std::string& get_error_msg() const;
+		void set_error_msg(const std::string&);
```
eisber: >void set_error_msg(const std::string&); [](start = 2, length = 39)

should be private

decision_service/include/ds_api_status.h
```diff
@@ -0,0 +1,36 @@
...
+		const int background_queue_overflow = 2;
+		const int eventhub_http_generic = 3;
+		const int eventhub_http_bad_status_code = 4;
+	}
+
+
+	class api_status {
+	public:
+		int get_error_code() const;
+		void set_error_code(int);
```
eisber: >void set_error_code(int); [](start = 2, length = 25)

should be private

decision_service/dsclientlib/ds_config_collection.cc
```diff
@@ -0,0 +1,22 @@
...
+    const auto it = _map.find(name);
+    if (it != _map.end()) 
+      return it->second.c_str();
+    return defval;
+  }
+
+  int config_collection::get_int(const char* name, int defval) const {
+    const auto it = _map.find(name);
+    if (it != _map.end())
+      return std::stoi(it->second);
```
eisber: >std::stoi [](start = 13, length = 9)

I had some issues that this doesn't work on all compilers

decision_service/dsclientlib/ds_driver.cc
```diff
@@ -0,0 +1,131 @@
...
+#include "error_callback_fn.h"
+
+
+//this macro assumes that success_code equals 0
+#define TRY_OR_RETURN(x) do { \
+  int retval = (x); \
+  if (retval != 0) { \
+    return retval; \
+  } \
+} while (0)
```
eisber: >while (0) [](start = 2, length = 9)

why?

decision_service/dsclientlib/ds_driver.cc
```diff
@@ -0,0 +1,131 @@
...
+#include "logger/ds_logger.h"
+
+#include <boost/uuid/uuid_generators.hpp> // generators
+#include <boost/uuid/uuid_io.hpp>         // streaming operators etc.
+#include "error_callback_fn.h"
+
+
+//this macro assumes that success_code equals 0
+#define TRY_OR_RETURN(x) do { \
+  int retval = (x); \
```
eisber: >retval [](start = 6, length = 6)

add __LINENR__ (or whatever it's called to make sure the variable name is unique

decision_service/dsclientlib/ds_driver.cc
```diff
@@ -0,0 +1,131 @@
...
+	private:
+	  	utility::config_collection _configuration;
+    	error_callback_fn _error_cb;
+    	logger _logger;
+
+	public:
+		int ranking_request(const char * uuid, const char * context, ranking_response & response, api_status * status)
+		{
+	    	//clear previous errors if any
+      		if (status != nullptr) status->clear();
```
eisber: >if (status != nullptr) status->clear(); [](start = 8, length = 39)

introduce a utility function similar to try_set?

decision_service/dsclientlib/ds_driver.cc
```diff
@@ -0,0 +1,131 @@
...
+			TRY_OR_RETURN(_logger.append_ranking(evt.serialize(), status));
+
+			response.set_uuid(uuid);
+			response.set_ranking(action_proba);
+
+			return error_code::success;
+		}
+
+		//here the uuid is auto-generated
+		int ranking_request(const char * context, ranking_response & response, api_status * status)
```
eisber: >const char * context, ranking_response & response, api_status * status [](start = 22, length = 70)

can we have consistent code style const char * vs const char* (space between *)

decision_service/dsclientlib/ds_driver.cc
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+	};
+
+	//driver implementation
+	driver::~driver() {}
+
+	driver::driver(const utility::config_collection& config, error_fn fn, void* err_context) 
+	{
+		_pimpl = std::unique_ptr<driver_impl>(new driver_impl(config,fn,err_context));
```
eisber: >std::unique_ptr<driver_impl> [](start = 11, length = 28)

can't we do simple pointers and deallocate in the destructor. that way we don't need to expose std::unique_ptr. Or do you gain anything else from unique_ptr?

decision_service/dsclientlib/ds_event.h
```diff
@@ -0,0 +1,33 @@
...
+#include <string>
+#include <vector>
+
+namespace decision_service {
+
+	//serializable ranking event
+	class ranking_event {
+	public:
+		ranking_event(const char* uuid, const char* context, const std::vector<std::pair<int, float>>& ranking, const std::string& model_id);
+		std::string serialize();
```
eisber: >std::string serialize(); [](start = 2, length = 24)

any chance we can avoid the string copy?

decision_service/dsclientlib/ds_ranking_response.cc
```diff
@@ -0,0 +1,44 @@
...
+
+	const std::string& ranking_response::get_uuid() const
+	{
+		return _uuid;
+	}
+
+	int ranking_response::get_top_action_id() const
+	{
+		if (_ranking.empty()) {
+			throw std::exception();
```
eisber: >throw std::exception(); [](start = 3, length = 23)

need to switch to api status

decision_service/dsclientlib/ds_ranking_response.cc
```diff
@@ -0,0 +1,44 @@
...
+	{
+		return _ranking;
+	}
+
+	void ranking_response::set_uuid(const std::string & uuid)
+	{
+		_uuid = uuid;
+	}
+
+	void ranking_response::set_ranking(const std::vector<std::pair<int, float>>& ranking)
```
eisber: >std::vector<std::pair<int, float>>& ranking [](start = 42, length = 43)

are we exposing stl containers on the public api?

decision_service/dsclientlib/error_callback_fn.cc
```diff
@@ -0,0 +1,36 @@
+#include "error_callback_fn.h"
+
+using namespace std;
+namespace decision_service {
+
+  void error_callback_fn::set(error_fn fn, void* context)
+  {
+    lock_guard<mutex> lock(_mutex);
```
eisber: >lock_guard<mutex> lock(_mutex); [](start = 4, length = 31)

this is to make sure we have a consistent combination of fn/context?

decision_service/dsclientlib/error_callback_fn.cc
```diff
@@ -0,0 +1,36 @@
...
+    _fn = fn;
+    _context = context;
+  }
+
+  void error_callback_fn::report_error(api_status& s)
+  {
+    if (_fn == nullptr) 
+      return;
+
+    lock_guard<mutex> lock(_mutex);
```
eisber: >lock_guard<mutex> lock(_mutex); [](start = 4, length = 31)

that looks heavy.

decision_service/dsclientlib/error_callback_fn.h
```diff
@@ -0,0 +1,28 @@
...
+      error_callback_fn(error_fn, void*);
+      ~error_callback_fn() = default;
+
+    private:
+      error_callback_fn(const error_callback_fn&) = default;
+      error_callback_fn(error_callback_fn&&) = default;
+      error_callback_fn& operator=(const error_callback_fn&) = default;
+      error_callback_fn& operator=(error_callback_fn&&) = default;
+      
+      std::mutex _mutex;
```
eisber: >std::mutex _mutex; [](start = 6, length = 18)

what's the semantics of this when copied?

decision_service/dsclientlib/Makefile
```diff
@@ -0,0 +1,28 @@
+TARGET = dsclientlib
+LIBS = -lm
+CXX = g++
+CXXFLAGS = -g -Wall -std=c++11
```
eisber: >LAGS = -g -Wall -std=c++11 [](start = 4, length = 26)

can you re-use the Makefile infrastructure from VW?

eisber:
> 

do we need this?

---
Refers to: decision_service/dsclientlib/targetver.h:9 in f2590c2. [](commit_id = f2590c207896cf26b695fa937ef7731628b89436, deletion_comment = False)

decision_service/dsclientlib/logger/ds_async_batcher.h
```diff
@@ -0,0 +1,120 @@
...
+
+		int append(const std::string& evt, api_status* status = nullptr)
+		{
+			if (_queue.size() < _queue_max_size) {
+				_queue.push(evt);
+				return error_code::success;
+			}
+
+			//report errors
+			api_status::try_update(status, error_code::background_queue_overflow, "dropped event: " + evt);
```
eisber: >api_status::try_update(status, error_code::background_queue_overflow, "dropped event: " + evt); [](start = 3, length = 95)

at least in the past we adjusted the probabilities if we're dropping events.

decision_service/include/ds_driver.h
```diff
@@ -0,0 +1,31 @@
...
+		// request the decision service, in order to rank the N actions provided in the context_json string
+		int ranking_request(const char * uuid, const char * context_json, ranking_response&, api_status* = nullptr);
+		int ranking_request(const char * context_json, ranking_response&, api_status* = nullptr);//uuid is auto-generated
+
+		// report the reward for the top action
+		int report_outcome(const char* uuid, const char* outcome_data, api_status* = nullptr);
+		int report_outcome(const char* uuid, float reward, api_status* = nullptr);
+	  
+	private:
+		std::unique_ptr<driver_impl> _pimpl;
```
yannstad: The driver is not copyable, so you could add a private copy constructor

decision_service/dsclientlib/logger/ds_async_batcher.h
eisber: >"dropped event: " + evt [](start = 73, length = 23)

that works?

decision_service/dsclientlib/logger/ds_async_batcher.h
```diff
@@ -0,0 +1,120 @@
...
+		void flush()//flush all batches
+		{
+		  const auto queue_size = _queue.size();
+			if (queue_size == 0) return;
+
+			//handle batching
+			std::string batch, next;
+			_queue.pop(&batch); //there is at least one element
+
+			for (size_t i = 1; i < queue_size; ++i)
```
eisber: >1 [](start = 19, length = 1)

why start at 1?

decision_service/dsclientlib/logger/ds_async_batcher.h
```diff
@@ -0,0 +1,120 @@
...
+					//the batch is about to reach the max size: send it
+					if(_sender.send(batch,&status) != error_code::success)
+					{
+            if (_perror_cb) _perror_cb->report_error(status);
+					}
+          
+					batch = next;
+				}
+				else
+					batch += "\n" + next;
```
eisber: >batch += "\n" + next; [](start = 5, length = 21)

switch to ostrstream to avoid copying and re-allocation?

decision_service/dsclientlib/logger/ds_async_batcher.h
```diff
@@ -0,0 +1,120 @@
...
+				_queue.pop(&next);
+
+				size_t batch_size = batch.length() + next.length();
+				if (batch_size > _batch_max_size)
+				{
+          api_status status;
+					//the batch is about to reach the max size: send it
+					if(_sender.send(batch,&status) != error_code::success)
+					{
+            if (_perror_cb) _perror_cb->report_error(status);
```
eisber: > if (_perror_cb) _perror_cb->report_error(status); [](start = 11, length = 50)

use try_set_... function

eisber: > if (_perror_cb) _perror_cb->report_error(status); [](start = 11, length = 50)

maybe introduce a try_report_error function similar to api_status?

decision_service/dsclientlib/logger/ds_async_batcher.h
```diff
@@ -0,0 +1,120 @@
...
+		async_batcher(TSender& pipe, error_callback_fn* perror_cb = nullptr,
+                  size_t batch_max_size = (256 * 1024 - 1), size_t batch_timeout_ms = 1000, size_t queue_max_size = (8 * 1024))
+			: _sender(pipe),
+      _perror_cb(perror_cb),
+	    _batch_max_size(batch_max_size),
+			_batch_timeout_ms(batch_timeout_ms),
+			_queue_max_size(queue_max_size)
+		{
+			_thread_is_running = true;
+			_background_thread = std::thread(&async_batcher::timer, this);
```
eisber: >_background_thread = std::thread(&async_batcher::timer, this); [](start = 3, length = 62)

can this throw?

decision_service/include/include.vcxproj
```diff
@@ -0,0 +1,122 @@
...
+    <WindowsTargetPlatformVersion>8.1</WindowsTargetPlatformVersion>
+  </PropertyGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
+    <ConfigurationType>Application</ConfigurationType>
+    <UseDebugLibraries>true</UseDebugLibraries>
+    <PlatformToolset>v140</PlatformToolset>
+    <CharacterSet>MultiByte</CharacterSet>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
```
eisber: >'Release|Win32'" [](start = 60, length = 16)

do we support Win32?

eisber:
> <?xml version="1.0" encoding="utf-8"?>

we should be able to remove this file, right?

---
Refers to: decision_service/include/include.vcxproj.filters:1 in f2590c2. [](commit_id = f2590c207896cf26b695fa937ef7731628b89436, deletion_comment = False)

decision_service/unit_test/ds_eventhub_client_test.cc
```diff
@@ -0,0 +1,36 @@
...
+
+
+using namespace decision_service;
+
+
+BOOST_AUTO_TEST_CASE(send_something)
+{
+	//start a http server that will receive events sent from the eventhub_client
+	http_helper http_server;
+	http_server.on_initialize(U("http://localhost:8080"));
```
eisber: >U [](start = 27, length = 1)

isn't that a macro provided by cpprest? do we want to leak that?

decision_service/dsclientlib/logger/ds_eventhub_client.h
```diff
@@ -0,0 +1,33 @@
...
+
+	private:
+		std::string& authorization();
+
+	private:
+		web::http::client::http_client _client;
+		
+		const std::string _eventhub_host;          //e.g. "ingest-x2bw4dlnkv63q.servicebus.windows.net"
+		const std::string _shared_access_key_name; //e.g. "RootManageSharedAccessKey"
+		const std::string _shared_access_key;      //e.g. "2MNeQafvOsmV9jgbUOtx4I1KBAtQEWXqU2e6Om8M/n4="
```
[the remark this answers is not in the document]
yannstad: these are fake values

decision_service/unit_test/ReadMe.txt
```diff
@@ -0,0 +1,40 @@
+========================================================================
```
eisber: remove this file

decision_service/unit_test/targetver.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
```
eisber: do we need this file?

decision_service/unit_test/http_server.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
```
eisber: is this replaced by the C++ http server?

yannstad: yes, this one can be removed

decision_service/include/ds_config_collection.h
```diff
@@ -0,0 +1,19 @@
...
+  class config_collection
+  {
+  public:
+    void set(const char* name, const char* value);
+
+    const char* get(const char* name, const char* defval) const;
+    int get_int(const char* name, int defval) const;
+
+  private:
+    std::unordered_map<std::string, std::string> _map;
```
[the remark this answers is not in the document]
JohnLangford: What is the motivation for avoiding stl in the interface?

decision_service/dsclientlib/ds_config_collection.cc
```diff
@@ -0,0 +1,22 @@
+#include "ds_config_collection.h"
+
+namespace decision_service { namespace utility {
+
+  void config_collection::set(const char* name, const char* value) {
+    _map[name] = value;
```
JohnLangford: This seems semantically dangerous because the map has keys and values that can be mutated externally.

decision_service/dsclientlib/ds_config_collection.cc
```diff
@@ -0,0 +1,22 @@
+#include "ds_config_collection.h"
+
+namespace decision_service { namespace utility {
+
```
JohnLangford: I'm a little bit unclear on the utility of this class.  If you discard the class, then you don't need to even define set since _map[name]=value is fine.  Defining two helper get_default() in a namespace seems plenty straightforward.

decision_service/dsclientlib/ds_driver.cc
```diff
@@ -0,0 +1,131 @@
...
+	public:
+    	using error_fn = void(*)(const api_status&,void* user_context);
+
+	private:
+	  	utility::config_collection _configuration;
+    	error_callback_fn _error_cb;
+    	logger _logger;
+
+	public:
+		int ranking_request(const char * uuid, const char * context, ranking_response & response, api_status * status)
```
JohnLangford: I thought the plan was to have api_status with a default nullptr value?

decision_service/dsclientlib/ds_api_status.cc
```diff
@@ -0,0 +1,41 @@
...
+
+	int api_status::get_error_code() const {
+		return _error_code;
+	}
+
+	void api_status::set_error_code(int new_code) {
+		_error_code = new_code;
+	}
+
+	const std::string& api_status::get_error_msg() const {
```
rajan-chari: >std::string [](start = 7, length = 11)

Since api_status is part of our public interface, should we be exposing a char* instead of std::string?

decision_service/dsclientlib/ds_api_status.cc
```diff
@@ -0,0 +1,41 @@
+
+#include "ds_api_status.h"
+
+
+namespace decision_service {
+
+	int api_status::get_error_code() const {
+		return _error_code;
```
rajan-chari: In general this makes sense.  Since api_status is part of our public interface, does it make sense to keep all implementation in cc files, especially since we might be shipping binaries?

decision_service/dsclientlib/logger/ds_async_batcher.h
yannstad: Because the  `for` loop must iterate `queue_size-1` times, in order to dequeue all remaining elements
(The first one is dequeued before the loop, doing `_queue.pop(&batch)`.)

decision_service/unit_test/ds_eventhub_client_test.cc
yannstad: i think this macro is provided by the standard library, not cpprest:
http://en.cppreference.com/w/cpp/language/string_literal

decision_service/include/ds_config_collection.h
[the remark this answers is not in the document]
eisber: I got burned in the past with linking against different versions of STL (e.g. VS 2013 vs 2015). if the internal structure you get very hard to debug crashes

---
In reply to: [186772707](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186772707) [](ancestors = 186772707)

decision_service/dsclientlib/ds_api_status.cc
eisber: +1

---
In reply to: [186782757](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186782757) [](ancestors = 186782757)

decision_service/dsclientlib/logger/ds_async_batcher.h
eisber: ahhh... thanks!

---
In reply to: [186785374](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186785374) [](ancestors = 186785374)

decision_service/include/ds_config_collection.h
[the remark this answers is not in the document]
rajan-chari: Confirming my understanding of the issue:
  
1) Inside api-consumer code if new config_collection() is called or a config_collection is created on the stack, the compiler will look at this header to determine how much space to allocate.
2) It will then call the constructor for config_collection which is in dsclientlib.  
3) config_collection constructor which is compiled in to config_collection.obj will call the constructor for std::unordered_map is also compiled into config_collection.obj
4) So the concern is that if unordered_map header included in api-consumer code has different fields, it might allocate more or less memory than needed.  If it is more, it should not cause a big issue but if is less, then we might run into a large issue.
5) This assumes that ds-clientlib is included as a pre-compiled binary

Is this correct?  Is there something else that worries you?

---
In reply to: [186815209](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186815209) [](ancestors = 186815209,186772707)

decision_service/ds.sln
rajan-chari: Ran all unittests in x86, Release and Debug.  It seems to work.  Leaving it in till I hit a x86 implementation incompatibility.

---
In reply to: [186733028](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186733028) [](ancestors = 186733028)

decision_service/dsclientlib/dsclientlib.vcxproj.filters
rajan-chari: Helps organize files in visual studio.  Any reason to remove it?

---
In reply to: [186733204](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186733204) [](ancestors = 186733204)

decision_service/dsclientlib/ds_restapi_data_transport.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+#include "model_mgmt/ds_model_mgmt.h"
+
+namespace decision_service { namespace model_mangement {
+  class restapi_data_tranport : public i_data_transport {
+  public:
+    model_data get_data() override;
+  };
+}}
```
JohnLangford: Why is dsclientlib, a binary file, included in the pull request?

rajan-chari: Removed.  Yann will help me tomorrow from removing it from commit history.

---
In reply to: [186873323](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186873323) [](ancestors = 186873323)

JohnLangford:
Are there any tests this is supposed to pass?  I see nothing in the .yml files?

decision_service/include/ds_api_status.h
rajan-chari: removed method

---
In reply to: [186733796](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186733796) [](ancestors = 186733796)

decision_service/include/ds_api_status.h
rajan-chari: removed method

---
In reply to: [186733743](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186733743) [](ancestors = 186733743)

decision_service/dsclientlib/ds_config_collection.cc
rajan-chari: switched to atoi

---
In reply to: [186734171](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186734171) [](ancestors = 186734171)

decision_service/dsclientlib/ds_driver.cc
rajan-chari: Makes macros safe under several scenarios.  See [https://stackoverflow.com/questions/257418/do-while-0-what-is-it-good-for](https://stackoverflow.com/questions/257418/do-while-0-what-is-it-good-for)

---
In reply to: [186734789](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186734789) [](ancestors = 186734789)

decision_service/dsclientlib/ds_driver.cc
rajan-chari: retval is scoped by the do {} while block.  Is there a reason to make it unique?

---
In reply to: [186734938](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186734938) [](ancestors = 186734938)

decision_service/dsclientlib/ds_driver.cc
rajan-chari: done

---
In reply to: [186735215](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186735215) [](ancestors = 186735215)

decision_service/dsclientlib/ds_driver.cc
rajan-chari: done

---
In reply to: [186735496](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186735496) [](ancestors = 186735496)

decision_service/dsclientlib/ds_driver.cc
rajan-chari: Makes ownership and lifetime of the pointer clear.  The de-allocation is done automatically.  However unique_ptr does show up in the header file.  Removed it since its a simple change.

---
In reply to: [186736006](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186736006) [](ancestors = 186736006)

decision_service/dsclientlib/error_callback_fn.cc
rajan-chari: yes

---
In reply to: [186736980](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186736980) [](ancestors = 186736980)

decision_service/dsclientlib/error_callback_fn.cc
rajan-chari: a couple of issues here.  1) background thread is using the function pointer 2) we might have more than one background thread and user supplied error callback might not be reentrant. 

---
In reply to: [186737190](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186737190) [](ancestors = 186737190)

decision_service/dsclientlib/dsclientlib.vcxproj.filters
eisber: aren't those auto-generated?

---
In reply to: [186865764](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186865764) [](ancestors = 186865764,186733204)

decision_service/dsclientlib/ds_driver.cc
eisber: if I have a retval variable in the surrounding block it's clashing.

---
In reply to: [186922880](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186922880) [](ancestors = 186922880,186734938)

decision_service/dsclientlib/logger/ds_async_batcher.h
```diff
@@ -0,0 +1,123 @@
...
+
+		  return error_code::background_queue_overflow;
+		}
+
+	private:
+		void timer()//the timer triggers a queue flush (run in background)
+		{
+			while (_thread_is_running)
+			{
+				std::this_thread::sleep_for(std::chrono::milliseconds(_batch_timeout_ms));
```
JohnLangford: It would be better if this functioned on an amortized basis.  In particular, instead of a mystery parameter using _batch_timeout_ms = the median of the last 100 upload times seems good.

decision_service/dsclientlib/ds_config_collection.cc
eisber: there's a copy happening here due to map<string, string>

---
In reply to: [186774583](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186774583) [](ancestors = 186774583)

decision_service/include/ds_driver.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+#include "ds_ranking_response.h"
+#include "ds_config_collection.h"
+#include "ds_api_status.h"
+#include <memory>
+
+namespace decision_service {
+
+	class driver_impl;
```
eisber: >class driver_impl; [](start = 1, length = 18)

add doxygen comments

decision_service/dsclientlib/ds_event.h
```diff
@@ -0,0 +1,33 @@
...
+
+	private:
+		const char* _uuid;
+		const char* _context;
+		const std::vector<std::pair<int, float>>& _ranking;
+		const std::string& _model_id;
```
eisber: I'm worried about the life time of that reference

decision_service/dsclientlib/ds_driver_impl.cc
```diff
@@ -0,0 +1,80 @@
...
+      std::vector<std::pair<int, float>> action_proba;
+      std::string model_id = "model_id";
+      action_proba.push_back(std::pair<int, float>(2, 0.4f));
+      action_proba.push_back(std::pair<int, float>(1, 0.3f));
+      action_proba.push_back(std::pair<int, float>(4, 0.2f));
+      action_proba.push_back(std::pair<int, float>(3, 0.1f));
+
+      //send the ranking event to the backend
+      ranking_event evt(uuid, context, action_proba, model_id);
+      TRY_OR_RETURN(_logger.append_ranking(evt.serialize(), status));
```
eisber: switch to a regular function, avoid the object

decision_service/dsclientlib/ReadMe.txt
```diff
@@ -0,0 +1,37 @@
+========================================================================
+    STATIC LIBRARY : dsclientlib Project Overview
+========================================================================
+
+AppWizard has created this dsclientlib library project for you.
+
+This file contains a summary of what you will find in each of the files that
```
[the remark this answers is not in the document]
JohnLangford: I agree---this seems to be some autogenerated thing.

decision_service/dsclientlib/ds_api_status.cc
JohnLangford: I agree.

decision_service/dsclientlib/ds_config_collection.cc
JohnLangford: Right, thanks.

decision_service/dsclientlib/ds_driver_impl.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+#include "logger/ds_logger.h"
+
+namespace decision_service
+{
+  class ranking_response;
+  class api_status;
+
+  class driver_impl {
```
JohnLangford: Why is driver_impl the right name?  It sounds quite generic.

decision_service/dsclientlib/ds_driver_impl.h
```diff
@@ -0,0 +1,29 @@
...
+    using error_fn = void(*)( const api_status&, void* user_context );
+
+  private:
+    utility::config_collection _configuration;
+    error_callback_fn _error_cb;
+    logger _logger;
+
+  public:
+    
+    int ranking_request(const char* uuid, const char* context, ranking_response& response, api_status* status);
```
JohnLangford: Should we have =nullptr here?

decision_service/dsclientlib/ds_driver_impl.h
```diff
@@ -0,0 +1,29 @@
...
+    //here the uuid is auto-generated
+    int ranking_request(const char* context, ranking_response& response, api_status* status);
+    
+    int report_outcome(const char* uuid, const char* outcome_data, api_status* status);
+    int report_outcome(const char* uuid, float reward, api_status* status);
+    
+    explicit driver_impl(const utility::config_collection& config, error_fn fn = nullptr, void* err_context = nullptr);
```
JohnLangford: Why is explicit the right keyword to use?

rajan-chari:
I am adding support for linux build, run unit tests and Travis

---
In reply to: [387558412](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#issuecomment-387558412) [](ancestors = 387558412)

decision_service/dsclientlib/ds_restapi_data_transport.cc
```diff
@@ -0,0 +1,10 @@
+#include "ds_restapi_data_transport.h"
+
+namespace decision_service { namespace model_mangement {
+
```
JohnLangford: Is this a stub to be filled in?

decision_service/dsclientlib/logger/ds_async_batcher.h
```diff
@@ -0,0 +1,123 @@
...
+
+#include <string>
+#include <thread>
+#include <sstream>
+
+#include "ds_concurrent_queue.h"
+#include "ds_api_status.h"
+#include "../ds_error_callback_fn.h"
+
+namespace decision_service {
```
JohnLangford: A naming question: should we call it decision_service?

decision_service/dsclientlib/ds_restapi_data_transport.cc
rajan-chari: As part of model management PR (not in scope for the logging PR)

---
In reply to: [187104321](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187104321) [](ancestors = 187104321)

decision_service/dsclientlib/logger/ds_async_batcher.h
rajan-chari: vw_client , vw_client_lib more appropriate?  Other suggestions?  

---
In reply to: [187104723](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187104723) [](ancestors = 187104723)

decision_service/dsclientlib/logger/ds_concurrent_queue.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <queue>
+#include <mutex>
+
+
+namespace decision_service {
+
+	//a concurrent queue with locks and mutex
+	template <class T>
+	class concurrent_queue {
```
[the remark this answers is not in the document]
rajan-chari: yes.

---
In reply to: [187105931](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187105931) [](ancestors = 187105931)

decision_service/dsclientlib/ds_event.cc
```diff
@@ -0,0 +1,52 @@
...
+
+#include <sstream>
+
+
+namespace decision_service {
+	ranking_event::ranking_event(const char* uuid, const char* context, const std::vector<std::pair<int, float>>& ranking, const std::string& model_id)
+		: _uuid(uuid), _context(context), _ranking(ranking), _model_id(model_id)
+	{}
+
+	std::string ranking_event::serialize()
```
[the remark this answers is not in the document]
rajan-chari: Agree on both counts.  

---
In reply to: [187103009](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187103009) [](ancestors = 187103009)

decision_service/dsclientlib/logger/ds_async_batcher.h
JohnLangford: Yes, I'm worried about the time complexity here.  This: https://stackoverflow.com/questions/15082170/c-stdstring-append-vs-push-back seems to suggest that append amortizes to the additional bytes required which is reasonablish.  However, you might be creating a string and then adding it, which is doubling up on the work.  
Using an ostrstream does seem like it might work better.

Even better would be using some of the techniques in VW.   Instead of a queue, have a char* ring buffer and modify serialize to write directly into the ring buffer.  The modify flush() to just write an extent.  Locking would apply to the pointers indicating beginning and end of the queue. 

Even better would of course be sending bytes instead of strings...

decision_service/dsclientlib/model_mgmt/ds_model_mgmt.h
```diff
@@ -0,0 +1,47 @@
+#pragma once
+#include "../utility/ds_object_factory.h"
+
+// Declare const pointer for internal linkage
```
JohnLangford: The semantics of this is unclear to me---I think I need a walkthrough.

decision_service/dsclientlib/main.cc
```diff
@@ -0,0 +1,84 @@
...
+
+void error_handler(const api_status& error, void* user_context)
+{
+  std::cout << "An error happened in the background thread. Code:" << error.get_error_code() 
+    << " Details:" << error.get_error_msg() << std::endl;
+}
+
+int main()
+{
+  const auto config = init_from_json(R"({"eventhub_host":"localhost:8080"})");
```
JohnLangford: I'd suggest we walk through this carefully.

I'm a bit worried that initialization from json is adding work making use of the interface more difficult.

decision_service/include/ds_api_status.h
```diff
@@ -0,0 +1,33 @@
...
+namespace decision_service {
+  class api_status
+  {
+    public:
+      api_status();
+
+      int get_error_code() const;
+      const char* get_error_msg() const;
+
+      static void try_update(api_status* status, int new_code, const char* new_msg);
```
JohnLangford: What is the meaning of these functions?

decision_service/include/ds_ranking_response.h
```diff
@@ -0,0 +1,27 @@
...
+#include <vector>
+
+namespace decision_service {
+
+	//Ranking response returned by the decision service
+	class ranking_response {
+
+	public:
+		ranking_response();
+		ranking_response(const std::string& uuid, const std::vector<std::pair<int, float>>& ranking);
```
JohnLangford: When we discussed the name of this previously, I thought Rajan had a more neutral name thank 'ranking'?

JohnLangford:
I finished a first pass reading through things.

decision_service/unit_test/http_server/stdafx.cc
```diff
@@ -0,0 +1,12 @@
+/***
+* Copyright (C) Microsoft. All rights reserved.
```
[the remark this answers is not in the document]
rajan-chari: Pre-compiled header file used by windows. Speeds up re-compiles in windows when headers are included here.  

---
In reply to: [187120679](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187120679) [](ancestors = 187120679)

decision_service/include/ds_ranking_response.h
rajan-chari: I remember choose_rank() method name.  I don't recall response class name ... Maybe @yannstad ?

---
In reply to: [187118946](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187118946) [](ancestors = 187118946)

rajan-chari: >const std::string& uuid, const std::vector<std::pair<int, float>>& ranking [](start = 19, length = 74)

Todo: Remove stl and stdlib classes from the interface?

decision_service/include/ds_api_status.h
rajan-chari: They are helper functions used to check for nullptr api_status.  

---
In reply to: [187117708](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187117708) [](ancestors = 187117708)

decision_service/dsclientlib/main.cc
rajan-chari: Design principle was the following:

1) Initialize ds-clientlib using a name/value pair collection that has no knowledge of how configuration is stored.

2) The name/value pair collection can be populated in place (unit tests) or from other means e.g. a database table, csv config file, or a json config file etc.

3) We will supply utility methods so the current json config file stored in the user's app can be parsed and used to poplulate the config object.



---
In reply to: [187117181](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187117181) [](ancestors = 187117181)

decision_service/dsclientlib/model_mgmt/ds_model_mgmt.h
rajan-chari: Good explanation here.
[https://stackoverflow.com/a/998457](https://stackoverflow.com/a/998457)

---
In reply to: [187115294](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187115294) [](ancestors = 187115294)

decision_service/include/ds_ranking_response.h
yannstad: I don't recall either.
If we rename the class `ranking_response` in `ranking`, i suggest to also rename the `get_ranking()` and `set_ranking()` methods from the `ranking_response`.

decision_service/dsclientlib/logger/ds_async_batcher.h
rajan-chari: Makes sense.  Need to understand what the failure modes.  Feels like scope creep for this PR.  Will add a work item for ds_clientlib project.  Does that work?

---
In reply to: [187034482](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187034482) [](ancestors = 187034482)

decision_service/dsclientlib/ds_error_callback_fn.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+#include <mutex>
+
+namespace decision_service {
+  class api_status;
+
+  class error_callback_fn
+  {
+    public:
+      using error_fn = void(*)(const api_status&, void*);
```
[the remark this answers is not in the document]
rajan-chari: The way I imagine this being used by an API consumer is to send the API an opaque pointer along with the function to call back.  Generally this is a pointer to an object that the user will cast back into that object (since they know what they sent)  and then switch to c++ code to handle the error.

1) Since error_fn is going to be called back with the value passed in as void*, this order felt more natural to me.  However I am happy to change it if this seems odd.

2) We cannot carry the type information from the user into the background thread without needing to recompile the entire library.  Should be possible to loose the type information at the interface layer.   However that would mean the driver constructor will be recompiled by the client.  

---
In reply to: [187100996](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187100996) [](ancestors = 187100996)

decision_service/dsclientlib/ds_driver.cc
rajan-chari: makes sense

---
In reply to: [187031419](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187031419) [](ancestors = 187031419,186922880,186734938)

decision_service/dsclientlib/ds_driver_impl.h
rajan-chari: This will prevent driver_impl from getting constructed accidentally/implicitly from config_collection.  For example C++ compiler overload resolution will consider a function taking driver_impl as valid resolution if the parameter passed in is config_collection since other two parameters have default value.  Not strictly necessary, just a precaution.

---
In reply to: [187098796](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187098796) [](ancestors = 187098796)

decision_service/dsclientlib/ds_driver_impl.h
rajan-chari: This is an internal class that always gets called with an argument (by driver).  So not as much value in having a default value since its not meant to be used directly but can add it here. 

---
In reply to: [187098631](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187098631) [](ancestors = 187098631)

decision_service/dsclientlib/logger/ds_async_batcher.h
rajan-chari: Switch to ostrstream and add client lib work item to switch to bytes?  Probably makes sense to schedule that with moving to a binary format?

---
In reply to: [187109810](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187109810) [](ancestors = 187109810)

decision_service/dsclientlib/ds_config_collection.cc
rajan-chari: Would not be needed right now if we are fine with exposing stl in our interface.
In the long run, I was thinking about this as a message that can be sent to microservices that will allow us to change behavior at runtime.  For example it would be nice to change the log level at runtime.  

---
In reply to: [186775278](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186775278) [](ancestors = 186775278)

JohnLangford: Ok.

decision_service/dsclientlib/ds_driver_impl.h
JohnLangford: Perhaps the 'driver' part is the issue.   A library that has 'driver' as a primary object seems quite generic.  Is there a better descriptor for what 'driver' is supposed to do?

decision_service/dsclientlib/ds_driver_impl.h
JohnLangford: Cool.

decision_service/dsclientlib/ds_driver_impl.h
JohnLangford: Ok.

decision_service/dsclientlib/logger/ds_async_batcher.h
JohnLangford: That seems plausible.  Note that switching to ostrstream isn't quite enough---you want to have it defined at the class level not the function level so as to reuse memory.

decision_service/dsclientlib/logger/ds_concurrent_queue.h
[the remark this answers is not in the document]
JohnLangford: It seems ok to use then.  I'm skeptical we'll want to use it in the long term since we'd like to shift towards extents on a ring buffer.  In the short term, locking is a relatively expensive operation so eliminating that could be desirable.

decision_service/include/ds_api_status.h
JohnLangford: Ok, sounds good.

JohnLangford: (Maybe add a comment.)

decision_service/dsclientlib/logger/ds_async_batcher.h
rajan-chari: Makes sense.  Will need to consider concurrency.  

---
In reply to: [187339305](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187339305) [](ancestors = 187339305)

decision_service/include/ds_api_status.h
rajan-chari: Yep

---
In reply to: [187340727](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187340727) [](ancestors = 187340727)

decision_service/include/ds_ranking_response.h
JohnLangford: choose_rank() was what I was thinking of.   Why aren't we using that in ds_driver.h?

decision_service/dsclientlib/ds_driver_impl.h
rajan-chari: A couple of suggestions:
a) decision_service
b) ds_client

decision_service/dsclientlib/ds_error_callback_fn.h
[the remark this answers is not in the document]
JohnLangford: Perhaps we should have both an alternate header-only type-checking interface and the void* interface?  That will help document what the void* is supposed to do and provides easy type checking for those wanting to compile?

.travis.yml
```diff
@@ -2,9 +2,23 @@ language: cpp
...
+  - sudo apt-get install g++ git make zlib1g-dev libboost-all-dev libssl-dev cmake
+  - git clone https://github.com/Microsoft/cpprestsdk.git casablanca
+  - cd casablanca/Release
+  - mkdir build
+  - cd build
+  - cmake .. 
+  - make
+  - sudo make install
+  - cd ../../../
+  #### end cpprest
```
eisber: I remember that we've been tight on time budget with travis. Do you know how long this takes?

decision_service/include/ds_driver.h
```diff
@@ -0,0 +1,31 @@
...
+	//decision service client
+	class driver {
+
+	public:
+    using error_fn = void(*)(const api_status&, void*);
+    explicit driver(const utility::config_collection& config, error_fn fn = nullptr, void* err_context = nullptr);
+		~driver();
+
+		// request the decision service, in order to rank the N actions provided in the context_json string
+		int ranking_request(const char * uuid, const char * context_json, ranking_response&, api_status* = nullptr);
```
eisber: >ranking_request [](start = 6, length = 15)

rename to choose_rank

decision_service/dsclientlib/Makefile
eisber: upgrade root Makefile to c++11

---
In reply to: [186742232](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186742232) [](ancestors = 186742232)

decision_service/dsclientlib/main.cc
```diff
@@ -0,0 +1,87 @@
...
+    << " Details:" << error.get_error_msg() << std::endl;
+}
+
+int main()
+{
+	const auto config = init_from_json(R"({"eventhub_host":"localhost:8080"})");
+
+	auto error_cntxt = 1;
+	// Create a ds driver, and initialize with configuration
+	driver ds(config, error_handler, (void*)(&error_cntxt));
```
eisber: >driver [](start = 1, length = 6)

rename driver to ? (maybe personalization_client?)

decision_service/dsclientlib/logger/ds_logger.cc
```diff
@@ -0,0 +1,37 @@
...
+        c.get("eventhub_interaction_name", "interaction")),
+      _outcome_client(
+        c.get("eventhub_host", "localhost:8080"),
+        c.get("shared_access_key_name", ""),
+        c.get("shared_access_key", ""),
+        c.get("eventhub_observation_name", "observation")),
+      _async_batcher(
+        _ranking_client,
+        perror_cb,
+        c.get_int("batch_max_size", 8 * 1024),
```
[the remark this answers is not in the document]
eisber: defaults are 4*1024*1024 in C#

---
In reply to: [187148368](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187148368) [](ancestors = 187148368,187114050)

decision_service/dsclientlib/logger/ds_logger.cc
```diff
@@ -0,0 +1,37 @@
...
+        c.get_int("batch_timeout_ms", 1000),
+        c.get_int("queue_max_size", 1000 * 2))
```
eisber: >       c.get_int("queue_max_size", 1000 * 2)) [](start = 1, length = 45)

performance test of transport to determine reasonable values

decision_service/dsclientlib/logger/ds_eventhub_client.h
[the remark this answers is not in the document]
eisber: alternative is to point to EventHub documentation/reference

---
In reply to: [186749618](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186749618) [](ancestors = 186749618)

decision_service/dsclientlib/logger/ds_eventhub_client.cc
```diff
@@ -0,0 +1,142 @@
...
+
+			//report error
+			if (status)
+			{
+				std::ostringstream error_msg;
+				error_msg << e.what();
+				error_msg << "post_data: " << post_data;
+
+				status->set_error_code(error_code);
+				status->set_error_msg(error_msg.str());
```
eisber: can we introduce a macro/class/whatever that makes this more elegant? REPORT(status, code) << ".... " << "... ;

decision_service/dsclientlib/logger/ds_eventhub_client.cc
```diff
@@ -0,0 +1,142 @@
...
+	eventhub_client::eventhub_client(const std::string& host, const std::string& key_name, const std::string& key, const std::string& name)
+		: _client(build_url(host, name)),
+		_eventhub_host(host), _shared_access_key_name(key_name), _shared_access_key(key), _eventhub_name(name),
+		_authorization_valid_until(0)
+	{
+		//init authorize token
+		authorization();
+	}
+
+	std::string& eventhub_client::authorization()
```
eisber: >std::string& eventhub_client::authorization() [](start = 1, length = 45)

deal with concurrency. the actually work of re issuing the authorization token is in-frequent

decision_service/dsclientlib/logger/ds_concurrent_queue.h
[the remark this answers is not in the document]
JohnLangford: test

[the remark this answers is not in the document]
eisber: test 2

---
In reply to: [187641806](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187641806) [](ancestors = 187641806)

decision_service/dsclientlib/logger/ds_async_batcher.h
```diff
@@ -0,0 +1,123 @@
...
+		{
+			if (_queue.size() < _queue_max_size) {
+				_queue.push(evt);
+				return error_code::success;
+			}
+
+			//report errors
+		  std::ostringstream os;
+		  os << "Dropped event: " << evt;
+			api_status::try_update(status, error_code::background_queue_overflow, os.str().c_str());
```
eisber: use the REPORT macro

decision_service/dsclientlib/logger/ds_async_batcher.h
eisber: maybe introduce a similar macro to REPORT for callback.

What to do if there is a stack of call? call at each level? call at the end?

---
In reply to: [186743495](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186743495) [](ancestors = 186743495)

decision_service/dsclientlib/logger/ds_async_batcher.h
eisber: it can throw according to Yann. we need to handle

---
In reply to: [186743714](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186743714) [](ancestors = 186743714)

decision_service/dsclientlib/ds_error_callback_fn.h
```diff
@@ -0,0 +1,28 @@
...
+      error_callback_fn(error_fn, void*);
+      ~error_callback_fn() = default;
+
+    private:
+      error_callback_fn(const error_callback_fn&) = default;
+      error_callback_fn(error_callback_fn&&) = default;
+      error_callback_fn& operator=(const error_callback_fn&) = default;
+      error_callback_fn& operator=(error_callback_fn&&) = default;
+      
+      std::mutex _mutex;
```
eisber: >std::mutex _mutex; [](start = 6, length = 18)

we should not copy.

decision_service/dsclientlib/ds_event.h
eisber: switch to plain function instead of class

---
In reply to: [187049450](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187049450) [](ancestors = 187049450)

decision_service/dsclientlib/ds_event.h
```diff
@@ -0,0 +1,33 @@
...
+	};
+
+
+	//serializable outcome event
+	class outcome_event {
+	public:
+		outcome_event(const char* uuid, const char* outcome_data);
+		std::string serialize();
```
eisber: switch to plain function?

.travis.yml
eisber: upgrading the ubuntu version [https://github.com/travis-ci/travis-ci/issues/6469](https://github.com/travis-ci/travis-ci/issues/6469)

---
In reply to: [187608097](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187608097) [](ancestors = 187608097)

decision_service/dsclientlib/logger/ds_async_batcher.h
eisber: add work item and schedule

---
In reply to: [186742634](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186742634) [](ancestors = 186742634)

.travis.yml
rajan-chari: Builds successfully.  Total build time is 37 mins.  We can close this.

---
In reply to: [187651899](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187651899) [](ancestors = 187651899,187608097)

rajan-chari:
Done.

---
In reply to: [387798616](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#issuecomment-387798616) [](ancestors = 387798616,387558412)

decision_service/dsclientlib/ds_driver.cc
rajan-chari: Done

---
In reply to: [187047467](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187047467) [](ancestors = 187047467,186922638,186734789)

decision_service/dsclientlib/ds_event.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include <string>
+#include <vector>
+
+namespace decision_service {
+
+	//serializable ranking event
+	class ranking_event {
```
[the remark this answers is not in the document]
rajan-chari: Done

---
In reply to: [187049963](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187049963) [](ancestors = 187049963)

decision_service/dsclientlib/Makefile
rajan-chari: Done

---
In reply to: [187632446](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187632446) [](ancestors = 187632446,186742232)

decision_service/dsclientlib/ReadMe.txt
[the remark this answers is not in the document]
rajan-chari: Done

---
In reply to: [186742360](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r186742360) [](ancestors = 186742360)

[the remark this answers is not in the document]
rajan-chari: Done

---
In reply to: [187317496](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187317496) [](ancestors = 187317496,187095438)

decision_service/dsclientlib/logger/ds_async_batcher.h
rajan-chari: Added work item under logging.

---
In reply to: [187653180](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187653180) [](ancestors = 187653180,186742634)

rajan-chari: Done.

---
In reply to: [187774785](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187774785) [](ancestors = 187774785,187653180,186742634)

decision_service/include/ds_driver.h
rajan-chari: Done

---
In reply to: [187628839](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187628839) [](ancestors = 187628839,186742650)

decision_service/include/ds_driver.h
rajan-chari: Done

---
In reply to: [187626574](https://github.com/JohnLangford/vowpal_wabbit/pull/1479#discussion_r187626574) [](ancestors = 187626574)

Makefile
```diff
@@ -71,39 +71,45 @@ else
...
-all:	vw library_example java spanning_tree
+all:	vw library_example java spanning_tree ds_clientlib
 
 %.1:	%
 	help2man --no-info --name="Vowpal Wabbit -- fast online learning tool" ./$< > $@
 
 export
 
+ds_clientlib:
+	cd decision_service/dsclientlib; $(MAKE) -j $(NPROCS) things
```
rajan-chari: >; [](start = 32, length = 1)

Convert to && through out

decision_service/unit_test/Makefile
```diff
@@ -0,0 +1,27 @@
+TARGET = dsclient-test.out
```
rajan-chari: Need Makefile.am for integration with automake

decision_service/dsclientlib/model_mgmt/ds_model_mgmt.cc
```diff
@@ -0,0 +1,30 @@
...
+
+namespace decision_service { namespace model_mangement
+{
+  i_data_transport* blob_tranport_create(const config_collection&)
+  { return new restapi_data_tranport(); }
+
+  object_factory<i_data_transport> model_data_factory()
+  {
+    object_factory<i_data_transport> ret;
+    ret.register_type(value::AZURE_STORAGE_BLOB, blob_tranport_create);
```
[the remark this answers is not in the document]
rajan-chari: Not in scope for this PR.

decision_service/unit_test/ReadMe.txt
rajan-chari: Done.

decision_service/unit_test/targetver.h
rajan-chari: Done

PR #1480, opened 2018-05-11
Update build rules to enable building alongside conda in Linux

mlisovyi:
Hi,

this PR originates from the trouble that people encounter when they install VW having conda in their environment. The intention is not to break existing building rules, but to allow to build with relevant libs and compiler from conda directly.

The top-level `Makefile` is modified to use `BOOST_XXX` environmental variables and to follow the original values if those are not present. This allows to pass location in a docker image without a need to directly modify the `Makefile`, is is suggested in https://github.com/JohnLangford/vowpal_wabbit#osx-python-binding-installation-with-anaconda.

The `vowpalwabbit/Makefile` is modified to avoid building of dependency files, when make clean is called and to make use of the user-preferred boost includes

The installation docs have not been updated, but I can add them to the PR, if you think it would be useful

Cheers,
Misha


JohnLangford:
Yes, please add doc updates to this.

JohnLangford:
Thanks!  @Scott-Graham-Bose can you take a look?

arielf:
Hi @mlisovyi 

Thanks so much for doing all this.

I wonder whether automatic detection of conda can be done to save the user from having to set env variables.

something along the line of (bash)
```
python_used="$(which python)"
case "$python_used" in
    (*conda/*) # user prefers anaconda
                     # figure out the paths to include and lib dirs using realpath and dirname ...
                     BOOST_INCLUDE=...
                     BOOST_LIBRARY=...
                     ;;
esac
```

Makefile
```diff
@@ -19,8 +19,12 @@ endif
 UNAME := $(shell uname)
 ARCH_UNAME := $(shell uname -m)
 LIBS = -l boost_program_options -l pthread -l z
-BOOST_INCLUDE = -I /usr/local/include/boost -I /usr/include
-BOOST_LIBRARY = -L /usr/local/lib -L /usr/lib
+ifeq ($(BOOST_INCLUDE),)
+  BOOST_INCLUDE = -I /usr/local/include/boost -I /usr/include
```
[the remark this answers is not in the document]
Scott-Graham-Bose: getting this sorted for conda would be huge, although I'm wondering if it might be better to build vowpalwabbit as a conda package instead of trying to support installation from the source

some instructions here: https://conda.io/docs/user-guide/tutorials/build-pkgs.html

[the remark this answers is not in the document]
mlisovyi: Thanks, you are absolutely right, I did not think of other OS's. My bad. A simple work around would be to have different environmental variable names, e.g. VW_BOOST_XXX, to check them after OS-specific settings are done around [line 60](https://github.com/mlisovyi/vowpal_wabbit/blob/8fcea2734efa59a1c554e5bb13251569a65c61b2/Makefile#L60) and to overwrite the OS-specific values of BOOST_XXX that have been set.

Building a conda package would be ultimately the best solution. However, I have never done it before and I'm sure there will be some technical issues, which I do not have time to fight with right now. So if there is someone with the relevant  expertise and willing to dedicate a fraction of his/her time to it- wonderful.

mlisovyi:
@arielf yes, we can detect presence of conda automatically in the way that you have suggested. I can implement it on Monday (not much work needed, but I have other plans for the weekend). The only issue with that approach is that the user **must** have packages installed in conda, because we will not include other system directories, whereas with env vars we give the user freedom to choose what to use (although non-conda packages will with high probability fail in the presence of conda). If you think it is fine, I will go forward with it.

arielf:
Or a script to install `vw` in anaconda using all the needed `conda install ...`

python/README.rst
```diff
@@ -103,10 +103,33 @@ For Mac OSX
...
+    git clone https://github.com/JohnLangford/vowpal_wabbit.git
+    # build the C++ binaries and python package
+    cd vowpal_wabbit
+    python python python/setup.py install
+
+This procedure was tested with python3, but is expected to also work with python2.
+Typically, you can find `/your_conda_env_path` by either checking the `$CONDA_PREFIX` environmental variable
+or `conda info --base` if you do not have any environments (beware, environments make your life better).
+
+
```
mlisovyi: @arielf There is no conda script provided, but there are instructions added to the docs. I can complete those instructions to automatically deduce `BOOST_XXX from the python path as you have suggested. Did you suggest a real script in the repository? I do not know what are the policies about it and if that is encouraged/allowed?

python/README.rst
```diff
@@ -103,10 +103,31 @@ For Mac OSX
...
+    # make a soft link to the compiler, since Makefiles internally use `which g++`
+    [ ! -z ${GXX} ] && ln -s ${GXX} `dirname ${GXX}`/g++
+    # set BOOST_XXX variables, that will be used to find boost libs in linking
+    export BOOST_INCLUDE='-I /your_conda_env_path/include/boost -I /your_conda_env_path/include'
+    export BOOST_LIBRARY='-L /your_conda_env_path/lib'
+    # get the package from github
+    git clone https://github.com/JohnLangford/vowpal_wabbit.git
+    # build the C++ binaries and python package
+    cd vowpal_wabbit
+    python python python/setup.py install
```
arielf: redundant `python`?

python/README.rst
```diff
@@ -103,10 +103,31 @@ For Mac OSX
...
-
+Having `conda` on a Linux-based system requires a build with conda-provided packages to ensure consistency:
+.. code-block:: bash
+    # create/activate a conda environment (depends on your choice, no example is given)
+    # boost packages
+    conda install boost libboost py-boost
+    # a consistent compiler (g++ 7.2.0)
+    conda install gxx_linux-64
+    # pick up the newly installed compiler
+    source activate YourEnvironmentName
```
arielf: Can `YourEnvironmentName` be automatically determined inside conda?

python/README.rst
arielf: Thanks @mlisovyi 
Ideally the build is uniform/short and can be supported directly in the `Makefile` with only some env var changes.
But since anaconda is so involved, have it own tools and need additional operations, it may be preferable to keep all this unusual and non-standard complexity out of the standard `Makefiles`.  The procedure itself, may call some make targets (if they work as is, there's no need to repeat them in the conda script) but given the unique differences, I'd personally prefer a separate `conda-install.sh` over filling the Makefiles with env-specific cruft.

python/README.rst
mlisovyi: Yes, will be done in the proposed script

python/README.rst
mlisovyi: Sure, a typo on my side. Will be fixed in the proposed script

python/README.rst
mlisovyi: @arielf could you please suggest where would be an appropriate location to place such a script? I would imagine directly in the `python` folder, as conda is used by python users.

arielf: I think both in the python dir or the top dir would make sense. top dir may be more visible/obvious to new anaconda users trying to figure out things.  Please make sure to name the script clearly (like the suggestion above, or similar) so its purpose is obvious for random observers. Thanks for all your help on this. Anaconda has been a pain point for many, but no one of the problem reporters, stakeholders, has contributed a solution. Really glad someone is taking responsibility and fixing this  :+1:

python/conda_install.sh
```diff
@@ -33,4 +33,5 @@ export USER_BOOST_INCLUDE="-I $BASE_BOOST/include/boost -I $BASE_BOOST/include"
 export USER_BOOST_LIBRARY="-L $BASE_BOOST/lib"
 
 # build the C++ binaries and python package
-python python/setup.py install
+python vowpal_wabbit/python/setup.py install
+pip install -U vowpal_wabbit/python/
```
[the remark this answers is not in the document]
mlisovyi: That is something i do not understand either. But without it there is a local package build as well as `vowpalwabbit-8.5.0-py3.6-linux-x86_64.egg` among conda packages, but the package itself is not installed into the the conda packages. Thus, `import vowpalwabbit` fails since it can not find the package.

mlisovyi:
Ok, i've replaces instructions in the text with a complete building script. It still requires to get some variables from the environment and also a small tweak to the setup.py script to force skiping of `find_library` usage. The latter change is due to `find_library` is not customisable, has different search behaviour on different OS's according to the manual and is tricky to guide it to find the conda-installed library. For example, I also have a central system version of boost-python, but it is different from the conda version (python version is different). The central version was discovered by `find_library` and added into linking, but of course this crashed compilation as inedeed the conda version.
Current PR is ready for review. It is not perfect (as @Scott-Graham-Bose already started to discover), but it seems to work...

python/conda_install.sh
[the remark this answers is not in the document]
mlisovyi: That works to some extend, I have tried that as well:) But it doesn't install the package inside the conda environment. Instead, it keeps the build locally and sets an egg-link, if I remember correctly. This allows to import the package. But I'm not sure how robust is it for all possible use cases, i.e. wrapping in into a docker image or whatever else people might want to do with it. I have close-to-zero experience with docker, so I might just misunderstand how it functions. If you think that is it good-enough solution, than I can switch to that

Scott-Graham-Bose:
I'm trying this on a docker container with ubuntu 16.04 and anaconda (python2), it does not seem to be pulling in the boost header files, is there a conda install package to get libboost-dev that is compatible with the rest of the packages, or am I doing something wrong?

also, I'm noticing python 3.6 is being installed into the conda environment, this doesn't seem right?

Scott-Graham-Bose:
ok, i was able to get this working in a docker container running ubuntu 16.04 and anaconda2, but it installs python 3.6 on top of the previous version of python 2. I suspect it is from the py-boost package that conda is installing? do you know how to install py-boost using python2?

mlisovyi:
@Scott-Graham-Bose i also use anaconda2, but i tried it inside a python3.6 environment. I will try it with a python2 environment hopefully today and will get back to you. To be complete, which py2.X version did you use?

mlisovyi:
@Scott-Graham-Bose 
Creating a new environment based on python2.7, the whole script runs and builds the VW package. In such fresh setup i do not see an attempt to download python3. I get all packages aligned with python2.7, and in particular `py-boost                  1.65.1           py27hf484d3e_4`. Note that all boost packages on conda seem to exist only with python2.7 builds, could it be that you have python2.6 or even earlier version in your conda environment?

Scott-Graham-Bose:
```
root@db1e58c83583:/# source activate vw
(vw) root@db1e58c83583:/# python --version
Python 2.7.14 :: Anaconda, Inc.
(vw) root@db1e58c83583:/# conda install py-boost
Solving environment: done


==> WARNING: A newer version of conda exists. <==
  current version: 4.4.10
  latest version: 4.5.4

Please update conda by running

    $ conda update -n base conda



## Package Plan ##

  environment location: /root/anaconda2/envs/vw

  added / updated specs:
    - py-boost


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    certifi-2018.4.16          |           py36_0         142 KB
    libedit-3.1.20170329       |       h6b74fdf_2         172 KB
    python-3.6.5               |       hc3d631a_2        29.4 MB
    libgcc-ng-7.2.0            |       hdf63c60_3         6.1 MB
    ca-certificates-2018.03.07 |                0         124 KB
    libgfortran-ng-7.2.0       |       hdf63c60_3         1.2 MB
    pip-10.0.1                 |           py36_0         1.8 MB
    mkl_random-1.0.1           |   py36h629b387_0         373 KB
    wheel-0.31.1               |           py36_0          62 KB
    setuptools-39.1.0          |           py36_0         550 KB
    openssl-1.0.2o             |       h20670df_0         3.4 MB
    libboost-1.65.1            |       habcd387_4        19.2 MB
    intel-openmp-2018.0.0      |                8         620 KB
    libstdcxx-ng-7.2.0         |       hdf63c60_3         2.5 MB
    sqlite-3.23.1              |       he433501_0         1.5 MB
    numpy-1.14.2               |   py36hdbf6ddf_1         4.1 MB
    py-boost-1.65.1            |   py36hf484d3e_4         297 KB
    mkl_fft-1.0.1              |   py36h3010b51_0         140 KB
    xz-5.2.4                   |       h14c3975_4         366 KB
    ncurses-6.1                |       hf484d3e_0         943 KB
    mkl-2018.0.2               |                1       205.2 MB
    ------------------------------------------------------------
                                           Total:       278.2 MB

The following NEW packages will be INSTALLED:

    bzip2:           1.0.6-h9a117a8_4
    ca-certificates: 2018.03.07-0
    certifi:         2018.4.16-py36_0
    icu:             58.2-h9c2bf20_1
    intel-openmp:    2018.0.0-8
    libboost:        1.65.1-habcd387_4
    libedit:         3.1.20170329-h6b74fdf_2
    libffi:          3.2.1-hd88cf55_4
    libgcc-ng:       7.2.0-hdf63c60_3
    libgfortran-ng:  7.2.0-hdf63c60_3
    libstdcxx-ng:    7.2.0-hdf63c60_3
    mkl:             2018.0.2-1
    mkl_fft:         1.0.1-py36h3010b51_0
    mkl_random:      1.0.1-py36h629b387_0
    ncurses:         6.1-hf484d3e_0
    numpy:           1.14.2-py36hdbf6ddf_1
    openssl:         1.0.2o-h20670df_0
    pip:             10.0.1-py36_0
    py-boost:        1.65.1-py36hf484d3e_4
    python:          3.6.5-hc3d631a_2
    readline:        7.0-ha6073c6_4
    setuptools:      39.1.0-py36_0
    sqlite:          3.23.1-he433501_0
    tk:              8.6.7-hc745277_3
    wheel:           0.31.1-py36_0
    xz:              5.2.4-h14c3975_4
    zlib:            1.2.11-ha838bed_2

Proceed ([y]/n)? y
```

if i proceed, after this installs I get this:
```
(vw) root@db1e58c83583:/# python --version
Python 3.6.5 :: Anaconda, Inc.
```
=(

Scott-Graham-Bose:
ok i think the issue was how I created the conda env
I did this: 
```conda create --name vw```
which did use python 2.7.14, but for some reason allowed conda to update to python 3.6 during installation

when I did this:
```conda create --name vw python=2.7```
it used python 2.7.15 and did not update to python 3.6 when I installed py-boost

**Comment by Scott-Graham-Bose** [comment]
I think using pip install -e and not using python setup.py install at all is the way to go.

**Comment by mlisovyi** [comment]
sure, i will switch to that later today (CET time)

**Comment by mlisovyi** [comment]
Hi @arielf Could you please have a look at the updated PR? Thanks in advance

**Comment by arielf** [comment]
Hi @mlisovyi 

Thanks for making this a script!
Hopefully it will be helpful for future conda users.

A few comments from reading this script:
- Always quote variables like this before you pass them on, just in case they include spaces which may break commands expecting a single-arg:
     if [ -z "$variable_name" ] 
- Change backquoted-commands to modern bash command substitution:  $(command args...)
- Instead of echo "I do not know what to do... Exiting before VW build" which isn't helpful (doesn't tell the user what's wrong) say exactly what the problem is: echo "\$CONDA_PREFIX is not set - aborting"
- Errors should go to stderr, so error messages should look like:  echo ... >&2
- symlinking g++ to the conda gcc creates an intrusive side-effect (possibly damaging a user env), instead, simply prepend the directory name of the conda g++ to the PATH so the rest of the script uses it, but it doesn't leave side-effects behind after the script exits.
- I wonder whether the boost version used by conda BOOST_V="1.65" can be discoverable instead of hardwired.

HTH and thanks again for your contribution.

**Comment by mlisovyi** [comment]
@arielf Awesome! thanks for the thorough look! Let me put right away a few feedback points on the issues, that are not straightforward, or where I would like to iterate:

> symlinking g++ to the conda gcc creates an intrusive side-effect (possibly damaging a user env), instead, simply prepend the directory name of the conda g++ to the PATH so the rest of the script uses it, but it doesn't leave side-effects behind after the script exits.

Small technical note, that does not change you main point: _symlinking g++ to the **conda g++**_. The problem is that conda g++ does not have `g++` name, but a complex name including architecture and OS. The directory, where is lives (`dirname ${GXX}`) is already prepended to the `PATH` by setting up a conda environment. But VW Makefiles explicitly look for `g++`. So the installed conda `g++` can not be found unless we symlink to the name expected by VW. Does it make sense?

> I wonder whether the boost version used by conda BOOST_V="1.65" can be discoverable instead of hardwired.

I started by letting conda decide which versions of the packages to get. It worked well in python3.5, however in python3.6 there was boost version 1.67, while others were version 1.65 only. And I did not manage to make VW compile with those package versions scattered around.

> source activate ... seems wrong. source should be passed a single file arg to source

This is the standard way a conda environment is set up. `activate` is a conda setup script and it takes a single argument as input (the environment name).

I would like to hear back from you or get your blessing not to implement those suggestions.
Thanks in advance!

**Comment by arielf** [comment]
Thanks @mlisovyi 

-  `source activate ...` - ok.  I misunderstood. I didn't realize this is how conda works. You may pass further args to `source`
- Regarding the hardwired `1.65` boost version: `vw` needs to be built with the same boost version as conda, whatever that is. boost in notoriously version incompatible.  So by hardwiring, the script becomes brittle: when boost gets upgraded in `conda`, the script might break.
- Regarding the symlinking.  If `GXX` in not actually called `g++`, but you know what the full name is (since you link to it) - you could pass it (and PATH) down in the environment to the `vw` bulild like this:
```
    PATH=/the/conda/tools/dir:$PATH  CXX="$GXX" build command and args ...
```
(Note that `vw` uses `CXX` in all Makefiles rather than `GXX`).
For example, when I want to build `vw` with `clang++` and static linking, instead of `g++` and dynamic linking, I use (example):
```
         CXX='clang++ -static' make clean vw test
```
That said, I'm not very familiar with everything going on in the python `setup`/`pip` steps so that compiler env passing through the python layers may not work.

**Comment by mlisovyi** [comment]
@arielf the small issues (bullet-points 1--4 from your original proposal) have been implemented in the latest update. The bigger issues need to be iterated over for me to understand how to proceed.

**Comment by JohnLangford** [comment]
@Scott-Graham-Bose do you have ideas for how to address the hardcoded boost version issue?

**Comment by Scott-Graham-Bose** [comment]
no, unfortunately the necessary conda packaging doesn't allow for a lot of flexibility here as @mlisovyi described. i suppose the problem is if you change the boost version of conda in the future it could break the usage of vw. the plus side is this is isolated to the virtual environment used for this, so i think it's as good as we can get it for now.

**Comment by JohnLangford** [comment]
@arielf any ideas?  Otherwise, I expect we should merge---it seems better than before.

**Comment by arielf** [comment]
Sure we could merge.
I agree it is much better than before.

@mlisovyi  good catch: you're right the 1st line in the top `Makefile` unconditionally sets CXX to `which g++`   This should be fixed to:
```
CXX ?= $(shell which g++)
```
Which means any passing of `CXX=<some_value>` either via the environment or the make command:  `make CXX=NewValue ...` should override that setting.  `?=` in a gnu-make `Makefile` means set iff it is not yet set which is the right thing to do.

As for finding which version of boost is installed: this works for me (for the system boost) 
You may be able to do something similar for the `conda` boost:
```
grep BOOST_LIB_VERSION /usr/include/boost/version.hpp | grep -Eo '[0-9]+\_[0-9]+'  | tr  _  .
```

mlisovyi:
@arielf Thanks for detailed instructions. I appreciate it, as my Makefile-writing skills are close to zero :)
**The `$CXX` choice**: does everyone agree to implement the conditional compiler choice as suggested by @arielf? I'm just starting with VW and thus I do not know what was the historic reason to do it this way originally. I can switch to the proposal and try it out. This should be as easy as setting `$CXX`to `$GXX` instead of setting a symlink (conda environment will set something into `$CXX` once `gxx_linux-64` is installed, but we can overwrite it with `$GXX`)
**The boost version**: I can try to play those tricks installing different versions of different boost packages in conda. But this will take time, as I already forgot what was the actual problem with version mismatch. How about we merge the current version (I also want to port VW into kaggle kernels) and then I will try to make it work with scattered boost versions at a later stage and discuss it in a separate issue?

mlisovyi:
Ok, I see the `CXX` conditional setting is already submitted as a separate PR #1490. Could you please also push it to `mlisovyi:update_build_rules` once it is decided to merge that PR? Thanks in advance

JohnLangford:
Merged, thanks :-)

PR #1487, opened 2018-05-22
Update automake for hash moving into explore/

mnutt:
I noticed that the `Makefile`s had been updated to include the `explore` directory, but the makefile inputs had not. Also removes `vowpalwabbit/hash.h` which is now gone.


JohnLangford:
Merged thanks.  This is yet another example of how our somewhat confused build system causes issues.

PR #1488, opened 2018-05-22
get_action_score from c wrapper

mnutt:
This is opened as a PR but is more of a question as I have no idea what the best API for this might be. I was hoping to get the `action_scores` for a prediction via the c API (my app is in golang) and noticed that it was possible to get the data through c++ and python but not through c.

I was having trouble coming up with a better API that didn't involve having to allocate more memory. Is there some better way to do it with a `v_array` instead?


JohnLangford:
This is a reasonable interface.  The more C++ish version uses iterators as per this: 
https://github.com/rajan-chari/vowpal_wabbit/blob/7438e0a2ff0279f67115863239afdefa31bcbe43/decision_service/include/ranking_response.h but in a C interface, this seems reasonable.  

@rajan-chari @eisber comments on this?

vowpalwabbit/parser.cc
```diff
@@ -1088,6 +1088,20 @@ uint32_t* get_multilabel_predictions(example* ec, size_t& len)
...
 }
 
+float get_action_score(example* ec, size_t i)
+{
+  ACTION_SCORE::action_scores scores = ec->pred.a_s;
+
+  if(i < scores.size()) {
+    return scores[i].score;
+  } else {
+    return 0.0;
```
rajan-chari: Is there a way to signal to the user that index is out of bounds?

rajan-chari:
get_size() and get_item_by_idx(), the way you have it, seems like the simplest approach for a C interface.

JohnLangford:
Merged, thanks.

PR #1493, opened 2018-05-30
Fix libboost-name on MacOS

marsewe:


JohnLangford:
Is there a way to autodetect and do the right thing?

marsewe:
Determining the boost-library name on MacOS by 
`boost_lib = 'boost_python' + str(sys.version_info[0]) + str(sys.version_info[1]) + '-mt' `
works with recent versions of "brew", so with `brew install boost-python3` or "brew install boost-python" (for Python 2).
Personally I do not know how to detect on a Mac whether boost-python(3) was installed via "brew" or e.g. via Anaconda.

JohnLangford:
Ok, thanks.

PR #1495, opened 2018-06-04
Reinforcement Learning model management

rajan-chari:
Changes to reinforcement learning library
1) background thread to download models
2) updated factory pattern with static factory objects
3) cleaner error reporting with less repetitive code
4) refactored some code into different files and namespaces


decision_service/include/constants.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+namespace reinforcement_learning {
+  using str_const = const char * const;
```
eisber: >using str_const = const char * const; [](start = 2, length = 37)

why not just use the type below?

decision_service/include/err_constants.h
```diff
@@ -0,0 +1,44 @@
+#pragma once
+
+namespace reinforcement_learning { namespace error_code {
+  //success code
+  const int success = 0;
+
+  //error code
+  const int invalid_argument            = 1;
```
eisber: >const int invalid_argument            = 1; [](start = 2, length = 42)

can we introduce some macro so that the error codes and strings are correctly mapped and we have some consistent mapping?

decision_service/include/factory_resolver.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+#include "model_mgmt.h"
+#include "object_factory.h"
+
+namespace reinforcement_learning  {
+  using dtfactory = utility::object_factory<model_management::i_data_transport>;
+  using modelfactory = utility::object_factory<model_management::i_model>;
+
+  extern dtfactory& data_transport_factory;
```
eisber: >extern dtfactory& data_transport_factory; [](start = 2, length = 41)

why extern?

decision_service/include/model_mgmt.h
```diff
@@ -0,0 +1,32 @@
...
+    class i_data_transport{
+    public:
+      virtual int get_data(model_data& data, api_status* status = nullptr) = 0;
+      virtual ~i_data_transport() {}
+    };
+
+    class i_model {
+    public:
+      virtual int init(model_data& data, api_status* status = nullptr) = 0;
+      virtual int choose_rank(int& action, char* features, int actions[], api_status* status = nullptr) = 0;
```
eisber: >int& action [](start = 30, length = 11)

what's that int?

decision_service/include/object_factory.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+#include "config_collection.h"
+#include "api_status.h"
+#include <functional>
```
eisber: why in the middle? can we do include <..> include "..." afterwards?

decision_service/include/constants.h
rajan-chari: I wanted to declare str_const in the reinforcement_learning namespace and use it in reinforcement_learning::name and reinforcement_learning::value namespaces.  Not sure if that is your question.

---
In reply to: [193405380](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193405380) [](ancestors = 193405380)

decision_service/include/factory_resolver.h
rajan-chari: Every cpp module that needs to refer to data_transport_factory will include this file but it will only be instantiated in one cpp module.  

---
In reply to: [193405978](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193405978) [](ancestors = 193405978)

decision_service/include/object_factory.h
rajan-chari: Changed

---
In reply to: [193420010](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193420010) [](ancestors = 193420010)

decision_service/include/model_mgmt.h
rajan-chari: Not needed.  Changed.

---
In reply to: [193406164](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193406164) [](ancestors = 193406164)

decision_service/include/constants.h
eisber: I was questioning if the typedef makes it more readable... maybe if it is that close, it's fine... 

---
In reply to: [193425684](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193425684) [](ancestors = 193425684,193405380)

decision_service/include/factory_resolver.h
eisber: so how does this work if we have multiple instances of the DS in the same process. Does the transport factory need to be the same? This looks like a big singleton too me.

---
In reply to: [193427006](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193427006) [](ancestors = 193427006,193405978)

decision_service/rlclientlib/bg_model_download.h
```diff
@@ -0,0 +1,21 @@
...
+  class bg_model_download {
+  public:
+    bg_model_download(i_data_transport* ptrans, data_callback_fn* pdata_cb);
+    ~bg_model_download();
+    bg_model_download(bg_model_download&& temp) noexcept;
+    bg_model_download& operator=(bg_model_download&& temp) noexcept;
+
+    int run_once(api_status* status) const;
+  private:
+    i_data_transport* _ptrans;
```
eisber: >i_data_transport* _ptrans; [](start = 4, length = 26)

maybe use unique_ptr and safe and avoid the delete in the destructor?

decision_service/rlclientlib/bg_model_download.cc
```diff
@@ -0,0 +1,32 @@
...
+      _ptrans      = temp._ptrans;
+      temp._ptrans = x;
+    }
+    return *this;
+  }
+
+  int bg_model_download::run_once(api_status* status) const {
+    model_data md;
+    const auto scode = _ptrans->get_data(md, status);
+    TRY_OR_RETURN(scode);
```
eisber: can you put the statement into the macro?

decision_service/rlclientlib/bg_model_download.cc
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  bg_model_download& bg_model_download::operator=(bg_model_download&& temp) noexcept {
+    if (&temp != this) {
+      const auto x = _ptrans;
+      _ptrans      = temp._ptrans;
+      temp._ptrans = x;
+    }
+    return *this;
+  }
```
eisber: are those used or just for good measure?

decision_service/rlclientlib/data_callback_fn.h
```diff
@@ -0,0 +1,22 @@
...
+    ~data_callback_fn() = default;
+
+    data_callback_fn(const data_callback_fn&) = delete;
+    data_callback_fn(data_callback_fn&&) = delete;
+    data_callback_fn& operator=(const data_callback_fn&) = delete;
+    data_callback_fn& operator=(data_callback_fn&&) = delete;
+
+  private:
+    data_fn _fn;
+    void* _context;
```
eisber: >void* _context; [](start = 4, length = 15)

should we use shared_ptr to make lifespan management of the context easier?

decision_service/rlclientlib/factory_resolver.cc
```diff
@@ -0,0 +1,70 @@
...
+namespace reinforcement_learning { 
+  namespace m = model_management;
+  namespace u = utility;
+  // For proper static intialization 
+  // Check https://en.wikibooks.org/wiki/More_C++_Idioms/Nifty_Counter for explanation
+  static int init_guard;  // guranteed to be zero when loaded
+
+  // properly aligned memory for the factory object 
+  template <typename T>
+  using natural_align = std::aligned_storage<sizeof(T), alignof ( T )>;
```
eisber: why is this needed?

decision_service/rlclientlib/factory_resolver.cc
```diff
@@ -0,0 +1,70 @@
...
+
+  static natural_align<dtfactory>::type dtfactory_buf;
+  static natural_align<modelfactory>::type modelfactory_buf;
+
+  // reference should point to the allocated memory to be initalized by placement new in factory_initializer::factory_initializer()
+  dtfactory& data_transport_factory = (dtfactory&)( dtfactory_buf );
+  modelfactory& model_factory = (modelfactory&)( modelfactory_buf );
+
+  factory_initializer::factory_initializer() {
+    if ( init_guard++ == 0 ) {
```
eisber: >if ( init_guard++ == 0 ) { [](start = 4, length = 26)

can't we just do smart pointers?

decision_service/rlclientlib/factory_resolver.cc
lokitoth: It looks like a ref-count for the purposes of initializing/destructing the singleton. My question is - why not use a shared pointer or similar for this?

decision_service/rlclientlib/factory_resolver.cc
```diff
@@ -0,0 +1,70 @@
...
+    if ( uri == nullptr ) {
+      api_status::try_update(status, error_code::http_uri_not_provided, error_code::http_uri_not_provided_s);
+      return error_code::http_uri_not_provided;
+    }
+    auto pret = new m::restapi_data_tranport(uri);
+    const auto scode = pret->check(status);
+    if(scode != error_code::success) {
+      delete pret;
+      return scode;
+    }
```
eisber: the complete block feels a bit weird. why doesn't the transport not invoke check and return the status?

decision_service/rlclientlib/periodic_bg_proc.h
```diff
@@ -0,0 +1,90 @@
...
+#include "interruptable_sleeper.h"
+
+namespace reinforcement_learning {
+  class error_callback_fn;
+}
+
+namespace reinforcement_learning { namespace utility {
+
+  template<typename BGProc>
+  class periodic_bg_proc {
```
eisber: >periodic_bg_proc [](start = 8, length = 16)

can you rename to periodic_background_proc?

decision_service/rlclientlib/periodic_bg_proc.h
```diff
@@ -0,0 +1,90 @@
...
+    if ( _thread_is_running ) {
+      _thread_is_running = false;
+      _sleeper.interrupt();
+      _background_thread.join();
+    }
+  }
+
+  template <typename BGProc>
+  periodic_bg_proc<BGProc>::~periodic_bg_proc() {
+    stop();
```
eisber: do you think this should join the background_thread too?

decision_service/rlclientlib/periodic_bg_proc.h
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  template <typename BGProc>
+  void periodic_bg_proc<BGProc>::time_loop() {
+    while ( _thread_is_running ) {
+      api_status status;
+      // Run the background task once
+      if ( _proc.run_once(&status) != error_code::success ) {
```
eisber: > _proc.run_once(&status) [](start = 10, length = 24)

I'm unclear on the naming. run_once sounds to me like a method that should only ever run once per lifetime of the object. since it's in a loop it seems like it's executed many times?

decision_service/rlclientlib/data_callback_fn.h
rajan-chari: This is provider by the user of the class and returned to the user during callback.  Not owned by the callback.

---
In reply to: [193485951](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193485951) [](ancestors = 193485951)

decision_service/rlclientlib/restapi_data_transport.cc
```diff
@@ -1,10 +1,138 @@
...
+    ::utility::datetime curr_last_modified;
+    ::utility::size64_t curr_datasz;
+    auto err = get_data_info(curr_last_modified, curr_datasz, status);
+    TRY_OR_RETURN(err);
+
+    if ( curr_last_modified == _last_modified && curr_datasz == _datasz )
+      return error_code::success;
+
+    // Build request URI and start the request.
+    pplx::task<int> requestTask = _httpcli.request(methods::GET)
```
eisber: I think you get rid of it of the get_data_info call using HEAD and better use this: [https://docs.microsoft.com/en-us/rest/api/storageservices/specifying-conditional-headers-for-blob-service-operations](https://docs.microsoft.com/en-us/rest/api/storageservices/specifying-conditional-headers-for-blob-service-operations)

decision_service/rlclientlib/periodic_bg_proc.h
rajan-chari: My goal is to change the other background thread to use the same timed thread infrastructure.  run_once() runs once during each iteration of the loop.  The class that implments run_once() can be tested without the background thread.  

---
In reply to: [193489575](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193489575) [](ancestors = 193489575)

decision_service/rlclientlib/restapi_data_transport.cc
```diff
@@ -1,10 +1,138 @@
...
+      curr_last_modified = ::utility::datetime::from_string(iter->second);
+      if ( curr_last_modified.to_interval() == 0 )
+        RETURN_ERROR(status, error_code::last_modified_invalid, error_code::last_modified_invalid_s);
+
+      curr_datasz = response.headers().content_length();
+      if ( curr_datasz <= 0 )
+        RETURN_ERROR(status, error_code::bad_content_length, error_code::bad_content_length_s);
+
+      const auto buff = new char[curr_datasz];
+      ret.data = buff;
```
eisber: what happens to this memory in case response.body().read_to_end(rb).get throws?

decision_service/rlclientlib/vw_model.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+#include "model_mgmt.h"
+
+namespace reinforcement_learning { namespace model_management {
+  class vw_model : public i_model{
+  public:
+    int init(model_data& data, api_status* status = nullptr) override;
+    int choose_rank(int& action, char* features, int actions[], api_status* status = nullptr) override;
```
eisber: >int& action, [](start = 20, length = 12)

what is this parameter?

decision_service/include/factory_resolver.h
rajan-chari: Yes that's right.  This is a c++ stdlib technique for singletons.    The intention is to allow users to register new transports, models etc before using client lib.

---
In reply to: [193484768](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193484768) [](ancestors = 193484768,193427006,193405978)

decision_service/rlclientlib/vw_model.h
rajan-chari: It was but I changed it after looking at the vw interface.  No longer there.

---
In reply to: [193495501](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193495501) [](ancestors = 193495501)

decision_service/rlclientlib/logger/async_batcher.h
```diff
@@ -57,11 +58,9 @@ namespace reinforcement_learning {
...
       catch(const std::exception& e) {
         _thread_is_running = false;
-        std::ostringstream os;
-        os  << "Unable to start background thread to log events. "
-            << e.what();
-        api_status::try_update(status, error_code::background_thread_start, os.str().c_str());
-        return error_code::background_thread_start;
+        return report_error(status, error_code::background_thread_start,
+                            "Unable to start background thread to log events. ",
+                            e.what());
```
eisber: I looked at the macro and it looks fancy, but the usage is kind of unexpected. can't we do something like report_error(status, error_code::...) << "Unable to start.." << e.what()? which looks more natural?

decision_service/rlclientlib/bg_model_download.cc
rajan-chari: not sure what you mean.  what statement are you thinking about?



---
In reply to: [193485345](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193485345) [](ancestors = 193485345)

decision_service/rlclientlib/bg_model_download.cc
rajan-chari: I might have used it in some tests at some point.  

---
In reply to: [193485495](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193485495) [](ancestors = 193485495)

decision_service/include/err_constants.h
rajan-chari: Many of the errors are not mapped to static strings.  What did you have in mind?  can you provide an example?

---
In reply to: [193405817](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193405817) [](ancestors = 193405817)

decision_service/rlclientlib/factory_resolver.cc
rajan-chari: normally new allocates properly aligned memory and invokes the constructor.  This template is used to make sure that statically memory allocated for factory objects have the right alignment for that object.

---
In reply to: [193486212](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193486212) [](ancestors = 193486212)

decision_service/rlclientlib/factory_resolver.cc
rajan-chari: smart pointers do reference counted de-allocation.  the use case here is around order of initialization of static object.  I don't think it applies.

---
In reply to: [193487698](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193487698) [](ancestors = 193487698)

decision_service/rlclientlib/factory_resolver.cc
rajan-chari: This is a technique used to initialize static objects because the initialization order is not guaranteed across translation units.  This an issue for example if the users of this library decide to use this object inside their own global object.

For example std::cout uses this method for initializing.  

more details here: [https://en.wikibooks.org/wiki/More_C++_Idioms/Nifty_Counter](https://en.wikibooks.org/wiki/More_C++_Idioms/Nifty_Counter)


---
In reply to: [193488308](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193488308) [](ancestors = 193488308)

decision_service/rlclientlib/factory_resolver.cc
rajan-chari: Although I am not sure what you are referring to in this file location...

I found that exceptions are not propagated as with value based continuations.  I switched to task based continuations everywhere now.  Which means I explicitly get the task and check the return value.  

---
In reply to: [193488341](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193488341) [](ancestors = 193488341)

decision_service/rlclientlib/bg_model_download.h
rajan-chari: although destructor would be simpler (no delete), move constructor and assignment would be more complicated... no?

---
In reply to: [193485057](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193485057) [](ancestors = 193485057)

decision_service/rlclientlib/factory_resolver.cc
rajan-chari: If you mean why doesn't the constructor return a status if it did not get constructed properly, that might lead to potential usage of a poorly constructed object.  

---
In reply to: [193488341](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193488341) [](ancestors = 193488341)

decision_service/rlclientlib/periodic_bg_proc.h
rajan-chari: Done

---
In reply to: [193488786](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193488786) [](ancestors = 193488786)

decision_service/rlclientlib/logger/async_batcher.h
rajan-chari: Yes thought about that, but report_error() would have to return a reference to object with overloaded << operator but then it also needs to return an error code.

---
In reply to: [193497746](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193497746) [](ancestors = 193497746)

decision_service/rlclientlib/bg_model_download.h
rajan-chari: Moved out ownership of these pointers (in the next PR)

---
In reply to: [193517810](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193517810) [](ancestors = 193517810,193485057)

decision_service/rlclientlib/bg_model_download.cc
rajan-chari: Done

---
In reply to: [193499415](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193499415) [](ancestors = 193499415,193485345)

decision_service/unit_test/model_mgmt_test.cc
```diff
@@ -5,76 +5,177 @@
...
+#include "periodic_bg_proc.h"
+#include "bg_model_download.h"
+#include "data_callback_fn.h"
+
+namespace r = reinforcement_learning;
+namespace m = reinforcement_learning::model_management;
+namespace u =  reinforcement_learning::utility;
+namespace e = reinforcement_learning::error_code;
 
 const char * A_BLOB_URI   = "https://storagex6ilrr3obwyra.blob.core.windows.net/mwt-models/current?sv=2017-04-17&sr=b&sig=n9udcFiyjhahgnK38elMIEbg9FYWfmmAWte1MP87v1s%3D&st=2018-03-19T17%3A51%3A18Z&se=2028-03-19T17%3A52%3A18Z&sp=r";
```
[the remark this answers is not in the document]
rajan-chari: Done.

---
In reply to: [193497377](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193497377) [](ancestors = 193497377)

decision_service/rlclientlib/restapi_data_transport.cc
rajan-chari: Good point.  Fixed

---
In reply to: [193495029](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193495029) [](ancestors = 193495029)

decision_service/rlclientlib/restapi_data_transport.cc
rajan-chari: Will add a workitem for this and schedule.

---
In reply to: [193494507](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193494507) [](ancestors = 193494507)

decision_service/rlclientlib/periodic_bg_proc.h
rajan-chari: Change to run_iteration()

---
In reply to: [196400208](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r196400208) [](ancestors = 196400208,193494540,193489575)

decision_service/include/factory_resolver.h
```diff
@@ -0,0 +1,23 @@
...
+  struct factory_initializer {
+    factory_initializer();
+    ~factory_initializer();
+    private:
+    static void register_default_factories();
+  };
+  // Every translation unit gets a factory_initializer
+  // only one translation unit will initialize it
+  static factory_initializer _init;
+}
```
rajan-chari: Consider passing in factories to rlclientlib.

decision_service/include/constants.h
rajan-chari: Done

---
In reply to: [193484325](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193484325) [](ancestors = 193484325,193425684,193405380)

decision_service/include/factory_resolver.h
rajan-chari: I thought about what you said some more.  

I will add a constructor to rlclientlib that takes factories in addition to what's there.  When they are not provided, I will use the default factories.  This way the simple case is simple and the complex case is still possible.

---
In reply to: [193496818](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r193496818) [](ancestors = 193496818,193484768,193427006,193405978)

decision_service/include/factory_resolver.h
rajan-chari: Will add a constructor that takes factories for the complex case.  The simple case will use the default factories.

---
In reply to: [196488055](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r196488055) [](ancestors = 196488055)

decision_service/rlclientlib/periodic_bg_proc.h
rajan-chari: Done

---
In reply to: [196476514](https://github.com/JohnLangford/vowpal_wabbit/pull/1495#discussion_r196476514) [](ancestors = 196476514,196400208,193494540,193489575)

JohnLangford:
Merged, thanks :-)

PR #1499, opened 2018-06-18
RegCB exploration algorithms

albietz:
Implementations of the RegCB algorithms of Foster et al., 2018 (https://arxiv.org/abs/1803.01088).


vowpalwabbit/learner.h
```diff
@@ -63,7 +63,7 @@ struct learn_data
 };
 
 struct sensitivity_data
-{ using fn = float(*)(void* data, base_learner& base, void* ex);
+{ using fn = float(*)(void* data, base_learner& base, example& ex);
```
JohnLangford: Is there an inherent reason to assume that sensitivity only applies to example and not multi_ex?

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -483,11 +678,13 @@ base_learner* cb_explore_adf_setup(arguments& arg)
     data->explore_type = EXPLORE_FIRST;
   else if (arg.vm["softmax"].as<bool>())
     data->explore_type = SOFTMAX;
+  else if (arg.vm["regcb"].as<bool>() || arg.vm["regcbopt"].as<bool>())
+    data->explore_type = REGCB;
```
JohnLangford: Can you add a test case for this?

JohnLangford:
I'm not sure what the test fail is.   After you address the other comments, let's investigate that more deeply.

vowpalwabbit/learner.h
albietz: At least from my experience implementing RegCB and active e-greedy, I do not see a generic way to define sensitivity for multi_ex, and even if we wanted it to be algorithm-specific, it should probably return a vector instead of a single float, meaning that the `void*` approach still wouldn't be suitable. There might be a better way to approach things long term, but this seemed like a simpler fix for now.

JohnLangford:
Ok, sounds reasonable.

albietz:
Regarding the test fail, basically having `--cb_type dr --cb_type ips` should fail but it does not fail anymore. It seems like it's due to this line:
https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parser_helper.cc#L138
which doesn't always return the duplicate option, since the vm contains default options as well (in this case it returns the new `cb_max_cost` option instead of `cb_type`, since it is first alphabetically). Is there a way to obtain a vm only for provided options, excluding default ones?

JohnLangford:
FYI, I'm working with this patch now and have resolved several issues.

Is there a good reason to not have randomtie be always on?

albietz:
Not much, other than the fact that randomtie doesn't seem to be necessary for cover, and I've seen it slightly degrade the performance of cover on a few datasets. But I guess having a disable option instead of enable, or just having it always on except for cover, could work.

JohnLangford:
I merged my version, Thanks :-)

Issue #1484, opened 2018-05-22
Bug constructing label from example in pyvw.py


aadyotb:
In the `from_example()` method of the classes `cost_sensitive_label` and `cbandits_label` in `pyvw.py`, you have the following for loops on lines 465 and 495, respectively:
```
for i in range(ex.get_costsensitive_num_costs):
...
```
and
```
for i in range(ex.get_cbandits_num_costs):
...
```

This results in a crash because the objects in the `range(...)` are methods, not integers. An easy fix is to just replace these calls with
```
for i in range(ex.get_costsensitive_num_costs()):
...
```
and
```
for i in range(ex.get_cbandits_num_costs()):
...
```
respectively.

JohnLangford:
@Scott-Graham-Bose your thoughts?   A patch is welcome.

Scott-Graham-Bose:
yep, that's an easy fix

gramhagen:
should be all set now, thanks @aadyotb

JohnLangford:
Thanks @gramhagen and @aadyotb .

PR #1503, opened 2018-06-22
fixing pyvw range bug and cleaning up learning with lists of examples

gramhagen:
addressing #1484 

but during testing I discovered two other pre-existing bugs in how predict and learn was being called on lists of examples, so I cleaned up the logic there

and some small adjustment to setup.py to offer a bit more robustness in finding boost-python library for mac osx


JohnLangford:
Merged, thanks :-)

PR #1530, opened 2018-07-13
enforce all boost packages to come from conda-forge

mlisovyi:
This is a tiny fix, that addresses that on some systems different boost packages will be pulled from different conda channels/repositories and for some reason vw build fails, due to  `boost_program_options` not being found in linking.

### DIFF
--- a/python/conda_install.sh
+++ b/python/conda_install.sh
@@ -13,7 +13,7 @@ source activate ${CONDA_DEFAULT_ENV}
 
 # boost packages
 BOOST_V="1.65"
-conda install --yes boost=$BOOST_V libboost=$BOOST_V py-boost=$BOOST_V
+conda install -c conda-forge --yes boost=$BOOST_V libboost=$BOOST_V py-boost=$BOOST_V
 
 # make a soft link to the compiler, since Makefiles internally use `which g++`
 if [ ! -z "${GXX}" ]; then


JohnLangford:
@Scott-Graham-Bose this looks good to me, but can you take a quick look?

Scott-Graham-Bose:
i did a bit of rework to the build system which impacts conda installs #1533 

i actually removed the conda install script in favor of instructing users to manually install boost using the anaconda channel
```conda install -y -c anaconda boost```

this provides all the boost dependencies needed under mac osx and linux, and I updated setup.py to set the appropriate environment variables to use the correct libraries. I tested it under a few systems, but would be great to get someone else to verify. If that's all working then we shouldn't need to worry about this change.

**Comment by mlisovyi** [comment]
@Scott-Graham-Bose that sounds promising. When the install script was under development, `conda install -y -c anaconda boost` alone was not enough to get all needed packages in the desired version. But I'm happy to hear you see no issue there. Let's discuss in the PR that you created

**Comment by JohnLangford** [comment]
Is this patch obsolete?

**Comment by gramhagen** [comment]
I've incorporated this into #1533 by putting the conda_install script back in place and updating to use conda-forge. There's no reason to remove the script if it's working for folks.

However, I think if people are using mac or ubuntu and install anaconda then relying on boost from the anaconda channel is still simpler and effective (this however doesn't seem to work when using the anaconda base docker image). So I'll keep the documentation as is.

Example docker file I used to test this for reference:
```
FROM ubuntu:16.04

RUN apt-get update && \
    apt-get install -y build-essential \
    curl \
    make \
    git
    
RUN curl -O https://repo.anaconda.com/archive/Anaconda3-5.2.0-Linux-x86_64.sh && \
    bash Anaconda3-5.2.0-Linux-x86_64.sh -b

ENV PATH /root/anaconda3/bin:$PATH

RUN conda create -n vw -y -c anaconda boost

RUN git clone https://github.com/gramhagen/vowpal_wabbit -b 1528_python_boost
```
docker build -t vw .
docker run -it vw /bin/bash
> source activate vw
> pip install -e vowpal_wabbit/python
> python -c 'from vowpalwabbit import pyvw; print(pyvw.vw(quiet=True))'
(<vowpalwabbit.pyvw.vw object at 0x7fe2e8559490>)

JohnLangford:
Ok, I'm closing.  Thanks @mlisovyi and @Scott-Graham-Bose

PR #1531, opened 2018-07-13
First two arguments are the wrong way around

jackgerrits:
- Fix ordering of context and uuid arguments
- Clean up some trailing whitespace while I'm in the file


JohnLangford:
@rajan-chari can you examine?

rajan-chari:
Yes makes sense.  Spoke to @jackgerrits earlier.

Issue #1475, opened 2018-05-02
[Python] [sklearn] A few issues (and minor bugs?) in the sklearn_vw.VW.fit() method


mlisovyi:
Hi,

playing around with the sklearn interface of VW it seems there a few issues (applies to v 8.5.0 and the current master branch):
1. A **very minor bug**: it seems that the code intends to shuffle examples between epochs/passes. This is done [here in the code](https://github.com/JohnLangford/vowpal_wabbit/blob/master/python/vowpalwabbit/sklearn_vw.py#L307). However, `range(n)` starts with `0`, so the `n > 1` condition has to be changed to `n >= 1` otherwise the second pass is done without reshuffling.
2. An **issue/bug** (?): the current implementation of shuffling doesn't work in a general case for pd.DataFrame or pd.Series as an input. It will work if all indices are sequential starting from 0. However, if one attempts to use a KFold splitting from sklearn, then it breaks down. A minimal example to reproduce the problem is given at the end of the post. This feature can be worked around by passing `pd.Series.values` but at the moment it is not clear from the documentation that pandas objects directly have only a limited support. Pandas support also relates to observations in #1472.
3. An **issue**: even if a `np.array` is passed directly, the `VW.fit()` with several passes shuffles examples in the original array. The problem with this is that one wants to evaluate performance on tthe training dataset, thus doing `clf.predict()` and to compare to a vector (=array/pd.Series/etc.) true labels. True labels are usually available separately in a sklearn nomenclature (i.e. 0/1 labels instead of -1/1). However, those have not been shuffled and thus the order of predictions and true labels is broken and performance metric evaluation doesn't make sense. Is it possible to create a local copy of the input data object inside the `.fit()` method and clear it at the end? Otherwise, the documentation has to be very explicit about this behaviour.

Thanks in advance for feedback

**Environmental setup**
Ubuntu 16.04 with conda 4.5.1 and python3.5
vowpalwabbit 8.5.0 (installed from pypi), scikit-learn 0.19.1, numpy 1.14.2

**Example of the issue 2**
```python
import pandas as pd
import numpy as np
from vowpalwabbit.sklearn_vw import VWClassifier
from sklearn.model_selection import KFold

# The dummy example X has already been adjusted to the native VW format, so convert_to_vw=False
X = pd.Series(['1 |Pet cat', '-1 |Pet dog', '1 |Pet cat', '1 |Pet cat'], name='catdog')
[...]

PR #1532, opened 2018-07-13
[python] Fix shuffle between passes and add a test for it

mlisovyi:
Addresses #1475. 
Feedback from @Scott-Graham-Bose is welcome.

### DIFF
--- a/python/tests/test_sklearn_vw.py
+++ b/python/tests/test_sklearn_vw.py
@@ -186,6 +186,17 @@ class TestVWClassifier:
 
         assert np.allclose(expected, actual)
 
+    def test_shuffle_list(self):
+        # dummy data in vw format
+        X = ['1 |Pet cat', '-1 |Pet dog', '1 |Pet cat', '1 |Pet cat']
+
+        # Classifier with multiple passes over the data
+        clf = VWClassifier(passes=3, convert_to_vw=False)
+        clf.fit(X)
+
+        # assert that the dummy data was not perturbed
+        assert X == ['1 |Pet cat', '-1 |Pet dog', '1 |Pet cat', '1 |Pet cat']
+
     def test_shuffle_pd_Series(self):
         # dummy data in vw format
         X = pd.Series(['1 |Pet cat', '-1 |Pet dog', '1 |Pet cat', '1 |Pet cat'], name='catdog')
--- a/python/tests/test_sklearn_vw.py
+++ b/python/tests/test_sklearn_vw.py
@@ -2,10 +2,12 @@ import os
 
 from collections import namedtuple
 import numpy as np
+import pandas as pd
 import pytest
 
 from vowpalwabbit.sklearn_vw import VW, VWClassifier, VWRegressor, tovw
 from sklearn import datasets
+from sklearn.model_selection import KFold
 from scipy.sparse import csr_matrix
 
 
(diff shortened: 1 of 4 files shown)


python/vowpalwabbit/sklearn_vw.py
```diff
@@ -340,7 +343,7 @@ def predict(self, X):
             if not using convert_to_vw, X is expected to be a list of vw formatted feature vector strings with labels
 
         Returns
-        -------
+        -------hon/vowpalwabbit/sklearn_vw.py#L307
```
Scott-Graham-Bose: something weird happened here

python/tests/test_sklearn_vw.py
```diff
@@ -184,6 +186,20 @@ def test_decision_function(self, data):
...
+
+        kfold = KFold(n_splits=3, random_state=314, shuffle=False)
+        for train_idx, valid_idx in kfold.split(X):
+            X_train = X[train_idx]
+            # Classifier with multiple passes over the data
+            clf = VWClassifier(passes=3, convert_to_vw=False)
+            # Test that there is no exception raised in the fit on folds
+            try:
+                clf.fit(X_train)
+            except KeyError:
```
Scott-Graham-Bose: catching exceptions is not really necessary under pytest, if an exception occurs it will be caught and the test marked as a failure

python/tests/test_sklearn_vw.py
```diff
@@ -184,6 +186,20 @@ def test_decision_function(self, data):
 
         assert np.allclose(expected, actual)
```
Scott-Graham-Bose: would be good to see a test of normal list input, something like this
```
    def test_shuffle_list(self):
        # dummy data in vw format
        X = ['1 |Pet cat', '-1 |Pet dog', '1 |Pet cat', '1 |Pet cat']

        # Classifier with multiple passes over the data
        clf = VWClassifier(passes=3, convert_to_vw=False)
        clf.fit(X)

        # assert that the dummy data was not perturbed
        assert X == ['1 |Pet cat', '-1 |Pet dog', '1 |Pet cat', '1 |Pet cat']
```

python/tests/test_sklearn_vw.py
mlisovyi: OK, but the intention was to clarify with the error message what is the cause of the problem (as otherwise the exception itself is hard to trace). Did I fail to make an instructive message?

python/tests/test_sklearn_vw.py
mlisovyi: Makes sense. Copied over the test suggestion.

python/tests/test_sklearn_vw.py
Scott-Graham-Bose: it's fine, it's just a level of rigor in the test harness that I've never seen before =)
under pytest it's easy to trace the problem back to this test, but we can keep it here.

Scott-Graham-Bose:
looks good to merge

mlisovyi:
It seems that tests on travis fail due to missing pandas installation:
https://travis-ci.org/JohnLangford/vowpal_wabbit/builds/404370272#L5390-L5398
@Scott-Graham-Bose could you please point me where one defines required packages for installation? Thanks in advance!

Scott-Graham-Bose:
since this is only required during testing you can make the change here:
https://github.com/JohnLangford/vowpal_wabbit/blob/543dc5e023932146829b501be73ac32824820599/.travis.yml#L45

mlisovyi:
OK, Travis is happy after adding `pandas` to the list of packages to be installed. C# and Java fails are not related to these changes, I believe

JohnLangford:
Merged.  Thanks @mlisovyi and @Scott-Graham-Bose :-)

PR #1544, opened 2018-07-23
Changes to fix Make test under python 2 and 3

Scott-Graham-Bose:
Updates:
save_resume_test.py script to work under python 2 and 3
parser_helper to recognize uint64 arg types in oaa and ect models
fixing test reference for oaa inverse hash check (tests/RunTests:1253)


JohnLangford:
For unclear reasons, this lacks an 'update to master' button, so I merged it manually.  Thanks :-)

PR #1549, opened 2018-07-26
1) report_outcome() should log in background thread. 2) Bug fixes and cleanups

rajan-chari:
1) Move report outcome to background thread.  (feature)
2) Use object pool for buffer since this can be accessed from multiple threads.  (thread safety)
3) std::ostringstream reset does not work correctly. Replaced with data_buffer (bug fix)
4) use size_t for action id since it indexes an array.  (clean up compile)


rajan-chari:
Removed basic_usage_cpp.out

reinforcement_learning/examples/rl_sim_cpp/rl_sim.cc
```diff
@@ -157,7 +161,7 @@ rl_sim::rl_sim(boost::program_options::variables_map vm) :_options(std::move(vm)
 std::string get_dist_str(const reinforcement_learning::ranking_response& response) {
   std::string ret;
   ret += "(";
-  for (auto& ap_pair : response) {
+  for (auto ap_pair : response) {
```
ataymano: const auto& ?

rajan-chari: Makes sense.

reinforcement_learning/rlclientlib/utility/data_buffer.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+#include <string>
+#include <vector>
+
+namespace reinforcement_learning { namespace utility {
+  class data_buffer {
+  public:
+
+    void reset();
+    std::string str();
```
ataymano: const?

rajan-chari: Makes sense.

PR #1551, opened 2018-07-26
Fixed bug in CLI parsing of --csoaa_ldf multiline

marco-rossi29:


JohnLangford:
Tests are barfing on this: https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/csoaa.cc#L850

We need to either make multiline explicit or rework the if/then/else branching.

marco-rossi29:
Based on the tests it seems that "multiline" is required. I would just remove commit de09b0b to preserve minimal change. Do you agree?

JohnLangford:
Sounds good.

(It seems to be required by the if/then structure but not by the defaults
variable values....)

-John

PR #1559, opened 2018-08-03
Extract stable_unique to own function for clarity 

jackgerrits:


jackgerrits:
@JohnLangford I would like to add a test to verify stable_unique functions as expected, but it doesn't seem like there are any unit tests for the cpp code at the moment. Do you have a preferred way I can set this up?

JohnLangford:
Why do we want to extract the stable_unique function into a .h file?  For a function which is only used once, I tend to either leave it inline or (for clarity reasons) extract it into a separate function within the same file.

jackgerrits:
Putting it in a separate file allows us to easily test it and means it can be easily consumed in future elsewhere in the lib. Considering how generic this is I think it's worth it

PR #1562, opened 2018-08-08
rl client tests fix

ataymano:
Not fix yet, but error message improvement.

### DIFF
--- a/reinforcement_learning/unit_test/live_model_test.cc
+++ b/reinforcement_learning/unit_test/live_model_test.cc
@@ -36,7 +36,7 @@ BOOST_AUTO_TEST_CASE(live_model_ranking_request)
 {
 	//start a http server that will receive events sent from the eventhub_client
 	http_helper http_server;
-	http_server.on_initialize(U("http://localhost:8080"));
+	BOOST_CHECK(http_server.on_initialize(U("http://localhost:8080")));
   r::api_status status;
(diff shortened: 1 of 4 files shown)


reinforcement_learning/unit_test/live_model_test.cc
```diff
@@ -74,17 +74,14 @@ BOOST_AUTO_TEST_CASE(live_model_ranking_request)
 	ds.choose_rank(uuid, JSON_CONTEXT, response, &status);
 	BOOST_CHECK_EQUAL(status.get_error_code(), 0);
```
jackgerrits: This will send off the ranking request to the logger, I think we should put a wait in before tearing down the webserver

jackgerrits: In fact that may be the issue, that the log request is not able to make it to the web server because it is torn down almost instantly.

jackgerrits: Since the async_batcher destructor flushes to the network this change should fix it 👍

JohnLangford: Is a wait required here?  I'd like to avoid one if possible as they typically are really covering up for a wait condition.

JohnLangford:
Merged, thanks :-)

PR #1584, opened 2018-08-29
Add build + install instructions to Python binding readme

jackgerrits:


marco-rossi29:
UnitTest: TestString Failed, very weird as pr adds only comments! @jackgerrits Any ideas?

```
Failed   TestString
Passed   TestStringFeatureGroup
Passed   TestStringNamespace
Passed   TestStringEscape
Passed   TestStringSplit
Passed   TestStringIncludeName
Passed   TestDictionary
Passed   TestCustomType
Passed   TestEnumerableString
At line:4 char:4
+    vstest.console /Platform:x64 /inIsolation vowpalwabbit\x64\Release ...
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
   Assert.IsFalse failed. 
Stack Trace:
   at cs_unittest.VowpalWabbitExampleValidator`1.Validate(String line, TExample example, ILabel label) in C:\vowpal_wabbit\cs\unittest\VowpalWabbitExampleValidator.cs:line 122
   at cs_unittest.TestMarshalling.TestString() in C:\vowpal_wabbit\cs\unittest\TestMarshalling.cs:line 39
```

jackgerrits:
I am not familiar with the test in question, but given that master is passing and this PR doesn't modify any code then the test must be at least a little flaky. I'm going to rerun.

marco-rossi29:
Test passed! Ready to be merged!

reinforcement_learning/bindings/python/README.md
```diff
@@ -1 +1,14 @@
+# Python Bindings
+
+## Build + Install
+1. [Install pip for python3.6](https://askubuntu.com/questions/889535/how-to-install-pip-for-python-3-6-on-ubuntu-16-10)
+2. `sudo python3.6 -m pip install wheel`
```
tyclintw: Can we add a "requirements.txt" and replace this with "pip install -r requirements.txt"? This will prevent the need to update this document if the dependencies change.

reinforcement_learning/bindings/python/README.md
```diff
@@ -1 +1,14 @@
...
+3. `sudo python3.6 -m pip install setuptools`
+4. Edit reinforcement_learning/bindings/python/setup.py to point to a directory on your machine (Will install dependencies here)
```
tyclintw: We shouldn't be asking people to modify a file to build. Can we modify setup.py to accept a dirs. argument instead. We should also edit the current directory string '/home/ataymano/restore_e2e_1' to remove usernames.

reinforcement_learning/bindings/python/README.md
```diff
@@ -1 +1,14 @@
...
+5. `./reinforcement_learning/bindings/python/build_linux/restore.sh`
+6. `make clean`
+7. `make rl_python`
+8. `sudo pip3.6 install reinforcement_learning/bindings/python/dist/rl_client-0.0.5-cp36-cp36m-linux_x86_64.whl`
```
tyclintw: For consistency can we match all pip commands "sudo python3.6 -m pip"?

reinforcement_learning/bindings/python/README.md
jackgerrits: setup.py is called by the makefile which is called by the global vw makefile. So there isn't an easy way to parametize this. Best I can think of right now is a config file that this is picked up from?

jackgerrits: Went with an environment variable - see the latest update

PR #1586, opened 2018-08-30
[client lib] Separate Python binding and C++ docs into different doxygen projects

jackgerrits:
- CPP and Python are now in their own separate doxygen project and can be navigated between using the header

image


JohnLangford:
Merged, thanks.

PR #1591, opened 2018-08-31
OpenSsl linking problem workaround

ataymano:
Issue was introduced by installing openssl-1.1 on appveyor images which is wrongly picked up by CMake.

Workaround copied from here:
https://github.com/Microsoft/vcpkg/issues/4189#issuecomment-417462822

Fix is not released yet:
https://github.com/appveyor/ci/issues/2600

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,9 @@ shallow_clone: true
 clone_folder: C:\vowpal_wabbit
 # need to install nuget packages before Visual Studio starts to make ANTLR targets available.
 build_script:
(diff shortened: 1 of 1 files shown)


jackgerrits:
Based on the linked appveyor issue it seems as though this is mitigated on their end now right?

ataymano:
Not sure (but we can try to trigger build again). As far as I understand here is the proper link we should track (in terms of deployment from their side):
https://github.com/appveyor/ci/milestone/95

PR #1593, opened 2018-09-07
When parsing dsjson, skip lines not starting with "{"

marco-rossi29:


jackgerrits:
Can you provide an explanation of why we want to do this?

marco-rossi29:
@jackgerrits We need this because with the new Joiner/Trainer refactoring we are saving one line in the azure storage logs with checkpoint information and that is currently breaking VW without this change

JohnLangford:
I'm merging this for now, but I'd like us to work out a long term format for the logs...

PR #1595, opened 2018-09-10
C# Bindings: Native Object lifecycle management and Configuration

lokitoth:
* E2E command-line build and P/Invoke (Windows)
* Stand up lifecycle contract in NativeObject (todo: simplify verbosity)
* Bind Configuration lifecycle and JSON-initialization

### DIFF
--- a/reinforcement_learning/bindings/cs/rl.net.cli/App.config
+++ b/reinforcement_learning/bindings/cs/rl.net.cli/App.config
@@ -1,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+    <startup> 
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.2" />
+    </startup>
+</configuration>
--- a/reinforcement_learning/bindings/cs/rl.net.cli/rl.net.cli.csproj
+++ b/reinforcement_learning/bindings/cs/rl.net.cli/rl.net.cli.csproj
@@ -1,0 +1,58 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <PlatformTarget>x64</PlatformTarget>
+    <Platform >x64</Platform>
+    <ProjectGuid>{D0A78C03-F8E9-447F-99D4-795F6E33D8CE}</ProjectGuid>
+    <OutputType>Exe</OutputType>
+    <RootNamespace>RL.Net.Cli</RootNamespace>
+    <AssemblyName>rl</AssemblyName>
+    <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <OutputPath>$(SolutionDir)$(Platform)\$(Configuration)</OutputPath>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <!-- What are you?! -->
+    <Deterministic>true</Deterministic>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <DefineConstants>DEBUG;TRACE</DefineConstants>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
+    <PlatformTarget>x64</PlatformTarget>
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <DefineConstants>TRACE</DefineConstants>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
+    <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Data.DataSetExtensions" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Data" />
+    <Reference Include="System.Net.Http" />
+    <Reference Include="System.Xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="EntryPoints.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="App.config" />
+    <Copy Include="$(SolutionDir)examples\basic_usage_app\client.json" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\rl.net\rl.net.csproj">
+      <Project>{dfe0ee0d-dd39-4a6e-a716-4486ee86c317}</Project>
+      <Name>RL.Net</Name>
+    </ProjectReference>
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+</Project>
--- a/reinforcement_learning/bindings/cs/rl.net.native/rl.net.global.h
+++ b/reinforcement_learning/bindings/cs/rl.net.native/rl.net.global.h
@@ -1,0 +1,8 @@
+#pragma once
+
+#include "rl.net.native.h"
+
+// Global exports
+extern "C" {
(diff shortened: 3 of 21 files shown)


reinforcement_learning/rl.sln
```diff
@@ -0,0 +1,58 @@
...
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|x64 = Debug|x64
+		Release|x64 = Release|x64
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+        {1E205806-7F80-47DD-A38D-FC08083F3593}.Debug|x64.ActiveCfg = Debug|x64
+		{1E205806-7F80-47DD-A38D-FC08083F3593}.Debug|x64.Build.0 = Debug|x64
+		{1E205806-7F80-47DD-A38D-FC08083F3593}.Release|x64.ActiveCfg = Release|x64
+		{1E205806-7F80-47DD-A38D-FC08083F3593}.Release|x64.Build.0 = Release|x64
+        {C40E3263-B6FF-46D5-92F7-D67E6F9FCEDD}.Debug|x64.ActiveCfg = Debug|x64
```
jackgerrits: nit: mixed spaces and tabs

reinforcement_learning/bindings/cs/rl.net.cli/rl.net.cli.csproj
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
jackgerrits: Is it possible for this to be sdk based now? Given that's where we are moving everything else

reinforcement_learning/bindings/cs/rl.net.native/rl.net.global.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include "rl.net.native.h"
+
+// Global exports
+extern "C" {
+    API void SayHello(void);
```
jackgerrits: Should we be exporting New and Delete?

reinforcement_learning/bindings/cs/rl.net/Configuration.cs
```diff
@@ -0,0 +1,32 @@
...
+        public Configuration() : base(new New<Configuration>(CreateConfig), new Delete<Configuration>(DeleteConfig))
+        {
+        }
+
+        public static bool TryLoadConfigurationFromJson(string json, out Configuration config)
+        {
+            config = new Configuration();
+
+            int result = LoadConfigurationFromJson(json.Length, json, config.handle);
+            return true;
```
jackgerrits: Check result or add a TODO

reinforcement_learning/bindings/cs/rl.net.cli/rl.net.cli.csproj
```diff
@@ -0,0 +1,58 @@
...
+    <OutputType>Exe</OutputType>
+    <RootNamespace>RL.Net.Cli</RootNamespace>
+    <AssemblyName>rl</AssemblyName>
+    <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <OutputPath>$(SolutionDir)$(Platform)\$(Configuration)</OutputPath>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <!-- What are you?! -->
```
yannstad: why this comment?

reinforcement_learning/bindings/cs/rl.net/Configuration.cs
```diff
@@ -0,0 +1,32 @@
...
+    public sealed class Configuration: NativeObject<Configuration>
+    {
+        [DllImport("rl.net.native.dll")]
+        private static extern IntPtr CreateConfig();
+
+        [DllImport("rl.net.native.dll")]
+        private static extern void DeleteConfig(IntPtr config);
+
+        [DllImport("rl.net.native.dll")] // We need to decide on accepted string encodings here. By default, I'm declaring this a unicode string.
+        private static extern int LoadConfigurationFromJson(int jsonLength, [MarshalAs(UnmanagedType.LPWStr)] string json, IntPtr config);
```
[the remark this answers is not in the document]
lokitoth: Remember that P/Invoke is c-style. LP just means that it is char*/wchar_t*/tchar*. Also P/Invoke works at the machine level, so it has no notion of const-correctness.

The real question is what string encoding is desired: We have the choice of ANSI, UTF-16, UTF-8, or "platform-dependent" (ANSI/UTF-16, e.g. tchar*). Have we specified what we support here?

reinforcement_learning/bindings/cs/rl.net/Native/Global.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+using System.Threading;
+using System.Runtime.InteropServices;
+
+namespace Rl.Net.Native {
+    internal partial class NativeMethods
+    {
+        [DllImport("rl.net.native.dll")]
+        public static extern void SayHello();
```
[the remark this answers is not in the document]
lokitoth: Nope; it looks like Globals might be able to go away.

reinforcement_learning/bindings/cs/rl.net/Native/NativeObject.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        private readonly Delete<THandle> operatorDelete;
+
+        protected NativeObject(New<THandle> operatorNew, Delete<THandle> operatorDelete) 
+        : base(operatorNew(), ownsHandle: true)
+        {
+            this.operatorDelete = operatorDelete;
+
+            // TODO: PRECHECKIN: Check nulls in Debug
+            Console.Out.WriteLine($"New object at at {this.handle.ToInt64():x}");
```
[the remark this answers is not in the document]
lokitoth: No, good catch

reinforcement_learning/bindings/cs/rl.net/Native/NativeObject.cs
```diff
@@ -0,0 +1,57 @@
...
+            if (localHandle == null)
+            {
+                return true; // What does the contract say we need to do here, semantically?
+            }
+
+            // Here, we must obey all rules for constrained execution regions.
+            Console.Out.WriteLine($"Deleting object at at {localHandle.ToInt64():x}");
+            operatorDelete(localHandle);
+            
+            return true;
```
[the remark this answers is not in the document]
lokitoth: This is part of the SafeHandle contract - we cannot change the return type. It seems that the primary purpose of returning false is to trigger a Managed Debugging Assistant in the case of failure. Technically, SafeHandle is supposed to handle properly only calling Release once, so we can just return true.

reinforcement_learning/bindings/cs/rl.net.cli/rl.net.cli.csproj
lokitoth: Yes, I just had a bit of a hard time generating an exe out of the SDK-style project, because I forgot I had to explicitly set net462 as my target framework, rather than NetCoreApp.

reinforcement_learning/bindings/cs/rl.net.native/rl.net.global.h
lokitoth: No - as we'd discussed, because we do not carry over typing information between managed and unmanaged, we would only be able to have procedures that take in void*, which will fail to call the proper destructor.

Using templates will not provide much value here, because it is unclear how to ensure the name is not mangled (it becomes impossible to P/Invoke in a compiler-portable way if there is any mangling happening), and we would still have to explicitly request instantiation for each type we project. We could make it all auto-generated on build, but it's not really worth it given that we only have so many types.

reinforcement_learning/bindings/cs/rl.net.cli/rl.net.cli.csproj
lokitoth: I do not get what the <Deterministic> property is controlling. I'll get rid of the comment - I wanted to take a second look later, but forgot.

reinforcement_learning/bindings/cs/rl.net.native/rl.net.native.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <iostream>
+#include <fstream>
+#include "config_utility.h"
+#include "live_model.h"
+
+// Namespace manipulation for brevity
```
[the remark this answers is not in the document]
lokitoth: I intended to remove that after finishing coding. Good catch.

lokitoth:
I'm going to close this and create a new one (since I will want to rebase onto the changes I want to push out to simplify/consolidate-outputs for the Windows build project files.

reinforcement_learning/bindings/cs/rl.net.cli/App.config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+    <startup> 
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.2" />
+    </startup>
+</configuration>
```
yannstad: [nit] general comment applicable to almost all files: missing new line at the end (to make POSIX happy)

PR #1598, opened 2018-09-10
Mock http calls for live model tests without background callback

jackgerrits:


rajan-chari:
Looks good.  Merged.  Thanks.

PR #1599, opened 2018-09-12
Clean windows build and unify output paths

lokitoth:
This simplifies the Windows build

* Remove Azure projects from solution - they are not well-maintained, and need retargeting / rethinking
* Remove Setup projects from solution - they require WiX to be installed, and this does not seem valuable to most people who will be building VW
* Remove VisualLeakDetector projects - this requires VLD extension to be installed, and the projects do not build
* Sets up an environment variable to inject vcpkg when invoking msbuild from the command line
* Remove unused projects/configurations from solution
* Reconcile core projects' output paths, and simplify project files
* vw_dynamic renamed to vw_core - it is no longer a dynamic library (that is libvw), so the name was unclear
* Fix malformed xml in comment warnings
* Fix missing dll error in boost unit test-based project

PR #1601, opened 2018-09-20
RLLib C# Bindings

lokitoth:
Implement the Initial Set of C# bindings to the RL Library

- The strategy here is direct P/Invoke to ease transition to .Net Standard, and enable use of these bindings xPlat.
- Initial Bindings only targeting the E2E use case as demonstrated by basic_usage_app (implemented into rl.net.cli, to be replaced with the RL Simulator C# implementation).

Known Issues:
- Unclear how to expose the ranking enumerator, since it uses size_t (and other unsigned integers, which are not CLS-Compliant), or whether we event want CLS Compliance.


reinforcement_learning/bindings/cs/rl.net.cli/EntryPoints.cs
```diff
@@ -0,0 +1,78 @@
...
+        // TODO: Pull this out to a separate sample.
+        public static void BasicUsageExample(string [] args)
+        {
+            const float outcome = 1.0f;
+            const string eventId = "event_id";
+            const string contextJson = @"({
+                                  ""GUser"":{""id"":""a"",""major"":""eng"",""hobby"":""hiking""},
+                                  ""_multi"":[ { ""TAction"":{""a1"":""f1""} },{""TAction"":{""a2"":""f2""}}]})";
+
+            if (args.Length != 1) {
```
yannstad: [nit] the brace-style is not consistent

reinforcement_learning/bindings/cs/rl.net.cli/EntryPoints.cs
```diff
@@ -0,0 +1,78 @@
...
+                WriteErrorAndExit("Missing path to client configuration json");
```
yannstad: Maybe the error message should specify that the argument is missing? The fact that the path is missing may not be that helpful.

reinforcement_learning/bindings/cs/rl.net/LiveModel.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+        }
+
+        public bool TryInit(ApiStatus apiStatus = null)
+        {
+            int result = InitLiveModel(this.NativeHandle, apiStatus.ToNativeHandleOrNullptr());
+            return result == NativeMethods.SuccessStatus;
+        }
+
+        // TODO: This is a bit of an akward API, as it requires pre-construction of the RankingResponse.
```
jackgerrits: >akward [](start = 37, length = 6)

awkward

reinforcement_learning/bindings/cs/rl.net/LiveModel.cs
```diff
@@ -0,0 +1,53 @@
...
+        public bool TryChooseRank(string actionId, string contextJson, RankingResponse response, ApiStatus apiStatus = null)
```
jackgerrits: >RankingResponse response [](start = 71, length = 24)

Would it make more sense for this to be an out parameter?

reinforcement_learning/bindings/cs/rl.net.cli/EntryPoints.cs
```diff
@@ -0,0 +1,78 @@
...
+                WriteStatusAndExit(apiStatus);
+            }
+
+            RankingResponse rankingResponse = new RankingResponse();
+            if (!liveModel.TryChooseRank(eventId, contextJson, rankingResponse, apiStatus))
+            {
+                WriteStatusAndExit(apiStatus);
+            }
+
+            long actionId;
```
yannstad: Is there any reason why we use `long` instead of `int`?

reinforcement_learning/bindings/cs/rl.net/RankingResponse.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.Threading;
+using System.Runtime.InteropServices;
+
+using Rl.Net.Native;
+
+namespace Rl.Net {
+    [StructLayout(LayoutKind.Sequential)]
+    internal struct ActionProbability
+    {
+        public UIntPtr ActionId; // If we expose this publically, this will not be CLI-compliant
```
jackgerrits: >publically [](start = 54, length = 10)

publicly

reinforcement_learning/bindings/cs/rl.net/RankingResponse.cs
```diff
@@ -0,0 +1,131 @@
...
+            }
+
+            object System.Collections.IEnumerator.Current => this.Current;
+
+            public bool MoveNext()
+            {
+                // The contract of result is to return 1 if true, 0 if false.
+                int result = RankingEnumeratorMoveNext(this.NativeHandle);
+
+                return result == 1;
```
jackgerrits: >1 [](start = 33, length = 1)

NativeMethods success?

reinforcement_learning/bindings/cs/rl.net/rl.net.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <OutputPath>$(SolutionDir)$(Platform)\$(Configuration)</OutputPath>
+    <Platforms>AnyCPU;x64</Platforms>
```
jackgerrits: >AnyCPU [](start = 15, length = 6)

is AnyCPU supported?

reinforcement_learning/bindings/cs/rl.net.cli/rl.net.cli.csproj
```diff
@@ -0,0 +1,57 @@
...
+    <Copy Include="$(SolutionDir)examples\basic_usage_app\client.json" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\rl.net\rl.net.csproj">
+      <Project>{dfe0ee0d-dd39-4a6e-a716-4486ee86c317}</Project>
+      <Name>RL.Net</Name>
+    </ProjectReference>
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+</Project>
```
jackgerrits: Can this be VS2017 style?

reinforcement_learning/bindings/cs/rl.net.native/rl.net.config.cc
```diff
@@ -0,0 +1,19 @@
+#include "rl.net.config.h"
+
+API u::configuration* CreateConfig()
```
yannstad: Is it possible to have a better name instead of `u` here?

reinforcement_learning/bindings/cs/rl.net.native/rl.net.global.cc
```diff
@@ -0,0 +1,3 @@
+#include "rl.net.global.h"
+
+#include <iostream>
```
jackgerrits: Why is iostream added here?

reinforcement_learning/bindings/cs/rl.net.native/rl.net.config.h
```diff
@@ -0,0 +1,17 @@
+#include "rl.net.native.h"
+
+#pragma once
+
+//int load_file(const std::string& file_name, std::string& file_data);
```
yannstad: [nit] do you plan to remove these comments?

reinforcement_learning/bindings/cs/rl.net.native/rl.net.native.h
```diff
@@ -0,0 +1,14 @@
...
+#include "config_utility.h"
+#include "live_model.h"
+
+// Namespace manipulation for brevity
+namespace r = reinforcement_learning;
+namespace u = r::utility;
+namespace cfg = u::config;
+namespace err = r::error_code;
+
+#define API __declspec(dllexport)
```
jackgerrits: nit: newline at end of file

reinforcement_learning/bindings/cs/rl.net.native/rl.net.global.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include "rl.net.native.h"
+
+// Global exports
+extern "C" {
+}
```
jackgerrits: It seems as though global can be removed?

jackgerrits:
Are a large number of these changes from PR #1599 ? The diff will be easier to consume once that is merged

reinforcement_learning/bindings/cs/rl.net.native/rl.net.ranking_response.cc
```diff
@@ -0,0 +1,79 @@
...
+    r::ranking_response::ranking_iterator end;
+
+public:
+    inline ranking_enumerator_adapter(r::ranking_response* response) : current{response->begin()}, end{response->end()}
+    {}
+
+public:
+    inline int move_next()
+    {
+        if ((this->current != this->end) &&
```
yannstad: I am wondering why you need 2 conditions here. Cannot be done with one test only?

reinforcement_learning/bindings/cs/rl.net.native/rl.net.ranking_response.h
```diff
@@ -0,0 +1,31 @@
...
+    API r::ranking_response* CreateRankingResponse();
+    API void DeleteRankingResponse(r::ranking_response* ranking);
+
+    // TODO: We should think about how to avoid extra string copies; ideally, err constants
+    // should be able to be shared between native/managed, but not clear if this is possible
+    // right now.
+    API const char* GetRankingEventId(r::ranking_response* ranking);
+    API const char* GetRankingModelId(r::ranking_response* ranking);
+
+    // TODO: Would it be more clear to call it GetRankingActionCount, or to keep parallel with
```
yannstad: +1 on your comment. `GetRankingActionCount` sounds more clear to me

reinforcement_learning/bindings/cs/rl.net/RankingResponse.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        // TODO: Why does this method call, which seems like a "get" of a value, have an API status?
+        public bool TryGetChosenAction(out long action, ApiStatus status = null)
+        {
+            action = -1;
+            UIntPtr chosenAction;
+            int result = GetRankingChosenAction(this.NativeHandle, out chosenAction, status.ToNativeHandleOrNullptr());
+
+            if (result == NativeMethods.SuccessStatus)
```
yannstad: should it not be != instead of == ?

reinforcement_learning/bindings/cs/rl.net.cli/EntryPoints.cs
lokitoth: This is due to `int` being defined as 32-bit integer in C#; in vw we use size_t, which is technically uint64 (or rather UIntPtr, which varies depending on platform, but we only target x64 with C# right now). I'm trying to avoid exposing uint64 right now, since these types are not CLSCompliant.

reinforcement_learning/bindings/cs/rl.net/RankingResponse.cs
lokitoth: Nope, this is a standard bool value. It might make sense to switch to that contract, but I'd need to create S_FALSE in that case, for full semantic equivalence. Will see if P/Invoke does the right thing if I annotate it bool, though.

reinforcement_learning/bindings/cs/rl.net/rl.net.csproj
lokitoth: No, but nuget restore did not seem to do the right thing here unless I kept AnyCPU. Will try deleting again.

reinforcement_learning/bindings/cs/rl.net.native/rl.net.config.cc
lokitoth: Ah, whoops, I forgot to yank the shorthand

reinforcement_learning/bindings/cs/rl.net.native/rl.net.global.cc
lokitoth: This used to have SayHello as a first-pass "binding work" test. Can probably be removed outright at this point.

reinforcement_learning/bindings/cs/rl.net.native/rl.net.config.h
lokitoth: Yes

reinforcement_learning/bindings/cs/rl.net.native/rl.net.ranking_response.cc
lokitoth: We only overload the "not equals" operator for this ranking iterator. Given that we need to decide how we want to deal with unsigned integer types in the C# bindings, I ended up not pushing this part through yet.

reinforcement_learning/bindings/cs/rl.net/RankingResponse.cs
```diff
@@ -0,0 +1,131 @@
...
+
+            public bool MoveNext()
+            {
+                // The contract of result is to return 1 if true, 0 if false.
+                int result = RankingEnumeratorMoveNext(this.NativeHandle);
+
+                return result == 1;
+            }
+
+            public void Reset()
```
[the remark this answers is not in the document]
lokitoth: `Reset` is part of the `IEnumerable` interface. I cannot avoid having the method, even though it is not a possible operation, but the usual way of specifying that Reset is not supported is to throw `NotSupportedException`. `foreach` does not use Reset - it's a one-shot pass through the collection, so I have never seen this construction having an issue.

reinforcement_learning/bindings/cs/rl.net.cli/EntryPoints.cs
yannstad: got it

reinforcement_learning/bindings/cs/rl.net/RankingResponse.cs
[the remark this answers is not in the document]
yannstad: Ok, I guess it would be an issue if someone explicitly call the `Reset` method, what is not recommended. (Just for sanity, `Reset` is part of `IEnumerator`, not `IEnumerable`)

reinforcement_learning/bindings/cs/rl.net.native/rl.net.live_model.h
```diff
@@ -0,0 +1,14 @@
...
+
+#include "rl.net.native.h"
+
+// Global exports
+extern "C" {
+    // NOTE: THIS IS NOT POLYMORPHISM SAFE!
+    API reinforcement_learning::live_model* CreateLiveModel(reinforcement_learning::utility::configuration* config);
+    API void DeleteLiveModel(reinforcement_learning::live_model* livemodel);
+
+    API int InitLiveModel(reinforcement_learning::live_model* livemodel, reinforcement_learning::api_status* status = nullptr);
```
tyclintw: >InitLiveModel [](start = 12, length = 13)

nit: LiveModelInit would be more inline with the naming conventions.

reinforcement_learning/bindings/cs/rl.net.cli/rl.net.cli.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net462</TargetFramework>
```
jackgerrits: Are we able to make this .net standard?

reinforcement_learning/bindings/cs/rl.net.native/rl.net.config.cc
```diff
@@ -0,0 +1,19 @@
...
+
+API void DeleteConfig(reinforcement_learning::utility::configuration* config)
+{
+    delete config;
+}
+
+API int LoadConfigurationFromJson(const int length, const char* json, reinforcement_learning::utility::configuration* config, reinforcement_learning::api_status* status)
+{
+    // This is a deep copy, so it is safe to push a pinned-managed string here.
+    const std::string json_str (json, length);
```
jackgerrits: nit: remove space after variable name

reinforcement_learning/bindings/cs/rl.net.cli/rl.net.cli.csproj
lokitoth: No, because .Net Standard does not specify executable targets. Our only choices are NetCoreApp and Net462. If we swap it to NetCoreApp, it makes it much harder to run on the command, as it does not generate an executable, unless we use .Net Native to AOT-compile it, and I'd rather not do that.

reinforcement_learning/bindings/cs/rl.net.native/rl.net.config.cc
lokitoth: Fixed in latest push

JohnLangford:
Ugh, there are conflicts here.   Fix?

lokitoth:
There were not earlier. Did another PR get merged in?

PR #1605, opened 2018-09-28
Change VW factory to construct using model data and to not use seed_vw_model

jackgerrits:


reinforcement_learning/include/model_mgmt.h
```diff
@@ -26,20 +26,21 @@ namespace reinforcement_learning { namespace model_management {
...
             _refresh_count(other._refresh_count) {}
 
         model_data& operator=(model_data&& other) noexcept {
-          if (this == &other)
-            return *this;
-          _data = other._data;
-          _data_sz = other._data_sz;
-          _refresh_count = other._refresh_count;
+          if (this != &other) {
+            _data = other._data;
```
rajan-chari: Should we not swap content here to be correct?

jackgerrits: We don't want to set the state of other to what the current object holds. Unless we explicitly set ourselves to null/default values and then do a swap

rajan-chari: Will this cause a memory leak if a temporary is assigned to a named model_data?


---
In reply to: [222718213](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1605#discussion_r222718213) [](ancestors = 222718213)

jackgerrits: This is a good point. We need to free any existing memory before we accept from the other object. We also need to be careful to set the other object to null so that when it is destroyed it doesnt clear the memory. I'll fix this up

rajan-chari: doing the swap should take care of that for you.

---
In reply to: [222725336](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1605#discussion_r222725336) [](ancestors = 222725336)

jackgerrits: Not if the current object already has its members populated though right?

reinforcement_learning/rlclientlib/vw_model/safe_vw.h
```diff
@@ -28,10 +29,11 @@ namespace reinforcement_learning {
   };
 
   class safe_vw_factory {
-    std::shared_ptr<safe_vw> _master;
+    model_management::model_data _master_data;
 
   public:
-    safe_vw_factory(const std::shared_ptr<safe_vw>& master);
+    // model_data is copied and stored in the factory object.
+    safe_vw_factory(const model_management::model_data& master_data);
```
rajan-chari: Not super important.  An r-value override could produce a copy free model data.

PR #1609, opened 2018-10-01
Fixed _labelIndex out of bound error message

marco-rossi29:
_label_index -> _labelIndex
added clarification that _labelIndex must be smaller than number of examples


jackgerrits:
:shipit:

PR #1613, opened 2018-10-03
Update Readme with details on installing boost dependencies

whiletruelearn:
- Installing `vowpalwabbit` directly with pip fails currently if boost dependencies are not installed. Updating the Readme with instructions for installing dependencies would be helpful for future users.


JohnLangford:
Merged, thanks :-)

PR #1615, opened 2018-10-04
Fix cpprestsdk at version 2.10.1 because of build break

jackgerrits:
Fixes cpprestsdk version to mitigate break checked into Microsoft/cpprestsdk#883


rajan-chari:
ok.  will merge.

jackgerrits:
This was resolved in #1620

PR #1617, opened 2018-10-05
update bindings

dwaijam:


reinforcement_learning/bindings/cs/rl.net/ILiveModel.cs
```diff
@@ -0,0 +1,11 @@
+namespace Rl.Net
+{
+    public interface ILiveModel
+    {
+        bool TryChooseRank(string actionId, string contextJson, out RankingResponse response, ApiStatus apiStatus = null);
+        bool TryChooseRank(string actionId, string contextJson, RankingResponse response, ApiStatus apiStatus = null);
+        bool TryInit(ApiStatus apiStatus = null);
+        bool TryReportOutcome(string actionId, object outcome, ApiStatus apiStatus = null);
```
tyclintw: Why not create two methods?

```csharp
        bool TryReportOutcome(string actionId, float outcome, ApiStatus apiStatus = null);
        bool TryReportOutcome(string actionId, string outcome, ApiStatus apiStatus = null);
```

reinforcement_learning/bindings/cs/rl.net/LiveModel.cs
```diff
@@ -49,10 +53,24 @@ public bool TryChooseRank(string actionId, string contextJson, RankingResponse r
             return result == NativeMethods.SuccessStatus;
         }
 
-        public bool TryReportOutcome(string actionId, float outcome, ApiStatus apiStatus = null)
+        public bool TryReportOutcome(string actionId, object outcome, ApiStatus apiStatus = null)
         {
-            int result = LiveModelReportOutcome(this.NativeHandle, actionId, outcome, apiStatus.ToNativeHandleOrNullptr());
+            string objStr = JsonConvert.SerializeObject(outcome);
+            int result;
```
tyclintw: Why would we serialize the outcome just to reparse it? Wouldn't it make more sense to simply check the object for type float?

reinforcement_learning/bindings/cs/rl.net/LiveModel.cs
```diff
@@ -49,10 +53,24 @@ public bool TryChooseRank(string actionId, string contextJson, RankingResponse r
...
+            }
+            else
+            {
+                result = LiveModelReportOutcome(this.NativeHandle, actionId, objStr, apiStatus.ToNativeHandleOrNullptr());
+            }
             return result == NativeMethods.SuccessStatus;
         }
+
+        public new void Dispose()
+        {
```
tyclintw: What is the point of overloading this method if you don't have any new functionality besides "base.Dispose()"?

reinforcement_learning/bindings/cs/rl.net/ILiveModel.cs
dwaijam: mwt client had object as the outcome type because sometimes int was also getting passed from the decision service api.
I wanted to have Jacob's opinion on what he had in mind to support that. I was conetemplating float and string myself before I came across a failed test which made me think otherwise

reinforcement_learning/bindings/cs/rl.net/LiveModel.cs
dwaijam: float casting throws exception if int value is passed. Maybe I can do Convert.ToSingle which should handle both and then if it throws then do Convert.ToString

reinforcement_learning/bindings/cs/rl.net/LiveModel.cs
dwaijam: the interface ILiveModel needs to have Dispose as it will be called from DecisionService. this Dispose goes and calls SafeHandle's Dispose method

lokitoth:
The changes from this PR are pulled in by #1618

lokitoth:
Closing this, since #1618 has been merged.

PR #1618, opened 2018-10-05
Port rl_sim_cpp to C# in RL.Net.Cli

lokitoth:
Port the simulator to C# to enable validation against C++.


reinforcement_learning/bindings/cs/rl.net/LiveModel.cs
```diff
@@ -22,13 +22,35 @@ private static New<LiveModel> BindConstructorArguments(Configuration config)
         private static extern int LiveModelInit(IntPtr liveModel, IntPtr apiStatus);
 
         [DllImport("rl.net.native.dll")]
-        private static extern int LiveModelChooseRank(IntPtr liveModel,  [MarshalAs(NativeMethods.StringMarshalling)] string eventId,  [MarshalAs(NativeMethods.StringMarshalling)] string contextJson, IntPtr rankingResponse, IntPtr apiStatus);
+        private static extern int LiveModelChooseRank(IntPtr liveModel, [MarshalAs(NativeMethods.StringMarshalling)] string eventId, [MarshalAs(NativeMethods.StringMarshalling)] string contextJson, IntPtr rankingResponse, IntPtr apiStatus);
 
         [DllImport("rl.net.native.dll")]
-        private static extern int LiveModelReportOutcome(IntPtr liveModel,  [MarshalAs(NativeMethods.StringMarshalling)] string eventId,  float outcome, IntPtr apiStatus);
+        private static extern int LiveModelReportOutcomeF(IntPtr liveModel, [MarshalAs(NativeMethods.StringMarshalling)] string eventId, float outcome, IntPtr apiStatus);
```
dwaijam: Can't we have the same name with different signatures ?

dwaijam:
>     public sealed class LiveModel: NativeObject<LiveModel>

this needs to implement an Interface to be mockable for unit tests

---
Refers to: reinforcement_learning/bindings/cs/rl.net/LiveModel.cs:8 in 7638351. [](commit_id = 7638351076519fa40ef39efebb7c0dabef49c4dc, deletion_comment = False)

reinforcement_learning/bindings/cs/rl.net/LiveModel.cs
```diff
@@ -51,8 +73,39 @@ public bool TryChooseRank(string actionId, string contextJson, RankingResponse r
...
+        public bool TryReportOutcome(string actionId, string outcomeJson, ApiStatus apiStatus = null)
+        {
+            int result = LiveModelReportOutcomeJson(this.NativeHandle, actionId, outcomeJson, apiStatus.ToNativeHandleOrNullptr());
+            return result == NativeMethods.SuccessStatus;
+        }
+
+        private event EventHandler<ApiStatus> BackgroundErrorInternal;
+
+        // TODO: This class need a pass to ensure thread-safety (or explicit declaration of non-thread-safe)
+        public event EventHandler<ApiStatus> BackgroundError
```
jackgerrits: It looks like there can only ever be one callback added? The convention is to be able to add any number, so this may be unexpected behavior for a consumer.

lokitoth: The aggregation logic is invisible, because it is auto-implemented in the `BackgroundErrorInternal` event. If you note, we pass it the delegate (`value`) from `BackgroundError`. Then in the actual managed-side callback, `WrapStatusAndRaiseBackgroundError`, raises the -`Internal` event, rather than the outer one consumers of the API use to register.

reinforcement_learning/bindings/cs/rl.net/LiveModel.cs
lokitoth: We cannot, because we have to have this function be treated as a c-style function (to avoid name mangling, so P/Invoke can find it), and c does not allow function overloading, so if we try to compile the rl.net.native project with two exports of LiveModelReportOutcome we get compile errors.

PR #1620, opened 2018-10-05
Use a prebuilt docker image for travis build

jackgerrits:
Docker image contains all dependencies pre-installed and reduces build time from ~40 minutes to ~20 minutes


Dockerfile
```diff
@@ -0,0 +1,57 @@
+FROM ubuntu:16.04 AS build
```
rajan-chari: Should we switch to ubuntu 14.x?

jackgerrits: We should, working on an update now.

rajan-chari:
Looks good.  Thanks

JohnLangford:
Merged, thanks :-)  The smaller travis build times alone make this very useful :-)

PR #1631, opened 2018-10-12
Ataymano/deferred action

ataymano:
Deferred action support in rlclient


reinforcement_learning/rlclientlib/live_model.cc
```diff
@@ -43,34 +44,34 @@ namespace reinforcement_learning
...
   }
 
   int live_model::choose_rank(const char* context_json, ranking_response& response, api_status* status)
   {
     INIT_CHECK();
-    return _pimpl->choose_rank(context_json, response, status);
+    return choose_rank(context_json, action_flags::DEFAULT, response, status);
   }
 
   //not implemented yet
```
lokitoth: [Minor] Can be removed

reinforcement_learning/rlclientlib/live_model.cc
```diff
@@ -43,34 +44,34 @@ namespace reinforcement_learning
...
   int live_model::choose_rank(const char* event_id, const char* context_json, unsigned int flags, ranking_response& response,
```
lokitoth: Why not use the enum type actionFlags here, rather than unsigned int?

ataymano: But this is public interface - having enum exposed here can cause backward compatibility issues here, no?

ataymano: Just realized that it is actually not relevant now since this is static lib. Otherwise it introduces additional requirement for consumer and rl_lib to be built with the same version of enum.

lokitoth: Yep, that makes sense. Was asking because it was static lib.

reinforcement_learning/rlclientlib/live_model.cc
ataymano: you mean both methods with legacy signature (without flags)?

lokitoth: The comment about "not implemented yet" is superfluous in both.

PR #1634, opened 2018-10-12
Further improve build cache by saving packages

jackgerrits:


jackgerrits:
It seems the packages directory is greater than 1gb compressed and so cant be cached. I think this is still worth checking in and trying to reduce that size a bit.

JohnLangford:
Why is the packages directory huge?

jackgerrits:
I would say it is mostly because of the number of Boost packages we bring in https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/packages.config. The packages folder on my machine is ~3.5gb. I think we only actually require program_options, system, unit_test_framework, foreach and align. I'll try removing all of the extra ones and seeing if we can still build.

PR #1639, opened 2018-10-15
Ataymano/deferred action parse

ataymano:


cs/common/VowpalWabbitDecisionServiceInteractionHeader.cs
```diff
@@ -32,5 +32,10 @@ public sealed class VowpalWabbitDecisionServiceInteractionHeader
         /// Probability of drop extracted from Decision Service Interaction JSON data.
         /// </summary>
         public float ProbabilityOfDrop { get; set; }
+
+        /// <summary>
+        /// Flag if this example is used for learning.
+        /// </summary>
+        public bool Deferred { get; set; }
```
lokitoth: [Minor] There's a bit of a mismatch between the language used for the comment "if this is example is used for learning" implying true => learning will happen, and the name, "deferred", indicating true => learning might not happen.

vowpalwabbit/parse_example_json.h
```diff
@@ -931,6 +951,11 @@ class DecisionServiceState : public BaseState<audit>
 			else if (length == 11 && !_stricmp(str, "_labelIndex"))
 				return &ctx.label_index_state;
 		}
+		else if (length == 9 && !strcmp(str, "_deferred"))
```
rajan-chari: >strcmp [](start = 27, length = 6)

strncmp is more secure

JohnLangford:
This looks good to me.  Merge when the tests pass.

PR #1643, opened 2018-10-17
Fix cb explore adf segfault

JohnLangford:


vowpalwabbit/cb_explore_adf.cc
```diff
@@ -219,6 +219,8 @@ void get_cost_ranges(std::vector<float> &min_costs,
 
 size_t fill_tied(cb_explore_adf& data, v_array<action_score>& preds)
 {
+  if (preds.size() == 0 )
```
yannstad: It fixes the segfault issue, but it looks to me that the counting of examples would not be correct anymore?

JohnLangford:
This function returns the number of tied probability actions.   It
previously assumed that the number of actions was at least 1.

It doesn't affect counting of examples.

-John

PR #1644, opened 2018-10-17
rl_sim.py: Improvements

marco-rossi29:
- Fixed namespaces in context
- Report outcome only if non-zero
- Added stats
- Added number of rounds as input
- Slimmer argparse invokation


reinforcement_learning/examples/python/rl_sim.py
```diff
@@ -51,40 +50,45 @@ def __init__(self, args):
 
     def loop(self):
         round = 0
-        while (True):
+        stats = {person._id : [[0,0] for _ in self._actions] for person in self._people}
```
jackgerrits: Thanks for moving out the initialization 👍

reinforcement_learning/examples/python/rl_sim.py
```diff
@@ -51,40 +50,45 @@ def __init__(self, args):
...
+        ctr = [0,0]
+        while round != self._options.num_rounds:
```
jackgerrits: Even though I am sure this would be fine because the variable is changed as the first thing in the loop, this would feel a tiny safer if it was less than:
```suggestion
        while round <= self._options.num_rounds:
```

Feel free to wontfix this!

PR #1645, opened 2018-10-18
Harden rllib's test_cpp in validation mode

lokitoth:
* Add throttling to verify-mode by allowing sleeping between loop run-throughs.
* Add background error callback to avoid getting into permanent error mode after the first background error


reinforcement_learning/examples/test_cpp/test_loop.cc
```diff
@@ -99,6 +100,10 @@ void test_loop::validity_loop(size_t thread_id)
     }
 
     test_inputs.log(thread_id, i, response, loggers[thread_id]);
+    
+    if (sleep_interval) {
```
jackgerrits:
```suggestion
    if (sleep_interval > 0) {
```

I'm not a fan of this c-style bool check. I'd rather it be explicit that we are comparing an integer

PR #1646, opened 2018-10-18
Make VW setup projects buildable from the command-line

lokitoth:
It is still a little finicky, and the solution does not always like to load in Visual Studio. There is more work to be done here, but this is a good place to pull and release a current VW installer package.

* Fixes the redistributable package version to properly pick up vc140, to match what we target at build-time
* Fixes a bug with the setup where the executable will fail if the vcrt package is already installed (this did not impact the MSI)


JohnLangford:
Merged, thanks :-)

PR #1648, opened 2018-10-19
 nothing

maverick100:


jackgerrits:
I'm sorry but this is not a helpful PR. If you would like to contribute to the project please describe the changes in the title and description, ensure the build passes and make sure it isn't change for the sake of change. If you're looking for a simple change then I've noticed small issues like spelling mistakes here and there.

lokitoth:
Closing this PR since there does not appear to be action on it. If necessary, please resubmit with title/description of the change.

JohnLangford:
@Scott-Graham-Bose is there a reason to prefer print to an exception?

Scott-Graham-Bose:
The idea is to abandon installation if there is a problem with finding the correct dependencies or system. If we want to support additional systems we should identify them and include them directly here instead of allowing installation to continue for unknown configurations.

JohnLangford:
Cool.

PR #1649, opened 2018-10-20
Add an option to block instead of dropping events when queue is full

yannstad:
Currently in the RL client lib, events are statistically dropped if the background queue is full.

This PR introduces an additional optional behavior:
* events are never dropped
* adding an event to the queue blocks, if the queue if full, until an empty space is freed


reinforcement_learning/rlclientlib/logger/async_batcher.h
```diff
@@ -63,8 +65,14 @@ namespace reinforcement_learning {
 
   template<typename TEvent>
   int async_batcher<TEvent>::append(TEvent&& evt, api_status* status) {
+    if (_queue_dropping_disable && _queue.size() > _queue_max_size) { // block if queue is full
+      std::unique_lock<std::mutex> lk(_m);
+      _cv.wait(lk, [this] { return _queue.size() < _queue_max_size; });
+    }
     _queue.push(std::move(evt));
-    prune_if_needed();
+    if (!_queue_dropping_disable && _queue.size() > _queue_max_size) { // drop events if queue is full
```
ataymano: why do we need to split this logic into 2 blocks?
if (queue is full)
  if (drop) drop else wait;
looks cleaner.

reinforcement_learning/rlclientlib/logger/async_batcher.h
```diff
@@ -37,8 +36,8 @@ namespace reinforcement_learning {
                   error_callback_fn* perror_cb = nullptr,
                   size_t send_high_water_mark = (1024 * 1024 * 4),
                   size_t batch_timeout_ms = 1000,
-                  size_t queue_max_size = (8 * 1024));
-
+                  size_t queue_max_size = (8 * 1024),
+                  bool queue_dropping_disable = false);
```
ataymano: double negation "disable + false" sounds a bit complicated. Maybe it is better either to rename _queue_dropping_disable to something like _queue_blocking_available or switch to enum with explicit mode names (queue_mode::BLOCKING / queue_mode::DROP)

reinforcement_learning/rlclientlib/logger/async_batcher.h
yannstad: Because:
* the prunning is done after the enqueuing, whereas the waiting must happen before. 
* I tried to avoid modifying the prunning logic.

If you think that the prunning logic can be safely moved from "after the enqueuing" to "before the enqueing", then yes, the logic does not need to be splitted.

ataymano: I do not see big difference for moving any of blocks either before or after push, since effectively it is just shifting _queue_max_size by 1 (which is not a problem for me).

reinforcement_learning/rlclientlib/logger/async_batcher.h
yannstad: I like your idea using enum. Looks definitely better indeed.

reinforcement_learning/rlclientlib/logger/async_batcher.h
yannstad: +1

reinforcement_learning/rlclientlib/logger/async_batcher.h
```diff
@@ -63,8 +65,16 @@ namespace reinforcement_learning {
...
+    if (_queue.size() > _queue_max_size) {
+      if (_queue_dropping_disable) { // block if queue is full
+        std::unique_lock<std::mutex> lk(_m);
+        _cv.wait(lk, [this] { return _queue.size() < _queue_max_size; });
+      }
+      else { // drop events if queue is full
+        _queue.prune(_pass_prob);
+      }
+    }
     _queue.push(std::move(evt));
```
jackgerrits: Doesn't this change introduce the side effect of the appended event can never be the one that is pruned?

jackgerrits:
Why do we wan't to explicitly block if we have reached max size? Could we simply just uncap the queue limit? Or do we want to explicitly protect against an unbounded queue?

reinforcement_learning/rlclientlib/logger/async_batcher.h
yannstad: Yes. It looks to me more like a choice than a statistical issue.

yannstad:
Do you see any reason why we would prefer to uncap the queue limit?

jackgerrits:
It would really just be for simplicity. Much simpler to unbound than block and wait. This is really just a design decision so wanted to put it out there. I am happy with how it is.

yannstad:
I agree. But adding the "block" behavior is finally pretty simple to implement. So in my opinion it is worth doing it.

reinforcement_learning/rlclientlib/async_batcher.cc
```diff
@@ -0,0 +1,11 @@
+#include "logger/async_batcher.h"
+
+namespace reinforcement_learning {
+    queue_mode_enum to_queue_mode_enum(const char* queue_mode_str) {
```
ataymano: probably just 'queue_mode' is better:
1) it is consistent with other types / enums naming (see include/action_flags.h)
2) it is confusing to have queue_mode_enum as typename and queue_mode_str as variable name in a single scope.

PR #1650, opened 2018-10-22
Windows test fixes, fix error reporting in logger send

jackgerrits:
- Fix corruption of value returned buffer
- Fix error reporting of asynchronous send


reinforcement_learning/rlclientlib/live_model_impl.cc
```diff
@@ -177,9 +177,9 @@ namespace reinforcement_learning {
     // Generate egreedy pdf
     size_t action_count = 0;
     RETURN_IF_FAIL(utility::get_action_count(action_count, context, _trace_logger.get(), status));
-    
+
     vector<float> pdf(action_count);
-    // Generate a pdf with epsilon distributed between all action.  The top action 
+    // Generate a pdf with epsilon distributed between all action.  The top action
```
yannstad: [ultra minor] would be easier to read if ` The top action` is moved to the next line

yannstad:
The title of this PR is confusing. You do more than fixing test here. You fix the actual implementation, aren't you?

jackgerrits:
I do, I'll update the title

reinforcement_learning/rlclientlib/model_mgmt/model_mgmt.cc
```diff
@@ -40,8 +40,10 @@ namespace reinforcement_learning {
     }
 
     void model_data::free() {
-      delete[] _data;
-      _data = nullptr;
+      if (_data != nullptr) {
```
yannstad: was this causing an issue in windows test?

jackgerrits: Potentially. Jacob made this change. The behavior I was seeing that the returned `model_data` object from `get_model_data_from_raw` contained garbage memory despite being copied to right before. This feels like a compiler bug to me because not only should that object have been valid, it should have also followed copy elision and so not have to be copied and deleted. Ah well, this is a simple mitigation.

reinforcement_learning/test_tools/sender_test/test_loop.cc
```diff
@@ -3,6 +3,7 @@
 #include "constants.h"
 #include "config_utility.h"
 #include "factory_resolver.h"
+#include "error_callback_fn.h"
```
yannstad: [minor] includes should be in lexicographical order

reinforcement_learning/rlclientlib/model_mgmt/model_mgmt.cc
lokitoth: I was seeing a free on nullptr here. It might no longer be an issue with the real fix.

jackgerrits: It doesn't hurt to do the check. We don't want to double free

reinforcement_learning/rlclientlib/factory_resolver.cc
```diff
@@ -89,26 +89,28 @@ namespace reinforcement_learning {
...
-  int interaction_sender_create(i_sender** retval, const u::configuration& cfg, i_trace* trace_logger, api_status* status) {
+  int interaction_sender_create(i_sender** retval, const u::configuration& cfg, error_callback_fn* _error_cb, i_trace* trace_logger, api_status* status) {
     *retval = new eventhub_client(
       cfg.get(name::INTERACTION_EH_HOST, "localhost:8080"),
       cfg.get(name::INTERACTION_EH_KEY_NAME, ""),
       cfg.get(name::INTERACTION_EH_KEY, ""),
       cfg.get(name::INTERACTION_EH_NAME, "interaction"),
       cfg.get_int(name::INTERACTION_EH_TASKS_LIMIT, 16),
       trace_logger,
+      _error_cb,
```
lokitoth: [Minor][Nit] _error_cb => error_cb (+others)

Not important enough to block this PR. Please fix in followup.

lokitoth:
Thanks, merged!

PR #1653, opened 2018-10-24
Refactor eventhub message format to use flatbuffers

homezcx:
This PR is refactoring eventhub messages format to use flatbuffers.

1. It refactors async_batcher to send flatbuffer messages
2. It refactors eventhub_client to send data in binary format
3. It refactors data_buffer to use unsigned char, instead of assuming text format (char)
4. choose rank, report outcome API response format and log format are not touched.


rajan-chari:
CI failing for windows and linux

jackgerrits:
Please do not check in the flatbuffer source code into the repo.

If you require the code to be in the source tree then please add a git submodule to the ext_libs directory. Once we move to cmake (very soon) then consuming this dependency will be easier.

Alternatively can this be installed as a dependency on the system? I see there is a vcpkg port available for this.

jackgerrits:
I would really like to see this implemented as perhaps a new sender implementation that uses a preamble to multiplex. I would rather that we didn't simply replace the existing text based sending functionality if the intention is to keep this around as a debug transport option

reinforcement_learning/examples/override_interface/override_interface.cc
```diff
@@ -32,7 +32,7 @@ class ostream_sender : public r::i_sender {
   }
 
 protected:
-  virtual int v_send(std::string&& data, r::api_status* status) override {
+  virtual int v_send(const std::string& data, r::api_status* status) override {
```
rajan-chari: What's the reason for this change?

reinforcement_learning/unit_test/event_test.cc
```diff
@@ -71,10 +75,14 @@ BOOST_AUTO_TEST_CASE(serialize_empty_ranking)
   ranking.set_model_id("model_id");
 
   utility::data_buffer oss;
+<<<<<<< HEAD
```
rajan-chari: This won't compile

homezcx: this file is not included in the build and should be deleted. As there it will be replaced by flatbuffer and flatbuffer has its own unit tests for this case.

reinforcement_learning/examples/override_interface/override_interface.cc
homezcx: it was the master code before. I think it's merge override. I will recover it.

homezcx:
Hi Jack, to implement flatbuffer using preamble header is not in the scope of this PR. We can discuss the preamble header implementation offline.

---
In reply to: [432779804](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#issuecomment-432779804) [](ancestors = 432779804)

reinforcement_learning/include/sender.h
```diff
@@ -9,14 +10,20 @@ namespace reinforcement_learning {
   public:
     virtual int init(api_status* status) = 0;
 
-    int send(std::string&& data, api_status* status = nullptr)
+    int send_byte(std::vector<unsigned char>& data, api_status* status = nullptr)
```
ataymano: why we cannot have move semantics here? if moving is impossible, it probably should be at least const &.

reinforcement_learning/rlclientlib/logger/async_batcher.h
```diff
@@ -55,9 +56,11 @@ namespace reinforcement_learning {
     std::unique_ptr<i_sender> _sender;
 
     event_queue<TEvent> _queue;       // A queue to accumulate batch of events.
+    utility::event_batcher<TEvent> _event_batcher;
     utility::data_buffer _buffer;           // Re-used buffer to prevent re-allocation during sends.
     size_t _send_high_water_mark;
     size_t _queue_max_size;
+    size_t _batch_size = 400;        // initial batch size used by flatterbuffers
```
ataymano: Looks more like event_batcher member.

reinforcement_learning/rlclientlib/logger/event_batcher.h
```diff
@@ -0,0 +1,87 @@
...
+    };
+
+    template<typename TEvent>
+    void event_batcher<TEvent>::batch_serialize(data_buffer& oss, size_t& remaining, event_queue<TEvent>& queue, size_t _send_high_water_mark, size_t batch_size) {
+
+      if (batch_size <= 0) {
+        throw("event_batcher::batch_serialize batch_size is 0 - message is too large.");
+      }
+
+      if (std::is_same<TEvent, ranking_event>::value) {
```
ataymano: It doesn't seem like a proper way to customize behavior based on type. It has to be either:
- Hidden behind some common interface within ranking_event/interaction_event implementations
or
- using template specialization: https://en.cppreference.com/w/cpp/language/template_specialization

reinforcement_learning/rlclientlib/logger/eventhub_client.cc
```diff
@@ -18,13 +18,13 @@ namespace reinforcement_learning {
   eventhub_client::http_request_task::http_request_task(web::http::client::http_client& client,
     const std::string& host,
     const std::string& auth,
-    std::string&& post_data)
+    const std::vector<unsigned char>& post_data)
```
ataymano: Overall comment to all these changes. Looks like review will look much simpler if it can be splitted in few stages, where the first step is introducing new abstraction (message/buffer/...) on top of current std::string (even with just typedef). It is quite broad change, but with 0 breaking changes and very easier to validate. After that replacing it to vector<char> (or anything else) turns into quite local changes.

reinforcement_learning/rlclientlib/logger/event_batcher.h
ataymano: Also, these two branches look like exact copy paste of each other (except some type names, that can be 100% unified using templates)

reinforcement_learning/rlclientlib/logger/event_batcher.h
```diff
@@ -0,0 +1,87 @@
...
+  namespace utility {
+    template<typename TEvent>
+    class event_batcher {
+    public:
+      event_batcher() {};
+      void batch_serialize(data_buffer& oss, size_t& remaining, event_queue<TEvent>& queue, size_t _send_high_water_mark, size_t batch_size);
+    };
+
+    template<typename TEvent>
+    void event_batcher<TEvent>::batch_serialize(data_buffer& oss, size_t& remaining, event_queue<TEvent>& queue, size_t _send_high_water_mark, size_t batch_size) {
```
ataymano: it is not clear how responsibilities of event_batcher and async_batcher are related to each other? With current implementation event_batcher cannot be used out of existing scenario because it is tightly coupled with async_batcher implementation details (event_queue / send_high_water_mark/etc). In this case we probably do not need separate class and just additional private method inside async_batcher is ok. 
Other option is to make them orthogonal to each other - but in this case only event_batcher should be owner of all batch-related parameters (send_high_watermark / batch_size / ...)

reinforcement_learning/include/sender.h
homezcx: I tried but moving using && by vector is not supported. Let me know if there is a better way. I can change it to const &.

---
In reply to: [228194531](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228194531) [](ancestors = 228194531)

reinforcement_learning/include/sender.h
```diff
@@ -9,14 +10,20 @@ namespace reinforcement_learning {
...
+    {
+      return v_send(data, status);
+    }
+
+    int send_string(std::string &&data, api_status* status = nullptr)
```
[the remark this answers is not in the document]
homezcx: It was for compatible purpose. I will verify the usage to see if we can remove it. Will replay inline later.

---
In reply to: [228195623](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228195623) [](ancestors = 228195623)

reinforcement_learning/include/sender.h
```diff
@@ -9,14 +10,20 @@ namespace reinforcement_learning {
...
     {
       return v_send(std::move(data), status);
     }
 
     virtual ~i_sender() = default;
-  
+
   protected:
-    virtual int v_send(std::string&& data, api_status* status) = 0;
+    virtual int v_send(std::vector<unsigned char>& data, api_status* status) { return 0; };
```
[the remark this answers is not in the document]
homezcx: This should be caused by rebasing. There were tons of conflicts when I rebased. I will change it.

---
In reply to: [228195878](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228195878) [](ancestors = 228195878)

reinforcement_learning/rlclientlib/logger/eventhub_client.cc
homezcx: True. The reason is that no one would agree to the abstraction change unless people seeing flatbuffer changes need. It's kinda chicken and egg problem.

---
In reply to: [228201579](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228201579) [](ancestors = 228201579)

reinforcement_learning/rlclientlib/logger/event_batcher.h
homezcx: I was not able to find a good so we can simplify the code. But I will check your specialization here or we can discuss later.

---
In reply to: [228199517](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228199517) [](ancestors = 228199517)

homezcx: Same as above.

---
In reply to: [228204171](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228204171) [](ancestors = 228204171)

reinforcement_learning/rlclientlib/logger/event_batcher.h
homezcx: I agree and I think I would move them into async_batcher. The reason why we had that was we might want to support both ds-json string and flatbuffer at the same time. But this was not the case anymore. 

---
In reply to: [228219231](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228219231) [](ancestors = 228219231)

reinforcement_learning/rlclientlib/logger/eventhub_client.cc
ataymano: It looks like mandatory change anyways. You have already done heavy work to discriminate all std::string between sender message and other ones, but the next person who will need to introduce next change on this level have to do the same now.

reinforcement_learning/rlclientlib/logger/event_batcher.h
ataymano: It looks like you do not event need specialization here - just single template method (with probably some type traits tricks) should work

reinforcement_learning/rlclientlib/logger/eventhub_client.cc
ataymano: So having this abstraction is 100% good thing even if all other changes will not be accepted for some reason.

reinforcement_learning/rlclientlib/schema/OutcomeEvent.fbs
```diff
@@ -0,0 +1,12 @@
...
+
+namespace VW.Events;
+
+table OutcomeEvent {
+    EventId:string;
+    v:string;
+}
+
+table OutcomeEventBatch {
+	Events:[OutcomeEvent];
```
sharathmalladi: >	 [](start = 0, length = 1)

nit: inconsistent tabs vs spaces in this file.

reinforcement_learning/rlclientlib/logger/event_batcher.h
```diff
@@ -0,0 +1,87 @@
...
+
+      if (batch_size <= 0) {
+        throw("event_batcher::batch_serialize batch_size is 0 - message is too large.");
+      }
+
+      if (std::is_same<TEvent, ranking_event>::value) {
+        std::vector<flatbuffers::Offset<RankingEvent>> events_offset;
+        flatbuffers::FlatBufferBuilder builder(_send_high_water_mark);
```
ataymano: also, can you make flatbuffers not visible outside of ranking_event / interaction event level?

reinforcement_learning/rlclientlib/logger/eventhub_client.cc
ataymano: Vector of unsigned char is not abstraction - if we decide to change it to
`class NewMessage {vector<unsigned char> data; bool someFlag;}`
it requires almost the same amount of changes as you are doing now.

I am talking about having first PR something based on:
`class i_sender
{
  using message_t = std::string;
  int send(message_t&& data, api_status* status = nullptr);
  ...
}`
This review is quite simple and will cover almost every signature change that you have in your current PR with 0 logical changes.

Remaining part will be almost completely within ranking_event.cc

homezcx: To me this abstraction is good to have, but may not a must. If we want this, I prefer we can have a follow up PR to address this.  We can discuss this offline.

I do agree if we all agree flatbuffer approach at the begining (e.g. a month back) and want to do this PR incrementally, this is the best approach.

---
In reply to: [228333588](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228333588) [](ancestors = 228333588)

reinforcement_learning/rlclientlib/schema/OutcomeEvent.fbs
homezcx: yes. Will fix this.

---
In reply to: [228273114](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228273114) [](ancestors = 228273114)

reinforcement_learning/rlclientlib/ranking_event.cc
```diff
@@ -56,79 +55,133 @@ namespace reinforcement_learning {
...
+    , _a_vector(std::move(other._a_vector))
+    , _p_vector(std::move(other._p_vector))
+    , _model_id(std::move(other._model_id))
+  { }
 
   ranking_event& ranking_event::operator=(ranking_event&& other) {
     if (&other != this) {
       event::operator=(std::move(other));
       _body = std::move(other._body);
+      _context = std::move(other._context);
```
ataymano: why do we need another copies of message content here and postpone flatbuffer message building?

reinforcement_learning/rlclientlib/ranking_event.cc
```diff
@@ -56,79 +55,133 @@ namespace reinforcement_learning {
...
     }
 
     //add model id
     oss << R"(],"VWState":{"m":")" << resp.get_model_id() << R"("})";
-
     return ranking_event(event_id, pass_prob, oss.str());
-	}
+  }
+
+  void ranking_event::serialize(u::data_buffer& oss) {
```
ataymano: Merge conflict: This method is not used anymore. On the other side there is no pdrop support in the new serialization.

reinforcement_learning/rlclientlib/ranking_event.h
```diff
@@ -32,28 +41,39 @@ namespace reinforcement_learning {
     ranking_event();
     ranking_event(ranking_event&& other);
     ranking_event& operator=(ranking_event&& other);
+    ranking_event(utility::data_buffer& oss, const char* event_id, const char* context,
```
ataymano: Merge conflict: these public constructors were replaced with static ranking_event::choose_rank, outcome_event::report_action_taken, outcome_event::report_outcome and are not called anymore.

reinforcement_learning/rlclientlib/logger/event_batcher.h
```diff
@@ -0,0 +1,38 @@
...
+    private:
+      size_t _batch_size = 400;
+      template<typename TEvent, typename TResultObject, typename TResultObjectBuilder>
+      void batch_serialize_internal(data_buffer& oss, size_t& remaining, event_queue<TEvent>& queue, size_t _send_high_water_mark, size_t batch_size);
+    };
+
+    template<typename TEvent>
+    void event_batcher::batch_serialize(data_buffer& oss, size_t& remaining, event_queue<TEvent>& queue, size_t _send_high_water_mark)
+    {
+      // This is used by unit_test
```
ataymano: Why do we need this method?

reinforcement_learning/rlclientlib/logger/event_batcher.h
```diff
@@ -0,0 +1,38 @@
...
+#include "event_queue.h"
+#include "ranking_event.h"
+#include "utility/data_buffer.h"
+
+namespace reinforcement_learning {
+  namespace utility {
+    class event_batcher {
+    public:
+      event_batcher();
+      void batch_serialize(data_buffer& oss, size_t& remaining, event_queue<ranking_event>& queue, size_t _send_high_water_mark);
```
ataymano: Can we make it as either single template method or single method within template class here?

reinforcement_learning/rlclientlib/logger/event_batcher.h
```diff
@@ -0,0 +1,38 @@
...
+      void batch_serialize(data_buffer& oss, size_t& remaining, event_queue<outcome_event>& queue, size_t _send_high_water_mark);
+
+      template<typename TEvent>
+      void batch_serialize(data_buffer& oss, size_t& remaining, event_queue<TEvent>& queue, size_t _send_high_water_mark);
+
+    private:
+      size_t _batch_size = 400;
```
ataymano: Can you double check what is going on between _batch_size and _send_high_water_mark? Looks like they have to be collapsed into one thing.

reinforcement_learning/rlclientlib/logger/event_batcher.h
homezcx: I spend most of my time yesterday on this.  The problem is that the logic need to know the type of the event so It can call different flatbuffer object builders. Not sure how we can do that in one method.

---
In reply to: [228610394](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228610394) [](ancestors = 228610394)

reinforcement_learning/rlclientlib/logger/event_batcher.h
homezcx: This is known issue and had been discussed for a while. The problem here is we won't know the flatbuffer size before we call the build to serialize it.

It is possible to remove _send_high_water_mark but this could be a breaking change. I remembered Rajan mentioned we will have plan to do fix length transport. In this case, this may be still needed.

With Rajan's proposed preamble header, we may be able to solve it.

---
In reply to: [228611953](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228611953) [](ancestors = 228611953)

reinforcement_learning/rlclientlib/logger/event_batcher.h
homezcx: The problem is async_batcher actually is designed to support any events. Unit tests are defining different events contains of string data, which does not have a flatbuffer schema.

---
In reply to: [228610098](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228610098) [](ancestors = 228610098)

reinforcement_learning/rlclientlib/ranking_event.cc
homezcx: This is single event data. Flatbuffer is defining an flatbuffer object contains "an array of object". It can only be built when the array is available. There is no way to build it here.

---
In reply to: [228608357](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228608357) [](ancestors = 228608357)

reinforcement_learning/rlclientlib/ranking_event.h
homezcx: I will double check this.

---
In reply to: [228609821](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228609821) [](ancestors = 228609821)

reinforcement_learning/rlclientlib/ranking_event.cc
homezcx: I will double check that. I ported the event schema from previous event_test.cc. If it was not tested there, it will likely be missed since we didn't have a schema before. Is it float?

---
In reply to: [228608998](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228608998) [](ancestors = 228608998)

reinforcement_learning/rlclientlib/logger/event_batcher.cc
```diff
@@ -0,0 +1,53 @@
...
+    {
+      batch_serialize_internal<ranking_event, RankingEvent, VW::Events::RankingEventBatchBuilder>(oss, remaining, queue, _send_high_water_mark, _batch_size);
+    }
+    void event_batcher::batch_serialize(data_buffer& oss, size_t& remaining, event_queue<outcome_event>& queue, size_t _send_high_water_mark)
+    {
+      batch_serialize_internal<outcome_event, OutcomeEvent, VW::Events::OutcomeEventBatchBuilder>(oss, remaining, queue, _send_high_water_mark, _batch_size);
+    }
+
+    template<typename TEvent, typename TResultObject, typename TResultObjectBuilder>
+    void event_batcher::batch_serialize_internal(data_buffer& oss, size_t& remaining, event_queue<TEvent>& queue, size_t _send_high_water_mark, size_t batch_size)
```
ataymano: since _batch_size is the member of the class, we do not need it here as input argument.
The same has to be applied to send_high_water_mark (and it has to be removed from async_batcher since this is the guy who owns batching logic)

reinforcement_learning/rlclientlib/logger/event_batcher.h
ataymano: Oh I see, they are both used in a different way - then please see the new comment on event_batcher.cc

build-linux.sh
```diff
@@ -3,11 +3,25 @@ set -e
...
+cd ..
+git clone git@github.com:google/flatbuffers.git
+cd ./flatbuffers
+git checkout 925c1d77fcc72636924c3c13428a34180c30f96f  # release commit
+cmake 
+make
+sudo make install
+sudo cp ./flatc /usr/local/bin/
+cd ..  # to ../
+cd ./vovowpal_wabbit
```
tyclintw: merge to one command 'cd ../vowpal_wabbit'. Also, vowpal_wabbit is spelled incorrectly.

vowpalwabbit/vw.sln
```diff
@@ -273,16 +273,14 @@ Global
...
 		{C5D19F4E-174D-40F0-8CD9-B92CBCE721E0}.Release|x86.ActiveCfg = Release|Win32
 		{C5D19F4E-174D-40F0-8CD9-B92CBCE721E0}.Release|x86.Build.0 = Release|Win32
-		{E86A833A-514D-424D-9FE7-154225031662}.Debug|Any CPU.ActiveCfg = Debug|Win32
+		{E86A833A-514D-424D-9FE7-154225031662}.Debug|Any CPU.ActiveCfg = Debug|x64
 		{E86A833A-514D-424D-9FE7-154225031662}.Debug|x64.ActiveCfg = Debug|x64
 		{E86A833A-514D-424D-9FE7-154225031662}.Debug|x64.Build.0 = Debug|x64
-		{E86A833A-514D-424D-9FE7-154225031662}.Debug|x86.ActiveCfg = Debug|Win32
-		{E86A833A-514D-424D-9FE7-154225031662}.Debug|x86.Build.0 = Debug|Win32
-		{E86A833A-514D-424D-9FE7-154225031662}.Release|Any CPU.ActiveCfg = Release|Win32
+		{E86A833A-514D-424D-9FE7-154225031662}.Debug|x86.ActiveCfg = Debug|x64
```
tyclintw: >x64 [](start = 69, length = 3)

Why is x86 targeting x64? This should stay Win32.

reinforcement_learning/rlclientlib/rlclientlib.vcxproj.filters
```diff
@@ -60,12 +62,19 @@
     <ClInclude Include="..\include\sender.h" />
     <ClInclude Include="..\include\object_factory.h" />
     <ClInclude Include="utility\watchdog.h" />
+    <ClInclude Include="generated\RankingEvent_generated.h" />
+    <ClInclude Include="logger\event_batcher.h" />
```
tyclintw: Included twice.

reinforcement_learning/rlclientlib/ranking_event.cc
ataymano: I mean building of RankingEvent/OutcomeEvent (basically serialize_eventhub_message logic). Why cannot we do in constructor and make reinforcement_learning::ranking event be the holder of RankingEvent only (and the same for outcome_event)?

homezcx: It cannot. We won't serialize RankingEvent/OutComeEvent. What we serialize is RankingEventBatch and OutComeEventBatch. Serialize  RankingEventBatch and "serialize RankingEvent and batch them" are not equivalent. The batch size is determined by event queue size. Batch size in the class is only a max queue size.

---
In reply to: [228634268](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228634268) [](ancestors = 228634268)

reinforcement_learning/rlclientlib/logger/event_batcher.h
homezcx: 1) I think it is. async_batcher only providers numbers of queue to event_batcher. EventBatcher is the flatbuffer serialization.
2) It might not be needed now. Rajan will have flatbuffer preamble header soon. I don't see a point to have flatbuffer to replace all unit tests now.

---
In reply to: [228641030](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228641030) [](ancestors = 228641030)

reinforcement_learning/rlclientlib/logger/event_batcher.cc
homezcx: We will need that. The logic right now is using batch_size to determine

1. with a maximum of 400 events will it fill high_water_mark? 
2. If yes, try a recursive call with half of the value. That is why batch_size is needed in the signature.


---
In reply to: [228620136](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1653#discussion_r228620136) [](ancestors = 228620136)

reinforcement_learning/rlclientlib/model_mgmt/restapi_data_transport.h
```diff
@@ -14,7 +14,7 @@ namespace reinforcement_learning {
     int check(api_status* status);
   private:
     using time_t = std::chrono::time_point<std::chrono::system_clock>;
-    int get_data_info(::utility::datetime& last_modified, ::utility::size64_t& sz, api_status* status);
+    int get_data_info(::utility::datetime& last_modified, ::utility::size64_t& sz, api_status* status);
```
homezcx: extra space.

build-linux.sh
```diff
@@ -3,11 +3,24 @@ set -e
...
 
+# add flat buffers
+cd ..
+git clone https://github.com/google/flatbuffers.git
+cd ./flatbuffers
+git checkout 925c1d77fcc72636924c3c13428a34180c30f96f  # release commit
+cmake .
+make
+sudo make install
+sudo cp ./flatc /usr/local/bin/
```
jackgerrits: Does the install target not install the executable?

build-linux.sh
```diff
@@ -3,11 +3,24 @@ set -e
...
+cd ../vw
+#
```
jackgerrits: remove

build-linux.sh
```diff
@@ -3,11 +3,24 @@ set -e
 
 export PATH="$HOME/miniconda/bin:$PATH"
 
+# add flat buffers
+cd ..
+git clone https://github.com/google/flatbuffers.git
+cd ./flatbuffers
+git checkout 925c1d77fcc72636924c3c13428a34180c30f96f  # release commit
```
jackgerrits: Can you change the comment to which release of flatbuffers we are targeting?

reinforcement_learning/rlclientlib/Makefile
```diff
@@ -5,9 +5,12 @@ INCLUDE = -I ../include -I ../../explore -I ../explore -I ../../rapidjson/includ
 
 .PHONY: default all clean
 
-default: $(TARGET)
+default: pre-build $(TARGET)
 all: default
 
+pre-build:
```
jackgerrits: The makefile rule should specify the output of this command as the target. This target should then be a dependency for every target that relies on it

reinforcement_learning/rlclientlib/logger/event_batcher.cc
```diff
@@ -0,0 +1,53 @@
+#include "event_batcher.h"
+
+#include "generated/RankingEvent_generated.h"
+#include "generated/OutcomeEvent_generated.h"
+
+namespace reinforcement_learning {
+  namespace utility {
+
+    event_batcher::event_batcher()
```
jackgerrits: This is empty, make it default

reinforcement_learning/rlclientlib/logger/event_batcher.cc
```diff
@@ -0,0 +1,53 @@
...
+    void event_batcher::batch_serialize(data_buffer& oss, size_t& remaining, event_queue<outcome_event>& queue, size_t _send_high_water_mark)
+    {
+      batch_serialize_internal<outcome_event, OutcomeEvent, VW::Events::OutcomeEventBatchBuilder>(oss, remaining, queue, _send_high_water_mark, _batch_size);
+    }
+
+    template<typename TEvent, typename TResultObject, typename TResultObjectBuilder>
+    void event_batcher::batch_serialize_internal(data_buffer& oss, size_t& remaining, event_queue<TEvent>& queue, size_t _send_high_water_mark, size_t batch_size)
+    {
+      if (_batch_size <= 0) {
+        throw("event_batcher::batch_serialize batch_size is 0 - message is too large.");
```
jackgerrits: This shouldn't throw but it should return an error code to be consistent with the rest of the codebase

reinforcement_learning/rlclientlib/logger/event_batcher.cc
```diff
@@ -0,0 +1,53 @@
...
+      }
+
+      std::vector<flatbuffers::Offset<TResultObject>> events_offset;
+      flatbuffers::FlatBufferBuilder builder(_send_high_water_mark);
+      TEvent evt;
+      for (int i = 0; remaining > 0 && oss.size() < _send_high_water_mark; i++) {
+        if (i == batch_size) break;
```
jackgerrits: Prefer to avoid break in the loop if you can. And if it is needed, split it into several lines and add a comment

reinforcement_learning/rlclientlib/ranking_event.cc
```diff
@@ -51,100 +50,127 @@ namespace reinforcement_learning {
...
     return *this;
   }
 
-  void ranking_event::serialize(u::data_buffer& oss) {
-    oss << _body;
-    if (_pass_prob < 1) {
-      oss << R"(,"pdrop":)" << (1 - _pass_prob);
-    }
-    oss << R"(})";
+  flatbuffers::Offset<RankingEvent> ranking_event::serialize_eventhub_message(flatbuffers::FlatBufferBuilder& builder) {
```
jackgerrits: I am not convinced we need to tie the implementation of our events to flatbuffers so tightly. Can this be a static method elsewhere or a different class?

reinforcement_learning/rlclientlib/ranking_event.h
```diff
@@ -33,27 +41,30 @@ namespace reinforcement_learning {
     ranking_event(ranking_event&& other);
     ranking_event& operator=(ranking_event&& other);
 
-    virtual void serialize(utility::data_buffer& buffer) override;
-  
+    virtual flatbuffers::Offset<RankingEvent> serialize_eventhub_message(flatbuffers::FlatBufferBuilder& builder);
```
jackgerrits: Why is this virtual?

reinforcement_learning/rlclientlib/ranking_event.h
```diff
@@ -33,27 +41,30 @@ namespace reinforcement_learning {
...
       unsigned int flags, const ranking_response& resp, float pass_prob = 1);
 
   private:
-    ranking_event(const char* event_id, float pass_prob, const std::string& body);
+    ranking_event(const char* event_id, float pass_prob, const char* context, const ranking_response& response);
 
-  private:
-    std::string _body;
+    std::string _context;
+    std::vector<uint64_t> _a_vector;
```
jackgerrits: can you please make these names clearer?

reinforcement_learning/rlclientlib/utility/data_buffer.cc
```diff
@@ -3,50 +3,43 @@
...
+    void data_buffer::remove_last() { _buffer.pop_back(); }
 
-  buffer_factory::buffer_factory(const translate_func& _translate) : translate(_translate) {}
+    void data_buffer::append(const unsigned char* data, size_t len) {
+      _buffer.reserve(_buffer.size() + len);
+      _buffer.insert(_buffer.end(), data, data + len);
+    }
 
-  data_buffer* buffer_factory::operator()() const { return new data_buffer(translate); }
+    std::vector<unsigned char> data_buffer::buffer() {
```
jackgerrits: Should this be a copy? Shouldn't we be returning a reference or pointer here?

reinforcement_learning/rlclientlib/utility/data_buffer.cc
```diff
@@ -3,50 +3,43 @@
...
-    return _buffer.size();
-  }
-
-  void data_buffer::remove_last() { _buffer.pop_back(); }
+    void data_buffer::remove_last() { _buffer.pop_back(); }
 
-  buffer_factory::buffer_factory(const translate_func& _translate) : translate(_translate) {}
+    void data_buffer::append(const unsigned char* data, size_t len) {
+      _buffer.reserve(_buffer.size() + len);
+      _buffer.insert(_buffer.end(), data, data + len);
```
jackgerrits: Since we are only reserving as much space as we are adding - is there a perf gain by doing this in two steps?

reinforcement_learning/rlclientlib/utility/data_buffer.cc
```diff
@@ -3,50 +3,43 @@
 #include <iterator>
 #include <iostream>
 
-namespace reinforcement_learning { namespace utility {
-  data_buffer::data_buffer() = default;
+namespace reinforcement_learning {
+  namespace utility {
+    data_buffer::data_buffer() = default;
 
-  data_buffer::data_buffer(const translate_func& _translate) : translate(_translate) {}
```
jackgerrits: @ataymano Why did we add the translate function initially? Is it okay to remove?

reinforcement_learning/rlclientlib/utility/data_buffer.cc
```diff
@@ -3,50 +3,43 @@
...
+      return *this;
+    }
 
-  data_buffer& data_buffer::operator<<(float rhs) { return operator<<(std::to_string(rhs)); }
+    data_buffer& data_buffer::operator<<(const char* data) { return operator<<(std::string(data)); }
+    data_buffer& data_buffer::operator<<(size_t rhs) { return operator<<(std::to_string(rhs)); }
+    data_buffer& data_buffer::operator<<(float rhs) { return operator<<(std::to_string(rhs)); }
 
-  translate_func::translate_func() : empty(true) {}
+    buffer_factory::buffer_factory() {}
```
jackgerrits: This can be removed or made default

reinforcement_learning/rlclientlib/logger/event_batcher.h
```diff
@@ -0,0 +1,38 @@
+#pragma once
+#include "event_queue.h"
+#include "ranking_event.h"
+#include "utility/data_buffer.h"
+
+namespace reinforcement_learning {
+  namespace utility {
+    class event_batcher {
```
jackgerrits: I'm not sure I understand the difference between event_batcher and async_batcher. Could you explain?

build-linux.sh
homezcx: no, it won't.

reinforcement_learning/rlclientlib/logger/event_batcher.cc
homezcx: This will only happen in theory. The caller is async_batcher which does not have return code propagated to the upper caller, not like the other API response case.

reinforcement_learning/rlclientlib/logger/event_batcher.cc
homezcx: This is required. Please see detains in the discussion with Alexey.

reinforcement_learning/rlclientlib/utility/data_buffer.cc
homezcx: It was reusing the same logic as before, reserve before insert. Might need to go back to history to figure why we had it.

reinforcement_learning/rlclientlib/utility/data_buffer.cc
homezcx: As far as I can answer, previous implementation only works for ds json. The translate function is to process "\n".

reinforcement_learning/rlclientlib/logger/event_batcher.h
homezcx: event_batcher is to create the eventhub messages (which is an flatbuffer object, that logically contains a batch of events).

async_batcher was just previous logic, which process the eventqueue and send messages.

reinforcement_learning/rlclientlib/ranking_event.cc
homezcx: It needs to access all fields in events. If we move it to anther class or static method, it will break the encapsulation of events anyway. In this sense, I think staying here should be fine.

reinforcement_learning/rlclientlib/utility/data_buffer.cc
homezcx: it's git hub diff confused you. The real diff of this line should be line 43.

reinforcement_learning/rlclientlib/utility/data_buffer.cc
ataymano: Yes, it was required when we used \n as delimiters (to remove it from message bodies). Should be removed here.

reinforcement_learning/rlclientlib/logger/event_batcher.cc
homezcx: This is quite self-explained. It will loop to the batch_size defined in the code.

build-linux.sh
homezcx: added.

build-linux.sh
homezcx: removed.

reinforcement_learning/rlclientlib/Makefile
homezcx: Not sure what does this mean. Let's take it offline.

reinforcement_learning/rlclientlib/logger/event_batcher.cc
homezcx: done

reinforcement_learning/rlclientlib/ranking_event.h
homezcx: removed.

reinforcement_learning/rlclientlib/ranking_event.h
homezcx: These are exact names "a", "p" we are using in ds json. Changing to other name will cause more confusion.

reinforcement_learning/rlclientlib/utility/data_buffer.cc
homezcx: done

reinforcement_learning/rlclientlib/ranking_event.h
homezcx: See previous definition before flatbuffer:
R"({"Version":"1","EventId":"event_id","a":[],"c":{context},"p":[],"VWState":{"m":"model_id"}})"

reinforcement_learning/rlclientlib/logger/event_batcher.cc
jackgerrits: In that case it should use the background error callback

reinforcement_learning/rlclientlib/ranking_event.cc
jackgerrits: The data should be able to be accessed externally from ranking_event. It is essentially a data struct. I don't think it is acceptable leaving this in the class.

reinforcement_learning/rlclientlib/ranking_event.h
jackgerrits: Please add a comment then, the ds json isn't exactly self explanatory either

reinforcement_learning/rlclientlib/logger/event_batcher.h
jackgerrits: Given this changes how they interact can you please rename to make this clearer?

reinforcement_learning/rlclientlib/utility/data_buffer.cc
jackgerrits: I'm not sure I understand? We are still returning a complete copy of the buffer here. Generally a function named liked this would return a pointer to the underlying buffer

homezcx: This is a new method. Previous method is in line 43 in right patch, leaving unchanged.

homezcx: This can be changed to get_buffer()

homezcx: I think I got what you mean - you are suggesting returning "std::vector<unsigned char>&"?

reinforcement_learning/rlclientlib/schema/OutcomeEvent.fbs
```diff
@@ -0,0 +1,12 @@
+﻿// EventHubInteraction Schema used by FlatBuffer
+
+namespace VW.Events;
+
+table OutcomeEvent {
```
yannstad: The deferred action should also be defined in the OutcomeEvent.
@ataymano do you confirm?

reinforcement_learning/rlclientlib/utility/data_buffer.cc
jackgerrits: Yeah that or just a pointer

reinforcement_learning/rlclientlib/schema/OutcomeEvent.fbs
rajan-chari: Can we use c++ style naming convention please?  This library is external facing and would be good to not have to switch to different styles for generated code.

reinforcement_learning/rlclientlib/logger/event_batcher.h
```diff
@@ -0,0 +1,38 @@
+#pragma once
```
rajan-chari: event_batcher seems to have overlapping batching logic with other parts of code.  Also seems to violate the spirit of high water mark.

reinforcement_learning/rlclientlib/logger/event_batcher.cc
```diff
@@ -0,0 +1,51 @@
...
+    void event_batcher::batch_serialize_internal(data_buffer& oss, size_t& remaining, event_queue<TEvent>& queue, size_t _send_high_water_mark, size_t batch_size)
+    {
+      if (_batch_size <= 0) {
+        throw("event_batcher::batch_serialize batch_size is 0 - message is too large.");
+      }
+
+      std::vector<flatbuffers::Offset<TResultObject>> events_offset;
+      flatbuffers::FlatBufferBuilder builder(_send_high_water_mark);
+      TEvent evt;
+      for (int i = 0; remaining > 0 && oss.size() < _send_high_water_mark; i++) {
```
rajan-chari: >int [](start = 11, length = 3)

size_t ?

reinforcement_learning/rlclientlib/logger/event_batcher.cc
```diff
@@ -0,0 +1,51 @@
...
+
+    template<typename TEvent, typename TResultObject, typename TResultObjectBuilder>
+    void event_batcher::batch_serialize_internal(data_buffer& oss, size_t& remaining, event_queue<TEvent>& queue, size_t _send_high_water_mark, size_t batch_size)
+    {
+      if (_batch_size <= 0) {
+        throw("event_batcher::batch_serialize batch_size is 0 - message is too large.");
+      }
+
+      std::vector<flatbuffers::Offset<TResultObject>> events_offset;
+      flatbuffers::FlatBufferBuilder builder(_send_high_water_mark);
```
rajan-chari: During the performance review, I thought we agreed on using passing in a custom allocator so we can avoid allocation here.

reinforcement_learning/rlclientlib/logger/event_batcher.cc
```diff
@@ -0,0 +1,51 @@
...
+      auto orc = ranking_event_batch_builder.Finish();
+      builder.Finish(orc);
+
+      size_t size = builder.GetSize();
+      if (size > _send_high_water_mark) {
+        batch_serialize_internal<TEvent, TResultObject, TResultObjectBuilder>(oss, remaining, queue, _send_high_water_mark, batch_size / 2);
+        return;
+      }
+
+      oss.append(builder.GetBufferPointer(), builder.GetSize());
```
rajan-chari: >oss [](start = 6, length = 3)

Avoiding this copy gave us large factor performance gain in my tests.  I shared results in the POC onenote document.
Passing in this buffer to a custom allocator would avoid this copy.

reinforcement_learning/rlclientlib/logger/event_batcher.cc
rajan-chari: >oss.size() < _send_high_water_mark [](start = 39, length = 34)

This loop does not seem to write to oss.  This can exceed send_high_water_mark by a large number.

reinforcement_learning/rlclientlib/ranking_event.h
```diff
@@ -33,38 +41,45 @@ namespace reinforcement_learning {
...
   //serializable outcome event
   class outcome_event : public event {
   public:
     outcome_event();
+
     outcome_event(outcome_event&& other);
     outcome_event& operator=(outcome_event&& other);
 
-    virtual void serialize(utility::data_buffer& buffer) override;
+    std::string str();
```
rajan-chari: a) to_json_string might be a better name
b) Please see comment on FB serialization knowledge living outside this class.  This method could benefit from a similar mechanism.

rajan-chari: Why not pass in data_buffer reference?  Is this used somewhere?

reinforcement_learning/rlclientlib/schema/RankingEvent.fbs
```diff
@@ -0,0 +1,22 @@
+﻿// EventHubInteraction Schema used by FlatBuffer
+
+namespace VW.Events;
+
+// struct has better performace than table. But use table here in case format changes in future.
+table VWStateType {
```
rajan-chari: >VWStateType [](start = 6, length = 11)

Can we use c++ style naming convention please?  This library is external facing and would be good to not have to switch to different styles for generated code.

reinforcement_learning/rlclientlib/schema/RankingEvent.fbs
```diff
@@ -0,0 +1,22 @@
...
+    m:string;        // model_id
+}
+
+table RankingEvent {
+    Version:short = 1;
+    EventId:string;
+    DeferredAction:bool = false;
+    a:[uint64];
+    c:string;          // context
```
rajan-chari: >c [](start = 4, length = 1)

Should be called context_json_string

reinforcement_learning/rlclientlib/schema/RankingEvent.fbs
```diff
@@ -0,0 +1,22 @@
...
+    p:[float];
```
rajan-chari: >p [](start = 4, length = 1)

rename:  suggestions a) pdf b) probability_vector c) probabilities?

reinforcement_learning/rlclientlib/schema/RankingEvent.fbs
```diff
@@ -0,0 +1,22 @@
+﻿// EventHubInteraction Schema used by FlatBuffer
+
+namespace VW.Events;
```
rajan-chari: >VW. [](start = 10, length = 3)

rename:  reinforcement_learning.serialization.fb.events

reinforcement_learning/rlclientlib/schema/RankingEvent.fbs
rajan-chari: >VWStateType [](start = 6, length = 11)

rename: inference_model_state

reinforcement_learning/rlclientlib/schema/RankingEvent.fbs
```diff
@@ -0,0 +1,22 @@
...
+
+// struct has better performace than table. But use table here in case format changes in future.
+table VWStateType {
+    m:string;        // model_id
```
rajan-chari: rename: model_id

reinforcement_learning/rlclientlib/schema/RankingEvent.fbs
```diff
@@ -0,0 +1,22 @@
...
+}
+
+table RankingEvent {
+    Version:short = 1;
+    EventId:string;
+    DeferredAction:bool = false;
+    a:[uint64];
+    c:string;          // context
+    p:[float];
+    VWState:VWStateType;
```
rajan-chari: rename{ model_state

reinforcement_learning/rlclientlib/schema/RankingEvent.fbs
```diff
@@ -0,0 +1,22 @@
+﻿// EventHubInteraction Schema used by FlatBuffer
+
+namespace VW.Events;
+
+// struct has better performace than table. But use table here in case format changes in future.
+table VWStateType {
+    m:string;        // model_id
+}
+
+table RankingEvent {
```
rajan-chari: Are these normalized with Yann's design doc on model format?

reinforcement_learning/rlclientlib/schema/OutcomeEvent.fbs
ataymano: Yes, it should.

PR #1654, opened 2018-10-24
Support retrying failed web requests for eventhub_client sends

jackgerrits:
- Retry n times before counting request as failed
- Added unit tests for retry and ensuring order
- Small refactor of authorization code in eventhub_client


reinforcement_learning/rlclientlib/logger/eventhub_client.h
```diff
@@ -68,8 +102,10 @@ namespace reinforcement_learning {
     std::string _authorization;
     long long _authorization_valid_until; //in seconds
     std::mutex _mutex;
-    moving_queue<http_request_task> _tasks;
+    std::mutex _mutex_http_tasks;
```
ataymano: isn't it better to turn moving_queue into thread-safe one and have this mutex there?

ataymano: we actually had it done before (before previous moving_queue implementation turned into event_queue)

jackgerrits: The problem is that we pop and then join which may take some time. I am thinking now that the synchronization may need to occur at a slightly greater scope. Let me chat with you offline.

rajan-chari:
Merged Thanks

PR #1657, opened 2018-10-26
Ataymano/python interface fix

ataymano:


reinforcement_learning/doc/python/python.dox
```diff
@@ -58,28 +58,37 @@ Note: `https://aka.ms/rl_client_python` points to a binary targeting x64 Linux P
...
 \param context_json Contains action, action features and context features in JSON format
+\param deferred Identifies if action is deferred or not. Deferred one requires additional call of report_action_taken to be invovled in training.
 \exception RuntimeError Thrown if an internal error occurs
 \return model_id, chosen_action_id, actions_probabilities
 \return model_id - Get the model id. Every call to choose action is associated with a unique model used to predict. A unique model id is associated with each unique model. (This is set internally by the API)
 \return chosen_action_id - ID of action chosen by model
 \return actions_probabilities - List of pairs (action_id, probability) ranked by the model's prediction
 
-\fn rl_client::live_model::choose_rank(context_json)
+\fn rl_client::live_model::choose_rank(context_json, deffered = False)
```
yannstad: typo: `deffered` => `deferred`

reinforcement_learning/doc/python/python.dox
```diff
@@ -58,28 +58,37 @@ Note: `https://aka.ms/rl_client_python` points to a binary targeting x64 Linux P
...
 \memberof rl_client::live_model
 \brief Choose an action, given a list of actions, action features and context features. The inference library chooses an action by creating a probability distribution over the actions and then sampling from it.
 \param context_json Contains action, action features and context features in JSON format
+\param deferred Identifies if action is deferred or not. Deferred one requires additional call of report_action_taken to be invovled in training.
```
yannstad: typo: `invovled` => `involved`

reinforcement_learning/examples/python/basic_usage.py
```diff
@@ -16,23 +16,31 @@ def main():
     model = rl_client.live_model(config, test_cb)
     model.init()
 
-    event_id = "event_id"
+    uuid = "event_id"
```
yannstad: why did you change the var name?

reinforcement_learning/examples/python/basic_usage.py
```diff
@@ -16,23 +16,31 @@ def main():
...
 
-    model_id, chosen_action_id, actions_probabilities, event_id = model.choose_rank(context, deferred = True)
+    model_id, chosen_action_id, actions_probabilities, event_id = model.choose_rank(context, deferred = False)
+    model.report_action_taken(event_id)
+    print("event_id: " + event_id)
+    print("model_id: " + model_id)
+    print("chosen action id: " + str(chosen_action_id))
+    print("actions probabilities list: " + str(actions_probabilities))
+
+    model_id, chosen_action_id, actions_probabilities = model.choose_rank(context, deferred=True, event_id='asd')
```
yannstad: Maybe a more realistic event_id example is preferable?
Like for example: `e379df0b66fe48d1839e2a85a793c944y23fmJ`

reinforcement_learning/examples/python/basic_usage.py
tyclintw: Yes, let's just leave the original naming for clarity.

reinforcement_learning/examples/python/basic_usage.py
tyclintw: [nit] maintain the original argument ordering (context, event_id, deferred)

reinforcement_learning/doc/python/python.dox
marco-rossi29: deffered -> deferred

reinforcement_learning/examples/python/basic_usage.py
marco-rossi29: For clarity, I would do:
```
import uuid  # add this on top of file

uuid = str(uuid.uuid4())
```

reinforcement_learning/examples/python/basic_usage.py
marco-rossi29: I agree with Yann and Tyler. I would do:
import uuid
uuid = str(uuid.uuid4())
model_id, chosen_action_id, actions_probabilities = model.choose_rank(context, event_id=uuid, deferred=True)

marco-rossi29: I agree with Yann and Tyler. I would do:
import uuid
uuid = str(uuid.uuid4())
model_id, chosen_action_id, actions_probabilities = model.choose_rank(context, event_id=uuid, deferred=True)

yannstad: Actually, i thought we said that the deferred action logic will not be exposed except for advanced users.
Maybe we can omit to give an example where deferred is true?

marco-rossi29: That's a good point actually. @ataymano what do you think?

ataymano: are we exposing example? Even if we do, it is secondary thing compare to documentation in my opinion. Should I also remove it from docs?

ataymano: it was intentionally, to test that order is not important for named variables

ataymano: I can reorder though

yannstad: Neither one nor the other IMO

PR #1661, opened 2018-10-30
Ataymano/no train fix

ataymano:
Offline experimentation fix


JohnLangford:
Appveyor build is failing.

vowpalwabbit/example.cc
```diff
@@ -241,4 +242,11 @@ void clear_seq_and_finish_examples(vw& all, multi_ex& ec_seq)
   ec_seq.clear();
 }
 
+void return_unused_examples(vw& all, v_array<example*>& examples)
```
JohnLangford: return_multiline_example instead?

vowpalwabbit/parser.cc
```diff
@@ -787,6 +787,19 @@ example& get_unused_example(vw* all)
   }
 }
 
+void return_unused_example(vw& all, example& ec)
```
JohnLangford: Why can't we use finish_example() instead?

ataymano:
>           Appveyor build is failing.
It was cancelled. Now it passed

JohnLangford:
Why did the build fail?

PR #1668, opened 2018-11-02
api_context_format.dox - Small improvements

marco-rossi29:


reinforcement_learning/doc/cpp/api_context_format.dox
```diff
@@ -10,13 +10,14 @@ The format of this context is a JSON string, with a specific structure. This JSO
 
 Typically, the shared features describe a user, while the action features describe all the possible actions that can be taken for this user.
 
+In the following, we discuss the expected format for shared features and action features separately, and then at the end we provide an exampleof the expected context.
```
yannstad:
```suggestion
In the following, we discuss the expected format for shared features and action features separately, and then at the end we provide an example of the expected context.
```

reinforcement_learning/doc/cpp/api_context_format.dox
```diff
@@ -92,12 +97,12 @@ becomes
 ~~~~~
 
 
-A complete example with both shared and action features:
+A full example of the expexted context (note that newlines shouldn't be included):
```
yannstad:
```suggestion
A full example of the expected context (note that newlines shouldn't be included):
```

PR #1670, opened 2018-11-04
Use cerr for parser warnings

### LINKED ISSUE #1539: Parser warnings go to standard output instead of standard error, and corrupt the output of libvw users
I'm using Vowpal Wabbit as a library in a program that writes structured data to standard out.

That code managed to construct a Vowpal Wabbit example with a NaN in it. This didn't cause any errors (the program ran through successfully), but the output was corrupted, because Vowpal Wabbit injected its own text into it.

It looks like this warning code here, which prints to standard output, really [ought to](https://stackoverflow.com/q/1430956) print to standard error:

https://github.com/JohnLangford/vowpal_wabbit/blob/24233e238c3e957066014d85ce13cd3b67d996e5/vowpalwabbit/parse_example.cc#L78-L83

palmerlao:
closes #1539

### DIFF
--- a/vowpalwabbit/parse_example.cc
+++ b/vowpalwabbit/parse_example.cc
@@ -77,7 +77,7 @@ public:
 
   inline void parserWarning(const char* message, char* begin, char* pos, const char* message2)
   {
-    cout << message << std::string(begin, pos - begin).c_str() << message2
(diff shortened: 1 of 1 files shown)


JohnLangford:
Merged, thanks :-)

PR #1679, opened 2018-11-14
Fix crash when empty multi_ex is supplied for --cb_explore_adf

jackgerrits:
Will throw when an empty multi_ex is supplied

### DIFF
--- a/test/unit_test/CMakeLists.txt
+++ b/test/unit_test/CMakeLists.txt
@@ -1,4 +1,4 @@
-add_executable(vw-unit-test.out main.cc stable_unique_tests.cc)
+add_executable(vw-unit-test.out main.cc cb_explore_adf_test.cc explore_test.cc stable_unique_tests.cc)
 
 # Add the include directories from vw target for testing
 target_include_directories(vw-unit-test.out PRIVATE $<TARGET_PROPERTY:vw,INCLUDE_DIRECTORIES>)
--- a/vowpalwabbit/cb_explore_adf.cc
+++ b/vowpalwabbit/cb_explore_adf.cc
@@ -90,31 +90,38 @@ template<class T> void swap(T& ele1, T& ele2)
   ele1 = temp;
 }
 
+// Validates a multiline example collection as a valid sequence for action dependent features format.
 example* test_adf_sequence(multi_ex& ec_seq)
 {
+  if (ec_seq.size() == 0)
+    THROW("cb_adf: At least line must be provided for an example to be valid.");
+
   uint32_t count = 0;
   example* ret = nullptr;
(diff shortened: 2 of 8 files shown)


JohnLangford:
Why is appveyor barfing?

test/unit_test/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-add_executable(vw-unit-test.out main.cc stable_unique_tests.cc)
+add_executable(vw-unit-test.out main.cc cb_explore_adf_test.cc explore_test.cc stable_unique_tests.cc)
```
JohnLangford: Is there a reason to use extra tests of this sort rather than RunTests?  RunTests seems somehow cleaner since you don't need to recompile.

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -90,31 +90,38 @@ template<class T> void swap(T& ele1, T& ele2)
   ele1 = temp;
 }
 
+// Validates a multiline example collection as a valid sequence for action dependent features format.
 example* test_adf_sequence(multi_ex& ec_seq)
 {
+  if (ec_seq.size() == 0)
+    THROW("cb_adf: At least line must be provided for an example to be valid.");
```
JohnLangford: We should use a semantic comment instead.  "At least one action must be provided for an example to be valid".

JohnLangford:
Overall, this looks like a reasonable fix and certainly better than what we have.   Better in some ways would be modifying the parser to just skip examples with no actions.

jackgerrits:
It's possible that the parser actually catches this and ignores a line like this (I think this is very likely). This can occur when you use the apis directly like we do, instead of stdin

jackgerrits:
For appveyor, it looks like the unit test project is misconfigured. I'll fix it up.

test/unit_test/CMakeLists.txt
jackgerrits: I believe this sort of code issue is best tested using a unit test, especially since it might not repro with command line and stdin. I think we should aim to add a test/unit test for each bug as we fix them. Helps reduce the risk of regression and builds up the testing suite over time

lokitoth:
>           Overall, this looks like a reasonable fix and certainly better than what we have.   Better in some ways would be modifying the parser to just skip examples with no actions.

When I was trying to run the data set with this example through the executable, it wrote an error asking me to check my data.

JohnLangford:
Ok.  Maybe tweak the error message to be something more meaningful, fix the appveyor issue, and let's merge.

jackgerrits:
I've updated the comment and hopefully fixed the build issue.

JohnLangford:
Merged, thanks.

PR #1681, opened 2018-11-20
cbify: --cbify_ldf for multiline (csoaa_ldf) input datasets

albietz:
This PR adds support for reading cost-sensitive classification datasets in LDF format in cbify, which turns them into contextual bandit examples in ADF format.

cc @JohnLangford @alekh


vowpalwabbit/cbify.cc
```diff
@@ -284,3 +435,44 @@ base_learner* cbify_setup(options_i& options, vw& all)
 
   return make_base(*l);
 }
+
+base_learner* cbifyldf_setup(arguments& arg)
+{
+  auto data = scoped_calloc_or_throw<cbify>();
+
+  if (arg.new_options("Make csoaa_ldf into Contextual Bandit")
```
JohnLangford: This needs to be converted to the tweaked option parsing.

JohnLangford:
I'm finally getting to this---sorry for the long wait (backstory: we had a baby).  

The option parsing code has shifted a little bit since this was written.  Can you tweak to update?

JohnLangford:
Something failed in test 175 on appveyor.  Ideas?

albietz:
The test seems to run fine locally for me, can we retry with a clean build? Also the new test is labeled 176 in the RunTests file comment, not 175 -- is this discrepancy expected?

JohnLangford:
maybe try bumping it.

I suspect we have an off-by-one in accounting for the test numbers...?

albietz:
Is there a way to check that the parsed C# test is reading from the right file / doing the right thing? The error seems to suggest that the input data is in single-line format while the test expects multi-line, but it's hard to see what's going on without access to a windows machine...

JohnLangford:
Not without running on windows, as far as I know.

albietz:
It clearly seems like something odd is happening on the C# testing code, in particular, it seems `IsMultilineData` returns `false` in my test on this line https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/cs/unittest/RunTestsHelper.cs#L45, although other tests using the same data file seem to run without error.

I'm not sure I can figure this out myself, given that the only way for me to debug currently is to commit something and look at the appveyor output... Can someone with access to the C# tools take a look? Or do you have other suggestions for next steps otherwise? (maybe I can just comment out the unit test...?)

JohnLangford:
Let me see what I can do on this end.

-John

albietz:
Cool, thanks!

JohnLangford:
@yannstad will take a look at it.

yannstad:
@albietz : basically, you enabled (not intentionally) the test 175, that was ignored (by mistake/bug) on Windows. And this test is failing when it is enabled.
The PR #1751 fixes the test issue, it will be merged soon. Then you can update your PR.

albietz:
I see, I guess I was assuming that 175 was my own test (it is labeled 176 in the comment in RunTests, but it seems those numbers don't really correspond to the final ordering anyway... and in the travis tests the last one - i.e. mine - is 175). I'll wait for the other PR to be merged and will try again.

Thanks for looking into this!

albietz:
[ FYI - I added a PR to show the correct test labels in `make test` and Travis logs, see #1753. I suppose the C# tests are already doing this the right way based on @yannstad's comments. ]

yannstad:
The fix is merged now, feel free to update your PR anytime.

JohnLangford:
Merged, thanks :-)

PR #1682, opened 2018-11-24
add multiclass support

awant:
add support for multiclass classification problems.
Also, find out that we have to speed up script to make it actually useful.

### DIFF
--- a/utl/vw-hyperopt.py
+++ b/utl/vw-hyperopt.py
@@ -176,6 +176,7 @@ class HyperOptimizer(object):
         self.max_evals = max_evals
         self.searcher = searcher
         self.is_regression = is_regression
+        self.labels_clf_count = 0
 
         self.trials = Trials()
         self.current_trial = 0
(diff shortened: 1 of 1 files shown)


utl/vw-hyperopt.py
```diff
@@ -219,11 +220,16 @@ def get_hyperparam_string(self, **kwargs):
     def compose_vw_train_command(self):
         data_part = ('vw -d %s -f %s --holdout_off -c '
                      % (self.train_set, self.train_model))
+        if self.labels_clf_count > 2: # multiclass, should take probabilities
+            data_part += ('--oaa %s --loss_function=logistic --probabilities '
```
arielf: Would be nice to support other multiclass methods in addition to `--oaa`.
e.g.: `--ect`, `--recall_tree`, `--wap`, (with & without ldf) etc.
They all have similar interface (expect number of classes <k> as arg)

PR #1683, opened 2018-12-04
(1/2) Refactor arguments to a standalone interface

jackgerrits:
Creates a generic interface for configuration in VW and implements a concrete type for Boost program options


JohnLangford:
This pull request **introduces 1 alert** when merging f9a19e8d2bc9d212f01e1344b55fd6091bdb0779 into fac3c061e47f3feb65919ea5c807a63b59a47da2 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-d8bf0c48c2c812b5574fe3d10b5cd17514972f0b)

**new alerts:**

* 1 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

JohnLangford:
This pull request **introduces 1 alert** when merging 50fc574d90592daead7863dd4819e3f049ee68f1 into fac3c061e47f3feb65919ea5c807a63b59a47da2 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-f26e8920887e87e7a8677ed780f01745a9ed81a5)

**new alerts:**

* 1 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

vowpalwabbit/arguments.h
```diff
@@ -0,0 +1,157 @@
...
+
+    typed_argument& keep(bool keep = true) {
+      m_keep = keep; return *this;
+    }
+
+    bool value_supplied() {
+      return m_value.get() != nullptr;
+    }
+
+    typed_argument& value(T& value) {
```
jackgerrits: Should setting this also set the value for all of m_locations. This sets a dangerous precedent about event based parameter/variable synchronization in the library though.

vowpalwabbit/arguments.cc
```diff
@@ -0,0 +1,15 @@
+#include "arguments.h"
+
+using namespace VW;
+
+bool VW::operator==(const base_argument& lhs, const base_argument& rhs) {
+  return lhs.m_name == rhs.m_name
```
JohnLangford: It seems semantically confusing to have a top-level operator==() which just deals arguments.  Maybe this should be scoped to an arguments type?

vowpalwabbit/arguments.h
```diff
@@ -0,0 +1,157 @@
...
+    std::string m_help = "";
+    std::string m_short_name = "";
+    bool m_keep = false;
+
+    virtual ~base_argument() {}
+  };
+
+  template<typename T>
+  struct typed_argument : base_argument {
+    typed_argument(std::string name, T* location)
```
JohnLangford: Maybe use T& here?  There is no need to do pointer operations.

vowpalwabbit/arguments.h
```diff
@@ -0,0 +1,157 @@
...
+    typed_argument& value(T& value) {
+      m_value = std::make_shared<T>(value);
+      return *this;
+    }
+
+    T value(){
+      return m_value ? *m_value : T();
+    }
+
+    std::vector<T*> m_locations;
```
JohnLangford: Why is this a vector?

vowpalwabbit/arguments.h
JohnLangford: I'm not quite following this.

vowpalwabbit/arguments.h
```diff
@@ -0,0 +1,157 @@
...
+
+  struct arguments_serializer_i {
+    virtual void add(base_argument& argument) = 0;
+    virtual std::string str() = 0;
+    virtual const char* data() = 0;
+    virtual size_t size() = 0;
+  };
+
+  template<typename T>
+  bool operator==(typed_argument<T>& lhs, typed_argument<T>& rhs) {
```
JohnLangford: Why isn't this a member of typed_argument<T>?

JohnLangford:
This looks like a good first step.  

Aside from the question above I would prefer to wait until after you complete the refactor before merging everything.

JohnLangford:
One thing that I'm wondering about: This patch is replacing about half of what boost::program_options does.  Given that, is it worthwhile to replace all of boost::program_options (at least the portions we use)?

vowpalwabbit/arguments.cc
jackgerrits: I think it is worth being able to check equality of two base_arguments, even though the default value can't be checked at this point

vowpalwabbit/arguments.h
jackgerrits: Good point

vowpalwabbit/arguments.h
jackgerrits: If there are several places where an argument is specified AND their specification matches then they will both receive the value of the parsed value

vowpalwabbit/arguments.h
jackgerrits: There doesn't exist a need for it to be a member function and keeping it free like this is better encapsulated. http://www.drdobbs.com/cpp/how-non-member-functions-improve-encapsu/184401197?pgno=1

jackgerrits:
Makes sense, this can be merged all on one go. I figured it would be easier to review separated like this.

What is the other half of what boost::program_options does this isn't replacing? The intention is to hide **all** of what program_options does behind an interface.

vowpalwabbit/arguments.cc
JohnLangford: I'm not questioning the usefulness of the function, but rather it's scope.

vowpalwabbit/arguments.h
JohnLangford: I'm confused between options and arguments.   The file is named arguments.h, but it seems like this is about specifying options.   Between the two, what do you have in mind here?

vowpalwabbit/arguments.cc
jackgerrits: There exists an operator for both the base_argument type as well as typed_argument<T>, so that the best operator can be used depending on how much type information you have at the time. I may be misunderstanding the question though

vowpalwabbit/arguments.h
jackgerrits: I decided to use the arguments name because of the arguments object that exists already, so that naming wasn't changed too much. A typed_argument<T> object contains all of the metadata about a specific argument that can be passed on the command line as well as its value once parsed. 

Compared to boost::program_options the typed_argument<T> type maps closely to [option_description](https://www.boost.org/doc/libs/1_63_0/doc/html/boost/program_options/option_description.html)

vowpalwabbit/arguments.h
JohnLangford: Interesting.  

It seems like there are two things to worry about: 
(1) Polluting a class with functions reduces encapsulation of the data in the class.  This is a good point. 
(2) Polluting the top level namespace with function names.   I've been bitten by this a few times and it's tricky to debug because some binary you don't expect is getting linked.  

It seems like (1) and (2) can be dealt with by having a namespace for arguments? (Or should that be options for semantic clarity?)

vowpalwabbit/arguments.h
jackgerrits: Because of the fact we pass pointers to the location to write the value upon parsing, providing a function to specify the stored value begs the question of whether or not when giving this value if it is also written to the location pointers

JohnLangford:
I think we want every merge to master to improve master and we should actively think about how to make merges to master as small as possible under that constraint.  

W.r.t. B:po, there is presently code for in this patch for specifying the semantics of options.  B:po also applies the semantics to the commandline (the second half).   This is not a question of encapsulation, but of functionality.  Given the somewhat substantial overhead of matching the interface to B:po here and the weight of B:po, maybe we should scrap the B:po interface, B:po entirely, and just write the commandline parser?

vowpalwabbit/arguments.h
JohnLangford: No?  It seems like all writes should happen at parsing time.

JohnLangford: For consistency between defaults and non-defaults.

vowpalwabbit/arguments.cc
JohnLangford: Why should both operators be at the VW:: scope?

jackgerrits:
My attempt with the typed_argument/base_argument system was to replicate the metadata that we could want about an argument/option no matter what form it may be passed in as. I didn't see this as trying to match B:po but rather trying to convey all information about the option. 

I think getting into the business of writing a command line parser is tricky, and I feel that we would come out at the end of with something that resembles what B:po is doing for us

vowpalwabbit/arguments.h
jackgerrits: I 100% agree, a namespace for all of these argument related things would be good. Perhaps `config` or `configuration` to indicate the purpose of the classes

vowpalwabbit/arguments.h
JohnLangford: So, by 'argument' you mean 'option', and yes that's my fault.  It might be worth rectifying though since it's cleaner to have 'options' associated with the semantics and 'arguments' associated with the commandline. 

Reparsing, "If there are several places where an option is specified and their specification matches, then when parsing arguments, they will both receive the parsed value".   I understand now, but I'm still a little worried that this is not quite the right structure.  What happens if an option is specified twice, once as a float and once as an int, then an int is passed on the commandline?

JohnLangford:
Sure, but it's certainly a valid observation that this specification is a big part of what B:po does.  

Commandline parsing is somewhat tricky as you say, but it's not clear to me that you'll need significantly more lines of code than the present interface to B:po.

vowpalwabbit/arguments.h
JohnLangford: Ok, let's do that.

vowpalwabbit/arguments.cc
jackgerrits: They can be in the config name space

vowpalwabbit/arguments.h
jackgerrits: Sounds good, more than happy to make that naming change.

That is not possible with this implementation, if you attempted to do that it would throw. You can only have multiple locations if everything matches

JohnLangford: There's a question here about what the right semantics for options are.  Do we want to enforce something about an option across multiple reductions or not?  In the fully modular approach, we would not want to---essentially each reduction's options would live in it's own namespace.   Why should be not be fully modular?

vowpalwabbit/arguments.h
jackgerrits: I agree with you, all writes should happen at parsing time

vowpalwabbit/arguments.h
jackgerrits: It does feel strange to enforce this. The more modular we get here though, the harder it is to decide if an option is unrecognized.

In a fully modular system what do we do if one expects a string and the other an int?

jackgerrits:
There is a rather small amount of code that actually hooks this proposed interface in with boost. We would need to handle things like conversions, implicit values, collections

vowpalwabbit/arguments.h
jackgerrits: To throw on an unrecognized option, you need to know the complete pool of possible options. If you need the complete pool then it becomes impossible to have one option be typed differently based on other options.

JohnLangford: Sorry, I meant 'throw on an unrecognized argument value'.  Even the current approach allows for parsing with unrecognized options.

jackgerrits: I see, parsing with unrecognized options is easy. The problem is then being certain about it's final nonrecognition to allow us to throw an error. unrecognized argument value means a value not convertible to the desired type?

JohnLangford: Yes?  In a library situation where each reduction is a single thing you may only feed the arguments relevant to that reduction.  

But on a commandline, and possibly for convenience in a library situation, it's necessary to have multiple pools of options with each of them parsed against the arguments as an incomplete set.   

What's important is that every argument be parsed by _some_ (at least 1) option.  

To answer my own question, it seems like we should get rid of the vector and just have more options.  All those options in a pool whose argument value are of the correct type (or coercable into the correct type as per int->float) would be parsed and applied with those arguments (and argument values) marked as parsed-at-least-once.  

At the end of parsing from the pool, the present code treats things as an anonymous specification of a data file and barfs if it fails to open.  Something slightly more intelligent may be necessary in general, but we should definitely know about and do something sane (warn or barf) on an unrecognized commandline element.

jackgerrits: I originally had that behavior essentially, so that every reduction must be very explicit and provide specifications for EVERY option it requires. Knowing when something is unrecognized is still a little tricky but I do think it is possible. I want to make sure we are on the same page here, we can chat about it tomorrow?

test/unit_test/arguments_boost_po_test.cc
```diff
@@ -0,0 +1,319 @@
...
+#include <vector>
+
+using namespace VW;
+
+
+char** convert_to_command_args(char* command_line, int& argc, int max_args = 64) {
+  char** argv = new char*[max_args];
+  argc = 0;
+
+  char* current_arg = strtok(command_line, " ");
```
rajan-chari: nit:  prefer strtok_r to strtok

JohnLangford:
This pull request **introduces 1 alert** when merging 72bc6360dbffabeb0cf1b91ce378e29564b83a95 into f085178d12c778024a0e533b5bf7a44a79f3d816 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-aeca376b3f09c8997bf1c0cf807d56794e65b117)

**new alerts:**

* 1 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

JohnLangford:
This pull request **introduces 1 alert** when merging 1918a8c1abee07cc768f05f548ea521fcec9cecf into f085178d12c778024a0e533b5bf7a44a79f3d816 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-a5fa0f3bb41bdeb0e2d27400fa244bb4072c2e8d)

**new alerts:**

* 1 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

test/unit_test/arguments_boost_po_test.cc
jackgerrits: I can't seem to be able to find the header that the _r version is defined in?

JohnLangford:
This pull request **introduces 1 alert** when merging 12426017760f2d602caf36fde0464442b6c88831 into f085178d12c778024a0e533b5bf7a44a79f3d816 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-5574e632b307df2126bad7190d471b8e5be1d40a)

**new alerts:**

* 1 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

JohnLangford:
This pull request **introduces 1 alert** when merging 12426017760f2d602caf36fde0464442b6c88831 into f085178d12c778024a0e533b5bf7a44a79f3d816 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-5574e632b307df2126bad7190d471b8e5be1d40a)

**new alerts:**

* 1 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

jackgerrits:
closing in favor of #1706

PR #1686, opened 2018-12-11
bugfix:  cb_adf is not including some examples in stats calculation

rajan-chari:
bugfix:  cb_adf is not including examples without shared features in stats calculations and output_example


JohnLangford:
This looks good---I'm going to merge it.  Thanks :-)

PR #1688, opened 2018-12-11
Move trace_message from arguments to vw object

jackgerrits:
This is required for the larger task of replacing the arguments object

### DIFF
--- a/vowpalwabbit/accumulate.cc
+++ b/vowpalwabbit/accumulate.cc
@@ -118,7 +118,7 @@ void accumulate_weighted_avg(vw& all, parameters& weights)
 {
   if(!all.adaptive)
   {
-    all.opts_n_args.trace_message<<"Weighted averaging is implemented only for adaptive gradient, use accumulate_avg instead\n";
+    all.trace_message<<"Weighted averaging is implemented only for adaptive gradient, use accumulate_avg instead\n";
     return;
   }
   uint32_t length = 1 << all.num_bits; //This is the number of parameters
--- a/vowpalwabbit/audit_regressor.cc
+++ b/vowpalwabbit/audit_regressor.cc
@@ -152,7 +152,7 @@ void end_examples(audit_regressor_data& d)
 
 inline void print_ex(vw& all, size_t ex_processed, size_t vals_found, size_t progress)
 {
-  all.opts_n_args.trace_message << std::left
+  all.trace_message << std::left
                                 << std::setw(shared_data::col_example_counter) << ex_processed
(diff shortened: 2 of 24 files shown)


JohnLangford:
I may not quite understand the plan.  

trace_message was previously in the global datastructure and then was moved to the arguments datastructure when that was created.  The core reason for this is that many of the setup() functions need to leave messages and I wanted to shift away from having setup() be dependent on the global data structure.   Shifting back, you introduce a spurious dependence on the global datastructure for many reductions.  Why?

rajan-chari:
In the short term, @jackgerrits wanted to remove all non-args logic out of args implementation.  Also probably good from a seperation of concerns perspective.

Longer term, as we remove dependence on 'all', there a number of "dependency injection" approaches we can use.

One approach is here:

https://github.com/VowpalWabbit/reinforcement_learning/blob/d61b87ad619781c9c1037b76486850a717868e8a/rlclientlib/live_model_impl.cc#L128
This allows the library user to specify what kind of logging they want (if any).  

A second option is to let a driver pass a trace interface into the setup (factory) function that constructs the reduction

A third option used in some composable systems is a service_provider class that a component can query for the services it needs (trace logging in this case).  The same service_provider would also support registering services with the provider.  This would allow a driver (or client code) to implment and register trace providers and each reduction could ask for trace provider.

JohnLangford:
I agree with the end state but I'm uncomfortable with the intermediate
state (i.e. extra dependence on the global data structure).   This is not a
big deal here, but in the future please shift towards (a) making
trace_message an argument to setup (or one of the other approaches
suggested) (b) making the overall change more of an atomic transaction.

jackgerrits:
Okay that makes sense. As we work towards the larger refactoring changes that we've discussed the changes needed become prohibitively large in terms of complexity and size. As you said though we always want to stay in a state where we are functioning correctly and not left in a bad state. @rajan-chari (chime in if you disagree with me) and I will try and find a happy medium between manageable incremental changes and continuing the evolution of the codebase. I do agree with you though, we shouldn't ever be at a point which we fundamentally disagree with approach checked in.

rajan-chari:
I was thinking about this.  The intermediate states might be less than ideal.  Reviewing smaller transformations is optimal from a workflow perspective.  Perhaps pull requests to a refactoring branch?  We can merge periodically once we are in a good intermediate state.

JohnLangford:
Sounds good.  Separate branch with incremental commits is the way I've
worked informally when doing refactoring personally.

-John

PR #1693, opened 2018-12-13
Add bug report issue template

jackgerrits:


JohnLangford:
I'm slightly hazy on how you expect to use this.  More detail?

jackgerrits:
This will be offered as a template when someone goes to open an issue. Gives a bit of structure about the info reported

JohnLangford:
Merged, thanks.

PR #1697, opened 2018-12-21
Fix some VW initialization memory leaks

homezcx:


vowpalwabbit/global_data.cc
```diff
@@ -288,6 +288,7 @@ vw::vw()
   current_pass = 0;
   reduction_stack=v_init<LEARNER::base_learner* (*)(arguments&)>();
 
+  opts_n_args.~arguments(); // in-place delete
```
JohnLangford: Why delete before use?

jackgerrits: Using placement new is one of the only times in cpp where calling the destructor explicitly is actually required as far as I know. I am not sure if the ordering of events here 100% requires it

JohnLangford: I believe we want/need to call the destructor---the code was doing that.  

But why do we want to call the destructor just before calling new?  Restated, why should we call the destructor in two place (the destructor and the constructor)?

jackgerrits: Here's the bit that I'm unclear about. Does the destructor cause the member variables that aren't heap allocated to be destructed as well? Since placement new doesn't call the destructor I am wondering if those variables need to be dealt with.

JohnLangford: At this line.  vw::vw() is constructor.

jackgerrits: I see, I am assuming the destructor i called explicitly here because we are placement newing on the next line.

JohnLangford: That's not my understanding of how placement new is used on uninitialized memory.  
(a) Calling a destructor on uninitialized memory is inherently problematic because it may do the wrong thing. 
(b) Placement new is used here for relatively spurious reasons---I wanted to a spurious pointer-chase.  More typically it's used like here: https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/v_array.h#L94 in collections. In both cases, you don't destruct (on uninitialized memory) before construct.

jackgerrits: `opts_n_args` is initialized memory though, since it is default constructed when `vw` is constructed.

JohnLangford: Good, that gets us closer.  

Calling the destructor here seems pointless rather than buggy since the default constructor does no allocation.   Is the rationale that the default constructor might in the future change to initialize some heap memory?

jackgerrits: The default constructor allocates memory for all of the member variables of the class. std::vector<T> uses heap allocation internally, and those boost objects probably do too. Calling the destructor causes the destructor for all member variables to be called too as far as I know. So there almost certainly implicit heap memory being used here

homezcx: @JohnLangford @jackgerrits The leak happened from a object that creates a VW instance by calling 

```
vw* vw = new vw();
```

In that case, ```opts_n_args``` will be allocated on heap once. Then later on, VW::initialize is called that triggers in-place constructor to allocate another ```opts_n_args``` on heap without calling delete.

homezcx: in that case , there will be an opts_n_args leaked on the heap.

homezcx: Corrected my previous comment (deleted) a bit. The leak happens when VW is created on heap by ```vw* vw = new vw()```. Since we are replacing a default constructed variable, two ```opts_n_args``` will be created and the first one is not destructed.

JohnLangford: Ok, let's merge it.  Thanks :-)

vowpalwabbit/parser_helper.h
```diff
@@ -32,9 +32,18 @@ class arguments {
   vw* all;//backdoor that should go away over time.
 
   //initialization
- arguments(vw& all_in, std::string name_in=""):new_od(name_in), missing_critical(false), all(&all_in) {file_options = new std::stringstream;};
- arguments():missing_critical(false){};//this should not be used but appears sometimes unavoidable.  Do an in-place allocation with the upper initializer after it is used.
-  ~arguments(){ delete file_options;};
+ arguments(vw& all_in, std::string name_in=""):new_od(name_in), missing_critical(false), all(&all_in) {
+   file_options = new std::stringstream;
+ };
+ arguments():missing_critical(false){
```
tyclintw: Now that the default constructor is no longer one line, please reposition the comment.

JohnLangford:
Merged, thanks :-)

PR #1698, opened 2018-12-21
Fix memory leak in cb_explore_adf

homezcx:


vowpalwabbit/cb_explore_adf.cc
```diff
@@ -581,6 +581,7 @@ void finish(cb_explore_adf& data)
     data.prepped_cs_labels[i].costs.delete_v();
   data.prepped_cs_labels.delete_v();
   data.gen_cs.pred_scores.costs.delete_v();
+  data.gen_cs.mtr_ec_seq.~vector();
```
homezcx: This object is created in cb_explore_adf.cc by calling
```
  T* temp = calloc_or_throw<T>(1);
  new (temp) T();
```

and we only called ```free``` of this object in ``` learner::finish()``` method in ```learner.h```. We have to call destructor of each fields in this method. Followed the same pattern in line 570.

@JohnLangford @ataymano Alexey and I had a discussion that we may want to introduce destructors to struct in VW instead of having ```finish()``` method explicitly calling the destructor in each of the fields. This might be a long term goal. Open to discuss.

JohnLangford: Good catch.   

A destructor would prevent oversights of this sort.  

Thinking about it though, maybe this could be done incrementally by creating a destructor and just using finish() to call the destructor?

jackgerrits:
I agree destructors would be great, the number of explicit destructors being called makes me nervous

vowpalwabbit/cb_explore_adf.cc
ataymano: yes, that was the proposal for the first step.

PR #1701, opened 2019-01-03
Add clang-format

### LINKED ISSUE #1692: Create .clang-format file to enforce style consistency across codebase

palmerlao:
This is an attempt to fix #1692.

(I think) you can reproduce most of this PR by:

- Dumping a Google-style `.clang-format` file:
  `clang-format -style=Google -dump-config > .clang-format`
- Making a few arbitrary changes to somewhat suit the project's current style better:
```
~/vowpal_wabbit$ diff .clang-format <(clang-format -style=Google -dump-config)
14,15c14,15
< AllowShortFunctionsOnASingleLine: None
< AllowShortIfStatementsOnASingleLine: false
---
> AllowShortFunctionsOnASingleLine: All
> AllowShortIfStatementsOnASingleLine: true
36c36
< BreakBeforeBraces: Allman
---
> BreakBeforeBraces: Attach
39c39
< ColumnLimit:     120
---
> ColumnLimit:     80
```
- Format everything under `vowpalwabbit/`: 
  `find vowpalwabbit -type f -name '*.cc' -o -name '*.h' | xargs clang-format -i`

It also adds onto the end of the script that the Linux CI uses to check if running `clang-format` would result in any changes.


.clang-format
```diff
@@ -0,0 +1,90 @@
...
+ObjCSpaceBeforeProtocolList: false
+PenaltyBreakBeforeFirstCallParameter: 1
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 1000000
+PenaltyReturnTypeOnItsOwnLine: 200
+PointerAlignment: Left
+ReflowComments:  true
+SortIncludes:    true
```
jackgerrits: Currently include order can break the build, we should turn this off
```suggestion
SortIncludes:    false
```

vowpalwabbit/OjaNewton.cc
```diff
@@ -533,14 +526,14 @@ base_learner* OjaNewton_setup(arguments& arg)
...
   if (arg.new_options("OjaNewton options")
-      .critical("OjaNewton", "Online Newton with Oja's Sketch")
-      ("sketch_size", ON->m, 10, "size of sketch")
-      ("epoch_size", ON->epoch_size, 1, "size of epoch")
-      ("alpha", ON->alpha, 1.f, "mutiplicative constant for indentiy")
-      ("alpha_inverse", po::value<float>(), "one over alpha, similar to learning rate")
-      ("learning_rate_cnt", ON->learning_rate_cnt, 2.f, "constant for the learning rate 1/t")
-      ("normalize", ON->normalize, true, "normalize the features or not")
-      ("random_init", ON->random_init, true, "randomize initialization of Oja or not").missing())
+          .critical("OjaNewton", "Online Newton with Oja's Sketch")("sketch_size", ON->m, 10, "size of sketch")(
```
jackgerrits: Can we change an option to not cause this change?

vowpalwabbit/action_score.h
```diff
@@ -1,79 +1,102 @@
...
-class score_iterator :
-  public virtual std::iterator<
-	  std::random_access_iterator_tag,	// iterator_cateogry
-	  float,					// value_type
-	  long,						// difference_type
-	  float*,					// pointer
-	  float						// reference
-	  >
+class score_iterator : public virtual std::iterator<std::random_access_iterator_tag,  // iterator_cateogry
+                                                    float,                            // value_type
```
jackgerrits: I do not think we want this kind of alignment, can we change the option to just indent these one level instead of align

vowpalwabbit/array_parameters_dense.h
```diff
@@ -86,44 +123,61 @@ class dense_parameters
...
+  }
 
 #ifndef _WIN32
 #ifndef DISABLE_SHARED_WEIGHTS
   void share(size_t length)
   {
-    float* shared_weights = (float*)mmap(0, (length << _stride_shift) * sizeof(float),
-      PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
+    float* shared_weights = (float*)mmap(0, (length << _stride_shift) * sizeof(float), PROT_READ | PROT_WRITE,
+                                         MAP_SHARED | MAP_ANONYMOUS, -1, 0);
```
jackgerrits: wrapped lines should not be aligned, just indented

vowpalwabbit/best_constant.cc
```diff
@@ -3,82 +3,90 @@ using namespace std;
...
+      (all.loss == nullptr) || (all.sd == nullptr))
+    return false;
 
-  float label1 = all.sd->first_observed_label; // observed labels might be inside [sd->Min_label, sd->Max_label], so can't use Min/Max
-  float label2 = (all.sd->second_observed_label == FLT_MAX)?0: all.sd->second_observed_label; // if only one label observed, second might be 0
-  if (label1 > label2) {float tmp = label1; label1 = label2; label2 = tmp;} // as don't use min/max - make sure label1 < label2
+  float label1 = all.sd->first_observed_label;  // observed labels might be inside [sd->Min_label, sd->Max_label], so
+                                                // can't use Min/Max
+  float label2 = (all.sd->second_observed_label == FLT_MAX)
+                     ? 0
```
jackgerrits: I assume the same wrapping alignment option

vowpalwabbit/feature_group.h
```diff
@@ -50,179 +57,227 @@ struct features;
...
-  features_value_iterator(feature_value* begin)
-    : _begin(begin)
-  { }
 
-  features_value_iterator(const features_value_iterator& other)
-    : _begin(other._begin)
-  { }
+ public:
+  features_value_iterator(feature_value* begin) : _begin(begin)
+  {
```
jackgerrits: I think we should explore if there is an option for this to become 
`{}` instead

.clang-format
palmerlao: Great catch, thanks!

vowpalwabbit/action_score.h
palmerlao: I think this is now partially fixed - it depends on the interpretation of one level of indentation: https://github.com/VowpalWabbit/vowpal_wabbit/pull/1701/files#diff-2f6bec9b76c178b3a45f8d064aca5350L12

palmerlao:
I made a change to the `Dockerfile` in this repo to add `clang-format`, which I hope will be reflected in `vowpalwabbit/travis-base`, but it's not clear to me that this happens automatically. Would it be better to cut two PRs for this, one to install `clang-format` in the CI image, and another to apply the formatting changes?

vowpalwabbit/OjaNewton.cc
palmerlao: I haven't come across any options that I think would be relevant. Would adding comments to turn off `clang-format` around blocks that are roughly "tabular" be acceptable?

jackgerrits: What's involved with turning it off?

.clang-format
```diff
@@ -0,0 +1,90 @@
...
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Inline
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: true
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: true
+AlwaysBreakTemplateDeclarations: true
+BinPackArguments: true
```
jackgerrits: I'd prefer this to be false
```suggestion
BinPackArguments: false
```

.clang-format
```diff
@@ -0,0 +1,90 @@
...
+BinPackParameters: true
```
jackgerrits: I'd prefer this to be false
```suggestion
BinPackParameters: false
```

vowpalwabbit/array_parameters_dense.h
palmerlao: I'm not sure it can be verbatim the same. The previous version start a new unaligned line for the third argument. Using `AlignAfterOpenBracket: DontAlign` for some reason forces it to start a new indented line with the first argument. 

The other alternatives for the option I used (`AlignAfterOpenBracket`) to get indentation instead of alignment are `Align` (the intermediate change we don't want) and `AlwaysBreak` (probably will look the same as now).

jackgerrits: It probably forces it because the line is too long, not sure. `AlignAfterOpenBracket: DontAlign` seems best?

JohnLangford:
With #1711 finally in, I'd like to get this one in.  Is it easy to rerun the reformatting to deal with the merge conflicts created?

palmerlao:
@JohnLangford I hope it is pretty easy. I can try it soon (tonight or otherwise within a few days).

palmerlao:
Apologies for the git bungling. Think this is ready for another pass

vowpalwabbit/OjaNewton.cc
```diff
@@ -537,15 +530,18 @@ base_learner* OjaNewton_setup(options_i& options, vw& all)
...
-    .add(make_option("learning_rate_cnt", ON->learning_rate_cnt).default_value(2.f).help("constant for the learning rate 1/t"))
-    .add(make_option("normalize", ON->normalize).default_value(true).help("normalize the features or not"))
-    .add(make_option("random_init", ON->random_init).default_value(true).help("randomize initialization of Oja or not"));
+  new_options.add(make_option("OjaNewton", oja_newton).keep().help("Online Newton with Oja's Sketch"))
+      .add(make_option("sketch_size", ON->m).default_value(10).help("size of sketch"))
+      .add(make_option("epoch_size", ON->epoch_size).default_value(1).help("size of epoch"))
+      .add(make_option("alpha", ON->alpha).default_value(1.f).help("mutiplicative constant for indentiy"))
+      .add(make_option("alpha_inverse", alpha_inverse).help("one over alpha, similar to learning rate"))
+      .add(make_option("learning_rate_cnt", ON->learning_rate_cnt)
+               .default_value(2.f)
```
JohnLangford: Why is .default_value shifted to new line?  Any way to avoid that?

vowpalwabbit/array_parameters.h
```diff
@@ -22,153 +22,170 @@ typedef std::unordered_map<uint64_t, weight*> weight_map;
...
-	sparse_parameters()
-		: _map(), _weight_mask(0), _stride_shift(0), _seeded(false), _delete(false), default_data(nullptr), fun(nullptr)
-	{ default_value = calloc_mergable_or_throw<weight>(stride());}
+ private:
+  void (*fun)(const weight*, void*);
 
-	bool not_null() { return (_weight_mask > 0 && !_map.empty()); }
+ public:
+  sparse_parameters(size_t length, uint32_t stride_shift = 0)
+      : _map(),
```
JohnLangford: Can this take fewer lines?

vowpalwabbit/bfgs.cc
```diff
@@ -385,27 +403,34 @@ void bfgs_iter_middle(vw& all, bfgs& b, float* mem, double* rho, double* alpha,
...
     bfgs_iter_middle(all, b, mem, rho, alpha, lastj, origin, all.weights.sparse_weights);
   else
     bfgs_iter_middle(all, b, mem, rho, alpha, lastj, origin, all.weights.dense_weights);
 }
 
-template<class T>
-double wolfe_eval(vw& all, bfgs& b, float* mem, double loss_sum, double previous_loss_sum, double step_size, double importance_weight_sum, int &origin, double& wolfe1, T& weights)
+template <class T>
+double wolfe_eval(vw& all,
+    bfgs& b,
```
JohnLangford: The many lines here seem a little bit unfortunate.   Fixable?

JohnLangford:
I took a look through.  The patch seems definitely worth doing although we might want to make some tweaks around the edges.  Can you take a look at the questions I had?

It's probably also worthwhile stick this:
 find vowpalwabbit -type f -name '*.cc' -o -name '*.h' | xargs clang-format -i 
into utl/new_version

vowpalwabbit/array_parameters.h
jackgerrits: It it doesn’t all fit on one line then it splits across multiple. I’d prefer this to be one per line (as is) rather than bin packed

JohnLangford:
@jackgerrits did you have opinions on the questions I asked as well?

vowpalwabbit/bs.cc
```diff
@@ -233,19 +234,20 @@ void finish_example(vw& all, bs& d, example& ec)
 }
 
 void finish(bs& d)
-{ delete d.pred_vec; }
+{
```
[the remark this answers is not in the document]
palmerlao: My understanding of `BreakBeforeBraces: Allman` for `clang-format` is that every brace has a newline before it. 

For class member functions, there's a separate option that I've set as per @jackgerrits earlier [comment:](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1701#discussion_r245012922) `AllowShortFunctionsOnASingleLine: Inline`.

However, I've realized that the `clang-format-3.4` I installed in CI doesn't have this option, and that the script wasn't running correctly, which I'm close to fixing.

Would there be any objection to updating the Docker image so that we can get a more accessible and up-to-date version of `clang-format`?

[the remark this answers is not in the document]
JohnLangford: no objection here.  @jackgerrits ?

vowpalwabbit/array_parameters.h
JohnLangford: is bin packed an option?  The same thing is happening to option declarations which is a hefty chunk of the extra 2k lines of code.

jackgerrits: I think the fact this is split across lines is good, much easier to read. The option declarations also are very long lines, so it is kind of nice this is splitting this up. The increase in number of lines is not a bad thing in my opinion as it improves overall readability.

vowpalwabbit/bfgs.cc
jackgerrits: This function has heaps of parameters, I think the extra lines are okay

vowpalwabbit/bs.cc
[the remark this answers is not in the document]
jackgerrits: No objection at all! I recently setup auto-building for our docker image so if you could update the Dockerfile too that would be great

jackgerrits:
I had a couple of comments, I think the overall trend of my answers though is that I'm okay with more lines if it makes things clearer. Clearer is obviously subjective though.

JohnLangford:
I guess it's clearer to me if something which is conceptually a 'line' (multiple initializers, multiple option modifiers) stays on a conceptual 'line' rather than having a discontinuous shift in format.  

This does seem pretty subjective, so I think the question is: does that convince either of you?  Or does anyone else care?

Unrelated, there's a build fail here that we need to fix before merging.

palmerlao:
My (probably strange) opinion is that the value of the code formatter does not come from making the code more readable, but from making some style (even if it's slightly or completely different from the pre-existing style) automatically and consistently applicable. 

That is, I'm okay with (and to some degree, expect) the code getting slightly uglier.

jackgerrits:
I agree with consistent and automatically applied being valuable for sure.

vowpalwabbit/array_parameters.h
jackgerrits: @palmerlao we want to have the overflow to fill the entire line before it continues onto another, is this possible?

vowpalwabbit/bs.cc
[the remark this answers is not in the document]
jackgerrits: So for the short functions, it's okay if they are expanded out over several lines but is it possible for it to be consistent with class functions and free functions?

jackgerrits:
There are just 2 more comments and then I think we should be good to merge.
- Overflow fill entire line rather than one item per line
- Expand small functions consistently

JohnLangford:
Reading through clang-format docs, it looks like we want:
AllowShortFunctionsOnASingleLine: All
instead of:
AllowShortFunctionsOnASingleLine: Inline

and:
BinPackParameters: true 
instead of:
BinPackParameters: false

The logic here is that the format is a little bit more consistent and concise.  

Do you want to do it?  (I'll merge immediately to avoid having things drag on further...)

palmerlao:
I'll make some time for this tonight unless you want to try the changes you're suggesting out.
I'm not sure this is safe to merge as-is.

JohnLangford:
What's the safety concern?

palmerlao:
I guess if you merge master and format all the files it should be okay, but merging without re-formatting may cause the CI check to fail.

I also wanted to test the CI check more thoroughly and add the script to easily reformat everything under `vowpalwabbit` under  `utl/` like you mentioned earlier, but happy to do that in a separate PR if you're keen on getting the config in now.

JohnLangford:
I see.  Go ahead and take your time to get it right.

vowpalwabbit/active.cc
```diff
@@ -144,32 +144,31 @@ base_learner* active_setup(options_i& options, vw& all)
...
   bool simulation = false;
   option_group_definition new_options("Active Learning");
-  new_options
-    .add(make_option("active", active_option).keep().help("enable active learning"))
-    .add(make_option("simulation", simulation).help("active learning simulation mode"))
-    .add(make_option("mellowness",  data->active_c0).default_value(8.f).help("active learning mellowness parameter c_0. Default 8"));
+  new_options.add(make_option("active", active_option).keep().help("enable active learning"))
+      .add(make_option("simulation", simulation).help("active learning simulation mode"))
+      .add(make_option("mellowness", data->active_c0)
+               .default_value(8.f)
```
JohnLangford: This is still not doing what I expect :-(

How do we bin pack chained function calls?

JohnLangford: As far as I can tell, this is not possible with clang-format. 

Overall, this patch still seems worthwhile, so I'm going to merge.

JohnLangford:
(Thank you :-) )

PR #1704, opened 2019-01-07
Fix best constant and best constant's loss calculation when using ksvm

jackgerrits:
When using ksvm `hinge` is forced as the loss function, but isn't recorded into the argument string. When calculating best constant later the loss function is extracted from the argument string.


JohnLangford:
Thanks, merged :-)

PR #1705, opened 2019-01-08
Update assembly versions to match current version

jackgerrits:


JohnLangford:
This will be a chronic problem.  Is there any way to make it automatic?

kumpera:
> This action should be done as part of the change when we bump the version number in the future

Tracking this issue here: https://github.com/VowpalWabbit/vowpal_wabbit/projects/5#card-16435167

JohnLangford:
Ok.  /utl/new_version is where something like an existing release script lives.

JohnLangford:
Merged, thanks both :-)

PR #1706, opened 2019-01-08
`options_i` command line parsing refactor 

jackgerrits:
- Add `options_i` interface and `options_boost_po` implementation for options parsing/storing/querying
- Add `options_serializer_i` interface and `options_serializer_boost_po` implementation for serializing an options_i collection
- Migrate 45 reductions and setup tasks to new options workflow
- Default options are no longer serialized, updated several tests to accommodate this
- Update loss function to advertise type rather than reading the command line argument


JohnLangford:
This pull request **introduces 1 alert** when merging 02045b1b2061f6f7ce9ed08c2b53f27e0b28cec2 into 1e8742b400498f24f1181a2c8dfc144ecf6adb6e - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-f977b1aa265cbc6509e810f63622398169b792c0)

**new alerts:**

* 1 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

JohnLangford:
This pull request **introduces 1 alert** when merging b52586e0ab23e680e05fe590df7e40df32fa7fed into 1e8742b400498f24f1181a2c8dfc144ecf6adb6e - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-f654b9d0287f02309c3b0e90f61aede4beca8c95)

**new alerts:**

* 1 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

JohnLangford:
This pull request **introduces 1 alert** when merging 7f006c1ec54f32ae17d3e52e0a3f4a99bfe36d86 into 1e8742b400498f24f1181a2c8dfc144ecf6adb6e - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-e3a0d97eae885d316f20bdd88d1acb6b521fe0a7)

**new alerts:**

* 1 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

arielf:
Hi Jack,

Thanks for all the work.

Being able to pass a data-file without `-d` is a very useful feature (I added it originally many years ago).  I never really liked the need for the redundant `-d`

1) It is the most common minimal use case (`vw [options] trainset-file`) 
2) It is consistent with many other Unix utilities behavior (`-d` is in fact the less natural way to pass input)
3) Explicit files behave the same way as providing input via `stdin` which is a very common use-pattern too.
4) The change creates inconsistencies with common practice and many examples in emails and on the web that were shared over the past ~10 years.

So I just wanted to share that I feel the breaking change is a step backwards, not just in terms of user-friendliness, but much more than that.

Could we revert this particular change?
Thanks :)

JohnLangford:
This pull request **introduces 1 alert** when merging bf8b3f183df6d1d72b5026f41f5e6eee7dabf12b into 1e8742b400498f24f1181a2c8dfc144ecf6adb6e - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-3d59aacec7955a753810f431362adc6fdb00c09f)

**new alerts:**

* 1 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

jackgerrits:
Thanks so much for commenting @arielf, I'm glad you saw this. You make some excellent points, and I agree with you. I'm going to put back in the positional behavior of `-d`.

JohnLangford:
This pull request **introduces 1 alert** when merging 3f7cf1afff9ac0148c1b73d6e9f8c2b8503c6d08 into 1e8742b400498f24f1181a2c8dfc144ecf6adb6e - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-2538492870b185b2efe045d039b4f06555dc03e8)

**new alerts:**

* 1 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

JohnLangford:
This pull request **introduces 2 alerts** when merging d3bf228a5dcab6c903512b8c718d044bc1be6bbc into 1e8742b400498f24f1181a2c8dfc144ecf6adb6e - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-c91debda754699018dd3612dd81b5b4dbc9c5a51)

**new alerts:**

* 2 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

JohnLangford:
This pull request **introduces 2 alerts** when merging 78c7f2aa1b8b340d845a50998b19eec72ee7db22 into 1e8742b400498f24f1181a2c8dfc144ecf6adb6e - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-edeb8e68aaac43c435147f8f9890e9291d12b157)

**new alerts:**

* 2 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

vowpalwabbit/options.h
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    template<typename T>
+    option_group_definition& operator()(typed_option<T> op) {
+      add(op);
+      return *this;
+    }
+
+    std::string m_name;
+    std::vector<std::shared_ptr<base_option>> m_options;
```
rajan-chari: >shared_ptr [](start = 21, length = 10)

why vector<shared_ptr> and not just vector<option>?

jackgerrits: Options may be retrieved from the class and shouldn't be copied, but their lifetime needs to ensured. It may be possible to use references instead but I didn't go that far

rajan-chari: 1) what's the reason copies are not allowed
2) why not prevent copies using a private copy constructor?
3) since the vector is exposed, why not let the vector handle returning the reference?

---
In reply to: [246873582](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1706#discussion_r246873582) [](ancestors = 246873582)

rajan-chari: options_i returns a shared_ptr instead of a reference.  Is that the issue?


---
In reply to: [246889589](https://github.com/VowpalWabbit/vowpal_wabbit/pull/1706#discussion_r246889589) [](ancestors = 246889589,246873582)

cs/cli/vowpalwabbit.cpp
```diff
@@ -764,8 +764,9 @@ size_t hashstring(String^ s, size_t u)
 Func<String^, size_t, size_t>^ VowpalWabbit::GetHasher()
 { //feature manipulation
   string hash_function("strings");
-  if (m_vw->opts_n_args.vm.count("hash"))
-  { hash_function = m_vw->opts_n_args.vm["hash"].as<string>();
+  if (m_vw->options->was_supplied("hash"))
+  {
+    hash_function = m_vw->options->get_typed_option<string>("hash").value();
```
JohnLangford: Have you considered removing the extra ->?

cs/unittest/TestMarshalNumeric.cs
```diff
@@ -1,4 +1,5 @@
 
+
```
JohnLangford: Why the pointless whitespace change?

python/pylibvw.cc
```diff
@@ -70,14 +71,16 @@ const char* get_model_id(vw_ptr all) { return all->id.c_str(); }
...
-	{
-		args.append(s);
-		args.append(" ");
-	}
+  VW::config::options_serializer_boost_po serializer;
+  for (auto const& option : all->options->get_all_options())
+  {
+    if (all->options->was_supplied(option->m_name))
+    {
+      serializer.add(*option);
```
JohnLangford: Minor: I'd prefer to avoid {} for single operation blocks.

test/pred-sets/ref/inv_hash.cmp
```diff
@@ -5,8 +5,8 @@ bits:18
 lda:0
 0 ngram:
 0 skip:
-options: --hash_seed 0 --quadratic :: --oaa 3 --link identity
-Checksum: 2231945539
+options: --oaa 3 --quadratic ::
```
JohnLangford: I'm assuming you've made defaults implicit here.

vowpalwabbit/OjaNewton.cc
```diff
@@ -529,24 +529,30 @@ void save_load(OjaNewton& ON, io_buf& model_file, bool read, bool text)
...
-      ("alpha", ON->alpha, 1.f, "mutiplicative constant for indentiy")
-      ("alpha_inverse", po::value<float>(), "one over alpha, similar to learning rate")
-      ("learning_rate_cnt", ON->learning_rate_cnt, 2.f, "constant for the learning rate 1/t")
-      ("normalize", ON->normalize, true, "normalize the features or not")
-      ("random_init", ON->random_init, true, "randomize initialization of Oja or not").missing())
+
+  bool oja_newton;
+  float alpha_inverse;
+  VW::config::option_group_definition new_options("OjaNewton options");
+  new_options.add(VW::config::make_typed_option("OjaNewton", oja_newton).keep().help("Online Newton with Oja's Sketch"));
```
JohnLangford: This is adding quite a bit of verbage compared to what was there before.  

Why can't we use an operator() similar to previous?  Or at least .add() if you want to explicit?

And, the repeated use of VW::config:: is kind of unfortunate.

vowpalwabbit/allreduce.h
```diff
@@ -3,10 +3,13 @@ Copyright (c) by respective owners including Yahoo!, Microsoft, and
 individual contributors. All rights reserved.  Released under a BSD
 license as described in the file LICENSE.
  */
-// This implements the allreduce function of MPI.
 
+// This implements the allreduce function of MPI.
 #pragma once
+
 #include <string>
+#include <algorithm>
```
JohnLangford: Why is this change here?

vowpalwabbit/autolink.cc
```diff
@@ -34,17 +34,20 @@ void predict_or_learn(autolink& b, LEARNER::single_learner& base, example& ec)
...
+
+  if (!options.was_supplied("autolink"))
     return nullptr;
 
-  data->stride_shift = arg.all->weights.stride_shift();
+  data->stride_shift = all.weights.stride_shift();
 
   LEARNER::learner<autolink,example>& ret =
-    init_learner(data, as_singleline(setup_base(arg)), predict_or_learn<true>, predict_or_learn<false>);
+    init_learner(data, as_singleline(setup_base(*all.options, all)), predict_or_learn<true>, predict_or_learn<false>);
```
JohnLangford: It's somewhat inelegant to pass an argument and a field of the same argument.

vowpalwabbit/global_data.h
```diff
@@ -526,13 +532,14 @@ struct vw
...
   uint64_t affix_features[256]; // affixes to generate (up to 16 per namespace - 4 bits per affix)
-  bool     spelling_features[256]; // generate spelling features for which namespace
+  bool spelling_features[256]; // generate spelling features for which namespace
   std::vector<std::string> dictionary_path;  // where to look for dictionaries
   std::vector<feature_dict*> namespace_dictionaries[256]; // each namespace has a list of dictionaries attached to it
   std::vector<dictionary_info> loaded_dictionaries; // which dictionaries have we loaded from a file to memory?
 
   void(*delete_prediction)(void*); bool audit; //should I print lots of debugging information?
   bool quiet;//Should I suppress progress-printing of updates?
+  bool help_requested;
```
JohnLangford: Semantics?  Why do we need to add this?  (We generally want to remove fields here.)

vowpalwabbit/global_data.h
```diff
@@ -552,6 +559,7 @@ struct vw
   uint32_t holdout_period;
   uint32_t holdout_after;
   size_t check_holdout_every_n_passes;  // default: 1, but search might want to set it higher if you spend multiple passes learning a single policy
+  size_t early_terminate_passes;
```
JohnLangford: Semantics?

vowpalwabbit/kernel_svm.cc
```diff
@@ -855,47 +856,58 @@ void finish(svm_params& params)
...
-      ("pool_size", params->pool_size, (size_t)1, "size of pools for active learning")
-      ("subsample", params->subsample, (size_t)1, "number of items to subsample from the pool")
-      .keep("kernel", kernel_type, (string)"linear", "type of kernel (rbf or linear (default))")
-      .keep("bandwidth", bandwidth, 1.f, "bandwidth of rbf kernel")
-      .keep("degree", degree, 2, "degree of poly kernel").missing())
-      //.keep("lambda", params->lambda, "saving regularization for test time").missing())
+
+  bool ksvm = false;
+
+  option_group_definition kernel_svn_options("Kernel SVM");
```
JohnLangford: misspelling: svm not svn.

vowpalwabbit/options.cc
```diff
@@ -0,0 +1,15 @@
+#include "options.h"
```
JohnLangford: Is this worth a .cc file?

vowpalwabbit/options.h
```diff
@@ -0,0 +1,159 @@
...
+    virtual ~base_option() {}
+  };
+
+  template<typename T>
+  struct typed_option : base_option {
+    typed_option(std::string name, T& location)
+      : base_option(name, typeid(T).hash_code()), m_location{ location } {}
+
+    static size_t type_hash() {
+      return typeid(T).hash_code();
```
JohnLangford: This btw, is a cool feature that I wasn't aware of.

test/pred-sets/ref/inv_hash.cmp
jackgerrits: Yes, when default values are used they are no longer serialized into the model

vowpalwabbit/OjaNewton.cc
jackgerrits: using the namespace reduces noise here by quite a bit, I'd be happy to do that to all of them. Using operator() is possible here, I just avoided it because clang-format has some trouble formatting it nicely and really just personal style preferences.

What do you mean by?

> Or at least .add() if you want to explicit?

test/pred-sets/ref/inv_hash.cmp
jackgerrits: Yes, when default values are used they are no longer serialized into the model

vowpalwabbit/autolink.cc
jackgerrits: I agree it is inelegant, but I think it is preferable to just the all object no? It is more explicit about parameters. It is unfortunate that the option object must be a field of all still

vowpalwabbit/global_data.h
jackgerrits: In order to print help, all possible options must have been seen already. We only see the options available after `parse_modules` so we must save the intent as we parse the option and then carry out the action later.

vowpalwabbit/global_data.h
jackgerrits: This is an option several reductions care about, previously they read it out of the variable map. I moved to exposing the value off all for consumption.

vowpalwabbit/options.cc
jackgerrits: Perhaps not? They'd need to be marked inline in the header

jackgerrits:
Remaining test failures should be fixed by #1707

cs/cli/vowpalwabbit.cpp
jackgerrits: Which `->` are you referring to?

cs/unittest/TestMarshalNumeric.cs
jackgerrits: I'll remove this file since it is unrelated to the change

JohnLangford:
This pull request **introduces 2 alerts** when merging cb0b7c00f2f40e83023d4671cb8c1c1cc4f17d6c into f40ef64c9469f3bdb4cc78ac6bbf4120a23e614e - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-07dd902cf6e5eca9752cb8188ed686544219efbd)

**new alerts:**

* 2 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

jackgerrits:
It is very long, thanks for going through it. I'll address all of the comments.

Duplicate detection is done through registering all options as vectors on the implementation side so that boost::po saves all values. Then as we process the final list after parsing we can check if they disagree. It is linear over the number of supplied values for an argument, but is constant in regards to total number of different supplied arguments where it was n before. See here:
https://github.com/VowpalWabbit/vowpal_wabbit/pull/1706/files#diff-a567aac3b00bba949b9da1685197e413R188

JohnLangford:
This pull request **introduces 2 alerts** when merging 1aaafe7ca4227b23720c25784f939b83434d6b1f into f40ef64c9469f3bdb4cc78ac6bbf4120a23e614e - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-a04730529b203dde0cfd5fca410ceaa8f1fb4fae)

**new alerts:**

* 2 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

vowpalwabbit/options.h
jackgerrits: They are created as part of the option_group_definition, then passed in and ownership is taken over by the options class.

It's a tricky set of lifetimes. But I may be able to make it so that only rvalues are accepted when adding to the option_group_definition, and then the values stored in the option_group_definition are processed they are then moved to the options object.

As you can see the typed_option object gets moved around a fair bit, so I opted for using a pointer to help ensure it isn't copied and one value is kept alive.

JohnLangford:
That's elegant, nice. 

Maybe also look into the two LGTM errors?

vowpalwabbit/allreduce.h
jackgerrits: accidental change, will revert

vowpalwabbit/example.cc
```diff
@@ -4,6 +4,8 @@ individual contributors. All rights reserved.  Released under a BSD (revised)
 license as described in the file LICENSE.
  */
 #include <stdint.h>
+#include <algorithm>
```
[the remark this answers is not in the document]
jackgerrits: std::find requires it

vowpalwabbit/gd.cc
```diff
@@ -1070,96 +1076,103 @@ uint64_t ceil_log_2(uint64_t v)
...
+  if(options.was_supplied("feature_mask"))
     feature_mask_off = false;
 
-  if(!arg.all->holdout_set_off)
+  if(!all.holdout_set_off)
   {
-    arg.all->sd->holdout_best_loss = FLT_MAX;
-    g->early_stop_thres = arg.vm["early_terminate"].as< size_t>();
+    all.sd->holdout_best_loss = FLT_MAX;
+    // TODO save this into all
```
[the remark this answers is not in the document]
jackgerrits: removed

vowpalwabbit/options.cc
jackgerrits: I'm going to leave as is

vowpalwabbit/allreduce.h
jackgerrits: Actually this was needed for std::min!

JohnLangford:
This pull request **fixes 2 alerts** when merging 0bea867dc77a86110380c855dd0c70a5ee09b943 into 2b64c3c0938d1f03aa8e6c9ca1513174a14240e2 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-a948786f6d6334a0e5e16aa2f9c74ab8eefc5e57)

**fixed alerts:**

* 2 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

JohnLangford:
This pull request **fixes 2 alerts** when merging 4dec969eef3d28b03cc498aee8f374b621c7fcab into 2b64c3c0938d1f03aa8e6c9ca1513174a14240e2 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-a9943304b75367a943dea2b6ae987e2373707126)

**fixed alerts:**

* 2 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

JohnLangford:
This pull request **fixes 2 alerts** when merging 9b5215ca89c0e522305b674a5764bd793992a1af into 2b64c3c0938d1f03aa8e6c9ca1513174a14240e2 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-1e8fee7f8308afe4a6c3e60b9a0d0c80d9ba1997)

**fixed alerts:**

* 2 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

cs/cli/vowpalwabbit.cpp
JohnLangford: Ping on this.

jackgerrits: There is actually a situation where we need a pointer rather than a reference here. If a command line is supplied we must construct the options_i object for the user, and so we must delete when the vw object is deleted. We could could cast our reference to a pointer in this case but I think we should avoid that. Most usages of options_i in the system is through references

vowpalwabbit/autolink.cc
jackgerrits: I'll make the change

vowpalwabbit/global_data.h
jackgerrits: I am planning on tackling shared state soon, so I'll leave it as is because I would like to avoid too many duplicate option definitions. Best case would be this is passed as a parameter.

vowpalwabbit/parse_args.h
```diff
@@ -5,13 +5,30 @@ license as described in the file LICENSE.
  */
 #pragma once
 #include "global_data.h"
+#include "options.h"
+
+// Used in parse_source
+struct input_options {
+  bool daemon;
```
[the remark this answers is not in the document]
jackgerrits: This is an attempt to encapsulate related state, because it is an option parsed as part of this block I am going to keep it in the struct

vowpalwabbit/parse_dispatch_loop.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <functional>
+
```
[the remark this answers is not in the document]
jackgerrits: Required for std::function on line 5

JohnLangford:
This pull request **fixes 2 alerts** when merging dce63d8e29e43ee296bc7a6829006e454364a5f2 into 2b64c3c0938d1f03aa8e6c9ca1513174a14240e2 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-3ae18b10ee6d2170baf0e0c277b3fba6e420d206)

**fixed alerts:**

* 2 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

JohnLangford:
This pull request **fixes 2 alerts** when merging fa7f6d3207e7eb33834c7819a0842531c2b9bc86 into 2b64c3c0938d1f03aa8e6c9ca1513174a14240e2 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-e56e8bac04315c88866e04c003c3201b5b36eb81)

**fixed alerts:**

* 2 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

cs/cli/vowpalwabbit.cpp
JohnLangford: I'm not quite following this.   Deleting of members seems routine?

JohnLangford:
This pull request **fixes 2 alerts** when merging 9337bed4e587598d29dda6ef1e7dc9b24e61e90b into 6dbb7e5f96e701253f16397b19ecde70881a7057 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-602e27a8b66aa829c7080f688c27ce5528544edd)

**fixed alerts:**

* 2 for Inconsistent definition of copy constructor and assignment \(&#39;Rule of Two&#39;\)

---

JohnLangford:
I think the pointer vs. member question is the only thing left to worry about here.

JohnLangford:
Merged, thanks :-)  That's quite a monster.

arielf:
Hi there.  Indeed a big change. Thanks for all the effort!

I've been testing this a bit and I'm seeing another backward incompatibility break with `--quiet`:

```
$ ./vowpalwabbit/vw --quiet f20-315.tt.gz

finished run
number of examples = 0
weighted label sum = 0.000000
average loss = n.a.
total feature number = 0
the argument ('f20-315.tt.gz') for option '--quiet' is invalid. Valid choices are 'on|off', 'yes|no', '1|0' and 'true|false'
```

Would it be possible to make `--quiet` backward-compatible in the sense that it passed naked, it defaults to `=yes` and doesn't cause an abort/error ?

Thanks!

jackgerrits:
This isn’t restricted to quiet, but any Boolean option as we are expecting a following value. We can force Boolean options to be a switch (rather than an option that takes a value), but then it is not possible to set an option to be false. This is one of the tricky cases using a positional data option introduces. Currently Boolean options take an implicit value of true if the following token is an option, but a following value token is interpreted as a value for the Boolean option. The positional data value trips this up.

JohnLangford:
Boost manages to get this to work.  Why can't we?

PR #1707, opened 2019-01-11
Add tolerances for float comparison in VWActionScoresLearnerTest tests

jackgerrits:
In these tests the ActionScores are compared using direct equality, this change introduces a tolerance to the float comparisons to deal with rounding precision errors. This resolves test issues seen in #1706 

The tolerance is taken from the minor allowable tolerance in RunTests.

### DIFF
--- a/java/src/test/java/vowpalWabbit/learner/VWActionScoresLearnerTest.java
+++ b/java/src/test/java/vowpalWabbit/learner/VWActionScoresLearnerTest.java
@@ -4,16 +4,65 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 import vowpalWabbit.VWTestHelper;
+import vowpalWabbit.responses.ActionScore;
 import vowpalWabbit.responses.ActionScores;
 
 import java.io.IOException;
 
 import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertTrue;
 
 /**
  * Created by jmorra on 10/2/15.
  */
 public class VWActionScoresLearnerTest extends VWTestHelper {
+    private final float TOLERANCE = 0.001f;
+
+    // https://floating-point-gui.de/errors/comparison/
+    private static boolean equalWithTolerance(float a, float b, float epsilon) {
+        final float absA = Math.abs(a);
+        final float absB = Math.abs(b);
+        final float diff = Math.abs(a - b);
+
+        if (a == b) { // shortcut, handles infinities
+            return true;
+        } else if (a == 0 || b == 0 || diff < Float.MIN_NORMAL) {
+            // a or b is zero or both are extremely close to it
+            // relative error is less meaningful here
+            return diff < (epsilon * Float.MIN_NORMAL);
+        } else { // use relative error
+            return diff / Math.min((absA + absB), Float.MAX_VALUE) < epsilon;
+        }
+    }
+
+    private boolean equalWithTolerance(ActionScore a1, ActionScore a2, float tolerance) {
+        return (a1.getAction() == a2.getAction())
+                && equalWithTolerance(a1.getScore(), a2.getScore(), tolerance);
+    }
+
+    private boolean equalWithTolerance(ActionScores a1, ActionScores a2, float tolerance) {
+        ActionScore[] a1_scores = a1.getActionScores();
+        ActionScore[] a2_scores = a2.getActionScores();
+
+        for (int i = 0; i < a1_scores.length; i++) {
+           if(!equalWithTolerance(a1_scores[i], a2_scores[i], tolerance)){
+               return false;
+           }
+        }
+
+        return true;
+    }
+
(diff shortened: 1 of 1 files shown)


jackgerrits:
Comparing floats exactly is rarely reliable due to rounding errors while being operated on, this just makes the test resilient to this.

JohnLangford:
Comparing float values which are equal as reals but computed via a different sequence of operations is not reliable due to rounding.   Where is this happening?

The worry here: my experience is that float values in tests which are not _precisely_ equal end up creating cascading differences of arbitrary scale due to tie-breaking situations.  So, I don't want to paper over an issue that should be nailed down now.

java/src/test/java/vowpalWabbit/learner/VWActionScoresLearnerTest.java
```diff
@@ -4,16 +4,65 @@
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 import vowpalWabbit.VWTestHelper;
+import vowpalWabbit.responses.ActionScore;
 import vowpalWabbit.responses.ActionScores;
 
 import java.io.IOException;
 
 import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertTrue;
```
jmorra:
```suggestion
import static org.junit.Assert.assertEquals;
```

java/src/test/java/vowpalWabbit/learner/VWActionScoresLearnerTest.java
```diff
@@ -4,16 +4,65 @@
...
 
 /**
  * Created by jmorra on 10/2/15.
  */
 public class VWActionScoresLearnerTest extends VWTestHelper {
+    private final float TOLERANCE = 0.001f;
```
jmorra: None of this is necessary as it is already in `Assert`.

java/src/test/java/vowpalWabbit/learner/VWActionScoresLearnerTest.java
```diff
@@ -61,7 +110,7 @@ public void testCSOAA() throws IOException {
                         actionScore(1, 1.0227613f)
                 )
         };
-        assertArrayEquals(expected, pred);
+        assertTrue(equalWithTolerance(expected, pred, TOLERANCE));
```
jmorra:
```suggestion
        assertEquals(equalWithTolerance(expected, pred, TOLERANCE));
```

java/src/test/java/vowpalWabbit/learner/VWActionScoresLearnerTest.java
```diff
@@ -110,7 +159,7 @@ private void testCBADF(boolean withRank) throws IOException {
             )
         };
         vw.close();
-        assertArrayEquals(expectedTrainPreds, trainPreds);
+        assertTrue(equalWithTolerance(expectedTrainPreds, trainPreds, TOLERANCE));
```
jmorra:
```suggestion
        assertEquals(equalWithTolerance(expectedTrainPreds, trainPreds, TOLERANCE));
```

java/src/test/java/vowpalWabbit/learner/VWActionScoresLearnerTest.java
```diff
@@ -123,6 +172,6 @@ private void testCBADF(boolean withRank) throws IOException {
         };
 
         vw.close();
-        assertArrayEquals(expectedTestPreds, testPreds);
+        assertTrue(equalWithTolerance(expectedTestPreds, testPreds, TOLERANCE));
```
jmorra:
```suggestion
        assertEquals(equalWithTolerance(expectedTestPreds, testPreds, TOLERANCE));
```

jackgerrits:
@jmorra the assertEquals(Object, Object) overload is being used as it is not directly comparing a float. That's why I had to actually iterate over the objects to compare the values.

java/src/test/java/vowpalWabbit/learner/VWActionScoresLearnerTest.java
jackgerrits: See comment below about using the `Object` overload meaning tolerance cannot be used

jackgerrits:
@JohnLangford I have tried to reproduce this difference in values for quite some time but cannot nail down what is causing it. As far as I can tell the prediction values are identical in the cpp layer. I created an app doing the equivalent operations in cpp only and the values for the prediction before and after the options change are identical (on Windows, so it may be a Linux only difference). The only change I made to the Java binding layer was an exception being translated at the binding layer.

jmorra:
It would appear that a newer version of JUnit might have this functionality built in.

https://stackoverflow.com/questions/37627340/junit-assert-double-arrays

You would have to just pull out the values and you can do that generically.

jackgerrits:
Ah I see, so flatten and compare? I'll update

jackgerrits:
Closing in favor of #1709, this was actually a bug with the implementation of `exp` that was being linked. It seems as if we are okay not using tolerances in these tests for now.

PR #1708, opened 2019-01-11
Disable stdin processing for all but single instance CLI. Fixes #1300

### LINKED ISSUE #1300: JNI Layer throws Exceptions when close method is called in parallel ON DIFFERENT MODELS

kumpera:
Fixes #1300

### DIFF
--- a/java/src/main/java/vowpalWabbit/learner/VWLearners.java
+++ b/java/src/main/java/vowpalWabbit/learner/VWLearners.java
@@ -36,6 +36,8 @@ final public class VWLearners {
      */
     @SuppressWarnings("unchecked")
     public static <T extends VWLearner> T create(final String command) {
+        if(command.indexOf("--no_stdin") == -1)
+            command += " --no_stdin";
         long nativePointer = initialize(command);
         VWReturnType returnType = getReturnType(nativePointer);
 
--- a/java/src/main/java/vowpalWabbit/learner/VWLearners.java
+++ b/java/src/main/java/vowpalWabbit/learner/VWLearners.java
@@ -35,7 +35,7 @@ final public class VWLearners {
      * @return A VW Learner
      */
     @SuppressWarnings("unchecked")
-    public static <T extends VWLearner> T create(final String command) {
+    public static <T extends VWLearner> T create(String command) {
         if(command.indexOf("--no_stdin") == -1)
             command += " --no_stdin";
         long nativePointer = initialize(command);
--- a/cs/cli/vw_base.cpp
+++ b/cs/cli/vw_base.cpp
@@ -1,4 +1,4 @@
-﻿/*
+/*
 Copyright (c) by respective owners including Yahoo!, Microsoft, and
 individual contributors. All rights reserved.  Released under a BSD (revised)
 license as described in the file LICENSE.
@@ -60,7 +60,9 @@ VowpalWabbitBase::VowpalWabbitBase(VowpalWabbitSettings^ settings)
         m_model->IncrementReference();
       }
       else
-      { if (settings->ModelStream == nullptr)
+      { if (!settings->Arguments->Contains("--no_stdin"))
+		  string += " --no_stdin";
+	    if (settings->ModelStream == nullptr)
         { if (!settings->Verbose && !settings->Arguments->Contains("--quiet"))
             string.append(" --quiet");
 
@@ -68,8 +70,6 @@ VowpalWabbitBase::VowpalWabbitBase(VowpalWabbitSettings^ settings)
         }
         else
         { clr_io_buf model(settings->ModelStream);
-          if (!settings->Arguments->Contains("--no_stdin"))
-            string += " --no_stdin";
           m_vw = VW::initialize(string, &model, false, trace_listener, trace_context);
           delete settings->ModelStream;
 		  settings->ModelStream = nullptr;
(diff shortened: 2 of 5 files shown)


JohnLangford:
Do you know what the travis fail is?

kumpera:
I'll take a look at it asap.

java/src/main/java/vowpalWabbit/learner/VWLearners.java
```diff
@@ -36,6 +36,8 @@ private VWLearners() {}
      */
     @SuppressWarnings("unchecked")
     public static <T extends VWLearner> T create(final String command) {
+        if(command.indexOf("--no_stdin") == -1)
+            command += " --no_stdin";
```
jackgerrits: Travis build failure: `/vw/java/src/main/java/vowpalWabbit/learner/VWLearners.java:[40,13] final parameter command may not be assigned`

`command` is final so can't be changed, can just create a new string to append to to fix.

JohnLangford:
Status?

JohnLangford:
Status on this?

kumpera:
I'm swamped with other higher priority tasks. I plan to have time for this before EOM.

JohnLangford:
Sounds good.

kumpera:
@jackgerrits @JohnLangford I fixed the Java issue. Sorry for the delay

JohnLangford:
Great to have this one in, thanks :-)

PR #1709, opened 2019-01-14
Use std::exp instead of exp free function

jackgerrits:
This should resolve the rounding issues in #1706

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,32 @@
+---
+name: Bug report
+about: Report a bug in VW
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+#### Describe the bug
+A clear and concise description of what the bug is.
+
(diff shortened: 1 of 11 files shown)


kumpera:
@JohnLangford is VW's numerical stability across OS/releases important?
If it is, we should consider whether to build and static link a custom libm.

JohnLangford:
Yes.   The essential issue is that you need tests to verify that things are working and if the tests change across platforms it becomes a maintenance problem.

A custom libm seems like overkill?  Compared to little macros like this one.

kumpera:
By custom I mean have a submodule to an existing implementation.
It's can be a problem in the sense that even basic functions like `exp` are implemented in different ways across OS and that could produce precision issues.

JohnLangford:
I guess, the question is: are you thinking of proactively creating a library?  Or reactively?   If the former, there's an issue in that there are many functions we don't use.  If reactively, that informally exists---there are many little bits of code throughout that have a #ifdef depending on the OS.

PR #1713, opened 2019-01-17
Compile fixes

JohnLangford:
Fix the last compile warnings.


vowpalwabbit/parse_args.cc
```diff
@@ -1053,7 +1053,7 @@ void parse_output_preds(options_i& options, vw& all)
       if (options.was_supplied("binary"))
         all.trace_message << "Warning: --raw_predictions has no defined value when --binary specified, expect no output" << endl;
     }
-    if (raw_predictions.c_str() == "stdout")
+    if (strcmp(raw_predictions.c_str(),"stdout")==0)
```
jackgerrits: Why not just compare the string object to the temporary?

raw_predictions == “stdout”

JohnLangford: That sounds plausibly better.

JohnLangford: (Send a pull request?)

jackgerrits:
Will do

PR #1715, opened 2019-01-17
Change from strcmp to std::string operator== for comparison

jackgerrits:


JohnLangford:
Definitely cleaner, thanks :-)

PR #1717, opened 2019-01-18
Change to using a bool_switch for bool options

jackgerrits:
This should fix the issue @arielf brought up in #1706. The following should not work as expected:
```
$ ./vowpalwabbit/vw --quiet f20-315.tt.gz
```

This does mean that a value of `false` or any can no longer be supplied to boolean options, except for the 2 special cases in OjaNewton


JohnLangford:
This looks good to me.  

"Every boolean is a switch" represents some loss in expressiveness from "some are a switch and some take an argument", but that seems relatively minor which this patch seems to show.

PR #1721, opened 2019-01-22
Create Windows build scripts and update instructions

lokitoth:
* Creates restore and build scripts with support for custom toolchain integration
* Updates instructions for building on Windows (includes instructions for VS2017)
* Changes AppVeyor to use new build scripts


README.Windows.md
```diff
@@ -0,0 +1,41 @@
+﻿## Windows specific info
+
+This document details differences in pre-requisites and building instructions under Windows.
```
jackgerrits: Can this in be added in the wiki rather than here? https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Dependencies

lokitoth: Sure; I was just trying to be parallel to README.md. Keep in both, like for Linux?

lokitoth:
Closes #1721

README.Windows.md
jackgerrits: So the readme is a bit out of date at the moment, I've done work to try and make sure the wiki is up to date. I've utilized several pages in the wiki to make it a bit simpler. I was planning on making the front page readme a bit simpler and just using it to explain VW's mission and why use it, then link into the wiki for instructions

JohnLangford:
Having the information in a single location is conceptually simple, but
maybe the other direction is better?  That way it's available when you
download the code and need to operate offline?

-John

jackgerrits:
Totally agree, plus there's the added benefit of pull requests for changes when they are in files. So we can either migrate the existing wiki docs to a folder in the vowpalwabbit repo, or we can utilize the docs repo I recently created to host the rl_client docs. (See here https://vowpalwabbit.github.io/docs/, but this removes the benefit of automatically having it available locally for offline use) Or we can wait until the website is ready and think about how the pages fit in there? The wiki is useful for the sidebar aswell

PR #1730, opened 2019-01-29
Fix warnings in Windows MSVC x64 build

lokitoth:
* Fixes bool inequality comparison warning
* Annotates rapidjson.h as unmanaged when compiling with /clr under MSVC

Note: This does not address a warning about code analysis ruleset not
   found in some cases, as well as an issue with GitLink and the T4 tool

### DIFF
--- a/vowpalwabbit/parse_example_json.h
+++ b/vowpalwabbit/parse_example_json.h
@@ -13,8 +13,18 @@ license as described in the file LICENSE.
 //#define RAPIDJSON_SIMD
 //#define RAPIDJSON_SSE42
 
+// Let MSVC know that it should not even try to compile this as managed
+#if (_MANAGED == 1) || (_M_CEE == 1)
+#pragma managed(push,off)
+#endif
+
 #include <rapidjson/reader.h>
 #include <rapidjson/error/en.h>
+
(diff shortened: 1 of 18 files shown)


JohnLangford:
This looks good to me although I'm unfamiliar with the pragma.

lokitoth:
I added comments referencing the documentation for the pragma as well as the detection logic.

vowpalwabbit/parse_example_json.h
```diff
@@ -13,8 +13,20 @@ license as described in the file LICENSE.
 //#define RAPIDJSON_SIMD
 //#define RAPIDJSON_SSE42
 
+// Let MSVC know that it should not even try to compile RapidJSON as managed
+// - pragma documentation: https://docs.microsoft.com/en-us/cpp/preprocessor/managed-unmanaged?view=vs-2017
+// - /clr compilation detection: https://docs.microsoft.com/en-us/cpp/dotnet/how-to-detect-clr-compilation?view=vs-2017
+#if (_MANAGED == 1) || (_M_CEE == 1)
```
jackgerrits: This bit is nice, otherwise it would warn about an unknown pragma under gcc

JohnLangford:
Merged, thanks :-)

PR #1731, opened 2019-01-29
Move to std types for concurrency

### LINKED ISSUE #1725: Improve portability of VW lib - MUTEX and CV
We should replace `MUTEX` and `CV` with the standard library types `std::mutex` and `std::condition_variable` to reduce dependence on SDKs and system headers.

Beyond this, daemon mode should be extracted into it's on project/target so it can be only compiled when needed as this introduces system dependencies for the network calls.

We are actually quite close to vw_core being portable across OSs.

jackgerrits:
Closes #1725 

- Moves to `std::thread`, `std::condition_variable` and `std::mutex`
- Uses scope based locking for better exception safety
- Move from calloc to new so that destructor will be called and object local types will be freed
- Had to trick the compiler a bit when compiling for CLR

### DIFF
--- a/vowpalwabbit/global_data.h
+++ b/vowpalwabbit/global_data.h
@@ -14,6 +14,7 @@ license as described in the file LICENSE.
 #include <cstdio>
 #include <inttypes.h>
 #include <climits>
+#include <thread>
 
 #include "v_array.h"
 #include "array_parameters.h"
@@ -434,11 +435,8 @@ struct vw
 { shared_data* sd;
 
   parser* p;
-#ifndef _WIN32
-  pthread_t parse_thread;
-#else
-  HANDLE parse_thread;
-#endif
+  std::thread parse_thread;
+
   AllReduceType all_reduce_type;
   AllReduce* all_reduce;
 
--- a/vowpalwabbit/global_data.h
+++ b/vowpalwabbit/global_data.h
@@ -15,11 +15,10 @@ license as described in the file LICENSE.
 #include <inttypes.h>
 #include <climits>
 
+// Thread cannot be used in managed C++, tell the compiler that this is unmanaged even if included in a managed project.
 #ifdef _M_CEE
 #pragma managed(push,off)
-#	undef _M_CEE 
 #include <thread>
-#	define _M_CEE 001
 #pragma managed(pop)
 #else 
 #include <thread>
--- a/vowpalwabbit/global_data.h
+++ b/vowpalwabbit/global_data.h
@@ -14,7 +14,16 @@ license as described in the file LICENSE.
 #include <cstdio>
 #include <inttypes.h>
 #include <climits>
+
+#ifdef _M_CEE
+#pragma managed(push,off)
+#	undef _M_CEE 
 #include <thread>
+#	define _M_CEE 001
+#pragma managed(pop)
+#else 
+#include <thread>
+#endif 
 
 #include "v_array.h"
 #include "array_parameters.h"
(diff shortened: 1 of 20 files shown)


jackgerrits:
This actually contains a regression, I think a deadlock. I'm going to close it until I get time to figure it out

vowpalwabbit/parser.cc
```diff
@@ -1005,59 +932,51 @@ void finish_example(vw& all, example& ec)
...
     return p->examples + ring_index;
   }
   else
   {
     if (!p->done)
     {
-      condition_variable_wait(&p->example_available, &p->examples_lock);
-      mutex_unlock(&p->examples_lock);
+      p->example_available.wait(lock);
       return get_example(p);
```
kumpera: This extends for how long `examples_lock` is held.
What is the locking semantics of  `get_example` ?

vowpalwabbit/parser.h
```diff
@@ -34,11 +45,12 @@ struct parser
   example* examples;
   uint64_t used_index;
   bool emptylines_separate_examples; // true if you want to have holdout computed on a per-block basis rather than a per-line basis
-  MUTEX examples_lock;
-  CV example_available;
-  CV example_unused;
-  MUTEX output_lock;
-  CV output_done;
+
+  std::mutex examples_lock;
```
kumpera: It's nice to include a comment on what's protected by this lock.

vowpalwabbit/parser.h
```diff
@@ -34,11 +45,12 @@ struct parser
...
+  std::condition_variable example_available;
```
kumpera: A comment saying which lock to use with each cond_var goes a long way when reviewing for concurrency.

vowpalwabbit/global_data.h
```diff
@@ -15,6 +15,16 @@ license as described in the file LICENSE.
 #include <inttypes.h>
 #include <climits>
 
+#ifdef _M_CEE
+#pragma managed(push,off)
+#	undef _M_CEE
```
lokitoth: Consider adding a comment to give context for undefing and re-defing _M_CEE around including <thread>. I am curious what issues it causes otherwise.

jackgerrits: It's possible that the undef and def is redundant. Maybe just the managed bit is required?

lokitoth: Not sure. If all you care about is ensuring that /clr does not attempt to compile any of it as managed, per the docs, pushing managed to off should be sufficient. I do not know whether MSVC's <thread> does anything odd with _M_CEE

vowpalwabbit/parser.cc
```diff
@@ -147,17 +75,17 @@ bool is_test_only(uint32_t counter, uint32_t period, uint32_t after, bool holdou
...
-  ret.local_example_number = 0;
-  ret.in_pass_counter = 0;
-  ret.ring_size = 1 << 8;
-  ret.done = false;
-  ret.used_index = 0;
-  ret.jsonp = nullptr;
-
-  return &ret;
+  auto ret = new parser();
+  ret->input = new io_buf;
```
JohnLangford: This doesn't matter at a macro scale, but it's slightly more elegant to auto& ret then use '.' to dereference instead of '->'.

vowpalwabbit/parser.h
JohnLangford: Good idea.  The answer seems to be 'p->done', 'ec.in_use', p->begin_parsed_examples, p->end_parse_examples, and p->used_index.

JohnLangford:
This seems like a clear win from a code complexity and dependency standpoint.   Let's merge after addressing minor comments.

vowpalwabbit/global_data.h
jackgerrits: It seems like the _M_CEE  is required

PR #1732, opened 2019-01-29
Scripts for running tests and generating NuGet packages

lokitoth:
* Creates a test script to run C smoke tests, C++ unit tests and C# UTs
* Creates a package script to generate NuGet packages
* Updates the nuspec files to properly support injecting paths into them

Note: The test script is currently designed to be used by humans, and it
  may fail to error out a CI/CD process that simply executes the script.

### DIFF
--- a/.scripts/package.cmd
+++ b/.scripts/package.cmd
@@ -1,0 +1,28 @@
+@ECHO OFF
+IF DEFINED DebugBuildScripts (
+    @ECHO ON
+)
+
+SETLOCAL
+
+CALL %~dp0init.cmd
+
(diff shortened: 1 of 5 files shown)


lokitoth:
For some reason it is not picking up the AppVeyor build: https://ci.appveyor.com/project/JohnLangford/vowpal-wabbit/builds/21982213

Will just wait for another PR to be merged and trigger another merge of 'master'

JohnLangford:
This seems like an obviously good idea.   

I'd like to eventually reach the point where we can do a pushbutton release across all platforms.

PR #1733, opened 2019-01-29
Fix a small issue and cleanup usage of long long

kumpera:
Fix an issue in python bindings revealed by a warning (implicit cast from 64 to 32 bits int).

Remove usage of `long long` in favor of int64_t. The code base uses both and the later is both shorter and is more intuitive.


JohnLangford:
Looks good to go to me.

Issue #1728, opened 2019-01-25
Recent significant regression in vw training-time


arielf:
#### Describe the bug

Seeing a significant regression in training-times with latest `vw`.
The regression occurred somewhere in the past ~6 months.

#### To Reproduce

Build latest `vw` like this:
```
mkdir -p build
cd build
env CC=clang CXX=clang++ cmake -DCMAKE_BUILD_TYPE=Release -DSTATIC_LINK_VW=ON ..
make -j
# ignore some test related build errors, we only need vw itself
```

```
time ./build/vowpalwabbit/vw -k -c --passes 5000 --quiet --holdout_off -d  test/train-sets/0002.dat
18.84su 2.22ss 11.38s 185%
```

With older `vw` (built statically on July 8, 2018):
```
time ~/bin/vw -k -c --passes 5000 --quiet --holdout_off test/train-sets/0002.dat
4.65su 1.25ss 2.97s 198%
```

Used 5000 passes just to get a long enough running time.
(the regression also occurs is single pass; time is too short to reliably measure)
Progress output looks the same.

But new `vw` is nearly 4x slower to train.
 
#### Expected behavior

Expect `vw` to train as fast as it used to.

#### Observed Behavior

Training time for some data-sets like the example test above have become almost 4x slower.

#### Environment

Ubuntu 18.04
Build with `clang++` (when building with gcc, runtime is even slower) and statically linked.
Latest `vw` (cloned on 2019-01-24)

#### Additional context

I believe the problem is recent because I've been building `vw` and testing speed regularly.

JohnLangford:
I don't know any good reason for this.  

Can you binary search for the critical checkin?

arielf:
@JohnLangford yes, I plan to get to this.

Running `git bisect` got a bit complicated because of the switch to `cmake` in the middle of multiple other changes.

Speculation: I might be hitting a pathological L2/L3 cache miss because of creeping code size growth, or maybe one of the newer Intel CPU freq-down-scaling adjustments on the fly when the chip gets too hot.  This is running on a pretty new Intel(R) Core(TM) i7-8559U CPU with 256KiB L1 cache, 1MiB L2 cache, and 8MiB L3 cache.

At any rate, both are running on the same hardware.  I'll try to reproduce on different hardware and it would also help if someone else can reproduce using the older code.  I have old statically linked Linux binaries anyone to reproduce. at http://finance.yendor.com/ML/VW/Binaries

[...]

PR #1735, opened 2019-01-31
Fix build scripts forcing Debug builds. Add LTO mode and fix VW default visibility.

kumpera:
Release builds were not working for two reasons. First, it would always
override -DCMAKE_BUILD_TYPE with "Debug" Second, it would use CONFIG, which
is not set on OSX with cmake 3.13.1 at least.

Finally, set the default symbol visibility to hidden on unix - linux is a
misnomer in the build script... Without it, we export something like 700
extra symbols that corresponds to all public functions in VW. Exporting
them on unix means making them available to be interposed and requiring an
indirect call to be used at all call sites.

The LTO mode, which can be used with -DLTO=1, enables clang's thinlto mode.

Debug: 9.191kb
Relase+bad visibility: 3.728kb
Release+good visibility: 3.615kb (-113kb)
Release+lto+viz: 3.374kb (-354kb)

Doing `vw -c rcv1.train.raw.txt -b 22 --ngram 2 --skips 4 -l 0.25 --binary` on `http://hunch.net/~jl/VW_raw.tar.gz`. Removed cache file between runs.

Debug: 
real	0m30.582s
user	0m52.192s

Release bad visility:
real	0m7.007s
user	0m11.133s

Release good visibility:
real	0m6.588s
user	0m10.735s

Release + LTO:
real	0m6.602s
user	0m10.580s

LTO is odd, it's worse wallclock but better cpu utilization.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.5)
 # Only allow Release and Debug configurations
 set(CMAKE_CONFIGURATION_TYPES Debug Release CACHE TYPE INTERNAL FORCE)
 if(NOT CMAKE_BUILD_TYPE)
-  set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Choose the type of build." FORCE)
+  set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Choose the type of build." FORCE)
 endif()
 
 project(vowpal_wabbit C CXX)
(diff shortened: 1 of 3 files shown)


kumpera:
I tried full lto and perf was about the same, the binary shrank even further to 3.225kb (-149kb x thin) but perf stayed about the same.

JohnLangford:
This almost certainly addresses #1728 .   What is the definition of "bad visibility" exactly?  

And, Travis is having issues due to the python bindings: https://travis-ci.org/VowpalWabbit/vowpal_wabbit/builds/486765137#L2125  Easy fix?

kumpera:
This PR regressed the python bindings due to lack of export annotations.

JohnLangford:
Looks good, merged :-)

PR #1736, opened 2019-01-31
Add CCB label parsing, label and prediction types

jackgerrits:
- Adds label parser for CCB
- New label and prediction type
- Consolidate `substring_equal` usages
- Unit tests for ccb label parsing

[See here for a definition of the input format](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Contextual-Bandit-algorithms#conditional-contextual-bandit)

### DIFF
--- a/vowpalwabbit/CMakeLists.txt
+++ b/vowpalwabbit/CMakeLists.txt
@@ -34,7 +34,7 @@ set(vw_all_headers
   gen_cs_example.h parse_args.h topk.h cb_explore_adf.h parse_dispatch_loop.h unique_sort.h
   interact.h interactions.h parse_example_json.h cbify.h interactions_predict.h vw_allreduce.h
   classweight.h parse_regressor.h kernel_svm.h confidence.h label_dictionary.h
-  config.h.in primitives.h lda_core.h print.h vw_versions.h
+  config.h.in primitives.h lda_core.h print.h vw_versions.h conditional_contextual_bandit.h
 )
 
 set(vw_all_sources
@@ -50,7 +50,7 @@ set(vw_all_sources
   active_cover.cc cs_active.cc kernel_svm.cc best_constant.cc ftrl.cc svrg.cc lrqfa.cc interact.cc
   comp_io.cc interactions.cc vw_validate.cc audit_regressor.cc gen_cs_example.cc cb_explore.cc
   action_score.cc cb_explore_adf.cc OjaNewton.cc parse_example_json.cc baseline.cc classweight.cc
-  vw_exception.cc no_label.cc
+  vw_exception.cc no_label.cc conditional_contextual_bandit.cc
 )
 
 set(explore_all_headers
--- a/vowpalwabbit/conditional_contextual_bandit.cc
+++ b/vowpalwabbit/conditional_contextual_bandit.cc
@@ -3,6 +3,8 @@
 #include "example.h"
 #include "global_data.h"
 
+#include <numeric>
+
 bool CCB::ec_is_example_header(example& ec)
 {
   return ec.l.conditional_contextual_bandit.type == example_type::shared;
@@ -69,29 +71,43 @@ namespace CCB {
 
   void cache_label(void* v, io_buf& cache)
   {
-    /*char* c;
-    CCB::label* ld = static_cast<CCB::label*>(v);
-    cache.buf_write(c, sizeof(size_t) + sizeof(cb_class) * ld->costs.size());
-    bufcache_label(ld, c);*/
+    //char* c;
+    //CCB::label* ld = static_cast<CCB::label*>(v);
+    //cache.buf_write(c, sizeof(uint8_t) + sizeof(bool) + sizeof(float) + sizeof(uint32_t) + sizeof(ACTION_SCORE::action_score) * ld->probabilities.size());
+
+    //// type
+    //*(uint8_t*)c = static_cast<uint8_t>(ld->type);
+    //c += sizeof(uint8_t);
+
+
+    //*(bool*)c = (ld->type);
+    //c += sizeof(uint8_t);
   }
 
   void default_label(void* v)
   {
     CCB::label* ld = static_cast<CCB::label*>(v);
-    ld->outcomes.clear();
-    ld->excluded_actions.clear();
+    ld->outcome = nullptr;
+    ld->explicit_included_actions = v_init<uint32_t>();
+    ld->type = example_type::unset;
   }
 
   bool test_label(void* v)
   {
     CCB::label* ld = static_cast<CCB::label*>(v);
-    return ld->outcomes.size() == 0;
+    return ld->outcome == nullptr;
   }
 
   void delete_label(void* v)
   {
     CCB::label* ld = static_cast<CCB::label*>(v);
-    ld->outcomes.delete_v();
+    if (ld->outcome)
+    {
+      ld->outcome->probabilities.delete_v();
+      delete ld->outcome;
+      ld->outcome = nullptr;
+    }
+    ld->explicit_included_actions.delete_v();
   }
 
   void copy_label(void* dst, void* src)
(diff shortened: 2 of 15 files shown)


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,300 @@
...
+#include <numeric>
+
+bool CCB::ec_is_example_header(example& ec)
+{
+  return ec.l.conditional_contextual_bandit.type == example_type::shared;
+}
+
+void CCB::print_update(vw& all, bool is_test, example& ec, multi_ex* ec_seq, bool action_scores)
+{
+  // TODO: Implement for CCB
```
JohnLangford: This is the exponential backoff printout which VW does by default, so it's obviously important to implement.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,300 @@
...
+  throw std::runtime_error("CCB:print_update not implemented");
+}
+
+template <typename T>
+T read_object(io_buf& cache)
+{
```
JohnLangford: This is nice.  It would benefit other cache readers.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,300 @@
...
+  void cache_label(void* v, io_buf& cache)
+  {
+    char* c;
+    CCB::label* ld = static_cast<CCB::label*>(v);
+    size_t size =
+      sizeof(uint8_t) // type
+      + sizeof(bool) // outcome exists?
+      + (ld->outcome == nullptr ? 0 :
+        sizeof(float) // cost
+        + sizeof(size_t) // probabilities size
```
JohnLangford: We should avoid read/write of objects of nonfixed size to preserve compatibility across 32/64 bit platforms.

JohnLangford:
I'm not quite following the format alteration.  In particular, I'm worried that we need the ability to specify a sparse probability distribution.  That's not possible with this, which causes ambiguities?

Aside from addressing the comments, I'd like to hold off on merging this until the complete CCB patch is ready to merge.

vowpalwabbit/conditional_contextual_bandit.cc
JohnLangford: Yes.   Early tomorrow morning?

vowpalwabbit/conditional_contextual_bandit.cc
JohnLangford: (Maybe shift to cache.h?)

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: I am not sure how we can avoid this here?

JohnLangford: Arrange to use uint32_t or uint64_t.

jackgerrits: Should we just enforce that uint32_t is used here then?

JohnLangford: Yes, make the relevant variables have a fixed length.

JohnLangford: More generally, it seems better to use sizeof(variable_name) than sizeof(typename) since that better documents why a particular size is needed.

jackgerrits:
We can still specify a sparse distribution. The proposed changed is 
`ccb decision 1:0.8:0.8,0:0.2 1,2` | d:7 -> `ccb decision 1:0.8:0.8:0,0.2 1,2 | d:7` where the `:` and `,` are reversed. This is for 2 reasons. It is easier to parse, and the label is composed of logical blocks delimited by `:` where the action probability pairs are delimited by a `,`. Perhaps a different delimiting character would reduce confusion with this?

What is the rationale for waiting to merge? I was hoping to do it in smaller chunks so it is easier to review and then Yann can consume the changes easier as well.

vowpalwabbit/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ set(vw_all_headers
   gen_cs_example.h parse_args.h topk.h cb_explore_adf.h parse_dispatch_loop.h unique_sort.h
   interact.h interactions.h parse_example_json.h cbify.h interactions_predict.h vw_allreduce.h
   classweight.h parse_regressor.h kernel_svm.h confidence.h label_dictionary.h
-  config.h.in primitives.h lda_core.h print.h vw_versions.h
+  config.h.in primitives.h lda_core.h print.h vw_versions.h conditional_contextual_bandit.h
```
yannstad: Why not `ccb.h` instead of `conditional_contextual_bandit.h`

vowpalwabbit/CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ set(vw_all_sources
   active_cover.cc cs_active.cc kernel_svm.cc best_constant.cc ftrl.cc svrg.cc lrqfa.cc interact.cc
   comp_io.cc interactions.cc vw_validate.cc audit_regressor.cc gen_cs_example.cc cb_explore.cc
   action_score.cc cb_explore_adf.cc OjaNewton.cc parse_example_json.cc baseline.cc classweight.cc
-  vw_exception.cc no_label.cc
+  vw_exception.cc no_label.cc conditional_contextual_bandit.cc
```
yannstad: Why not `ccb.cc` instead of `conditional_contextual_bandit.cc`

vowpalwabbit/conditional_contextual_bandit.h
```diff
@@ -0,0 +1,50 @@
...
+
+struct vw;
+struct example;
+typedef std::vector<example*> multi_ex;
+
+namespace CCB {
+  // Each positon in outer array is implicitly the decision corresponding to that index. Each inner array is the result of CB for that call.
+  typedef v_array<ACTION_SCORE::action_scores> decision_scores_t;
+
+  struct conditional_contexual_bandit_outcome
```
yannstad: rename `conditional_contexual_bandit_outcome` in `ccb_outcome`?

vowpalwabbit/example.h
```diff
@@ -40,6 +41,7 @@ typedef union {
   MULTICLASS::label_t multi;
   COST_SENSITIVE::label cs;
   CB::label cb;
+  CCB::label conditional_contextual_bandit;
```
yannstad: rename `conditional_contextual_bandit` in `ccb` ?

vowpalwabbit/example.h
```diff
@@ -55,6 +57,7 @@ typedef union {
   v_array<float> scalars;           // a sequence of scalar predictions
   ACTION_SCORE::action_scores a_s;  // a sequence of classes with scores.  Also used for probabilities.
   uint32_t multiclass;
+  CCB::decision_scores_t decision_scores;
```
yannstad: It looks to me that `decision_scores_t d_s` would be more consistent with `action_scores a_s`.

vowpalwabbit/conditional_contextual_bandit.h
jackgerrits: I opted for the long form throughout as ccb is easy to mixup with cb. Thoughts?

yannstad: I am fine with both, so let's do what you prefer.
I prefer `ccb_outcome` regarding cosmetics mainly. Also because i prefer shorter names.

jackgerrits:
@JohnLangford I believe I have addressed the comments. I have implemented the parsing in the format you were expecting so you can disregard my previous comments about changing it.

I'm going to close this PR and open a new one that points to a ccb staging branch where we will keep this work until it is ready for master.

jackgerrits:
closing in favor of #1754

PR #1737, opened 2019-02-01
Shorten Readme

jackgerrits:
This removes all information about how to use VW from the readme and points to the wiki instead. Previously the readme was a very long document that was somewhat daunting to consume. Much of the information was out of date as well which was causing issues. I have migrated all instructions to the wiki in an easier to consume format and have updated the readme to provide a value proposition for vw and then links to the relevant places in the wiki.


JohnLangford:
Merged, thanks :-)

PR #1740, opened 2019-02-04
Small batch of cleanups to io_buff

kumpera:
As part of trying to make the IO stack modular, I'm trying to tame the design of io_buf in such a way that internals don't bleed all around.

This is a small step towards better abstracting the underlying data source and avoid some of the oddities we have when using it with non-file streams.

### DIFF
--- a/vowpalwabbit/OjaNewton.cc
+++ b/vowpalwabbit/OjaNewton.cc
@@ -513,7 +513,7 @@ void save_load(OjaNewton& ON, io_buf& model_file, bool read, bool text)
     ON.initialize_Z(all.weights);
   }
 
-  if (model_file.files.size() > 0)
+  if (model_file.num_files() > 0)
   {
     bool resume = all.save_resume;
     stringstream msg;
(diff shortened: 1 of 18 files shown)


JohnLangford:
This seems good, but something is barfing in the tests?

kumpera:
I'll double check the test results tomorrow.

palmerlao:
I think this question/drive-by suggestion may qualify as a small cleanup for `io_buf`, but if it isn't correct or doesn't fit here well, I'd be happy to open a separate PR. 

Is there a special use case for which it does not make sense to close all the file descriptors associated with an `io_buf` in the `io_buf`'s destructor? If not, I think it's safe to add `close_files` to the destructor since it's idempotent. Even better might be to have a separate RAII struct/class that holds a file descriptor instead of using the raw `int`s in `io_buf`, but maybe I'm glossing over some important details here?

kumpera:
@palmerlao I can't answer that question precisely as I'm still going though the codebase learning how io_buf is used.

kumpera:
@JohnLangford the tests are failing for me on master. I'll rather fix them first before blindly merging things

JohnLangford:
@palmerlao I don't know of a use case where you can't close all files in the destructor.

JohnLangford:
I'm confused about why this one is failing---it seems like such a simple patch.

kumpera:
The crash is due to parser.cc : finalize_source modifying `io_buff::files` but using `io_buff::num_files()` to check its size. I discussed it with @jackgerrits and we believe the right cleanup here is to replace files/gz_files.

PR #1742, opened 2019-02-04
Fix model file parser for proper treatment of negative-valued options

### LINKED ISSUE #1741: VM model files containing option values of the form '-<digit>' cannot be loaded
#### Describe the bug
Models containing lambda "-1" cannot be loaded, because the "-1" is being misinterpreted as a short option-key in the parser during the merge options logic.

#### To Reproduce
1. Get model file from https://github.com/lokitoth/lokitoth.github.io/blob/master/repros/vw/vowpal_wabbit/1741/model-unloadable.vw
2. Use the following vw configuration: vw -i model-unloadable.vw --readable_model <somepath> -t --no_stdin

#### Expected behavior
VW loads the specified model file and exports it as a human-readable model

#### Observed Behavior
VW errors out stating that '---1' is not a valid options for '--lambda'

#### Environment
Windows native command-line vw (vw.exe, built by vw.vcxproj) built off latest master, version-tagged 8.6.1

lokitoth:
The previous program options changes introduced an issue where models that have a stored value of the form "-&lt;digit&gt;" will have the options merge logic incorrectly treat this value as a key and merge it in as "---&lt;digit&gt;", which then breaks the final parser pass.

The fix is to add a rule that option key identifiers cannot start from a digit (similar rules to variable identifiers), and treat it as a value when running the merge logic.

This fixes #1741


lokitoth:
The CI is failing for a clang-format check failure.

I would like to push back against this format change as it makes it dramatically less readable:

(in code)
`if (opt.string_key.length() > 1 && // Skip empty options and ones without at least two characters in string_key
opt.string_key[0] == '-' &&
opt.string_key[1] >= '0' &&
opt.string_key[1] <= '9')`

(proposed by clang format)
`if (opt.string_key.length() > 1 && // Skip empty options and ones without at least two characters in string_key
opt.string_key[0] == '-' && opt.string_key[1] >= '0' && opt.string_key[1] <= '9')`

It seems to be a less consistent and less glanceable form of the original.

cc @JohnLangford @palmerlao

JohnLangford:
Perhaps we should also discuss whether clang-format should fail or advise.... personally I'm more inclined towards 'advise' with fixing things periodically.

lokitoth:
The comment is not that critical, since with the longer comment above it seems a bit superfluous. 

It is a bit of a red-herring, though: I would not be a big fan of putting all the checks on one line, given that there are a number of them, and it forces the reader to go through them linearly to understand what the full check is. 

If the structure was more complicated (parentheses, ||s, etc.) it would be even more desirable to be able to format it manually to indicate the boundaries of the subexpressions.

JohnLangford:
This sounds like opposition to the concept of code format?

I'm a little bit skeptical of that in the sense that I don't expect the least mature of programmers to necessarily format things sanely, and I don't want to police format actively.

lokitoth:
Not an opposition in principle, though I can understand how that might have come across that way. I would just prefer to have it err on the side of leaving things alone. If the formatter makes it harder to read the code, is the consistency worth it?

As an aside: It seems odd to me to be in a situation where we have a tool that will only accept one input as non-broken, and is also capable of changing the provided input to match the input it wants, but still have it break the CI, rather than just running the fix and checking in the changes.

JohnLangford:
(I was thinking more in terms of a cronjob although is also awkward in a different way.)

lokitoth:
If there was a non-platform-specific way to do it, maybe a git checkin hook?

JohnLangford:
It sounds plausible although I'm unclear on the specifics of using a git checkin hook.  Do you know how?

lokitoth:
I'll create a "feature request" issue to track figuring out how to make the experience more streamlined.

JohnLangford:
Merged, thanks.

palmerlao:
If you're particular about a few special cases, you can always turn clang-format off for certain sections of the code.

FWIW the repo I contribute to at work is similar in terms of formatting breaking CI. Some people use precommit hooks; I have my editor format on save.

JohnLangford:
In a work environment I think it makes more sense to require formatting.  For an open source project, I'm worried that it's perceived as excessive friction so we'd really like this to work automatically.

PR #1747, opened 2019-02-06
[tests] Make repeat.py compatible with python 3

kumpera:
Fixes `test_race_condition` on OSX High Sierra that comes with python 3 by default.


JohnLangford:
Merged, thanks :-)

PR #1751, opened 2019-02-07
Fix test 175

yannstad:
This PR fixes some test issues:
* Test 175 is not enabled on Windows, because the test parser expects an empty line after the test definition.
* When enabled, the test is failing, because an empty line is missing after the example (thus failing the multi-line detection).
* Test 59 does not exist, misleading the reading of Travis logs, where the test id is not the parsed id but a counter starting from 1.


rajan-chari:
Looks good to me.

PR #1752, opened 2019-02-07
Convert TopK reduction to be multiline example based

### LINKED ISSUE #1748: TopK should be a multi line reduction - illegal usage of example field in current state
This causes a memory leak as well as undefined behavior when examples are retrieved from the buffer as many of the fields of example have previous values left in them. Therefore if a reduction does not overwrite the prediction field the old value will be used.

https://github.com/VowpalWabbit/vowpal_wabbit/blob/c9110426d9e8585e8410403d12f0194d5e6673fa/vowpalwabbit/parser.cc#L908

jackgerrits:
Fixes #1748

### DIFF
--- a/test/train-sets/ref/topk-rec.stderr
+++ b/test/train-sets/ref/topk-rec.stderr
@@ -1,4 +1,4 @@
-creating quadratic features for pairs: MF 
+creating quadratic features for pairs: MF
 predictions = topk-rec.predict
 Num weight bits = 18
 learning rate = 0.5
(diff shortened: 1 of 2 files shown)


test/train-sets/ref/topk-rec.stderr
```diff
@@ -12,20 +12,17 @@ loss     last          counter         weight    label  predict features
...
 finished run
-number of examples = 12
-weighted example sum = 12.000000
+number of examples = 9
+weighted example sum = 9.000000
 weighted label sum = 15.000000
 average loss = 0.000002
 best constant = 1.666667
-total feature number = 39
+total feature number = 36
```
JohnLangford: This changes because you eliminate newline examples with a constant feature?

test/train-sets/ref/topk-rec.stderr
```diff
@@ -12,20 +12,17 @@ loss     last          counter         weight    label  predict features
...
-0.000003 0.000000            9            9.0   2.0000   2.0004        4
-0.000003 0.000000           10           10.0   1.0000   1.0003        4
-0.000002 0.000000           11           11.0   3.0000   2.9995        4
-0.000002     n.a.           12           12.0  unknown   0.0000        1
+0.000003 0.000003            4            4.0   0.0000   0.0018        4
+0.000004 0.000011            5            5.0   3.0000   2.9968        4
+0.000004 0.000001            6            6.0   1.0000   1.0007        4
+0.000003 0.000000            7            7.0   2.0000   2.0004        4
+0.000003 0.000000            8            8.0   1.0000   1.0003        4
+0.000002 0.000000            9            9.0   3.0000   2.9995        4
```
JohnLangford: The number of multiline examples should be 3, right?

test/train-sets/ref/topk-rec.stderr
jackgerrits: Yes exactly

test/train-sets/ref/topk-rec.stderr
jackgerrits: Yes it is 3, but I am recreating the previous behavior of printing statistics after each example. See line 91 in the reduction. What do you think is the correct thing to do here? Only print the statistics after each set of examples?

JohnLangford: Yes.   All predictions from a top-k call should (conceptually) be on a single line.  The idea of this reduction is that you are return the top-k-of-n things.

JohnLangford:
Merged, thanks :-)

PR #1753, opened 2019-02-08
RunTests: use test label number instead of counter

albietz:
In the testing logs (e.g. in Travis), this shows the test label shown in the comment above each test, rather than just a counter. For instance, the logs will show 175 for the test labeled `# Test 175:`, instead of the value of the counter. This avoids weird discrepancies when reading the Travis logs.

cc @yannstad

### DIFF
--- a/test/RunTests
+++ b/test/RunTests
@@ -764,6 +764,7 @@ exit $ErrorCount;
 #   You may break very long lines using \ at EOL.
 #
 #   # -------------------------------------
+#   # Test <test_id>: ...
 #   shell command which may include {VW} ...
 #       reference/file1
 #       reference/file2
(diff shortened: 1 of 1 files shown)


yannstad:
LGTM.
Thanks!

JohnLangford:
@arielf any comments?

yannstad:
At least it would be good to update the RunTests doc (line ~750 in the RunTests file).

jackgerrits:
@JohnLangford are we able to get this in?

JohnLangford:
Yes, let's merge.

JohnLangford:
That seems plausible as there is no need to be over-specific.  Send a pull request?

PR #1759, opened 2019-02-11
Small Json parser cleanup

jackgerrits:
- Remove unused cc file
- Remove goto
- Remove unused boost header
- Use RAII types

### DIFF
--- a/vowpalwabbit/parse_example_json.h
+++ b/vowpalwabbit/parse_example_json.h
@@ -29,7 +29,9 @@ license as described in the file LICENSE.
 
 #include "cb.h"
 #include "best_constant.h"
-#include <boost/algorithm/string.hpp>
+
+#include <algorithm>
+#include <vector>
 
 // portability fun
 #ifndef _WIN32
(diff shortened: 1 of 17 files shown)


vowpalwabbit/parse_example_json.h
```diff
@@ -1038,19 +1028,19 @@ struct Context
...
+      // Do not insert feature_group if it already exists.
+      if(std::find(ex->indices.begin(), ex->indices.end(), feature_group) == ex->indices.end())
+      {
+        ex->indices.push_back(feature_group);
+      }
     }
 
-  done:
-    return namespace_path.pop().return_state;
+    auto return_state = namespace_path.back().return_state;
```
ataymano: [minor] namespace_path.back()  ->   ns?

vowpalwabbit/parser.cc
```diff
@@ -171,7 +163,10 @@ void reset_source(vw& all, size_t numbits)
       else
       {
         int fd = input->files.pop();
-        if (!member(all.final_prediction_sink, (size_t)fd))
+        auto& fps = all.final_prediction_sink;
```
ataymano: [minor] const auto&?

vowpalwabbit/parse_example_json.h
jackgerrits: I'd like to leave it as is because it is the normal way you expect to see the back of a queue being popped

PR #1765, opened 2019-02-20
Migrate from v_array to std::vector in some places + other changes

jackgerrits:
- Migrate some usages of v_array to std:vector where possible
- Use std::stack for reduction stack
- Make tokenize templated to allow any container type

### DIFF
--- a/vowpalwabbit/parse_args.cc
+++ b/vowpalwabbit/parse_args.cc
@@ -1208,72 +1208,75 @@ void load_input_model(vw& all, io_buf& io_temp)
 
 LEARNER::base_learner* setup_base(options_i& options, vw& all)
 {
-  LEARNER::base_learner* ret = all.reduction_stack.pop()(options, all);
-  if (ret == nullptr)
+  auto setup_func = all.reduction_stack.top();
+  all.reduction_stack.pop();
+  auto base = setup_func(options, all);
+
+  if (base == nullptr)
     return setup_base(options, all);
   else
-    return ret;
+    return base;
 }
 
 void parse_reductions(options_i& options, vw& all)
 {
   // Base algorithms
-  all.reduction_stack.push_back(GD::setup);
-  all.reduction_stack.push_back(kernel_svm_setup);
-  all.reduction_stack.push_back(ftrl_setup);
-  all.reduction_stack.push_back(svrg_setup);
-  all.reduction_stack.push_back(sender_setup);
-  all.reduction_stack.push_back(gd_mf_setup);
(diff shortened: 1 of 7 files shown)


jackgerrits:
@JohnLangford, this is fixed now. I changed a bit of behavior to do with exception propagation which I have now reverted back to original behavior.

vowpalwabbit/parse_args.cc
```diff
@@ -1208,72 +1208,75 @@ void load_input_model(vw& all, io_buf& io_temp)
 
 LEARNER::base_learner* setup_base(options_i& options, vw& all)
 {
-  LEARNER::base_learner* ret = all.reduction_stack.pop()(options, all);
-  if (ret == nullptr)
+  auto setup_func = all.reduction_stack.top();
+  all.reduction_stack.pop();
+  auto base = setup_func(options, all);
```
JohnLangford: It's somewhat unfortunate that 1 line becomes 3.

jackgerrits:
I agree, I don’t understand why pop is void.

vowpalwabbit/parser.cc
```diff
@@ -127,11 +121,11 @@ uint32_t cache_numbits(io_buf* buf, int filepointer)
   }
   catch (...)
   {
-    t.delete_v();
+    // Prior to using a std::vector, a v_array was used and the catch statement was used to delete it.
+    // This caused any exception to be ignored. Bubbling up this exception causes tests to fail so
+    // I am going to swallow it to maintain previous behavior.
```
JohnLangford: I'm slightly shaky on this.  Are we ignoring exceptions that we should not be ignoring?

JohnLangford:
Overall, this looks good.  I'd like to understand the exception issue slightly more before merging.

vowpalwabbit/parser.cc
jackgerrits: There are 4 exceptions that are swallowed by this function, see below. These probably should not be ignored. Fixing this behavior breaks some tests, but those are probably using invalid cache files.
```
THROW("cache version too long, cache file is probably invalid");
THROW("cache version too short, cache file is probably 
THROW("failed to read");
THROW("data file is not a cache file");
```

JohnLangford: Do you want to fix in this patch or merge and fix later?

jackgerrits: I'd like to fix separately. I can add a TODO and link it to an issue to make sure we get back to it

JohnLangford:
Merged, thanks :-)

PR #1761, opened 2019-02-14
[CCB] Add ccb_explore_adf reduction

yannstad:
PR description:
* add a new option `--ccb_explore_adf`, that triggers `-cb_explore_adf` under the hood
* use the new label parser for CCB
* Implement the new ccb reduction that converts a ccb multiline example into a sequence of cb multi-line examples. Between 2 cb call, previously chosen actions are removed from all available actions
* implement the explicit inclusion logic, that overrides the list of available actions

Some remaining work should be done, in order to have a functional CB:
* decision features should be used (currently ignored)
* decision features should be crossed with all other features (shared and actions)
* handle console 'printing' - currently broken for CCB

PR #1767, opened 2019-02-21
[CCB] Dsjson and json parsing for CCB

jackgerrits:
- Implement JSON parsing in normal and dsjson mode for CCB
- Add tests for dsjson and json for simple, cb and ccb

Note: this will not build until #1761 is merged


vowpalwabbit/parse_example_json.h
```diff
@@ -640,16 +751,40 @@ class DefaultState : public BaseState<audit>
       if (ctx.key_length == 6 && !strcmp(str, "_multi"))
         return &ctx.multi_state;
 
+      if (ctx.key_length == 3 && !strcmp(str, "_df"))
+        return &ctx.df_state;
+
       if (ctx.key_length == 4 && !_stricmp(ctx.key, "_tag"))
         return &ctx.tag_state;
 
+      if (ctx.key_length == 4 && !_stricmp(ctx.key, "_inc"))
```
ataymano: is there any specific reason for using ctx.key in some places and str in others?

vowpalwabbit/parser.cc
```diff
@@ -549,19 +549,33 @@ void enable_sources(vw& all, bool quiet, size_t passes, input_options& input_opt
...
+        }
+
         // TODO: change to class with virtual method
         // --invert_hash requires the audit parser version to save the extra information.
         if (all.audit || all.hash_inv)
         {
           all.p->reader = &read_features_json<true>;
           all.p->audit = true;
-          all.p->jsonp = std::make_shared<json_parser<true>>();
+          auto parser_context = std::make_shared<json_parser<true>>();
```
JohnLangford: Minor, but why don't you assign to the final destination and then modify?

JohnLangford: Even better, should mode be an argument to the constructor?

vowpalwabbit/parse_example_json.h
jackgerrits: I don't believe there is a reason, especially since it is set at the top of the function

JohnLangford:
Maybe add a test an end-to-end test in RunTests?  That will document the usage pattern.

PR #1768, opened 2019-02-21
Offset tree

rajan-chari:
Implementation of offset_tree predict().
Offset tree paper for reference:  https://arxiv.org/abs/0812.4044

Some differences from the offset tree paper:
Instead of binary classifier at each node, 2 action contextual bandit exploration is used.
Since we would like to return scores over all actions, the predict() algorithm is different also.


vowpalwabbit/offset_tree_internal.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
+#include <cstdint>
+#include <vector>
+#include "learner.h"
+namespace offset_tree
+{
+  struct tree_node {
+    tree_node(uint32_t i, uint32_t l, uint32_t r, bool lf);
```
jackgerrits: Can you use more descriptive names here?

vowpalwabbit/offset_tree_internal.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
+#include <cstdint>
+#include <vector>
+#include "learner.h"
+namespace offset_tree
```
jackgerrits: This should be nested inside of the VW namespace.

Side comment: I would really like the see the base namespace be lowercase `vw`. Uppercase seems very out of place with everything else. But `vw` is already taken by the global object so we would need to rename it (probably not a bad idea)

vowpalwabbit/offset_tree_internal.h
```diff
@@ -0,0 +1,40 @@
...
+    bool is_leaf;
+  };
+
+  struct min_depth_binary_tree {
+    min_depth_binary_tree();
+    void build_tree(uint32_t num_nodes);
+    inline uint32_t internal_node_count() const;
+    inline uint32_t leaf_node_count() const;
+
+    std::vector< tree_node> nodes;
```
jackgerrits:
```suggestion
    std::vector<tree_node> nodes;
```

vowpalwabbit/offset_tree_internal.cc
```diff
@@ -0,0 +1,138 @@
...
+
+  uint32_t min_depth_binary_tree::internal_node_count() const {
+    return nodes.size() - _num_leaf_nodes;
+  }
+
+  uint32_t min_depth_binary_tree::leaf_node_count() const {
+    return _num_leaf_nodes;
+  }
+
+  void offset_tree::init(uint32_t num_actions) {
```
jackgerrits: Can you do this work in the constructor and remove the concept of not being initialized from this object? Is this done for failure handling?

vowpalwabbit/offset_tree_internal.cc
```diff
@@ -0,0 +1,138 @@
...
+
+  bool tree_node::operator==(const tree_node& rhs) const {
+    if (this == &rhs) return true;
+    return (id == rhs.id && left_id == rhs.left_id && right_id == rhs.right_id && is_leaf == rhs.is_leaf);
+  }
+
+  bool tree_node::operator!=(const tree_node& rhs) const { return !(*this == rhs); }
+
+  min_depth_binary_tree::min_depth_binary_tree()
+    : root_idx{0}, _num_leaf_nodes{0}, _initialized{false} {}
```
jackgerrits: consider adding these initializers in the header and leaving the constructor as default

vowpalwabbit/offset_tree_internal.cc
```diff
@@ -0,0 +1,138 @@
...
+    // deal with degenerate cases of 0 and 1 actions
+    if(_num_leaf_nodes == 0) {
+      _initialized = true;
+      return;
+    }
+
+    try {
+      nodes.reserve(2 * _num_leaf_nodes - 1);
+      std::vector<uint32_t> tournaments;
+      tournaments.reserve(_num_leaf_nodes); // Add all leaf nodes to node collection
```
jackgerrits: nit: prefer comment on own line

vowpalwabbit/offset_tree_internal.cc
```diff
@@ -0,0 +1,138 @@
...
+    _num_leaf_nodes = num_nodes;
+    // deal with degenerate cases of 0 and 1 actions
+    if(_num_leaf_nodes == 0) {
+      _initialized = true;
+      return;
+    }
+
+    try {
+      nodes.reserve(2 * _num_leaf_nodes - 1);
+      std::vector<uint32_t> tournaments;
```
jackgerrits: Can you explain the use of the work tournament?

vowpalwabbit/offset_tree_internal.cc
```diff
@@ -0,0 +1,138 @@
...
+    binary_tree.build_tree(num_actions);
+  }
+
+  int32_t offset_tree::learner_count() const {
+    return binary_tree.internal_node_count();
+  }
+
+  const std::vector<float>& offset_tree::predict(LEARNER::single_learner& base, example& ec) {
+    static thread_local std::vector<std::pair<float,float>> prediction_buffer(learner_count());
+    static thread_local std::vector<float> scores(binary_tree.leaf_node_count());
```
jackgerrits: I'm curious why use thread_local here?

vowpalwabbit/offset_tree_internal.cc
```diff
@@ -0,0 +1,138 @@
...
+  void offset_tree::init(uint32_t num_actions) {
+    binary_tree.build_tree(num_actions);
+  }
+
+  int32_t offset_tree::learner_count() const {
+    return binary_tree.internal_node_count();
+  }
+
+  const std::vector<float>& offset_tree::predict(LEARNER::single_learner& base, example& ec) {
+    static thread_local std::vector<std::pair<float,float>> prediction_buffer(learner_count());
```
jackgerrits: Can you add a comment explaining the values stored in the pair?

vowpalwabbit/offset_tree_internal.cc
```diff
@@ -0,0 +1,138 @@
+#include "offset_tree_internal.h"
+#include "vw_exception.h"
```
JohnLangford: Why is it worthwhile to have an internal/not distinction?  Spreading a single reduction across multiple files seems undesirable by default?

vowpalwabbit/offset_tree_internal.cc
JohnLangford: This is probably a leftover from the ECT code although it is valid to think of the offset tree as conducting a single elimination tournament over actions.

vowpalwabbit/offset_tree_internal.cc
JohnLangford: Throughout the VW reduction stack, I've generally avoided allocation on the predict/learn compute paths for efficiency reasons.   Can we avoid it here?

JohnLangford: In this particular case, you should be able to use the stack instead with a DFS search strategy.   To do this, you would compute the prediction at a node, multiply it by the predictions of parent nodes, then recurse left and then right.  In the terminal case you reach a leaf and just store the product of all predictions on the path to the leaf.  

This incidentally eliminates the 1-action special case.

vowpalwabbit/offset_tree_internal.cc
rajan-chari: Think of thread_local as a global static variable per thread when the function is called.   This particular use case can be thought of as an object pool for prediction_buffer and scores without locks.

vowpalwabbit/offset_tree_internal.cc
rajan-chari: See comment above.

Because of thread_local we are not allocating per call.  The objects will stabilize in memory and not have to allocate per call.

rajan-chari: @JohnLangford  I believe this approach is better.  Let me know if I should revisit this question.

vowpalwabbit/offset_tree_internal.cc
rajan-chari: Will introduce a header for min-depth tree and merge these into offset_tree.cc

vowpalwabbit/offset_tree_internal.h
rajan-chari: done

vowpalwabbit/offset_tree_internal.h
rajan-chari: Done

vowpalwabbit/offset_tree_internal.cc
rajan-chari: Needed for scoped_calloc_or_throw

vowpalwabbit/offset_tree_internal.cc
rajan-chari: Done

vowpalwabbit/offset_tree_internal.cc
rajan-chari: Done

vowpalwabbit/offset_tree_internal.cc
rajan-chari: done

vowpalwabbit/offset_tree_internal.cc
rajan-chari: Done.

rajan-chari:
I believe all PR comments are addressed.  I would like to merge to offset_tree branch unless there are other concerns.

vowpalwabbit/min_depth_tree.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
```
JohnLangford: Is there another consumer of this file?

vowpalwabbit/offset_tree.cc
```diff
@@ -0,0 +1,207 @@
...
+    if (t.leaf_node_count() == 1)
+    {
+      scores[0] = 1.0f;
+      return scores;
+    }
+
+    // Get predictions for all internal nodes
+    for (auto idx = 0; idx < t.internal_node_count(); ++idx)
+    {
+      base.predict(ec, idx);
```
JohnLangford: There is a potential optimization here in that we could use a multi_predict operation.   Probably, we should save that for later.

vowpalwabbit/offset_tree.cc
```diff
@@ -0,0 +1,207 @@
...
+    // Compute action scores
+    const auto lrnr_strt_offset = t.leaf_node_count();
+    for (auto rit = t.nodes.rbegin(); rit != t.nodes.rend(); ++rit)
+    {
+      // done processing all internal nodes
+      if (rit->is_leaf)
+        break;
+
+      // update probabilities for left node
+      const float left_p = prediction_buffer[rit->id - lrnr_strt_offset].first;
```
JohnLangford: Why do we need to constantly use lrnr_strt_offset?

JohnLangford:
I think we want an end-to-end test as well to demonstrate use and we should wait until the learning side is handled before merging so it's a functional addition.

vowpalwabbit/offset_tree.cc
rajan-chari: I added an enhancement request to track it.  Unclear to me what multi_predict does.  Do you mind elaborating?

rajan-chari:
I plan on adding End to End test once learn() is complete before merge to master.

vowpalwabbit/min_depth_tree.h
rajan-chari: No.  Will collapse it into offset_tree.h

vowpalwabbit/offset_tree.cc
rajan-chari: min_depth_tree's node collection contains internal and leaf nodes.  lrnr_strt_offset is the begining of all the internal nodes.

rajan-chari: prediction_buffer only needs to hold the predictions for the internal nodes.

rajan-chari: Decided to hide the offset arithmetic behind a class.

rajan-chari:
I believe all PR comments are addressed. I would like to merge to offset_tree branch unless there are other concerns.

PR #1770, opened 2019-02-21
[CCB] Enable interactions to be overridden by an example

jackgerrits:
This is a very trivial change enabling interactions to be overridden by an example. I'll follow this up with an object pooled approach. However I think the opt in approach here is good as it doesn't affect existing usage perfomance wise (except for one conditional).


JohnLangford:
Why not get rid of all.interactions?

jackgerrits:
This allows us to minimally affect the performance when the interactions are not overridden

PR #1777, opened 2019-03-01
Parser now uses object and prod/cons queue and object pool for example parsing

### LINKED ISSUE #1739: 8.6.1: --cb_adf silently makes two examples when single example has > 255 actions
#### Describe the bug
In VW ver. 8.6.1, when using `--cb_adf` and specifying a multiline example with over 254 actions, actions # 255 and beyond are silently interpreted as a new example

#### To Reproduce
Data used: https://gist.github.com/maxpagels/d432e5e16fdaef899c58d6c0b6795c3d
Run: `vw test.vw --cb_adf --progress 1`

#### Expected behavior
If 254 actions is a hard limit, a warning would perhaps be a good idea?

#### Observed Behavior
The output of VW indicates that the multiline example is in fact interpreted as two examples:

```
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = test.vw
num sources = 1
    n.a.     n.a.            1            1.0  unknown        0:0...     1012
    n.a.     n.a.            2            2.0  unknown        0:0...        4

finished run
number of examples = 2
weighted label sum = 0.000000
average loss = n.a.
total feature number = 1016
```

Removing the last action from the input data ( resulting in 254 actions), works as expected:
```
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = test.vw
num sources = 1
    n.a.     n.a.            1            1.0  unknown        0:0...     1012

finished run
number of examples = 1
weighted label sum = 0.000000
average loss = n.a.
total feature number = 1012
```

#### Environment
OS X, VW 8.6.1, CLI

jackgerrits:
- This replaces the fixed size example ring previously used
- Therefore there is now no limit on the number of lines in a multi line example
- Fixes #1739
- VW now uses generic queue based object pool for example objects
- Parser offers ready passed examples in a FIFO producer/consumer queue with bounded size (ring size is used)
- Performance impact of this change has not yet been tested
- Simplifies logic around handling the example objects and passing to learner through these abstractions
- `used_index` is no longer relevant
- `being_parsed_examples` and `end_parsed_examples` haven't been removed yet as they are used to set the `example_counter` and hasn't been replaced
- I used inheritance to represent the interface of the object pool as I thought it made it clear. It should not impact performance due to us using the class directly and a good compiler will devirtualize

@lokitoth and I discussed the following issue: The theoretical memory bound of examples is still maintained with this PR. `ring_size` is still used to control the size of the queue used by the parser and learner. However, there is an issue where if a multi line example is used, there is no upper bound on the number of samples it can contain and so an attacker could use this to increase memory consumption. Currently in this scenario memory will not increase unbounded but the example will be split at ring size which is just incorrect behavior. Therefore, we should do the work to put a hard limit on multiline example size in a strict mode or something for these sensitive situations.

### DIFF
--- a/vowpalwabbit/object_pool.h
+++ b/vowpalwabbit/object_pool.h
@@ -5,90 +5,68 @@
 
 namespace VW
 {
-
-// Not thread safe
-template <typename T, typename TInitializer>
+template <typename T>
 struct object_pool
 {
-  object_pool() = default;
-  object_pool(size_t initial_chunk_size, size_t chunk_size = 8)
-    : m_initial_chunk_size(initial_chunk_size),
-    m_chunk_size(chunk_size)
+  virtual void return_object(T*) = 0;
+  virtual T* get_object() = 0;
+  virtual bool empty() const = 0;
+  virtual size_t available() const = 0;
+  virtual size_t size() const = 0;
+  virtual bool is_from_pool(T*) const = 0;
+
+  virtual ~object_pool() {}
+};
+
+// Not thread safe
+template <typename T, typename TFactory>
+struct unbounded_object_pool : object_pool<T>
+{
+  unbounded_object_pool() = default;
+  unbounded_object_pool(size_t initial_size)
   {
-    new_chunk(initial_chunk_size);
+    for (size_t i = 0; i < initial_size; i++)
+    {
+      auto obj = m_factory();
+      m_pool_objects.insert(obj);
+      m_pool.push(std::unique_ptr<T>(obj));
+    }
   }
 
-  void return_object(T* obj)
+  void return_object(T* obj) override
   {
     assert(is_from_pool(obj));
-    m_pool.push(obj);
+    m_pool.push(std::unique_ptr<T>(obj));
   }
 
-  T* get_object()
+  T* get_object() override
   {
+    T* obj;
     if (m_pool.empty())
     {
-      new_chunk(m_chunk_size);
+      obj = m_factory();
+      m_pool_objects.insert(obj);
+    }
+    else
+    {
+      obj = m_pool.front().release();
+      m_pool.pop();
     }
 
-    auto obj = m_pool.front();
-    m_pool.pop();
     return obj;
   }
 
-  bool empty() const { return m_pool.empty(); }
+  bool empty() const override { return m_pool.empty(); }
 
-  size_t available() const { return m_pool.size(); }
+  size_t available() const override { return m_pool.size(); }
 
-  size_t size() const {
-    size_t size = 0;
-    auto num_chunks = m_chunk_bounds.size();
+  size_t size() const override { return m_pool_objects.size(); }
 
-    if(m_chunk_bounds.size() > 0 && m_initial_chunk_size > 0)
-    {
-      size += m_initial_chunk_size;
-      num_chunks--;
-    }
-
-    size += num_chunks * m_chunk_size;
-    return size;
-  }
-
-  bool is_from_pool(T* obj) const {
-    for(auto& bound : m_chunk_bounds)
-    {
-      if(obj >= bound.first && obj <= bound.second)
-      {
-        return true;
-      }
-    }
-
-    return false;
-  }
+  bool is_from_pool(T* obj) const override { return m_pool_objects.count(obj) != 0; }
 
  private:
-  void new_chunk(size_t size)
-  {
-    if(size == 0)
-    {
-      return;
-    }
-
-    m_chunks.push_back(std::unique_ptr<T[]>(new T[size]));
-    auto& chunk = m_chunks.back();
-    m_chunk_bounds.push_back({&chunk[0], &chunk[size - 1]});
-
-    for (size_t i = 0; i < size; i++)
-    {
-      m_pool.push(m_initializer(&chunk[i]));
-    }
-  }
-
-  std::queue<T*> m_pool;
-  std::vector<std::pair<T*, T*>> m_chunk_bounds;
-  std::vector<std::unique_ptr<T[]>> m_chunks;
-  TInitializer m_initializer;
-  size_t m_initial_chunk_size = 0;
-  size_t m_chunk_size = 8;
+  std::queue<std::unique_ptr<T>> m_pool;
+  std::set<T*> m_pool_objects;
+  TFactory m_factory;
 };
 }  // namespace VW
--- a/vowpalwabbit/object_pool.h
+++ b/vowpalwabbit/object_pool.h
@@ -11,8 +11,9 @@ template <typename T, typename TInitializer>
 struct object_pool
 {
   object_pool() = default;
-  object_pool(size_t initial_chunk_size, size_t chunk_size = 8)
-    : m_initial_chunk_size(initial_chunk_size),
+  object_pool(size_t initial_chunk_size, TInitializer initializer = {}, size_t chunk_size = 8)
+    : m_initializer(initializer),
+    m_initial_chunk_size(initial_chunk_size),
     m_chunk_size(chunk_size)
   {
     new_chunk(initial_chunk_size);
--- a/vowpalwabbit/object_pool.h
+++ b/vowpalwabbit/object_pool.h
@@ -81,6 +81,9 @@ struct object_pool
 
     for (size_t i = 0; i < size; i++)
     {
+      // TODO fix need to memset to 0
+      memset(&chunk[i], 0, sizeof(T));
+      new (&chunk[i]) T{};
       m_pool.push(m_initializer(&chunk[i]));
     }
   }
--- a/vowpalwabbit/object_pool.h
+++ b/vowpalwabbit/object_pool.h
@@ -5,68 +5,90 @@
 
 namespace VW
 {
-template <typename T>
-struct object_pool
-{
-  virtual void return_object(T*) = 0;
-  virtual T* get_object() = 0;
-  virtual bool empty() const = 0;
-  virtual size_t available() const = 0;
-  virtual size_t size() const = 0;
-  virtual bool is_from_pool(T*) const = 0;
-
-  virtual ~object_pool() {}
-};
 
 // Not thread safe
-template <typename T, typename TFactory>
-struct unbounded_object_pool : object_pool<T>
+template <typename T, typename TInitializer>
+struct object_pool
 {
-  unbounded_object_pool() = default;
-  unbounded_object_pool(size_t initial_size)
+  object_pool() = default;
+  object_pool(size_t initial_chunk_size, size_t chunk_size = 8)
+    : m_initial_chunk_size(initial_chunk_size),
+    m_chunk_size(chunk_size)
   {
-    for (size_t i = 0; i < initial_size; i++)
-    {
-      auto obj = m_factory();
-      m_pool_objects.insert(obj);
-      m_pool.push(std::unique_ptr<T>(obj));
-    }
+    new_chunk(initial_chunk_size);
   }
 
-  void return_object(T* obj) override
+  void return_object(T* obj)
   {
     assert(is_from_pool(obj));
-    m_pool.push(std::unique_ptr<T>(obj));
+    m_pool.push(obj);
   }
 
-  T* get_object() override
+  T* get_object()
   {
-    T* obj;
     if (m_pool.empty())
     {
-      obj = m_factory();
-      m_pool_objects.insert(obj);
-    }
-    else
-    {
-      obj = m_pool.front().release();
-      m_pool.pop();
+      new_chunk(m_chunk_size);
     }
 
+    auto obj = m_pool.front();
+    m_pool.pop();
     return obj;
   }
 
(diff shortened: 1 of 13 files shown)


vowpalwabbit/object_pool.h
```diff
@@ -0,0 +1,72 @@
...
+// Not thread safe
+template <typename T, typename TFactory>
+struct unbounded_object_pool : object_pool<T>
+{
+  unbounded_object_pool() = default;
+  unbounded_object_pool(size_t initial_size)
+  {
+    for (size_t i = 0; i < initial_size; i++)
+    {
+      auto obj = m_factory();
```
kumpera: We should allocate this initial set of objects through `new T[]` to make sure they are contiguous and packed.

vowpalwabbit/object_pool.h
```diff
@@ -0,0 +1,72 @@
...
+    return obj;
+  }
+
+  bool empty() const override { return m_pool.empty(); }
+
+  size_t available() const override { return m_pool.size(); }
+
+  size_t size() const override { return m_pool_objects.size(); }
+
+  bool is_from_pool(T* obj) const override { return m_pool_objects.count(obj) != 0; }
```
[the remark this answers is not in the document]
jackgerrits: Yes, this is used elsewhere in the codebase to determine if an example was from the ring (now the object pool) and act accordingly

vowpalwabbit/object_pool.h
```diff
@@ -0,0 +1,72 @@
...
+
+ private:
+  std::queue<std::unique_ptr<T>> m_pool;
+  std::set<T*> m_pool_objects;
```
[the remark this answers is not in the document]
jackgerrits: Is there a different data structure you would suggest? Since values should be unique and we want at least log(n) lookup time. I could use std::unordered_set for average constant lookup.

This data structure us just used to maintain the list of all objects this pool manages, irrespective of if they are currently borrowed.

vowpalwabbit/object_pool.h
```diff
@@ -0,0 +1,72 @@
...
+  {
+    T* obj;
+    if (m_pool.empty())
+    {
+      obj = m_factory();
+      m_pool_objects.insert(obj);
+    }
+    else
+    {
+      obj = m_pool.front().release();
```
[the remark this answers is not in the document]
jackgerrits: Yeah, that could be good. I am a little hesitant about the behavior of the pool being destroyed and destroying objects that are currently in use. It seems like a weird situation though. What would you expect to happen in that case?

vowpalwabbit/object_pool.h
jackgerrits: Is it possible to mix the initial block initialization with the normal factory usage in a generic way that encompasses both? I agree that contiguous and packed would be optimal but not 100% sure how to get it to work

kumpera: If we allocate in chunks we can test membership by bounds checking against them.
We can have a fast path with data inline in the object pool itself with overflow using a std::vector of tuples.

I'll leave as an exercise to check when the constant factor cut in the linear cost due to chunking matches the cost of the logarithmic lookup of std::set.

jackgerrits: Can you explain more what you mean by a fast path with data inline using tuples?

We can get away from log lookup in the set by using an unordered set, and since we are storing pointers in it the hash is trivial

vowpalwabbit/parser.h
```diff
@@ -23,16 +23,33 @@ license as described in the file LICENSE.
...
+};
+
 struct parser
 {
   v_array<substring> channels;  // helper(s) for text parsing
   v_array<substring> words;
   v_array<substring> name;
 
+  std::mutex pool_lock;
+  std::unique_ptr<VW::unbounded_object_pool<example, example_factory>> example_pool;
```
JohnLangford: Why do we want a pointer to the example pool rather than the example pool?

vowpalwabbit/parser.h
```diff
@@ -23,16 +23,33 @@ license as described in the file LICENSE.
...
+  std::unique_ptr<VW::ptr_queue<example>> ready_parsed_examples;
```
JohnLangford: Again, why a pointer?  This seems like unnecessary boxing.

vowpalwabbit/queue.h
```diff
@@ -0,0 +1,86 @@
...
+namespace VW {
+  template <typename T>
+  class ptr_queue
+  {
+  public:
+    ptr_queue(size_t max_size)
+      : max_size(max_size)
+    {}
+
+    T* pop()
```
JohnLangford: Why pointer-not-reference?

JohnLangford:
High level comments:
1) Set is not an efficient datastructure because it involves pointer chasing.
2) Why don't we just have a vector of pointers to examples?
3) There seems to be some unnecessary boxing going on---that it would be cleaner to get rid of.
4) I don't understand the point of the inheritance here---it seems like a class without inheritance has all the same capabilities and is simpler.

vowpalwabbit/queue.h
jackgerrits: A reference implies the object is still managed by the pool, a pointer makes it clearer that is now free and must be returned

vowpalwabbit/parser.h
jackgerrits: The example pool must be constructed with the default size, so it would need to be done in the constructor. The constructor is currently hidden behind a `new_parser` method. While typing this out I am realizing it would be a good change to move this logic to a constructor. I'll add another commit for this

jackgerrits:
1. Good point, would unordered_set suffice?
2. To replace the queue? Queue is just an adapter over the container. To replace the set? I wanted the is_from_pool call to be better than linear over the objects in the pool. If you'd rather us just scan a vector we can do that.
3. I agree, I am working on removing the boxing
4. The inheritance here is simply a documentation/maintainability tool, and allows us to test different implementations easily. Functionally it has no effect though, so I am more than happy to remove.

JohnLangford:
This pull request **introduces 2 alerts** when merging 770d43ca57e372ef6dc781f2b0521bb6e16af639 into 112f9fe03101e7e3f64b731b033b2a6ffc3ebdc9 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-018ab21a32279e7630467112aad19f57af36a320)

**new alerts:**

* 2 for Resource not released in destructor

---

JohnLangford:
This pull request **introduces 2 alerts** when merging 721e1f6fb80c1ee5b59648667a54bf19ab90dda7 into 112f9fe03101e7e3f64b731b033b2a6ffc3ebdc9 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-31017c12024aca5a2994c275383e68257de6ee31)

**new alerts:**

* 2 for Resource not released in destructor

---

lokitoth:
I am a little concerned that this breaks an unstated rule that you could deterministically control how much memory vw would consume at peak. Now with uncapped example pool, this is no longer the case. We may want to consider adding a feature (out of scope for this PR, maybe) to re-enable that kind of hard-cap.

jackgerrits:
The hard cap is implemented through the queue the parser uses to pass the
examples through to the learner. If it is full then it will sit and wait
for the learner to consume some

jackgerrits:
It seems there may be a deadlock in daemon test that hits sometimes, I need to look into that

vowpalwabbit/object_pool.h
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    auto obj = m_pool.front();
+    m_pool.pop();
+    return obj;
+  }
+
+  bool empty() const { return m_pool.empty(); }
+
+  size_t available() const { return m_pool.size(); }
```
JohnLangford: Why have this function?   It seems conceptually unnecessary?

jackgerrits: You're right, it isn't really needed anymore. I had it around when it was fixed size

vowpalwabbit/object_pool.h
```diff
@@ -0,0 +1,98 @@
...
+      return;
+    }
+
+    m_chunks.push_back(std::unique_ptr<T[]>(new T[size]));
+    auto& chunk = m_chunks.back();
+    m_chunk_bounds.push_back({&chunk[0], &chunk[size - 1]});
+
+    for (size_t i = 0; i < size; i++)
+    {
+      // TODO fix need to memset to 0
```
JohnLangford: The comment can be removed, right?

vowpalwabbit/object_pool.h
```diff
@@ -0,0 +1,98 @@
...
+
+ private:
+  void new_chunk(size_t size)
+  {
+    if(size == 0)
+    {
+      return;
+    }
+
+    m_chunks.push_back(std::unique_ptr<T[]>(new T[size]));
```
JohnLangford: Why is unique_ptr necessary here?

vowpalwabbit/parser.cc
```diff
@@ -673,18 +662,11 @@ namespace VW
...
-      return ret;
-    }
-    else
-      p->example_unused.wait(lock);
-  }
+  std::unique_lock<std::mutex> lock(p->pool_lock);
+  auto ex = p->example_pool.get_object();
+  ex->in_use = true;
+  p->begin_parsed_examples++;
+  return *ex;
```
JohnLangford: I'm slightly worried about the semantics here.  What happens when the parser is faster than the learner as far as memory usage?

vowpalwabbit/parser.cc
```diff
@@ -1005,47 +960,35 @@ size_t get_feature_number(example* ec) { return ec->num_features; }
...
+  ex->in_use = false;
+  ex->passthrough = nullptr;
+  ex->tag = v_init<char>();
+  ex->indices = v_init<namespace_index>();
+  memset(&ex->feature_space, 0, sizeof(ex->feature_space));
+  return ex;
 }
 
-void adjust_used_index(vw& all) { all.p->used_index = all.p->begin_parsed_examples; }
+void adjust_used_index(vw& all) { /* no longer used */ }
```
JohnLangford: Can we just remove this function if it's a noop?

vowpalwabbit/parser.h
```diff
@@ -23,82 +23,101 @@ license as described in the file LICENSE.
...
 };
 
-parser* new_parser();
-
 void enable_sources(vw& all, bool quiet, size_t passes, input_options& input_options);
 
-bool examples_to_finish();
-
-// only call these from the library form:
+/* [[deprecated]] */
```
JohnLangford: Why not nuke?

vowpalwabbit/object_pool.h
jackgerrits: Yes

vowpalwabbit/object_pool.h
jackgerrits: So that memory is automatically freed for us on destruction. If we can avoid writing delete then it's a win :)

vowpalwabbit/parser.cc
jackgerrits: If the parser is faster then the queue will fill until the limit given (ring_size) and wait to add any more until some has been drained

vowpalwabbit/parser.cc
jackgerrits: Yes I'd like to, I just need to get around to it since it is used in quite a few bindings

vowpalwabbit/parser.h
jackgerrits: yes it should be

vowpalwabbit/object_pool.h
JohnLangford: I understand what unique_ptr does.  The cost (I believe) is that you end up double boxing things making every pointer deref extra painful.  Avoiding unnecessary boxing is one of the first things I look for in making code run faster.

vowpalwabbit/parser.cc
JohnLangford: Cool.

vowpalwabbit/object_pool.h
jackgerrits: These unique_ptrs are never accessed except for on creation of a chunk and at deletion time, so extra boxing shouldn't affect things here. The raw pointers are returned for normal use

JohnLangford: So you can remove dependence on unique_ptr and call delete explicitly without issues?  That seems like a plausible win in terms of reducing the amount of code.

jackgerrits: `delete[]` would need to be called on the memory block pointer which `unique_ptr` is managing. This class passes out pointers to individual pieces within this, so the delete cannot be done on these individual pointers.

jackgerrits:
@JohnLangford I believe I have no resolved the deadlock.

JohnLangford:
Merged, thanks :-)

PR #1779, opened 2019-03-01
Fix resetting of cb_type when using dm with cb_explore_adf

### LINKED ISSUE #1778: when using --cb_type dm (cb_explore_adf) warning of resetting to --cb_type ips is incorrect
#### To Reproduce
Data:
[vw_bug.txt](https://github.com/VowpalWabbit/vowpal_wabbit/files/2921313/vw_bug.txt)

#### Expected behavior
As VW is warning that:
warning: cb_type must be in {'ips','dr','mtr'}; resetting to ips.

I expect --cb_type ips to behave exactly the same as --cb_type dm

#### Observed Behavior
```
>vw --cb_explore_adf --epsilon 0.05 -l 0.5 --dsjson vw_bug.txt -P 1 --cb_type ips
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = vw_bug.txt
num sources = 1
-0.135135 -0.135135            1            1.0    known        0:0.111111...      205
-0.070946 -0.006757            2            2.0    known        0:0.955556...      205
-0.130631 -0.250000            3            3.0    known        2:0.955556...      205
-10.847973 -43.000000            4            4.0    known        4:0.955556...      205

finished run
number of examples = 4
weighted example sum = 4.000000
weighted label sum = 0.000000
average loss = -10.847973               <----------------------------------------
total feature number = 820

>vw --cb_explore_adf --epsilon 0.05 -l 0.5 --dsjson vw_bug.txt -P 1 --cb_type dm
warning: cb_type must be in {'ips','dr','mtr'}; resetting to ips.
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = vw_bug.txt
num sources = 1
-0.135135 -0.135135            1            1.0    known        0:0.111111...      205
-0.070946 -0.006757            2            2.0    known        0:0.955556...      205
-0.130631 -0.250000            3            3.0    known        2:0.955556...      205
-0.160473 -0.250000            4            4.0    known        0:0.955556...      205

finished run
number of examples = 4
weighted example sum = 4.000000
weighted label sum = 0.000000
average loss = -0.160473               <----------------------------------------
total feature number = 820
```
#### Environment
What version of VW did you use? 8.6.1
What OS or language did you use? Windows command line

jackgerrits:
Fixes #1778 

This issue was caused by the fact that the valid cb_type was being checked after the rest of the reduction stack was initialized. It needed to be done before so the default of "ips" could be used.

### DIFF
--- a/vowpalwabbit/cb_explore_adf.cc
+++ b/vowpalwabbit/cb_explore_adf.cc
@@ -807,13 +807,6 @@ base_learner* cb_explore_adf_setup(options_i& options, vw& all)
     data->explore_type = EPS_GREEDY;
   }
 
-  multi_learner* base = as_multiline(setup_base(options, all));
-  all.p->lp = CB::cb_label;
-  all.label_type = label_type::cb;
-
-  // Extract from lower level reductions.
-  data->gen_cs.scorer = all.scorer;
-  data->cs_ldf_learner = as_multiline(all.cost_sensitive);
   data->gen_cs.cb_type = CB_TYPE_IPS;
   if (options.was_supplied("cb_type"))
   {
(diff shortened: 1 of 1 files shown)


lokitoth:
Reviewed in-person, but here are the comments: Looks good overall; please add a description of the issue and fix to the body.

marco-rossi29:
This fix looks good to me. I was thinking about it and before merging it I'd like to understand what was return by --cb_type dm without this fix. Was that garbage or was that actually something? In the first case, we are all fine. In the later case, we should enable --cb_type dm.

jackgerrits:
That may be a regression from the options refactor. It would be good to confirm if this was in 8.6.1, I'll check that

jackgerrits:
This bug existed in 8.6.1, so we should wait and clarify with @JohnLangford before continuing.

JohnLangford:
There are three confusing things going on here. 
1) --cb_type dm is not enumerated here leading to ill-defined behavior.  I believe this was an oversight so the defined semantics is unclear.  In this sense the patch seems fine since it's helping define something ill-defined.
2) --cb_explore_adf should not need to know about the cb_type since that isn't _used_ except in the cb_adf reduction.  Hence, there is a modularity failure.  This modularity failure involves cover, mtr, and dealing with shared features which together make a mess.   The clean way forward here is to break things out into   simpler separate reductions which may have potentially different requirements on base.  I'll open an issue around this to keep track of it.

JohnLangford:
A question about this one: 
Previously, the call to base was above the processing of --cb_type because --cb_type was not defined until setup_base was called.   Is this changed with the new options parsing?

jackgerrits:
Prior to the options change we dug out the "cb_type" before it was defined in the underlying reduction, so there is a small change there. The bug still existed then as well, and we broke the modularity of layers at that point too

JohnLangford:
It's structurally messy to have the option defined in a place where it's not (conceptually at least) needed.  We should try to remove cb_type from here going forward.

Issue #1749, opened 2019-02-07
Fail to compiling a static executable vw on macOS


silverguo:
#### Describe the bug
Following the instruction [here](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Building), I fail to build vw statically on macOS

#### To Reproduce
1. Install all dependencies as mentioned [here](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Dependencies#macos), `brew install boost --with-python` doesn't work for me, I just did `brew install boost`

2. Following the instructions
```
mkdir build
cd build
cmake .. -DSTATIC_LINK_VW=ON
make vw-bin -j
```
#### Expected behavior
A static vw

#### Observed Behavior
```
[ 97%] Linking CXX static library libvw.a
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: libvw.a(parse_example_json.cc.o) has no symbols
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: libvw.a(parse_example_json.cc.o) has no symbols
[ 97%] Built target vw
Scanning dependencies of target vw-bin
[ 98%] Building CXX object vowpalwabbit/CMakeFiles/vw-bin.dir/main.cc.o
[100%] Linking CXX executable vw
ld: library not found for -lcrt0.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[3]: *** [vowpalwabbit/vw] Error 1
make[2]: *** [vowpalwabbit/CMakeFiles/vw-bin.dir/all] Error 2
make[1]: *** [vowpalwabbit/CMakeFiles/vw-bin.dir/rule] Error 2
make: *** [vw-bin] Error 2
```

#### Environment
- vw version: clone from master
- OS: macOS Mojave
- cmake: 3.13.4
- boost: 1.68.0

#### Additional context
I try to build static with lastest vw release 8.6.1 (Jul 21, 2018), but I didn't find the clear readme to build static vw on macOS for this version.

arielf:
I believe this is a more general (Linux as well) not just Mac OS-X issue.

I wasn't able to build statically since the switch to `cmake` building was made. Neither with `clang++` nor with `g++`.

Prior to that, building statically was simple (just add `-static` to `CXX`) and always worked for me.

Building statically is very important for more system-independent production deployments.

silverguo:
@arielf 
Thanks for your reply!

[...]

PR #1787, opened 2019-03-12
Allow spanning_tree to be a static executable

jackgerrits:
This addresses an issue that was brought up in #1749


JohnLangford:
Merged, thanks :-)

Issue #1746, opened 2019-02-06
Tests failures on master


kumpera:
#### Describe the bug
When running the test suite on OSX against master 6dd314d20dfde9211b2d614a8b13d387fd93778d I hit a few issues.

#### To Reproduce
1. `mkdir build; cd build; cmake -GNinja ..; ninja`
1. `cd build/test; ctest --verbose`

#### Expected behavior
All tests passes

#### Observed Behavior
The following tests FAILED:
	  1 - RunTests_pass_1_onethread (Failed)
	  2 - RunTests_pass_2 (Failed)
	  4 - test_race_condition (Failed)

#### Environment
Master 6dd314d20dfde9211b2d614a8b13d387fd93778d
OSX High Sierra
Release build

JohnLangford:
Is this a heisenbug or a reliable fail?

kumpera:
It's a reliable failure. I believe it happens on other OS', still confirming that.

JohnLangford:
That's a relief---it should be much easier to debug.

jackgerrits:
Can you provide the full test failure output - using `make test_with_output`

kumpera:
@jackgerrits https://gist.github.com/kumpera/94bd4727e198e5db88c233163e008830

kumpera:
Links to the failures:

https://gist.github.com/kumpera/94bd4727e198e5db88c233163e008830#file-gistfile1-txt-L183
https://gist.github.com/kumpera/94bd4727e198e5db88c233163e008830#file-gistfile1-txt-L461
```
1: RunTests: test 91: lenient-diff.tmp: line 1: fuzzy-match missing data on one of the sides. Can't compare
1: e^x*f^a:96492:-0.0611589						      <
1: RunTests: test 91: saving lenient diff in 'test-91.lenient-diff' for later inspection
1: --- diff -u --minimal pred-sets/ref/inv_hash.cmp inv_hash.cmp
1: --- pred-sets/ref/inv_hash.cmp	2019-01-30 20:14:23.000000000 -0500
```

https://gist.github.com/kumpera/94bd4727e198e5db88c233163e008830#file-gistfile1-txt-L618
```
test 4
    Start 4: test_race_condition

4: Test command: /anaconda3/bin/python "repeat.py" "1024" "/Users/kumpera/src/vowpal_wabbit/build/test/../vowpalwabbit/vw" "train-sets/rcv1_small.dat" "--holdout_after" "100"
[...]

PR #1788, opened 2019-03-13
[audit] Make sure --audit output is reproducible across systems.

kumpera:
Interactions output was using std::sort which triggered issues when values were the same.


kumpera:
This fixes a bug I found while working on #1746

JohnLangford:
This looks good, but appveyor is barfing on 176?

kumpera:
Investigating it next.

kumpera:
The appveyor failure was due to some technical debt in our test harness. I added an issue to track this: https://github.com/VowpalWabbit/vowpal_wabbit/issues/1803

PR #1459, opened 2018-04-20
C# exploration

eisber:

PR #1790, opened 2019-03-17
Add example C# Simulator

jackgerrits:
This is a copy/paste/clean of @marco-rossi29's C# simulator.

This uses the C# exploration lib in mwt-ds, and therefore is forced to use the VW nuget that it consumes. This project therefore doesn't target the cs project but the 8.4.1 nuget. 

Types that are missing:
- ActionProbability (Do we have this available in the C# projects here?)
- PRG
- GenericTopSlotExplorer

If we migrate these types to core VW then we can remove this dependency, I think this should be done soon.


vowpalwabbit/vw.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
```
jackgerrits: This retargets us to VS2017 instead of VS2015. Is anyone against this? The new style C# project structure requires this.

marco-rossi29: If it's working I'm very happy to upgrade to latest version

marco-rossi29:
In terms of removing the dependency from mwt-ds, I believe that we should have already in vw core equivalent routines (maybe with different signatures) that we can use right away, reducing the effort needed. For example, I'm quite sure that the behavior we need is probably similar to functions used in rl_client_lib.

To be clear, the goal of this simulator is to mirror what the latest Decision Service (or VW used in an online learning fashion) would do (with minor differences: learning and update model after every event, rather than exporting every X minutes, etc) so there is no need to use mwt-ds or any other legacy code. 

Therefore, I'm very happy to nuke unnecessary dependencies. Ideally I'd like something that is as simple and self contained as possible so that this can be used almost as a test to check with different VW version that no regression is introduced.

JohnLangford:
Why is appveyor failing?

jackgerrits:
The build script is picking up msbuild14, but the new project format for c# requires msbuild15. The last commit fixed the build environment, but I still need to work out why the script isn't picking that up.

jackgerrits:
Okay should be fixed with latest commit

jackgerrits:
This PR is blocked on #1459 so that the types can be consumed from this repo and not depend on nugets

PR #1791, opened 2019-03-18
Fix usage of char instead of unsigned char for namespace index

jackgerrits:
This caused problems when the api was used with namespaces >127


JohnLangford:
Looks good.  Merge when all tests pass.

PR #1793, opened 2019-03-19
Offset tree

rajan-chari:
1) learn() implementation for offset_tree
2) example test file
3) unit test


test/unit_test/offset_tree_tests.cc
```diff
@@ -47,6 +49,27 @@ using predictions_t = vector<pair<float, float>>;
...
+  VW::offset_tree::offset_tree tree;
+  tree.init(3);
+  example ec;
+  ec.pred.a_s = v_init<ACTION_SCORE::action_score>();
+  ec.l.cb = CB::label();
+  ec.l.cb.costs.push_back({-1.0,1,0.5 ,0.0});
+  
+  tree.learn(*as_singleline(test_harness), ec);
+  
+  destroy_free<test_learner_t>(test_harness);
```
jackgerrits: Why can't we simply call delete?

vowpalwabbit/offset_tree.cc
```diff
@@ -186,20 +199,65 @@ namespace VW { namespace offset_tree {
...
+
+  void learn(offset_tree& tree, single_learner& base, example& ec)
   {
-    // get predictions for all internal nodes in binary tree.
     ec.pred.a_s.clear();
-    auto scores = ot.predict(base, ec);
-    for (uint32_t idx = 0; idx < scores.size(); ++idx){
-      ec.pred.a_s.push_back({idx,scores[idx]});
-    }
+    static thread_local offset_tree::scores_t saved_scores;
```
jackgerrits: I'm skeptical about use of thread local, why can't this just be on the stack?

rajan-chari: vector allocates on the heap.  So we would have a heap allocation on every call.

test/unit_test/offset_tree_tests.cc
rajan-chari: This is allocated following standard VW idiom ie. calloc().  The proper way to dealloc is free().

vowpalwabbit/csoaa.cc
```diff
@@ -35,7 +35,7 @@ inline void inner_loop(single_learner& base, example& ec, uint32_t i, float cost
 {
   if (is_learn)
   {
-    ec.weight = (cost == FLT_MAX) ? 0.f : 1.f;
+    (cost == FLT_MAX) ? ec.weight = 0.0f : ec.weight = 1.0f;
```
JohnLangford: We should just remove this line instead, right?  Overriding the weight directly invalidates setting the weight at higher levels.

vowpalwabbit/offset_tree.cc
```diff
@@ -138,13 +142,22 @@ namespace VW { namespace offset_tree {
       return scores;
     }
 
+    // Save and clear example label.  
+    // (CB algs predict_or_learn checks for valid label during learn.)
+    static thread_local CB::label saved_label;
```
JohnLangford: This shouldn't (conceptually) be necessary since label should not affect prediction.  Can we instead fix whatever the issue is in cb_algs?

vowpalwabbit/offset_tree.cc
```diff
@@ -186,20 +199,65 @@ namespace VW { namespace offset_tree {
...
   }
+}
 
-  void predict(offset_tree& ot, single_learner& base, example& ec)
+void predict(offset_tree& ot, single_learner& base, example& ec)
+{
+  // get predictions for all internal nodes in binary tree.
+  ec.pred.a_s.clear();
+  const auto& scores = ot.predict(base, ec);
+  copy_to_action_scores(scores, ec.pred.a_s);
```
JohnLangford: The pattern where we have unnecessary copies seems undesirable from LOC and efficiency perspectives.   Maybe instead using the pattern where pass in a return reference is better?

jackgerrits:
@rajan-chari should this be closed? I know you have other work that is getting closer to a mergeable state?

PR #1794, opened 2019-03-19
Fix to bug 1729: no bias regularization for BFGS not working

### LINKED ISSUE #1729: no_bias_regularization not working
#### Describe the bug
The option --no_bias_regularization for BFGS doesn't seem to affect anything. This option was added here: https://github.com/VowpalWabbit/vowpal_wabbit/pull/1151 as a fix to this bug https://github.com/VowpalWabbit/vowpal_wabbit/issues/919 

#### To Reproduce
Here is an example dataset that shows it has no effect, and this dataset also shows the nature of why I want to use the option. This is a simple toy example and is predicting height (in feet) from gender and name. 
```
5.4 | Female Alice
5.8 | Male Bob
5.9 | Male Charlie
```

VW command with and without --no_bias_regularization appear to have the exact same coefficients. 

`vw -k -c --passes 20 --bfgs --holdout_off -d train.vw --readable_model readable.txt --l2 10 --no_bias_regularization`

#### Expected behavior
Given a high value of L2 regularization, I would like the intercept to converge toward the global averages (somewhere around 5+ feet) and the other coefficients to shrink toward zero. This can improve generalization to new feature values and to missing features. In the example above, this would predict a Male Jamie to be 3 feet tall, and a Jamie with unknown gender to be 2 feet tall. 

#### Observed Behavior
Nothing in the log indicates that --no_bias_regularization is being utilized.

#### Environment
VW 8.6.1 on Linux via command line

#### Additional context
I have included an Scikit-Learn example as a reference toward expected behavior where regularization is by default not applied to the intercept term. Doing a brief survey, it appears that Scikit-Learn, Spark MLLib and many other packages exclude the intercept from regularization by default. Setting it equal to the global mean (e.g. using 'base' input format) doesn't work either since the intercept in regression isn't always supposed to be the global mean. 

`train_data = pd.DataFrame({'Male': [0, 1, 1], 'Female':[1,0,0], 'Alice':[1,0,0],'Bob':[0,1,0], 'Charlie':[0,0,1], 'Height':[5.4,5.8,5.9]})
regr = linear_model.Ridge(alpha=100)  
regr.fit(train_data.drop(columns=['Height']), train_data['Height'])
print(regr.intercept_)`
>>> 5.699

pavithraks:
fixes #1729 

Cause of problem: weights[constant] always returns 0, so there was no subtraction happening at all with the below two lines:

`(&weights[constant])[W_GT] -= regularization * weights[constant]; ret -= 0.5 * regularization * (weights[constant]) * (weights[constant]);`

This is because we weren't accessing it the right way. 

`weights.strided_index(constant)` gives us what we want. Thus, the correct code should probably be:

`(&weights.strided_index(constant))[W_GT] -= regularization * (weights.strided_index(constant));
      ret -= 0.5*regularization*(weights.strided_index(constant))*(weights.strided_index(constant));`

### DIFF
--- a/vowpalwabbit/bfgs.cc
+++ b/vowpalwabbit/bfgs.cc
@@ -461,14 +461,14 @@ double add_regularization(vw& all, bfgs& b, float regularization, T& weights)
   {
     if (b.regularizers == nullptr)
     {
-      (&weights[constant])[W_GT] -= regularization * weights[constant];
-      ret -= 0.5 * regularization * (weights[constant]) * (weights[constant]);
+      (&weights.strided_index(constant))[W_GT] -= regularization * (weights.strided_index(constant));
+      ret -= 0.5 * regularization * (weights.strided_index(constant)) * (weights.strided_index(constant));
     }
     else
     {
       uint64_t i = constant >> weights.stride_shift();
-      weight delta_weight = weights[constant] - b.regularizers[2 * i + 1];
-      (&weights[constant])[W_GT] -= b.regularizers[2 * i] * delta_weight;
+      weight delta_weight = (weights.strided_index(constant)) - b.regularizers[2 * i + 1];
+      (&weights.strided_index(constant))[W_GT] -= b.regularizers[2 * i] * delta_weight;
       ret -= 0.5 * b.regularizers[2 * i] * delta_weight * delta_weight;
     }
   }
--- a/vowpalwabbit/bfgs.cc
+++ b/vowpalwabbit/bfgs.cc
@@ -457,6 +457,7 @@ double add_regularization(vw& all, bfgs& b, float regularization, T& weights)
     }
 
   // if we're not regularizing the intercept term, then subtract it off from the result above
+  // when accessing weights[constant], always use weights.strided_index(constant)
   if (all.no_bias)
   {
     if (b.regularizers == nullptr)


vowpalwabbit/bfgs.cc
```diff
@@ -457,18 +457,19 @@ double add_regularization(vw& all, bfgs& b, float regularization, T& weights)
...
 
   // if we're not regularizing the intercept term, then subtract it off from the result above
+  // when accessing weights[constant], always use weights.strided_index(constant)
   if (all.no_bias)
   {
     if (b.regularizers == nullptr)
     {
-      (&weights[constant])[W_GT] -= regularization * weights[constant];
-      ret -= 0.5 * regularization * (weights[constant]) * (weights[constant]);
+      (&weights.strided_index(constant))[W_GT] -= regularization * (weights.strided_index(constant));
```
jackgerrits: Is there a use case to NOT using strided_index? If not then perhaps we should make the `operator[]` private to avoid this bug in future?

pavithraks: I don't know vw well enough to comment on that but like we discussed in our conversation earlier today, it would be a big change, probably out of scope of this PR

jackgerrits: This is definitely out of scope of the PR and if we decide we want to look into this I can create and issue to track it

JohnLangford: Unfortunately, there is.  Currently, in setup_example all features are multipled by a stride which allows direct access to the parameters via [] as per https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/gd_predict.h#L33

JohnLangford:
Pavithra, thanks for this.  Let's merge as soon as tests pass.

vowpalwabbit/bfgs.cc
jackgerrits: So does offset correspond with the stride? Maybe the default should be to access using the stride and then explicitly state when you want unstrided? It depends on what the more common/intuitive usage is

JohnLangford: Not exactly.  

Stride controls floats per parameter.  Offset indexes into logically distinct sets of parameters.

PR #1795, opened 2019-03-20
Implement finish example and output for CCB

jackgerrits:


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -211,6 +212,160 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
...
+    std::ostringstream pred_buf;
+    pred_buf << std::setw(all.sd->col_current_predict) << std::right << std::setfill(' ') << pred_str;
+
+
+    all.sd->print_update(all.holdout_set_off, all.current_pass, label_buf.str(), pred_buf.str(), num_features,
+        all.progress_add, all.progress_arg);
+  }
+}
+
+inline float get_unbiased_cost(CCB::conditional_contexual_bandit_outcome* outcome, uint32_t action, float offset = 0.)
```
JohnLangford: Why can't we share this function with other implementations?

jackgerrits: CCB stores the cost and probability in it's own outcome structure and so we need a function that understands this. I can definitely move it to a common location though

jackgerrits: So the difference between CCB class and CB class is that CCB _can_ contain the entire pdf for that slot. So far that information has not been utilized on the reduction, so do we actually need it? If we don't then CCB can simply use the CB class

jackgerrits: @JohnLangford can you comment on the above? ^

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -211,6 +212,160 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
...
+    return (outcome->cost - offset) / outcome->probabilities[0].score;
+  return 0.;
+}
+
+void output_example(vw& all, ccb& /*c*/, multi_ex& ec_seq)
+{
+  if (ec_seq.size() <= 0)
+    return;
+
+  std::vector<example*> decisions;
```
kumpera: Should we reserve an initial size based on the size of  `ec_seq`?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -211,6 +212,160 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
...
+
+  // What does it mean for not all of the decisions to be labeled? Does it become a non-labeled example at that point?
+  // Is it hold out?
+  bool labeled_example = true;
+  auto preds = ec_seq[0]->pred.decision_scores;
+  for (int i = 0; i < decisions.size(); i++)
+  {
+    if (decisions[i]->l.conditional_contextual_bandit.outcome != nullptr)
+    {
+      float l = get_unbiased_cost(decisions[i]->l.conditional_contextual_bandit.outcome, preds[i][0].action);
```
kumpera: 0 is a cool number, but a constant would be amazing.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -211,6 +212,160 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
...
+    }
+    ss << '\n';
+    ssize_t len = ss.str().size();
+    ssize_t t = io_buf::write_file_or_socket(f, ss.str().c_str(), (unsigned int)len);
+    if (t != len)
+      std::cerr << "write error: " << strerror(errno) << std::endl;
+  }
+}
+
+void print_update(vw& all, std::vector<example*> decisions, decision_scores_t decision_scores, size_t num_features)
```
kumpera: Pass `decisions` byref

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: We don't know how many decisions exist in the sequence and therefore may overshoot significantly. I don't think it is worth reserving

PR #1796, opened 2019-03-20
Shared example decision feature injection and auto crossing

jackgerrits:
- Decision features are added to the shared example for to each CB call
- Decision and shared features are automatically crossed


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -131,14 +134,78 @@ void clear_pred_and_label(ccb& data)
...
+void inject_decision_features(example* shared, example* decision)
+{
+  for (auto index : decision->indices)
+  {
+    // constant namespace should be ignored, as it already exists and we don't want to double it up.
+    if (index == constant_namespace)
+    {
+      continue;
+    }
+    else if (index == 32)  // Decision default namespace has a special namespace in shared
```
kumpera: No magic numbers

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -131,14 +134,78 @@ void clear_pred_and_label(ccb& data)
...
+    }
+    else
+    {
+      LabelDict::del_example_namespace(*shared, static_cast<char>(index), decision->feature_space[index]);
+    }
+  }
+}
+
+// Generates all pairs of the namespaces for the two examples
+void calculate_and_insert_interactions(example* shared, example* decision, std::vector<std::string>& vec)
```
kumpera: We should reserve size in vec. It's `len(shared->indices)` * `len(decision->indices)` + `len(vec)`

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -185,12 +252,29 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
   sanity_checks<is_learn>(data);
   create_cb_labels(data);
 
+  std::vector<std::string> generated_interactions;
```
kumpera: The perf of this is iffy. It's fine to use as-is for now. Are we tracking perf work that we'll need to do later?

jackgerrits: I agree it is iffy, I don't have a tracking item for CCB perf yet.

PR #1797, opened 2019-03-20
Sample from PDF and swap chosen as part of CCB

jackgerrits:
In order for subsequent decisions to be valid sampling from the pdf must occur as part of the process.

Ultimately it makes sense for this to be a reduction itself but in the interest of unblocking CCB I am going to implement it here for now.


JohnLangford:
It does seem like it would be nice to make this into its own reduction. 

I'm also concerned with the random seed.  We need a way to controllably seed things on a per-example basis.  Maybe using the tag?  The difficult aspect here is that the tag is used for many things.

One earlier proposal never implemented was to allow any number of tags which would relieve this issue.

jackgerrits:
I agree this should be it's own reduction and was hoping to punt on that extra bit of work until later, but time has a way of slipping away. Rather it done now?

I figured I'd use the global random_seed as it would be reproducible. It seems that tag is often not set no? That would be a real issue if each example uses the same RNG state

JohnLangford:
A global random_seed is not that reproducible---think about 1000 different
frontends each operating off the same global random_seed.  In addition, I
think we need to give users direct control over the random seed on a per
event basis, as is done in CB.  There are situations where that control is
fairly important (think about randomization on a per user basis...).

It seems better to take the time to make a separate reduction.  It
shouldn't be much, right?  "draw a random action from this distribution" is
pretty straightforward.

-John

jackgerrits:
Okay that makes sense, what do you propose we use for the seed?

The reduction would essentially just do the sample and swap correct? We can only get it to work in the context of one label type - so I gues cb_label is the only one that makes sense.

JohnLangford:
Maybe... Parse the tag against the pattern "random_seed=<variable>" ?   We
can defer allowing a list of tags to support alternatives although that
will become necessary down this route.

-John

jackgerrits:
Are there any other options than in the tag? Seems like a weird overloading of that field?

JohnLangford:
We could potentially do something with the label type, but it feels even
odder to have a nontrivial label when testing.

Sticking nonfeature stuff in the features seems even odder, particularly
because we could not pass a string in that fashion.

-John

jackgerrits:
Messing with the label seems tricky I agree, tag is at least agnostic to label type. I think it's our best bet right now

jackgerrits:
@JohnLangford this has now been implemented as it's own reduction

PR #1801, opened 2019-03-21
expand Boost_LIBRARIES for pkg-config libs

jrmarkle:
With the latest cmake build process the installed file `libvw.pc` contains all the required boost libraries in a semicolon-separated list. This causes the output of `pkg-config --libs libvw` on Ubuntu 18.04 to be:
```
-L/usr/local/lib -lvw -lallreduce -pthread /usr/lib/x86_64-linux-gnu/libboost_program_options.so\;/usr/lib/x86_64-linux-gnu/libboost_system.so\;/usr/lib/x86_64-linux-gnu/libboost_thread.so\;/usr/lib/x86_64-linux-gnu/libboost_unit_test_framework.so\;/usr/lib/x86_64-linux-gnu/libboost_chrono.so\;/usr/lib/x86_64-linux-gnu/libboost_date_time.so\;/usr/lib/x86_64-linux-gnu/libboost_atomic.so\;/usr/lib/x86_64-linux-gnu/libpthread.so -lz
```
which is not useful for passing to the linker.

This change separates the library paths with spaces so that a command such as `g++ $(pkg-config --libs libvw) ...` can work.

There might be a better, more idiomatic cmake way to fix this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,6 +150,10 @@ endif()
 
 # PkgConfig
 if(VW_INSTALL)
+  FOREACH(BOOST_LIBRARY ${Boost_LIBRARIES})
+    SET (BOOST_LINK_LIBRARIES "${BOOST_LINK_LIBRARIES} ${BOOST_LIBRARY}")
+  ENDFOREACH()
+
   configure_file(libvw.pc.in libvw.pc)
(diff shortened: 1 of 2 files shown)


jackgerrits:
Thanks for this fix @jrmarkle. I'm not aware of a more idiomatic way to do this, so looks good.

JohnLangford:
Merged, thanks :-)

PR #1802, opened 2019-03-22
get rid of separated add_label calls

ataymano:


jackgerrits:
Since you are getting rid of these calls can you provide guidance on what the correct way to add a label is?

ataymano:
> Since you are getting rid of these calls can you provide guidance on what the correct way to add a label is?

Hm, it looks like having separate add_label requires additional documentation, while without it interface is more consistent with any other entry points and there is no specific instructions except generic input format guide: https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Input-format

Are you suggesting to leave this link in the header file?

jackgerrits:
Oh no that's fine if it just has to be part of the text input. Wasn't sure if there was a function to replace that.

vowpalwabbit/vwdll.cpp
```diff
@@ -219,17 +219,6 @@ VW_DLL_MEMBER size_t VW_CALLING_CONV VW_HashFeatureStaticA(const char * s, unsig
   return VW::hash_feature_static(str, u, hash, num_bits);
 }
 
-VW_DLL_MEMBER void  VW_CALLING_CONV VW_AddLabel(VW_EXAMPLE e, float label, float weight, float base)
```
lokitoth: This may be breaking API change. We should probably go through a deprecation cycle for it.

JohnLangford:
It seems like this patch is just removing add_label, but I'm unclear on what's motivating that.  Why do that?

PR #1804, opened 2019-03-22
Use std::stable_sort when computing interactions.

kumpera:
This fixes test 92 on OSX where it was producing a different set of interactions.


kumpera:
This is the last fix needed to get OSX green #1776

PR #1805, opened 2019-03-22
Extend CCB parser for Dsjson/Json to support implicitly converting CB examples to CCB examples

jackgerrits:
- dsjson/json is the only input type that will implicitly convert CB to CCB examples


vowpalwabbit/parse_example_json.h
```diff
@@ -1080,8 +1081,29 @@ class DecisionServiceState : public BaseState<audit>
...
+  BaseState<audit>* EndObject(Context<audit>& ctx, rapidjson::SizeType /* memberCount */) override
   {
+    // If we are in CCB mode and there have been no decisions. Check label cost, prob and action were passed. In that
+    // case this is CB, so generate a single decision with this info.
+    auto jsonp = static_cast<json_parser<audit>*>(ctx.all->p->jsonp.get());
+    if (jsonp->mode == json_parser_mode::ccb)
+    {
+      auto num_decisions = std::count_if(ctx.examples->begin(), ctx.examples->end(),
+          [](example* ex) { return ex->l.conditional_contextual_bandit.type == CCB::example_type::decision; });
+      if (num_decisions == 0 && ctx.label_object_state.found_cb)
```
ataymano: Maybe we should handle inconsistency (num_decisions != 0 && found_cb) somehow (throw / report error)?

vowpalwabbit/parse_example_json.h
```diff
@@ -1080,8 +1081,29 @@ class DecisionServiceState : public BaseState<audit>
...
+      {
+        ctx.ex = &(*ctx.example_factory)(ctx.example_factory_context);
+        ctx.all->p->lp.default_label(&ctx.ex->l);
```
lokitoth: [Nit] Consider changing the name of this field to parser->label_parser to make it clear what this is and does.

vowpalwabbit/parse_example_json.h
```diff
@@ -1382,7 +1402,7 @@ void read_line_decision_service_json(vw& all, v_array<example*>& examples, char*
   json_parser<audit>* parser = static_cast<json_parser<audit>*>(all.p->jsonp.get());
 
   // As long as VW was configured correctly this should not occur.
-  if(parser->mode == json_parser_mode::standard)
+  if (parser->mode == json_parser_mode::standard)
```
lokitoth: [Nit] `json_parser_mode::standard` seems to imply that we run parser in JSON "Standard" and JSON "non-standard" mode, where the real selection is between non-*CB parsing and *CB parsing.

vowpalwabbit/parse_example_json.h
jackgerrits: found_cb is actually set in the ccb path, it is set for and _Label_... field that is encountered. I agree it's not great but is the case because of the natural overlap

PR #1807, opened 2019-03-22
[OSX] Improve support for static linking.

kumpera:
On OSX, it's not possible to statically link against system libraries as they are closed source and Apple doesn't provide `.a` versions of them.

So we do the best thing we can given those constraints. We fix the build and eliminate all dependencies we can.
The resulting binaries should work on factory reset OSX machines, which is as good as we can aim for.


kumpera:
This fixes #1749

We must update the wiki page to include `brew install zlib` on the OSX instructions if this is merged.

PR #1816, opened 2019-04-01
Conditional Contextual Bandit

jackgerrits:
Opening a draft PR for review and comment.

- Implements ccb_explore_adf reduction
- Implements CCB format for text, json and dsjson format

~~I want to point out that we are 100% leaking memory in CCB. Primarily around labels and predictions. I'll need to do a pass fixing them at some point.~~ Leaks should be fixed

Still left to do:
- [x] Remove random sampling from learn branch


yannstad:
I already reviewed and approved this code before this PR was submitted. Also, I wrote some part of it, so reviewing again is not helpful.
Can someone else have a look? @rajan-chari @kumpera @ataymano ?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,785 @@
...
+
+
+struct ccb
+{
+  example* shared;
+  std::vector<example*> actions, decisions;
+  uint32_t chosen_action_index;
+  std::unordered_map<uint32_t, uint32_t> origin_index;
+  CB::cb_class cb_label, default_cb_label;
+  std::unordered_set<uint32_t> exclude_list, include_list;
```
kumpera: Given the values in this set are known to be in the `[0 - data.actions.size() [`  range, this should be a bit-field such as ` std::vector<bool>`.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,785 @@
...
+using namespace VW;
+using namespace VW::config;
+
+
+struct ccb
+{
+  example* shared;
+  std::vector<example*> actions, decisions;
+  uint32_t chosen_action_index;
+  std::unordered_map<uint32_t, uint32_t> origin_index;
```
kumpera: Given the key to this map will be a number known to be in the range `[0 - data.actions.size() [`  we can use a fully associative type instead. This should be a `std::vector<uint32_t>`.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,785 @@
...
+
+namespace CCB
+{
+static constexpr uint32_t SHARED_EX_INDEX = 0;
+static constexpr uint32_t TOP_ACTION_INDEX = 0;
+
+void clear_all(ccb& data)
+{
+  data.shared = nullptr;
+  data.actions.clear();
```
kumpera: This could be a problem if we're dealing with a large enough set of actions/decisions that calling `clear` actually relocates the underlying array.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,785 @@
...
+  *(uint8_t*)c = static_cast<uint8_t>(ld->type);
+  c += sizeof(uint8_t);
+
+  *(bool*)c = ld->outcome != nullptr;
+  c += sizeof(bool);
+
+  if (ld->outcome != nullptr)
+  {
+    *(float*)c = ld->outcome->cost;
+    c += sizeof(float);
```
kumpera: This should be `sizeof(ld->outcome->cost)`

jackgerrits:
Had to unmark it is as draft so appveyor would build it

vowpalwabbit/cache.h
```diff
@@ -16,3 +16,18 @@ void cache_tag(io_buf& cache, v_array<char> tag);
...
+
+namespace VW
+{
+  template <typename T>
+  T read_object(io_buf& cache)
+  {
+    char* c;
+    size_t next_read_size = sizeof(T);
+    if (cache.buf_read(c, next_read_size) < next_read_size)
+      THROW("Failed to read CCB label cache");
```
JohnLangford: Maybe we should have a return reference?  The template will work better and we can return the number of bytes read (effectively, an error code) rather than throwing.

vowpalwabbit/cb_algs.h
```diff
@@ -69,6 +69,13 @@ inline float get_unbiased_cost(CB::cb_class* observation, COST_SENSITIVE::label&
   return get_unbiased_cost(observation, action);
 }
 
+inline float get_unbiased_cost(ACTION_SCORE::action_score& a_s, float cost, uint32_t action, float offset = 0.)
+{
```
JohnLangford: The name of this function doesn't make sense unless offset = 0.

vowpalwabbit/cb_sample.cc
```diff
@@ -0,0 +1,62 @@
...
+
+  auto action_scores = examples[0]->pred.a_s;
+
+  uint64_t seed = data.random_seed;
+  if (examples[0]->tag.size() > 0)
+  {
+    if (strncmp(examples[0]->tag.begin(), "seed=", 5) == 0)
+    {
+      substring tag_seed{examples[0]->tag.begin() + 5, examples[0]->tag.begin() + examples[0]->tag.size()};
+      seed = static_cast<uint64_t>(int_of_substring(tag_seed));
```
JohnLangford: The seed need not be an int.  We should be doing exactly the same manipulations as in the exploration library so the seed is used in the same way.

vowpalwabbit/cb_sample.cc
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  // Sampling is done after the base learner has generated a pdf.
+  uint32_t chosen_action;
+  if (exploration::sample_after_normalizing(seed + data.random_seed_counter++,
+          ACTION_SCORE::begin_scores(action_scores), ACTION_SCORE::end_scores(action_scores), chosen_action))
+    THROW("Failed to sample from pdf");
+
+  if (exploration::swap_chosen(action_scores.begin(), action_scores.end(), chosen_action))
+    THROW("Failed to swap top action_score chosen during sampling");
```
JohnLangford: Failure to swap seems like a silly reason to throw an exception.  Can we avoid this?

vowpalwabbit/cb_sample.cc
```diff
@@ -0,0 +1,62 @@
...
+
+base_learner* cb_sample_setup(options_i& options, vw& all)
+{
+  auto data = scoped_calloc_or_throw<cb_sample_data>();
+  bool cb_sample_option = false;
+  uint64_t override_seed;
+
+  option_group_definition new_options("CB Sample");
+  new_options.add(make_option("cb_sample", cb_sample_option).keep().help("Sample from CB pdf and swap top action."));
+  new_options.add(make_option("cb_sample_seed", override_seed)
```
JohnLangford: I'm not following why this is needed.  We can already control the overall seed in VW.

vowpalwabbit/example.h
```diff
@@ -85,6 +90,9 @@ struct example : public example_predict  // core example datatype.
   bool end_pass;  // special example indicating end of pass.
   bool sorted;    // Are the features sorted or not?
   bool in_use;    // in use or not (for the parser)
+
+  // Interactions can be overriden on a per example basis by setting this pointer.
+  std::vector<std::string>* override_interactions;
```
JohnLangford: I believe this should just always be used so we only have one mechanism for specifying interactions.

vowpalwabbit/parser.cc
```diff
@@ -541,21 +541,31 @@ void enable_sources(vw& all, bool quiet, size_t passes, input_options& input_opt
 
       if (input_options.json || input_options.dsjson)
       {
+        auto mode = json_parser_mode::standard;
+        if (all.label_type == label_type::cb)
```
JohnLangford: Where are we at in trying to project cb into ccb automatically?

vowpalwabbit/parse_args.cc
```diff
@@ -1249,7 +1250,7 @@ void parse_reductions(options_i& options, vw& all)
   all.reduction_stack.push(lrqfa_setup);
   all.reduction_stack.push(stagewise_poly_setup);
   all.reduction_stack.push(scorer_setup);
-  // Reductions
+  // ReductionsS
```
JohnLangford: Typo

vowpalwabbit/conditional_contextual_bandit.h
```diff
@@ -0,0 +1,53 @@
...
+    shared = 1,
+    action = 2,
+    decision = 3
+  };
+
+  struct label {
+    example_type type;
+    // Outcome may be unset.
+    conditional_contexual_bandit_outcome* outcome;
+    v_array<uint32_t> explicit_included_actions;
```
JohnLangford: What is this exactly?

vowpalwabbit/conditional_contextual_bandit.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include "label_parser.h"
+#include "v_array.h"
+#include "action_score.h"
+#include "options.h"
+
+
+namespace LEARNER {
+  template<class T, class E> struct learner;
+  using base_learner = learner<char, char>;
```
JohnLangford: What is this?

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: The size of the vector should be sufficiently small for it not to matter here, but I agree this is not great. Would be good if there was a way to clear it and guarantee the memory is not deallocated

test/RunTests
```diff
@@ -1656,35 +1656,39 @@ echo "1 | feature:1" | {VW} -a --initial_weight 0.1 --initial_t 0.3
...
 
-# Test 184 warm_cb warm start with CYC supervised corruption
+# Test 185 warm_cb warm start with CYC supervised corruption
 {VW} --warm_cb 10 --cb_explore_adf --cb_type mtr --epsilon 0.05 --warm_start 3 --interaction 7 --choices_lambda 8 --warm_start_update --interaction_update --corrupt_type_warm_start 2 --corrupt_prob_warm_start 0.5 -d train-sets/multiclass
     train-sets/ref/warm_cb_cyc.stderr
 
-# Test 185 warm_cb warm start with input cost-sensitive examples
+# Test 186 warm_cb warm start with input cost-sensitive examples
 {VW} --warm_cb 3 --cb_explore_adf --cb_type mtr --epsilon 0.05 --warm_start 1 --interaction 2 --choices_lambda 8 --warm_start_update --interaction_update --warm_cb_cs -d train-sets/cs_cb
     train-sets/ref/warm_cb_cs.stderr
```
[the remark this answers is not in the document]
jackgerrits: We definitely are, this is on the todo

vowpalwabbit/cache.cc
```diff
@@ -209,3 +209,12 @@ void cache_features(io_buf& cache, example* ae, uint64_t mask)
 
   for (namespace_index ns : ae->indices) output_features(cache, ns, ae->feature_space[ns], mask);
 }
+
+uint32_t VW::convert(size_t number)
+{
```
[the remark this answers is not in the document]
jackgerrits: The symbol is exposed in the header. It is used in the `cache_label` function in conditional_contextual_bandit.cc

vowpalwabbit/cache.h
jackgerrits: The pollutes usage of this function a lot. Can we actually recover by knowing number of bytes read if we fail reading at this point?

vowpalwabbit/cb_algs.h
jackgerrits: I simply copied from similar implementations above, what would you suggest?

vowpalwabbit/example.h
jackgerrits: I agree, but the way to do it properly us quickly becoming out of scope in the timeframe we have left for CCB. Would it be acceptable to make this a follow up issue?

vowpalwabbit/parser.cc
jackgerrits: This is done in the json parser. Only JSON can be projected

vowpalwabbit/conditional_contextual_bandit.h
jackgerrits: Forward declaration due to headers being included in a weird order

vowpalwabbit/conditional_contextual_bandit.h
jackgerrits: That is for specifying the set of actions that are valid for each slot but it seems that that portion is being cut for now

vowpalwabbit/cache.cc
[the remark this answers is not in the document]
JohnLangford: Cool

vowpalwabbit/cache.h
JohnLangford: The cache code is generally setup so that failure to read => end of cache => end pass.  Letting an error create this semantics seems reasonable?

vowpalwabbit/cb_algs.h
JohnLangford: Maybe "cost_estimate"?

vowpalwabbit/example.h
JohnLangford: I'd rather not, for two reasons.

The first is that I believe this can fixed with a single line of code in the example initializer.

The second is just general aversion to incurring additoinal technical debt.

jackgerrits: For the single line I assume you mean copying the command line provided interactions into this per-example store? I agree, this is easy and I'm totally happy to do that. My concerns is this incurs a copy for every example, CCB or not and regardless if the reduction stack wishes to mutate the interactions.

vowpalwabbit/cb_sample.cc
jackgerrits: I agree it doesn't make sense. But we need to explicitly handle all error that may come out of this function.

vowpalwabbit/cb_sample.cc
jackgerrits: Since we control exploration in this reduction, we need to provide a mechanism to specify the seed for reproducability. Up until now the sampling was done externally with a different RNG, this just maintains that functionality.

vowpalwabbit/cb_sample.cc
jackgerrits: I'll take a look at the explore lib and update this

vowpalwabbit/cb_algs.h
jackgerrits: For the new function? Or to rename existing usage too?

vowpalwabbit/cache.h
jackgerrits: Sounds good, I'll fix that up

vowpalwabbit/example.h
JohnLangford: No copy---just have have a pointer and declare the memory at that address not owned by the example.

vowpalwabbit/cb_sample.cc
JohnLangford: Since this is on the example path, we shouldn't use an exception. 

Alternatives: 
1) Prove that the function can't err.
2) Define a default semantics when it errs.

vowpalwabbit/cb_sample.cc
JohnLangford: This however is inherently the wrong seed since it is not defined on a per-example basis.  

We want our seed to be handled exactly the same way as with the exploration library.

vowpalwabbit/cb_algs.h
JohnLangford: It seems more consistent to do both.

vowpalwabbit/cache.h
jackgerrits: Essentially all of the helpfulness of this function is lost when the return needs to be an out param (see below). To be fair though the semantics do change a bit. Since the code path isn't setup to handle exceptions to indicate that the cache file has ended or there is a failure then it seems like the function I proposed is not really an option.

```c++
auto size_probs = read_object<uint32_t>(cache);
read_count += sizeof(size_probs);

```
---
```c++
uint32_t size_probs;
if(read_object(cache, size_probs) < sizeof(size_probs))
  return 0;
read_count += sizeof(size_probs);
```

JohnLangford: I see.   I'm not sure what's better than the last.  Maybe a macro?  Or maybe just give up on read_object.

jackgerrits: Yeah I think I'll give up on read_object and just stick with the current convention. It didn't reduce code that much anyway

vowpalwabbit/cb_algs.h
jackgerrits: Sounds good

vowpalwabbit/cb_sample.cc
jackgerrits: While I agree we need to be careful on the example path and this exception shouldn't be here. I think there are always going to be functions that can fail, and I don't think there is always a way to prevent that. 

I this case I believe a failure indicates it is totally broken, should we simply panic?

vowpalwabbit/cb_sample.cc
jackgerrits: In that case the only options available are to specify the seed in the tag, and if that is not present we fall back to the global seed?

JohnLangford: Sounds good.

vowpalwabbit/example.h
jackgerrits: For some reason I had it in my mind that there would be several concurrent usages of such a piece of memory, but thinking about it now I don't think that's the case.

Then as long as the example puts it back how it found it we are good. I'll implement this.

vowpalwabbit/cb_sample.cc
JohnLangford: Proving that we can't fail and then ignoring the possibility of failure seems more clean/efficient?

vowpalwabbit/example.h
JohnLangford: I can't quite tell if this is what you have in mind but using COW semantics seems plausibly sanest.  By default, reductions don't touch this field.  Any that do should create their own interactions vector and change the example to point to it, then restore the pointer to the default interactions when finishing.

vowpalwabbit/cb_sample.cc
jackgerrits: Proving we can't fail seems a bit tricky, but the only sensible course of action here based on our policy.

jackgerrits: It still feels like a bug waiting to happen if we ignore error codes.

vowpalwabbit/example.h
jackgerrits: COW was the sort of thing I had in mind too. I think we are on the same page here.

vowpalwabbit/cb_sample.cc
jackgerrits: We do other sanity checks in this file and throw if the invariants are not met, therefore I think leaving these here would be reasonable. They are a similar invariant check, if they are violated we cannot proceed. Thoughts?

JohnLangford: Maybe just use assert?  That way it has no runtime impact.

The general policy is no exceptions on the example handling path, which this certainly is.

jackgerrits: Assert is reasonable, I like that

JohnLangford:
Why is appveyor failing for this one?

test/RunTests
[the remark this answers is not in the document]
JohnLangford: Still missing...

vowpalwabbit/gd.h
```diff
@@ -75,9 +75,9 @@ inline void foreach_feature(vw& all, example& ec, R& dat)
 {
   return all.weights.sparse
       ? foreach_feature<R, S, T, sparse_parameters>(all.weights.sparse_weights, all.ignore_some_linear,
-            all.ignore_linear, all.interactions, all.permutations, ec, dat)
+            all.ignore_linear, ec.interactions ? *ec.interactions : all.interactions, all.permutations, ec, dat)
```
JohnLangford: Let's remove the 'if'.

vowpalwabbit/global_data.h
```diff
@@ -534,8 +535,12 @@ struct vw
   size_t passes_complete;
   uint64_t parse_mask;  // 1 << num_bits -1
   bool permutations;    // if true - permutations of features generated instead of simple combinations. false by default
+
+  // Referenced by examples as their set of interactions. Can be overriden by reductions.
   std::vector<std::string> interactions;
+  // TODO #1863 deprecate in favor of only interactions field.
```
JohnLangford: Can this be done in a computationally efficient manner?

jackgerrits: Which bit? I don't follow

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,907 @@
...
+
+  // Stash the CCB labels before rewriting them.
+  data.stored_labels.clear();
+  for (auto ex : examples)
+  {
+    data.stored_labels.push_back({ex->l.conditional_contextual_bandit.type,
+        ex->l.conditional_contextual_bandit.outcome, ex->l.conditional_contextual_bandit.explicit_included_actions});
+  }
+
+  // Shared example must include history namespace:
```
JohnLangford: Nuke?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,907 @@
...
+  if (std::find(data.shared->indices.begin(), data.shared->indices.end(), ccb_history_namespace) ==
+      data.shared->indices.end())
+  {
+    data.shared->indices.push_back(ccb_history_namespace);
+  }
+  data.shared->feature_space[ccb_history_namespace].push_back(1, constant);
+
+  // This will overwrite the labels with CB.
+  create_cb_labels(data);
```
JohnLangford: Uses v_init -> invalid?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,907 @@
...
+
+  // Reset exclusion list for this example.
+  data.exclude_list.assign(data.actions.size(), false);
+
+  auto decision_scores = v_init<ACTION_SCORE::action_scores>();
```
JohnLangford: Causes memory allocation.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,907 @@
...
+    {
+      data.generated_interactions.clear();
+      // TODO be more efficient here
+      std::copy(data.original_interactions->begin(), data.original_interactions->end(),
+          std::back_inserter(data.generated_interactions));
+      // TODO currently all action namespaces are used adding redundant interactions.
+      calculate_and_insert_interactions(data.shared, slot, data.actions, data.generated_interactions);
+      size_t removed_cnt;
+      size_t sorted_cnt;
+      INTERACTIONS::sort_and_filter_duplicate_interactions(data.generated_interactions, true, removed_cnt, sorted_cnt);
```
JohnLangford: Why do we need to call this here?  Can't we just guarantee that there are no duplicate interactions directly?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,907 @@
...
+
+    if (has_action(cb_ex))
+    {  // the cb example contains at least 1 action
+      multiline_learn_or_predict<is_learn>(base, cb_ex, examples[0]->ft_offset);
+      save_action_scores(data, decision_scores);
+      clear_pred_and_label(data);
+    }
+    else
+    {  // the cb example contains no action => cannot decide
+      auto empty_action_scores = v_init<ACTION_SCORE::action_score>();
```
JohnLangford: Invalid, causes memory allocation.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,907 @@
...
+    {
+      ex->interactions = data.original_interactions;
+    }
+    remove_slot_features(data.shared, slot);
+
+    // Put back the original shared example tag.
+    std::swap(data.shared->tag, slot->tag);
+  }
+
+  delete_cb_labels(data);
```
JohnLangford: memory deallocation...

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,907 @@
...
+    num_features += ec->num_features;
+
+    if (ec->l.conditional_contextual_bandit.type == CCB::example_type::slot)
+    {
+      slots.push_back(ec);
+    }
+  }
+
+  // What does it mean for not all of the slots to be labeled? Does it become a non-labeled example at that point?
+  // Is it hold out?
```
JohnLangford: Having only a subset of examples labeled seems like a malformed example.  Maybe test and warn?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,907 @@
...
+{
+  if (ec_seq.size() > 0)
+  {
+    output_example(all, data, ec_seq);
+    CB_ADF::global_print_newline(all);
+  }
+
+  for (auto a_s : ec_seq[0]->pred.decision_scores)
+  {
+    a_s.delete_v();
```
JohnLangford: Should not be here.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,907 @@
...
+
+  learner<ccb, multi_ex>& l =
+      init_learner(data, base, learn_or_predict<true>, learn_or_predict<false>, 1, prediction_type::decision_probs);
+
+  l.set_finish_example(finish_multiline_example);
+  l.set_finish(CCB::finish);
+  return make_base(l);
+}
+
+size_t read_cached_label(shared_data*, void* v, io_buf& cache)
```
JohnLangford: Maybe we should separate the ccb label type stuff from the ccb reduction?  That seems conceptually sound and agrees with how other reductions are done.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,907 @@
...
+  if (!substring_equal(words[0], "ccb"))
+  {
+    THROW("ccb labels require the first word to be ccb");
+  }
+
+  auto type = words[1];
+  if (substring_equal(type, "shared"))
+  {
+    if (words.size() > 2)
+      THROW("shared labels may not have a cost");
```
JohnLangford: Should we be throwing or warning?

vowpalwabbit/global_data.h
JohnLangford: Shifting quadratics and cubics into interactions.  That was nontrivial the first time, which is why it did not happen.

jackgerrits: Ah I see, I am not sure of the efficiencies of being able to do this. I more mean to maintain only a single vector rather than 3. When generating the interactions, we can treat the cases of size 2,3 and n differently to maintain the current efficiencies.

Having pairs, triples and interactions kind of leaks this information and I assume is there from before we could handle arbitrary length?

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: They are separate, just happen to be in the same file. I agree, this file is messy because of it. Will move

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: This was a first pass attempt, the logic is simpler but less efficient. To do this without sort and filter, we would need to binary search to find if it exists and then insert if it does'nt. So it would be nlogn, unless we make a pass and build a map of indices then perhaps we could do it in n but with extra memory. I assume sort+filter is nlogn, so I figured it's not too bad to fall back to this way.

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: Okay, can warn about this

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: I had quite a bit of trouble with leaks when I was reusing memory, so I decided to use alloc+dealloc. I can take another pass at reducing this

vowpalwabbit/global_data.h
JohnLangford: They are, yes.

vowpalwabbit/cb_sample.cc
```diff
@@ -0,0 +1,97 @@
...
+  {
+    if(examples[i]->l.cb.costs.size() > 0)
+    {
+      // Must remove 1 because of shared example index.
+      labelled_action = i - 1;
+    }
+  }
+
+  // If we are learning and have a label, then take that action as the chosen action. Otherwise sample the distribution.
+  if (is_learn && labelled_action != -1)
```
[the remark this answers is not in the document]
jackgerrits: When sampling in the learn pipeline, we don't actually want to sample but choose the thing that was labelled/chosen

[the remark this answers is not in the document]
JohnLangford: So, this should be an opportunity to check reproducibility, yes?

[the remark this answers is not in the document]
jackgerrits: Yes it is, but the rng state needs to be the same for both calls. If a predict is followed by a learn, the rng will be advanced and so the draw will be different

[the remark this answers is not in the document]
JohnLangford: Sure, but we should be trying to do that.  Can we add the check?

test/RunTests
```diff
@@ -1657,43 +1657,47 @@ echo "1 | feature:1" | {VW} -a --initial_weight 0.1 --initial_t 0.3
...
 
-# Test 186 test counting examples with holdout_after option
+# Test 187 test counting examples with holdout_after option
 {VW} -k -P 100 --holdout_after 500 -d train-sets/0002.dat
     train-sets/ref/holdout_after.stderr
 
-# Test 187 test counting examples with holdout_after option with 2 passes on the training set
+# Test 188 test counting examples with holdout_after option with 2 passes on the training set
```
yannstad: There are 2 tests # 188. This is breaking the appveyor build apparently.

jackgerrits: Thanks, I need to update the test numbers to fix this. Bad merge

test/RunTests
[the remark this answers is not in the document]
jackgerrits: I've been waiting to settle on the learning algorithm before adding the test

vowpalwabbit/cb_sample.cc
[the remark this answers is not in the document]
jackgerrits: I'm not sure what you mean by adding the check?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,907 @@
...
+// Copy anything in default namespace for slot to ccb_slot_namespace in shared
+// Copy other slot namespaces to shared
+void inject_slot_features(example* shared, example* slot)
+{
+  for (auto index : slot->indices)
+  {
+    // constant namespace should be ignored, as it already exists and we don't want to double it up.
+    if (index == constant_namespace)
+    {
+      continue;
```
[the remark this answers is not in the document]
jackgerrits: I don't think injecting the constant namespace makes sense? It is the same for every example.

vowpalwabbit/interactions.cc
```diff
@@ -237,7 +237,7 @@ void eval_count_of_generated_ft(vw& all, example& ec, size_t& new_features_cnt,
   if (all.permutations)
   {
     // just multiply precomputed values for all namespaces
-    for (std::string& inter : all.interactions)
+    for (std::string& inter : ec.interactions ? *ec.interactions : all.interactions)
```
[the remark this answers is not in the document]
jackgerrits: So there's a tricky problem here - there are quite a few places where reductions (nn, search) create examples and use them but don't actually pass them through the setup_example function. So when they get here the interaction pointer will be `nullptr`. I had a go at ensuring it is initialized but often in search I don't have access to the all object to get the interaction pointer. Not sure what the right way forward with this is.

vowpalwabbit/cb_sample.cc
```diff
@@ -0,0 +1,97 @@
...
+  auto action_scores = examples[0]->pred.a_s;
+  uint32_t chosen_action;
+
+  int labelled_action = -1;
+  // Find that chosen action in the learning case, skip the shared example.
+  for(size_t i = 1; i < examples.size(); i++)
+  {
+    if(examples[i]->l.cb.costs.size() > 0)
+    {
+      // Must remove 1 because of shared example index.
```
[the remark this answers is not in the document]
jackgerrits: We cannot because the probabilities used to sample are stored in the shared example.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,907 @@
...
+        THROW("ccb_adf_explore: badly formatted example - invalid example type");
+    }
+  }
+}
+
+template <bool is_learn>
+void sanity_checks(ccb& data)
+{
+  if (data.slots.size() > data.actions.size())
+    THROW("ccb_adf_explore: badly formatted example - number of actions "
```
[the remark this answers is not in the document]
jackgerrits: The only alternative we have right now is to put an assert in? Does that seem reasonable?

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: I really do think we should throw here. The data is malformed and we should not continue in an undefined state

jackgerrits:
@JohnLangford removing the if for interactions is still blocked on Search using non-setup examples. Ideas?

vowpalwabbit/cb_sample.cc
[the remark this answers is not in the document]
jackgerrits: This check is still required for safety, as if -t wasn't passed then this may cause a crash.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,666 @@
...
+        break;
+      case example_type::action:
+        data.actions.push_back(ex);
+        break;
+      case example_type::slot:
+        data.slots.push_back(ex);
+        break;
+      default:
+        THROW("ccb_adf_explore: badly formatted example - invalid example type");
+    }
```
rajan-chari: Return an error here to avoid throw.  Perhaps skip processing example?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,666 @@
...
+{
+  clear_all(data);
+  split_multi_example(examples, data);  // split shared, actions and slots
+
+#ifndef NDEBUG
+  sanity_checks<is_learn>(data);
+#endif
+
+  // Stash the CCB labels before rewriting them.
+  data.stored_labels.clear();
```
rajan-chari: Can be moved to clear_all?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,666 @@
...
+  for (auto ex : examples)
```
rajan-chari: >for [](start = 2, length = 3)

Move this into split_multi_examples?  You are iterating the example collection there anyway.  Might require renaming split_multi_example()

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,666 @@
...
+  // for each slot, re-build the cb example and call cb_explore_adf
+  int slot_id = 0;
+  for (example* slot : data.slots)
+  {
+    // Namespace crossing for slot features.
+    // If the slot example only has the constant namespace, there will be no extra crossing and so skip that logic.
+    if (!(slot->indices.size() == 1 && slot->indices[0] == constant_namespace))
+    {
+      data.generated_interactions.clear();
+      std::copy(data.original_interactions->begin(), data.original_interactions->end(),
```
rajan-chari: Use vector assignment operator?  Easier to understand.  Might have vector specific optimizations depending on impl.  (might alloc memory at the beginning, for example)

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,666 @@
...
+    {
+      LabelDict::del_example_namespace(*shared, index, slot->feature_space[index]);
+    }
+  }
+}
+
+// Generates quadratics between each namespace and the slot id as well as appends slot id to every existing interaction.
+void calculate_and_insert_interactions(example* shared, std::vector<example*> actions, std::vector<std::string>& vec)
+{
+  std::vector<std::string> new_interactions;
```
rajan-chari: Use static thread_local, with clear pattern to avoid allocation here?

For example:
[https://github.com/rajan-chari/vowpal_wabbit/blob/5c21b2dfda90cd8beb1372a0926f4a445dffb708/vowpalwabbit/offset_tree.cc#L128](https://github.com/rajan-chari/vowpal_wabbit/blob/5c21b2dfda90cd8beb1372a0926f4a445dffb708/vowpalwabbit/offset_tree.cc#L128)

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,666 @@
...
+    }
+    else
+    {
+      LabelDict::del_example_namespace(*shared, index, slot->feature_space[index]);
+    }
+  }
+}
+
+// Generates quadratics between each namespace and the slot id as well as appends slot id to every existing interaction.
+void calculate_and_insert_interactions(example* shared, std::vector<example*> actions, std::vector<std::string>& vec)
```
rajan-chari: >vec [](start = 113, length = 3)

nit: name this generated_interactions or something like that?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,666 @@
...
+{
```
rajan-chari: Add a unit test for this?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,666 @@
...
+  std::vector<std::string> new_interactions;
+  for(auto interaction : vec)
```
rajan-chari: Should this be auto&?

vowpalwabbit/conditional_contextual_bandit.cc
rajan-chari: There are a number of strings created and destroyed in this function.   v_array<string> w/ a modified resize() of v_array where string destructors are not called + only string assignment operations which cause a memcpy into existing string storage would save those allocations.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,666 @@
...
+  for (auto index : slot->indices)
+  {
+    // constant namespace should be ignored, as it already exists and we don't want to double it up.
+    if (index == constant_namespace)
+    {
+      continue;
+    }
+    else if (index == default_namespace)  // slot default namespace has a special namespace in shared
+    {
+      LabelDict::add_example_namespace(*shared, ccb_slot_namespace, slot->feature_space[32]);
```
rajan-chari: Replace with well known constant?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,666 @@
...
+          std::back_inserter(data.generated_interactions));
+      calculate_and_insert_interactions(data.shared, data.actions, data.generated_interactions);
+      data.shared->interactions = &data.generated_interactions;
+      for (auto ex : data.actions)
+      {
+        ex->interactions = &data.generated_interactions;
+      }
+    }
+
+    multi_ex cb_ex;
```
rajan-chari: Use static thread_local, with clear pattern to avoid allocation here?

For example:
[https://github.com/rajan-chari/vowpal_wabbit/blob/5c21b2dfda90cd8beb1372a0926f4a445dffb708/vowpalwabbit/offset_tree.cc#L128](https://github.com/rajan-chari/vowpal_wabbit/blob/5c21b2dfda90cd8beb1372a0926f4a445dffb708/vowpalwabbit/offset_tree.cc#L128)

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: I moved to using a set due to duplicate issues, assignment can no longer be used.

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: I do not think this is worth it given the scale that this function operates at. I also recently moved to a set, so the number of strings that are copied into the final generated_interactions is further reduced.

JohnLangford: a) 'set' is generally a bad data structure because it involves lots of pointer chasing.
b) When -q :: is used, the number of strings is quite high and running through them is the dominant computational factor.

jackgerrits: In this function we want to maintain a unique collection of namespaces to use, which a set lends itself to well. Unless we determine this to be a significant perf bottleneck then I think a set should be used.

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: I don't believe we have a system to skip processing?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,666 @@
...
+    auto current_index_str = "index"+std::to_string(id);
+    index = VW::hash_feature(*data.all, current_index_str, data.id_namespace_hash);
+    data.hashes[id] = index;
+  }
+  else
+  {
+    index = data.hashes[id];
+  }
+
+  shared->feature_space[ccb_id_namespace].push_back(1., index);
```
rajan-chari: What's this?

jackgerrits: This is where we insert the slot id feature

vowpalwabbit/conditional_contextual_bandit.cc
JohnLangford: I expect this is a significant perf bottleneck when -q :: is active.  

But, I'm slightly confused about 'unique collection of namespaces to use'.  Why/when would the set of namespaces not be unique by default?

vowpalwabbit/conditional_contextual_bandit.cc
rajan-chari: What if we returned actions in the same order presented?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,662 @@
...
+  }
+}
+
+template <bool audit>
+void inject_slot_id(ccb& data, example* shared, int id)
+{
+  // id is zero based, so the vector must be of size id + 1
+  if(id + 1 > data.hashes.size())
+  {
+    data.hashes.resize(id + 1, 0);
```
rajan-chari: >hashes [](start = 9, length = 6)

If this is only used to store slotid hashes, rename to slotid_hashes.  It's nice that this needs to be only populated once.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,662 @@
...
+void clear_pred_and_label(ccb& data)
+{
+  return_v_array(data.shared->pred.a_s, data.action_score_pool);
+
+  // This just needs to be cleared as it is reused.
+  data.actions[data.action_with_label]->l.cb.costs.clear();
+}
+
+// true if there exists at least 1 action in the cb multi-example
+bool has_action(multi_ex& cb_ex) { return cb_ex.size() > 1; }
```
rajan-chari: vector::empty() probably has a more efficient implementation

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -0,0 +1,666 @@
...
+// Generates quadratics between each namespace and the slot id as well as appends slot id to every existing interaction.
+void calculate_and_insert_interactions(example* shared, std::vector<example*> actions, std::vector<std::string>& vec)
+{
+  std::vector<std::string> new_interactions;
+  for(auto interaction : vec)
+  {
+    interaction.push_back((char)ccb_id_namespace);
+    new_interactions.push_back(interaction);
+  }
+  vec.insert(vec.end(), new_interactions.begin(), new_interactions.end());
```
peterychang: good place to use a move_iterator?

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: They would not be unique because each action contains some collection of namespaces and there is almost certainly overlap between them

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: For the vec end iterator or the set iterator?

peterychang: Oops, sorry I was looking at an old version of the file. Ignore my comment

vowpalwabbit/conditional_contextual_bandit.cc
JohnLangford: Different actions would certainly commonly have the same namespace interactions.  Why do you care whether they are unique?

jackgerrits: The vector of interactions must be unique - we need to ensure it remains as a unique set. We can either use a set as we go or once we are done remove duplicates

JohnLangford: 1) It need not be unique---nothing breaks if it's not unique, you just spend more computation.  In particular, sometimes people explicitly choose to have (say) -q ua and -q au for redundancy reasons.
2) Given that it starts unique for individual example (=action and/or shared example) how do you lose uniqueness?

jackgerrits: 1. In my interpretation those are still unique, UA and UA would be duplicates and not unique. Wouldn't it do too many updates for the interaction feature? Regardless we don't want to do wasted work there.
2. Uniqueness is lost because several examples may have the same namespace (very likely) and since we need to scan over every action to find the namespaces we may end up with duplicates.

I did a performance test using google benchmark. I compared using a set as I am now with string copies, and using the vector directly and using binary search to find whether to insert every interaction into the set. Interestingly, the results showed almost exactly equal performance (several runs averages out the difference). 

```
2019-06-26 13:42:50
Running ./perf
Run on (40 X 2195 MHz CPU s)
Load Average: 0.52, 0.58, 0.59
----------------------------------------------------------------
Benchmark                      Time             CPU   Iterations
----------------------------------------------------------------
BM_Set                     64458 ns        64174 ns        11200
BM_VectorBinarySearch      54268 ns        54688 ns        10000
```

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: This doesn't communicate the end user that this is a skipped iteration. It could be interpreted incorrectly easily.

vowpalwabbit/conditional_contextual_bandit.cc
JohnLangford: I'm still not understanding why you expect to get duplicates.  

Assuming the input interactions are unique and not using the private namespace for CCB, can you get duplicates?  How?

W.r.t. benchmark, what's the computational time when you do no duplicate detection?

jackgerrits: The duplicate comes from appending slot_id_namespace with action namespaces. Each action has namespace "Action" so after scanning all n actions there are n A + slot_id_namespace  interactions.

Doing no deduping:
```
2019-06-26 14:08:54
Running ./perf
Run on (40 X 2195 MHz CPU s)
Load Average: 0.52, 0.58, 0.59
----------------------------------------------------------------
Benchmark                      Time             CPU   Iterations
----------------------------------------------------------------
BM_Set                      5184 ns         5162 ns       112000
BM_VectorBinarySearch       5132 ns         5156 ns       100000
BM_NoUnique                 2582 ns         2623 ns       280000
```

JohnLangford: So this is definitely a 'hot' spot.  

If the interactions for every action are the same, then why do we need to do everything n times?  Can we just skip that and never have dups?

jackgerrits: We don't know what interactions are present for any example until we see it. 

Perhaps we have a misunderstanding here. Should I instead be simply inserting a single interaction in the form {':','slot_id_namespace'}? Currently I am scanning for what's present and calculating the interactions?

JohnLangford: I agree that we don't know interactions in advance, but we do know that all the interactions of any individual example are unique if they should be.  

You can't use ':' because that needs to be expanded at this level. 

I'm still fundamentally confused about why we should care about duplicates.  It seems like we should be able to process each action independently without ever suffering through a union.

test/unit_test/ccb_parser_test.cc
```diff
@@ -0,0 +1,143 @@
...
+#include <boost/test/unit_test.hpp>
+#include <boost/test/test_tools.hpp>
+
+#include "test_common.h"
+
+#include <vector>
+#include "conditional_contextual_bandit.h"
+#include "parser.h"
+
+CCB::label parse_label(parser* p, std::string label)
```
lokitoth: Should this label parser be defined here? 

It seems like we would want to be testing the ccb_label_parser from ccb_label.h/ccb_label.cc

jackgerrits: This is correct, it is a test helper method.

vowpalwabbit/conditional_contextual_bandit.cc
rajan-chari: v_array?

vowpalwabbit/conditional_contextual_bandit.cc
rajan-chari: I was reviewing GPerf output.  It turns out that Time is time / iteration.  The number of iterations is not relevant except that you don't want something like one iteration.  So BM_VectorBinarySearch is better.

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: I don't think using v_array is necessary here. v_array still reallocates as well after a number of clears

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: Ah I see, that makes sense. Thanks for the clarification!

We're using a constant sized vector to mark now, and we don't need to dedupe so we are in a better place.

JohnLangford:
Is this mergable today?

jackgerrits:
I think so, at this point I've addressed all comments.

jackgerrits:
I added a benchmark for the bool bookkeeping case. Turns out that with the knowledge of the fixed size we can be significantly more efficient.
```
2019-07-02 09:51:53
Running ./perf_tests
Run on (40 X 2195 MHz CPU s)
Load Average: 0.52, 0.58, 0.59
----------------------------------------------------------------------
Benchmark                            Time             CPU   Iterations
----------------------------------------------------------------------
set                               1106 ns         1099 ns       640000
vector_binary_search_insert       1025 ns         1025 ns       746667
no_deduplication                  1058 ns         1050 ns       640000
fixed_bool_bookkeeping             430 ns          430 ns      1600000
```

JohnLangford:
Good.  If you profile, is the computational complexity not in this reduction?

jackgerrits:
Here is the runtime for learn down the entire stack. The thing to keep in mind is that since they are recursive the numbers accumulate.
image

Internally to CCB, the primary cost is the interactions generation. So it is a very good thing that we went with the implementation that we did.
image

jackgerrits:
@JohnLangford Do you think this is ready to merge?

JohnLangford:
Yep merged.  Sorry about the slowness---I ended up sick for the last week-or-so. 

This is a big one.  Congrats :-)

PR #1818, opened 2019-04-02
Fix L2S HookTask in c-library and Python bindings

### LINKED ISSUE #1413: Hook interface
Is it possible that the search_hooktask interface is currently broken in master?  

I've been trying build master with python3.6 and I haven't been able to get the python learning-to-search examples to work.  I also tried to run the library/test_search.cc example that also uses the hook interface and found that one was segfaulting too.  

The other tests all seem fine in my build environment but if the hook interface is working properly for everyone else then I'm definitely doing something wrong.

### LINKED ISSUE #1671: library/test_search segfault on master
Perhaps caused by the same issue as issue "Hook interface" #1413, building vw and then doing 
```
cd library
make
./test_search
``` 
Also results in a core dump 

This is on ArchLinux, 64bit, with GCC 8.2.1 and Boost 1.68

lokitoth:
Some time ago the "hook" L2S task was broken, and subsequently explicitly disabled. This change fixes it.

The underlying issues were around setting up the pointers to hook into correctly and update to the new multi-line example API, as the code changed due to various refactorings.

Fixes #1413, though there are still issues with LDF.
Fixes #1671


vowpalwabbit/search.cc
```diff
@@ -2737,7 +2737,7 @@ base_learner* setup(options_i& options, vw& all)
   options.add_and_parse(new_options);
 
   // If hook was passed as the search_task then do not activate reduction. This task can only be used in library form.
-  if (!options.was_supplied("search_task") || task_string == "hook")
+  if (!options.was_supplied("search_task") /* || task_string == "hook" */)
```
jackgerrits: Can the commented part just be removed outright now?

lokitoth:
Went over this in person with @hal3.

JohnLangford:
Merged, thanks :-)

PR #1819, opened 2019-04-03
Create clang-format CI pipeline

jackgerrits:
Currently clang-format is used in the Linux CI to test how many style issues there are. This is pretty much always ignored. This PR introduces a new azure-pipeline [(see here)](https://dev.azure.com/vowpalwabbit/Vowpal%20Wabbit/_build?definitionId=17&_a=summary), which we can add alongside Windows and Linux, but can make non essential for merging. I believe it will be easier to see this info in this location, reduces noise in the Linux build and is system agnostic.


jackgerrits:
Going to close as this isn't the behavior we want to surface style formatting as. Perhaps a bot that pushes a style commit to a PR upon request is a better way to tackle this

PR #1828, opened 2019-04-05
Consume RapidJson through submodule

### LINKED ISSUE #1689: Remove rapidjson from repo
Rapidjson should either be consumed by a submodule or through a package manager like apt-get

jackgerrits:
- Rapidjson is now consumed as a submodule at v1.1.0
- When using CMake the submodule is automatically checked out if it isn't already there
- When using VisualStudio the submodule must be explicitly checked out

Fixes #1689


jackgerrits:
Running a CI image update now, once that's done I will be able to fix the travis build

jackgerrits:
It seems the apt-get package isn't available on 14.04. Perhaps this should be done through a submodule instead then

jackgerrits:
Guidance on the project website is to just copy the `rapidjson/include` directory. So perhaps as a good first effort we can simply remove all of the extra files that don't need to be there?

kumpera:
Wasn't the idea to consume it through a submodule?

jackgerrits:
You're right, was hoping we could do it a little easier with packages but it doesn't seem like the way to go

lokitoth:
+40 -53113

^ The best kind of code change

Are we also upgrading the version here, or is it a like-for-like substitution?

jackgerrits:
It's tricky to know, I am targeting the last released version which was in 2016... It seems like we were on something like that as well in the source.

lokitoth:
We may want to do a manual test pass around good and bad input, and/or make a smoke-test?

jackgerrits:
A few of our tests go through json parsing, but yes there is potential for issues if behavior changed. It seems like versioning is not very good on Rapidjson but I think we are on the same major and minor version

lokitoth:
Okay, that seems reasonable. Expectation should be that if major/minor remain constant, behaviour should be the same up to bug fixes.

jackgerrits:
Just diffed the submodule against the old checked in source and the only changes are whitespace as far as I can tell

JohnLangford:
We should merge as soon as tests pass.

JohnLangford:
This seems to be causing issues with lgtm that would be good to fix.  https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/logs/rev/pr-9c14123a062578394b2332b974752376dc9fc6a0/lang:cpp/stage:Build%20merge_78d7529bf560c71feaef2324f27ffb47a1a9d50c

jackgerrits:
The LGTM failure is a known issue and will be resolved when we merge the PR. See this forum post for details: https://discuss.lgtm.com/t/git-submodule-sync-missing/1989

PR #1830, opened 2019-04-05
fix bug 1784: return probabilities for actions in the correct order in python binding

pavithraks:
This is a fix to bug https://github.com/VowpalWabbit/vowpal_wabbit/issues/1784

The problem was that the actual scores list was indexed by action. In the python binding, in `pyvw.cc`, in the function `ex_get_action_scores`, we were appending the probabilities to a list ignoring the index that specified the action. I have added a new array that stores both the index and the score and use this new array to generate the list that is returned.


jackgerrits:
For reference this is a resubmit on a new branch for this PR: #1829

Issue #1592, opened 2018-09-01
--Ojanewton breaks with -b 24


arielf:
In the vw source tree:

```
# takes about 0.25 sec, average loss = 0.089070
vw  --OjaNewton --sketch_size=10 --alpha_inverse=1.0 -d test/train-sets/0002.dat
```

Same command with -b 24:
```
# Takes over 16 seconds, has NaNs in all examples & error jumps to 0.283062
vw  -b 24 --OjaNewton --sketch_size=10 --alpha_inverse=1.0 -d test/train-sets/0002.dat
```

arielf:
Also noticed that `-b 20` still works reasonably well (although slower than default) and the breaking point is `-b 21`.

JohnLangford:
I'm not sure what's wrong here.  A careful examination of the computation is required.  @haipengl do you want to take a look?

yannstad:
The independent sample `r1` in the box-muller transform must be strictly positive. But it is chosen using merand48, that possibly returns 0:
https://github.com/VowpalWabbit/vowpal_wabbit/blob/6ef7676bb222d8f11edaf89165fdb9e925e7b247/vowpalwabbit/OjaNewton.cc#L82

PR #1833, opened 2019-04-09
Fix NaN issue in box-muller tranform

yannstad:
Fix issue #1592

### DIFF
--- a/vowpalwabbit/OjaNewton.cc
+++ b/vowpalwabbit/OjaNewton.cc
@@ -73,7 +73,6 @@ struct OjaNewton
       // more complicated initialization: orthgonal basis of a random matrix
 
       const double PI2 = 2.f * 3.1415927f;
-      const float FLOAT_MIN = (std::numeric_limits<float>::min)();
 
       for (uint32_t i = 0; i < length; i++)
       {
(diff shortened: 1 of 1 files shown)


vowpalwabbit/OjaNewton.cc
```diff
@@ -81,6 +82,15 @@ struct OjaNewton
         {
           float r1 = merand48(all->random_state);
           float r2 = merand48(all->random_state);
+
+          //in the box-muller tranform, r1 should be strictly positive: https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform
+          //redraw until r1 > 0
+          while (r1 < FLOAT_MIN)
```
kumpera: Why not check for zero?

yannstad: In order to avoid numeric instabilities when `0 < r1 < FLOAT_MIN`, especially for the `log(r1)` calculation happening next.

kumpera: That's impossible to happen.
FLOAT_MIN, as you defined, is, by design, the smallest number bigger than zero that can be represented by a float.

yannstad: Ok, thanks. I will check for zero instead.

Issue #1482, opened 2018-05-15
Java API oddity (Contextual Bandits)


matanster:
I might be hallucinating, but it seems that the java API is inconsistent here with regard to the index semantics of the actions:

When reporting costs to the model (method `learn`), it needs to be 1-indexed; otherwise you get: 
> invalid action: 0

However, in the distribution returned when predicting (method `predict`), the actions are clearly 0-indexed, as seen in this example value of the returned distribution object:
> [ActionProb{action=0, probability=0.05}, ActionProb{action=1, probability=0.95}]

This seems to be the case for models initialized with `--cb_explore`, `--epsilon`.

palmerlao:
We've also observed this with `--cb_adf`.

matanster:
@palmerlao thanks for the note. Obviously both our code-bases are already adapted to it, but this makes it a hell when initially developing with the API, not to mention any logging and debug processes....

JohnLangford:
This is an issue that clearly needs fixing.

matanster:
Thanks and just to note that even being a fix, it should probably be treated as a _breaking api change_, in terms of communication upon fixing; to avoid excessive application breakage.

JohnLangford:
I agree.

-John

jackgerrits:
This issue is being tracked by #1841

PR #1835, opened 2019-04-10
Change learner output action to be 1 indexed

cheng-tan:
Fix issue #1482

### DIFF
--- a/java/src/main/c++/vowpalWabbit_learner_VWActionProbsLearner.cc
+++ b/java/src/main/c++/vowpalWabbit_learner_VWActionProbsLearner.cc
@@ -16,7 +16,7 @@ jobject action_probs_prediction(example* vec, JNIEnv *env)
   jclass action_probs_class = env->FindClass("vowpalWabbit/responses/ActionProbs");
   for (uint32_t i=0; i<num_values; ++i)
   { ACTION_SCORE::action_score a = a_s[i];
-    jobject j_action_prob = env->NewObject(action_prob_class, action_prob_constructor, a.action, a.score);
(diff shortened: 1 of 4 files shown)


JohnLangford:
I think we would rather change everything to be zero indexed?

cheng-tan:
yes, that requires c++ layer changed to be zero index which is a breaking change. This is a temporary fix to make java API action index be consistent. Once we change everything to be zero indexed, this commit should be removed.

jackgerrits:
I very much agree that indices need to be normalized to 0 based (but not in this PR). We will 100% need to bump the major version number though as it is a hard breaking change for existing data files and scripts that use this, and we may need a migration story in place.

jackgerrits:
I've created an issue to track the larger piece of work about fixing the indexing properly.

#1841

cheng-tan:
@jmorra and @deaktator Please take a look when you get a chance. There is an inconsistency for java API when calling `learn`, for input action needs to be 1-indexed and output is 0-indexed. This commit change the output to be 1-indexed as well for consistency. Everything will be changed to be 0-indexed when we are ready for the migration.

JohnLangford:
It seems like we have three options before us:
(1) Leave as-is and document. 
(2) Shift everything to be 1-indexed and then shift everything to be 0-indexed.  
(3) Shift everything to be 0-indexed. 

I think we are in agreement that 0-indexed is better in the long term?  Given that, shouldn't we be preferring (3) over (2)?

jon-morra-zefr:
This seems fine with me.  I'd favor consistency everywhere and then make the breaking change later.

JohnLangford:
Closing this for now, since we want to avoid a double switch in the interface.

PR #1837, opened 2019-04-10
Implement finish_example, update scripts to match

### LINKED ISSUE #1810: Implement example.finish in Python bindings
Currently the finish method is not implemented for the example class, specifically on the CPP side my_finish_example is not implemented. Currently this means examples are not released in the bindings.

https://github.com/VowpalWabbit/vowpal_wabbit/blob/04b3f605e45ebadbc75f10abe313784ee9bc5ac0/python/pylibvw.cc#L173

jackgerrits:
- Implements finish_example
- Required an interface change because example and multi_ex must be treated differently and so the single Python example object can no longer expose the ability to finish directly. (Calling finish on a single example in a multi_ex will not work)

Fixes #1810


peterychang:
lgtm

jackgerrits:
@Scott-Graham-Bose if you have a chance to review that would be good

PR #1843, opened 2019-04-11
Fix conda_install.sh script

peterychang:


python/vowpalwabbit/conda_install.sh
```diff
@@ -12,8 +12,8 @@ conda install --yes gxx_linux-64
 source activate ${CONDA_DEFAULT_ENV}
 
 # boost packages
-BOOST_V="1.65"
-conda install -c conda-forge --yes boost=$BOOST_V libboost=$BOOST_V py-boost=$BOOST_V
+BOOST_V="1.67"
```
jackgerrits: Why do we hardcode a boost version? Is that required?

peterychang: removed the requirement with a fixup

PR #1854, opened 2019-04-25
cleaning up broken python3 vcxproj files

peterychang:
I'm leaving in the python27 vcxproj files since the CMake/vcpkg method requires a manual edit of a vcpkg CONTROL file.
https://github.com/Microsoft/vcpkg/issues/3495

I'll remove the old workflow once I can find a programmatic way of handling it


python/README.rst
```diff
@@ -194,4 +194,6 @@ Due to limitations in the current version of boost-python, some manual changes m
 
 .. code-block:: bat
 
-    > python setup.py --vcpkg-root=[vcpkg-dir] install
\ No newline at end of file
+    > python setup.py --vcpkg-root=[vcpkg-dir] install
+    
+.. _experimental Windows CMake build: https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/README.Windows.md#experimental-cmake-build
```
jackgerrits: It would be better to link here: https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Dependencies#experimental-cmake-build-system-on-windows

jackgerrits:
This seems like a good opportunity for us to contribute to vcpkg, the change seems easy and it will help us a lot here

jackgerrits:
Looks like the failures are due to the fact the submodule was removed in the merge

PR #1859, opened 2019-05-02
Ataymano/c wrapper fix2

ataymano:
c wrapper:
- missing feature space construction functions
- Add label deprecation comments
- parse_mask usage unification


vowpalwabbit/parse_example.cc
```diff
@@ -69,6 +69,7 @@ class TC_parser
   bool* spelling_features;
   v_array<char> spelling;
   uint32_t hash_seed;
+  uint64_t parse_mask;
```
jackgerrits: Would you be able to explain what the parse_mask is for?

vowpalwabbit/vwdll.h
```diff
@@ -120,7 +126,9 @@ extern "C"
   VW_DLL_MEMBER float VW_CALLING_CONV VW_Learn(VW_HANDLE handle, VW_EXAMPLE e);
   VW_DLL_MEMBER float VW_CALLING_CONV VW_Predict(VW_HANDLE handle, VW_EXAMPLE e);
   VW_DLL_MEMBER float VW_CALLING_CONV VW_PredictCostSensitive(VW_HANDLE handle, VW_EXAMPLE e);
+  //deprecated. Please use either VW_ReadExample for parsing, or VW_ImportExample for example construction
```
JohnLangford: Not deprecated, right?

vowpalwabbit/vwdll.cpp
```diff
@@ -77,19 +77,30 @@ VW_DLL_MEMBER void      VW_CALLING_CONV VW_Finish(VW_HANDLE handle)
...
 
-VW_DLL_MEMBER VW_EXAMPLE VW_CALLING_CONV VW_ImportExample(VW_HANDLE handle, const char * label, VW_FEATURE_SPACE* features, size_t len)
+VW_DLL_MEMBER VW_EXAMPLE VW_CALLING_CONV VW_ImportExample(VW_HANDLE handle, const char * label, VW_FEATURE_SPACE features, size_t len)
 { vw * pointer = static_cast<vw*>(handle);
   VW::primitive_feature_space * f = reinterpret_cast<VW::primitive_feature_space*>( features );
   return static_cast<VW_EXAMPLE>(VW::import_example(*pointer, label, f, len));
 }
 
+VW_DLL_MEMBER VW_FEATURE_SPACE VW_CALLING_CONV VW_InitializeFeatureSpaces(size_t len)
+{
```
JohnLangford: Why are these needed precisely?

JohnLangford: In other words, what is the missing functionality amongst existing interfaces?

JohnLangford:
This one is breaking on a python test related to lrq which needs to be updated. 

I'm also hazy on a detail: why do we need the extra functions in the interface?  

It does seem plausible we could easily get this one done quickly.

vowpalwabbit/vwdll.cpp
ataymano: With existing API the only way to "construct" example - VwExample ImportExample(FeatureSpace, Label);
But the only way to get feature space - FeatureSpace ExportExample(VwExample).
So I do not see any way to construct example without introducing functions required for initializing, adding features / namespaces etc to feature space. Please correct me if I am wrong.

JohnLangford: I guess the question is: What is the api?  Is it vwdll.cpp?  Or is it vw.h?  To the extent that vwdll.cpp requires additional functions exposing vw.h that seems fine.  To the extent that vw.h is missing things, I'd like to understand more deeply.  Is vw.h missing things?  If not, can these extra interface functions just call the vw.h interface?

ataymano: My understanding that cpp api is vw.h while c one is vwdll.
vw.h is not missing anything - at least allows to create example, since it is exposing full primitive_feature_space implementation:
https://github.com/VowpalWabbit/vowpal_wabbit/blob/92064660a778d343285e92620fa8077d41db841c/vowpalwabbit/vw.h#L54
So given current vw.h code, direct manipulations with primitive_feature space is technically "using vw.h interface".
vw.h can be refactored to hide its feature_space implementation details behind the interfaces, but still looks like corresponding flat c methods have to be added to vwdll.

JohnLangford: Good.  So, can we make sure the added vwdll functions use functions in vw.h?  Currently, it seems they are not, so this is sort-of creating an alternate interface.

ataymano: Sure, let me try.

ataymano: Sent the update draft, please take a look.

JohnLangford:
We need to update this test: https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/tests/test_sklearn_vw.py#L144 as well, given the implied changes t lrq.

JohnLangford:
And it looks like a C# test is failing, although the reason why is unclear... https://ci.appveyor.com/project/JohnLangford/vowpal-wabbit/builds/25040837#L1329

ataymano:
> And it looks like a C# test is failing, although the reason why is unclear... https://ci.appveyor.com/project/JohnLangford/vowpal-wabbit/builds/25040837#L1329

Yes, I am looking into this - was not able to reproduce locally so far

lokitoth:
> > And it looks like a C# test is failing, although the reason why is unclear... https://ci.appveyor.com/project/JohnLangford/vowpal-wabbit/builds/25040837#L1329
> 
> Yes, I am looking into this - was not able to reproduce locally so far

I have seen some flakiness out of this test in the past.

PR #1862, opened 2019-05-04
Use libz.a when STATIC_LINK_VW set

palmerlao:
This change allows me to run `cmake .. -DSTATIC_LINK_VW=ON` and `make -j 7 vw-bin` successfully. Other binaries may still be broken.

The output of `file vowpalwabbit/vw` is then
```
vowpalwabbit/vw: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), statically linked, for GNU/Linux 3.2.0, BuildID[sha1]=d703aa11a699f0ba5d9c501944de0e63b0c1d1a9, with debug_info, not stripped
```

There is a warning output during the build, but I believe this caveat applies to all static builds involving glibc?
```
[100%] Linking CXX executable vw
libvw.a(network.cc.o): In function `open_socket(char const*)':
network.cc:(.text+0xa1): warning: Using 'gethostbyname' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
[100%] Built target vw-bin
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,6 +109,7 @@ find_package(Threads REQUIRED)
 
 set(LINK_THREADS Threads::Threads)
 if(STATIC_LINK_VW)
+  find_package(ZLIB REQUIRED libz.a)
   if(APPLE)
     set(unix_static_flag "")
     #Guess ZLIB_LIBRARY to be the one provided by homebrew if none was provided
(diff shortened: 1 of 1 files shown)


jackgerrits:
The current setup is supposed to find the static version of zlib, is that not happening for you? When I build with the CMake command you supplied I get a vw binary that `file` says:
```
vowpalwabbit/vw: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), statically linked, for GNU/Linux 2.6.32, BuildID[sha1]=dd79f7f56236fdc110658f1a9675bd0c11ffb5e0, not stripped
```

The line with the arrow below is the one that specifically causes CMake to find the static zlib.
```
if(STATIC_LINK_VW)
  set(Boost_USE_STATIC_LIBS ON)
  SET(CMAKE_FIND_LIBRARY_SUFFIXES ".a") <--------
  SET(BUILD_SHARED_LIBS OFF)
else()
```

Yes, I think you're right about the glibc caveat.

palmerlao:
:thinking: I can't seem to reproduce the error I was seeing earlier. My mistake, sorry!

jackgerrits:
Not a problem at all, I appreciate the work you put into VW

PR #1864, opened 2019-05-06
facilitate mle experiments with pyvw

pmineiro:
Here's my workaround allowing me to use cbify (on the cb_bakeoff datasets) while getting the action distribution via pyvw.


pmineiro:
I want to tweak this some more before merging.

PR #1869, opened 2019-05-09
Implement history rollup and higher order interactions for CCB

jackgerrits:


JohnLangford:
This pull request **introduces 1 alert** when merging 3214ac74d201bd1b79c7258626dbab8edf434fc3 into eaf556f5651aa23d19bd8a3fd26462c9280f2c08 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-808602d91d52497fba8444754645e02e3fe907d1)

**new alerts:**

* 1 for FIXME comment

---

vowpalwabbit/example.h
```diff
@@ -36,6 +36,7 @@ const unsigned char dictionary_namespace = 135;    // this is \x87
 const unsigned char node_id_namespace = 136;       // this is \x88
 const unsigned char message_namespace = 137;       // this is \x89
 const unsigned char ccb_decision_namespace = 139;
+const unsigned char ccb_history_namespace = 140;
```
lokitoth: A bit of an aside: Is it expected that the user never provide features in these namespaces? Or are we treating collisions as something that "might just happen" but rarely?

jackgerrits: All namespaces above 125 are reserved and it is assumed a user does not use them

lokitoth: We may want to brainstorm how to deal with namespace names that start with a unicode character that is non-ASCII. Well out of the scope of this change, though.

jackgerrits: We will need to extend the range of values in that case. But yes that is a discussion that will need to happen eventually

Issue #1714, opened 2019-01-17
Update new_version script to update version.txt and other instances of version number


jackgerrits:

lokitoth:
Ideally, all the other places should grab the version number from version.txt, wherever possible. Are there situations where this cannot be done?

jackgerrits:
I think the extra ones are mainly AssemblyInfo.*

lokitoth:
There should be a task to generate AssemblyInfo* that I believe is used. We might be able to do a bit of archeology to find it and drive it manually based on this version file.

jackgerrits:
`cs\version.props` also contains the version number and will need to be considered for this

jackgerrits:
https://github.com/VowpalWabbit/vowpal_wabbit/pull/1884/files

`python/docs/conf.py` needs to be updated as well

thealphadollar:
@jackgerrits Is this issue complete?

jackgerrits:
Release management is still something we are working on. The new_version script has issues and as such it has not really been used. Ideally release pipelines will be setup eventually but we need to find time to work that out.

PR #1870, opened 2019-05-09
Enable Version and Tag override for NuGet pack

lokitoth:
This provides a mechanism to insert version/tag during the CI/CD pipeline (thereby inserting a version generated by the pipeline).

Longer-term this will be handed by a "version generation" step, which will read the target version from a file, and create the appropriate artifacts (config.h, build.props, etc.) to be used by the appropriate build systems to inject the version. In the short term, this unblocks work on release pipeline for Windows artifacts.

### DIFF
--- a/.scripts/package.cmd
+++ b/.scripts/package.cmd
@@ -9,9 +9,14 @@ CALL %~dp0init.cmd
 
 PUSHD %~dp0..
 
-REM TODO: these should be read from the version props file, or similar
-SET Version=8.6.1
-SET Tag=-INTERNALONLY
+REM TODO: these should be read from the version props file, or similar, with the ability to overload via env
+IF NOT DEFINED Version (
+    SET Version=8.6.1
+)
+
+IF NOT DEFINED Tag (
+    SET Tag=-INTERNALONLY
+)
 
 SET RootRelativeOutputDirX64=%vwRoot%\vowpalwabbit\out\target\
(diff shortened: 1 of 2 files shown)


.scripts/package.cmd
```diff
@@ -9,9 +9,14 @@ CALL %~dp0init.cmd
 
 PUSHD %~dp0..
 
-REM TODO: these should be read from the version props file, or similar
-SET Version=8.6.1
-SET Tag=-INTERNALONLY
+REM TODO: these should be read from the version props file, or similar, with the ability to overload via env
```
jackgerrits: Perhaps mention issue #1714 here and update the bug to mention this?

.scripts/package.cmd
```diff
@@ -9,9 +9,14 @@ CALL %~dp0init.cmd
...
+IF NOT DEFINED Version (
+    SET Version=8.6.1
+)
+
+IF NOT DEFINED Tag (
+    SET Tag=-INTERNALONLY
```
jackgerrits: Is tag an env var that nuget reads when packaging?

lokitoth: Yes, but only because we explicitly set it in the nuspec file: https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/cs/cs/cs.nuspec

`    <version>$version$$Tag$</version>`

jackgerrits: Ah got it, thanks

PR #1873, opened 2019-05-13
Shared example merge reduction

peterychang:


vowpalwabbit/shared_feature_merger.cc
```diff
@@ -0,0 +1,79 @@
...
+      ec_seq[0]->pred.a_s.clear();
+      ec_seq[0]->pred.a_s.push_back(tmp[0]->pred.a_s[0]);
+      std::swap(ec_seq[0]->pred.a_s, tmp[0]->pred.a_s);
+    }
+  }
+}
+
+LEARNER::base_learner* shared_feature_merger_setup(config::options_i& options, vw& all)
+{
+  if (!options.was_supplied("csoaa_ldf"))
```
peterychang: These options are the ones used by csldf_setup. As far as I can tell, its the only reduction that converts multi line examples to single line examples.

vowpalwabbit/cost_sensitive.cc
```diff
@@ -206,17 +206,8 @@ void print_update(vw& all, bool is_test, example& ec, multi_ex* ec_seq, bool act
...
-      // These should be already included in each example (TODO: including quadratic and cubic).
-      // TODO: code duplication csoaa.cc LabelDict::ec_is_example_header
-      example** ecc = &((*ec_seq)[0]);
-      const example& first_ex = *(*ec_seq)[0];
-
-      v_array<COST_SENSITIVE::wclass> costs = first_ex.l.cs.costs;
-      if (costs.size() == 1 && costs[0].class_index == 0 && costs[0].x < 0)
-        ecc++;
-
-      for (; ecc != &(*ec_seq->cend()); ecc++) num_current_features += (*ecc)->num_features;
```
peterychang: This line should crash since you can't dereference an end iterator. Switched it to use a range iterator since we remove the shared example earlier now

vowpalwabbit/label_dictionary.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "example_predict.h"
+#include "v_hashmap.h"
```
lokitoth: [Minor][Nit] Is this change needed?

vowpalwabbit/parse_args.cc
```diff
@@ -1278,6 +1279,7 @@ void parse_reductions(options_i& options, vw& all)
   all.reduction_stack.push(explore_eval_setup);
   all.reduction_stack.push(ExpReplay::expreplay_setup<'c', COST_SENSITIVE::cs_label>);
   all.reduction_stack.push(Search::setup);
+  all.reduction_stack.push(VW::shared_feature_merger::shared_feature_merger_setup);
```
lokitoth: This may be more of a @JohnLangford question, but is this the right place in the stack for this setup to run? It seems like we would want to only run this for reductions that inject csoaa_ldf or wap_ldf, even if it not provided at the command-line, but if this reduction is above them in the stack, it will never know that this argument gets injected.

lokitoth:
Looking at the checks, I am a little concerned about this:

From TravisCI: https://travis-ci.org/VowpalWabbit/vowpal_wabbit/builds/531903613#L4088
> RunTests: RUNTIME REGRESSION: 103.19 sec vs last time 10.83 sec. (852.81% worse)
> RunTests runtime: user 100.55, system 2.64, total 103.19 sec (+852.81% vs. last)

This could easily be a transient issue, but make sure that after merge/re-run-checks that this is not still there.

peterychang:
After merging with master and running the tests locally:
2: RunTests: RUNTIME REGRESSION: 24.01 sec vs last time 23.00 sec. (4.39% worse)
2: RunTests runtime: user 8.06, system 15.95, total 24.01 sec (+4.39% vs. last)
2/5 Test #2: RunTests_pass_2 ..................   Passed   28.97 sec

Looks like a transient issue, but I'll keep an eye on it here the next time my branch builds on travis

vowpalwabbit/label_dictionary.h
peterychang: shared_feature_merger.cc won't build without this. We can either include the header or forward declare v_hashmap.h. I typically avoid forward declarations unless absolutely necessary

peterychang:
There's actually an interaction I wasn't expecting here. It looks like parts of the examples are being modified in either the predict or learn workflow, and then being used as an input again on the other one. So when I do a swap on example.pred, crashes occur in some places

peterychang:
So it looks like moving the `add_example_namespaces_from_example` step earlier in the call stack may not actually be feasible.
At the very least, the cb_adf results appear to change depending on if the shared example have been merged into the other examples or not
on test 87:
```
1: --- diff -u --minimal pred-sets/ref/cb_adf_rank.predict cb_adf_rank.predict
1: --- pred-sets/ref/cb_adf_rank.predict        2019-05-06 14:19:52.455400300 -0400
1: +++ cb_adf_rank.predict      2019-05-20 11:59:34.965295400 -0400
1: @@ -1,6 +1,6 @@
1:  0:0,1:0,2:0
1:
1: -1:0,0:0.239436
1: +1:0,0:0.195903
1:
1: -1:0,0:0.359153
1: +1:0,0:0.293854
1:
1: RunTests: test 87: FAILED: ref(pred-sets/ref/cb_adf_rank.predict) != (cb_adf_rank.predict)
1:      cmd: /usr/bin/timeout 80 /mnt/c/git/vowpal_wabbit/build/test/../vowpalwabbit//vw --onethread --cb_adf --rank_all -d train-sets/cb_test.ldf -p cb_adf_rank.predict --noconstant
```

Closing this PR while I investigate whether this change is feasible or if we'll need to take a different approach

jackgerrits:
I've been thinking about this and I think this can coexist in its current form well with ccb. As long as it sits between cb and ccb in the reduction stack then it doesn't change anything. We don't get the benefit of being able to ignore shared examples in ccb, but that is impossible since we need to update that example to constuct the cb learning problem from ccb.

JohnLangford:
I'm not following the impossibility?

JohnLangford:
Ah, I see the issue is the "slot actions".
-John

vowpalwabbit/parse_args.cc
peterychang: Moved the reduction further down the stack (after the cbify/warm_cb reductions), traced and added the options that would result in calls into csoaa::do_actual_learning

peterychang: Its not a perfect solution, but it should work until we can refactor the reduction stack into something more explicit

vowpalwabbit/shared_feature_merger.cc
```diff
@@ -0,0 +1,112 @@
...
+#include "parse_args.h"
+#include "vw.h"
+
+#include <iterator>
+
+namespace VW
+{
+namespace shared_feature_merger
+{
+static const std::vector<std::string> option_strings = {
```
JohnLangford: Have an explicitly enumerated list is kind of subpar because it will inevitably fail to be updated properly in the future. 

Can we instead test directly for whether or not the parent reduction is multiline?

vowpalwabbit/shared_feature_merger.cc
```diff
@@ -0,0 +1,112 @@
...
+  {
+    if (options.was_supplied(opt))
+      return true;
+  }
+  return false;
+}
+
+struct sfm_data
+{
+  LEARNER::multi_learner* base;
```
JohnLangford: Why do you need a base separate from the base learner?

vowpalwabbit/shared_feature_merger.cc
```diff
@@ -0,0 +1,112 @@
...
+
+  const bool has_example_header = CB::ec_is_example_header(*ec_seq[0]);
+  if (has_example_header)
+  {
+    source = &tmp;
+    // copy vector and pass it down instead
+    tmp.reserve(ec_seq.size() - 1);
+    // skip first element since its a shared feature
+    auto startIter = std::next(ec_seq.begin());
+    tmp.insert(tmp.end(), startIter, ec_seq.end());
```
JohnLangford: This is doing allocation on the example path.  Can we avoid that?

vowpalwabbit/shared_feature_merger.cc
```diff
@@ -0,0 +1,112 @@
...
+    // merge sequences
+    for (size_t k = 0; k < tmp.size(); k++) LabelDict::add_example_namespaces_from_example(*tmp[k], *ec_seq[0]);
+    tmp[0]->pred = ec_seq[0]->pred;
+  }
+  if (source->size() == 0)
+    return;
+  if (is_learn)
+    base.learn(*source);
```
JohnLangford: Instead of dealing with source, why not just pass ec_seq?

vowpalwabbit/shared_feature_merger.cc
```diff
@@ -0,0 +1,112 @@
...
+  {
+    source = &tmp;
+    // copy vector and pass it down instead
+    tmp.reserve(ec_seq.size() - 1);
+    // skip first element since its a shared feature
+    auto startIter = std::next(ec_seq.begin());
+    tmp.insert(tmp.end(), startIter, ec_seq.end());
+    // merge sequences
+    for (size_t k = 0; k < tmp.size(); k++) LabelDict::add_example_namespaces_from_example(*tmp[k], *ec_seq[0]);
+  }
```
JohnLangford: Doing a copy multiple times per example seems pretty undesirable.  Why?

vowpalwabbit/shared_feature_merger.cc
peterychang: Unless I missed something, there's no way to know which other reductions are being used in the setup function unless we pass it through the all object. Since that would require a similar level of maintenance, I opted for options checking where we could at least encapsulate the check in a single place

vowpalwabbit/shared_feature_merger.cc
peterychang: The current design requires us to manipulate the examples before running the finish functions on them, this base is used to call into the original finish if necessary

vowpalwabbit/shared_feature_merger.cc
peterychang: This is one of two ways I can think of to handle this operation. The alternative would be to remove the shared example from the vector before passing it down to the rest of the stack.

Removing the shared example has the benefit of being a little faster and not necessarily requiring any memory allocations, the gains are fairly minimal in this case since the vector only maintain pointers to examples (and removing the first element of a vector requires a memcpy as well)
I had originally thought of changing multi_ex to be a deque instead, but Jack pointed out that the object is exposed in the various interfaces, which would complicate that change

vowpalwabbit/shared_feature_merger.cc
JohnLangford: I think we should leave this for a later refactoring.

vowpalwabbit/shared_feature_merger.cc
JohnLangford: Let's find a way to avoid this ---reporting should not be doubling our conversation.

vowpalwabbit/shared_feature_merger.cc
JohnLangford: Let's avoid memory allocation.

There is a third approach which is a bit unusual----you adjust pointers directly.

peterychang: re-arranging the pointers should have a similar (or worse) performance to just erasing and re-inserting the first element. Looking at the docs, erase has some guarantees around iterator invalidation, which means memory will never be reallocated for an erase; in turn this shouldn't affect the capacity at all, which would mean the insertion shouldn't cause a realloc either.

fix_cbtest_feature_count.patch
```diff
@@ -0,0 +1,319 @@
+diff --git b/test/train-sets/ref/cb_adf_crash2.stderr a/test/train-sets/ref/cb_adf_crash2.stderr
```
peterychang: Adding a patch file so we can fix the unit tests once the finish_examples function is made viable

jackgerrits: Perhaps put this patch file in the test folder?

vowpalwabbit/shared_feature_merger.cc
```diff
@@ -0,0 +1,106 @@
...
+    return;
+  if (is_learn)
+    base.learn(ec_seq);
+  else
+    base.predict(ec_seq);
+
+  if (has_example_header)
+  {
+    for (auto& example : ec_seq) LabelDict::del_example_namespaces_from_example(*example, *shared_example);
+    shared_example->pred = ec_seq[0]->pred;
```
JohnLangford: If you add example->num_features += shared_example->num_features at this line, you'll have a better accounting of the number of features (although it will not account for quadratics).

vowpalwabbit/shared_feature_merger.cc
```diff
@@ -0,0 +1,106 @@
...
+    ec_seq.insert(ec_seq.begin(), shared_example);
+  }
+}
+
+// Currently unused due to inefficiency.
```
JohnLangford: Remove please.

vowpalwabbit/shared_feature_merger.cc
peterychang: Since the examples can be passed through the reductions more than once, and since we're passing the examples by pointers here, I'm not sure its safe to modify the num_features directly without having a mechanism to undo the change elsewhere.

If I attempt to do the same operation in finish_multiline_example(), several other (seemingly unrelated) tests fail with bad feature counts.
As it is right now, the `current features` are the same as before this PR. The only thing that changes is the `total feature number` on several tests

JohnLangford: In the long term, I think we want the update rule itself to keep track of the number of features, in which case multiple uses of the same example results in incrementing.  

I agree this is not essential, but I don't know a good reason to not do it.

peterychang: In that case, the feature count increment should occur in finish_example, not predict_or_learn.

Incrementing the count here will change the number of features seen by upstream reductions. If an example is used in both predict and learn, we'd be incrementing these features multiple times as well.

JohnLangford: That seems fine.

JohnLangford:
Merging for now.   We can worry about feature count accounting at a later date.  Thanks :-)

PR #1874, opened 2019-05-13
Update gd_mf_weights.cc right namespace

szipkin:
Use right indices to write out features for the right namespace, correcting bug found in https://github.com/VowpalWabbit/vowpal_wabbit/issues/1871

### DIFF
--- a/library/gd_mf_weights.cc
+++ b/library/gd_mf_weights.cc
@@ -98,7 +98,7 @@ int main(int argc, char *argv[])
 
       right_quadratic << right.space_names[i].get()->second;
       for (size_t k = 1; k <= rank; k++)
-        right_quadratic << '\t' << weights[(right.indicies[i] + k)];
+        right_quadratic << '\t' << weights[(right.indicies[i] + k + rank)];
     }
     right_linear << endl;
     right_quadratic << endl;


library/gd_mf_weights.cc
```diff
@@ -94,11 +94,11 @@ int main(int argc, char *argv[])
     // write out features for right namespace
     features& right = ec->feature_space[right_ns];
     for (size_t i = 0; i < right.size(); ++i)
-    { right_linear << right.space_names[i].get()->second << '\t' << weights[left.indicies[i]];
+    { right_linear << right.space_names[i].get()->second << '\t' << weights[right.indicies[i]];
 
       right_quadratic << right.space_names[i].get()->second;
       for (size_t k = 1; k <= rank; k++)
-        right_quadratic << '\t' << weights[(left.indicies[i] + k + rank)];
```
ataymano: Oh no, sorry for confusion in issue discussion.
Looks like '+ rank' should stay here.
@JohnLangford , could you please confirm? (does it mean that interactions are not symmetric in case of factorization?)

ataymano: confirmed with John offline.
factorization weights are written sequentially with the right ones shifted by rank to avoid collisions - please see [mf.cc](https://github.com/VowpalWabbit/vowpal_wabbit/blob/45a8a7f5507015a17d585f7698eab20c963b592b/vowpalwabbit/mf.cc#L86) for details.
So, could you please revert '+ rank' removal?

ataymano:
Looks good, thank you.

Issue #1858, opened 2019-05-01
Generate Python docs with Sphinx and host on ReadTheDocs.io


jackgerrits:
The python bindings are decently documented with doc strings. It would be useful to be able to access those in the normal python docs way. 

ReadTheDocs.io and Sphinx seem to be the standard way.

jackgerrits:
For the time being the docs will be hosted statically on github pages. Once the python module install story is improved we can revisit using ReadTheDocs as it seems to require installing the module before generating docs

PR #1884, opened 2019-05-17
Add python documentation generation

jackgerrits:
#1858


jackgerrits:
Added it to #1714 to track fixing this as well as in other places

PR #1892, opened 2019-05-24
fix to make invert_hash work with bfgs (issue number 1891)

agroh1:
Fixes issue where models trained with bfgs could not invert the hash.

I copied the code from ftrl.cc which does do this correctly.  

There is also code in ftrl that does this for multipredict that I did not copy.  I was not sure when that was used.

### DIFF
--- a/vowpalwabbit/bfgs.cc
+++ b/vowpalwabbit/bfgs.cc
@@ -943,12 +943,16 @@ void end_pass(bfgs& b)
 }
 
 // placeholder
+template <bool audit>
 void predict(bfgs& b, base_learner&, example& ec)
 {
   vw* all = b.all;
   ec.pred.scalar = bfgs_predict(*all, ec);
+  if (audit)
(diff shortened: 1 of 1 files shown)


jackgerrits:
It does seem like the other base reductions are doing this operation in their predict function too.

@JohnLangford @marco-rossi29 Are you aware of any differences in the way invert_hash is done between 8.5.0 and 8.6.0? Was it previously done in one of the GD::* calls that are being done here?

JohnLangford:
invert_hash is essentially done via the audit pipeline.  It's always functioned in this way.

jackgerrits:
This block is present in the predict function of all base learners except bfgs though?
```
  if (audit)
      GD::print_audit_features(*(b.all), ec);
```

JohnLangford:
This is an inconsistency.

-John

jackgerrits:
Sounds good, going to merge this in as it fixes that.

PR #1896, opened 2019-05-30
Fix resource.rc encoding

lokitoth:
Some time ago we changed the *.rc files encoding to request git to set it to the proper encoding: UTF-16LE. Unfortunately, the autocrlf breaks this conversion, which breaks the version stamping. The fix is to admit that git does not deal with UTF-16 in a functional way, and just treat resource files as binary.


jackgerrits:
Thanks for fixing this Jacob!

PR #1897, opened 2019-06-02
strict parse flag

palmerlao:
One of the artifacts we use is the output of VW on `stderr`. When lots of examples are malformed, VW happily prints a large amount of warning messages (which individually can be large) to disk. This ends in a slow (VW continues to look at the whole dataset) and strange failure - VW keeps printing warning messages, and the job appears to die because it ran out of disk space.

This PR adds a `--strict_parse` flag, which turns parser warnings into exceptions when set.

Remaining work:
- I'm not sure what the preferred practice is for adding a test to check failure since `RunTests` seems to expect all its VW invocations to exit with code 0.


JohnLangford:
To test, I'd recommend creating a script that invokes VW and negates the standard definition of failure, then invoking the script as a test.    (There are several test cases in RunTests invoked by script.)

JohnLangford:
@rajan-chari @jackgerrits have you had a chance to look at this?  Your thoughts?

rajan-chari:
Looking.
    On Monday, June 3, 2019, 9:07:38 AM EDT, John <notifications@github.com> wrote:  
 
 
@rajan-chari @jackgerrits have you had a chance to look at this? Your thoughts?

jackgerrits:
Will take a look today

test/RunTests
```diff
@@ -1710,4 +1710,16 @@ printf '3 |f a b c |e x y z\n2 |f a y c |e x\n' | {VW} --oaa 3 -q ef --audit
...
+# Test 191: malformed examples, onethread, strict_parse failure
+./negative-test.sh {VW} -d train-sets/malformed.dat --onethread --strict_parse
+    train-sets/ref/malformed-onethread-strict_parse.stderr
+
+# Test 192: malformed examples, strict_parse failure
+./negative-test.sh {VW} -d train-sets/malformed.dat --strict_parse
+    train-sets/ref/malformed-strict_parse.stderr
+
+# Test 193: malformed examples success
+{VW} -d train-sets/malformed.dat --onethread
```
palmerlao: This test didn't pass for me without `--onethread`; my guess is there is some non-determinism in printing `parserWarning` messages from the parse thread and printing progress messages?

JohnLangford:
I suspect the problem is that the test, because it uses a script, does not run properly on windows.  It needs to be disabled, plausibly here: https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/cs/unittest/RunTests.tt#L27  Doable?

test/RunTests
JohnLangford: Yes.  The parser thread runs asynchronous of the learning thread unless you use --onethread.

test/negative-test.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+# this script runs a command and ensures that its exit code is not zero.
+
+if $@; then
```
jackgerrits: I get the idea behind this, but this gets us further from parity on Windows. Probably a no-op for this PR though

vowpalwabbit/parse_example.cc
```diff
@@ -76,9 +77,17 @@ class TC_parser
...
   {
-    cerr << message << std::string(begin, pos - begin).c_str() << message2 << "in Example #"
-         << this->p->end_parsed_examples << ": \"" << std::string(this->beginLine, this->endLine).c_str() << "\""
-         << endl;
+    if (p->strict_parse) {
+      std::stringstream exc_msg;
+      exc_msg << message << std::string(begin, pos - begin).c_str() << message2 << "in Example #"
+              << this->p->end_parsed_examples << ": \"" << std::string(this->beginLine, this->endLine).c_str() << "\""
+              << endl;
+      throw std::domain_error(exc_msg.str());
```
jackgerrits: I'd prefer this to be a VW specific exception of some type like `strict_parse_error`

vowpalwabbit/main.cc
```diff
@@ -131,6 +131,9 @@ int main(int argc, char* argv[])
 
     for (vw* v : alls)
     {
+      if (v->p->exc_ptr) {
+        std::rethrow_exception(v->p->exc_ptr);
```
jackgerrits: I'm not sure I follow why the exception needs to be stashed and rethrown? Why can it not just bubble up out of the parser?

palmerlao: I was under the impression that this is slightly safer. Don't exceptions that are not caught in `std::thread` just result in the program being `std::terminate`d?

vowpalwabbit/parse_example.cc
palmerlao: I've tried this in fede09b. Do you think it should be a subclass of `vw_exception`?

JohnLangford:
Why is #193 failing in appveyor?

rajan-chari:
When we discussed exception policy last we decided to switch to returning errors codes in the hot path.  Consistent with that, given that parsing examples is a high frequency operation, the preference is to return error codes instead of throwing exceptions.

Having said that throwing exceptions is still much better than writing to stderr.

vowpalwabbit/main.cc
jackgerrits: Yes they do, but you'll notice there are two catch statements in main which make sure the exception info is printed before exiting. I think letting the exception bubble up is the most idiomatic way of handling this. It also makes library use of the function much simpler

jackgerrits:
But in this case the point of the PR is to stop execution if a parse warning is encountered, so speed/hot path is less relevant

vowpalwabbit/parse_example.cc
jackgerrits: Yep, in the same way the other exception types in vw_exception.h are

JohnLangford:
If "throw an exception" is the preferred behavior (as indicated by a flag), it seems reasonable to allow an exception?

vowpalwabbit/parse_example.cc
```diff
@@ -76,9 +76,15 @@ class TC_parser
...
   {
-    cerr << message << std::string(begin, pos - begin).c_str() << message2 << "in Example #"
-         << this->p->end_parsed_examples << ": \"" << std::string(this->beginLine, this->endLine).c_str() << "\""
-         << endl;
+    std::stringstream ss;
+    ss << message << std::string(begin, pos - begin).c_str() << message2 << "in Example #"
+       << this->p->end_parsed_examples << ": \"" << std::string(this->beginLine, this->endLine).c_str() << "\""
+       << endl;
+    if (p->strict_parse) {
+      throw VW::strict_parse_exception(ss.str());
```
rajan-chari: A couple of concerns with this behavior.

1) There will be two substantially different code paths if exceptions/errors are enabled vs not.  While this was probably an acceptable tradeoff for memory allocation issues, this is not desireable in general.  For example we cannot recover and continue processing examples.

A couple of ways to avoid this issue:
a) Always throw.  Stop parsing the current example.  Let the outer loop either terminate or recover.  This would require some modifications so that the parser can continue parsing.
b) Always return error.  Stop parsing the current example.  Let the outer loop terminate or continue.
c) Accumulate the errors and continue parsing.  Continue or terminate in the outer loop based on a command line flag

2) If it is desireable to recover and continue parsing, then the performance impact of throwing compared to error codes is substantial.  (For example: https://github.com/rajan-chari/perf_tests) In high density type applications, this would require substantial CPU overallocation to be safe.  

I get that this might fall outside the scope of this PR.

lokitoth:
The Windows side of the 193 test is failing because "onethread" is defined in vw (the executable) not vw_core (the library). The C# bindings in Windows wrap vw_core, which does not include onethread. This causes the exception to be thrown.

vowpalwabbit/parse_example.cc
JohnLangford: I'm not entirely clear on (1) in this case.  In particular, adding an exception in the use case that @palmerlao has in mind is a slightly more verbose form of exit() in my understanding.

If we had a global "terminate on error" option for VW, then none of the performance impact is relevant?  (The goal of such an option is fast debugging of a data pipeline.)

cs/unittest/RunTests.tt
```diff
@@ -22,7 +22,7 @@ var skipList = new[] { 13, 32, 39, 40, 41, 59, 60, 61, 66, 68, 90,
 	14, 16, 17, 31, 33, 34,53, 101, 102, 103, 105, 106, 111, 112, // float delta
 	71, // --examples to test parser
 	143, 144, 146, 158, 189, // native json parsing 
-	149, 152, 156, // bash script
+	149, 152, 156, 191, 192, // bash script
 	188 // possibly float delta
```
JohnLangford: Given that --onethread doesn't work in the windows tests, we should just add an exception for 193 as well.

vowpalwabbit/parse_example.cc
rajan-chari: Clarifying (1)
As an application, terminate_on_error() makes sense.  As a library, I am not so sure.  It would be good to have continue_on_error() for example.  The exception leaves the parser in an undefined state.  It would be better to have some mechanism for letting the error bubble up.  exit_on_error() or continue_on_error() can be decided higher up the stack.

JohnLangford: I agree. 

The present default behavior is warn-and-continue.  This adds an option for terminate_on_error().  

In the longer term, having the warnings bubbled up for a later decision does seem better.  Probably beyond the scope of this pull request though?

rajan-chari: Yes makes sense.  It's a larger change.  We would likely need this for atomization.  I would like to revisit this usage of exceptions at that point if there are no objections.  I will add an issue to track this.

JohnLangford:
Sounds good.

cs/unittest/RunTests.tt
lokitoth: I would put it under a different category, though: For example, "executable-only feature used".

test/negative-test.sh
lokitoth: We may want to have the negative test be a RunTests-level concept: In other words, like we specify stderr/predictions to check, the test should be able to specify if it is expecting a successful or a failure exit. With that said, this would not address the issue of existing shell script-based tests.

JohnLangford:
Added an exception in #1906

palmerlao:
I'll close this since it seems like work is continuing in #1906, but I'm happy to contribute (albeit at a very intermittent rate on the weekdays) so let me know if I can help. Thanks!

PR #1898, opened 2019-06-03
memory leaks and warnings

JohnLangford:
Fixes several memory leaks and warnings.


vowpalwabbit/cbify.cc
```diff
@@ -88,15 +88,19 @@ void finish(cbify& data)
...
+
   if (data.use_adf)
   {
     for (size_t a = 0; a < data.adf_data.num_actions; ++a)
-    {
-      data.adf_data.ecs[a]->pred.a_s.delete_v();
-      VW::dealloc_example(CB::cb_label.delete_label, *data.adf_data.ecs[a]);
-      free_it(data.adf_data.ecs[a]);
-    }
+      {
```
lokitoth: [Minor][Nit] Is the extra indentation intended?

vowpalwabbit/cbify.cc
```diff
@@ -467,6 +477,7 @@ base_learner* cbifyldf_setup(options_i& options, vw& all)
 
   data->app_seed = uniform_hash("vw", 2, 0);
   data->all = &all;
+  data->use_adf=true;
```
lokitoth: Out of curiosity, what is the impact of not setting this?

vowpalwabbit/parse_example_json.h
```diff
@@ -314,11 +314,8 @@ struct LabelState : BaseState<audit>
 
   BaseState<audit>* StartObject(Context<audit>& ctx) { return ctx.label_object_state.StartObject(ctx); }
 
-  BaseState<audit>* String(Context<audit>& ctx, const char* str, rapidjson::SizeType /* len */, bool copy)
```
lokitoth: I think this copy is part of the API required by the rapidjson parser: http://rapidjson.org/md_doc_sax.html#CustomDataStructure. 

I am guessing this is what is causing the parser failures in the unit-tests.

vowpalwabbit/parser.cc
```diff
@@ -999,6 +999,17 @@ void free_parser(vw& all)
     output->currentname.delete_v();
   }
 
+  while (! all.p->example_pool.empty())
```
lokitoth: Is it possible to get into a weird state where parser is deallocated, but there is still a thread trying to pass some of these through learn?

(Do we define that as improper consumption of the library?)

vowpalwabbit/parse_example_json.h
lokitoth: Looks like you were way ahead of me.

vowpalwabbit/cbify.cc
JohnLangford: No.  

The problem with styles, in my experience, is that every editor has a different interpretation of precisely what a style is...

vowpalwabbit/cbify.cc
JohnLangford: The finish() doesn't work properly.

vowpalwabbit/parser.cc
JohnLangford: It shouldn't happen given the way shutdown occurs.

JohnLangford:
There's a not-locally-reproducible problem in dsjson tests.  Ideas?

vowpalwabbit/parser.cc
jackgerrits: Long term it would be better if the pool took a deallocate function and used that when its destructor is called

PR #1899, opened 2019-06-03
Change version-generation for C# bindings to use version.txt

lokitoth:
The version stamping strategy employed by the C# bindings is based on a hard-coded string in version.props. This is very easy to get out of sync with the version.txt file that the native build uses.

This change creates logic to determine VowpalWabbitAssemblyVersion by reading it from version.txt, and updates the projects that get built by the CI to use it.

Updating the Azure and Setup projects is out of scope:
* Azure because they are deprecated and not being built
* Setup because they have not been maintained well and is not something we expect consumers of VW to build; for v.Next (after this current push for ICML) I will create a proper setup-generation pipeline.


JohnLangford:
I'm confused about how this patch works---it looks like version is hardcoded in multiple locations?

lokitoth:
The hard-coded places are in AssemblyInfo.c(s|pp) and Resource.rc; these files get updated by the version generation script. In other words - if you change version.txt and run a build, it will update the files before compiling them.

Longer-term we want to get rid of these files from being checked into the build, but making that change would take longer, and I do not want to disrupt the release. This at least ensure that any assemblies built will always have the canonical version in version.txt, regardless of what the checked-in hardcoded files say.

JohnLangford:
Ok, merged.  Thanks :-)

PR #1658, opened 2018-10-29
Fix explore library and change rl_clientlib to use explore lib to swap actions

rajan-chari:
1) sample_after_normalizing will always normalize pdf
2) remove sample_after_normalizing that takes scores
3) add swap_chosen() to swap chosen action with first action
4) added unit tests
5) refactored ranking_response.  added a const iterator.
6) rl_clientlib will use swap from explore library to switch action ordering

PR #1902, opened 2019-06-04
C# Exploration

lokitoth:
* Rebases #1459 onto master
* Integrates C# exploration into new restore/build/test pipeline
* Pulls in fix from #1658 to ensure that sampling code does not leave PDF in partially normalized state


JohnLangford:
What is the status of blackbox testing?

@jackgerrits @eisber comments?

cs/explore/ExplorationStrategies.cs
```diff
@@ -0,0 +1,152 @@
+using System;
```
JohnLangford: Structurally, I was imagining grouping this by functionality (exploration) rather than by language (cs).  Should we make this a subdirectory of /explore instead?

lokitoth: Happy to move it, my only concern would be that we would get an odd duplication of hierarchies, since we would probably want to put a /cs/ folder under /explore?

lokitoth:
Black Box testing status: I worry that I will not have time to get to building the blackbox tests (and constructing testing criteria) before Firday.

jackgerrits:
I have not worked on the black box testing for this either unfortunately

cs/explore/ExplorationStrategies.cs
JohnLangford: explore/cs seems preferable to cs/explore since it will help us keep these in sync.  Furthermore, I expect we will refer people to the directory based on functionality more than language.

Although git lacks links, perhaps leaving a small file at cs/explore pointing to explore/cs is a reasonable way of handling this?

JohnLangford:
W.r.t. blackbox testing....

The immediate question is: Is it valuable to have the C# exploration in the master branch vs. a patch without testing?  I'm inclined towards 'no' at the moment.  Having it in a patch means it will be perceived as a little bit questionable which seems about right given the lack of testing.   

Given this, I'm inclined to leave it as a patch until we get testing working.

eisber:
Did you guys look at this: https://github.com/lokitoth/vowpal_wabbit/blob/dev/cs_exploration/cs/explore/TestDriver.cs ?

Not sure where the invocation is though…

lokitoth:
@eisber I did; the additional work is about building a test suite that runs black box against the C library as well as C#.

jackgerrits:
Closing for now. @lokitoth said he will recreate when ready.

PR #1903, opened 2019-06-04
Coin betting

bremen79:
I have implement the coin betting algorithm.

The algorithm used is the one in COCOB paper, with a different methodology to deal with increasing Lipschitz constants. This approach seems to be the best one among the ones I tried.

Note that the default alpha and beta parameters are good for the logistic loss, while for the square loss they must be bigger, like 16 and 4 respectively. 

Note that there is no "subtraction of the mean" here, it is still unclear how to run that efficiently.


JohnLangford:
Can you add a test case for it as well?

bremen79:
I added the train/test files.
I was adding the save/resume test as well, but it fails, as it fails for FTRL-Proximal as well. Is it a known issue with the save of FTRL models?

JohnLangford:
I vaguely recall "yes", but perhaps you can fix it?

bremen79:
Let me try

JohnLangford:
This: https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/test/save_resume_test.py#L156 
says that the save_resume failure is a known issue.  

However, it's not essential for this patch, so I'll merge and open an issue.  If you can fix it tomorrow, just send a separate pull request.

bremen79:
Sounds good

PR #1906, opened 2019-06-04
Optional exception

JohnLangford:
Tweaked #1897


jackgerrits:
I do not think rethrowing with an exception pointer is the idiomatic way to do this

vowpalwabbit/vw_exception.h
```diff
@@ -105,6 +105,15 @@ class vw_unrecognised_option_exception : public vw_exception
   ~vw_unrecognised_option_exception() _NOEXCEPT {}
 };
 
+class strict_parse_exception : public std::exception
```
jackgerrits: Should inherit from vw_exception

JohnLangford: That makes some sense.

JohnLangford:
Can you use the command line instructions to create your own variation on this patch?  (I'm busy with kids right now.)

jackgerrits:
Yeah, sure thing

jackgerrits:
So I made the changes, but the print statement used for the vw_exception prints the file and line number of the exception. This changes depending on filesystem on every machine. Not sure how to resolve that here.

vw (/mnt/c/w/repos/vowpal_wabbit_jackgerrits/vowpalwabbit/parse_example.cc:85): malformed example! '|',space, or EOL expected after : "| x:0.7"in Example #0: "| x:0.7"

JohnLangford:
See comment here: https://github.com/jackgerrits/vowpal_wabbit/commit/f42678d4e57ddcdccaba7cacc5001cb920b89d12#diff-5e0723881d65ba028f55a90faa0f6d93R112

Also note: it looks like you commented out example dealloc -> create memory leak.

jackgerrits:
I just realized changing the negation script was wrong -- please wait to merge

test/negative-test.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+# this script runs a command and ensures that its exit code is not zero.
+
+if $@; then
+    exit 0
```
jackgerrits:
```suggestion
    exit 1
```

test/negative-test.sh
```diff
@@ -0,0 +1,9 @@
...
+else
+    exit 1
```
jackgerrits:
```suggestion
    exit 0
```

JohnLangford:
Thanks @palmerlao @jackgerrits @rajan-chari   

@rajan-chari opening an issue around how we handle errors in parsers seems like a good idea.

PR #1907, opened 2019-06-05
Update to vw_exception

jackgerrits:
I realized that the rethrowing is necessary since it occurs on the parse thread, and so to make it to the main thread and exit in that same way then we stash+rethrow.

Note: this is merging into optional_exception, then you can merge that one.


jackgerrits:
I actually need to check this - hang on for merging

PR #1799, opened 2019-03-20
Contextual Memory Tree

wensun:
contextual memory tree (CMT) implementation with scripts located under demo/memory_tree

PR #1908, opened 2019-06-05
Cmt bump

JohnLangford:
This is #1799  For unclear reasons, appveyor didn't like the little changes @lokitoth proposed, so I'm bumping the build.


jackgerrits:
Appveyor has a restart build button if that's what you are doing

PR #1910, opened 2019-06-05
fix closing invalid file descriptor with memory_io_buf

jrmarkle:
This is a small fix to an old PR of mine (#1189). When using memory_io_buf the parent close will attempt to close(-1). It didn't seem to cause any problems but I recently noticed this reported in valgrind output. This change overrides close_file to not make the close() syscall.


JohnLangford:
Merged, thanks.

PR #1912, opened 2019-06-05
Fix to save the state of FTRL models

bremen79:
I have made a fix to save the state of FTRL models. Currently the state is not saved at all because it calls the the gd save function that doesn't save the state unless it finds a gd structure, that is absent in ftrl models.

I have added a integer in the vw structure the indicates the size of the ftrl model to save and changed read/write of gd accordingly.

Note that the saving for Pistol and Coin betting still does not work: It correctly save the state, but the test fails. Not sure why, I think it is the loss in precision due to the saving and the exponential nature of these algorithms. On the other hand, the saving state for FTRL Proximal now works perfectly.


test/save_resume_test.py
```diff
@@ -153,8 +153,9 @@ def do_test(filename, args, verbose=None, repeat_args=None, known_failure=False)
         errors += do_test(filename, '--loss_function logistic --link logistic')
         errors += do_test(filename, '--nn 2')
         errors += do_test(filename, '--binary')
-        errors += do_test(filename, '--ftrl', known_failure=True)
+        errors += do_test(filename, '--ftrl')
         errors += do_test(filename, '--pistol', known_failure=True)
```
JohnLangford: Are pistol and coin still in known failure?

vowpalwabbit/global_data.h
```diff
@@ -565,6 +565,7 @@ struct vw
   bool adaptive;            // Should I use adaptive individual learning rates?
   bool normalized_updates;  // Should every feature be normalized
   bool invariant_updates;   // Should we use importance aware/safe updates
+  uint32_t ftrl_size;
```
JohnLangford: Could we pass this as an argument to save_load instead?  That seems more elegant than sticking it in the global data structure.

JohnLangford:
The build 67 failure is fixed in the current patch that I'm working on.

vowpalwabbit/global_data.h
bremen79: Done.

test/save_resume_test.py
bremen79: Yes. As I write above, the data structure is saved correctly, but the weights seem to be different if trained continuously or trained, saved, resumed. The difference does not happen immediately, but after some samples. This is why I suspect numerical issues.

JohnLangford: This violates my understanding of computers so I suspect there is something that we're missing.  I probably won't be able to debug before the next release, but it should get done...

bremen79: I am trying to debug this issue a bit more

vowpalwabbit/global_data.h
JohnLangford: Oops, I merged.   Should I revert, or do you want to do a separate pull request?

bremen79: Found the bug, not sure what best way to fix it, see mail

PR #1913, opened 2019-06-05
fix static library build

jrmarkle:
The static library cmake build (`-DSTATIC_LINK_VW=On`) was broken recently.


JohnLangford:
@jackgerrits your comment?

jackgerrits:
What is the use case where this should not be a dynamic lib? This target is explicitly named as dll and designed to be used that way

jrmarkle:
I've been using libvw_c_wrapper.a to build a static binary (on Ubuntu 16.04). Using `cmake -DSTATIC_LINK_VW=On` worked for this purpose until recently. Right now setting that option causes the build to fail for me because the linker tries to combine static boost libraries into a dynamic VW library.

jackgerrits:
Is that binary using C or C++? If it is C++ I would recommend linking against the vw target which outputs libvw.

So I guess my main question is are the C interfaces defined in vwdll needed or is it suitable to use the core lib?

jrmarkle:
I'm actually using [cgo](https://golang.org/cmd/cgo/) so I do need a C wrapper.

jackgerrits:
Ah I see, thanks for helping me understand what you need. This looks good to me then.

PR #1928, opened 2019-06-07
Python distributions

peterychang:
Rearranging the repo to make python distributions work
setup.py and all its supporting metadata files need to be at the common ancestor directory for all files that will be included in the source distribution


peterychang:
@gramhagen @Scott-Graham-Bose Does this change look sane to you? I was trying to find a way to keep most/all the python specific files in the python/ subdirectory, but it seemed like setup.py makes a lot of assumptions that can't be changed. Do you know if thats correct?

setup.py
```diff
@@ -75,12 +76,18 @@ def build_cmake(self, ext):
 
         # example of cmake args
         config = 'Debug' if self.debug else 'Release'
+        
+        boost_root = ''
+        if 'CONDA_PREFIX' in os.environ and not 'BOOST_ROOT' in os.environ:
```
peterychang: boost has some issues with anaconda. If we're using anaconda and BOOST_ROOT isn't already set, point it at the conda env

gramhagen: this looks good, though a slightly simpler way might be to do this after cmake_args is defined, then:
```
if 'CONDA_PREFIX' in os.environ and 'BOOST_ROOT' not in os.environ:
    cmake_args.append('-DBOOST_ROOT={}'.format(os.environ['CONDA_PREFIX']))
```

peterychang:
Confirmed this change will allow us to make source distributions and wheels for
linux: python2.7, 3.5, 3.6, 3.7
windows: python3.7 (limited to 3.7 due to vcpkg)

gramhagen:
that's excellent, I think there was some concern about statically linking boost when building wheels so they would be more resilient to different versions / installations. I don't know where that stands relative to this, @jackgerrits do you have comments on how that should work with these changes?

peterychang:
with a little bit of setup when building the weels, vcpkg will also let us build/release binary distributions for python3.6 on Windows (use vcpkg commit aa095559917a495b160986e9ad50556431509ace).

vcpkg has never supported python3.5, so that one's out
vcpkg has never supported boost-python for python2.7, so that one's also out
Switching to pybind11 would let us build against python2.7 on Windows. But with support for python2 ending in a few months, I don't think it's a high priority.

peterychang:
One bit of weirdness my vcpkg changes make though, is installing a VW wheel requires a different command from building/installing VW from the source distribution.

Via wheel: pip install vowpalwabbit
Via sdist: pip install --global-option="--vcpkg-root=path/to/vcpkg" vowpalwabbit

It looks like giving a global-option will cause pip to build from source even if a binary distribution is available

jackgerrits:
Do the wheels work when boost is not installed locally?

peterychang:
Yes. I'm pretty sure linux links against boost statically, so there's no problems there
For windows though, I couldn't get the static linking to work, so its actually a little gross. Installing the wheel will give you the following
```
python/lib/site-packages
    - boost_program_options.dll
    - boost_python.dll
    - pylibvw.pyd
    + vowpalwabbit
        - *.py
```

So the wheel will package the boost library files, but puts them directly into the site-packages subdirectory instead of the vw one

jackgerrits:
Oh I see, that's a bit odd. But if it works perhaps we just want to move
forward with that?

peterychang:
I think that would be ok, maybe with a follow-up task of getting boost statically linking on windows

peterychang:
Moving requirements to setup.py per packaging conventions
https://packaging.python.org/discussions/install-requires-vs-requirements/

> Whereas install_requires defines the dependencies for a single project, Requirements Files are often used to define the requirements for a complete Python environment.

peterychang:
@jackgerrits Actually, I messed up when I was uninstalling boost while testing for requirements on linux.

Boost is dynamically linked in python, and the .so's aren't included. I'll check if we can enable static linking for the linux/python build.

A bigger problem is that pypi doesn't like linux_x86_64 platforms, since there are many flavors of linux. Instead, we'd have to figure out how to create manylinux1 platform wheels

peterychang:
I wasn't aware only an optional part of the python wrapper required the other python packages. Those rightly belong in requirements.txt

peterychang:
It looks like we won't be able to release binary distributions for linux since we would need to build against the manylinux platform. That requires compilation against CENTOS 5, which only supported up to gcc 4.1 (before the inclusion of c++11).

We should be able to build and distribute the windows wheels, but linux users will still need to compile the source distribution

gramhagen:
@peterychang can we build for manylinux2010? That uses CENTOS 6. 
more info here: https://github.com/pypa/manylinux

peterychang:
@gramhagen I've tried getting a build on manylinux2010 as well. The problem I ran into there is that the version of boost available on CENTOS6 is too old for what we need, and attempting to build a newer version is proving to be more problematic than I had anticipated.

We can always revisit this issue as a follow-up task if anything changes in the future

gramhagen:
got it, makes sense. perhaps leveraging conda forge will be more viable for linux

peterychang:
@lokitoth This change is ready for review

JohnLangford:
Let's merge as soon as tests pass.

PR #1930, opened 2019-06-13
Fix clear labels the correct way

### LINKED ISSUE #1821: Inconsistent failure caused by cb examples not being left in good state
#### Describe the bug
I've seen this issue come up in different forms. The latest is when replaying a log file and using more examples than the ring size there is a inconsistent failure caused by the label still containing a cost from the previous usage of the example object.

#### To Reproduce
Steps to reproduce the behavior (On my machine at least):
1. Command: `-d data_example_stale.json --dsjson --cb_explore_adf --cb_adf --csoaa_ldf multiline --csoaa_rank`
2. [data_example_stale.json](https://gist.github.com/jackgerrits/b19eb3a6e890f4a60c373fe24bb28e14)

#### Expected behavior
Training completes successfully

#### Observed Behavior
Exception: `THROW("cb_adf: badly formatted example, only one line can have a cost");`

#### Environment
What version of VW did you use? Latest Master
What OS or language did you use? Linux

jackgerrits:
Fixes #1821

### DIFF
--- a/vowpalwabbit/cb.cc
+++ b/vowpalwabbit/cb.cc
@@ -113,6 +113,7 @@ bool substring_eq(substring ss, const char* str)
 void parse_label(parser* p, shared_data*, void* v, v_array<substring>& words)
 {
   CB::label* ld = (CB::label*)v;
+  ld->costs.clear();
   for (size_t i = 0; i < words.size(); i++)
   {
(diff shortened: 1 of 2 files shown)


JohnLangford:
Looks good, merged :-)

PR #1933, opened 2019-06-14
cs/cli/vw_label.h: Avoid to throw for precision issues

marco-rossi29:


cs/cli/vw_label.h
```diff
@@ -70,9 +70,14 @@ public ref class ContextualBanditLabel sealed : ILabel
 
     void set(float value)
     { if (value < 0 || value >1)
-        throw gcnew ArgumentOutOfRangeException("invalid probability: " + value);
-
-      m_probability = value;
+      {
+        if (value > 1 && value < 1.00001)
```
kumpera: Is 0.00001 documented anywhere as our error margin?
Shouldn't it be a constant to make it consistent across label value operations?

marco-rossi29: Yes, I really like to add a constant somewhere. Where is the right place to do that?

lokitoth: There is a constant.h file - maybe that is the right place?

marco-rossi29: image
Why do we have some many constant.h files? Shouldn't they all be combined? If not, then which one is the "best" constant.h for me to use?

marco-rossi29:
Ready to be reviewed and merged...

vowpalwabbit/constant.h
```diff
@@ -10,6 +10,8 @@ const int cubic_constant2 = 37663;
 const int affix_constant = 13903957;
 const uint64_t constant = 11650396;
 
+const float tollerance = 1e-5f;
```
jackgerrits: Spelling: tolerance

cs/cli/vw_label.h
```diff
@@ -70,9 +71,14 @@ public ref class ContextualBanditLabel sealed : ILabel
 
     void set(float value)
     { if (value < 0 || value >1)
-        throw gcnew ArgumentOutOfRangeException("invalid probability: " + value);
-
-      m_probability = value;
+      {
+        if (value > 1 && value - 1 < tollerance)
```
jackgerrits: Consider extracting out to a float equals function that takes into account precision tolerance

lokitoth:
It looks like the reason build is failing is that other source files (stagewisepoly) also define a top-level constant called `tolerance`. Maybe can name it `float32_tolerance`? Or `probability_tolerance`?

cs/cli/vw_label.h
marco-rossi29: This line is slightly different then an equal function since the tolerance is only on the right side. I.e., it's not abs(value - 1) < tolerance.
Also, it is only used here, so I think abstracting it out it's a little bit overkill for now. Maybe for the future if we are using this over and over in many places. What do you think?

jackgerrits: That makes sense! Yeah let’s leave this

Issue #1840, opened 2019-04-11
Multiple instance mode is not working for cb_adf dsjson


ataymano:
#### Describe the bug
Multiple instances mode is not working for cb_adf dsjson

#### To Reproduce
vw.exe --args args.txt

args.txt:
--cb_adf --dsjson -d file.txt
--cb_adf --dsjson -power_t 0.5 file.txt

#### Expected behavior
Two passes over file.txt with both commandlines

#### Observed Behavior
Error: unrecognised option '--args'

#### Environment
Windows / 8.6.1

jackgerrits:
This issue is fixed by updating line 96 in main.cc
```c++
std::unique_ptr<options_boost_po> ptr(new options_boost_po(l_argc, l_argv));
```

We spoke offline though and you mentioned there are further issues with the driver and multi_ec

jackgerrits:
Would you be able to submit a PR for the partial fix I outlined above? If you'd like I am happy to submit it, just let me know

jackgerrits:
Fixed in #1934

PR #1934, opened 2019-06-15
Multiinstance mode for multiline examples

ataymano:
Multiinstance mode for multiline examples
#1840


vowpalwabbit/cb_adf.cc
```diff
@@ -399,7 +399,7 @@ void finish_multiline_example(vw& all, cb_adf& data, multi_ex& ec_seq)
     output_example_seq(all, data, ec_seq);
     global_print_newline(all);
   }
-  VW::clear_seq_and_finish_examples(all, ec_seq);
```
jackgerrits: Why do we no longer want to clear the sequence?

ataymano: It is moved to learner.cc::multi_example_handler.
clear_seq_and_finish_example did 2 things:
- finish examples (only if example is from certain vw instance's ring)
- clear sequence without any condition.
So,
1) in general, it is potential memory leak (if ec_seq is the only reference to non-ring examples).
2) particularly for multiinstance execution, it breaks the trick with using instance that allocated example as the last one.

ataymano: Also, ec_seq is std vector, so there is no memory management reason for doing explicit clean. We are doing it just because  the same ec_seq instance is used for every multi_line_example. But in this case it is purely piece of driver's business logic.

vowpalwabbit/learner.cc
```diff
@@ -36,207 +30,251 @@ const char* to_string(prediction_type_t prediction_type)
...
+  class custom_examples_queue {
+  public:
+    custom_examples_queue(v_array<example*> examples)
+    : _examples(examples) {}
+    
+    example* pop() {
+      return _index < _examples.size() ? _examples[_index++] : nullptr;
+    }
+  private:
+    v_array<example*> _examples;
```
jackgerrits: Can this be a std::deque or std::list?

vowpalwabbit/learner.cc
```diff
@@ -36,207 +30,251 @@ const char* to_string(prediction_type_t prediction_type)
...
+    multi_ex ec_seq;
+  };
+  
+  class ready_examples_queue {
+  public:
+    ready_examples_queue(vw& master)
+    : _master(master) {}
+    
+    example* pop() {
+      return !_master.early_terminate ? VW::get_example(_master.p) : nullptr;
```
jackgerrits: `VW::get_example` is already a thin wrapper around popping from the queue. If we are refactoring this a bit then I think you can skip using it?

```C++
example* get_example(parser* p) { return p->ready_parsed_examples.pop(); }
```

vowpalwabbit/learner.cc
```diff
@@ -36,207 +30,251 @@ const char* to_string(prediction_type_t prediction_type)
...
+        _context.template process<example, learn_ex>(*ec);
+    }
+  private:
+    context_type _context;
+  };
+  
+  template<typename context_type>
+  class multi_example_handler {
+  private:
+    bool complete_multi_ex(example* ec)
```
jackgerrits: I think I am misunderstanding but I don't see where multi_ex actually get's passed to learn/predict?

vowpalwabbit/learner.cc
```diff
@@ -36,207 +30,251 @@ const char* to_string(prediction_type_t prediction_type)
...
-/* example headers have the word "shared" */
-bool ec_is_example_header(example& ec)
-{
-  v_array<CB::cb_class> costs = ec.l.cb.costs;
-  if (costs.size() != 1)
+    VW::finish_example(all, ec);
+  }
+  
+  /* example headers have the word "shared" */
+  bool ec_is_example_header(example& ec)
```
jackgerrits: I don't think this is going to work once CCB is in. But I have been running and testing with CCB for some time with an equivalent code path...

ataymano: not sure if I get it - this function is not changed at all and CCB PR does not have any changes in learner.cc, so it should have exactly the same version of code for it.

vowpalwabbit/learner.cc
ataymano: It technically can, but the only reason for this class to exist - to use it as adapter between driver and parse_dispatch_loop.h:parse_dispatch. So, it is easier to have the same container here as we fill in par_dispatch. It can be replaced there, but in this case diff is going to expand further (at least to all parser interfaces), so it is probably better to do in separate pr.

vowpalwabbit/learner.cc
ataymano: there are 2 goals of this class:
1) early terminate logic
2) have common interface with custom_queue below.
So I am not sure what is the best way to make it simpler (without broader refactoring)

jackgerrits:
What's the status of this?

vowpalwabbit/learner.cc
```diff
@@ -36,207 +30,251 @@ const char* to_string(prediction_type_t prediction_type)
...
+
+  void drain_examples(vw& all) {
+    if (all.early_terminate) { // drain any extra examples from parser.
+      example* ec = nullptr;
+      while ((ec = VW::get_example(all.p)) != nullptr) VW::finish_example(all, *ec);
+    }
+    all.l->end_examples();
+  }
+  
+  class single_instance_context {
```
JohnLangford: What is the goal with these classes?

ataymano:
> 1. Has valgrind blessed this change?
> 2. You are making a bunch of utility classes, but it seems like you are ending up with a more complex solution.  What is the logic for each of the utility classes?  And should we really have all of them?

1. Yes
2. I do not see easier solution here. Let's discuss offline.

JohnLangford:
This pull request **introduces 1 alert** when merging ff4dee445dd11f1ccf4806b5aa4ded3e52ab7a5b into 2459ec5101bc8dcf1e87a982d2650811683b0781 - [view on LGTM.com](https://lgtm.com/projects/g/VowpalWabbit/vowpal_wabbit/rev/pr-c36e9e4d3742de24664476fcf410d9ad88261a6f)

**new alerts:**

* 1 for Comparison result is always the same

---

**Warning** - Automated code review for VowpalWabbit/vowpal_wabbit will be disabled on October 1, 2019. You can avoid this by <a href="https://github.com/apps/lgtm-com/installations/new">installing the LGTM.com GitHub App</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

PR #1938, opened 2019-06-19
Reuse memory in ccb rather than alloc/dealloc

jackgerrits:
Verified using Valgrind

### DIFF
--- a/vowpalwabbit/conditional_contextual_bandit.cc
+++ b/vowpalwabbit/conditional_contextual_bandit.cc
@@ -9,7 +9,7 @@
 #include "cb_adf.h"
 #include "cb_algs.h"
 #include "constant.h"
-#include "object_pool.h"
+#include "v_array_pool.h"
 
 #include <numeric>
 #include <algorithm>
@@ -21,44 +21,10 @@ using namespace VW;
 using namespace VW::config;
 
 template <typename T>
-struct v_array_initializer
-{
-  v_array<T>* operator()(v_array<T>* array)
-  {
-    *array = v_array<T>();
-    return array;
-  }
-};
-
-template <typename T>
-struct v_array_cleanup
-{
-  void operator()(v_array<T>* array) { array->delete_v(); }
-};
-
-template <typename T>
-using v_array_pool = VW::no_lock_object_pool<v_array<T>, v_array_initializer<T>, v_array_cleanup<T>>;
-
-template <typename T>
-void return_v_array(v_array<T>& array, std::queue<v_array<T>*>& in_use_queue, v_array_pool<T>& pool)
+void return_v_array(v_array<T>& array, VW::v_array_pool<T>& pool)
 {
   array.clear();
-  auto return_item = in_use_queue.front();
-  in_use_queue.pop();
-  return_item->_begin = array._begin;
-  return_item->_end = array._end;
-  return_item->end_array = array.end_array;
-  return_item->erase_count = array.erase_count;
-  pool.return_object(return_item);
-}
-
-// Intentionally returns shallow copy of pooled array
-template <typename T>
-v_array<T> get_v_array(std::queue<v_array<T>*>& in_use_queue, v_array_pool<T>& pool)
-{
-  auto pooled_item = pool.get_object();
-  in_use_queue.push(pooled_item);
-  return *pooled_item;
+  pool.return_object(array);
 }
 
 struct ccb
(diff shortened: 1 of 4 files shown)


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -9,15 +9,58 @@
...
 using namespace LEARNER;
 using namespace VW;
 using namespace VW::config;
 
+template <typename T>
+struct v_array_initializer
+{
+  v_array<T>* operator()(v_array<T>* array)
+  {
+    *array = v_array<T>();
```
JohnLangford: This seems ... confusing?  You pass in a pointer and then you assign to it, and then you return it.  Why assign?  And why pass in anything?  Interface-wise it looks like a memory leak trap.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -9,15 +9,58 @@
...
+};
+
+template <typename T>
+using v_array_pool = VW::no_lock_object_pool<v_array<T>, v_array_initializer<T>, v_array_cleanup<T>>;
+
+template <typename T>
+void return_v_array(v_array<T>& array, std::queue<v_array<T>*>& in_use_queue, v_array_pool<T>& pool)
+{
+  array.clear();
+  auto return_item = in_use_queue.front();
```
JohnLangford: Why do you need an in_use_queue?  Why not just pool.return_object(array)?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -29,6 +72,24 @@ struct ccb
...
+  // place. Therefore, we can't use the pooled object directly but must use the buffer within.
+  // 1. Remove object from pool
+  // 2. Shallow copy into location to be used
+  // 3. Place onto corresponding queue When done:
+  // 1. Shallow copy onto an item from queue
+  // 2. Return pooled queue item to pool
+  //
+  // It doesn't matter which queue object is used to return it as it is essentially just a container
+  // for the internal buffer that we are using.
+  std::queue<v_array<CB::cb_class>*> in_use_cb_label;
```
JohnLangford: I don't understand a fundamental reason why we need in_use_* here.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -529,23 +596,36 @@ void finish(ccb& data)
 {
   data.actions.~vector<example*>();
   data.slots.~vector<example*>();
-  data.origin_index.~map<uint32_t,uint32_t>();
+  data.origin_index.~map<uint32_t, uint32_t>();
```
JohnLangford: What is this map for?  It seems best to avoid map if we can.

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: The object works on pre-allocated memory. This is an initializer for the object, and so the convention would be to do a placement new into the memory. The difficultly is that technically you need to use the return value of the placement new to reference the object not the original pointer. This is why there is a return value.

The expected behavior is to modify the pointer passed in to initialize it and then return it.

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: The pool implementation being used requires that objects returned to the pool originated from the pool. If we did as you suggested it is invalid as it would be a copy of array and therefore a different object and address and not have originated from the pool.

I agree it is cumbersome, a different approach would be to accept returned objects by value and copy them into the pool which has it's own costs. And this is only valid for v_array because of its POD structure over a buffer.

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: Explained in the previous comment but, all objects returned to the pool must have originated from the pool. This tracks those objects so we can return them.

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: Origin index is used to track the original index of each action as they are removed after each iteration. Some sort of book keeping of this type is required.

JohnLangford:
It seems there is a large amount of overhead associated with using the pool object.  

As an alternative, how about we just use a v_array<> as a pool?  Then all the weirdness/overhead goes away?

jackgerrits:
If there are gaps in the pool api then I think that should be addressed rather than running a custom solution here.

JohnLangford:
Can you figure out how to eliminate the need for *_in_use and map?  Can we avoid the super-unintuitive operator()?  All of this is possible with the right structure.

Look here: http://hunch.net/~jl/projects/cover_tree/cover_tree.tar.gz in "stack.h" for the very first implementation of v_array<...> as a part of the cover_tree codebase.  This implementation was used to support a memory pool as careful examination of pop() reveals.  The entire implementation is smaller than this patch.

jackgerrits:
We can definitely fix those two things by providing an implementation that focuses on those constraints, I'll do that now. 

The reason the current implementation has the characteristics it does is because of the requirement to do chunk based allocation for the example pool and is designed to be more general.

JohnLangford:
It looks closer but there is still a map<> to remove?

JohnLangford:
I'm also a little hazy on why we want/need a v_array_pool and modifications to the pool.  (I might be looking at this too quickly...)

jackgerrits:
For the remaining map, there are different ways to achieve this. Such as a vector and erasing items as we go to maintain the mapping. However, I think a map is a good choice here.

We can't alloc and dealloc on each iteration, so we need to reuse. Using pooling is the easiest way to achieve this in my mind.

JohnLangford:
I'm a little bit confused by the diff of a diff.  What is the semantics of origin_index?

jackgerrits:
When actions are passed to cb, some actions are removed so the id/index of each action is different to what it was in the original set of actions. In order to construct the decision response we need the original index, `origin_index` is book keeping so we can get back to the original index easily.

JohnLangford:
So, why not use a v_array instead?  

I expect the computational requirements of map<> are well over an order of magnitude more.

jackgerrits:
Good point, trivial change especially since we never check for non-existent keys

JohnLangford:
Merged, thanks :-)

Issue #1943, opened 2019-06-20
Tau first strategy should count only learn calls (not predict)


marco-rossi29:
Issue:
In [cb_explore_adf.cc](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/cb_explore_adf.cc#L253) tau is decreased no matter if predict or learn is selected

To repro:
Using the C# simulator (which call predict+learn at every iteration) with input:
`simulator.exe "--cb_explore_adf --ignore XA -q UB --first 40000" 10 10 0.03 0.4 1 0 0 50000 10000 0`
Uniform exploration is only done untill 20000 iterations but tau=40000 (the last entry in each line should be around 1000 when uniform exploration is performed - when it's 10000 it means that no exploration is performed) . 
--cb_explore_adf --ignore XA -q UB --first 40000,10,10,1,0,0,0,**10000**,0.0653,990,**990**
--cb_explore_adf --ignore XA -q UB --first 40000,10,10,1,0,0,0,**20000**,0.06625,1996,**1006**
--cb_explore_adf --ignore XA -q UB --first 40000,10,10,1,0,0,0,**30000**,0.178,11996,**10000**
--cb_explore_adf --ignore XA -q UB --first 40000,10,10,1,0,0,0,**40000**,0.235,21996,**10000**
--cb_explore_adf --ignore XA -q UB --first 40000,10,10,1,0,0,0,**50000**,0.26612,31996,**10000**

PR #1944, opened 2019-06-20
Tau first should count only learn examples (not predict)

marco-rossi29:


marco-rossi29:
Solves issue #1943 

I'm open to a different implementation of the fix, but the end result is that data.tau shouldn't be decreased if we are just doing predict

marco-rossi29:
Incidentally, why are we checking:
`data.gen_cs.known_cost.probability < 1 && test_adf_sequence(examples) != nullptr`
in the learn/predict if?
Shouldn't these checks throw an error somewhere else and not force to `predict`?

vowpalwabbit/cb_explore_adf.cc
```diff
@@ -250,7 +250,8 @@ void predict_or_learn_first(cb_explore_adf& data, multi_learner& base, multi_ex&
   {
     float prob = 1.f / (float)num_actions;
     for (size_t i = 0; i < num_actions; i++) preds[i].score = prob;
-    data.tau--;
+    if (is_learn && data.gen_cs.known_cost.probability < 1 && test_adf_sequence(examples) != nullptr)
```
JohnLangford: Why the extra clauses in the 'if'?

marco-rossi29: I copy pasted from line 241. I'm happy to remove them but I wanted to be in sync with how we determine if learn or predict

marco-rossi29: Should the extra clauses be removed also from line 241?

JohnLangford: Likely yes.   "learn" should only be triggered when there is a valid label.

marco-rossi29: `data.gen_cs.known_cost.probability < 1` should be removed since we are not checking this in any other exploration mode
`test_adf_sequence(examples) != nullptr` is checking if the sequence is valid. If we remove it then who is responsible to check if we have a valid label. Do you mean that whoever is calling `predict_or_learn_first` with is_learn=true must be responsible for the validity of the label?

JohnLangford:
This seems clearly better than what we had before, so merging.  If you want to fix the similar issue in cover, that's welcome as well.

PR #1945, opened 2019-06-20
Inject slot id and revise automatic interactions to slot id with all existing namespaces and interactions

jackgerrits:


JohnLangford:
Travis is barfing?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -184,40 +225,29 @@ void remove_slot_features(example* shared, example* slot)
   }
 }
 
-// Generates all combinations of 4th order interactions between namespaces in [shared,history,slot,action]
+// Generates interaction between all namespaces and interactions with the slot id namespace.
```
JohnLangford: The comment is wrong?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -184,40 +225,29 @@ void remove_slot_features(example* shared, example* slot)
...
+    interaction.push_back((char)ccb_id_namespace);
+    new_interactions.push_back(interaction);
+  }
+  vec.insert(vec.end(), new_interactions.begin(), new_interactions.end());
 
-  for (auto shared_index : shared->indices)
+  for (auto& action : actions)
   {
-    for (auto action : actions)
+    for (auto& action_index : action->indices)
```
JohnLangford: I'm structurally confused by the function.  It seems like you should only map slot features into shared features?  Why all the extra work?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -163,6 +168,42 @@ void inject_slot_features(example* shared, example* slot)
   }
 }
 
+void inject_slot_id(ccb& data, example* shared, int id)
+{
+  auto current_index_str = "index"+id;
```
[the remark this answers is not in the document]
jackgerrits: Integer key makes much more sense

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -29,6 +30,10 @@ struct ccb
   std::vector<std::string>* original_interactions;
   std::vector<CCB::label> stored_labels;
   size_t action_with_label;
+
+  // All of these hashes are with a hasher seeded with the below namespace hash.
+  std::map<std::string, uint64_t> hashes;
```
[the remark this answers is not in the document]
jackgerrits: Yes a vector will work

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -163,6 +168,42 @@ void inject_slot_features(example* shared, example* slot)
...
+    data.hashes[current_index_str] = index;
+  }
+  else
+  {
+    index = data.hashes[current_index_str];
+  }
+
+  auto& indices = shared->feature_space[ccb_id_namespace].indicies;
+  auto val_index = std::distance(indices.begin(), std::find(indices.begin(), indices.end(), index));
+  shared->feature_space[ccb_id_namespace].values[val_index] = 0.f;
```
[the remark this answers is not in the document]
jackgerrits: There is no way to remove values from a v_array, the api is unimplemented. So this was a quick way to get the behavior. Should I spend the time to implement remove behavior?

Find is required as it seems indices and values are correlated by index and so I need to find the index to modify.

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: My comment is clumsy, struggled to word it in a way that makes sense.

"Generates quadratics between each namespace and the slot id as well as appends slot id to every existing interaction"?

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: This function is determining all interactions that need to occur. Moving the slot features to the shared example is done with `inject_slot_features`. This specific block is creating the quadratics for actions.

vowpalwabbit/conditional_contextual_bandit.cc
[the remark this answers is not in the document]
JohnLangford: Why isn't the answer "always the last feature"?  That can be removed efficiently without a linear time search.

vowpalwabbit/conditional_contextual_bandit.cc
JohnLangford: Sounds good.

vowpalwabbit/conditional_contextual_bandit.cc
JohnLangford: Ah, good, I understand now.

vowpalwabbit/conditional_contextual_bandit.cc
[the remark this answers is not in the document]
jackgerrits: I /should/ be the last feature, but can we rely on this? If we can, decrementing the end pointer is sufficient

jackgerrits:
I want to see how travis goes after these changes

vowpalwabbit/conditional_contextual_bandit.cc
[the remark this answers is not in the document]
JohnLangford: All lower level reductions should undo whatever changes to the example they make, so yes.  (That's an important contract to have/enforce, precisely for efficiency reasons.)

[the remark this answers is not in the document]
jackgerrits: In that case `pop` works perfectly :)

jackgerrits:
Since this changes the algo, I need to update the test to match the new results

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -163,6 +168,36 @@ void inject_slot_features(example* shared, example* slot)
   }
 }
 
+void inject_slot_id(ccb& data, example* shared, int id)
+{
+  // id is zero based, so the vector must be of size id + 1
+  if(id + 1 > data.hashes.size())
```
JohnLangford: This is minor, but the code is cleaner if you remove 'if', change the next 'if' to do the array bounds check, and then use push_back() inside the next 'if'.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -163,6 +168,36 @@ void inject_slot_features(example* shared, example* slot)
...
+
+  shared->feature_space[ccb_id_namespace].push_back(1., index);
+}
+
+// Since the slot id is the only thing in this namespace, the popping the value off will work correctly.
+void remove_slot_id(example* shared)
+{
+  shared->feature_space[ccb_id_namespace].indicies.pop();
+  shared->feature_space[ccb_id_namespace].values.pop();
+}
```
JohnLangford: It's probably worthwhile to also handle the --audit case both (potentially) when adding the feature and (potentially) when removing it.   People need to be able to understand what is happening...

jackgerrits: Yes audit is definitely worth it, I'll work out how to add that.

JohnLangford: The trick is adding it in a way which does not computationally impact the audit-free pipeline.  On the upside memory allocation is fine in the audit path :-)

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: Yes we could do that with the assumption that id is always increasing one at a time

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: I tried adding audit, but it seems I can't get the printout to actually occur. can you take a look at the most recent commit?

JohnLangford: I don't see a bug here.  Is shared_features_merger working properly with audit?

jackgerrits: I get audit output for action features that are merged with the shared example so I assume not?

jackgerrits: Is it because the value of the namespace is 140

JohnLangford: It should not matter what the namespace is.

JohnLangford: Maybe use debugger/print? (Is namespace 140 somehow being lost from indices?)

jackgerrits: The entire feature_space is lost in the stack somewhere, it's going to take me longer to work out what's going on here

JohnLangford: Good to find this now, obviously...

jackgerrits: The issue was that I didn't append to the example indices, so when the shared example was merged the feature_space was skipped. I've pushed a fix

JohnLangford:
Merged, thanks :-)

PR #1946, opened 2019-06-20
Enable suppressing NuGet version tag for official builds

lokitoth:
This enables running the Build-Windows-NuGet pipeline in "Official Build" mode, which generates a NuGet with an un-tagged, non-prerelease version: `8.7.0` vs. `8.7.0--ac162bffab30636a4a59725cdd7c899fcc4114ea--1277`

It is a NoOp for AppVeyor / non-Windows builds.


JohnLangford:
Merged, thanks.

PR #1947, opened 2019-06-21
CCB Parser needs to fix interaction object in order to function correctly

jackgerrits:


vowpalwabbit/parse_example_json.h
```diff
@@ -1028,6 +1028,8 @@ class CCBOutcomeList : public BaseState<audit>
   BaseState<audit>* old_root;
 
  public:
+  DecisionServiceInteraction* interactions;
```
JohnLangford: What is this exactly?  (And why does it need to be here?)

jackgerrits: The DSJSON format essentially outputs some info in this struct. I am not sure if this is actually still used or not (@yannstad can you comment?). The logic of the parser is that if no actions are seen then it skips the line, so ccb needs to output actions.

In my opinion this business logic should not be in the parser and we should revisit/refactor this parser soon.

PR #1949, opened 2019-06-25
Spark/JNI multipass fixes and AllReduce quiet support

eisber:


vowpalwabbit/allreduce_sockets.cc
```diff
@@ -287,7 +338,7 @@ void AllReduceSockets::broadcast(char* buffer, const size_t n)
       int read_size = recv(socks.parent, buffer + parent_read_pos, (int)count, 0);
       if (read_size == -1)
       {
-        cerr << " recv from parent: " << strerror(errno) << endl;
+        THROW(" recv from parent: " << strerror(errno));
```
jackgerrits: Why change these to throw?

eisber: in reality any silent failure will cause my spark application to hang. throwing will bubble things up

PR #1951, opened 2019-06-27
Add git commit to output of --version

lokitoth:


CMakeLists.txt
```diff
@@ -21,6 +21,24 @@ file(READ version.txt PACKAGE_VERSION)
...
 message(STATUS "VowpalWabbit Version: ${PACKAGE_VERSION}")
 
+# Grab git commitish into variable
+IF(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git)
+  FIND_PACKAGE(Git)
+  IF(GIT_FOUND)
+    EXECUTE_PROCESS(
+      COMMAND ${GIT_EXECUTABLE} rev-parse --short HEAD
+      WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+      OUTPUT_VARIABLE "vw_GIT_COMMIT"
```
jackgerrits: Why lowercase `vw`?

CMakeLists.txt
```diff
@@ -21,6 +21,24 @@ file(READ version.txt PACKAGE_VERSION)
...
+      ERROR_QUIET
+      OUTPUT_STRIP_TRAILING_WHITESPACE)
+    MESSAGE( STATUS "Git Version: ${vw_GIT_COMMIT}" )
+  ELSE(GIT_FOUND)
+    SET(vw_GIT_COMMIT "")
```
jackgerrits: Set variable before `FIND_PACKAGE(Git)` then it doesn't need to be duplicated

CMakeLists.txt
lokitoth: That seems to be how other libraries do it when they want to be able to have the files that define them be includable: <targetname>_<PROPERTY_NAME>.

See https://github.com/microsoft/onnxruntime/blob/master/cmake/onnxruntime.cmake#L30, though they also have it all uppercase. Having it be lower_UPPER_CAMEL does make it easier to segment it when reading quickly.

With all that said, I do not have a strong preference one way or the other, and am happy to change it.

jackgerrits: That's reasonable, happy to keep it how it is

PR #1952, opened 2019-06-28
cd_adf: Added importance weight probability clipping for cb_type mtr, ips, dr

marco-rossi29:


JohnLangford:
What's the goal here?

marco-rossi29:
To be able to clip the important weight in MTR and to be able to use it as a cmd arg when needed.

JohnLangford:
Why limit this to only mtr?

marco-rossi29:
Only implementaton time reasons, but I'm happy to expand to any cb_type as soon as possibile. Implementing it for dr and ips was much more involved than mtr since for dr and ips the probability is used outside of `cd_adf.cc`

JohnLangford:
It seems better to have a relatively universal implementation in the master branch?  Maybe we can discuss what's needed next week.

marco-rossi29:
Sounds good!

vowpalwabbit/gen_cs_example.h
```diff
@@ -11,6 +11,8 @@
 #include "cb_algs.h"
 #include "vw_exception.h"
 
+using namespace std;
```
lokitoth: It looks like you are using explicit std::max throughout your code additions. Is the using still necessary? If not, prefer to remove it.

vowpalwabbit/gen_cs_example.h
```diff
@@ -137,14 +139,14 @@ void gen_cs_label(cb_to_cs& c, example& ec, COST_SENSITIVE::label& cs_ld, uint32
         ((c.known_cost->cost - wc.x) * (c.known_cost->cost - wc.x) - c.avg_loss_regressors);
     c.last_pred_reg = wc.x;
     c.last_correct_cost = c.known_cost->cost;
-    wc.x += (c.known_cost->cost - wc.x) / c.known_cost->probability;
+    wc.x += (c.known_cost->cost - wc.x) / (std::max)(c.known_cost->probability, clip_p);
```
lokitoth: Is there a reason why you sometimes use safe_probability() and sometimes not?

marco-rossi29: I've keep everything as is. Not editing the use of safe_probability()

JohnLangford:
Merged, thanks :-)

PR #1970, opened 2019-07-19
Test and unify usage of ec_is_example_header

### LINKED ISSUE #1968: Investigate is_example_header

jackgerrits:
- Add tests for all variants of ec_is_example_header
- Remove unused and unimplemented LabelDict::ec_is_example_header function
- Add CCB implementation of ec_is_example_header
- Add VW::ec_is_example_header implementation that tests all specific label variants

Fixes #1968


jackgerrits:
@ataymano, I believe this should resolve the question I had about your PR if you use the VW::ec_is_example_header function

vowpalwabbit/parse_args.cc
```diff
@@ -557,6 +557,13 @@ const char* are_features_compatible(vw& vw1, vw& vw2)
 
   return nullptr;
 }
+
+// Compares the header against CB, CCB and CS header formats.
+bool ec_is_example_header(example& ec)
+{
+  return CB::ec_is_example_header(ec) || COST_SENSITIVE::ec_is_example_header(ec) || CCB::ec_is_example_header(ec);
```
peterychang: This is horrifically unsafe. CB and CS are reasonably safe because have exactly the same in-memory layout. However, CCB is totally different, which means this check is undefined behavior

jackgerrits: You're right, it's totally unsafe. What do you think we should do to fix this properly?

peterychang:
Not quite clear on how label_type is set, but I don't see anything obviously wrong with this
lgtm

PR #1973, opened 2019-07-22
Fix some compiler warnings

jackgerrits:


vowpalwabbit/cb_sample.cc
```diff
@@ -40,7 +40,7 @@ void learn_or_predict(cb_sample_data& data, multi_learner& base, multi_ex& examp
     for (size_t i = 0; i < action_scores.size(); i++)
     {
       auto& a_s = action_scores[i];
-      if (a_s.action == labelled_action)
+      if (a_s.action == static_cast<uint32_t>(labelled_action))
```
peterychang: kinda sketchy, but i guess its the behavior we have right now anyways..

peterychang:
some nitpicks regarding int<->uint conversions, but its current behavior so...
lgtm

jackgerrits:
I agree, could do with improvement but the scope of this pr is just to fix the warning

PR #1974, opened 2019-07-22
[ccb] Use a bitset and move it to the stack.

kumpera:


jackgerrits:
#1976 addresses this as well as the other thread_local in ccb

PR #1977, opened 2019-07-25
Remove hard requirement for git during Windows build

### LINKED ISSUE #1975: git is required to compile on windows
Attempting to build VW in visual studio without git installed results in the following error:

Error	MSB3073
The command "win32\make_config_h.exe:VCEnd" exited with code -532462766."

Unhandled Exception: System.ComponentModel.Win32Exception: The system cannot find the file specified
   at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
   at Program.StartGitRevParse()
   at Program.Main(String[] args)

lokitoth:
The fix is to default to empty string when git is unavailable (same as what happens during CMake build)

Fixes #1975


peterychang:
lgtm

PR #1981, opened 2019-07-30
Add Azure pipeline for Linux CI 

jackgerrits:
- Add Linux pipeline definition
- Uses Ubuntu 16.04 as base
- Enables testing for Python 3.6 (as well as existing 2.7)
- When BUILD_JAVA is on, JNI should be required and report an error when not found


.pipelines/build-linux.yml
```diff
@@ -0,0 +1,32 @@
...
+
+pool:
+  vmImage: 'ubuntu-16.04'
+
+container:
+  image: vowpalwabbit/ubuntu1604-build:0.1.0
+  endpoint: DockerHub
+
+steps:
+- bash: ./.scripts/linux/clang-format-check.sh $(Build.SourcesDirectory)
```
lokitoth: Is there a reason why you are relying on passing Build.SourcesDirectory to each of the scripts? It seems like it would make it more difficult to use those scripts locally if they relied on relative structure.

jackgerrits: The scripts accept the repository directory as their only argument so it can know the base path and find the things it needs

lokitoth: Would it be worthwhile to do something similar to how init.cmd detects its location, then sets up the various environment variables that the rest of the build scripts depend on?

From https://stackoverflow.com/questions/59895/get-the-source-directory-of-a-bash-script-from-within-the-script-itself:

`DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"`

jackgerrits: This is a cool solution, I like it!

PR #1984, opened 2019-07-31
Fix implicit fallthrough warning and unused variable warning in GCC

jackgerrits:


vowpalwabbit/gen_cs_example.h
```diff
@@ -144,7 +144,7 @@ void gen_cs_label(cb_to_cs& c, example& ec, COST_SENSITIVE::label& cs_ld, uint32
 }
 
 template <bool is_learn>
-void gen_cs_example_dr(cb_to_cs& c, example& ec, CB::label& ld, COST_SENSITIVE::label& cs_ld, float clip_p=0.f)
+void gen_cs_example_dr(cb_to_cs& c, example& ec, CB::label& ld, COST_SENSITIVE::label& cs_ld, float /*clip_p*/ =0.f)
```
lokitoth: Does the default value still need to be here even though the variable is unused?

explore/hash.h
```diff
@@ -112,14 +112,20 @@ inline uint64_t uniform_hash(const void * key, size_t len, uint64_t seed)
 
   switch (len & 3)
   {
-  case 3: k1 ^= tail[2] << 16;
-  case 2: k1 ^= tail[1] << 8;
+  case 3:
+    k1 ^= tail[2] << 16;
+    // fall through
```
lokitoth: Does the presence of the comment suppress the warning, or does the command have to be exactly `// fall through`? If not the latter, could you add a note about this suppressing the warning, so it does not get removed later?

jackgerrits: Yep! GCC see's this as an annotation and silences the warning. There is also a gnu specific attribute available. In C++17 a [[fallthrough]] attribute was added for this exact purpose. 

I figured a comment is simple and suffices for now. I'll add a description so we don't accidentally remove it.

vowpalwabbit/gen_cs_example.h
jackgerrits: Yes, it allows the function to be called without the last parameter (as is done in existing code). If I remove it the last parameter becomes required.

jackgerrits: Yes, it allows the function to be called without the last parameter (as is done in existing code). If I remove it the last parameter becomes required.

lokitoth: Makes sense.

PR #1985, opened 2019-07-31
Add constexpr and noexcept to some functions, cleanup unused functions

jackgerrits:
Many more functions can be made constexpr when we move to C++14 as the rules were relaxed significantly


vowpalwabbit/topk.cc
```diff
@@ -17,7 +17,7 @@ using scored_example = pair<float, v_array<char>>;
 
 struct compare_scored_examples
 {
-  bool operator()(scored_example const& a, scored_example const& b) const { return a.first > b.first; }
+  constexpr bool operator()(scored_example const& a, scored_example const& b) const { return a.first > b.first; }
```
lokitoth: We have places where we use "function structs" and places where we use bare function pointers. Should we have an issue about deciding on one of these two patterns and unifying?

vowpalwabbit/search.cc
```diff
@@ -628,7 +628,7 @@ void reset_search_structure(search_private& priv)
   if (priv.adaptive_beta)
   {
     float x = -log1pf(-priv.alpha) * (float)priv.total_examples_generated;
-    static const float log_of_2 = (float)0.6931471805599453;
+    static constexpr float log_of_2 = (float)0.6931471805599453;
```
lokitoth: Aside, and well out of scope of this change: We should do an audit of the magic numbers in VW, and see if we can unify some/any of them.

vowpalwabbit/kernel_svm.cc
```diff
@@ -553,15 +553,15 @@ bool update(svm_params& params, size_t pos)
...
 
-void copy_char(char& c1, const char& c2)
+void copy_char(char& c1, const char& c2) noexcept
 {
   if (c2 != '\0')
     c1 = c2;
 }
 
-void add_size_t(size_t& t1, const size_t& t2) { t1 += t2; }
+void add_size_t(size_t& t1, const size_t& t2) noexcept { t1 += t2; }
```
lokitoth: Do we have other places where we do an inplace add into the first argument? If so, can these be unified as a template?

vowpalwabbit/global_data.h
```diff
@@ -168,18 +168,18 @@ struct shared_data
...
-  static const int prec_since_last = 6;
-  static const int col_example_counter = 12;
-  static const int col_example_weight = col_example_counter + 2;
-  static const int prec_example_weight = 1;
-  static const int col_current_label = 8;
-  static const int prec_current_label = 4;
-  static const int col_current_predict = 8;
-  static const int prec_current_predict = 4;
-  static const int col_current_features = 8;
+  static constexpr int col_avg_loss = 8;
```
lokitoth: [Minor][Nit] There are inconsistencies between `static constexpr` and `constexpr static`. Is there a reason for that?

explore/hash.h
```diff
@@ -57,7 +57,7 @@ namespace MURMUR_HASH_3
 //-----------------------------------------------------------------------------
 // Finalization mix - force all bits of a hash block to avalanche
 
-static inline uint32_t fmix(uint32_t h)
+static inline uint32_t fmix(uint32_t h) noexcept
```
lokitoth: Is there a reason why this is not constexpr?

vowpalwabbit/search.cc
jackgerrits: Agreed, now they are easy to find with constexpr :)

vowpalwabbit/kernel_svm.cc
jackgerrits: Yes these simple functions are scattered around. Implementing them in one common place would be good

vowpalwabbit/global_data.h
jackgerrits: No reason, I just replaced const with cosntexpr so they were ordered the other way around before and I didn't notice

explore/hash.h
jackgerrits: In c++11 constexpr functions are **very** restricted. The function body can only contain a return statement and no other statements. This function is 100% constexpr but only when we can use the relaxed rules introduced in c++14

jackgerrits: In c++11 constexpr functions are very restricted. The function body can only contain a return statement and no other statements. This function is 100% constexpr but only when we can use the relaxed rules introduced in c++14

vowpalwabbit/topk.cc
jackgerrits: The functor used here is a requirement I believe for where it is used? Ideally I'd like to see *function* arguments templated so any form of callable can be supplied (functor, lambda, function pointer, std::function)

Issue #1567, opened 2018-08-13
Missing headers in homebrew install


coatless:
When trying to compile using vowpalwabbit's header, provided by installing on macOS using homebrew, a fatal error occurs that branches out from the inclusion of `global_data.h`. The initial error was:

```bash
/usr/local/include/vowpalwabbit/global_data.h:20:10: 
fatal error: 'array_parameters.h' file not found
```

After copying the header file to location, there were then subsequent errors that arose with compilation due to missing header files.  In particular, the following header files required by including `global_data.h` were not found in `/usr/local/include/vowpalwabbit/`:

- `array_parameters.h`
- `array_parameters_dense.h`
- `hash.h` (under `/explore`)
- `no_label.h`
- `example_predict.h`
- `error_reporting.h`
- `parser_helper.h`

I think the reason for the header files missing is because these files are not retained in Makefile list:

https://github.com/JohnLangford/vowpal_wabbit/blob/64ac2f8a1f566bb35c811f4f4d2fc8243b5dc5f6/Makefile.am#L3-L8

Header files provided by homebrew: 

```bash
ls -la /usr/local/include/vowpalwabbit/
-rw-r--r--   1 ronin  admin   1997 Aug 12 20:17 action_score.h
-rw-r--r--   1 ronin  admin   7934 Aug 12 20:17 allreduce.h
-rw-r--r--   1 ronin  admin   1078 Aug 12 20:17 cb.h
-rw-r--r--   1 ronin  admin    320 Aug 12 20:17 cb_explore.h
-rw-r--r--   1 ronin  admin    833 Aug 12 20:17 comp_io.h
-rw-r--r--   1 ronin  admin   2525 Aug 12 20:17 config.h
-rw-r--r--   1 ronin  admin    513 Aug 12 20:17 constant.h
-rw-r--r--   1 ronin  admin   1228 Aug 12 20:17 cost_sensitive.h
-rw-r--r--   1 ronin  admin    472 Aug 12 20:17 crossplat_compat.h
-rw-r--r--   1 ronin  admin   4076 Aug 12 20:17 example.h
-rw-r--r--   1 ronin  admin  11166 Aug 12 20:17 ezexample.h
-rw-r--r--   1 ronin  admin  10430 Aug 12 20:17 feature_group.h
-rw-r--r--   1 ronin  admin    443 Aug 12 20:17 floatbits.h
[...]

PR #1994, opened 2019-08-06
Install all headers as fix for missing headers in installed library

jackgerrits:
This fixes an issue where there are missing headers in the installed library. It is perhaps a bit heavy handed as it installs all headers. This will change once we define a proper public interface for the library.

#1567


peterychang:
lgtm. can you add an issue to clean up the installed headers once we clean up the interface?

jackgerrits:
By default once we create public headers in an include dir then we will need to update the install command, I'm going to hold off on the issue since we know we need the interface

PR #1995, opened 2019-08-06
Implement explicit included actions for CCB

jackgerrits:


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -327,34 +325,43 @@ void build_cb_example(multi_ex& cb_ex, example* slot, ccb& data)
...
-  // for (uint32_t included_action_id : slot->l.conditional_contextual_bandit.explicit_included_actions)
-  //   data.include_list.insert(included_action_id);
+  // Retrieve the action index whitelist (if the list is empty, then all actions are white-listed)
+  auto& explicit_includes = slot->l.conditional_contextual_bandit.explicit_included_actions;
+  if (explicit_includes.size() != 0)
+  {
+    // First time seeing this, initialize the vector with falses so we can start setting each included action.
+    if (data.include_list.size() == 0)
+    {
+      data.include_list.assign(data.actions.size(), false);
```
peterychang: Technically you can just do a resize() since that will default construct the bools (to false). But being explicit is better since this will probably optimize down to the same thing anyways

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -64,8 +64,6 @@ void clear_all(ccb& data)
   data.shared = nullptr;
   data.actions.clear();
   data.slots.clear();
-  // data.exclude_list.clear();
```
peterychang: add a comment why this isn't needed

peterychang:
lgtm

PR #1996, opened 2019-08-06
Move most deletion/cleanup code into parser and vw destructors

jackgerrits:


jackgerrits:
Debugged locally and it seems the failure is transient

jackgerrits:
Actually I do not think it is transient but I am having a lot of trouble reproing locally

jackgerrits:
Closing this until I have time to have another crack

PR #2015, opened 2019-08-12
Add VW-JNI SNAPSHOT publishing to build

eisber:


.scripts/linux/java.sh
```diff
@@ -7,10 +7,10 @@ REPO_DIR=$SCRIPT_DIR/../../
 cd $REPO_DIR
 
 # Run Java build and test
-mvn clean test -f java/pom.xml
+# mvn clean test -f java/pom.xml
```
jackgerrits: Why remove this?

.pipelines/build-linux.yml
```diff
@@ -17,6 +17,10 @@ steps:
   displayName: Test C++
 - bash: ./.scripts/linux/java.sh
   displayName: Build + test Java
+  env:
```
jackgerrits: Can you please add a comment describing:
- What these are for?
- What the publish destination is
- How to obtain the creds

eisber: - these are the credentials used to publish to oss.sonatype.org
- the creds are stored in the build

PR #2020, opened 2019-08-14
Ataymano/memory leaks fixes

ataymano:
Unit tests memory leaks fixes


test/unit_test/ccb_parser_test.cc
```diff
@@ -9,91 +9,127 @@
...
+  p.parse_name.delete_v();
 }
 
 BOOST_AUTO_TEST_CASE(ccb_cache_label)
 {
   io_buf io;
-  io.init();
-  io.space.resize(1000);
-  io.space.end() = io.space.begin() + 1000;
+  //io.init();      TODO: figure out and fix leak caused by double init()
```
jackgerrits: This line can just be removed

jackgerrits:
Thanks for these fixes @ataymano!

In my mind this PR just points out the critical need for RAII types in VW... The fact we can leak so easily by forgetting to call a function is an issue. Destructors fix this automatically

ataymano:
> Thanks for these fixes @ataymano!
> 
> In my mind this PR just points out the critical need for RAII types in VW... The fact we can leak so easily by forgetting to call a function is an issue. Destructors fix this automatically

Oh, I missed your PR about destructors. Will cancel this one then.

jackgerrits:
No no we still need this, I also am not sure if I will be able to fix that PR anytime soon.

test/unit_test/ccb_parser_test.cc
ataymano: I left it for TODO comment

jackgerrits:
Thanks @ataymano!

Issue #2003, opened 2019-08-07
Remove global object (all) - other reductions


rajan-chari:
Pick a reduction from this list and remove 'all' dependency.  This contains list of all reductions.  Add a new task with the desired reduction(s) and track them individually.

- [x] GD (ataymano)
- [ ] kernel_svm
- [ ] ftrl
- [ ] svrg
- [ ] sender
- [ ] gd_mf
- [ ] print
- [ ] noop
- [ ] lda
- [ ] bfgs
- [ ] OjaNewton
 
- [ ] baseline
- [ ] replay<'b', simple>
- [ ] active
- [ ] active_cover
- [ ] confidence
- [ ] nn
- [ ] mf
- [ ] marginal
- [x] autolink (#2047)
- [ ] lrq
- [ ] lrqfa
- [ ] stagewise_poly
- [ ] scorer

- [ ] bs
- [x] binary
- [ ] replay<'m', multi_class>
- [x] topk (#2050)
- [ ] oaa
- [ ] boosting
- [ ] ect
- [ ] log_multi
- [ ] recall_tree
- [ ] classweight
- [ ] multilabel_oaa
- [ ] cs_active
- [ ] CSOAA::csoaa
- [ ] interact
- [ ] CSOAA::csldf
- [ ] cb_algs
- [ ] cb_adf
- [ ] mwt
- [ ] cb_explore
- [x] cb_explore_adf (#2069)
- [ ] shared_feature_merger
- [ ] warm_cb
- [ ] cbify
- [ ] cbifyldf
- [ ] explore_eval
- [ ] replay<'c', cost_sensitive>
- [ ] Search
- [ ] audit_regressor

lalo:
Whoops, did a click by mistake so ignore all those operations.

JohnLangford:
A good idea.  No significant progress has been made in quite some time, so I'm closing this for now, but would welcome patches in this direction.

PR #2047, opened 2019-08-23
Atomize autolink reduction

jackgerrits:
Elevates autolink into a class which the setup function consumes.

#2003


jackgerrits:
@rajan-chari what are your thoughts?

vowpalwabbit/autolink.h
```diff
@@ -1 +1,24 @@
...
+    void predict(LEARNER::single_learner& base, example& ec);
+    void learn(LEARNER::single_learner& base, example& ec);
+
+  private:
+    void prepare_example(LEARNER::single_learner& base, example& ec);
+    void reset_example(example& ec);
+
+    // degree of the polynomial
+    const uint32_t m_d;
+    const uint32_t m_stride_shift;
```
rajan-chari: I have been using _ in most of the cpp code instead of m_.  I like it because it's one letter instead of two.

vowpalwabbit/autolink.h
```diff
@@ -1 +1,24 @@
+#pragma once
+
+#include "reductions.h"
```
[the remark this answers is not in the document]
peterychang: I created a reductions_fwd.h in another PR which forward declares several types and declares a few other typedefs. It contains all the bits that are commonly used in reductions

vowpalwabbit/autolink.h
peterychang: We should have a quick discussion to figure out all our naming conventions. Namespaces, object names, variable names, etc

jackgerrits: In the past I've used m_, s_ and c_ to indicate member, static and constant respectively.

vowpalwabbit/autolink.h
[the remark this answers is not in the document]
jackgerrits: That sounds good, I'll wait until Peter's change is in then switch over to using reductions_fwd.h

vowpalwabbit/autolink.cc
```diff
@@ -1,55 +1,75 @@
...
+  if (is_learn)
+    b.learn(base, ec);
+  else
+    b.predict(base, ec);
+}
+
 LEARNER::base_learner* autolink_setup(options_i& options, vw& all)
 {
-  free_ptr<autolink> data = scoped_calloc_or_throw<autolink>();
+  uint32_t d;
```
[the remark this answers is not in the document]
jackgerrits: Exactly :)

vowpalwabbit/autolink.cc
```diff
@@ -1,55 +1,75 @@
...
-    base.learn(ec);
-  else
-    base.predict(ec);
-
+void VW::autolink::reset_example(example& ec)
+{
+  features& fs = ec.feature_space[autolink_namespace];
   ec.total_sum_feat_sq -= fs.sum_feat_sq;
-  fs.clear();
+  ec.feature_space[autolink_namespace].clear();
```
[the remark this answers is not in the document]
jackgerrits: You're totally right, I missed that

vowpalwabbit/autolink.h
```diff
@@ -1 +1,24 @@
...
+
 LEARNER::base_learner* autolink_setup(VW::config::options_i& options, vw& all);
+
+namespace VW
+{
+  struct autolink
```
JohnLangford: Why do we want this struct in the header file?  If it's for testing purposes, shouldn't there be some tests?

jackgerrits: It's in the header so it can be independently instantiated as part of the library. I don't really understand enough semantically what this reduction does to know what I could test here.

vowpalwabbit/autolink.cc
```diff
@@ -1,55 +1,75 @@
...
+  if (is_learn)
+    b.learn(base, ec);
+  else
+    b.predict(base, ec);
+}
+
 LEARNER::base_learner* autolink_setup(options_i& options, vw& all)
 {
-  free_ptr<autolink> data = scoped_calloc_or_throw<autolink>();
+  uint32_t d;
```
JohnLangford: This seems like an elegance issue because you need to declare temporary variables and shuffle things into the more permanent structure.   Why not declare the struct in advance?

jackgerrits: Putting an integer on the stack is significantly cheaper than allocating a reduction object on the heap. It is best to delegate that to when we need it. We can also then make proper use of the constructor and have constant members in the reduction object set in the constructor.

jackgerrits: Currently we allocate every reduction and for almost all of them delete it right away

jackgerrits:
I'm not sure I follow what parts of the pattern are worrisome?

vowpalwabbit/autolink.cc
JohnLangford: This is computationally meaningless since it is done once per VW run.  (Or once per pipeline.)  

W.r.t. the constructor, I think it's important to keep things simple for people implementing reductions.  Adding setter/getter functions for every element is not simple.

vowpalwabbit/autolink.h
JohnLangford: I believe this is not the object we would want to instantiate.  Instead, it would be the learner which contains autolearn.

JohnLangford:
I'm worried about imposing unnecessary overhead in writing reductions.

vowpalwabbit/autolink.cc
jackgerrits: Sure it may be computationally insignificant compared to the hot predict/learn loop, it is still not needed though

Using a constructor gives us lots of wins in terms of guaranteeing state of objects, this gets us out of the setup/finish situation, where an object can be in an indeterminate state.

You'll notice there are no setters or getters in this PR. All of the state is encapsulated in the reduction itself. It allows us to reason about the state of the system in self contained components.

As far as for the people implementing reductions, the trend of changes we are making forces people to think in more of an object based C++ mindset rather than a more C style mindset.

JohnLangford: W.r.t. 'not needed', we also don't need the extra lines of code.  In general, I think reducing the number of lines of code at negligible computational cost is a good tradeoff.

Using a constructor seems fine, and you can do so here by initializing with d=1.

I agree that having the state encapsulated is great (seems orthogonal to the discussion)?

The general approach of encapsulation which object based C++ provides seems good if we can avoid the downsides.

jackgerrits:
I would agree the pattern is changing, like I said in my other comment it mainly seems to be an artifact of the fact we are embracing C++ more holistically across the codebase. If you are used to writing C then it may be trickier to write a reduction in this style. But we get so many benefits overall, and those benefits in turn actually make it easier to write new reductions.

Having a good developer guide to creating a reduction could help a lot here, but it does assume people know C++.

vowpalwabbit/autolink.cc
jackgerrits: I fail to see why lines of code is an important metric. In my mind speed, safety, correctness and maintainability are what I am trying to achieve. By having more lines here we can construct using the parsed option. This is both faster, and more safe and maintainable because we have a const member.

I'm not sure I follow the comment of initializing d=1. By making d a constructor parameter we can make it const, so that for the rest of the lifetime of that object d will never change.

vowpalwabbit/autolink.h
jackgerrits: A learner contains driver specific functionality, such as outputting and finishing examples. The reduction itself seems to be the core logic that is actually useful outside of the command line.

JohnLangford: I think my worry here is about type structure.  

"A learner is an object that instantiates learn() and predict()" may not be the right structure since you cannot create a collection due to nonuniform typing.

vowpalwabbit/autolink.cc
JohnLangford: This may be a difference in experience, but doing things without unnecessary shuffling seems more understandable (=maintainable?) to me.

Perhaps we should have a conversation in person in more depth.

vowpalwabbit/autolink.h
jackgerrits: Yeah I agree with that. There needs to be some sort of uniform type that can be used to reason here. Rajan, Peter and myself have discussed for some time the ideas of reductions being typed by what they support. (input/output, label type, pred type etc) There is definitely more work to come there, this is just a baby step to help get rid of all

vowpalwabbit/autolink.cc
jackgerrits: Yep I agree, we're probably thinking about this differently and aren't matching up 😄 
I'll schedule something for Monday or Tuesday and we can go through it.

PR #2050, opened 2019-08-23
Atomize topk reduction

jackgerrits:
Also fixes example copying bug.

#2003


vowpalwabbit/topk.h
```diff
@@ -3,4 +3,36 @@ Copyright (c) by respective owners including Yahoo!, Microsoft, and
...
+    struct compare_scored_examples
+    {
+      constexpr bool operator()(scored_example const& a, scored_example const& b) const { return a.first > b.first; }
+    };
+
+    void update_priority_queue(float pred, v_array<char>& tag);
+
+    // rec number
+    const uint32_t m_K;
+    std::priority_queue<scored_example, std::vector<scored_example>, compare_scored_examples> m_pr_queue;
```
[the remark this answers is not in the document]
rajan-chari: Seems like we are only interested in the top element.  A priority queue likely more efficient.  Also it's backed std::vector by default and would require less allocations.  

But as in all things performance... measure?

vowpalwabbit/topk.h
```diff
@@ -3,4 +3,36 @@ Copyright (c) by respective owners including Yahoo!, Microsoft, and
...
+  private:
+    struct compare_scored_examples
+    {
+      constexpr bool operator()(scored_example const& a, scored_example const& b) const { return a.first > b.first; }
+    };
+
+    void update_priority_queue(float pred, v_array<char>& tag);
+
+    // rec number
+    const uint32_t m_K;
```
rajan-chari: _ instead of m_?

vowpalwabbit/topk.h
```diff
@@ -3,4 +3,36 @@ Copyright (c) by respective owners including Yahoo!, Microsoft, and
...
+{
+  struct topk
+  {
+    using scored_example = std::pair<float, v_array<char>>;
+
+    topk(uint32_t K);
+
+    void predict(LEARNER::single_learner& base, multi_ex& ec_seq);
+    void learn(LEARNER::single_learner& base, multi_ex& ec_seq);
+    std::vector<scored_example> drain_queue();
```
rajan-chari: Should this be private?

vowpalwabbit/topk.h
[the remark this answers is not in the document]
jackgerrits: This reduction is not used very much, there is already a priority queue here. I think we should leave as is

vowpalwabbit/topk.h
jackgerrits: It doesn't seem right to link driver related output functions to the class itself. For the output function to work it needs access to the result of the reduction, hence why this function exists.

vowpalwabbit/topk.cc
```diff
@@ -5,51 +5,77 @@ license as described in the file LICENSE.
...
+  else if (m_pr_queue.top().first < pred)
+  {
+    m_pr_queue.pop();
+    m_pr_queue.push(std::make_pair(pred, tag));
+  }
+}
+
+std::vector<VW::topk::scored_example> VW::topk::drain_queue()
+{
+  std::vector<scored_example> result;
```
rajan-chari: Won't this cause allocations for every finish_example?

peterychang: This was the original reason I suggested the multimap. priority queue is fine for the top element, like you mentioned above, but this was bad

jackgerrits: Yep when using a pr it has to copy out into a vector

jackgerrits:
Changed to multimap, now there is no copy to output.

PR #2067, opened 2019-08-28
Remove unnecessary null checks

### LINKED ISSUE #1845: Remove unnecessary null pointer checks when deleting pointers
[An extra null pointer check is not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for null before delete p?") like the following.
- [delete_it](https://github.com/VowpalWabbit/vowpal_wabbit/blob/29c47678aae83cb8a6f64d12e95f63523dccfde2/vowpalwabbit/cbify.cc#L80 "delete_it function")
- [task_data](https://github.com/VowpalWabbit/vowpal_wabbit/blob/d0c9042ca1057c858c8158d0d4b7335a12abd491/vowpalwabbit/search_meta.cc#L80 "Destructor for task_data")

jackgerrits:
Closes #1845


elfring:
Thanks for your small source code improvement.

PR #2071, opened 2019-08-30
Remove all usages of "using namespace std"

jackgerrits:
### Why?
While it seems like it might just be style, there are practical reasons to not do `using namespace std`. The main reason is it makes the implementation of functions used ambiguous. In our codebase `min` and `max` had a definition in v_array.h, a macro in search, a macro in the Windows headers and the std implementations. It very often unclear which implementation was being used, and the semantics actually differed between them. If using the one in v_array.h, then it actually implicitly converted types to size_t. This was the cause of some very obscure bugs hit during development time.

Removing it from headers also important, otherwise you may include a header and actually change the implementations the cc file is using. Noone expects that.

### Changes:
- Remove all `using namespace std;`
- Use (below) to remove the need to surround `std::min` and `std::max` in parentheses.
    ```C++
    #define NOMINMAX
    #include <Windows.h>
    ```
- Remove duplicate implementations of `min` and `max`
    - In v_array.h
    - In search
- All places are now using `std::min`, `std::max`


JohnLangford:
Having a single implementation of min and max makes great sense.  

I'm a little bit less clear on the need to preface std on every std function though.  Can we mitigate this via 'using std::cerr' in files with (say) 5 or more uses of std::cerr?  (Or any other std function?)

W.r.t. my preference for print debugging, I'll just need to suck it up, I think.

jackgerrits:
This may be too heavy handed when it comes out output functions. It is a little easier to just be consistent, but for files with lots of them I can understand the desire for 'using std::cerr'. I'll go through and find the hardest hit places and try and alleviate things here.

Honestly, I think print debugging is great. Is the concern it's more cumbersome to write std::? We have the DBG macro, we can make it work for multiple variable values too to make it easier to print debug?

jackgerrits:
@JohnLangford would you be able to take a look at this iteration? Does it need more working to get the balance right?

library/ezexample_predict_threaded.cc
```diff
@@ -4,14 +4,12 @@
 
 #include <boost/thread/thread.hpp>
```
JohnLangford: Maybe tweak this one with using std::cerr using std:endl?

library/gd_mf_weights.cc
```diff
@@ -6,51 +6,49 @@
 #include <string.h>
```
JohnLangford: This file seems heavy also.

library/recommend.cc
```diff
@@ -14,7 +14,6 @@
```
JohnLangford: Feels heavy again.

library/search_generate.cc
```diff
@@ -8,7 +8,7 @@
 #include "../vowpalwabbit/ezexample.h"
 #include "libsearch.h"
```
JohnLangford: Feels heavy.

library/test_search.cc
```diff
@@ -6,23 +6,23 @@
 #include "libsearch.h"
```
JohnLangford: Feels heavy.

python/pylibvw.cc
```diff
@@ -20,8 +20,7 @@
 #define VWDLL_EXPORTS
 #include "../vowpalwabbit/vwdll.h"
```
JohnLangford: feels heavy.

vowpalwabbit/active_interactor.cc
```diff
@@ -11,6 +11,7 @@ license as described in the file LICENSE.
 #include <cerrno>
 #include <cstdlib>
 #ifdef _WIN32
+#define NOMINMAX
```
JohnLangford: feels heavy.

jackgerrits:
Most of the files you mentioned are in library and not the core codebase, they are not really used for anything.

JohnLangford:
They are certainly less used.

python/pylibvw.cc
jackgerrits: Most of these usages are for `std::string`, bringing that into the global namespace is dangerous because of how ubiquitous the name string is.

jackgerrits:
I've made the requested changes

vowpalwabbit/print.cc
```diff
@@ -2,7 +2,6 @@
 #include "float.h"
 #include "reductions.h"
```
JohnLangford: This file could use some help.

vowpalwabbit/parser.cc
```diff
@@ -17,6 +17,7 @@ license as described in the file LICENSE.
 #include <fstream>
```
JohnLangford: This could use help also.

vowpalwabbit/parse_args.cc
```diff
@@ -83,13 +83,16 @@ license as described in the file LICENSE.
 #include "options_boost_po.h"
```
JohnLangford: This could use help.

vowpalwabbit/kernel_svm.cc
```diff
@@ -7,6 +7,7 @@ license as described in the file LICENSE.
 #include <sstream>
```
JohnLangford: Could use help.

vowpalwabbit/search.cc
```diff
@@ -23,14 +23,10 @@ license as described in the file LICENSE.
 #include "vw_exception.h"
```
JohnLangford: Lots of std::endl here.

vowpalwabbit/parse_args.cc
jackgerrits: We should not bring string, equal, vector into the global namespace. We don't want a repeat of the minmax issues, namespaces guard us against ambiguity

jackgerrits:
I've addressed the latest comments minus one that should not be changed

JohnLangford:
(a) There's a failing test somewhere (?)
(b) Are there any other files where more specific 'using' could be help similarly helpful to those you've tweaked?

jackgerrits:
The failing test seems to have made it in with another PR. Looks like it may be from cb adf @peterychang?

I'll do another pass and see if there are any more files with lots of `std::cout`, `std::cerr` or `std::endl`. Other `std::` symbols should remain namespaced in my opinion though.

peterychang:
The tests are passing on Ubuntu 14, 18, OSX, and Windows, this seems to be a problem exclusively on ubuntu 16

jackgerrits:
Looks to me like the rest is reasonable, can we merge this?

vowpalwabbit/parse_args.cc
JohnLangford: Is this a "real" problem here?  In other words, is string or vector defined anywhere else in the codebase or in the include dependencies?

peterychang: I dont think this is a problem in our code base right now, but common object and function names like string or equal typically get overloaded eventually. Since the version you use is entirely dependent on the linker, that can result in really hard-to-find bugs in unrelated code.

The most common and least ambiguous rule I see is "don't 'using' anything in libraries". But there shouldn't be any issues 'using' unusual symbols like cout or endl

JohnLangford:
Thanks.   Merged the monster :-)

PR #2080, opened 2019-09-10
Add forwarding header for commonly used objects in reductions headers

peterychang:


vowpalwabbit/reductions_fwd.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+// Forward declaration header to declare the basic components used by VW reductions
+
+#include <vector>
+
+// forward declarations
+struct example;
+typedef std::vector<example*> multi_ex;
```
jackgerrits: using?

jackgerrits:
```suggestion
using multi_ex = std::vector<example*>;
```

PR #2081, opened 2019-09-11
Java Binding Improvements

eisber:
- expose perf statistics (e.g. avg loss)
- add additional unit tests


java/src/main/c++/jni_spark_vw.cc
```diff
@@ -135,9 +135,69 @@ JNIEXPORT jobject JNICALL Java_org_vowpalwabbit_spark_VowpalWabbitNative_getArgu
   jstring args = env->NewStringUTF(serializer.str().c_str());
 
   jclass clazz = env->FindClass("org/vowpalwabbit/spark/VowpalWabbitArguments");
-  jmethodID ctor = env->GetMethodID(clazz, "<init>", "(IILjava/lang/String;)V");
+  jmethodID ctor = env->GetMethodID(clazz, "<init>", "(IILjava/lang/String;DD)V");
```
jackgerrits: What does this DD mean?

eisber: int, int, string, double, double in the signature (2 doubles added)

eisber: it's the constructor signature.

jackgerrits: Got it, thanks

java/src/main/java/org/vowpalwabbit/spark/VowpalWabbitNative.java
```diff
@@ -96,21 +105,35 @@ public VowpalWabbitNative(String args, byte[] model) {
...
+     * Creates a new empty VW example associated with this this instance. This is
+     * used to mark the end of a multiline example.
+     * 
      * @return new {@code VowpalWabbitExample} object.
      */
     public VowpalWabbitExample createEmptyExample() {
         return new VowpalWabbitExample(this.nativePointer, true);
     }
 
+    private static final String[] performanceKeys = new String[] { "numberOfExamplesPerPass", "weightedExampleSum",
```
jackgerrits: Would it be better for this to be an object? numberOfExamplesPerPass and totalNumberOfFeatures should be integers and being forced into doubles seems odd.

eisber: reason I switched is that a) I'm forcing into double on spark side anyway b) I should start to deal with exceptions (e.g. when calling/creating java objects). 

Regardless, you have a fair point. I'll change it.

PR #2087, opened 2019-09-17
Move cb_sample to be class based

jackgerrits:


vowpalwabbit/cb_sample.cc
```diff
@@ -93,8 +102,7 @@ base_learner* cb_sample_setup(options_i& options, vw& all)
   if (!cb_sample_option)
     return nullptr;
 
-  data->seed_state = all.random_seed;
-
+  auto data = scoped_calloc_or_throw<cb_sample_data>(all.random_seed);
```
rajan-chari: Peter's PR https://github.com/VowpalWabbit/vowpal_wabbit/pull/2069 creates a shared random state object.  Perhaps we should be using that here?

vowpalwabbit/cb_sample.cc
```diff
@@ -8,82 +8,91 @@ using namespace LEARNER;
 using namespace VW;
 using namespace VW::config;
 
+namespace VW
+{
 struct cb_sample_data
```
rajan-chari: Some explanation of what this reduction does will be useful.

vowpalwabbit/cb_sample.cc
```diff
@@ -8,82 +8,91 @@ using namespace LEARNER;
...
+void learn_or_predict(cb_sample_data &data, multi_learner &base, multi_ex &examples)
+{
+  data.learn_or_predict<is_learn>(base, examples);
 }
 
-base_learner* cb_sample_setup(options_i& options, vw& all)
+base_learner *cb_sample_setup(options_i &options, vw &all)
 {
-  auto data = scoped_calloc_or_throw<cb_sample_data>();
   bool cb_sample_option = false;
```
rajan-chari: I don't see where this is set to true?  Does this setup function always return nullptr?

vowpalwabbit/cb_sample.h
```diff
@@ -1,3 +1,5 @@
 #pragma once
-
+
+#include "reductions_fwd.h"
```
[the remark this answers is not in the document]
jackgerrits: Yes, this header brings forward definitions used in the below declaration

vowpalwabbit/cb_sample.cc
jackgerrits: Yes, I'd love to consume that object here!

vowpalwabbit/cb_sample.cc
jackgerrits: This is set to true by the option parser on line 90

vowpalwabbit/cb_sample.cc
jackgerrits: Sounds good, I'll add a doc comment

vowpalwabbit/cb_sample.cc
```diff
@@ -8,82 +8,91 @@ using namespace LEARNER;
...
-  uint64_t seed_state;
-};
+  explicit cb_sample_data(uint64_t seedState) : _seed_state(seedState) {}
 
-template <bool is_learn>
-void learn_or_predict(cb_sample_data& data, multi_learner& base, multi_ex& examples)
-{
-  multiline_learn_or_predict<is_learn>(base, examples, examples[0]->ft_offset);
+  template <bool is_learn>
+  void learn_or_predict(multi_learner &base, multi_ex &examples)
```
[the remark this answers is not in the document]
jackgerrits: We don't make any other learn_or_predict functions inline, why this one?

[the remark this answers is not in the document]
rajan-chari: Good question.  By adding classes we now have one more method.  Perhaps there is more opportunity to inline now.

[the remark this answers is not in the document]
jackgerrits: I agree, the functions that are required to get learn/predict happening should be easily inlined

rajan-chari:
Merged thanks.

PR #2088, opened 2019-09-17
fix: softmax can overflow

pmineiro:
Encountered in practice #sadlife


explore/explore_internal.h
```diff
@@ -90,7 +90,8 @@ namespace exploration
       return E_EXPLORATION_BAD_RANGE;
 
     float norm = 0.;
-    float max_score = *std::max_element(scores_first, scores_last);
+    float max_score = lambda > 0 ? *std::max_element(scores_first, scores_last)
```
jackgerrits: Can you explain the reasoning behind the change? Why does it make sense to find the min element for max score?

pmineiro: The reasoning is: 
* Right now it doesn't work.  (See new test number 200; inspired by a problem reported by a customer).
* This routine was written assuming lambda was positive but it is actually negative in practice ("minimizing cost" instead of "maximizing reward").
* To caller-proof this routine I decided to make it handle any lambda properly.

rajan-chari: Sounds good.

jackgerrits:
@pmineiro it looks like file pred-sets/ref/cbe_adf_softmax_biglambda.predict is missing from this PR. Could you add this file in a new PR?

PR #2093, opened 2019-09-21
Make hashing constexpr in C++14 and unify rotl impl

jackgerrits:
- Add future_compat.h that allows us to optionally use future features through macros
- Make hashing constexpr in C++14
- (Most important) unify rotl implementation used on Linux + Windows

Let me know what thoughts are about this approach to using new std features. This was more of an experiment that worked out well. Unifying the functions used is really useful though.


explore/hash.h
```diff
@@ -29,59 +29,48 @@ license as described in the file LICENSE.
...
+  // Block read - if your platform needs to do endian-swapping or can only
+  // handle aligned reads, do the conversion here
+  constexpr static inline uint32_t getblock(const uint32_t * p, int i) noexcept
+  {
+    return p[i];
+  }
 }
 
-inline uint64_t uniform_hash(const void * key, size_t len, uint64_t seed)
+STD14_CONSTEXPR inline uint64_t uniform_hash(const void* key, size_t len, uint64_t seed)
```
peterychang: VW_STD14_CONSTEXPR to avoid naming collisions?

PR #2094, opened 2019-09-24
Migrate c arrays to std::array

jackgerrits:
Benefits:
- Value semantics
- Allows use of iterators
- Same performance as c arrays
- Object interface


peterychang:
LGTM

Issue #1978, opened 2019-07-25
TestJsonDictThreading fails sporadically during CI


lokitoth:
The .NET Bindings test, TestJsonDictThreading, fails sporadically (very rarely).

JohnLangford:
This is potentially fixed now, and I don't have a way to reproduce.  Please reopen if it reoccurs.

PR #2096, opened 2019-09-25
Add option to turn off sampling for CCB

jackgerrits:


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -652,7 +654,7 @@ base_learner* ccb_explore_adf_setup(options_i& options, vw& all)
     options.add_and_parse(new_options);
   }
 
-  if (!options.was_supplied("cb_sample"))
+  if (!options.was_supplied("cb_sample") && !do_not_sample)
```
yannstad: If `cb_sample` was supplied, then `do_not_sample` is ignored.
This behavior is not very user friendly. Maybe you could throw or log a warning in that case?

jackgerrits: Good point, when cb_sample is supplied explicitly it overrides this.

This option prevents CCB from adding the automatic cb_sample option. But if cb_sample is supplied in the model explicitly then it will also be a problem. Perhaps I should check this option in the cb_sample reduction itself?

yannstad: What is the scope of this option: global or CCB only?
If it is CCB only, i suggest to do it in the ccb_explore_adf reduction.
If it is global (CCB + CB), then it makes sense to do it in the cb_sample reduction.

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -635,12 +635,14 @@ base_learner* ccb_explore_adf_setup(options_i& options, vw& all)
...
   auto data = scoped_calloc_or_throw<ccb>();
   bool ccb_explore_adf_option = false;
+  bool do_not_sample = false;
   option_group_definition new_options(
       "EXPERIMENTAL: Conditional Contextual Bandit Exploration with Action Dependent Features");
   new_options.add(
       make_option("ccb_explore_adf", ccb_explore_adf_option)
           .keep()
           .help("EXPERIMENTAL: Do Conditional Contextual Bandit learning with multiline action dependent features."));
+  new_options.add(make_option("do_not_sample", do_not_sample).keep().help("Turn off sampling each CB call"));
```
yannstad: `do_not_sample` gives the ability to deactivate `cb_sample`.
For consistency, I would rename it into `cb_do_not_sample`.

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: Yes it depends on that. If it is CCB only, but the model contains the cb_sample flag, then do_not_sample will not do as intended. It either needs to remove the flag from the options or operate in the cb_sample reduction. I am leaning towards the latter as it is less complicated code with, and making this a global thing

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: That sounds good

jackgerrits:
@yannstad Does this new approach seems more reasonable to you?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -652,7 +654,7 @@ base_learner* ccb_explore_adf_setup(options_i& options, vw& all)
     options.add_and_parse(new_options);
   }
 
-  if (!options.was_supplied("cb_sample"))
+  if (!options.was_supplied("cb_sample") && !slate)
```
yannstad: If `cb`_sample was supplied, then `slate` is ignored. This behavior is confusing.
The 2 options are not compatible, so why not catching/throwing from here?

jackgerrits: I've updated with a throw

yannstad:
@jackgerrits  do you know why 'TestJsonDictThreading' failed? is it related to your change?

jackgerrits:
It is not related, it is a flaky test. See: #1978

PR #2097, opened 2019-09-26
Fix OSX builds when not using Anaconda.

peterychang:
Builds through Anaconda on OSX continues to have problems
- It looks like there are issues with system installed versions of boost and anaconda installed boost on OSX (possibly with any version of boost >= 1.70), likely due to the fact that boost changed its discovery method from cmake in 1.70.
- If Anaconda is NOT installed, this change fixes the build for OSX 10.14
- This change SHOULD be safe for all versions of boost. Tested in OSX (boost 1.71), Ubuntu (boost 1.67),  and Windows(boost 1.69). If this isn't safe, then there's not really a way to fix the build scripts since we can't tell which version of boost we have until we're actually in cmake

Also made a change that fixes a minor build issue on Windows. With VS2017 generators, we no longer need to set the VCTargetsPath variable when building

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,13 +21,22 @@ here = os.path.abspath(os.path.dirname(__file__))
 pkg_path = os.path.join(here, 'python')
 
 class Distribution(_distribution):
-    if system == 'Windows':
-        global_options = _distribution.global_options + [
-            ('vcpkg-root=', None, 'Path to vcpkg root. For Windows only.'),
-        ]
+    global_options = _distribution.global_options
 
+    global_options += [
+        ('enable-boost-cmake', None, 'Enable boost-cmake'),
+        ('cmake-options=', None, 'Additional semicolon-separated cmake setup options list'),
+    ]
+
+    if system == 'Windows':
+        global_options += [
+            ('vcpkg-root=', None, 'Path to vcpkg root. For Windows only'),
+        ]
+ 
     def __init__(self, attrs=None):
         self.vcpkg_root = None
+        self.enable_boost_cmake = None
+        self.cmake_options = None
         _distribution.__init__(self, attrs)
 
 class CMakeExtension(Extension):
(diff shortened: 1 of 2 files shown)


peterychang:
Fixes https://github.com/VowpalWabbit/vowpal_wabbit/issues/2089

peterychang:
Notably this does not fix the libstdc++ issue. I know what the cause of the problem is, however I'll need to find a way to reproduce the problem and verify any potential solution doesn't cause any other problems

setup.py
```diff
@@ -83,7 +83,8 @@ def build_cmake(self, ext):
             '-DPY_VERSION=' + '{v[0]}.{v[1]}'.format(v=version_info),
             '-DBUILD_PYTHON=On',
             '-DBUILD_TESTS=Off',
-            '-DWARNINGS=Off'
+            '-DWARNINGS=Off',
+            '-DBoost_NO_BOOST_CMAKE=ON'
```
jackgerrits: So this option is okay to be universally applied?

peterychang: It looks like it is. I tested against older versions of boost on linux and windows and no problems came up.

I could pipe a flag through setup,py to disable this, but that would require some plumbing. The pip command to use setup.py options is also pretty nasty, but it might be worth it?

jackgerrits: Is it the case that older versions just ignore the option? So it is either ignored or needed? In that case it is likely okay?

peterychang: This option has been in cmake since at least 3.0.2, ubuntu 18.04 is currently at 3.10.2.

It seems like the flag only affects boost versions built with cmake (the boost-cmake project). I'm not sure which package libraries use that method, so I can't test against them. So far its worked with OSX/homebrew, Ubuntu/anaconda, WIndows/vcpkg

peterychang: I'll add a flag in setup.py to let us disable this in case it causes problems for some people. I'll also see if I can extend it to add arbitrary cmake flags so we won't need to make a release for every bug we find

setup.py
```diff
@@ -21,13 +21,22 @@
 pkg_path = os.path.join(here, 'python')
 
 class Distribution(_distribution):
+    global_options = _distribution.global_options
+
+    global_options += [
+        ('enable-boost-cmake', None, 'Enable boost-cmake'),
```
jackgerrits: If the default adds the option, either the default value should be true or the name should be different

peterychang:
Added 2 new flags to setup.py. One to disable the original change I made (in case it causes problems) and another to pass arbitrary flags to cmake (semicolon separated)

Usage:
```python setup.py --enable-boost-cmake --cmake-options=--verbose;-DCMAKE_OPTION=ON```
from pip
```pip install --global-option="--enable-boost-cmake" --global-option="--cmake-options=--verbose;-DCMAKE_OPTION=ON" vowpalwabbit```

setup.py
peterychang: default is disabling boost-cmake.
```Boost_NO_BOOST_CMAKE=ON```

I wanted this to be a boolean flag and didn't want to check various 'false' flags

jackgerrits: Oh I see! Since the default is no

PR #2105, opened 2019-10-08
Fix an LGTM warning in recommend

jackgerrits:


jackgerrits:
This is a minor fix, with mostly formatting changes, and a rename to fix a warning.

PR #2111, opened 2019-10-11
Fix segfault in CCB - MTR must clean up predictions allocated for cost sensitive examples

### LINKED ISSUE #2106: CCB crashes when prediction is reused causing memory corruption
Repro: `ccb_test.txt --ccb_explore_adf  --epsilon 0.8 --onethread --ring_size 7`

[ccb_test.txt](https://github.com/VowpalWabbit/vowpal_wabbit/files/3713744/ccb_test.txt)

Crash occurs in loss calculation in output.

It seems the memory corruption may occur because something has a reference to decision_scores v_array buffer but is interpreting it as a different type.

Update:
- There is an extra prediction that is leaking out from an underneath reduction. This seems to be happening during MTR in cb_adf. Using DR does not repro the bug.

jackgerrits:
Fixes #2106


vowpalwabbit/cb_adf.cc
```diff
@@ -267,11 +269,13 @@ void cb_adf::learn_MTR(multi_learner& base, multi_ex& examples)
   examples[_gen_cs.mtr_example]->weight *= 1.f / clipped_p *
       ((float)_gen_cs.event_sum / (float)_gen_cs.action_sum);
 
+  std::swap(_gen_cs.mtr_ec_seq[0]->pred.a_s, _a_s_mtr_cs);
   // TODO!!! cb_labels are not getting properly restored (empty costs are dropped)
```
peterychang: Is this comment still relevant?

peterychang:
LGTM

vowpalwabbit/cb_adf.cc
jackgerrits: I think it is because this touches predictions and not labels

PR #2112, opened 2019-10-11
Remove redunant copy from CCB reduction

jackgerrits:


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -158,25 +158,23 @@ void attach_label_to_example(
 
 void save_action_scores(ccb& data, decision_scores_t& decision_scores)
 {
-  // save a copy
-  auto copy = data.action_score_pool.get_object();
-  copy_array(copy, data.shared->pred.a_s);
-  decision_scores.push_back(copy);
+  auto pred = data.shared->pred.a_s;
```
peterychang: This will do a copy if we switch from v_array to std::vector. Maybe use a ref?

PR #2117, opened 2019-10-15
Parser string view

peterychang:
Removed `substring` and `v_hashmap` objects
`substring` replaced with `boost::string_view` for Boost > 1.61.0. `boost::string_ref` for Boost 1.53~1.60
`v_hashmap` replaced with `std::unordered_map`

**Potential issues**
This PR will make the minimum supported boost to be 1.53.0. This should be ok since ubuntu 14's default boost is 1.54.

`boost::string_view` was introduced in 1.61. Since this is newer than many older OS's support (eg: ubuntu 14/16), the actual object being used will depend on the version of boost being used. For Boost versions 1.53~1.60, `boost::string_ref` will be used instead. The functionality and API is nearly identical, there are only a few missing convenience functions; notably the copy() function which allows you to safely copy substrings out of `string_view`. The workaround for this issue is to simply memcpy the data out.

I don't believe this should be a problem, but in the name of full transparency there is one difference in the behavior that _could_ be an issue. Prior to this change, the hashes of any object in the `v_hashmap` was totally consistent since we used `uniform_hash()`. This is no longer the case for all of our maps.
- Keys of type `string_view` use `uniform_hash()` with seed=0. The reason for this is because hashing support was not added to `boost::string_view` until boost 1.69.
- The map used in search (see below) uses a uses a `uniform_hash()` with the same seed as currently used
- All other key types use the standard hashing algorithm

**Rework around the `features` object**
I had to refactor the `features` object to use a destructor so they would get cleaned up correctly; this required extensive work around lifetime maintenance for the `features` objects.

Currently its in a really weird state where the object is malloc'd in some places and new'd in others.

Additionally, the `vw` object contains 2 member variables which hold pointers to the same set of `features`. Neither container object fully owns the `features` since they can be created into one or the other depending on the workflow. In this case, I had to use a `shared_ptr` to correctly maintain the lifetime of the objects

**POI in Search**
There is one point of interest in the search reduction. Originally we used a `v_hashmap<unsigned char*, scored_action>` where the key is a malloc'd byte array containing binary data that is packed in a very specific way. To maintain that behavior and have the automatic cleanup functionality, we now use a `std::unordered_map<std::unique_ptr<uint8_t[]>, scored_action>` with custom hasher and comparator

Finally, we're not quite able to convert the read_example interface to take in a `const char *` (nice to have since it would remove a lot of compiler warnings in the unit tests) because a function pointer we use underneath (that serves example parsing and json parsing) takes in a `char *` (because the json parser destructively parses the incoming `char*`)


peterychang:
Rather than using `boost::string_view` and `boost::string_ref` based on boost version, rajan pointed out that there are other stand-alone libraries that implement `string_view`. Abseil being one possibility.

vowpalwabbit/search.cc
```diff
@@ -32,6 +33,8 @@ using std::endl;
 
 namespace Search
 {
+typedef std::unique_ptr<uint8_t[]> byte_array;
```
jackgerrits: prefer `using`

vowpalwabbit/search.cc
```diff
@@ -714,7 +732,7 @@ void reset_search_structure(search_private& priv)
   {
     if (ar.repr != nullptr)
     {
-      ar.repr->delete_v();
+      //ar.repr->delete_v();
```
jackgerrits: Remove?

vowpalwabbit/v_hashmap.h
```diff
@@ -108,6 +108,13 @@ class v_hashmap
   {
```
jackgerrits: Can this file be removed?

JohnLangford:
Be careful here.  Those functions were introduced for computational reasons.

JohnLangford:
What's the status of this?  I see many failing tests?

setup.py
```diff
@@ -85,7 +85,8 @@ def build_cmake(self, ext):
         distutils.dir_util.mkpath(lib_output_dir)
 
         # example of cmake args
-        config = 'Debug' if self.debug else 'Release'
+        #config = 'Debug' if self.debug else 'Release'
+        config = 'RelWithDebInfo'
```
jackgerrits: revert before merge

peterychang:
I think a bad merge/push happened somewhere. I pushed to the remote to test my changes on mac, the last push wasn't intended to be a candidate for merging since there should be a few more bugs I need to work out

peterychang:
This PR is ready for final review

vowpalwabbit/feature_group.h
```diff
@@ -277,7 +277,48 @@ struct features
...
+   // a null state to prevent freeing of shallow copied v_arrays
+   features(features&& other) :
+   values(std::move(other.values)),
+   indicies(std::move(other.indicies)),
+   space_names(std::move(other.space_names)),
+   sum_feat_sq(other.sum_feat_sq)
+   {
+     // We need to null out all the v_arrays to prevent double freeing during moves
+     auto & v = other.values;
+     v._begin = v._end = v.end_array = nullptr;
```
jackgerrits: I'd prefer if we did not chain the assignments like this. We will get warnings for this too

jackgerrits: Replace with v_init to zero out?

vowpalwabbit/interact.cc
```diff
@@ -21,7 +21,7 @@ struct interact
 
   ~interact()
```
jackgerrits: Remove?

vowpalwabbit/marginal.cc
```diff
@@ -41,7 +41,7 @@ struct data
 
   ~data()
```
jackgerrits: remove?

vowpalwabbit/object_pool.h
```diff
@@ -114,13 +114,14 @@ struct no_lock_object_pool
...
   }
 
-  std::queue<T*> m_pool;
-  std::vector<std::pair<T*, T*>> m_chunk_bounds;
-  std::vector<std::unique_ptr<T[]>> m_chunks;
   TInitializer m_initializer;
   TCleanup m_cleanup;
-  size_t m_initial_chunk_size = 0;
-  size_t m_chunk_size = 8;
+  const size_t m_initial_chunk_size = 0;
```
jackgerrits: https://www.reddit.com/r/cpp/comments/8wbeom/coding_guideline_avoid_const_member_variables/

We should avoid const members

vowpalwabbit/parse_args.cc
```diff
@@ -206,8 +205,8 @@ void parse_dictionary_argument(vw& all, std::string str)
                                                          << ", opening failed");
   }
 
-  feature_dict* map = &calloc_or_throw<feature_dict>();
-  map->init(1023, nullptr, substring_equal);
+  std::shared_ptr<feature_dict> map = std::make_shared<feature_dict>();
```
jackgerrits:
```suggestion
  auto map = std::make_shared<feature_dict>();
```

vowpalwabbit/parse_args.cc
```diff
@@ -776,14 +769,14 @@ void parse_feature_tweaks(options_i& options, vw& all, std::vector<std::string>&
...
-      all.trace_message << endl;
+      all.trace_message << std::endl;
   }
 
   if (options.was_supplied("interactions"))
   {
     if (!all.quiet)
       all.trace_message << "creating features for following interactions: ";
-    for (std::vector<std::string>::iterator i = interactions.begin(); i != interactions.end(); ++i)
+    for (auto i = interactions.begin(); i != interactions.end(); ++i)
```
jackgerrits:
```suggestion
    for (auto& i = interactions.begin(); i != interactions.end(); ++i)
```

vowpalwabbit/parse_args.cc
```diff
@@ -1928,6 +1918,9 @@ void finish(vw& all, bool delete_all)
     if (all.final_prediction_sink[i] != 1)
       io_buf::close_file_or_socket(all.final_prediction_sink[i]);
   all.final_prediction_sink.delete_v();
+
+  // TODO: is the following code supposed to clear the loaded_dictionaries vector?
```
jackgerrits: TODO still relevant?

vowpalwabbit/parse_example.cc
```diff
@@ -51,36 +52,41 @@ template <bool audit>
...
-  parser* p;
-  example* ae;
-  std::array<uint64_t, NUM_NAMESPACES>* affix_features;
-  std::array<bool, NUM_NAMESPACES>* spelling_features;
-  v_array<char> spelling;
-  uint32_t hash_seed;
-  uint64_t parse_mask;
-
-  std::array<std::vector<feature_dict*>, NUM_NAMESPACES>* namespace_dictionaries;
+  const VW::string_view _line;
```
jackgerrits: const member here too

vowpalwabbit/parse_args.cc
peterychang: for this, I think an explicit definition is better for readability

vowpalwabbit/parse_args.cc
peterychang: changed to range-based loop

vowpalwabbit/cb_sample.cc
```diff
@@ -58,8 +59,8 @@ struct cb_sample_data
         if (strncmp(examples[0]->tag.begin(), SEED_IDENTIFIER.c_str(), SEED_IDENTIFIER.size()) == 0 &&
             examples[0]->tag.size() > SEED_IDENTIFIER.size())
         {
-          substring tag_seed{examples[0]->tag.begin() + 5, examples[0]->tag.begin() + examples[0]->tag.size()};
-          seed = uniform_hash(tag_seed.begin, substring_len(tag_seed), 0);
+          VW::string_view tag_seed(examples[0]->tag.begin() + 5, examples[0]->tag.size());
```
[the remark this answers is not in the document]
peterychang: No, string_view is just a `const char*` (to an existing string) and length. The consequence of this is that the underlying string needs to outlive the string_view since it owns no memory

vowpalwabbit/ccb_label.cc
```diff
@@ -274,40 +275,40 @@ CCB::conditional_contextual_bandit_outcome* parse_outcome(substring& outcome)
...
-void parse_label(parser* p, shared_data*, void* v, v_array<substring>& words)
+void parse_label(parser* p, shared_data*, void* v, v_array<VW::string_view>& words)
 {
   CCB::label* ld = static_cast<CCB::label*>(v);
   ld->weight = 1.0;
 
   if (words.size() < 2)
     THROW("ccb labels may not be empty");
-  if (!substring_equal(words[0], "ccb"))
+  if (!(words[0] == "ccb"))
```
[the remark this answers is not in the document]
peterychang: Current object-oriented C++ uses functions for the == operator, the performance should be exactly the same

vowpalwabbit/csoaa.cc
```diff
@@ -869,9 +868,8 @@ base_learner* csldf_setup(options_i& options, vw& all)
 
   all.p->emptylines_separate_examples = true;  // TODO: check this to be sure!!!  !ld->is_singleline;
 
-  features fs;
-  ld->label_features.init(256, fs, LabelDict::size_t_eq);
-  ld->label_features.get(1, 94717244);  // TODO: figure this out
+  ld->label_features.max_load_factor(0.25);
```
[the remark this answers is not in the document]
peterychang: These lines are just forcing unordered_map to mimic the v_hashmap behavior. v_hashmap in this case is allocating 256 elements of memory and has a max load factor of 25% before resizing

vowpalwabbit/global_data.h
```diff
@@ -33,97 +36,78 @@
...
-      std::cerr << '\'' << std::endl;
+      std::cerr << "warning: missing named label '" << s << '\'' << std::endl;
+      return 0;
     }
-    return v;
+    return iter->second;
   }
 
-  substring get(uint32_t v)
+  VW::string_view get(uint32_t v) const
```
jackgerrits: Can you make the two get functions use consistently either u32 or u64?

peterychang: Technically these should be size_t's since they're indices in a vector. But callers of these functions are using uint32_t's. Could possibly lead to issues someday, but I doubt we'll ever have more than ~4 Billion label names or run on an environment with a 16-bit size_t

vowpalwabbit/vw.h
```diff
@@ -135,41 +135,26 @@ void save_predictor(vw& all, io_buf& buf);
...
+  return getHasher(h)(s.data(), s.length(), u) & parse_mark;
 }
 
 inline uint64_t hash_feature_cstr(vw& all, char* fstr, uint64_t u)
 {
-  substring ss;
-  ss.begin = fstr;
-  ss.end = ss.begin + strlen(fstr);
-  return all.p->hasher(ss, u) & all.parse_mask;
+  return all.p->hasher(fstr, strlen(fstr), u) & all.parse_mask;
```
JohnLangford: I approve of all the simplification here :-)

vowpalwabbit/stagewise_poly.cc
```diff
@@ -75,7 +75,7 @@ struct stagewise_poly
     cout << "total feature number (after poly expansion!) = " << sum_sparsity << std::endl;
 #endif  // DEBUG
 
-    synth_ec.feature_space[tree_atomics].delete_v();
+    //synth_ec.feature_space[tree_atomics].delete_v();
```
JohnLangford: Memory leak?

JohnLangford: More generally, have you verified things are sound with valgrind?

vowpalwabbit/vw.h
peterychang: This was originally even simpler but slim couldn't handle string_views, so I had to convert it to a more generic interface

vowpalwabbit/queue.h
```diff
@@ -56,29 +55,33 @@ class ptr_queue
...
   }
 
   void set_done()
   {
-    done = true;
-
+    {
+      std::unique_lock<std::mutex> lock(mut);
+      done = true;
+    }
```
JohnLangford: Interesting.... Did this cause real bugs?

In essence, I'm unsure what counts as an atomic operation on a modern computer.

vowpalwabbit/stagewise_poly.cc
peterychang: `struct features` calls delete_v on its v_arrays in the destructor. So once `synth_ec` is destroyed, the features array goes out of scope and cleans itself

vowpalwabbit/parse_primitives.h
```diff
@@ -134,33 +115,50 @@ inline float parseFloat(char* p, char** end, char* endLine = nullptr)
...
+    {
+      *end_idx = end - start;
+    }
+    return ret;
+  }
+    
 }
 
-inline float float_of_substring(substring s)
+inline float parse_float_string_view(VW::string_view strview, size_t& end_idx)
```
JohnLangford: Is it worthwhile to have this intermediate function?

vowpalwabbit/parse_args.cc
```diff
@@ -288,11 +281,10 @@ void parse_dictionary_argument(vw& all, std::string str)
 
   if (!all.quiet)
     all.trace_message << "dictionary " << s << " contains " << map->size() << " item" << (map->size() == 1 ? "" : "s")
-                      << endl;
+                      << std::endl;
```
JohnLangford: Can we get rid of the many std::endl with a 'using' declaration?  It seems worthwhile in this file.

JohnLangford:
I finally finished reading through this!  

It's quite a patch---much more than just 'string_view'.  Almost everything seems quite worthwhile.  Can you look into my comments?  And have you done a valgrind test on the patch?

vowpalwabbit/stagewise_poly.cc
peterychang: > More generally, have you verified things are sound with valgrind?

Only insofar as to ensure there's no invalid memory usage or double freeing. I didn't check closely for leaks

vowpalwabbit/queue.h
peterychang: There weren't any bugs regarding this. These were issues I found while trying to figure out the race condition I was seeing (the one that was in neural net)

In general I don't think anything thats not explicitly called out as atomic can be relied on to be such. In this case, we might be able to get away with converting the done boolean to a `std::atomic<bool>`, but i'd have to do some more analysis on the consequences of that (especially on the read performance since this lock should only be hit once)

vowpalwabbit/parse_primitives.h
peterychang: we could remove it, the syntax is just a little bit more intuitive without having to call into begin and end

vowpalwabbit/parse_args.cc
peterychang: I can do that

peterychang:
Valgrind found a few broken places, mostly in Search. I'm looking at those now.

This change originally started as removing v_hashmap, which required removing substring, which snowballed into this monster. Hopefully I can keep the line changes down for future PRs.

vowpalwabbit/parse_primitives.cc
```diff
@@ -38,81 +21,52 @@ hash_func_t getHasher(const std::string& s)
...
-        // Regardless of whether the token was saved, we need to reset the current token.
-        current.begin = writing_head;
-        current.end = writing_head;
-      }
-      else if (current_character == '\\')
+      // always insert the next character after an escape if it exists
+      if(!s.empty())
       {
-        in_escape = !in_escape;
+	current.append(s.begin(), 1);
```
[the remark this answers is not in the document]
lokitoth: [Minor][Perf] Do we have a sense of what the performance differences between pre/post change? (It is not used in a hot path, so not necessarily an issue, but I am curious)

vowpalwabbit/parse_args.cc
```diff
@@ -204,8 +203,8 @@ void parse_dictionary_argument(vw& all, std::string str)
                                                          << ", opening failed");
   }
 
-  feature_dict* map = &calloc_or_throw<feature_dict>();
-  map->init(1023, nullptr, substring_equal);
+  std::shared_ptr<feature_dict> map = std::make_shared<feature_dict>();
+  map->max_load_factor(0.25);
```
lokitoth: 0.25 seems very sparse. Is this intended?

(We should consider adding a comment explaining the choice if we know the reasoning behind it)

vowpalwabbit/global_data.h
```diff
@@ -34,96 +37,77 @@
...
-inline void deleter(substring ss, uint64_t /* label */) { free_it(ss.begin); }
-
 class namedlabels
 {
  private:
-  std::vector<substring> id2name;
-  v_hashmap<substring, uint64_t> name2id;
-  uint32_t K;
+  // NOTE: This ordering is critical. m_id2name and m_name2id contain pointers into m_label_list!
+  std::string m_label_list;
```
lokitoth: [Minor][Nit] You add "m_" prefixes in this file, but in others you add "_" prefixes.

vowpalwabbit/global_data.h
lokitoth: That second one I could see happening. Could you add a static assert for sizeof(size_t) >= sizeof(uint32_t), or similar to call out the issue?

vowpalwabbit/v_array.h
```diff
@@ -112,8 +112,16 @@ struct v_array
...
   }
-  void push_back_unchecked(const T& new_ele)
-  { new (_end++) T(new_ele); }
+
+  void push_back_unchecked(const T& new_ele) { new (_end++) T(new_ele); }
+
+  template <class... Args>
+  void emplace_back(Args&&... args)
+  {
+    if (_end == end_array)
```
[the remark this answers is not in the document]
peterychang: no, one is a copy while the other is an in-place construction; I dont think we could switch one to use the other without losing performance. We could break out the capacity check/resize into another function if we wanted to though

vowpalwabbit/parse_primitives.cc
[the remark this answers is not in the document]
peterychang: We lose perf in the escaped tokenize function because we need to copy the strings instead of mutating it in place. The performance for the non-destructive tokenize function should be basically unchanged

vowpalwabbit/global_data.h
peterychang: We had a discussion late into the development of this branch to prefix with _ instead of m_. I'll leave this as-is for now, but we should make a sweep through the code to fix up the variable names at some point

vowpalwabbit/parse_primitives.h
```diff
@@ -85,14 +66,14 @@ hash_func_t getHasher(const std::string& s);
 //  - much faster (around 50% but depends on the  string to parse)
 //  - less error control, but utilised inside a very strict parser
 //    in charge of error detection.
-inline float parseFloat(char* p, char** end, char* endLine = nullptr)
+inline float parseFloat(const char* p, size_t* end_idx, const char* endLine = nullptr)
```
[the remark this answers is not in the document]
peterychang: rearranged the code a little bit and added a nullptr check.

vowpalwabbit/search.cc
```diff
@@ -138,6 +141,31 @@ std::ostream& operator<<(std::ostream& os, const action_cache& x)
...
+        return false;
+      return memcmp(A.get(), B.get(), sz_A) == 0;
+    }
+  };
+  struct cached_item_hash
+  {
+    size_t operator()(const byte_array& key) const
+    {
+      size_t sz = *key.get();
+      return uniform_hash(key.get(), sz, 3419);
```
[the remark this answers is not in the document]
peterychang: I'm not sure why this is 3419; this is just porting the existing behavior

vowpalwabbit/parse_primitives.h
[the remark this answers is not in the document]
peterychang: Actually, the checks incur a relatively large performance hit. Just switching it over to use a reference

vowpalwabbit/parse_primitives.h
JohnLangford: Personally, I'd remove it as a matter of style, just because being concise is good.   It's up to you.

vowpalwabbit/queue.h
JohnLangford: Where is the list of things explicitly called out as atomic?

vowpalwabbit/stagewise_poly.cc
JohnLangford: Definitely clean up all leaks before merging.

vowpalwabbit/queue.h
peterychang: std::atomic is a templated type. I believe all the primitive types have specializations while objects use a generic implementation.

According to https://en.cppreference.com/w/cpp/atomic/atomic there are specializations for pointers (including weak_ptr and shared_ptr), booleans, all integral types, and float types

vowpalwabbit/stagewise_poly.cc
peterychang: All the leaks found in RunTests have been cleaned up (other than the 5 test that we know will fail because of bash/python)

jackgerrits:
I think the only thing left here is John's sign off. @JohnLangford would you be able to give it a last pass?

JohnLangford:
It looks like there are some conflicts to resolve?  

And please fix the LGTM issue.

Other than that, please merge at will.

PR #2130, opened 2019-10-30
Fix slim build and various CMake fixes

jackgerrits:
- Create explore interface target
- Fix Slim cmake definition and master cmake file include
- Fix build issues


rajan-chari:
Looks good.  Merged.  Thanks

PR #2139, opened 2019-11-11
Python: Throw instead of silently append nullptr when types don't match

### LINKED ISSUE #2136: Passing a list of anything except example causes crash in Python
#### Describe the bug
When using the Python bindings, passing a list of anything by examples to predict will cause a crash. This is because of the following binding code, if it encounters anything except an example it replaces it with a nullptr. This should instead throw a Python exception saying the input is invalid.

https://github.com/VowpalWabbit/vowpal_wabbit/blob/19bf14e2a58eb6d73ffe09026da2463b12250e57/python/pylibvw.cc#L179
```C++
multi_ex unwrap_example_list(py::list& ec)
{
  multi_ex ex_coll;
  for (ssize_t i = 0; i<len(ec); i++)
  {
    py::object eci = ec[i];
    py::extract<example_ptr> get_ex(eci);
    example_ptr ecp;
    if (get_ex.check())
      ecp = get_ex();
    ex_coll.push_back(ecp.get());
  }
  return ex_coll;
}
```
1. The call to `.check()` returns false if the contained object is not convertible to `example_ptr`
2. Default constructed (nullptr) `example_ptr` is then pushed into the list
#### To Reproduce
This manifests especially easily when using `pyvw.vw.parse(...)`, since it returns a list of examples and is unintuitive when using it for vw format examples as you get a list of 1 example. Changing `append` to `extend` mitigates this, but discovering what is the problem is impossible when all you get is a segfault.
```Python
from vowpalwabbit import pyvw

model = pyvw.vw("--cb_explore_adf")

examples = []
examples.append(model.parse("shared | s_1", labelType=pyvw.pylibvw.vw.lContextualBandit))
examples.append(model.parse("0:1.5:0.25 | a:0.5 b:1", labelType=pyvw.pylibvw.vw.lContextualBandit))
examples.append(model.parse("| a:-1 b:-0.5", labelType=pyvw.pylibvw.vw.lContextualBandit))
examples.append(model.parse("| a:-2 b:-1", labelType=pyvw.pylibvw.vw.lContextualBandit))

pred = model.predict(examples, prediction_type=pyvw.pylibvw.vw.pACTION_SCORES)
```
#### Expected behavior
An exception thrown in Python saying the input is invalid.

#### Observed Behavior
Segmentation fault

#### Environment
What version of VW did you use? master, latest
What OS or language did you use? Python, Ubuntu 18.04

jackgerrits:
It turns out that `py::extract` will throw if the contained type is not convertible. We should rely on this instead of the current behavior of silently appending nullptrs.

Fixes #2136


peterychang:
LGTM

PR #2142, opened 2019-11-15
Use standard save/load functionality for sklearn Python lib

### LINKED ISSUE #2120: Incorrect predictions after loading model
Hi, thanks for this awesome library 👍

#### Describe the bug

I'm seeing incorrect predictions after loading a model with the Scikit-learn API.

#### To Reproduce

```python
from vowpalwabbit.sklearn_vw import VWRegressor

X = [[1, 2], [3, 4], [5, 6], [7, 8]]
y = [1, 2, 3, 4]

model = VWRegressor(l=100)
model.fit(X, y)
print(model.predict(X)) # [1. 2. 3. 4.]

model.save("/tmp/train.model")

model2 = VWRegressor(l=100)
model2.load("/tmp/train.model")
print(model2.predict(X)) # [1. 1. 1. 1.]
```

#### Expected behavior

The same predictions for both models: `[1. 2. 3. 4.]`

#### Observed Behavior

Different predictions for the loaded model: `[1. 1. 1. 1.]`

#### Environment

Master branch on MacOS 10.14.6 with the Python bindings

jackgerrits:
The Sklearn lib was saving and loading in a custom non standard way, I migrated to using the normal  model loading code. 

Fixes #2120

### DIFF
--- a/python/tests/test_sklearn_vw.py
+++ b/python/tests/test_sklearn_vw.py
@@ -246,7 +246,7 @@ def test_tovw():
     assert tovw(x=csr_matrix(x), y=y, sample_weight=w) == expected
 
 def test_save_load(tmp_path):
-    train_file = os.path.realpath(os.path.join(tmp_path, "train.model"))
+    train_file = str(tmp_path / "train.model")
 
     X = [[1, 2], [3, 4], [5, 6], [7, 8]]
     y = [1, 2, 3, 4]
--- a/python/tests/test_sklearn_vw.py
+++ b/python/tests/test_sklearn_vw.py
@@ -232,6 +232,24 @@ class TestVWRegressor:
         raw_model = VW()
         del raw_model
 
+    def test_save_load(self, tmp_path):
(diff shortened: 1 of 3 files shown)


python/tests/test_sklearn_vw.py
```diff
@@ -232,6 +232,24 @@ def test_delete(self):
...
+        model.fit(X, y)
+        before_saving = model.predict(X)
+
+        model.save(train_file)
+
+        model2 = VWRegressor(l=100)
+        model2.load(train_file)
+        after_loading = model2.predict(X)
+
+        assert all([a == b for a, b in zip(before_saving, after_loading)])
```
arielf: Simplify to?
   `assert before_loading == after_loading`

jackgerrits: I thought the same thing but when I do it fails with:
`ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`

PR #2157, opened 2019-11-22
Allow escaped command lines

### LINKED ISSUE #2124: Cannot supply filenames with spaces in them when using VW::initialize
Underlying issue is:

For the bindings, when initializing via `VW::initialize()`, we split the "command string" via whitespace-based tokenization, rather than the more complex CLI tokenization that happens when using the VW executable.

The fix is to add support for quoted tokens.

jackgerrits:
- Adds functionality for escaped strings when using the library initialize function
- Expose escaped command line through c interface
- Add `--quiet` to unit tests to fix output pollution
- Add deprecated macro

Fixes #2124

### DIFF
--- a/vowpalwabbit/parse_args.cc
+++ b/vowpalwabbit/parse_args.cc
@@ -1605,19 +1605,16 @@ char** to_argv(std::string const& s, int& argc)
   c[1] = ' ';
   strcpy(c + 2, s.c_str());
   substring ss = {c, c + s.length() + 2};
-  std::vector<substring> foo;
-  tokenize(' ', ss, foo);
-  escaped_tokenize(' ', ss, foo);
+  std::vector<substring> tokens = escaped_tokenize(' ', ss);
 
   char** argv = calloc_or_throw<char*>(tokens.size());
   for (size_t i = 0; i < tokens.size(); i++)
   {
-    *(foo[i].end) = '\0';
-    argv[i] = calloc_or_throw<char>(foo[i].end - foo[i].begin + 1);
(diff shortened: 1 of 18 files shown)


vowpalwabbit/parse_primitives.cc
```diff
@@ -43,6 +43,91 @@ hash_func_t getHasher(const std::string& s)
...
+      if(current_character == delim)
+      {
+        current.end = writing_head++;
+        *current.end = '\0';
+        if(current.begin != current.end || allow_empty)
+        {
+          tokens.push_back(current);
+        }
+
+        // Regardless of whether the token was saved, we nee to reset the current token.
```
lokitoth:
```suggestion
        // Regardless of whether the token was saved, we need to reset the current token.
```

vowpalwabbit/parse_args.cc
```diff
@@ -1598,7 +1598,29 @@ void cmd_string_replace_value(std::stringstream*& ss, std::string flag_to_replac
   }
 }
 
-char** get_argv_from_string(std::string s, int& argc)
+char** to_argv_escaped(std::string const& s, int& argc)
+{
+  char* c = calloc_or_throw<char>(s.length() + 3);
```
lokitoth: Since you will be reprocessing the escaped tokens anyways, do you need to start with copying the string with the prefix token? Could you just inject it right before the for loop below?

In other words:

```
char** argv = calloc_or_throw<char*>(tokens.size() + 1);
argv[0] = calloc_or_throw<char>(2); //or even point it to a constant string,
                                    //though that might run into cleanup issues
```

PR #2167, opened 2019-12-02
Fix memory leak in search.cc

jackgerrits:


vowpalwabbit/search.cc
```diff
@@ -301,6 +301,7 @@ search::~search()
     search_private& priv = *this->priv;
     clear_cache_hash_map(priv);
 
+    priv._random_state.~shared_ptr<rand_state>();
```
lokitoth: Since _random_state is a shared_pointer<>, should it not properly delete itself via RAII? Why do you need to call the destructor directly?

jackgerrits: The destructor of priv is never called. It is deallocated using `free`. This is on the todo list of "make sure destructors are used"

PR #2171, opened 2019-12-02
Unify throwing of exceptions to use vw_exception instead of bare std::exception

### LINKED ISSUE #2168: VW should not throw exceptions without a message
We have a number of places in code which raise a std::exception with no message.

We should standardize on VW::vw_exception, and make sure we always have a message in the exception.

Also we should opportunistically get rid of `throw 1` in array_parameters.h

jackgerrits:
Fixes #2168

### DIFF
--- a/vowpalwabbit/array_parameters.h
+++ b/vowpalwabbit/array_parameters.h
@@ -101,7 +101,7 @@ class sparse_parameters
 
   weight* first()
   {
-    THROW_OR_RETURN_VOID("Allreduce currently not supported in sparse");
+    THROW_OR_RETURN_NORMAL("Allreduce currently not supported in sparse", nullptr);
   }
 
   // iterator with stride
--- a/vowpalwabbit/array_parameters.h
+++ b/vowpalwabbit/array_parameters.h
@@ -101,7 +101,7 @@ class sparse_parameters
 
   weight* first()
   {
-    THROW("Allreduce currently not supported in sparse");
+    THROW_OR_RETURN_VOID("Allreduce currently not supported in sparse");
   }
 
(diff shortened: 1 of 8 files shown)


vowpalwabbit/array_parameters.h
```diff
@@ -98,7 +99,10 @@ class sparse_parameters
   sparse_parameters(const sparse_parameters& other) { shallow_copy(other); }
   sparse_parameters(sparse_parameters&&) = delete;
 
-  weight* first() { throw 1; }  // TODO: Throw better exceptions. Allreduce currently not supported in sparse.
+  weight* first()
+  {
+    THROW_OR_RETURN_VOID("Allreduce currently not supported in sparse");
```
lokitoth: This seems to be a weird macro name given that the function returns `weight*`. Is it the right one?

jackgerrits: Oh yes good catch, thanks. This is wrong

vowpalwabbit/array_parameters.h
```diff
@@ -101,7 +101,7 @@ class sparse_parameters
 
   weight* first()
   {
-    THROW_OR_RETURN_VOID("Allreduce currently not supported in sparse");
+    THROW_OR_RETURN_NORMAL("Allreduce currently not supported in sparse", nullptr);
```
lokitoth: Is there a THROW_OR_RETURN() that functions differently you are avoiding colliding with?

jackgerrits: Whoops, yes there is a macro that handles both these cases. Using that now.

PR #2172, opened 2019-12-02
Fix header install locations

### LINKED ISSUE #2165: Install does not place headers in correct location
Headers are installed to `/usr/local/lib`, but should be installed to `/usr/local/lib/vowpalwabbit/`

See install ouput:
```
...
-- Installing: /usr/local/include/accumulate.h
-- Installing: /usr/local/include/action_score.h
-- Installing: /usr/local/include/active_cover.h
-- Installing: /usr/local/include/active.h
-- Installing: /usr/local/include/allreduce.h
...
```

This needs to be fixed before we release.

jackgerrits:
- Fix pkgconfig files
- Install headers to vowpalwabbit subdir
- Install Slim headers

This change fixes a regression in 8.7.0. In the change to CMake install locations were accidentally modified, this change resolves that.

Fixes #2165

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,8 @@ if(WIN32)
 endif()
 
 project(vowpal_wabbit VERSION ${PACKAGE_VERSION} LANGUAGES CXX)
+set(VW_PROJECT_DESCRIPTION "Vowpal Wabbit Machine Learning System")
+set(VW_PROJECT_URL "https://vowpalwabbit.org")
 
 option(USE_LATEST_STD "Advanced: Override using C++11 with the latest standard the compiler offers. This can result in less than C++" OFF)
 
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -242,11 +244,11 @@ if(VW_INSTALL)
     SET (BOOST_LINK_LIBRARIES "${BOOST_LINK_LIBRARIES} ${BOOST_LIBRARY}")
   ENDFOREACH()
 
-  configure_file(libvw.pc.in libvw.pc)
-  configure_file(libvw_c_wrapper.pc.in libvw_c_wrapper.pc)
+  configure_file(libvw.pc.in libvw.pc @ONLY)
```
lokitoth: What does `@ONLY` do here?

vowpalwabbit/CMakeLists.txt
```diff
@@ -187,5 +187,5 @@ if(VW_INSTALL)
   # VW target headers
   install(
     FILES ${vw_all_headers} ${CMAKE_CURRENT_BINARY_DIR}/config.h
-    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/vowpalwabbit)
```
lokitoth: I am a little wary here, because that would mean that a version upgrade could break people's makefiles. Probably not a huge deal, particularly with pkgconfig / CMake (especially if we add a find macro module), but let's make sure we are doing this with deliberation.

jackgerrits:
Need to revert explore directory structure change

CMakeLists.txt
jackgerrits: @ONLY means CMake will only expand variables in the form @var@ and not ${var}. This is required becuase pkgconfig file use vars of the form ${var} and we don't want them expanded at configure time.

vowpalwabbit/CMakeLists.txt
jackgerrits: So in 8.6.1 this is actually the behavior that existed. This is a fix.

lokitoth: Ah, I got it. Could you add that to the commit description to explain we are backing out an change in 8.7?

CMakeLists.txt
lokitoth: Makes sense. Thanks for the explanation :)

PR #2173, opened 2019-12-03
Allow overriding --epsilon value from model file

### LINKED ISSUE #2101: Allow overriding the --epsilon value when there is already a value stored in the model
Plan:
1. Change epsilon to a list of values
2. Always use the last value in the list as the epsilon value
3. Check: When a command line and model are loaded which both specify the value, the command line value should take precedence
4. When serializing the value, only save the last value in the list

lokitoth:
* Enables overridable scalar options by specifying .allow_override().

Fixes #2101


jackgerrits:
When loading a model file do the model arguments or command line arguments get loaded last? This will only work if we allow command line to override and load those second

lokitoth:
@jackgerrits It works for the reductions, because they get parsed after the file options are merged, in `parse_modules`. We would need to use a different strategy for options declared before the file is parsed (and not redeclared later, like the update rule options)

The only options where this would not apply to are those defined in the root parse_args. We could move a chunk of them to after the file parse, but I am a bit worried about moving that much stuff right before release.

jackgerrits:
I mean in the notifier there is a list of options passed, the order of that list is critical. Do the command line options go after the model file?

lokitoth:
@jackgerrits Ah, I see. Right, I got it backwards because insert() always inserts it at the end. Good catch

lokitoth:
After offline discussion with @jackgerrits going to avoid pushing this before 8.8 to let it bake some more. Will continue updating to get it closer to shippable form in the meanwhile, but focusing on other issues for 8.8 first.

Issue #2175, opened 2019-12-04
Oracle generation is broken for LDF-mode in search_covington.ipynb


lokitoth:
Some time ago we changed the logic for oracle validation in Python to prevent zeroes from being passed in (see #1296 and #1042). This leads to broken oracle generation. For complex LDF cases, there seems to be additional issues once the oracle generation is fixed that prevents the system from learning correctly.

This issue is a tracking issue for the work to get search to a better place and update the examples to produce the expected output.

lokitoth:
Looks like at least part of the issue may be due to bad ternary operator as per #2374

jackgerrits:
@lokitoth Is there a way to reproduce this isssue?

lokitoth:
This gets reproduced in the search_covington.ipynb notebook: https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/docs/source/examples/search_covington.ipynb

In particular, see the end printout:

```
testing LDF
[1, 2, -1, 1, 2]
should have printed [ 1 2 -1 4 2 ]
```

There is strong reason to believe that the core of the search algorithm works properly - I spent time inside of the LDF-specific code-paths, and it did not appear that the issue was there. My current suspicion is that it is a problem either in the python binding layer, or the example is improperly put-together.

Updating title to make it clear what is going on.

JohnLangford:
Fixed in #4838

Issue #2144, opened 2019-11-17
RuntimeError: std::exception when running Learning_to_Search.ipynb


hsm207:
#### Describe the bug
I get a `RuntimeError: std::exception` when executing the cell:

```
for i in range(10):
    sequenceLabeler.learn(my_dataset)
```

In the [Learning_to_Search](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/examples/Learning_to_Search.ipynb) notebook.

#### To Reproduce
Run all the cells in the  [Learning_to_Search](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/examples/Learning_to_Search.ipynb) notebook

#### Expected behavior
No RuntimeException

#### Observed Behavior
This stack trace:

![image](https://user-images.githubusercontent.com/2398765/69001902-d0361f00-0921-11ea-8bb5-7f9b36bb7fe8.png)

#### Environment
What version of VW did you use? 
8.7.0.post1

What OS or language did you use?
Ubuntu 18.04.3 LTS with Python 2.7.15 | packaged by conda-forge | (default, Feb 28 2019, 04:00:11) [GCC 7.3.0]

jackgerrits:
@lokitoth do you think this is a duplicate of #1861?

lokitoth:
I will take a look, and dupe it against the other. Let me take it for now.

lokitoth:
The issue here is due to changes to how example disposal works in the Python bindings. We mistakenly kept __enter__ when removing __exit__ in #1837.

PR #2176. removes the __enter__ call and updates the example scripts to properly dispose the generated examples (without `with`, as that is no longer supported.) Unfortunately, it seems that there was no good way to implement finish_example without the breaking change.

lokitoth:
The samples have been updated with the correct code to use in this case with #2176. LDF issues with Covington DEP Parser, and Word Alignment are tracked by #2175.

Issue #1861, opened 2019-05-04
sequenceLabeler.learn producing run error


PramodParida:
Hi,

I am getting this error while running the below code:

#### Code:
```python
class SequenceLabeler(pyvw.SearchTask):
    def __init__(self, vw, sch, num_actions):
        # you must must must initialize the parent class
        # this will automatically store self.sch <- sch, self.vw <- vw
        pyvw.SearchTask.__init__(self, vw, sch, num_actions)
        
        # set whatever options you want
        sch.set_options( sch.AUTO_HAMMING_LOSS | sch.AUTO_CONDITION_FEATURES )

    def _run(self, sentence):   # it's called _run to remind you that you shouldn't call it directly!
        output = []
        for n in range(len(sentence)):
            pos,word = sentence[n]
            # use "with...as..." to guarantee that the example is finished properly
            with self.vw.example({'w': [word]}) as ex:
                pred = self.sch.predict(examples=ex, my_tag=n+1, oracle=pos, condition=[(n,'p'), (n-1, 'q')])
                output.append(pred)
        return output

vw = pyvw.vw("--search 4 --search_task hook --ring_size 1024") # 3 is the number of labels
sequenceLabeler = vw.init_search_task(SequenceLabeler)

for i in range(3):
    sequenceLabeler.learn(my_dataset)
```
#### Error:
```
  File "/home/pramod/anaconda3/lib/python3.6/site-packages/vowpalwabbit-8.6.1-py3.6-linux-x86_64.egg/vowpalwabbit/pyvw.py", line 24, in run
    def run(): self._output = self._run(my_example)

  File "<ipython-input-26-b725e17b2470>", line 15, in _run
    with self.vw.example({'w': [word]}) as ex:

AttributeError: __exit__

Traceback (most recent call last):

  File "<ipython-input-26-b725e17b2470>", line 26, in <module>
    sequenceLabeler.learn(my_dataset)

  File "/home/pramod/anaconda3/lib/python3.6/site-packages/vowpalwabbit-8.6.1-py3.6-linux-x86_64.egg/vowpalwabbit/pyvw.py", line 36, in learn
    self._call_vw(my_example, isTest=False);

  File "/home/pramod/anaconda3/lib/python3.6/site-packages/vowpalwabbit-8.6.1-py3.6-linux-x86_64.egg/vowpalwabbit/pyvw.py", line 31, in _call_vw
[...]

PR #2176, opened 2019-12-04
Fixes for Learning2Search Subsystem

lokitoth:
Fixes a set of issues in search bindings and corresponding examples. 

In PR #1837, we unified how examples in Python are disposed (to enable properly cleaning up multi example lists, and ensure we are doing the right type of dispose). This resulted in removing the ability to use the `with ... as` construction in Python, but we failed to update the example scripts.

This fixes the issues seen in #1861 and #2144 (commenting out the LDF examples which do not currently work).

As well, in PR #1296, we added a safety check in Python to ensure we do not pass oracles with values of 0. Similarly to above, we did not properly update the example scripts. As well, in LDF mode, we always expected the multi-example provided to have an example header, but this does not occur in the example cases. The fix is to check for the presence of a header and fixup the index to ensure that we address the right example.

Finally, in the case of CovingtonLDF / WordAlignment, fixing up the oracles exposes an additional issue. This additional work is tracked by #2175.

### DIFF
--- a/python/examples/Learning_to_Search.ipynb
+++ b/python/examples/Learning_to_Search.ipynb
@@ -114,9 +114,10 @@
     "        for n in range(len(sentence)):\n",
     "            pos,word = sentence[n]\n",
     "            # use \"with...as...\" to guarantee that the example is finished properly\n",
-    "            with self.vw.example({'w': [word]}) as ex:\n",
-    "                pred = self.sch.predict(examples=ex, my_tag=n+1, oracle=pos, condition=[(n,'p'), (n-1, 'q')])\n",
-    "                output.append(pred)\n",
+    "            ex = self.vw.example({'w': [word]})\n",
+    "            pred = self.sch.predict(examples=ex, my_tag=n+1, oracle=pos, condition=[(n,'p'), (n-1, 'q')])\n",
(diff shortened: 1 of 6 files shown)


python/examples/Learning_to_Search.ipynb
```diff
@@ -114,9 +114,10 @@
     "        for n in range(len(sentence)):\n",
     "            pos,word = sentence[n]\n",
     "            # use \"with...as...\" to guarantee that the example is finished properly\n",
-    "            with self.vw.example({'w': [word]}) as ex:\n",
-    "                pred = self.sch.predict(examples=ex, my_tag=n+1, oracle=pos, condition=[(n,'p'), (n-1, 'q')])\n",
-    "                output.append(pred)\n",
+    "            ex = self.vw.example({'w': [word]})\n",
+    "            pred = self.sch.predict(examples=ex, my_tag=n+1, oracle=pos, condition=[(n,'p'), (n-1, 'q')])\n",
+    "            self.vw.finish_example([ex]) # must pass the example in as a list because search is a MultiEx reduction\n",
```
jackgerrits: That's unfortunate... Especially since it is passed on its own above

PR #2181, opened 2019-12-07
New version

JohnLangford:


jackgerrits:
It's probably clang-format reordering headers causing it to break

JohnLangford:
Something's clearly quite wrong.

JohnLangford:
When running './utl/clang-format fix' locally, I see this repeated many times.

Error reading /home/jl/programs/vowpal_wabbit/.clang-format: Invalid argument
YAML:8:23: error: unknown key 'AlignEscapedNewlines'
AlignEscapedNewlines: Left
                      ^~~~

I was assuming it would not break the source, but that appears incorrect.

I'm planning to revert.

jackgerrits:
It is possible you need a newer version of clang format? Our build images use clang tools 7.0

JohnLangford:
I see: clang-format version 3.8.0-2ubuntu4 (tags/RELEASE_380/final)

jackgerrits:
Yes that is probably the issue. Some of the clang-format options were added later, which explains the error message you see

JohnLangford:
I'm still a bit shocked that clang-format will actively break your source if it doesn't recognize an option.

jackgerrits:
I think it probably defaults to true for the "SortIncludes" configuration option. Which can very easily break source

JohnLangford:
I can see how that would do it---still it's rather surprising to me.  

I'm looking into upgrading clang-format, but accidentally stepped into upgrade dependency bonus fun.  If it takes to long, I'll punt and just use this.

JohnLangford:
I installed clang-format-6.0.1, but it's not working properly.

JohnLangford:
Succeeded with clang-format 10.

PR #2195, opened 2019-12-14
Cleanup Headers

jackgerrits:
- Remove unused file primitives.cc
- Consistent header guards
- Move from header.h to cheader where possible
- Remove a redundant usage of boost::program_options


vowpalwabbit/gd.cc
```diff
@@ -3,12 +3,9 @@
 // license as described in the file LICENSE.
 #include "crossplat_compat.h"
 
-#include <float.h>
+#include <cfloat>
 #ifdef _WIN32
```
peterychang: Is this block needed at all anymore?

vowpalwabbit/global_data.cc
```diff
@@ -2,18 +2,24 @@
...
-#include <math.h>
-#include <assert.h>
+#include <cmath>
+#include <cassert>
 
 #include "global_data.h"
 #include "gd.h"
 #include "vw_exception.h"
 
+#ifdef _WIN32
```
peterychang: does there have to be an else to this?

jackgerrits: It seems the socket symbols are coming in from somewhere else on linux, I'll include them for safety though

PR #2203, opened 2019-12-18
Distributionally robust cb

pmineiro:
- Component that takes a stream of (w, r) pairs, maintains geometrically decaying statistics, and computes the chi^2-divergence dual variables on demand.
- Reduction that leverages the above component to reweigh training data as it streams.


test/unit_test/DistributionallyRobustUnitTest.ipynb
```diff
@@ -0,0 +1,288 @@
+{
```
jackgerrits: Can this just be a python script instead of a notebook?

Also can this be run with the other tests?

vowpalwabbit/cb_dro.cc
```diff
@@ -0,0 +1,154 @@
...
+    if (is_learn)
+    {
+      auto it = std::find_if(examples.begin(), examples.end(), [](example *item) { return !item->l.cb.costs.empty(); });
+
+      if (it != examples.end())
+      {
+        CB::cb_class logged = (*it)->l.cb.costs[0];
+        uint32_t labelled_action = std::distance(examples.begin(), it);
+
+        auto action_scores = examples[0]->pred.a_s;
```
jackgerrits: Const?

vowpalwabbit/cb_dro.cc
```diff
@@ -0,0 +1,154 @@
...
+    multiline_learn_or_predict<false>(base, examples, examples[0]->ft_offset);
+
+    if (is_learn)
+    {
+      auto it = std::find_if(examples.begin(), examples.end(), [](example *item) { return !item->l.cb.costs.empty(); });
+
+      if (it != examples.end())
+      {
+        CB::cb_class logged = (*it)->l.cb.costs[0];
+        uint32_t labelled_action = std::distance(examples.begin(), it);
```
jackgerrits: Const?

vowpalwabbit/cb_dro.cc
```diff
@@ -0,0 +1,154 @@
...
+    // "off policy learning of the stochastic exploration policy".
+    //
+    // Ergo, the following always optimizes the bound on the associated
+    // deterministic argmax score policy.
+
+    multiline_learn_or_predict<false>(base, examples, examples[0]->ft_offset);
+
+    if (is_learn)
+    {
+      auto it = std::find_if(examples.begin(), examples.end(), [](example *item) { return !item->l.cb.costs.empty(); });
```
jackgerrits: const?

vowpalwabbit/cb_dro.cc
```diff
@@ -0,0 +1,154 @@
...
+
+        chisq.update(chosen_action == labelled_action ? w : 0, r);
+
+        float qlb = chisq.effn() * chisq.qlb(w, r);
+
+        // avoid pathological cases
+        qlb = std::max(qlb, 0.01f);
+
+        // save the original weights and scale the example weights
+        std::vector<float> save_weight;
```
jackgerrits: You can reserve so that there aren't several allocation as it grows

vowpalwabbit/cb_dro.cc
```diff
@@ -0,0 +1,154 @@
...
+        std::transform(examples.begin(), examples.end(), std::back_inserter(save_weight), [](example *item) { return item->weight; });
```
jackgerrits:
```suggestion
        std::transform(examples.begin(), examples.end(), std::back_inserter(save_weight), [](example* item) { return item->weight; });
```

vowpalwabbit/cb_dro.cc
```diff
@@ -0,0 +1,154 @@
...
+        std::for_each(examples.begin(), examples.end(), [qlb](example *item) { item->weight *= qlb; });
```
jackgerrits:
```suggestion
        std::for_each(examples.begin(), examples.end(), [qlb](example* item) { item->weight *= qlb; });
```

vowpalwabbit/cb_dro.cc
```diff
@@ -0,0 +1,154 @@
...
+
+        // TODO: make sure descendants "do the right thing" with example->weight
+        multiline_learn_or_predict<true>(base, examples, examples[0]->ft_offset);
+
+        // restore the original weights
+        auto save_weight_it = save_weight.begin();
+        std::for_each(examples.begin(), examples.end(), [&save_weight_it](example *item) { item->weight = *save_weight_it++; });
```
jackgerrits:
```suggestion
        std::for_each(examples.begin(), examples.end(), [&save_weight_it](example* item) { item->weight = *save_weight_it++; });
```

vowpalwabbit/distributionally_robust.cc
```diff
@@ -0,0 +1,176 @@
...
+#include <type_traits>
+
+#include "distributionally_robust.h"
+
+namespace VW
+{
+namespace distributionally_robust
+{
+
+double ChiSquared::chisq_onedof_isf(double alpha)
```
jackgerrits: Can you mark this function as VW_STD14_CONSTEXPR?

vowpalwabbit/distributionally_robust.cc
```diff
@@ -0,0 +1,176 @@
...
+{
+  // data = Table[{ alpha, InverseCDF[ChiSquareDistribution[1], 1 - alpha] }, { alpha, 0.001, 0.999, 0.0005 }]
+  // lm = LinearModelFit[data, { Log[alpha], Log[alpha]^2, Log[alpha]^3, Log[alpha]^4 , Log[alpha]^5, Log[alpha]^6, Log[alpha]^7, Log[alpha]^8, Exp[alpha], Exp[alpha]^2, Exp[alpha]^3, Exp[alpha]^4, Exp[alpha]^5, Exp[alpha]^6, Exp[alpha]^7, Exp[alpha]^8}, alpha]
+  // ListPlot[lm["FitResiduals"]]
+  // lm["BestFitParameters"]
+  // Show[Plot[InverseCDF[ChiSquareDistribution[1], 1 - alpha], { alpha, 0 , 1}],  Plot[lm[alpha], { alpha, 0, 1 }], Frame->True]
+
+  double constants[] = {-1.73754,-1.40684,0.0758363,0.00726577,0.000468688,0.0000214395,1.0643e-6,6.43011e-8,2.0475e-9,1.16356,-0.575446,0.329796,-0.136076,0.0396764,-0.00763232,0.00087113,-0.0000445128};
```
jackgerrits:
```suggestion
  constexpr double constants[] = {-1.73754,-1.40684,0.0758363,0.00726577,0.000468688,0.0000214395,1.0643e-6,6.43011e-8,2.0475e-9,1.16356,-0.575446,0.329796,-0.136076,0.0396764,-0.00763232,0.00087113,-0.0000445128};
```

vowpalwabbit/distributionally_robust.cc
```diff
@@ -0,0 +1,176 @@
...
+      rv += constants[i] * logalphapow;
+      rv += constants[i + 8] * expalphapow;
+      logalphapow *= logalpha;
+      expalphapow *= expalpha;
+    }
+
+  return rv;
+}
+
+bool isclose(double x, double y, double atol=1e-8)
```
jackgerrits: VW_STD14_CONSTEXPR

vowpalwabbit/distributionally_robust.h
```diff
@@ -0,0 +1,136 @@
...
+#include <algorithm>
+#include <limits>
+#include <stdexcept>
+
+namespace VW
+{
+namespace distributionally_robust
+{
+
+class ChiSquared
```
jackgerrits: This entire class could likely be marked constexpr, but probably not worth the effort at the moment :)

vowpalwabbit/distributionally_robust.cc
```diff
@@ -0,0 +1,176 @@
...
+          double z = phi + 1 / (2 * n);
+
+          if (isclose(y * z, 0, 1e-9))
+            {
+              y = 0;
+            }
+
+          if (z <= 0 && y * z >= 0)
+            {
+              double kappa = std::sqrt(y / (2 * z));
```
ataymano: according to condition above, z can be 0?

vowpalwabbit/distributionally_robust.cc
pmineiro: constexpr function 'VW::distributionally_robust::ChiSquared::chisq_onedof_isf' cannot result in a constant expression

vowpalwabbit/distributionally_robust.cc
pmineiro: constexpr function 'VW::distributionally_robust::isclose' cannot result in a constant expression

vowpalwabbit/cb_dro.cc
pmineiro: Done.

vowpalwabbit/cb_dro.cc
pmineiro: Done.

vowpalwabbit/cb_dro.cc
pmineiro: Done.

vowpalwabbit/cb_dro.cc
pmineiro: Done.

vowpalwabbit/cb_dro.cc
pmineiro: Done.

vowpalwabbit/distributionally_robust.cc
pmineiro: Done.

vowpalwabbit/distributionally_robust.h
pmineiro: Not done.

vowpalwabbit/cb_dro.cc
pmineiro: Done.

vowpalwabbit/cb_dro.cc
pmineiro: Done.

vowpalwabbit/distributionally_robust.cc
jackgerrits: Ah std::abs is not i think, oh well

vowpalwabbit/distributionally_robust.cc
jackgerrits: Maybe because its a member function? Not important though, leave it out

test/unit_test/DistributionallyRobustUnitTest.ipynb
pmineiro: converted to script.  it just generates the test cases.  added comments to this effect.

vowpalwabbit/cb_adf.cc
```diff
@@ -203,9 +203,9 @@ void cb_adf::learn_SM(multi_learner& base, multi_ex& examples)
     _backup_nf.push_back((uint32_t)examples[current_action]->num_features);
 
     if (current_action == chosen_action)
-      examples[current_action]->weight = example_weight * (1.0f - action_score.score);
+      examples[current_action]->weight *= example_weight * (1.0f - action_score.score);
```
JohnLangford: This is unrelated.   What's the bug exactly?

vowpalwabbit/distributionally_robust.h
```diff
@@ -0,0 +1,136 @@
...
+        sumwsq(0),
+        sumwr(0),
+        sumwsqr(0),
+        sumwsqrsq(0),
+        delta(chisq_onedof_isf(alpha)),
+        duals_stale(true)
+      {
+        if (alpha > 1 || alpha <= 0)
+          {
+            throw std::invalid_argument("invalid alpha value");
```
JohnLangford: @rajan-chari what's your suggestion for avoiding throws here?

vowpalwabbit/distributionally_robust.cc
```diff
@@ -0,0 +1,176 @@
...
+
+double ChiSquared::chisq_onedof_isf(double alpha)
+{
+  // data = Table[{ alpha, InverseCDF[ChiSquareDistribution[1], 1 - alpha] }, { alpha, 0.001, 0.999, 0.0005 }]
+  // lm = LinearModelFit[data, { Log[alpha], Log[alpha]^2, Log[alpha]^3, Log[alpha]^4 , Log[alpha]^5, Log[alpha]^6, Log[alpha]^7, Log[alpha]^8, Exp[alpha], Exp[alpha]^2, Exp[alpha]^3, Exp[alpha]^4, Exp[alpha]^5, Exp[alpha]^6, Exp[alpha]^7, Exp[alpha]^8}, alpha]
+  // ListPlot[lm["FitResiduals"]]
+  // lm["BestFitParameters"]
+  // Show[Plot[InverseCDF[ChiSquareDistribution[1], 1 - alpha], { alpha, 0 , 1}],  Plot[lm[alpha], { alpha, 0, 1 }], Frame->True]
+
+  constexpr double constants[] = {-1.73754,-1.40684,0.0758363,0.00726577,0.000468688,0.0000214395,1.0643e-6,6.43011e-8,2.0475e-9,1.16356,-0.575446,0.329796,-0.136076,0.0396764,-0.00763232,0.00087113,-0.0000445128};
```
JohnLangford: What is this exactly?  Maybe add a comment about where these numbers come from.

vowpalwabbit/cb_dro.cc
```diff
@@ -0,0 +1,155 @@
...
+
+      if (it != examples.end())
+      {
+        const CB::cb_class logged = (*it)->l.cb.costs[0];
+        const uint32_t labelled_action = std::distance(examples.begin(), it);
+
+        const auto action_scores = examples[0]->pred.a_s;
+
+        // cb_explore_adf => want maximum probability
+        // cb_adf => want minimum cost
```
JohnLangford: minimum cost is not necessarily defined.   It works when we have a CSOAA reduction, but there are others for which it does not.   Instead, can we just pick the preferred action of the cb_adf policy?

vowpalwabbit/distributionally_robust.h
jackgerrits: Rather than std::invalid_argument, we should be using the THROW macro here

vowpalwabbit/cb_dro.cc
pmineiro: now using action_scores.begin()

JohnLangford:
@rajan-chari have you had a chance to think about exception removal here?

rajan-chari:
> @rajan-chari have you had a chance to think about exception removal here? 

Yes.  Apologies.  Just got to it.  Since you cannot return error codes from constructors, the only option is two stage construction with init() after construction where error codes are returned.  

For examples in rlclientlib:

https://github.com/VowpalWabbit/reinforcement_learning/blob/545950a3caffdddd9268e1eeb9f33a1c27510d4f/examples/rl_sim_cpp/rl_sim.cc#L184

https://github.com/VowpalWabbit/reinforcement_learning/blob/545950a3caffdddd9268e1eeb9f33a1c27510d4f/rlclientlib/live_model.cc#L42

jackgerrits:
But VW is not architected to support error codes right now, so I don’t think two stage construction buys us anything? I prefer using factory methods for a lib with error codes as it means we cannot be left with an uninitialized object accidentally.

Until we get in a state where error codes are viable either throwing or asserting seems like the only option here.

rajan-chari:
Having made the decision to move away from throwing, we should really start returning errors. Throwing in constructors seems especially problematic since its harder to unwind if the practice becomes widely used.  

Adding generic error handling to VW as a gating condition for these kinds changes will only make it harder to get there.

In this case a simple method that creates an instance and then calls init seems like a simple place to start the process.

jackgerrits:
The throws are happening on the setup path and not the predict or learn path.There are already many potential throw sites in the setup path.

Throwing exceptions out of constructors is standard practice when exceptions are being used. I personally view two stage initialization as an anti pattern as it introduces two states for every object (pre and post initialization) and opportunities for associated bugs. Constructors were designed to deal with this problem in the language, which is why I’m opposed to error codes as they prevent this valid state guarantee through construction. 

Given we’ve decided to use error codes perhaps there’s no point discussing this. If we do use factories and private constructors though we can guarantee received objects are in post initialization state.

pmineiro:
>> The throws are happening on the setup path and not the predict or learn path.There are already many potential throw sites in the setup path.

Based upon this, I removed the throw-s from the constructor for ChiSquared defined in distributionally_robust.h in case it wants to be used in another path someday.  However I retained the throws in cb_dro_setup().

rajan-chari:
Makes sense.  Thanks.

vowpalwabbit/cb_dro.cc
rajan-chari: If save_weight is made a member variable of cb_dro_data, and clear()ed here.  Allocations in steady state should go to zero in predict_learn().

PR #2206, opened 2019-12-20
Fix delete behavior for examples created using parse function

### LINKED ISSUE #2193: del and finish are behaving differently in Python (del causes crash)
Running the following in a cell in a Jupyter notebook twice fails:
```Python
from vowpalwabbit import pyvw

vw = pyvw.vw('--cb_explore_adf --epsilon 0.2 --ring_size 4000')
with open('data.txt') as f:
    examples = f.read().strip().split('\n\n')

for example in examples:
    pred_example = vw.parse(example, pyvw.vw.lContextualBandit)
    pmf = vw.predict(pred_example)
    vw.finish_example(pred_example)

    learn_example = vw.parse(example, pyvw.vw.lContextualBandit)
    vw.learn(learn_example)
    vw.finish_example(learn_example)

vw.finish()
```

Whereas the following works:
```Python
from vowpalwabbit import pyvw

vw = pyvw.vw('--cb_explore_adf --epsilon 0.2 --ring_size 4000')
with open('data.txt') as f:
    examples = f.read().strip().split('\n\n')

for example in examples:
    pred_example = vw.parse(example, pyvw.vw.lContextualBandit)
    pmf = vw.predict(pred_example)
    vw.finish_example(pred_example)

    learn_example = vw.parse(example, pyvw.vw.lContextualBandit)
    vw.learn(learn_example)
    vw.finish_example(learn_example)

del vw
```
The only difference is the last line.

### LINKED ISSUE #2199: Crash in cb_explore_adf in Python
This issue was originally reported as being dependent on ring size. Working at 400000, but failing at 4000. However, I am seeing it fail at the default 256.

data.txt: https://gist.github.com/AlexFridman/7d48306ef6ac04c9c520ae9782076da4

This crashes at about the 56th cb adf example

```Python
from vowpalwabbit import pyvw
vw = pyvw.vw('--cb_explore_adf --epsilon 0.2')
with open('data.txt') as f:
    examples = f.read().strip().split('\n\n')
for example in examples:
    pred_example = vw.parse(example, pyvw.vw.lContextualBandit)
    pmf = vw.predict(pred_example)
    vw.finish_example(pred_example)
    learn_example = vw.parse(example, pyvw.vw.lContextualBandit)
    vw.learn(learn_example)
    vw.finish_example(learn_example)
del vw
```

### LINKED ISSUE #2202: VW::finish_example getting called twice in Python bindings
This causes an assert to hit in `clean_example` (`assert(ec.in_use);`) as it has already been returned. Therefore I think multiple pointers are being put back into the pool.

To repro, run the following with a debug build of the Python bindings. When using single examples this is unlikely to cause problems. But as soon as you use multi_ex this is a very dangerous.

```Python
from vowpalwabbit import pyvw
vw = pyvw.vw("--quiet")
vw.learn("1 |x a b")
```

jackgerrits:
There were several Python issues and it seem that the root cause is the deleter for the objects. I incorrectly put in a fix for a memory leak some time ago (PR #2153), but the leak itself was fixed elsewhere in 8.8.0 and the deleter I used was wrong. It caused double finish behavior which broke other things.

Fixes #2193 
Fixes #2199 
Fixes #2202

### DIFF
--- a/python/pylibvw.cc
+++ b/python/pylibvw.cc
@@ -247,8 +247,10 @@ py::list my_parse(vw_ptr& all, char* str)
   for (auto ex : examples)
   {
     VW::setup_example(*all, ex);
+    // Examples created from parsed text should not be deleted normally. Instead they need to be
+    // returned to the pool using finish_example.
     example_collection.append(
-        boost::shared_ptr<example>(ex, [all](example* example) { VW::finish_example(*all.get(), *example); }));
+        boost::shared_ptr<example>(ex, dont_delete_me));
   }
   examples.clear();
   examples.delete_v();


peterychang:
lgtm

PR #2207, opened 2019-12-20
Fix learn semantics

rajan-chari:
Still working on reducing perf gap due to this change.

**Summary of changes:**
1) Entry point to learn (global_data.cc) calls predict and then learn.  All the other changes stem from this change.
2) VW_DBG to output debug logs in hierarchical form to debug different reductions
3) Parser for simple label had to change.
4) Changed some names because ... too manys 'l's and 'p's.  p->example_parser l->lbl_parser


vowpalwabbit/cb_explore_adf_cover.cc
```diff
@@ -160,6 +177,17 @@ void cb_explore_adf_cover::predict_or_learn_impl(LEARNER::multi_learner& base, m
   sort_action_probs(_action_probs, _scores);
   for (size_t i = 0; i < num_actions; i++) preds[i] = _action_probs[i];
 
+  if (VW_DEBUG_LOG)
```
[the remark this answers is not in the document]
rajan-chari: Normal compiler optimization should do the same thing without contexpr.  Are there other reasons besides that?  For example see https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAM1QDsCBlZAQwBtMQBGAFlICsupVs1qhkAUgBMAISnTSAZ0ztkBPHUqZa6AMKpWAVwC2tQVvQAZPLUwA5YwCNMxEJIDspAA6oFhdbT1DE0FvXzU6Kxt7IycXd0VlTFV/BgJmYgJA41NOBJVw2lT0gki7R2dXDwU0jKzg3Ori0ujYyoBKRVQDYmQOAHIpAGZrZEMsAGpxQZ11auJMZiMp7HEABgBBNfW5g1VxgDcAdwB9dAdgLfE3WQ3x8bRaavHGtWRxh1R9cawqZgNWAjHVioYB3MFTAAi41%2BrCUUxu6zuDyeLzwbw%2BX3Y6RsxDBeLukOhbDhgwRSLoKLSr3en1YzwWPQQ%2BPxhJhJLJ9wpBGeVLRNK%2BwHQzOFk0GULZmHhW3Jj25qPRtPGguOnnm%2BD2LLFRNhktJ0s5sp5zGpGLpDms6QAniKCVqJVLbgbKca%2Bab7g5jsx0FQbaLxcTdRzkXLeQqvsgFKhmMxfX7tez9cGjSbFRHPcl9phhayAw7No6k/L%2BXSCMRmMgANbHRqVuM/P4AoEgyZuHTjLHEHEtttKdLIJlXNuC7tK9AqtVo7mD94W4jW6fID1en0LyPRkdp8tqTN5q4Q3cbIY/axZgBqAHVjhDsNIAKoAcWOFgA8veDiczqAQPX/oDgRdDw2aw5QARwMdJMAgNoWw5O48CoCAjlOc4QBADscTafVmXmAhulocZJE4eFJkdOCEKQz9UNLcsqxrCtMNI/EcLwgjJGIrDxngiALyvG8HyfV8GMRYVmM7AjBnYwDhLuUT8O4CS9UPNwIT6DpWBAPoAFY%2BlIUw%2BlWHTUA0nQ5DkZ4uh6LMhk4HSCA0gy2g6CsQE01YhA07gdL0gzSCMvodIUEA3Ls/TVNIOBYCQNAjE8PB2DICgIGi2L4pAYAAE5JFIKg4oIZxAogBx7J081aCtDSbNIaKjC0Ahn1oVhLWK0gsCMERgHYZr8HmDNMEC0LSEwAAPJIDDyiqdOA5RmtYPAHDLOc9CwCbSFLPAjAmjoaHoJg2A4Hh%2BEEYRRBQUyZCEObAsgDpUE8Ap%2BoAWmfSQAsSZINAgcw6hyQbtGacoXFyUI/Dob6Qh8EHaH%2BmIKgaN6CiKWp9GyQQlHyFIahKE8WlhxRMbBhpMeh1pOA6SNul6Lg1I07TdOavyhoADgANge5nuCVZA3nSgA6SRxggXBCBISZJEGXJxj0GK4ucUXxegkyZDkWzisc0hnNc9y%2Bk8umBr8gKgtW1Xqb6F7dZ8/WjdCtXM2IXwNG4IA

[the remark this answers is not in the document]
jackgerrits: They are logically compile time available constants and so should be constexpr. The compiler can deal with them and optimize it which is good but if we wanted to use them in some larger constexpr expression it would not be possible without constexpr.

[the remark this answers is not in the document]
rajan-chari: Done.  FYI...  Under msvc 2019 this still does not get compiled out unless -O is specified.  gcc seems to do the right thing.

vowpalwabbit/cb_explore_adf_greedy.cc
```diff
@@ -67,6 +65,19 @@ void cb_explore_adf_greedy::predict_or_learn_impl(LEARNER::multi_learner& base,
     preds[0].score += 1.f - _epsilon;
 }
 
+
+template <bool is_learn>
+void cb_explore_adf_greedy::predict_or_learn_impl(LEARNER::multi_learner& base, multi_ex& examples)
+{
+  // Explore uniform random an epsilon fraction of the time.
+  if(is_learn)
```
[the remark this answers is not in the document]
rajan-chari: Tried this out.  However doing the following: 

if VW_STD17_CONSTEXPR (is_learn)

Seems to make code unnecassarily cryptic.  Release mode with optimization properly compiles this out in msvc and gcc.  If adding VW_STD17_CONSTEXPR makes an expression more efficient, it would make sense to add it then.

vowpalwabbit/binary.cc
```diff
@@ -4,22 +4,38 @@
...
 using namespace VW::config;
+using std::endl;
 
+namespace VW { namespace binary {
 template <bool is_learn>
 void predict_or_learn(char&, LEARNER::single_learner& base, example& ec)
 {
   if (is_learn)
+  {
+    VW_DBG(ec) << "binary: before-base.learn() " << simple_label_to_string(ec) << features_to_string(ec) << endl;
```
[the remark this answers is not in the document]
rajan-chari: There is preceedence in VW for logging code with this style.  See for example search.cc.  This style seems more readable.

vowpalwabbit/debug_log.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+struct vw_dbg
+{
+  const static bool default_log     = false;
```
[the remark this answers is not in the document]
rajan-chari: Done.

vowpalwabbit/example.cc
```diff
@@ -197,6 +197,91 @@ void free_flatten_example(flat_example* fec)
...
+std::string simple_label_to_string(const example& ec)
+{
+  std::stringstream strstream;
+  strstream << "[l=" << ec.l.simple.label << ",w=" << ec.weight << "]";
+  return strstream.str();
+}
+
+std::string depth_indent_string(uint32_t depth)
+{
+  const std::string indent_str = "- ";
```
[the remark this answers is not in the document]
rajan-chari: Done.

vowpalwabbit/example.cc
```diff
@@ -197,6 +197,91 @@ void free_flatten_example(flat_example* fec)
...
+  std::stringstream strstream;
+  strstream << "[p=" << ec.pred.scalar << ", pp=" << ec.partial_prediction << "]";
+  return strstream.str();
+}
+
+std::string a_s_pred_to_string(const example& ec)
+{
+  std::stringstream strstream;
+  strstream << "ec.pred.a_s[";
+  for (uint32_t i = 0; i < ec.pred.a_s.size(); i++)
```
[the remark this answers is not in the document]
rajan-chari: Need to output the index in the loop.

vowpalwabbit/example.h
```diff
@@ -49,11 +49,58 @@ typedef union
...
+};
+
 struct example : public example_predict  // core example datatype.
 {
   // input fields
   polylabel l;
 
+  // Notes: TLDR; needed to make predict() indpendent of label (as it should theoretically should be)
+  // 1) initial used to be in label_data (simple label)
+  // 2) gd.predict() used to use this to load intial value
```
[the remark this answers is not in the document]
rajan-chari: Fixed.

vowpalwabbit/example.h
```diff
@@ -123,3 +172,12 @@ namespace VW
 {
 void return_multiple_example(vw& all, v_array<example*>& examples);
 }  // namespace VW
+std::string features_to_string(const example& ec);
```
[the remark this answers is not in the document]
rajan-chari: Some of these yyy_to_string() functions need access to different parts example.  Seems more clean to do it this way, rather than add additional parameters.

vowpalwabbit/gd.cc
```diff
@@ -38,6 +42,9 @@ using namespace VW::config;
 // 4. Factor various state out of vw&
 namespace GD
 {
+uint32_t stack_depth;
```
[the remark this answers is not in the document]
rajan-chari: Good point.  Fixed.

vowpalwabbit/learner.h
```diff
@@ -142,25 +184,53 @@ struct learner
...
   bool is_multiline;  // Is this a single-line or multi-line reduction?
+  std::string name;   // Name of the reduction.  Used in VW_DBG to trace nested learn() and predict() calls
+  bool predict_before_learn;  // Most reductions need to call predict() before learn().  The prediction
+                                      //   is used for progressive validation.  Some reductions do not
+                                      //   need to call predict() before learn().  For example active.cc
 
   using end_fptr_type = void (*)(vw&, void*, void*);
   using finish_fptr_type = void (*)(void*);
 
+  void print_reduction_name(example& ec, const std::string& msg)
```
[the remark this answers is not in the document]
rajan-chari: Is log_reduction_entry() better?  given that depth is only used for logging?

vowpalwabbit/simple_label.h
```diff
@@ -10,14 +10,23 @@ struct vw;
...
 
-extern label_parser simple_label;
+extern label_parser simple_label_parser;
 
 bool summarize_holdout_set(vw& all, size_t& no_win_counter);
 void print_update(vw& all, example& ec);
 void output_and_account_example(vw& all, example& ec);
+
+namespace VW {
+  const float NA_0 = 0.f;   // constant to signal intitializing unused member
```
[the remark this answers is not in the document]
rajan-chari: Done.

vowpalwabbit/simple_label.h
```diff
@@ -10,14 +10,23 @@ struct vw;
...
+  const float NA_1 = 1.f;    // constant to signal intitializing unused member
```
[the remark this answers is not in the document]
rajan-chari: Done.

vowpalwabbit/v_array.h
```diff
@@ -72,7 +72,7 @@ struct v_array
       T* temp = (T*)realloc(_begin, sizeof(T) * length);
       if ((temp == nullptr) && ((sizeof(T) * length) > 0))
       {
-        THROW_OR_RETURN("realloc of " << length << " failed in resize().  out of memory?");
+        THROW_OR_RETURN_VOID("realloc of " << length << " failed in resize().  out of memory?");
```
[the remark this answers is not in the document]
rajan-chari: Yeah.  Unfortunately, this usage creates tons of warnings in linux so had to revert back here.

vowpalwabbit/learner.h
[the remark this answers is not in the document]
rajan-chari: Changed to log_entry()

vowpalwabbit/cb_explore_adf_greedy.cc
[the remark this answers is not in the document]
jackgerrits: No problem, sounds good.

vowpalwabbit/learner.h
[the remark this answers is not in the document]
jackgerrits: I like that name 👍

vowpalwabbit/v_array.h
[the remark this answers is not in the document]
jackgerrits: Which issue? The warning were fixed I believe

vowpalwabbit/binary.cc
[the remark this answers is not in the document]
jackgerrits: Hmm there are examples, but we are nowhere near consistent within VW for this kind of thing. THROW and THROW ex use it in the way I described.

vowpalwabbit/gd.cc
```diff
@@ -56,11 +56,16 @@ struct gd
   bool adaptive_input;
   bool normalized_input;
   bool adax;
-  uint32_t stack_depth;
-
   vw* all;  // parallel, features, parameters
 };
 
+uint32_t gd_stack_depth;
```
jackgerrits: Does this need to be a global?

vowpalwabbit/cb_explore_adf_common.h
```diff
@@ -113,12 +110,16 @@ inline void cb_explore_adf_base<ExploreType>::learn(
   example* label_example = CB_ADF::test_adf_sequence(examples);
   if (label_example != nullptr)
   {
+    // Notes:  Label exists so call learn()
     data._known_cost = CB_ADF::get_observed_cost(examples);
     // learn iff label_example != nullptr
     data.explore.learn(base, examples);
   }
-  else
+  else if (!examples[0]->predict_called_before_learn)
```
JohnLangford: Why does the reduction not know whether or not it needs to call predict first?

vowpalwabbit/binary.cc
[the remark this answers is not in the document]
rajan-chari: I was planning to leave it in.  It is designed to have no perf impact when disabled which is the default.

Details below:
1) VW_DBG is protected by if(compile_time_constant).  This should allow for zero runtime impact if disabled.
2) These constants usually have cc file level granularity so can be turned on and off at various levels.
3) When predict/learn debug statements are included, it prints indented debug statements along with predict/learn reduction in context.

vowpalwabbit/cb_algs.h
```diff
@@ -21,10 +21,10 @@ template <bool is_learn>
 float get_cost_pred(
     LEARNER::single_learner* scorer, CB::cb_class* known_cost, example& ec, uint32_t index, uint32_t base)
 {
+  VW_DBG(ec) << "get_cost_pred:" << is_learn << std::endl;
```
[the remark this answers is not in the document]
rajan-chari: I was planning to leave it in.  Details addressed in comments above.

vowpalwabbit/cb_explore.cc
```diff
@@ -94,6 +97,9 @@ void predict_or_learn_greedy(cb_explore& data, single_learner& base, example& ec
     base.predict(ec);
 
   // pre-allocate pdf
+
+  VW_DBG(ec) << "cb_explore: " << (is_learn ? "learn() " : "predict() ") << multiclass_pred_to_string(ec) << endl;
```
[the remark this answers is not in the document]
rajan-chari: I tried that first.  It became noisy and impractical with complex reduction stacks and issues in large example files.  

So I shifted to being able to control this at a granular level.

test/train-sets/ref/ccb_test.predict
```diff
@@ -87,8 +87,8 @@
 1:0.966667,0:0.0166667,3:0.0166667
 
 4:0.96,2:0.01,1:0.01,0:0.01,3:0.01
-2:0.9625,0:0.0125,3:0.0125,1:0.0125
-1:0.966667,0:0.0166667,3:0.0166667
+1:0.0125,0:0.0125,3:0.0125,2:0.9625
```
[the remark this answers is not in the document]
rajan-chari: I vaguely recall that being the case.  Will followup with the commits around it to be sure.

vowpalwabbit/cb_explore_adf_common.h
rajan-chari: I am not very happy with having to expose this.  This was part of the changes made to improve efficiency.  I need to refresh my memory about the issue.  Will provide details once I review my commits.

explore/future_compat.h
```diff
@@ -12,8 +12,13 @@
 
 #ifdef HAS_STD17
 #define ATTR(name) [[ name ]]
+// NOTE: c++ 17 detection defined above is not working properly for msvc.
```
jackgerrits, line 15: I recently made changes that should fix this - can you please check and revert this if it is fixed?

vowpalwabbit/cb_adf.cc
```diff
@@ -223,6 +228,12 @@ void cb_adf::learn_SM(multi_learner& base, multi_ex& examples)
   }
 }
 
+void cb_adf::save_prediction(const action_scores& a_s)
+{ copy_array(_a_s, a_s); }
```
jackgerrits, line 232: I think with the style guidelines for braces we have in place it should either be a single line (with the rest of the signature) or fully expanded. `clang-format` will be able to confirm :)

vowpalwabbit/cb_sample.cc
```diff
@@ -20,82 +20,106 @@ struct cb_sample_data
...
   explicit cb_sample_data(std::shared_ptr<rand_state> &&random_state) : _random_state(random_state) {}
 
-  template <bool is_learn>
-  inline void learn_or_predict(multi_learner &base, multi_ex &examples)
+  inline void predict(multi_learner &base, multi_ex &examples)
   {
-    multiline_learn_or_predict<is_learn>(base, examples, examples[0]->ft_offset);
+    // Call base predict() to get action scores for all actions
+    multiline_learn_or_predict<false>(base, examples, examples[0]->ft_offset);
+    example *first_example = examples[0];
```
jackgerrits, line 27:
```suggestion
    example* first_example = examples[0];
```

vowpalwabbit/cb_sample.cc
```diff
@@ -20,82 +20,106 @@ struct cb_sample_data
...
+    // Seed not found in the example.  Get the random seed from the default provider.
+    const uint64_t seed = _random_state->get_current_state();
+    // Advance the random state since we are going to use the seed 
+    _random_state->get_and_update_random();
+    return seed;
+  }
+
+  void sample(example *first_example)
+  {
+    auto action_scores = first_example->pred.a_s;
```
jackgerrits, line 94: Should be a ref

```suggestion
    auto& action_scores = first_example->pred.a_s;
```

vowpalwabbit/csoaa.cc
```diff
@@ -54,12 +59,11 @@ inline void inner_loop(single_learner& base, example& ec, uint32_t i, float cost
 template <bool is_learn>
 void predict_or_learn(csoaa& c, single_learner& base, example& ec)
 {
-  // std::cerr << "------------- passthrough" << std::endl;
   COST_SENSITIVE::label ld = ec.l.cs;
   uint32_t prediction = 1;
   float score = FLT_MAX;
   size_t pt_start = ec.passthrough ? ec.passthrough->size() : 0;
-  ec.l.simple = {0., 0., 0.};
+  ec.l.simple = {0., VW::UNUSED_1, VW::UNUSED_0};
```
jackgerrits, line 66: What is the deprecated plan for `VW::UNUSED_1`?

vowpalwabbit/ect.cc
```diff
@@ -66,7 +66,7 @@ struct ect
 
 bool exists(v_array<size_t> db)
```
jackgerrits, line 67:
```suggestion
bool exists(cosnt v_array<size_t>& db)
```

vowpalwabbit/example.cc
```diff
@@ -197,6 +197,91 @@ void free_flatten_example(flat_example* fec)
...
+    for (; ind_iter != f.indicies.cend(); ++ind_iter, ++val_iter)
+    {
+      strstream << "[h=" << *ind_iter << ","
+                << "v=" << *val_iter << "]";
+    }
+  }
+  return strstream.str();
+}
+
+std::string cb_label_to_string(const example& ec)
```
jackgerrits, line 217: Parameter should be the cb label

vowpalwabbit/example.cc
```diff
@@ -197,6 +197,91 @@ void free_flatten_example(flat_example* fec)
...
+  strstream << "ec.pred.a_s[";
+  for (uint32_t i = 0; i < ec.pred.a_s.size(); i++)
+  {
+    strstream << "(" << i << " = " << ec.pred.a_s[i].action << ", " << ec.pred.a_s[i].score << ")";
+  }
+  strstream << "]";
+  return strstream.str();
+}
+
+std::string multiclass_pred_to_string(const example& ec)
```
jackgerrits, line 278: Parameter should be the specific type

vowpalwabbit/example.cc
```diff
@@ -197,6 +197,91 @@ void free_flatten_example(flat_example* fec)
...
+std::string depth_indent_string(const multi_ex& ec) { return depth_indent_string(*ec[0]); }
+
+std::string scalar_pred_to_string(const example& ec)
+{
+  std::stringstream strstream;
+  strstream << "[p=" << ec.pred.scalar << ", pp=" << ec.partial_prediction << "]";
+  return strstream.str();
+}
+
+std::string a_s_pred_to_string(const example& ec)
```
jackgerrits, line 266: Parameter should be the specific type

vowpalwabbit/example.h
```diff
@@ -49,11 +49,58 @@ typedef union
...
+    return false;
+  }
+
+  void inline set_value(uint64_t ft_offset, const polyprediction& pred)
+  {
+    if (_cache.size() <= ft_offset)
+    {
+      _cache.resize(ft_offset + 1);
+    }
+    _cache[ft_offset] = pred;
```
jackgerrits, line 73: Copying a union seems like a dangerous pattern to be using here

vowpalwabbit/example.h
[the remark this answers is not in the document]
jackgerrits: I still don't see why intentionally restricting where these functions can be used is worth it? They should probably be implemented near the types anyway.

After looking at them, I can see some of then *do* require example, but definitely not all.

vowpalwabbit/gd.cc
jackgerrits: Bump on here

vowpalwabbit/memory_tree.cc
```diff
@@ -669,6 +673,11 @@ float F1_score_for_two_examples(example& ec1, example& ec2)
 
 void predict(memory_tree& b, single_learner& base, example& ec)
 {
+  // measure predict performance if test mode is set (i.e.  we lead memory tree model from file)
+  clock_t begin;
+  if (b.test_mode == false)
```
jackgerrits, line 678: Can this be placed behind a macro or compile time switch?

vowpalwabbit/learner.h
```diff
@@ -205,7 +274,9 @@ struct learner
     assert((is_multiline && std::is_same<multi_ex, E>::value) ||
         (!is_multiline && std::is_same<example, E>::value));  // sanity check under debug compile
     increment_offset(ec, increment, i);
+    log_entry(ec, "learn");
```
lalo, line 277: this call and log_exit would still happen if you disable VW_DBG

lalo:
can you merge back the latest master branch to get rid of the conflicts that block the CI from running?

vowpalwabbit/example.h
```diff
@@ -100,6 +147,8 @@ struct example : public example_predict  // core example datatype.
       "in_use has been removed, examples taken from the pool are assumed to be in use if there is a reference to them. "
       "Standalone examples are by definition always in use.")
   bool in_use = true;
+
+  bool predict_called_before_learn; // If the driver calls predict before learn, this flag will be set
```
lalo, line 151: where is this being used?

vowpalwabbit/learner.h
```diff
@@ -465,101 +550,110 @@ VW_WARNING_STATE_POP
...
   }
 };
 
 template <class T, class E, class L>
 learner<T, E>& init_learner(free_ptr<T>& dat, L* base, void (*learn)(T&, L&, E&), void (*predict)(T&, L&, E&),
-    size_t ws, prediction_type_t pred_type)
+    size_t ws, prediction_type_t pred_type, const std::string& name, bool predict_before_learn = true)
 {
-  auto ret = &learner<T, E>::init_learner(dat.get(), base, learn, predict, ws, pred_type);
+  auto ret = &learner<T, E>::init_learner(dat.get(), base, learn, predict, ws, pred_type, name, predict_before_learn);
```
lalo, line 565: what exactly do you mean with "predict_before_learn" when it is true vs when it is false? is this state necessary?

vowpalwabbit/log_multi.cc
```diff
@@ -303,7 +303,8 @@ void predict(log_multi& b, single_learner& base, example& ec)
 {
   MULTICLASS::label_t mc = ec.l.multi;
 
-  ec.l.simple = {FLT_MAX, 0.f, 0.f};
+  ec.l.simple = {FLT_MAX, VW::UNUSED_1, VW::UNUSED_0};
```
lalo, line 306: if they are unused why does the naming convention matter? what does _1 and _0 represent?

vowpalwabbit/learner.h
lalo: I read the code, but I'm curious about the intent of keeping track of this state.

lalo (changes requested):
Can you also add tests that exercise the new functionality? That actually would be good to add on a separate PR and merge that asap to get a better diff of improvement in this PR.

vowpalwabbit/cb_adf.cc
```diff
@@ -288,18 +293,13 @@ example* test_adf_sequence(multi_ex& ec_seq)
...
 }
 
-template <bool is_learn>
-void cb_adf::do_actual_learning(multi_learner& base, multi_ex& ec_seq)
+void cb_adf::learn(multi_learner& base, multi_ex& ec_seq)
 {
   _offset = ec_seq[0]->ft_offset;
   _gen_cs.known_cost = get_observed_cost(ec_seq);  // need to set for test case
-  if (is_learn && test_adf_sequence(ec_seq) != nullptr)
+  if (test_adf_sequence(ec_seq) != nullptr)
```
lalo, line 300: this refactor is not equivalent, not sure if it's by design. What happens if in the past you would have to support call of `is_learn = true and test_adf_sequence(ec_seq) == nullptr`. After this change they would produce different calls, since the new learn() does not have the else clause anymore.

vowpalwabbit/example_predict.h
```diff
@@ -42,7 +43,8 @@ struct example_predict
 
   v_array<namespace_index> indices;
   std::array<features, NUM_NAMESPACES> feature_space;  // Groups of feature values.
-  uint64_t ft_offset;                                  // An offset for all feature values.
+  uint64_t ft_offset;                                  // SAn offset for all feature values.
+  uint32_t stack_depth;
```
lalo, line 47: why is stack_depth needed as part of this class, not following that clearly but intuitively feels like coupling of reduction stack with example

vowpalwabbit/global_data.h
```diff
@@ -1,531 +1,535 @@
...
+  bool progress_add;   // additive (rather than multiplicative) progress dumps
+  float progress_arg;  // next update progress dump multiplier
+
+  std::map<uint64_t, std::string> index_name_map;
+
+  label_type_t label_type;
+
+  polyprediction _predict_buffer;
+  void cleanup_prediction();
+  void copy_prediction(const polyprediction& from_pred);
```
jackgerrits, line 504: Can these be made private?

vowpalwabbit/global_data.cc
```diff
@@ -272,6 +300,86 @@ vw_ostream::vw_ostream() : std::ostream(&buf), buf(*this), trace_context(nullptr
...
+      break;
+    case prediction_type_t::scalar:
+      _predict_buffer = from_pred;
+      break;
+    default:
+        THROW("Unhandled prediction type");
+  }
+}
+
+void vw::cleanup_prediction()
```
jackgerrits, line 345: This whole function just hammers home how desperately we need these types to be able to have destructors

vowpalwabbit/global_data.cc
```diff
@@ -272,6 +300,86 @@ vw_ostream::vw_ostream() : std::ostream(&buf), buf(*this), trace_context(nullptr
 VW_WARNING_STATE_PUSH
 VW_WARNING_DISABLE_DEPRECATED_USAGE
 
+void vw::copy_prediction(const polyprediction& from_pred)
```
jackgerrits, line 303: Same for this one and copy constructors

vowpalwabbit/cb_sample.cc
```diff
@@ -78,14 +81,34 @@ struct cb_sample_data
...
     auto result = exploration::swap_chosen(action_scores.begin(), action_scores.end(), chosen_action);
     assert(result == S_EXPLORATION_OK);
 
+    VW_DBG(examples) << "cb " << cb_decision_to_string(examples[0]->pred.a_s) <<
+                    " rnd:" << _random_state->get_current_state() << std::endl;
+
     _UNUSED(result);
   }
 
+  std::string cb_decision_to_string(const ACTION_SCORE::action_scores& action_scores)
```
jackgerrits, line 96: You can just add an overload for `operator<<` in v_array.h to handle this in a uniform way that will be discoverable and usable elsewhere in the codebase

vowpalwabbit/global_data.h
```diff
@@ -1,531 +1,535 @@
...
-void noop_mm(shared_data*, float label);
-void get_prediction(VW::io::reader* f, float& res, float& weight);
-void compile_gram(
-    std::vector<std::string> grams, std::array<uint32_t, NUM_NAMESPACES>& dest, char* descriptor, bool quiet);
-void compile_limits(std::vector<std::string> limits, std::array<uint32_t, NUM_NAMESPACES>& dest, bool quiet);
-
-VW_DEPRECATED("Use print_tag_by_ref instead")
-int print_tag(std::stringstream& ss, v_array<char> tag);
-int print_tag_by_ref(std::stringstream& ss, const v_array<char>& tag);
+// Copyright (c) by respective owners including Yahoo!, Microsoft, and
```
jackgerrits, line 1: Looks like a whitespace change leaves the whole file as modified. Can you please fix this?

vowpalwabbit/example_predict.h
rajan-chari: I use VW_DBG to debug learn/predict and show state of reductions in the context of processing a given example.  Tracking this in example as it traverses the stack seemed like a good way to do it.

jackgerrits: +1 to this

vowpalwabbit/global_data.h
rajan-chari: Done.

vowpalwabbit/parse_args.cc
```diff
@@ -219,6 +219,7 @@ void parse_dictionary_argument(vw& all, const std::string& str)
   {
     fd = VW::io::open_file_reader(fname);
   }
+
```
jackgerrits, line 222: remove this added newline between try and catch

vowpalwabbit/cb_sample.cc
rajan-chari: The following does not seem appropriate in varray.h since it is a specific interpretation of what the action, score collection means here.

        ostrm << "chosen[" << a_s.action << "," << a_s.score << "] [ "

jackgerrits: Yes, I agree. But you can reasonably provide a `operator<<` impl for `action_score` and then the existing `operator<<` for `v_array<T>` would utilize this when printing.

vowpalwabbit/example_predict.h
lalo: The fact that a 'debug build' debug statements now force abstractions to get coupled with other concepts of other abstractions is a hard sell for me. I suggest you either flesh out a design doc about debug statements to see how to address it or remove it and leave it for ad-hoc debugging since it's been helpful for you.

vowpalwabbit/learner.h
```diff
@@ -186,18 +228,44 @@ struct learner
...
   std::shared_ptr<void> learner_data;
+
   learner(){};  // Should only be able to construct a learner through init_learner function
+
  public:
   prediction_type_t pred_type;
   size_t weights;  // this stores the number of "weight vectors" required by the learner.
   size_t increment;
   bool is_multiline;  // Is this a single-line or multi-line reduction?
+  std::string name;   // Name of the reduction.  Used in VW_DBG to trace nested learn() and predict() calls
```
lalo, line 240: my comment on example_predict.h applies here too, I find it concerning that debug only build is now enforcing requirements on abstractions. I might be too strict though.

vowpalwabbit/global_data.h
```diff
@@ -1,531 +1,535 @@
...
+
+  // Set by --progress <arg>
+  bool progress_add;   // additive (rather than multiplicative) progress dumps
+  float progress_arg;  // next update progress dump multiplier
+
+  std::map<uint64_t, std::string> index_name_map;
+
+  label_type_t label_type;
+
+  polyprediction _predict_buffer;
```
lalo, line 502: where do you see this _predict_buffer landing after `all` struct stops being the one place to dump everything?

vowpalwabbit/global_data.h
rajan-chari: Where is the extra white space?  I can't find it.

jackgerrits: Sorry I meant line endings

vowpalwabbit/global_data.h
rajan-chari: Good question.  This should live in the "orchestrator".  Since vw is currently the orchestrator, this seems like an appropriate place for it.    In the future if there is a separate orchestrator we should revisit this.

Should probably be private also since the functions using this are now private.  Will do that.

vowpalwabbit/global_data.h
lalo: it's probably a windows style vs unix style line endings, so whatever is on this branch it probably is the opposite? depending on your git config it might muck with them too so if you did initially this change in another computer it might be a bit harder to decipher.

vowpalwabbit/global_data.cc
```diff
@@ -167,7 +167,20 @@ void vw::learn(example& ec)
...
   else
-    VW::LEARNER::as_singleline(l)->learn(ec);
+  {
+    if (l->predict_before_learn)
+    {
+      VW::LEARNER::as_singleline(l)->predict(ec);
+      copy_prediction(ec.pred);
+      auto restore_guard = VW::scope_exit([&ec, this]
+      {
+        std::swap(ec.pred, _predict_buffer);
```
jackgerrits, line 177: If `_predict_buffer` is only used in the predict and learn functions it really should be a local variable.

vowpalwabbit/example_predict.h
rajan-chari: This works in debug and release mode.

I view this as general low impact trace facility for developing reductions.  The lack of trace facility in reductions was a substantial issue both in developing continuous actions and with predict/learn changes.  

There are other changes planned to predict/learn in the near future where this will be important.  One of these changes is that learn() can avoid a substantial number of predictions in library mode.    

Perhaps best discussed during the VW session as I know this has been a point of contention in the past.

vowpalwabbit/global_data.cc
rajan-chari: Yeah.  But unions...

vowpalwabbit/global_data.cc
rajan-chari: Yeah.  But unions...

vowpalwabbit/learner.h
rajan-chari: Yes I understand the general object to having this.  I responded to you earlier comments.  This is a point of contention.  We should discuss during the VW meeting.

vowpalwabbit/global_data.cc
rajan-chari: Having a non-local variable prevents allocations in steady state.

jackgerrits: It's really quite a small allocation for this and in an effort to avoid it you are adding another field (union at that) to the god object we are trying to get rid of/reduce

vowpalwabbit/gd.cc
rajan-chari: Removed.

vowpalwabbit/global_data.h
rajan-chari: Got it..  Will check out the line endings.  Thanks.

rajan-chari:
> Which issue? The warning were fixed I believe

vowpalwabbit/v_array.h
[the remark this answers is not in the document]
rajan-chari: Fixed

test/train-sets/ref/ccb_test.predict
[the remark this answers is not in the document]
rajan-chari: Resolved.

vowpalwabbit/cb_adf.cc
```diff
@@ -90,9 +93,11 @@ struct cb_adf
   void learn_DR(multi_learner& base, multi_ex& examples);
   void learn_DM(multi_learner& base, multi_ex& examples);
   void learn_SM(multi_learner& base, multi_ex& examples);
-
   template <bool predict>
   void learn_MTR(multi_learner& base, multi_ex& examples);
+
+  void save_prediction(const action_scores& a_s);
```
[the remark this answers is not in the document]
rajan-chari: Removed.  Predictions are saved in global_data.cc where learn() is called after predict().  Prediction is restored there after learn() is called since sometimes learn() changes predictions.

vowpalwabbit/active.cc
```diff
@@ -171,11 +171,11 @@ base_learner* active_setup(options_i& options, vw& all)
   // Create new learner
   learner<active, example>* l;
   if (options.was_supplied("simulation"))
-    l = &init_learner(data, base, predict_or_learn_simulation<true>, predict_or_learn_simulation<false>);
+    l = &init_learner(data, base, predict_or_learn_simulation<true>, predict_or_learn_simulation<false>, "active-simulation", false);
```
lalo, line 174: so how would you explain to someone making a new reduction whether or not they need to call into predict before learning? is it based on the reduction itself they are coding or on the stack or what else? or how are you planning on document this change in learner by adding "predict_before_learn" bool

vowpalwabbit/learner.h
lalo, line 240: if I have two instances of learner objects, and I fetch the name of each and both reply back with the same exact name what can one infer from that?

lalo:
It would be nice if you update the description of this PR with pending work, I see a couple of comments spread around files (some seem permanent, some other seem temporary, or stale) so it would be good if you double check all new comments added (and remove stale ones) on this PR and then be more descriptive of next steps / or offshoot work triggered by this PR (ccb?). Also, there's small 'nit' level feedback on a lot of files which I didn't leave on github since its mostly extra white space, or missing white space in between chars, funky indentation. It would also be good if you do a pass to clean that up.

vowpalwabbit/cb_sample.cc
rajan-chari: Done.

vowpalwabbit/global_data.h
rajan-chari: Done.

vowpalwabbit/parse_args.cc
rajan-chari: Done.

vowpalwabbit/global_data.cc
rajan-chari: It's possible to have predictions with many elements.  With cb_adf_explore, ccb, and continuous actions it's possible to have non-trivial amount of elements in predictions.  This pattern is used in many reductions to avoid allocations. 

_predict_buffer is used by vw for coordinating between learn, predict and finish_example at a driver level.  This is different than adding something to struct vw that is use by one or even a couple of reductions.  It seems appropriate to add it here.

vowpalwabbit/active.cc
rajan-chari: In new reductions, if learn does not need prediction, the default work.  This is the common expected use case and default arguments for init functions are setup this way.

If reductions need predictions during learn, then things need to considered more carefully for efficiency reasons.  If there is a lot of extra setup for predictions, and the same work needs to happen for learn then predict_before_learn is set to false (for example ccb).  In this case learn for a reduction must call predict during learn.

vowpalwabbit/learner.h
rajan-chari: This is used to generate trace of execution stack and report state of reductions in context.  No assumption is made about the uniqueness of names.

Here is an example:

    - [shared_feature_merger.learn]
        - [cb_sample.learn]
            - [cb_explore_adf-greedy.learn]
                - [cb_adf.learn]
                  - cs_ldf_learn: ex=141, offset=0, id=0
                    - [csoaa_ldf_rank.predict]
                        - [scorer-identity.predict]
                            - [gd.predict]
                            -
                        -
                        - [scorer-identity.predict]
                            - [gd.predict]
                            -
                        -
                        - [scorer-identity.predict]
                            - [gd.predict]
                            -
                        -
                    -

vowpalwabbit/cb_adf.cc
rajan-chari: test_adf_sequence checks for the existence of a valid label.  This is necessary condition for all the learn functions in the switch statement below.

Correctness was verified by RunTests that exercises this code.

vowpalwabbit/learner.h
rajan-chari: I covered this in another comment.

vowpalwabbit/log_multi.cc
rajan-chari: The constants are defined as 1 and 0

vowpalwabbit/learner.h
rajan-chari: All the statements inside the function get compiled out if constants are false.

vowpalwabbit/example.h
rajan-chari: This is was an experiment.  Not used.  Removed.

vowpalwabbit/learner.h
lalo: I'm aware that's the only current usage of 'name' now but having a public member variable called 'name' on an object called 'learner' has implications for all the code base and for future work. 'name' has very strong connotations and I believe it will cause confusion at some point since 'name' only returns the top most level reduction name and hiding away whatever else is on that learner stack. Two learner instances might return the same name but will behave differently (when calling predict/learn) if they have different stacks but only coincide on the top level which is a bit dissonant.

vowpalwabbit/global_data.cc
jackgerrits: Without a performance test showing that this is required and the effects of this in real world use I still feel opposed to this driver level state.

vowpalwabbit/learner.h
jackgerrits: I agree with @lalo. A publicly accessible field called `name` seems dangerous here.

vowpalwabbit/cb_sample.cc
jackgerrits: @rajan-chari My comment was that the entire `cb_decision_to_string` function is not necessary. Now that you have an `operator<<` for `action_score` this entire function can be replaced with simply passing a `v_array<T>` to `operator<<`

vowpalwabbit/global_data.cc
rajan-chari: I understand the nature of objection.  I will setup a session Monday to review the contentious changes.   I will add this to the discussion list.  Does that work?

vowpalwabbit/learner.h
rajan-chari: "name" does make a strong statement and does not reflect the semantics.  Does trace_name or trace_reduction_name work?  If we find other uses for this in the future and provide stronger guarantees, we can change the name to reflect the semantics.  Does that address the concern?

vowpalwabbit/global_data.cc
```diff
@@ -167,7 +167,20 @@ void vw::learn(example& ec)
   if (ec.test_only || !training)
     VW::LEARNER::as_singleline(l)->predict(ec);
   else
-    VW::LEARNER::as_singleline(l)->learn(ec);
+  {
+    if (l->predict_before_learn)
```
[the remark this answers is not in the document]
rajan-chari: The goal for this PR was to introduce predict/learn semantics and switch to it where possible.  Initially this covered many more reductions but due to efficiency reasons this had to be rolled back.  This flag was introduced to address the efficiency issues.

Next steps:
1) All new reductions from now on should use the new semantics
2) There are some high value changes around library mode (predict file, progressive validation) that can take advantage of new semantics.  Along with these changes, I expect more reductions to migrate to new semantics.  The exact order will depend on prioritization and scheduling.
3) Other reductions beyond what's in step (2) will have to be addressed depending on priority.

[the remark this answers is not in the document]
lalo: Can you answer the first question though? was that intended?

vowpalwabbit/learner.h
olgavrou: Isn't passing in a name at the learner for each reduction opposite of what we were discussing for this PR (https://github.com/VowpalWabbit/vowpal_wabbit/pull/2531)?

vowpalwabbit/learner.h
```diff
@@ -122,23 +127,59 @@ inline float noop_sensitivity(void*, base_learner&, example&)
...
+inline void decrement_depth(multi_ex& ec_seq)
+{
+  if(vw_dbg::track_stack){
+    for (auto& ec : ec_seq)
+    {
+      --ec->stack_depth;
+    }
+  }
+}
+
```
olgavrou, line 160: I am concerned about all of this debug functionality. It seems like there is state and logic around the debug logs. It looks like we are adding cognitive load to the developers with this. If we forget to increment/decrement what are the chances of us spending time debugging the debugging logs? One thing could be to leave the functionality here should we need it in the future, but remove it from the reduction code (I think someone else proposed this too).

vowpalwabbit/learner.h
rajan-chari: The objection I am aware of was in exposing more metadata about the reduction in parse_args (i.e. outside the reduction).

olgavrou: I think it was about not asking the reduction writer to pass in a name and having them have to think about this too, no? @lalo ?

lalo: Yes, both of you are right. We can ask John what he thinks since he was the one that raised concerns with that in the previous discussion.

vowpalwabbit/learner.h
rajan-chari: It's unfortunate that I called this VW_DBG instead of VW_TRACE.  

It's true that we are tracking the depth of stack for trace output.  This is isolated in the learner and is transparent to reduction writers.

At this point using VW_DBG in a reduction just works and produces output in the context of the reduction stack.  Trying to trace how prediction or learn works in a reduction stack is substantially harder without this.

vowpalwabbit/global_data.cc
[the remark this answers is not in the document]
rajan-chari: Yes.  This is intended.  Predictions in top-most reduction are used in progressive validation, output to screen, written to predict files so always needed.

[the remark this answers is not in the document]
lalo: wouldn't this imply that if you have a stack with the following:
`reduction(predict_before_learn=True) -> reduction(predict_before_learn=False) -> ... -> base_reduction()` you would effectively add an extra call to predict with zero benefit?

vowpalwabbit/debug_log.h
```diff
@@ -2,9 +2,50 @@
...
+      get_pmf     | sample_pdf  | cb_explore_get_pmf  | cb_sample_pdf | ccb        |
+      scorer      | cb_sample;
+};
+
+#define VW_DEBUG_LOG vw_dbg::default_log
+
+#define VW_LOG_SINK std::cout
+
+#define VW_DBG(e)                       \
+  if VW_STD17_CONSTEXPR (VW_DEBUG_LOG)  \
```
olgavrou, line 46: if hasn't `STD17` what happens?

vowpalwabbit/example.h
```diff
@@ -80,6 +117,16 @@ struct example : public example_predict  // core example datatype.
   // input fields
   polylabel l;
 
+  // Notes: TLDR; needed to make predict() independent of label (as it should theoretically should be)
```
olgavrou, line 120: do we need this comment to be committed to the code base? now that `initial` is not part of label it seems like the comment is just keeping history around (explanation of this change could be a comment on the PR)

vowpalwabbit/oaa.cc
```diff
@@ -34,7 +33,7 @@ void learn_randomized(oaa& o, VW::LEARNER::single_learner& base, example& ec)
   if (ld.label == 0 || (ld.label > o.k && ld.label != (uint32_t)-1))
     std::cout << "label " << ld.label << " is not in {1," << o.k << "} This won't work right." << std::endl;
 
-  ec.l.simple = {1., 0.f, 0.f};  // truth
+  ec.l.simple = {1., VW::UNUSED_1, VW::UNUSED_0};  // truth
```
olgavrou, line 36: `UNUSED_0` for both?

vowpalwabbit/recall_tree.cc
```diff
@@ -313,7 +312,7 @@ predict_type predict_from(recall_tree& b, single_learner& base, example& ec, uin
   MULTICLASS::label_t mc = ec.l.multi;
   uint32_t save_pred = ec.pred.multiclass;
 
-  ec.l.simple = {FLT_MAX, 0.f, 0.f};
+  ec.l.simple = {FLT_MAX, VW::UNUSED_1, VW::UNUSED_0};
```
olgavrou, line 315: `UNUSED_0` for both?

vowpalwabbit/recall_tree.cc
```diff
@@ -355,9 +354,19 @@ float train_node(recall_tree& b, single_learner& base, example& ec, uint32_t cn)
...
-  float imp_weight = fabs((float)(delta_left - delta_right));
 
-  ec.l.simple = {route_label, imp_weight, 0.};
+  // Bug?
+  // Notes: imp_weight does not seem to be used since it was not set to
+  // ec.weight before.  ec.l.simple.weight was not used in gd.
+  // float imp_weight = fabs((float)(delta_left - delta_right));
+
+  ec.l.simple = {route_label, VW::UNUSED_1, VW::UNUSED_0};
+  // Bug?
```
olgavrou, line 364: is this a bug? should it be documented?

vowpalwabbit/recall_tree.cc
olgavrou: oh, they are unused members so I guess it doesn't matter?

vowpalwabbit/example.h
```diff
@@ -60,6 +60,43 @@ typedef union
   VW::continuous_actions::probability_density_function_value pdf_value;  // probability density value for a given action
 } polyprediction;
 
+class gd_prediction_cache
```
olgavrou, line 63: where is this `get_prediction_cache` used? I can't find it being used anywhere

vowpalwabbit/global_data.cc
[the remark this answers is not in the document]
rajan-chari: A reduction with predict_before_learn == True supports new semantics for predict/learn.  If the framework calls predict before learn, the top level reduction will likely induce calls to predict all the way down the chain.  Since the framework calls learn after predict, it's possible that one of reductions down the chain makes extra calls to predict.  However, caching predictions downstream fails because many calls to base predict from reductions deeper in the stack.  As a rough experiment, I attempted to cache predictions in gd.  This failed because reductions can modify examples and make predict calls to the same learner.
The solution to more efficient learn is to a) detect conditions where predict is not needed for learn b) let the framework avoid base level predict call.

vowpalwabbit/debug_log.h
rajan-chari: The keyword constexpr will not get used since it won't be compatible with earlier compilers.

vowpalwabbit/oaa.cc
rajan-chari: Yes this was deliberate.  One is weight (1) the other is cost (0).  Did not make sense to initialize weight to 0.  Although I get that this is not a strict replacement of the code before.

vowpalwabbit/recall_tree.cc
rajan-chari: Addressed this in an earlier comment.  This is deliberate.

vowpalwabbit/example.h
rajan-chari: This was an experiment.  It has been removed.

rajan-chari: I see what happened.  Did not show up as a change in my vscode.  Fixed.

vowpalwabbit/recall_tree.cc
rajan-chari: Let's ask John during PR review.  A can be raised in vw repo if needed.

vowpalwabbit/debug_log.h
```diff
@@ -2,9 +2,50 @@
...
+  static constexpr bool csoaa               = false;
+  static constexpr bool cs_active           = false;
+  static constexpr bool cats_tree           = false;
+  static constexpr bool cats_pdf            = false;
+  static constexpr bool cats                = false;
+  static constexpr bool cb_explore_pdf      = false;
+  static constexpr bool cb_explore          = false;
+  static constexpr bool cbify               = false;
+  static constexpr bool get_pmf             = false;
+  static constexpr bool sample_pdf          = false;
```
lalo, line 26: so you have to manually add a boolean here that sort of matches the name used in that file that sort of matches the string name passed to learner? isn't that a bit excessive

vowpalwabbit/debug_log.h
```diff
@@ -2,9 +2,50 @@
...
+#include "future_compat.h"
 #pragma once
 
-#define VW_DEBUG_ENABLE(x) static bool VW_DEBUG_LOG = x;
+struct vw_dbg
+{
+  static constexpr bool default_log         = false;
+  static constexpr bool learner             = false;
+  static constexpr bool search              = false;
+  static constexpr bool gd                  = false;
```
lalo, line 13: is this a list of reductions that currently have vw_dbg logging inside of it, and if there's a reduction name missing from here what is the meaning you can infer from it?

vowpalwabbit/debug_log.h
rajan-chari: These booleans define different scopes that can be enabled or not.  If you don't need a new scope it uses the default_log scope.

vowpalwabbit/debug_log.h
rajan-chari: Any trace statements in other reductions will be enabled/disabled by default_log.

rajan-chari:
> Can you also add tests that exercise the new functionality? That actually would be good to add on a separate PR and merge that asap to get a better diff of improvement in this PR.

All tests in RunTests exercise the new functionality since it's a change at the framework level.

test/Tests.sh
```diff
@@ -0,0 +1,202 @@
+( ../build/vowpalwabbit/vw -k -l 20 --initial_t 128000 --power_t 1 -d train-sets/0001.dat      -f models/0001_1.model -c --passes 8 --invariant      --ngram 3 --skips 1 --holdout_off ) >outputs/1_Test.stdout.track 2>outputs/1_Test.stderr.track
```
jackgerrits, line 1: Please remove one off testing scripts unless they can be adapted to be generally applicable

jackgerrits:
Benchmarks are showing some regressions in the CB path. https://github.com/VowpalWabbit/vowpal_wabbit/pull/2207/checks?check_run_id=2026663809

vowpalwabbit/example.h
```diff
@@ -84,6 +84,20 @@ struct example : public example_predict  // core example datatype.
...
+  // 2) gd.predict() used to use this to load initial value
+  // 3) It also used it as an accumulator and modified it.
+  // 4) This cause two breaches of label independence abstraction during
+  // predict()
+  //      a) All reductions depending on gd had to initialize example.l to sane
+  //      values before base.predict()
+  //      b) All reductions had to save label state before calling
+  //      base.predict()
+  // Making it impossible to remove dependence of predict on label
+  float initial = 0.f;
```
jackgerrits, line 99: I really wish this move of initial from the label to the example was its own PR. It affects a lot and adds significant noise to the other important changes here.

vowpalwabbit/csoaa.cc
```diff
@@ -416,112 +429,128 @@ multi_ex process_labels(ldf& data, const multi_ex& ec_seq_all);
...
+  if (!test_ldf_sequence(data, ec_seq))
+  {
+    if (data.is_wap)
+      do_actual_learning_wap(data, base, ec_seq);
+    else
+      do_actual_learning_oaa(data, base, ec_seq);
+  }
+}
+
+void convert_to_probabilities(multi_ex ec_seq)
```
jackgerrits, line 450: Currently this is a vector copy. Change to reference or const reference

vowpalwabbit/csoaa.cc
```diff
@@ -416,112 +429,128 @@ multi_ex process_labels(ldf& data, const multi_ex& ec_seq_all);
...
-      }
+      ec_seq[k]->pred.a_s = std::move(data.stored_preds[k]);
+      ec_seq[0]->pred.a_s.push_back(data.a_s[k]);
     }
-  }
 
-  /////////////////////// learn
-  if (is_learn)
+    ////////////////////// compute probabilities
+    if (data.is_probabilities) convert_to_probabilities(ec_seq);
```
jackgerrits, line 542:
```suggestion
    if (data.is_probabilities) { convert_to_probabilities(ec_seq); }
```

vowpalwabbit/csoaa.cc
```diff
@@ -416,112 +429,128 @@ multi_ex process_labels(ldf& data, const multi_ex& ec_seq_all);
...
+  if (ec_seq.empty()) return;  // nothing more to do
+
+  uint32_t K = (uint32_t)ec_seq.size();
 
+  /////////////////////// do prediction
+  data.a_s.clear();
+  data.stored_preds.clear();
+
+  auto restore_guard = VW::scope_exit([&data, &ec_seq, K] {
     qsort((void*)data.a_s.begin(), data.a_s.size(), sizeof(action_score), score_comp);
```
jackgerrits, line 532: This should use std::sort instead of qsort.

vowpalwabbit/csoaa.cc
```diff
@@ -738,11 +767,13 @@ void inline process_label(ldf& data, example* ec)
 }
 
 /*
- * The begining of the multi_ex sequence may be labels.  Process those
+ * The beginning of the multi_ex sequence may be labels.  Process those
  * and return the start index of the un-processed examples
  */
 multi_ex process_labels(ldf& data, const multi_ex& ec_seq_all)
 {
+  if (ec_seq_all.empty()) return ec_seq_all;  // nothing to do
```
jackgerrits, line 775:
```suggestion
  if (ec_seq_all.empty()) { return ec_seq_all; }  // nothing to do
```

vowpalwabbit/example.h
```diff
@@ -119,6 +133,7 @@ struct vw;
 struct flat_example
 {
   polylabel l;
+  float weight;  // a relative importance weight for the example, default = 1
```
jackgerrits, line 136: Set the default value?

vowpalwabbit/gen_cs_example.h
```diff
@@ -280,11 +276,13 @@ void call_cs_ldf(VW::LEARNER::multi_learner& base, multi_ex& examples, std::vect
     }
   });
 
-  // 2nd: predict for each ex
-  // // call base.predict for all examples
   if (is_learn)
+  {
+    if (predict_first) base.predict(examples, (int32_t)id);
```
jackgerrits, line 281:
```suggestion
    if (predict_first) { base.predict(examples, (int32_t)id); }
```

vowpalwabbit/multilabel.cc
```diff
@@ -135,11 +137,10 @@ void print_update(vw& all, bool is_test, example& ec)
     if (is_test)
       label_string << " unknown";
     else
-      for (size_t i = 0; i < ec.l.multilabels.label_v.size(); i++) label_string << " " << ec.l.multilabels.label_v[i];
+      for (unsigned int i : ec.l.multilabels.label_v) label_string << " " << i;
```
jackgerrits, line 140: Use the same type as is contained
```suggestion
      for (uint32_t i : ec.l.multilabels.label_v) { label_string << " " << i; }
```

vowpalwabbit/multilabel.cc
```diff
@@ -135,11 +137,10 @@ void print_update(vw& all, bool is_test, example& ec)
...
 
     std::stringstream pred_string;
-    for (size_t i = 0; i < ec.pred.multilabels.label_v.size(); i++)
-      pred_string << " " << ec.pred.multilabels.label_v[i];
+    for (unsigned int i : ec.pred.multilabels.label_v) pred_string << " " << i;
```
jackgerrits, line 143: Use the same type as is contained
```suggestion
    for (uint32_t i : ec.l.multilabels.label_v) { label_string << " " << i; }
```

vowpalwabbit/multilabel.cc
```diff
@@ -154,8 +155,8 @@ void output_example(vw& all, example& ec)
   if (!test_label(ld))
   {
     // need to compute exact loss
-    labels preds = ec.pred.multilabels;
-    labels given = ec.l.multilabels;
+    const labels preds = ec.pred.multilabels;
+    const labels given = ec.l.multilabels;
```
jackgerrits, line 159:
```suggestion
    const labels& preds = ec.pred.multilabels;
    const labels& given = ec.l.multilabels;
```

vowpalwabbit/oaa.cc
```diff
@@ -69,50 +68,75 @@ void learn_randomized(oaa& o, VW::LEARNER::single_learner& base, example& ec)
...
-      ec.l.simple = {(mc_label_data.label == i) ? 1.f : -1.f, 0.f, 0.f};
-      ec.pred.scalar = o.pred[i - 1].scalar;
-      base.update(ec, i - 1);
-    }
-  }
-
+  // Print predictions to a file
   if (print_all)
   {
+    std::stringstream outputStringStream;
```
jackgerrits, line 125: Use snake case for variable names

vowpalwabbit/parser.cc
```diff
@@ -738,6 +739,11 @@ void setup_example(vw& all, example* ae)
   INTERACTIONS::eval_count_of_generated_ft(all, *ae, new_features_cnt, new_features_sum_feat_sq);
   ae->num_features += new_features_cnt;
   ae->total_sum_feat_sq += new_features_sum_feat_sq;
+
+  // Notes: Allows the label parser to update the example after parsing is done.
+  // For example we set the initial value used in gd during predict() (if set)
+  // simple_label_parser is defined with this post parse setup step.
+  if (all.example_parser->lbl_parser.post_parse_setup != nullptr) all.example_parser->lbl_parser.post_parse_setup(ae);
```
jackgerrits, line 746:
```suggestion
  if (all.example_parser->lbl_parser.post_parse_setup != nullptr) { all.example_parser->lbl_parser.post_parse_setup(ae); }
```

jackgerrits (changes requested):
These binary files were added it seems by accident. Some are very large, it's important they are removed before merging.
- test/cb_dm.reg
- test/inv_hash_load_model.vw
- test/marginal_model

Additionally, the refactor around initial is useful but not relevant to "Fix learn semantics". Please submit those changes as a separate PR.

vowpalwabbit/example.h
jackgerrits, line 99: This new field needs to be added to `copy_example_metadata`

vowpalwabbit/example.h
jackgerrits, line 136: This new field needs to be added to `copy_example_metadata`

vowpalwabbit/simple_label.h
```diff
@@ -11,8 +11,14 @@ struct vw;
...
-  float weight = 0.f;
-  float initial = 0.f;
+  // only used for serialization and parsing.  example.weight is used for
+  // computation
+  // DeSerialized/Parsed values are copied into example in VW::setup_example()
+  float serialized_weight = 0.f;
+  // Only used for serialization and parsing.  example.initial is used for
+  // computation
+  // DeSerialized/Parsed values are copied into example in VW::setup_example()
+  float serialized_initial = 0.f;
```
jackgerrits, line 21: Rather than creating a post-parse step to handle cache files the cache file should just be updated to include this value or we should decide it does not need to be cached.

vowpalwabbit/multilabel.cc
rajan-chari: changed to uint32_t.  change to reference label look like a mistake.

vowpalwabbit/multilabel.cc
rajan-chari: done.

vowpalwabbit/example.h
rajan-chari: it's already there.

PR #2208, opened 2019-12-23
Fix most of the deprecation warnings for sklearn 0.22

peterychang:
Some functionality is not easily replaceable, so a future fix will need to be done

Specifically sklearn.linear_model.base will no longer be available in 0.24 and the two objects we pull in from there will be part of their private API, meaning we'll lose the ability to use them directly.


peterychang:
@gramhagen if you can think of any way to rework the sklearn bindings to avoid using the now private API, I'd love the help

python/tests/test_sklearn_vw.py
```diff
@@ -201,7 +201,7 @@ def test_shuffle_pd_Series(self):
         # dummy data in vw format
         X = pd.Series(['1 |Pet cat', '-1 |Pet dog', '1 |Pet cat', '1 |Pet cat'], name='catdog')
 
-        kfold = KFold(n_splits=3, random_state=314, shuffle=False)
+        kfold = KFold(n_splits=3, random_state=314, shuffle=True)
```
peterychang: We need to either set (or leave) random_state=None or set shuffle=True. Seeing the name of the test function, I decided to set shuffle=True

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -347,7 +349,14 @@ def predict(self, X):
             Output vector relative to X.
         """
 
-        check_is_fitted(self, 'fit_')
+        if sklearn.__version__ < '0.22':
```
peterychang: In sklearn <0.22, the second parameter is required. In sklearn ==0.22, it does nothing, and in sklearn >=0.23 it will be removed altogether. We need to do a version check to maintain compatibility between all three versions.

The error message being thrown is copied from the sklearn check_is_fitted function

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -7,9 +7,11 @@
 import io
 
 from scipy.sparse import csr_matrix
+import sklearn
+from sklearn.exceptions import NotFittedError
 from sklearn.base import BaseEstimator, RegressorMixin
 from sklearn.linear_model.base import LinearClassifierMixin, SparseCoefMixin
```
peterychang: These two imports will no longer be available in `sklearn >=0.24`. I'm not sure what we'll need to do to replace them

gramhagen:
@peterychang imo it's not worth the effort of calling check_is_fitted at this point. We can just check if the fit_ attribute is set and otherwise raise the not fitted error for all versions of sklearn.

peterychang:
@gramhagen Sure, I can remove the check_is_fitted call. The bigger problem will be how to handle the removal of these imports from the public API
`from sklearn.linear_model.base import LinearClassifierMixin, SparseCoefMixin`

gramhagen:
yeah, understood. I think what we will need to do is identify which methods we are leveraging from the private apis and internalize them if necessary. I'm not sure what's driving the privatization of those APIs, and it makes me nervous that we could set this up to start diverging from sklearn functionality if not careful.

python/vowpalwabbit/sklearn_vw.py
kumpera: Where's the reference for that?

peterychang: Its in the deprecation warning when importing sklearn.linear_model.base using sklearn 0.22. LinearClassifierMixin and SparseCoefMixin aren't in their public API so we'll lose them when they close off their private API

for the public API:
https://scikit-learn.org/dev/modules/classes.html#module-sklearn.linear_model

PR #2214, opened 2020-01-06
Add chain hash option

### LINKED ISSUE #1935: hashed feature output from vw should output strings that are easy to scrub
[FeatureImportance.zip](https://github.com/VowpalWabbit/vowpal_wabbit/files/3293809/FeatureImportance.zip)
#### Describe the bug
The invert hash strings returned by calling vw with the invert_hash option is returning strings that are difficult to interpret programmatically.

#### To Reproduce
1. Call vw with invert_hash option.

#### Expected behavior
The returned strings must have a delimiter between feature name and hash value. Example:
FromUrl^timeofday=Morning*FromUrl^name=Alice

#### Observed Behavior
Invert hash values that don't have a delimiter between feature name and value. Example:
FromUrl^timeofdayMorning*FromUrl^nameAlice

#### Environment
What version of VW did you use? 8.7.0
What OS or language did you use? Linux

#### Additional context
Attached is a zip file showing output from different examples.

cheng-tan:
Closes #1935


JohnLangford:
This is creating a divergence between the text and json input formats.  In particular, the semantics of key/value pairs where value is a string is different in json.  Can we also enable the same semantics in the text parser?

vowpalwabbit/parse_args.cc
```diff
@@ -441,7 +441,10 @@ input_options parse_source(vw& all, options_i& options)
               .help(
                   "use gzip format whenever possible. If a cache file is being created, this option creates a "
                   "compressed cache file. A mixture of raw-text & compressed inputs are supported with autodetection."))
-      .add(make_option("no_stdin", all.stdin_off).help("do not default to reading from stdin"));
+      .add(make_option("no_stdin", all.stdin_off).help("do not default to reading from stdin"))
+      .add(make_option("enable_json_chain_hash", parsed_options.enable_json_chain_hash)
```
lokitoth: [Minor][Nit] Consider using a shorter argument key (it makes it easier to use on the command-line, if needed), for example: "chain_hash_json".

vowpalwabbit/parse_example.cc
```diff
@@ -148,20 +170,63 @@ class TC_parser
...
       VW::string_view feature_name = read_name();
-      _v = _cur_channel_v * featureValue();
+      VW::string_view string_feature_value;
+
+      float float_feature_value = 0.f;
+      bool is_feature_float = isFeatureValueFloat(float_feature_value);
+
+      if (_chain_hash)
+      {
+        if (is_feature_float)
```
lokitoth: [Minor] Consider unifying conditional into the root if, to avoid duplicating `_v = _cur_channel_v * float_feature_value;`.

vowpalwabbit/global_data.h
```diff
@@ -359,6 +359,8 @@ struct vw
   AllReduceType all_reduce_type;
   AllReduce* all_reduce;
 
+  bool chain_hash = false;
```
lokitoth: [Minor] Since this seems to only be consumed by the parser, shoudl this live on the `parser* p` object above (like p->hasher), see parser_args.cc:670 in your branch.

PR #2218, opened 2020-01-10
Align prediction and label enum types

jackgerrits:
This is the first change in a series for my prediction and label refactor.

The main change here is the move to `enum class` over direct `enum`.


vowpalwabbit/global_data.h
```diff
@@ -334,10 +334,8 @@ enum AllReduceType
 
 class AllReduce;
 
-// avoid name clash
-namespace label_type
-{
-enum label_type_t
+
+enum class label_type_t
```
peterychang: nitpick: we should specify the underlying type so we know the explicit size

python/pylibvw.cc
```diff
@@ -136,15 +137,15 @@ size_t my_get_label_type(vw*all)
...
-  { case prediction_type::scalar:          return pSCALAR;
-    case prediction_type::scalars:         return pSCALARS;
-    case prediction_type::action_scores:   return pACTION_SCORES;
-    case prediction_type::action_probs:    return pACTION_PROBS;
-    case prediction_type::multiclass:      return pMULTICLASS;
-    case prediction_type::multilabels:     return pMULTILABELS;
-    case prediction_type::prob:            return pPROB;
-    case prediction_type::multiclassprobs: return pMULTICLASSPROBS;
-    case prediction_type::decision_probs:  return pDECISION_SCORES;
+  { case prediction_type_t::scalar:          return pSCALAR;
```
peterychang: We can/should expose these through enums in the bindings (boost_python and pybind both have the concept of python enums). Since thats an interface breaking change though, probably add it as a secondary method and deprecate this one

peterychang:
LGTM

python/pylibvw.cc
jackgerrits: I agree, that would be a better way of exposing

PR #2224, opened 2020-01-16
winsdk update

ataymano:
WinSdk version upgrade (to VS 2019 minimal one)


cs/cli/vw_clr.vcxproj
```diff
@@ -48,7 +48,7 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
   <Import Project="$(ProjectDir)\..\..\vowpalwabbit\Build.props" />
   <PropertyGroup Label="Globals">
-    <WindowsTargetPlatformVersion>10.0.10240.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.16299.0</WindowsTargetPlatformVersion>
```
lokitoth: Please make sure to set these before importing Microsoft.Cpp[.Default].props (also on output line 33), otherwise it will (for a few operations) default to 8.1 and can fail in the case where 8.1 is not installed on the system.

jackgerrits:
Could you also update the version in CMakeLists.txt?
https://github.com/VowpalWabbit/vowpal_wabbit/blob/8d6f103f15dc70537e1691cda24a4ef5cd0e53f6/CMakeLists.txt#L20

PR #2227, opened 2020-01-17
Save and restore test_only on examples

jackgerrits:
When reusing examples in library mode - if an example is passed to predict and then subsequently passed to learn then test_label should equal what was originally set in setup_example.

This bug was seen while using the C# bindings to initially predict with then add a label and learn.


jackgerrits:
The `predict_modifying_state` test shows this is not the correct fix. In fact, I don't think we can make this change at all. `finish_example` need to know whether this example is test only or not. But only predict can know whether to force this or not. Based on this `setup_example` must be called again when reusing an example after using it.

PR #2244, opened 2020-02-03
[Draft] Refactor example's label and prediction unions

jackgerrits:
## I am exploring a new approach, see #2275 for the first change 

- Make polyprediction and polylabel typesafe, RAII friendly types. With destructors! 
    - To access fields it changes from a field to function call 
        - In release mode this is essentially no different due to conditionally compiled checks and inlining
    - Convert, audit and test every reduction in VW to conform to the new syntax 
        - This is one of the big wins. Types in the label and prediction are now very clear, explicit and intentional. This paves the way for the next big thing in atomization, changing the signature of predict to return a prediction and learn to accept a label as input
        - This process uncovered several bugs where we are interpreting the union as one type when it certainly contains a different type
- Learners now explicitly set their required label type and prediction type and debug assertions verify examples are well formed coming in and out of reduction. Compiled out of release builds for performance
- Since polylabel and polyprediction are able to destroy their contained types, I was able to turn almost every type in VW to have scoped destruction, allow moves and copies. Including: 
    - v_array
    - example
    - features
    - All labels and prediction types
- Because of the tagged union, you can delete an example without having to specify what’s in it. Therefore, delete_prediction and delete_label are deprecated and no longer needed
- Several reductions migrated to RAII types
- Chance of leaking is now reduced greatly
- Significantly reduced risk of example reuse bugs

### Deprecations
- vw.h items:
    - copy_example_label
    - free_flatten_example
    - dealloc_example
    - alloc_example
    - copy_example_data
    - copy_example_label
- v_array items:
    - delete_v
    - v_init
    - pop
    - v_array_contains
- label_parser items:
    - delete_label
    - copy_label
    - label_size
- global_data items:
    - delete_prediction

### Initial performance
- ~3% reduction in cb workflow
- ~1% reduction in RunTests script

### Points of interest
- label.h
- prediction.h
- v_array.h

This change exposed this bug #2245

To Do:
- This change introduces issues with the c api - as it always tries to retrieve the scalar prediction


jackgerrits:
Note: I am rearchitecting this PR due to perf issues. Do review until it's ready.

PR #2247, opened 2020-02-04
[PREVIEW]Add framework for managing new/proposed features data structures

peterychang:


vowpalwabbit/features_data.h
```diff
@@ -0,0 +1,147 @@
...
+ *   Create an instance of a data structure:
+ *     features_data fd;
+ *     fd.create<data_type>(constructor_parameters);
+ *
+ *   Accessing an instance of existing data:
+ *     auto data = fd.get<data_type>();
+ */
+
+/////////////////////////////////////////////
+// sample placeholder structs. For demonstrative purposes only
```
peterychang: These placeholders will need to be removed before this goes out for final review. They're here to demonstrate how the TYPE_MAPPINGS macro will need to look

vowpalwabbit/features_data.h
```diff
@@ -0,0 +1,147 @@
...
+  void clear();
+};
+/////////////////////////////////////////////
+
+/*
+ * Escape the last line and add new data structure->enum mappings to the end of this list
+ */
+#define TYPE_MAPPINGS        \
+  MAP(basic_features, basic) \
+  MAP(ccb_features, ccb)
```
peterychang: Contributors will need to escape the last line before adding a new one (otherwise you get compiler warnings). Failing to do so will result in (somewhat cryptic) compiler errors.

Is that acceptable?

vowpalwabbit/features_data.h
```diff
@@ -0,0 +1,147 @@
...
+// Escape the last line and add new mappings to the end of this list
+
+
+////////////////////////////////////////////////////////////////////
+// ----------------- DO NOT MODIFY BELOW THIS LINE -----------------
```
peterychang: We could theoretically stick all this into another header, but it'll look a little weird.. basically the header structure would become

```
#include <std_stuff>

#define TYPE_MAPPINGS        \
  MAP(basic_features, basic) \
  MAP(ccb_features, ccb)

#include <actual_features_data.h>

#undef TYPE_MAPPINGS
```

vowpalwabbit/features_data.h
```diff
@@ -0,0 +1,147 @@
...
+template <typename E>
+constexpr typename std::underlying_type<E>::type to_underlying(E e) noexcept {
+    return static_cast<typename std::underlying_type<E>::type>(e);
+}
+
+// Used to check the existence of a method called `clear()`
+template <typename T>
+struct has_clear {
+private:
+  typedef char YesType[1];
```
jackgerrits: nit: prefer using

vowpalwabbit/features_data.h
```diff
@@ -0,0 +1,147 @@
...
+ typename std::enable_if<has_clear<T>::value, T>::type*
+  create(Args&&... args) {
+   T* ret = get<T>();
+   if(ret) {
+     ret->clear();
+     new (ret) T(std::forward<Args>(args)...);
+   }
+   else
+   {
+     ret = (T*)(_data[type_traits<T>::index] = std::make_shared<T>(std::forward<Args>(args)...)).get();
```
jackgerrits: I'd prefer if the assignment to `_data` and then the assignment of the created value to `ret` were different statements.

vowpalwabbit/features_data.h
peterychang: Another problem we could have with this setup is if `actual_features_data.h` got included anywhere else, this all falls apart... maybe... depending on the order the includes are processed in

peterychang:
closed in favor of https://github.com/VowpalWabbit/vowpal_wabbit/pull/2282

PR #2248, opened 2020-02-05
Support finding system version of RapidJSON

jackgerrits:
- RapidJSON can now consumed as a system dependency
- This is critical for supporting package manager scenarios
- Ultimately, I think we should remove the RapidJSON submodule. It really isn't needed anymore, as CMake makes this much easier than the old make system.


kumpera:
Whether we remove the submodule or not. We must support building with specific versions of RapidJSON so we can have reproducible builds.

jackgerrits:
Yep removing or not is neither here nor there. But at least now different versions and package managers can be supported.

CMakeLists.txt
```diff
@@ -174,44 +175,48 @@ find_package(ZLIB REQUIRED)
...
-  option(GIT_SUBMODULE "Check submodules during build" ON)
-  if(GIT_SUBMODULE)
-    message(STATUS "Submodule update")
-    execute_process(COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive
-                    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-                    RESULT_VARIABLE GIT_SUBMOD_RESULT)
-    if(NOT GIT_SUBMOD_RESULT EQUAL "0")
-      message(FATAL_ERROR "git submodule update --init failed with ${GIT_SUBMOD_RESULT}, please checkout submodules")
+if(RAPIDJSON_SYS_DEP)
+  find_package(RapidJSON 1.1.0 CONFIG REQUIRED)
```
lokitoth: What happens if the version is not 1.1.0? Can we soften the requirement to be 1.x? (Or does RapidJSON not use SemVer?)

jackgerrits: Since `EXACT` isn't supplied, it means that any version compatible with this is valid, based on RapidJSONs version configuration. Based on their configuration anything equal to or greater than is compatible.

However, RapidJSON hasn't released in years so this is very unlikely to be an issue in this case.

PR #2250, opened 2020-02-06
Fix segfault when reading from cache, using interactions and audit

### LINKED ISSUE #2115: Seg fault while running certain tests with --audit
When running
```
./vw --onethread -c -k -d test/train-sets/library_train -f test/models/library_train.w -q st --passes 100 --hash all --noconstant --csoaa_ldf m --holdout_off --audit
```
VW seg faults on line 143 because first.space_names is empty
```
#1  0x00000000080b2bc3 in INTERACTIONS::generate_interactions<GD::audit_results, unsigned long const, &GD::audit_feature, true, &GD::audit_interaction, dense_parameters> (
    interactions=std::vector of length 1, capacity 1 = {...}, permutations=false, ec=..., dat=..., weights=...) at /mnt/c/git/vowpal_wabbit/vowpalwabbit/interactions_predict.h:143
138
139               for (size_t i = 0; i < first.indicies.size(); ++i)
140               {
141                 feature_index halfhash = FNV_prime * (uint64_t)first.indicies[i];
142                 if (audit)
143                   audit_func(dat, first.space_names[i].get());
144                 // next index differs for permutations and simple combinations
145                 feature_value ft_value = first.values[i];
146                 features::features_value_index_audit_range range = second.values_indices_audit();
147                 features::iterator_all begin = range.begin();
(gdb) print first
$1 = (features &) @0x7ffffd980d48: {values = {_begin = 0x89124c0, _end = 0x89124cc, end_array = 0x89124cc, erase_count = 0}, indicies = {_begin = 0x89124e0, _end = 0x89124f8, end_array = 0x89124f8, erase_count = 0},
  space_names = {_begin = 0x0, _end = 0x0, end_array = 0x0, erase_count = 0}, sum_feat_sq = 3}
```

jackgerrits:
Fixes #2115


vowpalwabbit/feature_group.h
```diff
@@ -194,17 +194,23 @@ class features_value_index_audit_iterator : public features_value_index_iterator
   features_value_index_audit_iterator& operator++()
   {
     features_value_index_iterator::operator++();
-    _begin_audit++;
+    if (_begin_audit != nullptr)
```
kumpera: Can't we make sure we never construct an iterator with a null _begin_audit?

jackgerrits: The problem is that when you are using a cache file you don't really want to deal with audit strings and definitely care about speed. We could potentially template that those functions to allow that, but that is a larger chunk of work. In that case we would most likely just be inserting the same pairs of "","".

PR #2259, opened 2020-02-11
Multiline parsing in python

### LINKED ISSUE #2026: Unintuitive Python parse API
As as a user I would expect all of the following calls to behave indentically. Only the first two calls works, the other two are incorrect. All of these should be able to parse a multiline example.

```
ex = vw.parse("| a:1 b:0.5\n0:0.1:0.75 | a:0.5 b:1 c:2")
print(ex)

ex = vw.parse("""| a:1 b:0.5
0:0.1:0.75 | a:0.5 b:1 c:2""")
print(ex)

ex = vw.parse("""
| a:1 b:0.5
0:0.1:0.75 | a:0.5 b:1 c:2
""")
print(ex)

ex = vw.parse(["| a:1 b:0.5", "0:0.1:0.75 | a:0.5 b:1 c:2"])
print(ex)
```

```
[<vowpalwabbit.pyvw.example object at 0x7f2fc4f0cea0>, <vowpalwabbit.pyvw.example object at 0x7f2fc4f0cf10>]
[<vowpalwabbit.pyvw.example object at 0x7f2fc4f0cf10>, <vowpalwabbit.pyvw.example object at 0x7f2fc4f0ce68>]
[<vowpalwabbit.pyvw.example object at 0x7f2fc4f14810>, <vowpalwabbit.pyvw.example object at 0x7f2fc4f1a538>, <vowpalwabbit.pyvw.example object at 0x7f2fc4f37068>, <vowpalwabbit.pyvw.example object at 0x7f2fc4f370a0>]
Segmentation fault (core dumped)
```

Smit-create:
Fixes #2026 
#### Changes made
* python
  * Fixes to parse multiple line input


jackgerrits:
Thanks for having a look at the issue!

The fix seems incorrect to me though. With the code you've proposed, a given multiline example would actually be collapsed from N lines to 1 which shouldn't happen. A multiline example is a concept in VW which allows for examples with more rich structure to be expressed.

There are several ways a user could conceivably pass examples to the parse function. The issue highlights the differences between them and ideally they would all provide the same result.

If you try the example in the issue with the code you've proposed you'll notice the results differ.

Smit-create:
Okay, thanks for reviewing. I will try to fix it trying some other way

Smit-create:
```
diff --git a/python/vowpalwabbit/pyvw.py b/python/vowpalwabbit/pyvw.py
index e750fd0f..57653c45 100644
--- a/python/vowpalwabbit/pyvw.py
+++ b/python/vowpalwabbit/pyvw.py
@@ -114,12 +114,13 @@ class vw(pylibvw.vw):
     def parse(self, str_ex, labelType=pylibvw.vw.lDefault):
         """Returns a collection of examples for a multiline example learner or a single
         example for a single example learner."""
+        str_ex = "\n".join(str_ex) if not isinstance(str_ex, str) else str_ex
         str_ex = str_ex.replace('\r', '')
         ec = self._parse(str_ex)
         ec = [example(self, x, labelType) for x in ec]
         for ex in ec:
             ex.setup_done = True

-        if not self._is_multiline():
+        if not ('\n' in str_ex):
             if len(ec) == 1:
                 ec = ec[0]
             else:
```
With the above diff, I get the expected output for all the 4 cases.
I think the problem was with `self._is_multiline()` which returned False in the some cases though the input was multiline.

jackgerrits:
`self._is_multiline()` returns true if the learner is a multiline learner. I think that part is important to keep.

Smit-create:
I ran it with `print(self._is_multiline())` and returns `False` in all of four cases locally

jackgerrits:
How did you create the `vw` object?

Smit-create:
> How did you create the `vw` object?

`vw = pyvw.vw(quiet=True)`

jackgerrits:
So that's not a multiline learner, it's a singleline learner. It's a little bit tricky to know when things a multiline/singleline, you pretty much need to look at the c++ code and see if it uses a single example or `multi_ex`. Anyway, if you initialize vw with `cb_adf=True` then you'll get a multiline learner.

Smit-create:
```
diff --git a/python/vowpalwabbit/pyvw.py b/python/vowpalwabbit/pyvw.py
index e750fd0f..57653c45 100644
--- a/python/vowpalwabbit/pyvw.py
+++ b/python/vowpalwabbit/pyvw.py
@@ -114,12 +114,13 @@ class vw(pylibvw.vw):
     def parse(self, str_ex, labelType=pylibvw.vw.lDefault):
         """Returns a collection of examples for a multiline example learner or a single
         example for a single example learner."""
+        str_ex = "\n".join(str_ex) if not isinstance(str_ex, str) else str_ex
         str_ex = str_ex.replace('\r', '')
         ec = self._parse(str_ex)
         ec = [example(self, x, labelType) for x in ec]
         for ex in ec:
             ex.setup_done = True

        if not self._is_multiline():
             if len(ec) == 1:
                 ec = ec[0]
             else:
```
Running above diff:
```
vw = pyvw.vw(quiet=True, cb_adf=True)
ex = vw.parse("| a:1 b:0.5\n0:0.1:0.75 | a:0.5 b:1 c:2")
print(ex)

ex = vw.parse("""| a:1 b:0.5
0:0.1:0.75 | a:0.5 b:1 c:2""")
print(ex)

ex = vw.parse("""
| a:1 b:0.5
0:0.1:0.75 | a:0.5 b:1 c:2
""")
print(ex)

ex = vw.parse(["| a:1 b:0.5", "0:0.1:0.75 | a:0.5 b:1 c:2"])
print(ex)
```
Output:
```
[<vowpalwabbit.pyvw.example object at 0x7f39fb6eccb0>, <vowpalwabbit.pyvw.example object at 0x7f39fad4b4b0>]
[<vowpalwabbit.pyvw.example object at 0x7f39fad4b530>, <vowpalwabbit.pyvw.example object at 0x7f39fad4b970>]
[<vowpalwabbit.pyvw.example object at 0x7f39fad4b9b0>, <vowpalwabbit.pyvw.example object at 0x7f39fad4bf30>]
[<vowpalwabbit.pyvw.example object at 0x7f39fad4bf70>, <vowpalwabbit.pyvw.example object at 0x7f39fad51470>]
```

jackgerrits:
Rather than checking if it is not a string, checking if it is a list will be safer way to achieve this behavior. Then you'll be able to throw a TypeError if an unsupported type is passed in.

Smit-create:
Okay, I will make that change and commit the changes.

jackgerrits:
Could you throw a TypeError for unsupported types?

Smit-create:
Yes, I will add that too.

jackgerrits:
Could you please also add a test for this? Here is the file: https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/tests/test_pyvw.py

Smit-create:
Yes, I will add it.

Smit-create:
@jackgerrits  Can we run the tests for a particular file locally?

jackgerrits:
To run the tests locally you need to install the edited package locally:
```sh
python3 setup.py install --user
python3 -m pytest ./python/tests/
```

python/vowpalwabbit/pyvw.py
```diff
@@ -114,7 +114,12 @@ def format_input(key, val):
     def parse(self, str_ex, labelType=pylibvw.vw.lDefault):
         """Returns a collection of examples for a multiline example learner or a single
         example for a single example learner."""
+        if not (isinstance(str_ex, list) or isinstance(str_ex, str)):
+            raise TypeError('Unsupported type. List or string object must be passed.')
+        if isinstance(str_ex, list):
+            str_ex = "\n".join(str_ex)
         str_ex = str_ex.replace('\r', '')
+        str_ex = str_ex.strip()
```
jackgerrits: Whats the reasoning behind needing strip?

Smit-create: I have used it because while running the tests and observing the error, I found that the error was due to string ending with '\n' as in 3rd case. So I used strip to remove trailing spaces and '\n'. I think it does effect the results of other formats because it only removes trailing spaces in a string.

Smit-create: I gone through the tutorial of `--cd explore_adf` which stated that each of the example starts with as new line. Here if the string ends with '\n' , then it would consider that another example has been encountered and number of examples in 3rd case will rise to 3.

jackgerrits: Yep that makes sense, thanks!

Smit-create:
Thanks @jackgerrits

PR #2261, opened 2020-02-12
Run Linux build in both release and debug mode

jackgerrits:
Currently, CI only runs release builds. There are a number of checks that only run in debug configuration. This change simply adds a matrix for the linux build enabling these checks to at least run somewhere on PRs. Ideally the CI would do this for each platform.


jackgerrits:
Reviewed by @peterychang in person

PR #2270, opened 2020-02-18
Fixes test failure in test_pyvw.py

Smit-create:
* python
  * Fixes test failure in test_pyvw.py when ran locally

### DIFF
--- a/python/tests/test_pyvw.py
+++ b/python/tests/test_pyvw.py
@@ -153,7 +153,7 @@ def test_regressor_args():
     del new_model
 
     # clean up
-    os.remove('{}.cache'.format(data_file))
+    os.remove('{}'.format(data_file))
     os.remove('tmp.model')
 
 


jackgerrits:
I'm not sure what your test failure was locally, but deleting the data file won't fix it. You'll notice that CI is failing for this PR. Perhaps you need to check if the file exists before trying to delete it?

Smit-create:
I had following test failure before the change I commited:
```
===================================================== FAILURES =====================================================
_______________________________________________ test_regressor_args ________________________________________________

    def test_regressor_args():
        # load and parse external data file
        data_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'resources', 'train.dat')
        model = vw(oaa=3, data=data_file, passes=30, c=True, k=True)
        assert model.predict('| feature1:2.5') == 1

        # update model in memory
        for _ in range(10):
            model.learn('3 | feature1:2.5')

        assert model.predict('| feature1:2.5') == 3

        # save model
        model.save('tmp.model')
        del model

        # load initial regressor and confirm updated prediction
        new_model = vw(i='tmp.model', quiet=True)
        assert new_model.predict('| feature1:2.5') == 3
        del new_model

        # clean up
>       os.remove('{}.cache'.format(data_file))
E       FileNotFoundError: [Errno 2] No such file or directory: '/home/smit/Smit Lunagariya/vowpal_wabbit/python/tes
ts/resources/train.dat.cache'
```

Smit-create:
> Perhaps you need to check if the file exists before trying to delete it

It exists in my branch

jackgerrits:
File existing as in on disk and not the file being checked into the repo. Deleting the data file is not the right thing to do here.

Smit-create:
Okay got it. Should I close it then?

jackgerrits:
I think either close it or add an existence check

PR #2275, opened 2020-02-20
Change Polylabel to a struct of objects instead of union

jackgerrits:
This is the new approach towards refactoring label. The idea is that we simply should not use a union. Examples are around 26kB, the different of moving from a union to a flat struct is about 150 bytes, which is less than 1% difference in object size. But by removing the union we remove all of the issues we've had.

This PR is the smallest unit of work to move in this direction. This approach allows incremental work without perf loss. There is actually significant perf to be gained through this new approach as we can stop overwriting labels, we can actually maintain all of the labels for the callstack in parallel.

- All label parser functions are strongly typed
- Label parsers are defined with lambdas that forward the call


rajan-chari:
1) Can the label size increase cause it to not fit in a single cache line?  What was the size of the union before?

2) Labels will still need to be rewritten and saved in 2 cases:
   a) two reductions using the same label type but with different values.  for example this happens with simple label
   b) reductions with multiple learners will change labels with different learners.

For a reduction to be truly safe, if there are label modifications, the old label has to be saved and restored.  Either that or reductions at the top of the stack cannot assume that lower reductions won't modify labels.

For example:

Assume there are 2 reductions M and N where M reduces to N.  There are other reductions above M and below N.

Reduction N uses label 'n' which is set by reduction M.  

In order for M to work safely in all possible reduction stacks, when M writes to label 'n', it must save and restore 'n'.  

The only way I can think to make this a little better is to say every reduction that needs a label gets its own label.  Then you can atleast say that no other reduction would have touched that label.  But this comes with a host of it's own problems.

Even in this case, if multiple learners are involved (a tree for example), we might have to save and restore labels.  

In final analysis, I think this is a positive change because unions are harder to work with for many reasons.  

But this might not have all the benefits that you are hoping for.  In pariticular, save/restore label is not something this solves unless we make a ton of assumptions.  Instead of saving label state to a typeless construct (union), you can save/restore state of a single sub-component. 'n' in the example above.

I wonder if the missing "feature" is encapsulation around saving/restoring labels ....

rajan-chari:
One other way is to pass label as a parameter.  i.e. learn(simple_label&, example&).  Some changes would have to be made because base learners erase type and of course trivially this means function signatures need to change everywhere.

jackgerrits:
> Can the label size increase cause it to not fit in a single cache line? What was the size of the union before?

Perhaps, but I don't think this is really the issue because it was already quite large. Here are the sizes:
```
Example: 27152 bytes
Label struct: 232 bytes
Label union: 56 bytes
```
From testing this PR I did not notice any perf hits, there were some perf gains though. And more perf gains to be had when we work out when we can avoid saving and restoring labels. CCB is a prime example where this will improve perf.

> Labels will still need to be rewritten and saved in 2 cases:
a) two reductions using the same label type but with different values. for example this happens with simple label
b) reductions with multiple learners will change labels with different learners.

Yep, 100% this does not remove the need to save and restore labels. It does reduce/remove the need in lots of places. It isn't clear when you can and cant do that though as you pointed out.

This change does in fact make things safe in the sense of memory safety and leaks because when you replace the label the old one is destroyed. But it does _not_ provide statistical safety in the sense that the label may be incorrect due to some mistake and change results. 

I _really_ want to get to a place where each reduction gets a reference to just the label it needs. `learn(simple_label&, example&)`. This is extremely hard to pull off like you said, so this is the first step in us getting there. This PR is definitely not the final destination, I'm hoping to do things incrementally where possible (that other PR was too big of a change). Encapsulating saving and restoring of labels was primarily what the other PR was about, and the perf cost of that in that attempt was prohibitive (~5% - ~15%)

vowpalwabbit/label_parser.h
```diff
@@ -12,19 +12,20 @@
...
-  void (*delete_label)(void*);
-  float (*get_weight)(void*);
-  void (*copy_label)(void*, void*);  // copy_label(dst,src) performs a DEEP copy of src into dst (dst is allocated
+  void (*default_label)(polylabel*);
+  void (*parse_label)(parser*, shared_data*, polylabel*, v_array<VW::string_view>&);
+  void (*cache_label)(polylabel*, io_buf& cache);
+  size_t (*read_cached_label)(shared_data*, polylabel*, io_buf& cache);
+  void (*delete_label)(polylabel*);
+  float (*get_weight)(polylabel*);
+  void (*copy_label)(polylabel*, polylabel*);  // copy_label(dst,src) performs a DEEP copy of src into dst (dst is allocated
```
lalo: copying will effectively clear the label if both polylabel* end up pointing to the same thing (based on current impl of copy_array). Not sure if we should protect from this error on the api level.

vowpalwabbit/cost_sensitive.cc
```diff
@@ -147,7 +154,7 @@ void parse_label(parser* p, shared_data* sd, void* v, v_array<VW::string_view>&
         else
         {
           wclass f = {-FLT_MAX, 0, 0., 0.};
-          ld->costs.push_back(f);
+          ld.costs.push_back(f);
```
lalo: would using emplace_back save us a copy? here and in the other push_back

peterychang: emplace_back uses the input parameters as constructor params, there's no constructor for that object right now; but we should totally make that change when we can

vowpalwabbit/label_parser.h
jackgerrits: Hmm that's interesting, I think I'm going to hold off on an API fix for that because I have another branch lined up that will fix the copy behavior of v_array.

rajan-chari:
Just FYI
How big are various things in example?  This is what I found.  

example_predict:
ep.indices=32, ep.feature_space=26624, ep.ft_offset=8, ep.interactions=8, example_predict=26672

example
el.l=56, el.pred=32, el.weight=4, el.tag=32, el.example_counter=8, el.num_features=8, el.partial_prediction=4, el.updated_prediction=4, el.loss=4, el.total_sum_feat_sq=4, el.confidence=4, el.passthrough=8, el.test_only=1, el.end_pass=1, el.sorted=1, example=26856

feature_space is an array of 256 features, 104 bytes each.

---
In reply to: [590066446](https://github.com/VowpalWabbit/vowpal_wabbit/pull/2275#issuecomment-590066446) [](ancestors = 590066446)

rajan-chari:
Apologies.... these comments are long.

Here are are tradeoffs as I see it.

1) The labels in this PR (232 bytes)  compared to before (56 bytes) will not fit in a cache line (mostly 64 bytes).  It's true that features won't fit in cache lines.  However, the access pattern for features and labels are very different.  Features are accessed efficiently in gd for dot_product type operation.  Labels are accessed in many reductions (but not all).  
2) Construction/Destruction of labels now involve constructors/destructors for each type… no?  This might be the canonical approach for composition but not really how labels are used in vw.  Also only an issue when label is constructed, when example is constructed or when they are passed by value or saved locally.

3) It used to be that any time you modify a label in a reduction you have to save it and restore it after calling base .    This is still the case from what I can see.  For example, if you modify polylabel.cb, you have to save this label and restore it after base operations are done.   Or else you risk affecting the label for  upstream reductions.  

4) Not clear to me how this helps with moving to something like learn(cb_label, features).  All the work there is in the template machinery for calling learn/predict and changing a lot of function definitions.  From my perspective (and I think we are on the same page on this) this is the better change and one that we want to move towards.

I tried to look for changes that reduces save/restore.  I did not find it.  Is that part of another PR?


---
In reply to: [590964697](https://github.com/VowpalWabbit/vowpal_wabbit/pull/2275#issuecomment-590964697) [](ancestors = 590964697,590066446)

jackgerrits:
1. Yes I agree, this is a loss. However, given we are already committed to splitting out label types to be templated as a part of `learn`, then this is a temporary/short term negative. Given this is a step in the plan, I think this tradeoff is acceptable. Thoughts?

2. Labels are only constructed or destructed at the beginning and end of the life of the example. At all places in between they are simply reused. I would not say this deviates from the current usage pattern, we have just made the resource management less error prone. Examples are rarely copied and we should maintain that. But when they are copied it should behave as expected without pitfalls.

3. You may need to save and restore the label if reductions below you may use the same type. However, if you know that is not the case you don't need to save/restore. This is the case for CCB and slates, and gets big wins because of it. Again, this case will not be an issue when we get a templated split out label.

4. Yep, totally agree. The reason this gets us closer is that it makes all of the labels proper C++ types with constructors and destructors that properly utilize move and copy semantics. 

Updates to save and restore are not in this PR, they are sensitive and I want to do them seperately. I am trying to split up this work to as small chunks as I can, as having a huge PR like I did before is very hard to get effective review of. I think it's important to note that more work will follow this PR.

rajan-chari:
I think we landed at a good compromise.  I feel ok with this change since we are prioritizing the stack based label change which can be achieved incrementally following Rodrigo's approach.

---
In reply to: [592042555](https://github.com/VowpalWabbit/vowpal_wabbit/pull/2275#issuecomment-592042555) [](ancestors = 592042555)

lokitoth:
@rajan-chari/@jackgerrits: I am a bit confused - are we going with #2314 in place of this, or are they complementary?

jackgerrits:
They are complimentary, but #2314 will go in first then I will need to do some work to refactor this around it. That other PR is a condition on us doing this one.

test/unit_test/slates_parser_test.cc
```diff
@@ -179,44 +176,42 @@ BOOST_AUTO_TEST_CASE(slates_cache_slot_label)
 
   parser p{8 /*ring_size*/, false /*strict parse*/};
 
-  auto lp = VW::slates::slates_label_parser;
   auto label = scoped_calloc_or_throw<VW::slates::label>();
-  parse_label(lp, &p, "slates slot 0:0.5,1:0.25,2:0.25", *label.get());
-  lp.cache_label(label.get(), io_writer);
+  parse_slates_label(&p, "slates slot 0:0.5,1:0.25,2:0.25", *label.get());
```
lalo, line 180: does a unit test where we init a label but verify the rest of the fields make sense? what is the expectation of the state of the rest of the labels? (just started reading cr, so I might find out as I read)

jackgerrits: The label is memset to 0 during construction, and because of how it worked in the past that is a valid initial state.

test/unit_test/json_parser_test.cc
```diff
@@ -289,12 +289,13 @@ BOOST_AUTO_TEST_CASE(parse_json_cb_as_ccb)
...
   BOOST_CHECK_EQUAL(examples[4]->l.conditional_contextual_bandit.type, CCB::example_type::slot);
 
-  auto label1 = examples[4]->l.conditional_contextual_bandit;
+  auto& label1 = examples[4]->l.conditional_contextual_bandit;
   BOOST_CHECK_EQUAL(label1.explicit_included_actions.size(), 0);
   BOOST_CHECK_CLOSE(label1.outcome->cost, 1.f, .0001f);
   BOOST_CHECK_EQUAL(label1.outcome->probabilities.size(), 1);
   BOOST_CHECK_EQUAL(label1.outcome->probabilities[0].action, 0);
   BOOST_CHECK_CLOSE(label1.outcome->probabilities[0].score, .5f, .0001f);
+  // vw->finish_example(examples);
```
olgavrou, line 298: nit: remove commented out code

vowpalwabbit/cb.cc
```diff
@@ -9,18 +9,17 @@
...
 
 namespace CB
 {
-void parse_label(parser* p, shared_data*, void* v, std::vector<VW::string_view>& words, reduction_features&)
+void parse_label(parser* p, shared_data*, CB::label& ld, std::vector<VW::string_view>& words, reduction_features&)
 {
-  CB::label* ld = (CB::label*)v;
-  ld->costs.clear();
-  ld->weight = 1.0;
+  ld.weight = 1.0;
```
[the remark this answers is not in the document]
jackgerrits: I *think* default is called before parse, so this will already be cleared by the time we get here.

vowpalwabbit/multilabel.cc
```diff
@@ -5,13 +5,14 @@
 #include <cfloat>
 #include "gd.h"
 #include "vw.h"
+#include "example.h"
 
 namespace MULTILABEL
 {
```
lalo, line 11: do we care about the lack of testing reported in this file?

vowpalwabbit/cb_algs.cc
```diff
@@ -60,7 +61,10 @@ void predict_or_learn(cb& data, single_learner& base, example& ec)
     ec.l.cs = data.cb_cs_ld;
 
     // Guard example state restore against throws
-    auto restore_guard = VW::scope_exit([&ld, &ec] { ec.l.cb = ld; });
+    auto restore_guard = VW::scope_exit([&ld, &ec] {
+      ec.l.cs.costs = v_init<COST_SENSITIVE::wclass>();
```
lalo, line 65: why do we need to clear the costs in case of a throw?

jackgerrits: This isn't in case of a throw, this is something that must happen every time this function exits. If we don't reset it, it results in a double free.

vowpalwabbit/multilabel.cc
jackgerrits: We do, but I don't think we should tackle it in this pr.

vowpalwabbit/cb_explore.cc
```diff
@@ -235,6 +236,7 @@ void predict_or_learn_cover(cb_explore& data, single_learner& base, example& ec)
     }
   }
 
+  ec.l.cs.costs = v_init<COST_SENSITIVE::wclass>();
```
lalo, line 239: so basically you have to clear the costs of any label that was temporarily used by a reduction? if so you think its possible that we're missing this clear on other reductions? or how would you explain the need for this line

vowpalwabbit/cb_algs.cc
lalo: ah, right. keep getting misled by that comment and forget it's a scope guard at first

vowpalwabbit/cb_explore.cc
jackgerrits: There are some reductions which do a by value copy of the entire label and since v_arrays shallow copy currently this can result in double frees.

So you only need to clear if something else made a shallow copy of it. In future (as soon as this is merged) we will use move/copy to make this more sane.

lalo: is this also related to the double free with nn?

jackgerrits: No this was a different one. Caused by the fact that we reuse the label in the reduction data, hence the need to clear it so that at tear down there is only one reference left.

It boils down to at destruction time either only the example or only the reduction data needs the reference and not both

lalo (approved):

vowpalwabbit/cb_algs.cc
olgavrou: should this not be `ec.l.cs.costs.clear()`?

jackgerrits: Clear simply removes all the elements but still refers to the buffer. We actually want to completely remove the reference to the buffer here.

olgavrou: I am guessing the response is the same as: https://github.com/VowpalWabbit/vowpal_wabbit/pull/2275/files#r554070465

jackgerrits: Yep :)

vowpalwabbit/cb_explore.cc
olgavrou: so the reduction calls `delete_v` on the array?

olgavrou (approved):

vowpalwabbit/cb_explore.cc
jackgerrits: Yep the example will call delete_v on the array in the delete unions function

Issue #2268, opened 2020-02-14
Increase coverage of pyvw.py


Smit-create:
#### Describe the bug
The code coverage report of `pyvw.py` is:
```
----------- coverage: platform linux, python 3.7.4-final-0 -----------
Name                                                                                                           Stmts   Miss  Cover
----------------------------------------------------------------------------------------------------------------------------------
/home/smit/.local/lib/python3.7/site-packages/vowpalwabbit-8.8.0-py3.7-linux-x86_64.egg/vowpalwabbit/pyvw.py     417    227    46%

```
An html code coverage report can be generated locally with:
`/vowpal_wabbit/python$ pytest --cov=vowpalwabbit.pyvw --cov-report=html tests/test_pyvw.py`

Below is the code coverage report of `sklearn_vw.py` which looks good:
```
----------- coverage: platform linux, python 3.7.4-final-0 -----------
Name                                                                                                                 Stmts   Miss  Cover
----------------------------------------------------------------------------------------------------------------------------------------
/home/smit/.local/lib/python3.7/site-packages/vowpalwabbit-8.8.0-py3.7-linux-x86_64.egg/vowpalwabbit/sklearn_vw.py     163     20    88%

```
Does increasing the code coverage report seems to be a good idea to work upon?

jackgerrits:
Adding tests is always welcome! 😄

Smit-create:
Sure! will open a PR for that to work upon.

Smit-create:
I have started working on it. But without documentation is hard to run them. In `SearchTask()`, that type of object should be `sch`?

jackgerrits:
The search objects are complex, I would suggest you skip over them for now. If you would like to learn more have a look at these resources:
- https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Learning-to-Search-Sub-System
- https://hunch.net/~l2s/

Smit-create:
Okay!, will look over that

jackgerrits:
Looks like this was addressed in #2276

PR #2276, opened 2020-02-20
Added tests to increase coverage

Smit-create:
Related to #2268 
* python
   * Increase test coverage

After adding` test_numspace_id` coverage rises to `49%`
After adding` test_example_numspace` coverage rises to `55%`
After adding `test_simple_label` and `test_multiclass_label` coverage rises to `62%`.
Currently after adding `test_example` coverage rises to `70%`.


Smit-create:
I will add more tests in this PR and merge master

jackgerrits:
Hi @Smit-create,

Thanks for the PR!

This test should be asserting a few more things to test the state of the example:
1. Test for existance and value of the label
2. There are three namespaces which should be tested, 'a', 'b' and the default namespace
3. You could test number of features

python/tests/test_pyvw.py
```diff
@@ -183,3 +183,15 @@ def test_parse():
...
     del model
+
+def test_numspace_id():
+    vw_ex = vw(quiet=True)
+    ex = vw_ex.example('1 |a two features |b more features here')
+    nm1 = pyvw.namespace_id(ex, 2)
+    assert nm1.id == 2
+    assert nm1.ord_ns == 128
+    assert nm1.ns == '\x80'
+    nm2 = pyvw.namespace_id(ex, '')
```
jackgerrits: How does this obtain the default namespace? Seems odd that an empty string works here?

Smit-create:
> This test should be asserting a few more things to test the state of the example

Yes, I would add more of them upon current. Also I will add for other classes in `pyvw.py` in this PR.

python/tests/test_pyvw.py
Smit-create: Actually this too works here:
`nm2 = pyvw.namespace_id(ex, 'some_id')`
Should an empty string raise error for example passed here?

jackgerrits: Strings are valid, if you pass an empty string it is converted to a space. A space corresponds to the default namespace (32). The example you've provided doesn't included doesn't have any default namespaces.

Smit-create:
```
>>> nm2 = pyvw.namespace_id(ex, 'x')
assert nm2.id is None
assert nm2.ord_ns == 120
assert nm2.ns == 'x'

```
Adding this would be good?

Smit-create:
```
>>> nm2 = pyvw.namespace_id(ex, 'x')
assert nm2.ns == 'x'
assert nm2.ord_ns == 120
assert nm2.id is None
```
Would adding this test be good?

jackgerrits: You should read the implementation of `namespace_id` to get a better idea of what it does and how you should test it here

python/tests/test_pyvw.py
```diff
@@ -210,3 +210,33 @@ def test_example_namespace():
...
+def test_simple_label():
+    vw_ex = vw(quiet=True)
+    ex = vw_ex.example('1 |a two features |b more features here')
+    sl = pyvw.simple_label('c', weight=0.5)
+    assert sl.label == 'c'
+    assert sl.weight == 0.5
+    assert sl.prediction == 0.0
+    assert sl.initial == 0.0
+    assert str(sl) == 'c:0.5'
+    sl2 = pyvw.simple_label(ex)
```
jackgerrits: You are testing whether this is equal against itself? It will always be true.

python/tests/test_pyvw.py
```diff
@@ -210,3 +210,33 @@ def test_example_namespace():
     assert exm.pop_feature()
     exm.push_features(nmid, ['c', 'd'])
     assert exm.num_features_in() == 4
+
+def test_simple_label():
+    vw_ex = vw(quiet=True)
+    ex = vw_ex.example('1 |a two features |b more features here')
+    sl = pyvw.simple_label('c', weight=0.5)
```
jackgerrits: Passing a string as the first parameter is in fact a bug, this should only be a float

Smit-create: Okay. Then this should raise error in that case. I will change that and add a raise error if string is passed.

python/tests/test_pyvw.py
Smit-create: Can you please suggest some other way to test them? Or should I test it numerically as in `sl`

jackgerrits: Do as you did in the previous 3 lines, where you assert against known values

Smit-create:
```
class multiclass_probabilities_label(abstract_label):
    """Class for multiclass VW label with probabilities"""
    def __init__(self, label, prediction=None):
        abstract_label.__init__(self)
        if isinstance(label, example):
            self.from_example(label)
        else:
            self.prediction = prediction

    def from_example(self, ex):
->        self.prediction = ex.get_multiclass_probabilities() # should this be ex.get_prob ?
```
@jackgerrits The highlighted line raises error as no such method exits. Should it be `ex.get_prob`?

jackgerrits:
Interesting, I'm not sure what a multiclass probabilties labels corresponds to. Generally speaking the labels all line up to an entry here: https://github.com/VowpalWabbit/vowpal_wabbit/blob/7402da2c427418dc0c281391563391f368ba3af8/vowpalwabbit/example.h#L33

Smit-create:
[cov.pdf](https://github.com/VowpalWabbit/vowpal_wabbit/files/4241948/cov.pdf)
@jackgerrits Please have a look at the attached file. It contains the missing lines and a major portion of them is from `class SearchTask` and `init_search_task` method of `class vw`. 
Tests for `class multiclass_probabilities_label` are not added due to the above error I pointed. I would now work to add tests for `class example`. Currently it has 66% coverage

python/tests/test_pyvw.py
```diff
@@ -122,13 +122,19 @@ def test_multilabel_prediction_type():
 
 def test_cbandits_label():
     model = vw(cb=4, quiet=True)
-    assert pyvw.cbandits_label(model.example('1 |')).costs[0].label == 1
+    cbl = pyvw.cbandits_label(model.example('1 |'))
+    assert cbl.costs[0].label == 1
```
jackgerrits: There are 3 components to the label: cost, action and prob. Can you please test all 3?

python/tests/test_pyvw.py
```diff
@@ -122,13 +122,19 @@ def test_multilabel_prediction_type():
...
+    assert cbl.prediction == 0
+    assert str(cbl) == '1:3.4028234663852886e+38'
     del model
 
 
 def test_cost_sensitive_label():
     model = vw(csoaa=4, quiet=True)
-    assert pyvw.cost_sensitive_label(model.example('1 |')).costs[0].label == 1
+    csl = pyvw.cost_sensitive_label(model.example('1 |'))
```
jackgerrits: There are 3 components to the label: cost, action and prob. Can you please test all 3?

python/tests/test_pyvw.py
```diff
@@ -183,3 +189,60 @@ def test_parse():
...
+    assert nm2.ns == ' '
+
+def test_example_namespace():
+    vw_ex = vw(quiet=True)
+    ex = vw_ex.example('1 |a two features |b more features here')
+    nmid = pyvw.namespace_id(ex, 1)
+    exm = pyvw.example_namespace(ex, nmid)
+    assert isinstance(exm.ex, pyvw.example)
+    assert isinstance(exm.ns, pyvw.namespace_id)
+    assert exm.ns_hash is None
```
jackgerrits: Why is ns_hash None?

python/tests/test_pyvw.py
```diff
@@ -183,3 +189,60 @@ def test_parse():
...
+    assert exm.num_features_in() == 3
+    assert exm[2] == (11617, 1.0)
```
jackgerrits: What is 11617? Can you add a comment?

python/tests/test_pyvw.py
```diff
@@ -183,3 +189,60 @@ def test_parse():
...
+    assert exm.iter_features()
+    assert exm.pop_feature()
+    exm.push_features(nmid, ['c', 'd'])
+    assert exm.num_features_in() == 4
+
+def test_simple_label():
+    vw_ex = vw(quiet=True)
+    ex = vw_ex.example('1 |a two features |b more features here')
+    sl = pyvw.simple_label(2.0, weight=0.5)
```
jackgerrits: This is testing both a standalone simple label and one from an example- can you please split into two tests? Can you also do it for the other tests where this is the same?

python/vowpalwabbit/pyvw.py
```diff
@@ -441,6 +441,8 @@ def from_example(self, ex):
 class simple_label(abstract_label):
     """Class for simple VW label"""
     def __init__(self, label=0., weight=1., initial=0., prediction=0.):
+        if not (isinstance(label, float) or isinstance(label, example)):
```
jackgerrits: How much does this check hurt performance?

python/tests/test_pyvw.py
Smit-create: It takes the default argument as `None`. Should this be changed?

jackgerrits: I guess I'm asking, does None make sense here?

Smit-create: I think it should not be `None`.

python/vowpalwabbit/pyvw.py
Smit-create: Performance in terms of coverage or time?

jackgerrits: runtime perf

Smit-create: Can you please suggest some way to check it? Should it be checked with `%timeit`?

python/tests/test_pyvw.py
Smit-create: I have changed it to take the value as passed in the arguement.

python/vowpalwabbit/pyvw.py
Smit-create: With the check:
```
In [1]: %timeit pyvw.simple_label(1.0)
Out[1]: 1.6 µs ± 21.2 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
Without the check:
```
In [1]: %timeit pyvw.simple_label(1.0)
Out[1]: 1.5 µs ± 9 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

python/tests/test_pyvw.py
Smit-create: I have corrected the tests according to the docstring of namespace_id

Smit-create:
Currently:
 * Tests of class vw (specifically init_search_task method) and class SearchTask are left to be added.
 * Test of class multiclass_probabilities_label are left because of [ this comment.](https://github.com/VowpalWabbit/vowpal_wabbit/pull/2276#issuecomment-590093274)
 * Current coverage is `72%`.

Smit-create:
@jackgerrits  Please review this if this is ready to merge. I would study the parts of the tests that are left out as they are a bit complicated. The issue can be kept open to have a track of tests that are left out.

python/tests/test_pyvw.py
```diff
@@ -122,13 +123,23 @@ def test_multilabel_prediction_type():
 
 def test_cbandits_label():
     model = vw(cb=4, quiet=True)
-    assert pyvw.cbandits_label(model.example('1 |')).costs[0].label == 1
+    cbl = pyvw.cbandits_label(model.example('1 |'))
+    assert cbl.costs[0].label == 1
+    assert cbl.costs[0].probability == 0.0
+    assert cbl.costs[0].partial_prediction == 0.0
+    assert cbl.prediction == 0
+    assert str(cbl) == '1:3.4028234663852886e+38'
```
jackgerrits: Why does the produced string contain `3.4028234663852886e+38`? I don't see this value in the above assertions

python/tests/test_pyvw.py
```diff
@@ -20,6 +20,7 @@ def test_learn_predict(self):
         assert init == 0
         self.model.learn(ex)
         assert self.model.predict(ex) > init
+        assert self.model.predict({'x': "| a:1 b:0.5"}) == 0
```
jackgerrits: When using a dictionary to create an example it uses the value as the feature name, the parser is not run. Therefore this value does not make sense and is confusing to someone reading the tests

python/tests/test_pyvw.py
```diff
@@ -182,4 +193,94 @@ def test_parse():
 
     ex = model.parse(["| a:1 b:0.5", "0:0.1:0.75 | a:0.5 b:1 c:2"])
     assert len(ex) == 2
+
+    # To test the prediction on multiline examples
+    finish = model.finish_example(ex)
+    assert model.predict(ex) == [0.0, 0.0]
     del model
+
+def test_numspace_id():
```
jackgerrits:
```suggestion
def test_namespace_id():
```

python/tests/test_pyvw.py
```diff
@@ -182,4 +193,94 @@ def test_parse():
...
+    vw_ex = vw(quiet=True)
+    ex = vw_ex.example('1 |a two features |b more features here')
+    nm1 = pyvw.namespace_id(ex, 2)
+    assert nm1.id == 2
+    assert nm1.ord_ns == 128
+    assert nm1.ns == '\x80' # Represents string of ord_ns
+    nm2 = pyvw.namespace_id(ex, 'my_example')
```
jackgerrits: This does not make sense. Granted it is an unclear part of the API. You are invoking the branch which does not touch the ex object and so doesn't really make sense for this test case?

python/tests/test_pyvw.py
```diff
@@ -182,4 +193,94 @@ def test_parse():
...
+    assert nm2.id is None # It is kept None as because we don't want to do linear search required to find it
+    assert nm2.ord_ns == 109
+    assert nm2.ns == 'm' # First character is considered as namespace id in case of string
+
+def test_example_namespace():
+    vw_ex = vw(quiet=True)
+    ex = vw_ex.example('1 |a two features |b more features here')
+    nmid = pyvw.namespace_id(ex, 1)
+    exm = pyvw.example_namespace(ex, nmid, ns_hash=0)
```
jackgerrits: The hash of the namespace being 0 doesn't make sense?

jackgerrits:
I'd like you to take a step back and consider what each test should be testing, as some of them are testing not relevant things. I understand this is tricky with the current API, and so perhaps this is not a good first issue?

jackgerrits:
Didn't mean to click close :)

python/vowpalwabbit/pyvw.py
```diff
@@ -388,7 +388,7 @@ def __init__(self, ex, ns, ns_hash=None):
             raise TypeError
         self.ex = ex
         self.ns = ns
-        self.ns_hash = None
+        self.ns_hash = ns_hash
```
jackgerrits: Good catch!

python/tests/test_pyvw.py
```diff
@@ -182,4 +193,94 @@ def test_parse():
...
+def test_example():
+    vw_ex = vw(quiet=True)
+    ex = vw_ex.example('1 |a two features |b more features here')
+    ns = pyvw.namespace_id(ex, 1)
+    ns2 = pyvw.namespace_id(ex, 2)
+    assert isinstance(ex.get_ns(1), pyvw.namespace_id)
+    assert isinstance(ex[2],pyvw.example_namespace)
+    assert ex.setup_done is True
+    assert ex.num_features_in(ns) == 3
+    ex.unsetup_example() # unsetup an example as it is already setup
```
jackgerrits: Using unsetup and then learning is not a supported use case

python/vowpalwabbit/pyvw.py
lalo: this deserves more investigation, or a performance test with an e2e scenario

python/tests/test_pyvw.py
lalo: To reword Jack, this line should actually fail. So keep this test but revert the assert.

python/tests/test_pyvw.py
lalo: Let's get all tests in CI to pass, for example we're getting failures here https://dev.azure.com/vowpalwabbit/Vowpal%20Wabbit/_build/results?buildId=5684&view=logs&j=57e19422-7def-5009-e470-a832cad5a7e3&t=8e4bc9ff-39a8-50a5-9cb9-43a19053be7e

Let us know if you need help figuring out the test results.

Smit-create: The Test failure log shows this `AssertionError: assert '1:3.40282346639e+38' == '1:3.4028234663852886e+38'`. Maybe this can be because of some round-off, also I use Python3 whereas I saw that the tests run on Python 2. 
Locally all the tests pass on my computer.

python/tests/test_pyvw.py
Smit-create: Okay, for now, I will keep this for later when I would go for left out ones.

python/tests/test_pyvw.py
Smit-create: Locally when I changed according to the output of log, I get:
```
>       assert str(csl) == '1:3.40282346639e+38'
E       AssertionError: assert '1:3.4028234663852886e+38' == '1:3.40282346639e+38'
E         - 1:3.4028234663852886e+38
E         ?               ^^^^^^
E         + 1:3.40282346639e+38
E         ?               ^

```

python/tests/test_pyvw.py
Smit-create: So should it be changed to this `assert self.model.predict({"| a:1 b:0.5":'a'}) == 0`?

python/tests/test_pyvw.py
Smit-create: Okay, I would change the order of tests.

python/vowpalwabbit/pyvw.py
Smit-create: Thanks!

python/tests/test_pyvw.py
Smit-create: Can you please suggest for what value of `ns_hash` would it work?. Currently, I am unable to figure it out

jackgerrits: `ns_hash` should equal the output of the hash function from the string namespace. It's something that the parser would populate, that's why this particular object creation is hard to do. I'm not sure if this is something people would directly create?

python/tests/test_pyvw.py
jackgerrits: Comparing floats by string is error prone due to precision differences between environments which is why you are seeing these failures

python/tests/test_pyvw.py
Smit-create: Okay, for test purpose (coverage) can we use this one then `exm = pyvw.example_namespace(ex, nmid, ns_hash=nmid.ord_ns)`. As I was a similar thing implemented in `example.feature()`

python/tests/test_pyvw.py
Smit-create: Should I remove it because of the precision error? or I can do something like this by adding a comment:
```
#This test may fail locally because of different precision in some environment
assert str(csl) == '1:3.40282346639e+38' # Actual value is '1:3.4028234663852886e+38'
```
This will pass the CI tests

jackgerrits: So, I just had a closer look at the `cbandit_label` class, and it is actually incorrectly implemented. Thanks for finding this! I've created an issue here: #2295. Getting the string function to output the correct fields is critical to making sure the test is right. Would you like to work on that issue and submit a different PR for it?

Also, with this test, the label that is being parsed is missing the other two components of a CB label. Generally they come in the form: `action:cost:probability`. It seems the test was originally written just looking at the action, but since you are testing more fields you should expand the label to all 3. See [here](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Logged-Contextual-Bandit-Example) for example of these CB labels.

Also, just a side note tests must pass both locally and in CI.

Smit-create: > Would you like to work on that issue and submit a different PR for it?

Yes!, I will take it and submit a PR.

Smit-create:
@jackgerrits I have made a few changes, Please review it.

python/vowpalwabbit/pyvw.py
jackgerrits: @Smit-create have you looked into this?

Smit-create: I want some help regarding this, how to do a performance test check. I have checked it using %timeit as above and it shows that after adding that check, it takes few microseconds more  than the current master

Smit-create: @jackgerrits  So should I remove this check for now? as this PR looks fine after that. Also, I will open an issue related to [this](https://github.com/VowpalWabbit/vowpal_wabbit/pull/2276#issuecomment-590093274) to have a track of it.

jackgerrits: You could run the bindings on some dataset with and without the change and measure the difference in performanc

jackgerrits:
Now that #2346 is merged, does it affect this PR?

Smit-create:
> Now that #2346 is merged, does it affect this PR?

Yes, merging that PR increased the coverage percentage. Now with this PR, I will do the performance check and after providing the results, I will get it ready to be merged.

python/tests/test_pyvw.py
```diff
@@ -20,6 +20,7 @@ def test_learn_predict(self):
         assert init == 0
         self.model.learn(ex)
         assert self.model.predict(ex) > init
+        assert self.model.predict({'x': ['b']}) == 0
```
jackgerrits: This assertion is tangential to the test. Either remove it or add it in its own test

python/tests/test_pyvw.py
```diff
@@ -192,4 +193,95 @@ def test_parse():
 
     ex = model.parse(["| a:1 b:0.5", "0:0.1:0.75 | a:0.5 b:1 c:2"])
     assert len(ex) == 2
+
+    # To test the prediction on multiline examples
+    finish = model.finish_example(ex)
```
jackgerrits: This test is for parsing examples not predicting. Either remove or move it to a different test

python/tests/test_pyvw.py
```diff
@@ -192,4 +193,95 @@ def test_parse():
...
+    assert model.predict(ex) == [0.0, 0.0]
     del model
+
+def test_namespace_id():
+    vw_ex = vw(quiet=True)
+    ex = vw_ex.example('1 |a two features |b more features here')
+    nm1 = pyvw.namespace_id(ex, 2)
```
jackgerrits: This example has 3 namespaces. Constant, a and b. If you are testing namespaces please assert all of them.

python/tests/test_pyvw.py
```diff
@@ -192,4 +193,95 @@ def test_parse():
...
+    assert nm1.id == 2
+    assert nm1.ord_ns == 128
+    assert nm1.ns == '\x80' # Represents string of ord_ns
+
+def test_example_namespace():
+    vw_ex = vw(quiet=True)
+    ex = vw_ex.example('1 |a two features |b more features here')
+    nmid = pyvw.namespace_id(ex, 1)
+    exm = pyvw.example_namespace(ex, nmid, ns_hash=vw_ex.hash_space(nmid.ns))
```
jackgerrits: The naming of these objects is unclear, please give them more descriptive names

python/tests/test_pyvw.py
```diff
@@ -192,4 +193,95 @@ def test_parse():
...
+    assert isinstance(exm.ex, pyvw.example)
+    assert isinstance(exm.ns, pyvw.namespace_id)
+    assert exm.ns_hash == 2514386435
+    assert exm.num_features_in() == 3
+    assert exm[2] == (11617, 1.0) # represents (feature, value)
+    assert exm.iter_features()
```
jackgerrits: This isn't doing anything? I dont think it makes sense to have this in the test

python/tests/test_pyvw.py
```diff
@@ -192,4 +193,95 @@ def test_parse():
...
+    assert ml.label == 2
+    assert ml.weight == 0.2
+    assert ml.prediction == 1
+    assert str(ml) == '2:0.2'
+
+def test_multiclass_label_example():
+    vw_ex = vw(quiet=True)
+    ex = vw_ex.example('1 |a two features |b more features here')
+    ml2 = pyvw.multiclass_label(ex)
+    assert ml2.label == 1065353216
```
jackgerrits: This doesn't seem right? Where is this number coming from?

python/tests/test_pyvw.py
```diff
@@ -192,4 +193,95 @@ def test_parse():
...
+    assert ml2.weight == 1.0
+    assert ml2.prediction == 0
+    assert str(ml2) == '1065353216'
+
+def test_example():
```
jackgerrits: There is a lot of functionality being tested here. Ideally it would be better in different tests please

python/tests/test_pyvw.py
Smit-create: Sure!

python/tests/test_pyvw.py
Smit-create: Sure, I will test for all of them.

python/tests/test_pyvw.py
Smit-create: Okay, I will make it more clearer.

python/tests/test_pyvw.py
Smit-create: Actually, this produces a generator object. Should I change this to return list?

python/tests/test_pyvw.py
Smit-create: Sure, I will part them up. Thanks!

python/tests/test_pyvw.py
jackgerrits: But you need to assert it against some known value for this to be useful

python/tests/test_pyvw.py
```diff
@@ -259,40 +274,54 @@ def test_multiclass_label():
...
+    def test_example_learn(self):
+        ex = self.ex
+        ex.learn()
+        assert ex.setup_done is True
+        ex.unsetup_example() # unsetup an example as it is already setup
+        assert ex.setup_done is False
+
+    def test_example_label(self):
+        ex = self.ex
+        ex.set_label_string('new_label')
```
jackgerrits: This is not a valid simple label

python/tests/test_pyvw.py
```diff
@@ -259,40 +274,54 @@ def test_multiclass_label():
...
+        assert isinstance(ex.get_label(), pyvw.simple_label)
+
+    def test_example_features(self):
+        ex = self.ex
+        ns = self.ns
+        assert ex.get_feature_id(ns, 123) == 123
```
jackgerrits: Does that feature actually exist in the example?

python/vowpalwabbit/pyvw.py
Smit-create: I ran this on two cases 1. with the check and 2. without the check.
I ran this using `timeit` module on the (i) synthetic data set (ii) test_simple_label and measuring the time of run as well as importing the module.
Following is the data of 5 runs in each of the cases.
#### Case 1 : with the check
(i) On synthetic dataset i.e. `test_regressor_args()`
The time for 5 runs is:` [0.11250083799995991, 0.1103579630000695, 0.11039327899993623, 0.11107653700014453, 0.11065440400011539]`
(i) On `test_simple_label`
The time for 5 runs is: `[0.1577475270000832, 0.13949805800007198, 0.14299159899996994, 0.14037750700003926, 0.13951651699994727]`

#### Case 2: without the check
(i) On synthetic dataset i.e. `test_regressor_args()`
The time for 5 runs is: `[0.1079615889998422, 0.10585603900017304, 0.10556867700006478, 0.10575059799998598, 0.10852373700004136]`

(i) On `test_simple_label`
The time for 5 runs is:` [0.15763619900008052, 0.13817744999983006, 0.13808634600013647, 0.13846224700000676, 0.13818081999988863]`

jackgerrits: Okay based on those results, do you believe the checks are worth it?

Smit-create: I think this should be added as it doesn't hit the performance too hard. You view on this?

jackgerrits: I agree

python/tests/test_pyvw.py
Smit-create: That feature doesn't exists, but I think this was done to avoid raising an error in the case if it doesn't exist, I will change this to assert for an existing feature so it does not misleads someone reading this.

python/vowpalwabbit/pyvw.py
Smit-create: Thanks!

python/tests/test_pyvw.py
Smit-create: Is this correct one `'1:0'`?

python/tests/test_pyvw.py
```diff
@@ -259,40 +274,54 @@ def test_multiclass_label():
...
     ml2 = pyvw.multiclass_label(ex)
-    assert ml2.label == 1065353216
+    assert ml2.label == 1
     assert ml2.weight == 1.0
     assert ml2.prediction == 0
-    assert str(ml2) == '1065353216'
+    assert str(ml2) == '1'
+
+
+class TestExample:
```
peterychang: Correct me if I'm wrong, but it looks like several functions in this class have side effects. Which means the order in which they are called will affect the test results. If that's the case, then the behavior is wrong for unit tests. Each function call should be completely independent from any others, and any requisite state information needs to be explicitly passed in as parameters.

Smit-create: Should it be one complete test as in earlier commit?

peterychang: Its better to have many small, independent functions that tests a specific piece of functionality. If thats not possible, then one large function may be unavoidable. However, it should never be assumed that tests will be called in a specific order; in fact some testing frameworks don't run tests in the order they appear in code

Smit-create: I agree. But for the tests of example with this case, it seems good to have it one test. Is this good idea?

peterychang: A single function for `test_example` is far too general; there are a lot of components to an example and there's no way to tell what's actually being tested.

Performance in the testing code (so anything in `test_pyvw.py`) is not really a concern. If you can break the unit tests into smaller pieces by doing duplicate work (like setting up the same example in different functions), thats fine.

If there are multiple components you feel are functionally the same thing, and it makes more sense to keep the checks in the same function call, then please do. We may ask why in some cases, but as long as there's a reasonable justification for the decision, it should be fine.

jackgerrits:
@Smit-create it looks like the tests are failing

Smit-create:
Locally, it passes all the tests. May be due to Python 2/3?  However, I will look at the issue shown in the log about segmentation fault

python/vowpalwabbit/pyvw.py
gramhagen:
```suggestion
        if not isinstance(label, (example, float)):
```

python/vowpalwabbit/pyvw.py
```diff
@@ -503,13 +505,15 @@ def from_example(self, ex):
...
         if self.weight != 1.:
-            s += ':' + self.weight
+            s += ':' + str(self.weight)
         return s
 
 
 class multiclass_label(abstract_label):
     """Class for multiclass VW label with prediction"""
     def __init__(self, label=1, weight=1., prediction=1):
+        if not (isinstance(label, int) or isinstance(label, example)):
```
gramhagen:
```suggestion
        if not isinstance(label, (example, int)):
```

gramhagen:
tests pass for me with these changes using python 2.7 and 3, maybe try retriggering tests?

Smit-create:
Does this look to merge?

Smit-create:
Thanks, @jackgerrits. Working on this PR helped a lot to have a better taste of understanding of each class

PR #2282, opened 2020-02-21
Add framework for managing predict_data in the reduction stack

peterychang:
Composite struct instead of the design used in https://github.com/VowpalWabbit/vowpal_wabbit/pull/2247

If we anticipate that this class will contain more than a few types, we should switch to macro code generation to ease maintenance.

Currently, every new type will need to add code to 4 places, totaling ~12 lines of code

This design uses templatized getter functions


jackgerrits:
The standardized clear is very useful, but beyond that I am not sure I understand the value add here over public fields. I understand public fields are a code smell and have lead to issues in the past, but the boilerplate doesn't really add protection or data hiding?

peterychang:
Having a templatized get() call lets us access the types we want without worrying about differing naming conventions people may use (was it `.s`, `.simple`, `.simple_predict`, `simple_data`, or `.simple_predict_data`?).

Also, instead of `predict_data.o.x = 5` you'd see something more descriptive like `predict_data.get<onnx::predict_data>().x = 5`

You also know the benefits of encapsulation. At the very least, making it a function call is one more deterrent against doing things that can cause problems down the line. It would also let us change things down the line without having to rewrite all of the places we access the data

jackgerrits:
You're totally right, encapsulation is a very good thing. (And something we need more of) The naming is also an excellent point. Wasn't trying to say this isn't right, just wanted your take on it

peterychang:
The central object and one example usage is done.

There's a bunch of scaffolding that'll need to be put in place before we can actually start to use this though. We'll need to add new function pointers for parsing. I might need help from @jackgerrits or @rajan-chari next week to figure some of it out, if either of you have time

lokitoth:
[Nit] Consider changing the title to be a bit more descriptive of what is going on in the change.

vowpalwabbit/predict_data.h
```diff
@@ -0,0 +1,40 @@
...
+ *   Accessing an instance of existing data:
+ *     features_data fd;
+ *     auto& data = fd.get<data_type>();
+ */
+
+
+class predict_data
+{
+ private:
+  CCB::predict_data _ccb_predict_data;
```
lokitoth: Could we drive generating this (and associated getters) through a template and inheritance?

In other words:

```
class predict_data : predict_container<CCB:predict_data> /* ... */
{
}
```

And predict container looks like:

```
template <typename T>
class predict_container 
{
private:
   T _data;

public:
   inline T& get() { return _data; };
   inline const T& get() const { return _data; };

protected:
   inline void clear() { _data.clear(); };
};
```

Not completely sure how to deal with clear, but possibly through a variadic template (ignore the completely incorrect syntax - I wanted to get the idea out without being bogged down looking at syntax; will update):

```
template <typename T, ...>
class predict_data_frame : predict_container<T>
{
private:
   predict_data_frame<NextT, ...> next;

public:
   inline void clear() 
   {
      // chain to protected method
      ((predict_container<T>*)this)->clear(); // is there cleaner syntax?
      next.clear();
   }
}

Comment (lokitoth) [reply vowpalwabbit/predict_data.h]: I worked out how to make the template machinery work, but it is not the prettiest thing in the world.

https://gist.github.com/lokitoth/5aa5695b44a32b306d2e6341ee0f8053

**Comment by rajan-chari** [comment]
Hi,  Can we merge this?  It would have been useful for some cb_adf related changes where predict is manipulating the label.  These are moved to example for now.

**Review by jackgerrits — APPROVED** [review]

File: vowpalwabbit/ccb_label.h [hunk 2]
```diff
@@ -40,5 +42,13 @@ struct label
   float weight;
 };
 
+struct predict_data
+{
+  example_type type;
+  v_array<uint32_t> explicit_included_actions;
```
lalo, line 48: are we leaving this in even if its still not used?

peterychang: I think its fine to leave this here for now. There should be another PR later to remove the variables from label and switch the usage over to this one

olgavrou (approved):
some files are missing the copyright header

vowpalwabbit/example_predict.h
```diff
@@ -47,6 +49,7 @@ struct example_predict
   // Interactions are specified by this vector of vectors of unsigned characters, where each vector is an interaction
   // and each char is a namespace.
   std::vector<std::vector<namespace_index>>* interactions;
+  predict_data _predict_data;
```
olgavrou, line 52:
```suggestion
  predict_data predict_data;
```
this is public and intended to be accessed from outside of the class, so var name should not be prepended with underscore

olgavrou (commented):
We also don't call the `example_predict::predict_data.clear()` from anywhere, should it be called during example cleanup? in `parser.cc -> empty_example()`?

peterychang:
Yes, probably. I also realized I forgot the destructor for CCB::predict_data (need to delete the v_array)

PR #2283, opened 2020-02-21
Fix race condition in daemon mode between parsing and inference threa…

### LINKED ISSUE #2201: Heisenbug w/ docker container running Vowpal Wabbit daemon.
#### Describe the bug
Heisenbug when making requests to docker container running Vowpal Wabbit daemon.

#### To Reproduce
1. Login to ECR `$(aws ecr get-login --no-include-email --region <region> --registry-ids 462105765813)`
2. Pull Vowpal Wabbit docker image `docker pull 462105765813.dkr.ecr.<region>.amazonaws.com/sagemaker-rl-vw-container:vw-8.7.0-cpu`
3. Start Vowpal Wabbit daemon `docker run --rm -it -p 26542:26542 <image_id> vw --daemon --cb_explore 4 --foreground`
4. Send requests to daemon `echo "'tag| a:1 b:1" | nc 0.0.0.0 26542`
5. Repeat (step 4.) a dozen times

#### Expected behavior
I expect step 4 to respond with something like`0.962500 0.012500 0.012500 0.012500  tag` every time. 

#### Observed Behavior
However, after the 6th (or so) repetition of step 4, there is a random chance that there is no response, and 
the vowpal daemon throws a `write error: Bad File descriptor`.

#### Environment
ECR container of Vowpal Wabbit version 8.7.0 cpu.
Hosted on default AWS sagemaker instance,  ml.t2.medium
NAME="Amazon Linux AMI"
VERSION="2018.03"

#### Additional context

I observed the AWS example code flushing std-out and std-in just after each request, and their code is running fine w/ the same container. This makes me think the issue could be related to a buffer overflow. I'm not sure how to force the container to flush to test this hypothesis...

kumpera:
…ds. Fixes #2201

Once the parsing thread detects end-of-stream for the current socket, it triggers parser.cc::reset_source,
which closes the socket and accepts a new one.

Before it does so, it waits for the example queue to be empty. Though this is a necessary condition,
it's not enough as the last example will be in flight and the inference thread can run into a closed socket
when it tries to write to it.

The fix is to keep counts of both parsed and finished examples, then wait for them to be equal before closing
the current socket.

### DIFF
--- a/vowpalwabbit/parser.cc
+++ b/vowpalwabbit/parser.cc
@@ -166,7 +166,7 @@ void reset_source(vw& all, size_t numbits)
       // wait for all predictions to be sent back to client
       {
         std::unique_lock<std::mutex> lock(all.p->output_lock);
-        all.p->output_done.wait(lock, [&] { return all.p->ready_parsed_examples.size() == 0; });
+        all.p->output_done.wait(lock, [&] { return all.p->finished_examples == all.p->end_parsed_examples && all.p->ready_parsed_examples.size() == 0; });
       }
 
       // close socket, erase final prediction sink and socket
@@ -925,6 +925,7 @@ void finish_example(vw& all, example& ec)
 
   {
     std::lock_guard<std::mutex> lock(all.p->output_lock);
+    ++all.p->finished_examples;
     all.p->output_done.notify_one();
   }
 }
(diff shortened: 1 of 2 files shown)


rajan-chari:
Makes sense

Issue #1469, opened 2018-04-27
[Python] [sklearn] Improvements to Sklearn Python interface


mlisovyi:
Hi,

I'm just starting to use VW in python. I'm very happy that there is a python binding and one dores not need to go around to dump `.vw` files with custom functions, running from a command line and then reading in the predictions in python again. And in particular, the sklearn wrapper is very useful, as it allows to use multiple handy functionalities from that package and makes usage universal across many algorithms. So thanks a lot!

I have a few questions to this interface:
1. Is there a documentation of the API (pyvw in general, and sklearn_vw in particular)? Would it be possible to add auto-generated docs from the class and function description? For example, one can use [Sphinx](http://www.sphinx-doc.org/en/master/) for that or any other tool that you would fancy. This will help users to find information faster.
2. I wanted to understand how exactly training is handled by the sklearn interface. It seems that there are 2 regimes: 
2.1 VW native approach based on a `.vw` file. This seems to be enabled by the `data` arg in the constructor and learning done in the `pyvw.vw()` instantiation in `get_vw()`. Is that understanding correct? If so, this is very nice as it allows online training with a restricted memory footprint, as advertised by the tool. However, from [here](https://github.com/JohnLangford/vowpal_wabbit/blob/master/python/vowpalwabbit/sklearn_vw.py#L251), it seems that even in this case one runs one (and always only one) _additional_ pass over the data provided to the `.fit()` function. Would it make sense to set the number of passes to 0 in this case?
2.2 The sklearn-like interface based on in-memory X and y arrays. In this case one iterates over `pyvw.vw.learn()` calls. That's fine for small datasets, I imagine. It also makes start of VW usage for people with sklearn experience much faster and smoother. The only thing here, would it be possible to add an additional parameter to the constructor that would be specific to the sklearn API and would allow one to convert `[0,1]` labels in the classification into `[-1,1]`. This simplifies life a lot, since in most cases data have `0` label instead of `-1`. One can maybe even reuse the existing `convert_to_vw_` attribute to trigger this conversion, but this would not be backward-compatible (I'm not sure how many people use the sklearn interface of VW by now).

[...]

PR #2285, opened 2020-02-24
Move VW object creation to sklearn constructor

Sharad24:
Reference Issue #1469


python/vowpalwabbit/sklearn_vw.py
```diff
@@ -508,14 +508,31 @@ class VWClassifier(SparseCoefMixin, ThresholdingLinearClassifierMixin, VW):
     note - don't try to apply link='logistic' on top of the existing functionality
     """
 
-    def __init__(self, **params):
+    def __init__(self, X=None, y=None, sample_weight=None, **params):
```
jackgerrits: Why is init taking the features and labels? What if you want to predict/learn with different examples?

Sharad24: My initial thought was if the data is passed the initialized model could be trained as well. If its not passed, user can still call fit/predict. Do we only require object creation or should it be training+testing in the constructor requiring X_train, y_train, X_test, y_test with defaults as None?

jackgerrits: The constructor should not be training the model

jackgerrits:
@Scott-Graham-Bose does this seem good to you?

jackgerrits:
I'm not sure if you noticed @Sharad24, @gramhagen commented on the issue. It looks like we do actually want to keep the lazy initialization. Here is @gramhagen's response quoted below:

> @jackgerrits I dug in a bit more on item 1. I believe lazy initialization is the right construct for this estimator, and it was added to support sklearn pipelines. We should keep the vw model construction out of init to maintain support for this. Handling the case where a data file is provided makes this a little funky, but I think it's better to stick to a single conceptual model, where training only happens when fit is called (and the vw model only needs to be instantiated at this point).

Sharad24:
Yes, I saw. It definitely makes sense to allow for building pipelines in sklearn. I'll revoke those changes. I'll push changes with point 4 in the issue, on addition of the `convert_labels` parameter.

PR #2286, opened 2020-02-24
Move destruction of search_private to its destructor

jackgerrits:


vowpalwabbit/search.cc
```diff
@@ -136,6 +136,8 @@ std::ostream& operator<<(std::ostream& os, const action_cache& x)
   return os;
 }
 
+void clear_memo_foreach_action(search_private& priv);
+
 struct search_private
 {
   //FIXME: uncomment this private once this struct has a proper destructor
```
peterychang: Can you uncomment this private?

vowpalwabbit/search.cc
```diff
@@ -298,6 +300,59 @@ struct search_private
                   // the "real" decision step but this really only matters for caching purposes
   v_array<v_array<action_cache>*>
       memo_foreach_action;  // when foreach_action is on, we need to cache TRAIN trajectory actions for LEARN
+
+  ~search_private(){
+    if(all)
```
peterychang: nit: Can we make sure we're destroying stuff in the correct order?

jackgerrits: Once the v_array change comes in most of this destructor goes away so I'll leave this work to then

vowpalwabbit/search.cc
```diff
@@ -298,6 +300,59 @@ struct search_private
...
+      active_known.delete_v();
+
+      if (cb_learner)
+        allowed_actions_cache->cb.costs.delete_v();
+      else
+        allowed_actions_cache->cs.costs.delete_v();
+
+      train_trajectory.delete_v();
+
+      for (auto& ar : ptag_to_action) delete ar.repr;
```
peterychang: Add this to ~action_repr()?

jackgerrits: Going to hold off on this because action_repr is pushed around by value, so we'd need to workout proper ownership as well when we do this

PR #2288, opened 2020-02-25
Do not block stderr on Nan prediction + Non-logging logger

ataymano:
Main change is bringing dependency on quiet to finalize_prediction to avoid stderr lock on Nans in quiet mode.
At the same time, step 0 for creating separate logging compenent is done (to avoid increasing amount of struct vw dependencies):
vw_logger member is added to vw struct. (with quiet only there, but later trace_message and proper trace/log/error interfaces should be added).
All changes are mechanical replacement quiet -> logger.quiet for now.


vowpalwabbit/global_data.h
```diff
@@ -345,6 +344,16 @@ struct rand_state
   void set_random_state(uint64_t initial) noexcept { random_state = initial; }
 };
 
+struct vw_logger
+{
+  bool quiet;
```
jackgerrits: Can you initialize this?

PR #2293, opened 2020-02-26
Add convert_labels param in VW sklearn class

Sharad24:
Reference Issue #1469 
Added a parameter convert_labels to convert labels of the form [0,1] to [-1,1] to suit VW


jackgerrits:
@gramhagen Would you be able to take a look?

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -598,6 +602,10 @@ def tovw(x, y=None, sample_weight=None):
     if not isinstance(y, np.ndarray):
         y = np.array(y)
 
+    # convert labels of the form [0,1] to [-,1]
+    if convert_labels:
+        y = np.asarray([-1 if i==0 else 1 for i in y])
```
gramhagen: Using where might be faster
```np.where(y < 1, -1, y)```

Sharad24: Thanks! Just updated

jackgerrits:
Given it has a default and doesn't change existing functionality it makes sense for a minor version bump. Version is defined by https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/version.txt but we don't have a process for documenting api changes

Sharad24:
Version change from `8.8.0` to `8.9.0`? Not sure why the windows CI failed after merging.

jackgerrits:
No, the version will be changed when we next put out a release.

jackgerrits:
The Windows CI has a flaky test (#1978), I'll restart it

Sharad24:
@jackgerrits Is this good to merge?

jackgerrits:
Yep, I'll merge once CI finishes.

Issue #1868, opened 2019-05-08
Replace signed 'char' uses with 'unsigned char' or 'int8_t'


lokitoth:
It is likely fairly rare that we need an actual 8-bit integral value. If we do need one, we should prefer the explicitly-sized `int8_t`, that way we do not confuse "this is a character in a string, which is naturally unsigned", especially if we want to use it as an array index, as opposed to "this is an integer that is signed and one byte in size".

This will prevent us from having errors like trying to index into arrays using a signed char (which breaks for any value > 127), e.g. indexing into feature_space in the json parser.

jackgerrits:
Perhaps using `uint8_t` or `int8_t` explicitly would be clearer? Most places we use a char we actually just want an unsigned byte, right?

lokitoth:
From a cursory look, it seems that most of the cases are either strings or namespace index, so we could use `namespace_index` where it makes sense. The main thrust of this is to ensure that any place where we are treating the chars as elements of strings, we should always use `unsigned char`, or explicitly declare it to be a number with a fixed size.

Sharad24:
@jackgerrits Could you point to an example of a case matching this issue? It looks like there are _very_ few of them present now.

jackgerrits:
In example_predict.h you'll notice there is a typedef of `namespace_index` to `unsigned char`. Namespace indices are the majority offender for this issue I'd say.

So this issue is mostly a work item about auditing usages of namespace indices and making sure that we are doing the correct thing. Namespaces are stored as a `std::string`, and so is a list of signed characters. But we use them as an unsigned integer to index an array. Processing the interaction is here and you'll notice plenty of casts. https://github.com/VowpalWabbit/vowpal_wabbit/blob/bd98471f330ed5aae2dea5bdb3610761c396786e/vowpalwabbit/interactions_predict.h#L98

I think it might make sense to turn interactions into a vector of `uint8_t`, @lokitoth thoughts?

Sharad24:
Ah gotcha, thanks! Will take a look

olgavrou:
@jackgerrits can we close this issue?

jackgerrits:
Sounds good!

PR #2296, opened 2020-02-27
 Change Interactions from signed char to uint8_t to avoid casting

Sharad24:
Reference Issue #1868 

Interactions changed from `vector<string>` to `vector<vector<uint8_t>>`


Sharad24:
The Build fails when running a unit test for interactions. So that will have to be changed.
`check_vectors(interactions, compare_set);`

test/unit_test/test_common.h
```diff
@@ -18,3 +18,10 @@ template <typename T>
 void check_vectors(const std::vector<T>& lhs, const std::vector<T>& rhs) {
   BOOST_CHECK_EQUAL_COLLECTIONS(lhs.begin(), lhs.end(), rhs.begin(), rhs.end());
 }
+
+template <typename T>
+void check_vectors(const std::vector<std::vector<T>>& lhs, const std::vector<std::vector<T>>& rhs) {
+  for (int i=0; i<lhs.size();i++){
```
jackgerrits:
```suggestion
  for (size_t i=0; i < lhs.size(); i++){
```

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -38,8 +38,8 @@ struct ccb
   std::vector<uint32_t> origin_index;
   CB::cb_class cb_label, default_cb_label;
   std::vector<bool> exclude_list, include_list;
-  std::vector<std::string> generated_interactions;
-  std::vector<std::string>* original_interactions;
+  std::vector<std::vector<uint8_t>> generated_interactions;
```
jackgerrits: Most of the usages of `uint8_t` should really be `namespace_index` as that is what they logically are. Also, `namespace_index` should be `uint8_t`

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -303,7 +303,7 @@ void calculate_and_insert_interactions(
           !found_namespaces[action_index - INTERACTIONS::printable_start])
       {
         found_namespaces[action_index - INTERACTIONS::printable_start] = true;
-        generated_interactions.push_back({(char)action_index, (char)ccb_id_namespace});
+        generated_interactions.push_back({(unsigned char)action_index, (unsigned char)ccb_id_namespace});
```
jackgerrits: Cast to namespace_index?

vowpalwabbit/parse_args.cc
```diff
@@ -824,7 +825,12 @@ void parse_feature_tweaks(options_i& options, vw& all, std::vector<std::string>&
       all.interactions.clear();
     }
 
-    all.interactions = expanded_interactions;
+    // all.interactions = expanded_interactions;
```
jackgerrits: Remove commented out code

vowpalwabbit/parse_args.cc
```diff
@@ -824,7 +825,12 @@ void parse_feature_tweaks(options_i& options, vw& all, std::vector<std::string>&
...
+
+    for (std::vector<std::string>::iterator t=expanded_interactions.begin(); t!=expanded_interactions.end(); t++){
```
jackgerrits: `auto`

vowpalwabbit/parse_args.cc
```diff
@@ -824,7 +825,12 @@ void parse_feature_tweaks(options_i& options, vw& all, std::vector<std::string>&
...
+      std::vector<uint8_t> s(t->begin(),t->end());
+      all.interactions.push_back(s);
```
jackgerrits: Can this be emplaced?

vowpalwabbit/parse_regressor.cc
```diff
@@ -327,19 +327,21 @@ void save_load_header(
             bytes_read_write += bin_text_read_write_fixed_validated(
                 model_file, (char*)&inter_len, sizeof(inter_len), "", read, msg, text);
             if (!read)
-            {
-              memcpy(buff2, all.interactions[i].c_str(), inter_len);
+            { 
+              std::string s(all.interactions[i].begin(),all.interactions[i].end());
```
jackgerrits: This is a redundant copy, you can copy the buffer directly

vowpalwabbit/parse_regressor.cc
```diff
@@ -327,19 +327,21 @@ void save_load_header(
...
+              msg.write(s.c_str(), inter_len);
             }
 
             bytes_read_write += bin_text_read_write_fixed_validated(model_file, buff2, inter_len, "", read, msg, text);
 
             if (read)
             {
               std::string temp(buff2, inter_len);
-              all.interactions.push_back(temp);
+              std::vector<uint8_t> temp2(temp.begin(), temp.end());
```
jackgerrits: Can you remove this temporary?

vowpalwabbit/parse_regressor.cc
```diff
@@ -349,8 +351,20 @@ void save_load_header(
         else  // < VERSION_FILE_WITH_INTERACTIONS
         {
           // pairs and triples may be restored but not reflected in interactions
-          all.interactions.insert(std::end(all.interactions), std::begin(all.pairs), std::end(all.pairs));
-          all.interactions.insert(std::end(all.interactions), std::begin(all.triples), std::end(all.triples));
+          std::vector<std::vector<uint8_t>> newpairs_t;
```
jackgerrits: Why is the variable name suffixed with _t?

vowpalwabbit/parse_regressor.cc
```diff
@@ -349,8 +351,20 @@ void save_load_header(
...
+          std::vector<std::vector<uint8_t>> newtriples_t;
+
+          for (std::vector<std::string>::iterator t = all.pairs.begin(); t != all.pairs.end(); t++){
+            std::vector<uint8_t> s(t->begin(), t->end());
+            newpairs_t.push_back(s);
```
jackgerrits: Emplaceable?

vowpalwabbit/slim/include/vw_slim_predict.h
```diff
@@ -416,7 +416,14 @@ class vw_predict
       ns_copy_guard->feature_push_back(1.f, (constant << _stride_shift) + ex.ft_offset);
     }
 
-    score = GD::inline_predict<W>(*_weights, false, _ignore_linear, _interactions, /* permutations */ false, ex);
+    // create temporary interactions of type vector<vector<uint8_t>> to feed in GD::inline_predict
```
jackgerrits: This is a hot function. It is not acceptable to create a copy of the interactions on each run, can you find a way to fix this?

vowpalwabbit/search_dep_parser.cc
```diff
@@ -93,9 +93,24 @@ void initialize(Search::search &sch, size_t & /*num_actions*/, options_i &option
   all.pairs.swap(newpairs);
   all.triples.swap(newtriples);
 
+  // std::vector<uint8_t> temp_var(pair[0], pair[0]+1);
```
jackgerrits: Commented out code?

vowpalwabbit/search_dep_parser.cc
```diff
@@ -93,9 +93,24 @@ void initialize(Search::search &sch, size_t & /*num_actions*/, options_i &option
...
+  std::vector<std::vector<uint8_t>> newpairs_t(19);// = new std::vector<std::vector<uint8_t>>(20);
```
jackgerrits: Is this test code that was left in?

vowpalwabbit/search_dep_parser.cc
Sharad24: Apologize for this!

vowpalwabbit/search_dep_parser.cc
Sharad24: Yeah, was running into segfault cause of that extra element

vowpalwabbit/parse_regressor.cc
Sharad24: Will change that

vowpalwabbit/parse_args.cc
Sharad24: Okay, will change that.

vowpalwabbit/conditional_contextual_bandit.cc
Sharad24: Changed

vowpalwabbit/parse_regressor.cc
Sharad24: Changed that

vowpalwabbit/parse_regressor.cc
Sharad24: I've removed temp2 and I've used emplace_back here instead now

vowpalwabbit/conditional_contextual_bandit.cc
Sharad24: Apologies, I didn't understand. Are you asking me to change uint8_t to namespace_index? Or the way round?

vowpalwabbit/parse_regressor.cc
Sharad24: Removed

vowpalwabbit/parse_regressor.cc
Sharad24: Removed

vowpalwabbit/slim/include/vw_slim_predict.h
Sharad24: Should be resolved now. I changed interactions type in `vw_slim_predict.h` from `vector<string>` to `vector<vector<uint8_t>>`. This required overloading the`find_opt` function. Could you check if this would be ok?

vowpalwabbit/parse_regressor.cc
Sharad24: Changed

vowpalwabbit/parse_args.cc
Sharad24: Added

vowpalwabbit/parse_args.cc
Sharad24: Removed

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: `namespace_index` is just an alias, it should map to `uint8_t`, but the `namespace_index` should be used where appropriate

vowpalwabbit/slim/src/opts.cc
```diff
@@ -6,6 +6,52 @@
 namespace vw_slim
 {
 // && used to avoid constant string copying
+void find_opt(std::string const& command_line_args, std::string arg_name, std::vector<std::vector<uint8_t>> out_values)
```
jackgerrits: This function is almost entirely duplicated with the copy below, can you find aware to reduce the code duplication here?

vowpalwabbit/parse_args.cc
```diff
@@ -825,11 +825,8 @@ void parse_feature_tweaks(options_i& options, vw& all, std::vector<std::string>&
       all.interactions.clear();
     }
 
-    // all.interactions = expanded_interactions;
-
-    for (std::vector<std::string>::iterator t=expanded_interactions.begin(); t!=expanded_interactions.end(); t++){
-      std::vector<uint8_t> s(t->begin(),t->end());
-      all.interactions.push_back(s);
+    for (auto t=expanded_interactions.begin(); t!=expanded_interactions.end(); t++){
+      all.interactions.emplace_back(std::vector<uint8_t>(t->begin(),t->end()));
```
jackgerrits: You are providing a temporary to emplace_back? Can you supply the constructor args directly?

vowpalwabbit/parse_regressor.cc
```diff
@@ -329,19 +329,18 @@ void save_load_header(
             if (!read)
             { 
               std::string s(all.interactions[i].begin(),all.interactions[i].end());
-              memcpy(buff2, s.c_str(), inter_len);
+              memcpy(buff2, &all.interactions[i], inter_len);
```
jackgerrits: Are you certain `all.interactions` is at least size i+1?

vowpalwabbit/parse_regressor.cc
```diff
@@ -329,19 +329,18 @@ void save_load_header(
...
 
               msg << "interaction: ";
-              msg.write(s.c_str(), inter_len);
+              msg.write((char*)all.interactions[i].data(), inter_len);
```
jackgerrits: This should be a reinterpret_cast so we aren't hiding this risky operation

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -303,7 +303,7 @@ void calculate_and_insert_interactions(
           !found_namespaces[action_index - INTERACTIONS::printable_start])
       {
         found_namespaces[action_index - INTERACTIONS::printable_start] = true;
-        generated_interactions.push_back({(char)action_index, (char)ccb_id_namespace});
+        generated_interactions.push_back({(namespace_index)action_index, (namespace_index)ccb_id_namespace});
```
jackgerrits: Prefer `static_cast`

vowpalwabbit/interactions.cc
```diff
@@ -239,7 +239,7 @@ void eval_count_of_generated_ft(vw& all, example& ec, size_t& new_features_cnt,
   if (all.permutations)
   {
     // just multiply precomputed values for all namespaces
-    for (std::string& inter : *ec.interactions)
+    for (std::vector<uint8_t>& inter : *ec.interactions)
```
jackgerrits: Auto?

vowpalwabbit/parse_args.cc
```diff
@@ -824,7 +825,9 @@ void parse_feature_tweaks(options_i& options, vw& all, std::vector<std::string>&
       all.interactions.clear();
     }
 
-    all.interactions = expanded_interactions;
+    for (auto t=expanded_interactions.begin(); t!=expanded_interactions.end(); t++){
```
jackgerrits: Can this be a range based for?

vowpalwabbit/parse_regressor.cc
```diff
@@ -327,18 +327,19 @@ void save_load_header(
             bytes_read_write += bin_text_read_write_fixed_validated(
                 model_file, (char*)&inter_len, sizeof(inter_len), "", read, msg, text);
             if (!read)
-            {
-              memcpy(buff2, all.interactions[i].c_str(), inter_len);
+            { 
+              std::string s(all.interactions[i].begin(),all.interactions[i].end());
```
jackgerrits: This string is redundant now right?

vowpalwabbit/parse_regressor.cc
```diff
@@ -327,18 +327,19 @@ void save_load_header(
...
+              msg.write((char*)all.interactions[i].data(), inter_len);
             }
 
             bytes_read_write += bin_text_read_write_fixed_validated(model_file, buff2, inter_len, "", read, msg, text);
 
             if (read)
             {
-              std::string temp(buff2, inter_len);
+              std::vector<uint8_t> temp(buff2, buff2 + inter_len);
               all.interactions.push_back(temp);
```
jackgerrits: This is a redundant copy

vowpalwabbit/search_dep_parser.cc
```diff
@@ -93,9 +93,21 @@ void initialize(Search::search &sch, size_t & /*num_actions*/, options_i &option
   all.pairs.swap(newpairs);
   all.triples.swap(newtriples);
 
+  std::vector<std::vector<uint8_t>> newpairs_(19);
```
jackgerrits: Can you please explain where these magic numbers come from?

vowpalwabbit/parse_regressor.cc
```diff
@@ -349,8 +350,18 @@ void save_load_header(
         else  // < VERSION_FILE_WITH_INTERACTIONS
         {
           // pairs and triples may be restored but not reflected in interactions
-          all.interactions.insert(std::end(all.interactions), std::begin(all.pairs), std::end(all.pairs));
-          all.interactions.insert(std::end(all.interactions), std::begin(all.triples), std::end(all.triples));
+          std::vector<std::vector<uint8_t>> newpairs_(all.pairs.size());
```
jackgerrits: This doesn't fit in with naming conventions

```suggestion
          std::vector<std::vector<uint8_t>> new_pairs(all.pairs.size());
```

vowpalwabbit/slim/include/vw_slim_predict.h
```diff
@@ -297,10 +297,10 @@ class vw_predict
 
     // VW performs the following transformation as a side-effect of looking for duplicates.
     // This affects how interaction hashes are generated.
-    std::vector<std::string> vec_sorted;
-    for (const std::string& interaction : _interactions)
+    std::vector<std::vector<uint8_t>> vec_sorted;
+    for (const std::vector<uint8_t>& interaction : _interactions)
```
jackgerrits: You could make the loop arg a non-ref, so you get a by value copy then sort that

vowpalwabbit/search_dep_parser.cc
Sharad24: These are the lengths of the c-style arrays `pairs` and `triples` declared few lines above

vowpalwabbit/parse_regressor.cc
Sharad24: Updated

vowpalwabbit/interactions.cc
```diff
@@ -239,7 +239,7 @@ void eval_count_of_generated_ft(vw& all, example& ec, size_t& new_features_cnt,
   if (all.permutations)
   {
     // just multiply precomputed values for all namespaces
-    for (std::vector<uint8_t>& inter : *ec.interactions)
+    for (auto inter : *ec.interactions)
```
jackgerrits: This is now a copy on each iteration, use a ref

vowpalwabbit/parse_args.cc
```diff
@@ -825,8 +825,11 @@ void parse_feature_tweaks(options_i& options, vw& all, std::vector<std::string>&
       all.interactions.clear();
     }
 
-    for (auto t=expanded_interactions.begin(); t!=expanded_interactions.end(); t++){
-      all.interactions.emplace_back(std::vector<uint8_t>(t->begin(),t->end()));
+    // for (auto t=expanded_interactions.begin(); t!=expanded_interactions.end(); t++){
+    //   all.interactions.emplace_back(std::vector<uint8_t>(t->begin(),t->end()));
+    // }
+    for (auto i:expanded_interactions){
```
jackgerrits: This is a copy

vowpalwabbit/slim/include/vw_slim_predict.h
Sharad24: Right, this would avoid sorting the original interaction vectors. Updated that

vowpalwabbit/parse_args.cc
Sharad24: Updated

vowpalwabbit/interactions.cc
Sharad24: Updated

vowpalwabbit/slim/src/opts.cc
Sharad24: Defined a template function. Instantiated with `vector<string>` and `vector<vector<uint8_t>>`. Should resolve the code duplication issue. Could you check if this is ok/requires any changes?

vowpalwabbit/conditional_contextual_bandit.cc
Sharad24: Updated

Sharad24:
Major Changes as of now:
- Change of interactions datatype from vector<string> to vector<vector<namespace_index>>
- Addition of template methods in slim/src/opts{.h,.cc} and test/unit_test/test_common.h to support vector<vector<namespace_index>> types.

Minor Changes:
- Change of loop variables in `parse_args.cc` and `parse_regressor.cc` to use `auto` (const can also be added here) and use refs to avoid creation of temporary copies.
- Change `push_back` to `emplace_back` in `parse_args.cc`, `parse_regressor.cc` and `slim/src/opts.cc`

vowpalwabbit/slim/include/vw_slim_predict.h
```diff
@@ -297,10 +297,10 @@ class vw_predict
 
     // VW performs the following transformation as a side-effect of looking for duplicates.
     // This affects how interaction hashes are generated.
-    std::vector<std::string> vec_sorted;
-    for (const std::string& interaction : _interactions)
+    std::vector<std::vector<namespace_index>> vec_sorted;
+    for (auto interaction : _interactions)
```
Sharad24: Usage of ref here to avoid temporary copies. Will add this in the next commit

Sharad24:
Converage fails on merging with master. Not sure but probably due to #2325

jackgerrits:
@Sharad24 Please ensure all tests are passing and lgtm issues are resolved before we review the latest iteration.

Sharad24:
Sounds good. The Windows CI is failing due to the recent change of the function `check_vectors` to `check_collections_exact`. Will resolve the errors

vowpalwabbit/slim/include/opts.h
```diff
@@ -3,10 +3,13 @@
 #include <vector>
 #include <string>
 
+#include "example_predict.h"
+
 // command line option parsing
 namespace vw_slim
 {
-void find_opt(std::string const& command_line_args, std::string arg_name, std::vector<std::string>& out_values);
+template<class Tcontainer>
```
jackgerrits: I think just name this T, as it doesn't have to be a container

vowpalwabbit/parse_regressor.cc
```diff
@@ -349,8 +348,18 @@ void save_load_header(
         else  // < VERSION_FILE_WITH_INTERACTIONS
         {
           // pairs and triples may be restored but not reflected in interactions
-          all.interactions.insert(std::end(all.interactions), std::begin(all.pairs), std::end(all.pairs));
-          all.interactions.insert(std::end(all.interactions), std::begin(all.triples), std::end(all.triples));
+          std::vector<std::vector<namespace_index>> new_pairs(all.pairs.size());
```
jackgerrits: This doesn't reserve but creates N default elements. I think you meant to reserve here?

vowpalwabbit/parse_regressor.cc
```diff
@@ -349,8 +348,18 @@ void save_load_header(
...
+          std::vector<std::vector<namespace_index>> new_triples(all.triples.size());
+
+          for (auto &t : all.pairs){
+            new_pairs.emplace_back(t.begin(), t.end());
```
jackgerrits: Why can't you simply insert into  all.interactions directly from all.pairs and all.triples?

vowpalwabbit/search_dep_parser.cc
```diff
@@ -93,9 +93,21 @@ void initialize(Search::search &sch, size_t & /*num_actions*/, options_i &option
   all.pairs.swap(newpairs);
   all.triples.swap(newtriples);
 
+  std::vector<std::vector<namespace_index>> new_pairs(19); //19 is the length of c-style pairs array declared above
```
jackgerrits: Can you not just insert into all.interactions directly here as well?

jackgerrits:
Just a couple more comments then I think we can get this merged. @lokitoth have you had a chance to have a look?

vowpalwabbit/parse_regressor.cc
Sharad24: all.interactions is changed to now have a default type `vector<vector<namespace_index>>` instead of `vector<string>`. So it requires me to create the temporary `new_pairs` and `new_triples` objects.

jackgerrits: `all.interactions.emplace_back(t.begin(), t.end());` ?

Sharad24: Ah okay. I get it now. I'll change this.

vowpalwabbit/search_dep_parser.cc
Sharad24: Will change this as well. Thanks for pointing out.

vowpalwabbit/parse_args.cc
```diff
@@ -824,7 +825,9 @@ void parse_feature_tweaks(options_i& options, vw& all, std::vector<std::string>&
       all.interactions.clear();
     }
 
-    all.interactions = expanded_interactions;
+    for (auto &i:expanded_interactions){
+      all.interactions.emplace_back(i.begin(), i.end());
```
peterychang: Its really strange that half this code is treating an interaction as a string and the other half as a vector of uint8s. @jackgerrits would it make sense to covert the underlying functions in interactions.h/cc to work on `vector<vector<uint8_t>>` instead of `vector<string>`?

I know `expand_namespaces_with_recursion()` is looking for the `:` character, but AFAIK thats the only actual character comparison we do

peterychang: Also, below, `all.pairs` and `all.triples` are strings, not uint8s. Do we care about that?

vowpalwabbit/parse_regressor.cc
```diff
@@ -328,18 +328,17 @@ void save_load_header(
                 model_file, (char*)&inter_len, sizeof(inter_len), "", read, msg, text);
             if (!read)
             {
-              memcpy(buff2, all.interactions[i].c_str(), inter_len);
+              memcpy(buff2, &all.interactions[i], inter_len);
```
peterychang: since interactions is a vector of vectors, you'll need to memcpy the underlying buffer of one of its elements
`memcpy(buff2, all.interactions[i].data(), inter_len);`

vowpalwabbit/parse_regressor.cc
peterychang: I get why we're using emplace_back here, but its really just hacking around the fact that pairs/triples is an incompatible type. How feasible would it be to convert all the interactions (including pairs/triples/interactions function signatures) to uint8s?

vowpalwabbit/parse_args.cc
jackgerrits: Yes, I think ideally all usages of namespaces should be consistent in their treating them as vectors of bytes

Sharad24: @peterychang thanks for pointing it out, I will change that.

peterychang:
FYI: This PR will affect what you're working on here
https://github.com/VowpalWabbit/vowpal_wabbit/pull/2236

Sharad24:
My initial thinking was we'd want to remove pairs and triples as in that exact PR.

Sharad24:
@peterychang @jackgerrits Changing interactions completely would require overloading one of the default parsing functions for the type `vector<vector<uint8_t> >`. I'm not 100% sure how to do that. 

Adding an extra style parser will help in grouping the same ns indexes but they would still be read as `signed chars`. Same goes with overloading the `boost::po` validate function. The arguments are read by default as a string so it looks like we'd require casting at some level? As `istream` reads `string`

Please correct me if my interpretation is wrong here.

Sharad24:
It seems like if I add an option for `vector<vector<uint8_t> >` it builds, but its naturally not able to read something like `--quadratic Tf` but its able to read `--quadratic T f`. Goes without saying but over here it would put `T` and `f` in different vectors which isnt the desired behavior.

jackgerrits:
It's okay to define the option as std::vector<std::string>> but then that type should not leak further. After it gets read it is converted into its true type and stored

Sharad24:
Alright, sounds good! I'll update changes in the PR accordingly

peterychang:
To add to Jack's comment, we should be converting the input type once in a single place (important since it will make the code maintainable). The conversion should happen as early as possible. Everything before the conversion should treat the input as strings, and everything after should treat them as uint8_ts

Sharad24:
The input type `vector<string>` is converted to `vector<vector<namespace_index>>` now in `parse_args.cc`. 

For every type (`cubic`, `quadratic` & `interactions`), its only done once in `parse_args.cc`

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -305,7 +305,7 @@ void calculate_and_insert_interactions(
           !found_namespaces[action_index - INTERACTIONS::printable_start])
       {
         found_namespaces[action_index - INTERACTIONS::printable_start] = true;
-        generated_interactions.push_back({(char)action_index, (char)ccb_id_namespace});
+        generated_interactions.push_back({static_cast<namespace_index>(action_index), static_cast<namespace_index>(ccb_id_namespace)});
```
peterychang, line 308: Instead of static_casting the `action_index`, we should just make it the correct type

https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/example_predict.h#L6
We have at least 2 different definitions for what a `namespace_index` is. We need to track any additional definitions down and unify everything

vowpalwabbit/interactions.cc
```diff
@@ -239,7 +239,7 @@ void eval_count_of_generated_ft(vw& all, example& ec, size_t& new_features_cnt,
   if (all.permutations)
   {
     // just multiply precomputed values for all namespaces
-    for (std::string& inter : *ec.interactions)
+    for (auto &inter : *ec.interactions)
```
peterychang, line 242: nit: use const references wherever we can

vowpalwabbit/parse_regressor.cc
```diff
@@ -262,16 +262,16 @@ void save_load_header(
 
           if (!read)
           {
-            memcpy(pair, all.pairs[i].c_str(), 2);
-            msg << all.pairs[i] << " ";
+            memcpy(pair, all.pairs[i].data(), 2);
+            msg << all.pairs[i].data() << " ";
```
peterychang, line 266: Are you sure this will do what you want it to do? Is this vector null terminated?

Its dangerous to treat a container's data() section as a null-terminated string. Please don't do that

vowpalwabbit/parse_regressor.cc
```diff
@@ -291,15 +291,15 @@ void save_load_header(
 
           if (!read)
           {
-            msg << all.triples[i] << " ";
-            memcpy(triple, all.triples[i].c_str(), 3);
+            msg << all.triples[i].data() << " ";
```
peterychang, line 294: Are you sure this will do what you want it to do? Is this vector null terminated?

Its dangerous to treat a container's data() section as a null-terminated string. Please don't do that

vowpalwabbit/parse_regressor.cc
Sharad24: Would creating a temporary variable to store as a string be ok here, then?

vowpalwabbit/conditional_contextual_bandit.cc
Sharad24: We don't need a static cast here. I'll remove that.

Also, `example_predict.h` defines `namespace_index` as an `unsigned char`

vowpalwabbit/parse_regressor.cc
peterychang: It would probably be ok this the same way as later in the code; using `stringstream.write()` to write a set number of bytes into the stream.

The biggest danger to that method is the fact that its possible to write nulls into the stringstream, which would cause unintuitive behavior when converting it to a string. I would say add comments to that effect and call `write()`.

nit: Additionally for futureproofing, we should memcpy `sizeof(type)*2` rather than `2`

Sharad24: This would require casting right? Because write would take `char *` and we would have `unsigned char *`

peterychang: Yes, unfortunately there's no way to get around casting here since we're treating this as a printable string here. That should be fine since its only being used for debug purposes.

Sharad24: Got it. Updated

vowpalwabbit/parse_regressor.cc
```diff
@@ -291,15 +293,17 @@ void save_load_header(
 
           if (!read)
           {
-            msg << all.triples[i] << " ";
-            memcpy(triple, all.triples[i].c_str(), 3);
+            // Use write because all.triples[i] is vector of uint8_ts
```
peterychang, line 296: Add a comment explaining null characters can be written to the stringstream with `write()` and what might happen in that case

Sharad24: Just updated, can you check if its ok?

test/unit_test/test_common.h
```diff
@@ -33,3 +33,10 @@ void check_collections_exact(const ContainerOneT<T>& lhs, const ContainerTwoT<T>
 {
   BOOST_CHECK_EQUAL_COLLECTIONS(std::begin(lhs), std::end(lhs), std::begin(rhs), std::end(rhs));
 }
+
+template <typename T>
+void check_namespaces_exact(const std::vector<std::vector<T>>& lhs, const std::vector<std::vector<T>>& rhs) {
```
jackgerrits, line 38: The implementation of this has nothing to do with namespaces, could the name of the function be changed to reflect the generic nature?

vowpalwabbit/interactions.cc
```diff
@@ -239,7 +239,7 @@ void eval_count_of_generated_ft(vw& all, example& ec, size_t& new_features_cnt,
   if (all.permutations)
   {
     // just multiply precomputed values for all namespaces
-    for (std::string& inter : *ec.interactions)
+    for (const auto &inter : *ec.interactions)
```
jackgerrits, line 242: nit: 
```suggestion
    for (const auto& inter : *ec.interactions)
```

vowpalwabbit/interactions.cc
```diff
@@ -268,7 +268,7 @@ void eval_count_of_generated_ft(vw& all, example& ec, size_t& new_features_cnt,
     generate_interactions<eval_gen_data, uint64_t, ft_cnt>(all, ec, dat);
 #endif
 
-    for (std::string& inter : *ec.interactions)
+    for (auto &inter : *ec.interactions)
```
jackgerrits, line 271: nit:
```suggestion
    for (auto& inter : *ec.interactions)
```

vowpalwabbit/parse_args.cc
```diff
@@ -759,8 +759,13 @@ void parse_feature_tweaks(options_i& options, vw& all, std::vector<std::string>&
         all.trace_message << i << " ";
     }
 
+    std::vector<std::vector<namespace_index> > new_quadratics;
+    for (const auto & i : quadratics){
```
jackgerrits, line 763: nit:
```suggestion
    for (const auto& i : quadratics){
```

vowpalwabbit/parse_args.cc
```diff
@@ -777,8 +782,13 @@ void parse_feature_tweaks(options_i& options, vw& all, std::vector<std::string>&
         all.trace_message << *i << " ";
     }
 
-    std::vector<std::string> exp_cubic =
-        INTERACTIONS::expand_interactions(cubics, 3, "error, cubic features must involve three sets.");
+    std::vector<std::vector<namespace_index> > new_cubics;
```
jackgerrits, line 785: nit:
```suggestion
    std::vector<std::vector<namespace_index>> new_cubics;
```

vowpalwabbit/parse_args.cc
```diff
@@ -777,8 +782,13 @@ void parse_feature_tweaks(options_i& options, vw& all, std::vector<std::string>&
...
+    for (const auto & i : cubics){
```
jackgerrits, line 786: nit:
```suggestion
    for (const auto& i : cubics){
```

vowpalwabbit/parse_args.cc
jackgerrits: throughout

test/unit_test/test_common.h
Sharad24: I've changed it to `check_vector_of_vectors_exact` now. Is that ok?

kumpera:
> 
> 
> Looks ok to me. @jackgerrits any thoughts on the changes?
> @kumpera You are/were working on the pairs/triples change. How does this look to you?

Looks ok to me.

jackgerrits:
LGTM

PR #2297, opened 2020-02-27
Fix str function in cbandit to produce correct format

### LINKED ISSUE #2295: Python cbandit_label str function provides incorrect output
The `cbandit_label` type provides a str function to convert the object back to a string representation. That is currently structured as:
```python
' '.join([str(c.label) + ':' + str(c.cost) for c in self.costs])
```

I would expect this output to match the input form for these labels, which can be found [here](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Logged-Contextual-Bandit-Example). The use of the word `label` is also is misleading. It is actually the action, so a variable name of `action` would be clearer.

Smit-create:
Fixes #2295

### DIFF
--- a/python/tests/test_pyvw.py
+++ b/python/tests/test_pyvw.py
@@ -122,13 +122,25 @@ def test_multilabel_prediction_type():
 
 def test_cbandits_label():
     model = vw(cb=4, quiet=True)
-    assert pyvw.cbandits_label(model.example('1 |')).costs[0].label == 1
+    cbl = pyvw.cbandits_label(model.example('1 |'))
+    assert cbl.costs[0].action == 1
+    assert cbl.costs[0].probability == 0.0
+    assert cbl.costs[0].partial_prediction == 0.0
+    assert cbl.costs[0].cost # == 3.4028234663852886e+38 (actual value) This fails due to precision variation on different environments
+    # test issue 2295
+    assert str(cbl) # == '1:3.4028234663852886e+38:0.0' (actual value) This fails due to precision variation on different environments
     del model
 
 
 def test_cost_sensitive_label():
     model = vw(csoaa=4, quiet=True)
-    assert pyvw.cost_sensitive_label(model.example('1 |')).costs[0].label == 1
+    csl = pyvw.cost_sensitive_label(model.example('1 |'))
+    assert csl.costs[0].action == 1
+    assert csl.costs[0].wap_value == 0.0
+    assert csl.costs[0].partial_prediction == 0.0
+    assert csl.costs[0].cost # == 3.4028234663852886e+38 (actual value) This fails due to precision variation on different environments
+    # test issue 2295
+    assert str(csl) # == '1:3.4028234663852886e+38:0.0' (actual value) This fails due to precision variation on different environments
     del model
 
 
--- a/python/tests/test_pyvw.py
+++ b/python/tests/test_pyvw.py
@@ -122,25 +122,25 @@ def test_multilabel_prediction_type():
 
 def test_cbandits_label():
     model = vw(cb=4, quiet=True)
-    cbl = pyvw.cbandits_label(model.example('1 |'))
+    cbl = pyvw.cbandits_label(model.example('1:10:0.5 |'))
     assert cbl.costs[0].action == 1
-    assert cbl.costs[0].probability == 0.0
-    assert cbl.costs[0].partial_prediction == 0.0
-    assert cbl.costs[0].cost # == 3.4028234663852886e+38 (actual value) This fails due to precision variation on different environments
+    assert cbl.costs[0].probability == 0.5
+    assert cbl.costs[0].partial_prediction == 0
+    assert cbl.costs[0].cost == 10.0
     # test issue 2295
-    assert str(cbl) # == '1:3.4028234663852886e+38:0.0' (actual value) This fails due to precision variation on different environments
+    assert str(cbl) == '1:10.0:0.5'
     del model
 
 
 def test_cost_sensitive_label():
     model = vw(csoaa=4, quiet=True)
-    csl = pyvw.cost_sensitive_label(model.example('1 |'))
-    assert csl.costs[0].action == 1
+    csl = pyvw.cost_sensitive_label(model.example('2:5 |'))
+    assert csl.costs[0].action == 2
     assert csl.costs[0].wap_value == 0.0
     assert csl.costs[0].partial_prediction == 0.0
-    assert csl.costs[0].cost # == 3.4028234663852886e+38 (actual value) This fails due to precision variation on different environments
+    assert csl.costs[0].cost == 5.0
     # test issue 2295
-    assert str(csl) # == '1:3.4028234663852886e+38:0.0' (actual value) This fails due to precision variation on different environments
+    assert str(csl) == '2:5.0:0.0'
     del model
 
 
--- a/python/tests/test_pyvw.py
+++ b/python/tests/test_pyvw.py
@@ -127,7 +127,6 @@ def test_cbandits_label():
     assert cbl.costs[0].probability == 0.5
     assert cbl.costs[0].partial_prediction == 0
     assert cbl.costs[0].cost == 10.0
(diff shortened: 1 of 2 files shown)


python/vowpalwabbit/pyvw.py
```diff
@@ -564,7 +564,7 @@ def __init__(self, label, cost=0., partial_prediction=0., probability=0.):
             self.costs.append(wc)
 
     def __str__(self):
-        return ' '.join([str(c.label) + ':' + str(c.cost) for c in self.costs])
+        return ' '.join([str(c.action) + ':' + str(c.cost) + ':' + str(c.probability) for c in self.costs])
```
jackgerrits: Could you change this to using `format`? Will make the output a bit clearer

Smit-create: Sure!

python/tests/test_pyvw.py
```diff
@@ -122,13 +122,25 @@ def test_multilabel_prediction_type():
 
 def test_cbandits_label():
     model = vw(cb=4, quiet=True)
-    assert pyvw.cbandits_label(model.example('1 |')).costs[0].label == 1
+    cbl = pyvw.cbandits_label(model.example('1 |'))
```
jackgerrits: Since you are asserting all of the action cost and prob values below you should update this label with known values

Smit-create:
Is this good to merge?

python/vowpalwabbit/pyvw.py
```diff
@@ -533,7 +533,7 @@ def __init__(self, label, cost=0., partial_prediction=0., wap_value=0.):
             self.costs.append(wc)
 
     def __str__(self):
-        return ' '.join([str(c.label) + ':' + str(c.cost) for c in self.costs])
+        return ' '.join(["{}:{}:{}".format(c.action, c.cost, c.wap_value) for c in self.costs])
```
jackgerrits: CSOAA labels are not formed like this. You can read the parse function to get an idea of how these work. https://github.com/VowpalWabbit/vowpal_wabbit/blob/3c0f0ffb84c47701521abba6450a8cf7b39fba83/vowpalwabbit/cost_sensitive.cc#L124

jackgerrits:
So I suggested that action is a better name, logically that is true. However, changing the field name in Python is actually a breaking change so we cannot do it without deprecated the old one etc.

Smit-create:
Okay, I will change it accordingly using deprecated warnings

python/tests/test_pyvw.py
```diff
@@ -122,13 +122,25 @@ def test_multilabel_prediction_type():
 
 def test_cbandits_label():
     model = vw(cb=4, quiet=True)
-    assert pyvw.cbandits_label(model.example('1 |')).costs[0].label == 1
+    cbl = pyvw.cbandits_label(model.example('1:10:0.5 |'))
+    assert cbl.costs[0].action == 1
+    assert cbl.costs[0].probability == 0.5
+    assert cbl.costs[0].partial_prediction == 0
+    assert cbl.costs[0].cost == 10.0
+    # test issue 2295
```
lalo:
```suggestion
```

python/tests/test_pyvw.py
```diff
@@ -122,13 +122,25 @@ def test_multilabel_prediction_type():
...
 
 def test_cost_sensitive_label():
     model = vw(csoaa=4, quiet=True)
-    assert pyvw.cost_sensitive_label(model.example('1 |')).costs[0].label == 1
+    csl = pyvw.cost_sensitive_label(model.example('2:5 |'))
+    assert csl.costs[0].action == 2
+    assert csl.costs[0].wap_value == 0.0
+    assert csl.costs[0].partial_prediction == 0.0
+    assert csl.costs[0].cost == 5.0
+    # test issue 2295
```
lalo:
```suggestion
```

Smit-create:
@jackgerrits  Does this look good now?

python/vowpalwabbit/pyvw.py
```diff
@@ -597,11 +601,15 @@ def __init__(self, costs=[], prediction=0):
...
-                         probability=0.):
-                self.label = label
+                        action=None,
+                        cost=0.,
+                        partial_prediction=0.,
+                        probability=0.,
+                        **kwargs):
+                if kwargs.get('label', False):
+                    action = kwargs['label']
+                    warnings.warn("label has been deprecated. Please use 'action' instead.", DeprecationWarning)
```
jackgerrits: Deprecating should not break existing functionality. If someone tries to retrieve the `label` field it will fail now

Smit-create: Okay, I will fix it now!

python/vowpalwabbit/pyvw.py
```diff
@@ -562,11 +562,16 @@ def __init__(self, costs=[], prediction=0):
     def from_example(self, ex):
         class wclass:
             def __init__(self,
-                         label,
-                         cost=0.,
-                         partial_prediction=0.,
-                         wap_value=0.):
-                self.label = label
+                        action=None,
```
jackgerrits: Cost sensitive labels do not call this term an action - these are classes. I think for this class though you should revert the label ->action change to reduce churn/scope of this PR

Smit-create: Sure, I will remove that change

python/vowpalwabbit/pyvw.py
```diff
@@ -562,10 +562,10 @@ def __init__(self, costs=[], prediction=0):
     def from_example(self, ex):
         class wclass:
             def __init__(self,
-                         label,
-                         cost=0.,
-                         partial_prediction=0.,
-                         wap_value=0.):
+                        label,
```
jackgerrits: Please fix the spacing change

Smit-create:
Thanks! @jackgerrits :)

python/vowpalwabbit/pyvw.py
```diff
@@ -597,11 +597,16 @@ def __init__(self, costs=[], prediction=0):
...
-                         label,
+                         action=None,
                          cost=0.,
                          partial_prediction=0.,
-                         probability=0.):
-                self.label = label
+                         probability=0.,
+                         **kwargs):
+                if kwargs.get('label', False):
+                    action = kwargs['label']
```
peterychang: If action and label are both passed in, label will overwrite action. Is that the behavior we want?

python/vowpalwabbit/pyvw.py
```diff
@@ -597,11 +597,16 @@ def __init__(self, costs=[], prediction=0):
...
+                    warnings.warn("label has been deprecated. Please use 'action' instead.", DeprecationWarning)
+                self.label = action
```
peterychang: Do you know if there's any way to give a deprecation warning for an individual variable? We'll probably want to transition users away from calling object.label at some point.

If you don't know of any mechanism for it, its not a big deal though

python/vowpalwabbit/pyvw.py
Smit-create: Actually, I don't think that it might happen as there are two possibilities if any user is familiar with is old format then he might use them as below,
First,
`>>> wc = wclass(paramter1, parameter2, paramter3, parameter4)`
or second,
```
>>> wc = wclass(label=paramter1, cost=paramter2, partial_prediction=paramter3, probability=paramter4)
```
If someone knowing the current format would use the below two formats
`>>> wc = wclass(paramter1, parameter2, paramter3, parameter4)`
or,
```
>>> wc = wclass(action=paramter1, cost=paramter2, partial_prediction=paramter3, probability=paramter4)
```
So the all 4 cases works fine with the current implementation.

python/vowpalwabbit/pyvw.py
Smit-create: I am presently not familiar with any such way.

Smit-create:
Looks good to merge?

Smit-create:
Thanks @jackgerrits @peterychang :)

PR #2299, opened 2020-02-28
added quantile loss

adityauser:
Added quantile loss function.
I will also try to add supports for parallel search via MongoTrials.


lalo:
Would it be good to have a test in place @jackgerrits ? https://github.com/VowpalWabbit/vowpal_wabbit/blob/7402da2c427418dc0c281391563391f368ba3af8/test/RunTests

jackgerrits:
Yes, I think a test for this would be useful

adityauser:
@jackgerrits I'm trying to create tests for hypersearch, but the command `vw-hypersearch  1e-10  5e-4  vw  --l1 %  train.dat` is throwing an error `vw-hypersearch: command not found`, can you tell how to install vw-hypersearch? and what does sample training data look like?

jackgerrits:
How are you trying to invoke vw-hypersearch? It must not be in the PATH of the system you are using to call it

adityauser:
Thanks, @jackgerrits. The PATH was incorrect.

adityauser:
@lalo I have added tests.
I was not sure where I needed to add the test files so I added it in [unit_tests](https://github.com/VowpalWabbit/vowpal_wabbit/test/unit_test).
The dataset I have used is same as [0001.dat](https://github.com/VowpalWabbit/vowpal_wabbit/tree/master/test/train-sets/0001.dat), except 0 labels are now replaced with -1.

adityauser:
@jackgerrits do vw-hyperopt also works for CB? If yes, then can you please show an example.

jackgerrits:
I have not used vw-hyperopt, so I can't really assist there. Perhaps you should look at past contributors to the file?

adityauser:
Alright, although I don't think it does support CB algorithms, so I have opened a draft PR for it.

jackgerrits:
These tests are just asserting that nothing crashes? They are also not included in any CI.

adityauser:
I'm currently working on some other projects, I won't be able to work in this PR. Do you want me to close this?

PR #2301, opened 2020-02-28
[FIX] Make Python Examples Up To Date With Scikit-Learn Version 0.22.1

### LINKED ISSUE #2281: Python examples doesn't support recent version of scikit-learn i.e. 0.22
#### Describe the bug
Example in https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/examples/sklearn_vw.ipynb use outdated terminology(Look at the 2nd cell):
`RandomizedSearchCV `module is now under `sklearn.model_selection`, and not any more under `sklearn.grid_search`  .
Onwards scikit-learn version 0.18.1 `grid_scores_` is no longer in use. We can use `cv_results_` to get the results, which is similar to `grid_scores_` . 
#### To Reproduce

Run the ipython notebook. 

#### Observed Behavior
```
ModuleNotFoundError                       Traceback (most recent call last)

<ipython-input-55-3d5c4d474787> in <module>()
      1 from operator import itemgetter
      2 from time import time
      3 from sklearn.grid_search import RandomizedSearchCV
      4 from scipy.stats.distributions import uniform
      5 

ModuleNotFoundError: No module named 'sklearn.grid_search'

```
#### Environment

`VW 8.8.0 on Linux with the Python bindings`

#### Additional context
Do you want me to make a PR for this.

thealphadollar:
As per #2281, the current examples are not up to date with the latest scikit-learn version. The Pull Request implements the following changes in order to fix the issue and also some refactoring has been accommodated.

0. [fix] Add dist folder to .gitignore
	- This doesn't let a developer mistakenly commit newly build files to the repository

0. [fix] Make sklearn_vs example up to date with sklearn v0.22.1
	- Update outdated help link in cell 2
    - As per v0.22.1, change `from sklearn.grid_search import RandomizedSearchCV` to `from sklearn.model_
selection import RandomizedSearchCV`
    - Add Pandas to handle cv_results_ and update other data get operations accordingly.
    - Use itertuples since they are faster compared to iterrows in pandas DataFrame.
    
    - https://www.dataschool.io/how-to-update-your-scikit-learn-code-for-2018/#gridsearchandrandomizedsearchhavechangedhowtheyreportresults
    - https://stackoverflow.com/a/41022840
    - Github issue #2281

0. [dbg] Suppress FutureWarning for import from sklearn.linear_model.base
	- Importing From sklearn.linear_model.base has been depricated in v0.22 and will be removed in v0.24
    - We need to import from the same since LinearClassifierMixin and SparseCoefMixin haven't been added 
to linear_model module yet
    - Warning suppression is important since it hampers the aesthetics of the output and confuses beginne
rs
    
    - https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/linear_model/__init__.py
    - https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/linear_model/_base.py

0. [fix] Ignore models generated in python/examples in git history
	- Prevents developers from accidentally committing the saved models from examples

fixes #2281


jackgerrits:
Can you please clear the outputs?

thealphadollar:
@jackgerrits Output has been removed.

.gitignore
```diff
@@ -136,3 +136,4 @@ reinforcement_learning/unit_test/rlclient-test.out
 cs/azure_service/azure_service.ccproj.user
 
 build
+dist/
```
lalo: why is this needed? what is creating this dir?

thealphadollar:
@jackgerrits The PR is ready for review.

python/examples/sklearn_vw.ipynb
```diff
@@ -44,50 +39,31 @@
...
    "source": [
     "# Parameter Grid Search\n",
-    "# http://scikit-learn.org/stable/auto_examples/model_selection/randomized_search.html#example-model-selection-randomized-search-py\n",
+    "# https://scikit-learn.org/stable/auto_examples/model_selection/plot_randomized_search.html#sphx-glr-auto-examples-model-selection-plot-randomized-search-py\n",
     "\n",
+    "import pandas as pd\n",
     "from operator import itemgetter\n",
     "from time import time\n",
-    "from sklearn.grid_search import RandomizedSearchCV\n",
+    "from sklearn.model_selection import RandomizedSearchCV\n",
```
lalo: is there anywhere to specify the min version of sklearn supported? Not sure if this file is outdated https://github.com/VowpalWabbit/vowpal_wabbit/blob/5cba65d53d514a9f6273d12d059f524829393803/requirements.txt#L3

.gitignore
thealphadollar: The directory `dist/` is generated when `python setup.py install` is run. 

It is clearly mentioned [here](https://packaging.python.org/tutorials/packaging-projects/#generating-distribution-archives)

> command should output a lot of text and once completed should generate two files in the dist directory

We do not want this to be added to our git history.

python/examples/sklearn_vw.ipynb
thealphadollar: Certainly the file seems outdated; a good fix would be to update it to `>=0.22.0` which would imply that we need at least the version  0.22.0 for smooth working.

A small recommendation would be to refrain from changing the requirements.txt in this PR since I think we should keep minimum requirement updates in a separate PR in case something breaks :)

thealphadollar:
@jackgerrits Can you please review as you get time?

python/examples/sklearn_vw.ipynb
```diff
@@ -2,24 +2,19 @@
...
+   "outputs": [],
    "source": [
     "# Binary Classification\n",
+    "\n",
     "from __future__ import print_function\n",
     "\n",
+    "# import warnings filter\n",
+    "from warnings import simplefilter\n",
+    "# ignore all future warnings\n",
+    "simplefilter(action='ignore', category=FutureWarning)\n",
```
jackgerrits: Which future warnings are being ignored by this?

python/examples/sklearn_vw.ipynb
thealphadollar: As per discussion with @peterychang , this is not to be updated at the moment.

python/examples/sklearn_vw.ipynb
thealphadollar: Currently, we are importing LinearClassifierMixin and SparseCoefMixin from sklearn.linear_model.base. In version v0.24 this importing will be removed and currently it gives the deprecation warning. I thought that in example, when someone tries they should not be concerned with the warning but just the output to learn. I'm currently working on the fix for the issue as well.

python/examples/sklearn_vw.ipynb
peterychang: If you want to make this compatible with both the older and newer versions of sklearn, you can conditionally import the object based on the version of sklearn being used

thealphadollar: What do you say would be a good way to do the same; a try-catch block or checking the version and then making the import? I prefer the former.

thealphadollar: I've used the former method in 4befc30 to implement the same.

python/examples/sklearn_vw.ipynb
thealphadollar: @jackgerrits If you are satisfied with my response, can you please mark the comment as resolved?

jackgerrits: Can you please add a comment explaining this?

thealphadollar: Done.

PR #2304, opened 2020-02-28
[FIX] Remove Outputs From Example Notebooks

### LINKED ISSUE #2303: [IMPROVEMENT] Remove Outputs From Existing Python Examples
Some of the examples (for instance [Contextual_Bandit_Example_with_VW_Python_Wrapper](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/examples/Contextual_Bandit_Example_with_VW_Python_Wrapper.ipynb)) do not have the corresponding output with them. We should add outputs since,

- The output provide a base expected result and if something is an amiss user trying the example can easily note.
- It makes it readable without running the notebook locally which is useful for intuition based learning.

@jackgerrits Please give your views and, thereafter, I'll add the output to all of the python examples.

thealphadollar:
As per the conversation in #2303 , this PR removes the outputs from each of the cells of the example notebooks.

The following is the list of notebooks from which output has been removed.

- poisson_regression
- Uncertainty_Scores_and_Prediction_Errors
- VW_in_Python
- Learning_to_Search

Example output has been removed from sklearn_vm in #2301 

fixes #2303


thealphadollar:
@jackgerrits Can we merge this? Are there any more changes required?

PR #2305, opened 2020-02-29
Uniform docstyle in sklearn_vw.py

Smit-create:
Using numpy docstyle in `sklearn_vw.py`


python/vowpalwabbit/sklearn_vw.py
```diff
@@ -20,14 +20,17 @@
...
 class VW(BaseEstimator):
-    """Vowpal Wabbit Scikit-learn Base Estimator wrapper
-
-        Attributes
-        ----------
-        params : {dict}
-                 dictionary of model parameter keys and values
-        fit_ : {bool}
-               this variable is only created after the model is fitted
+    r"""
```
jackgerrits: Why do these need to become raw strings?

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -117,115 +120,192 @@ def __init__(self,
                  inpass=None,
                  meanfield=None,
                  multitask=None):
-        """VW model constructor, exposing all supported parameters to keep sklearn happy
+        r"""
```
jackgerrits: Why move the short summary to a new line?

jackgerrits:
Can you please provide an example of how this changes the output?

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -521,12 +618,14 @@ def predict(self, X):
         """Predict class labels for samples in X.
 
         Parameters
-        ----------
+        ==========
```
jackgerrits: Why change these to `=`, it looks like they are `-` in the numpy styleguide? https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

Smit-create: I have used it because when we use Sphinx for creating auto docs it treats them as headings and highlights them.
[http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html)

python/vowpalwabbit/sklearn_vw.py
Smit-create: This change will be useful when we will use Sphinx for auto docs. In Python docstrings, it will be necessary to escape any backslash characters so that they actually reach reStructuredText. The simplest way to do this is to use raw strings by adding the letter r in front of the docstring.
[https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html](https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html)

python/vowpalwabbit/sklearn_vw.py
Smit-create: Sorry, it can be reverted to original one.

python/vowpalwabbit/sklearn_vw.py
Smit-create: I figured that the actual problem why it didn't appear in the docs was that it has no 'numpydoc' extension provided by sphinx.

Smit-create:
This is the resulting doc. I have attached a pdf of it.
@jackgerrits  Does this look good?
[doc.pdf](https://github.com/VowpalWabbit/vowpal_wabbit/files/4314222/doc.pdf)

Smit-create:
@jackgerrits Thanks! Yeah, it actually looks good.

Smit-create:
Also, please have a look at the attached file which is the resulting doc of `pyvw.py`
[doc2.pdf](https://github.com/VowpalWabbit/vowpal_wabbit/files/4314296/doc2.pdf)

Smit-create:
Does it look good to merge?

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -122,112 +124,189 @@ def __init__(self,
...
-        ftrl_alpha: ftrl alpha parameter
-        ftrl_beta: ftrl beta parameter
-        learning_rate,l (float): Set learning rate
-        power_t (float): t power value
-        decay_learning_rate (float): Set Decay factor for learning_rate between passes
-        initial_t (float): initial t value
-        feature_mask (str): Use existing regressor to determine which parameters may be updated.
-                            If no initial_regressor given, also used for initial weights.
+
+        bfgs : use L-BFGS optimization algorithm
```
gramhagen: Can you add the missing types for these?

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -122,112 +124,189 @@ def __init__(self,
...
+            Use a sigmoidal feed-forward neural network with N hidden units
+        dropout : Train or test sigmoidal feed-forward network using dropout
+        inpass : Train or test sigmoidal feed-forward network with input pass-through
+        multitask : Share hidden layer across all reduced tasks
+        meanfield : Train or test sigmoidal feed-forward network using mean field
 
         Returns
         -------
+
         (BaseEstimator): Returns self
```
gramhagen: I don't think all the returns March the numpy docstrings format.?

Should be:

Returns
-------
int
    Description of anonymous integer return

Smit-create: Yes, I have changed that in numpydoc style

python/vowpalwabbit/sklearn_vw.py
Smit-create: Yes, I will add them

Smit-create:
Please have a look at the updated doc attached here. I have added a function in `conf.py` to document available options that are documented in` __init__`
[doc3.pdf](https://github.com/VowpalWabbit/vowpal_wabbit/files/4322925/doc3.pdf)

gramhagen:
Looks good. @jackgerrits how do we push the rendered version to vowpalwabbit.org?

jackgerrits:
The updating process is manual for now, so I will handle that once it is merged. I have started work on automating but haven't had time to finish it.

jackgerrits:
Thanks @Smit-create!

Issue #2264, opened 2020-02-12
Multiclass labels do not error when given a multilabel


jackgerrits:
The core of the issue here is that the number parser used for multiclass labels (and other places I'm sure) doesn't check that the number it parsed was the entire token. The parser should use the endptr strtol argument of `strtol` to check that the entire token was the number.

The label `1,2 | ...` which is a valid multilabel is interpreted as a multiclass label of `1` but instead there should be an error as this is a clear user pitfall.

RituRajSingh878:
I want to work on the issue, but I don't know form where I should start.

And if I make some changes in files(.cpp or .cc) then How should I check it or test it?
(we do for python files `python setup.py install --user `and `python3 - pytest ./python/tests/`)

Shivanshmundra:
@RituRajSingh878 once you change files, you can build the solution just like you install Vowpal Wabbit from source and see how is this working. [Here](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Building) is the link to do that.

Shivanshmundra:
@jackgerrits As I was looking into algorithms, I was looking into this problem. I could not work out the solution suggested by you because:
* ```strtol``` finds the number in the remaining string from ```endptr```.
* Without tokenization lines can't be separated as examples.
But using tokenization in ```multiclass.cc``` can solve this problem. I have solved using this and submitting PR.
@RituRajSingh878 Hope you don't mind me taking this issue as I needed to solve this for my work.

Issue #2262, opened 2020-02-12
vw --cbify does not properly handle multi-label dataset


haltux:
#### Describe the bug

Multi-label datasets for CB experimentation seems to be highly relevant considering that practically, in real life CB problem, several actions often lead to sucess, which unless I am missing something is not the case in experiments based on multi-class problems.

It seems that multi-label dataset can be generated to be used by Vowpwabbit for CB experiments. I have not found it documented anywhere, however such datasets are specifically generated for vw in this project: https://github.com/mwydmuch/datasets4vw 

Additionnaly, the paper "A contextual bandit bake off" cite experiments on multi-label data with vowpal wabbit, even though unfortunately it seems that they released the code only for single-label data.

The format is (from the previously cited project): 
```
1,2 | x:0 y:1
1,3 | x:1 y:1
```

This is properly parsed by vw and raises no error. Unfortunately I could not find a way to properly use it.

#### To Reproduce

Create this dataset and store it in /tmp/data.vw:
```
1,2,3 | x:0 y:0
```

Run:
```
vw --cbify 3 --cb_explore_adf /tmp/data.vw
```

#### Expected behavior
As all labels are associated to the unique data point, the loss should be 0.

#### Observed Behavior
```
average  since         example        example  current  current  current
loss     last          counter         weight    label  predict features
1.000000 1.000000            1            1.0        1        2        1
```

The current label is 1. Just like if the other labels where compeltely ignored. So as the cb algorithm predict 2, it gets a loss of 1, which is wrong.

If this is really the expected behavior, please tell me how it is actually possible to run vw on multi-label datasets.

#### Environment
8.8.0 on linux /  command line.

jackgerrits:
Hi @haltux,

So there are two things here. `--cbify` takes multiclass labels, and not multilabels. As you correctly noticed, the parser is actually extracting the first number and using that as the label here. This is clearly a problem, the parser should have given you an error instead. I've created issue #2264 to track this.

The second thing is how you can achieve the behavior you're looking for. I briefly discussed this with @pmineiro, I'll let him chime in here as I think he can do a better job explaining.

pmineiro:
Your question unfortunately requires a verbose response.

## Supervised Multi-class and Multi-label

[...]

PR #2306, opened 2020-03-01
[FIX] Multiclass labels do not error when given multilabels

Shivanshmundra:
As per discussion in #2264 and #2262 , this PR solves the issue when multi labels are given to multi-class algorithms where it should raise an error.

Previously it was just taking the label from the first index which was wrong.
Instead of using ```strtol``` function, I tokenized the string and looked for number of labels in an example.


RituRajSingh878:
you should also add some tests case for the same.

jackgerrits:
The issue raises the problem where the label parser does not fail if the whole token is not consumed. This is not just when accidentally passing a multilabel, it is for any input where the whole token isn't consumed. For example, it should fail for `1a` as well. So the label parser can't just be enlightened to commas.

Shivanshmundra:
So checking whether label string is number or not will check for every case, right?
Something like this function:
```
bool is_number(const std::string& s)  
{  
      std::string::const_iterator it = s.begin();
      while (it != s.end() && std::isdigit(*it)) ++it;
      return !s.empty() && it == s.end();
}

```
But the issue is to get `endptr` to that string, as in this line https://github.com/VowpalWabbit/vowpal_wabbit/blob/e3f4038dbfb890058ff128d69a44ffec7c251236/vowpalwabbit/multiclass.cc#L82 ```words[0]``` contain the whole database as a string and ```endptr``` only return after encountering the first number. 

Should I just check if the string before `|` is a number by using a regular expression or there is some better way?

jackgerrits:
You don't need to do pre-validation. The strtol function return the position in the string that it consumed up to. You will need to modify things so we are able to check if the entire string was consumed or not.

> words[0] contain the whole database as a string and endptr only return after encountering the first number.

Exactly, use the fact that endptr is set after finishing the number conversion.

Shivanshmundra:
@jackgerrits I've changed code to check whether the whole token is consumed or not.
See if this looks fine to you.

vowpalwabbit/parse_primitives.h
```diff
@@ -160,3 +160,22 @@ inline int int_of_string(VW::string_view s)
...
+  if (end <= s.begin() && s.size() > 0)
+  {
+    std::cout << "warning: " << s << " is not a good int, replacing with 0"
+              << std::endl;
+    i = 0;
+  }
+
+  // checking if whole string is consumed
+  if (s.end() != end)
+    return -1;
```
jackgerrits: If the user inputted a "-1" how do you distinguish between invalid and valid input?

vowpalwabbit/parse_primitives.h
```diff
@@ -160,3 +160,22 @@ inline int int_of_string(VW::string_view s)
 
   return i;
 }
+
+inline int int_of_string_ml(VW::string_view s)
```
jackgerrits: What does `ml` mean?

vowpalwabbit/parse_primitives.h
Shivanshmundra: Yes, I thought about it but then -1 won't be a label in general. I could resolve this by raising error in the `int_of_string_ml` itself.

vowpalwabbit/parse_primitives.h
Shivanshmundra: `ml` in `int_of_string_ml` represents multiclass here. As multilabel can have multiple labels, so the error should only invoke in multilabel one. Hence I made another function to get int from the string.

Shivanshmundra:
I changed the name for better interpretability and improved error checking approach to solve the valid and invalid input issues.

Shivanshmundra:
@jackgerrits I was looking at why test cases have been failing from this addition. Actually the issue is with `named_labels`. For example,
 ```
det | the 
noun | monster
```
here label is not an integer but a name. Before this commit, it was giving a warning message ` **label** is not a good float, replacing with 0` which is not correct as it should assign each named label to corresponding integer label.

Shivanshmundra:
@jackgerrits I have rectified all errors. Please have a look.

jackgerrits:
There should not be an implementation of int_of_string that is specialized to this use case. Can you please make it generic through an optional out parameter of the end of the consumed string? That callers can optionally determine whether the whole string was consumed.

jackgerrits:
You have not addressed the concern:
> Can you please make it generic through an optional out parameter of the end of the consumed string?

You have added a bool to check whether the whole token was consumed or not. However, we don't want this function to throw. Please allow the caller to optionally retrieve the endptr that was determined by strtol.

Shivanshmundra:
@jackgerrits Sorry for not getting it right. I still have some confusion regarding your objective.
If I am correct, you want function to return `endptr` of consumed string optionally?

jackgerrits:
Yes, the caller of `int_of_string` should have some mechanism to inspect the value of `endptr` that results from `strtol`

Shivanshmundra:
@jackgerrits One mechanism which I could think of is: 
Since we can't return multiple things in C++ function, declaring a char null pointer before calling function and passing it by reference. It would then store endptr value and then we can do whatever with that ptr.
But this would force to define char variable at every place where `int_of_string` is called which would not be that good idea. Do you have anything good in your mind?

jackgerrits:
You can use an overload to make it optional?

jackgerrits:
Hi @Shivanshmundra. This has been dormant for a while, are you still interested in working on this? If I don't hear from you I am going to go ahead and close it in 7 days.

Shivanshmundra:
Hi @jackgerrits, Actually I was busy in my internship work. Sorry for keeping it dormant.
I couldn't get how to return multiple values in that function. If you can show me any such example which has a similar use case like that of what we are talking about here, I can give a shot.

jackgerrits:
You can effectively return multiple values in C++ using out parameters, references or pointers.

PR #2307, opened 2020-03-01
added  repr in BaseEstimator from sklearn

RituRajSingh878:
Added repr from sklearn in VW. Later, I will use it with` ipython/Jupiter notebook` to output it as sklearn do for its model.

```julia
In [19]: model = LogisticRegression()                                                                    

In [20]: model.fit(data.x, data.y)                                                                       
Out[20]: 
LogisticRegression(C=1.0, class_weight=None, dual=False, fit_intercept=True,
                   intercept_scaling=1, l1_ratio=None, max_iter=100,
                   multi_class='warn', n_jobs=None, penalty='l2',
                   random_state=None, solver='warn', tol=0.0001, verbose=0,
                   warm_start=False)
```

So by doing this, user will get more information about the model.

### DIFF
--- a/python/vowpalwabbit/sklearn_vw.py
+++ b/python/vowpalwabbit/sklearn_vw.py
@@ -12,6 +12,7 @@ from sklearn.base import BaseEstimator, RegressorMixin
 from sklearn.linear_model.base import LinearClassifierMixin, SparseCoefMixin
 from sklearn.datasets import dump_svmlight_file
 from sklearn.utils import shuffle
+from sklearn.utils._pprint import _EstimatorPrettyPrinter
 from vowpalwabbit import pyvw
 
 DEFAULT_NS = ''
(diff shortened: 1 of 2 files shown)


python/vowpalwabbit/sklearn_vw.py
```diff
@@ -379,8 +380,52 @@ def __str__(self):
...
+                # categoric...
+                # handle_unknown='ignore',
+                # so we add [^\n]*\n which matches until the next \n
+                regex += r'[^\n]*\n'
+                right_lim = re.match(regex, repr_[::-1]).end()
+
+            ellipsis = '...'
+            if left_lim + len(ellipsis) < len(repr_) - right_lim:
+                # Only add ellipsis if it results in a shorter repr
+                repr_ = repr_[:left_lim] + '...' + repr_[-right_lim:]
```
thealphadollar: Hey @RituRajSingh878 

This Pull Request looks really good to me. One thing I'm not sure about is that do you think we need so many comments? Most of them seem redundant to me and could be skipped.

Here is a good resource to consult on the same: https://www.freecodecamp.org/news/code-comments-the-good-the-bad-and-the-ugly-be9cc65fbf83/

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -12,6 +12,7 @@
 from sklearn.linear_model.base import LinearClassifierMixin, SparseCoefMixin
 from sklearn.datasets import dump_svmlight_file
 from sklearn.utils import shuffle
+from sklearn.utils._pprint import _EstimatorPrettyPrinter
```
thealphadollar: Also, I believe `_pprint` is meant to be a private method and should not be imported as such. I think @jackgerrits would be a better person to comment than me though.

python/vowpalwabbit/sklearn_vw.py
RituRajSingh878: thanks for the suggestions, I will check it out and will remove the unnecessary comments.

python/vowpalwabbit/sklearn_vw.py
RituRajSingh878: I think you are right that's why it is failing on Travis. But it is working on local-

```julia
In [5]: model.fit(data.x, data.y)                                                                        
Out[5]: VWClassifier(convert_to_vw=True, loss_function='logistic', passes=1, quiet=True)
```

thealphadollar: I believe it works because of the difference between the Python version.

jackgerrits: We should not depend on private methods in sklearn, they can change between versions which is exactly the issue you are seeing.

RituRajSingh878: Okay, got it. So I think I have to implement something like https://github.com/scikit-learn/scikit-learn/pull/11705 in VW and then I can use it from VW itself.

lalo:
@RituRajSingh878 any idea on the test failures on CI of this PR?

RituRajSingh878:
@jackgerrits  Can I copy the utils code of sklearn into VW directly?

jackgerrits:
We would like to avoid copying sklearn code into our branch

RituRajSingh878:
@jackgerrits We can go with simple repr for now. I will update it later with full implementation.

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -482,7 +482,10 @@ def __str__(self):
             return str(self.params)
 
     def __repr__(self):
-        return self.__str__()
+        if self.params is not None:
+            repr_items_ = dict(sorted(self.params.items()))
+            repr_items_ = str(repr_items_).replace("{","").replace("}", "")
+            return "%s(%s)" % (self.__class__.__name__, repr_items_)
```
jackgerrits: Please use a {} based format string

RituRajSingh878:
the test is failing due to the python version as the order of` dict` is not fixed in different python versions (due to hashing).  I can write the code to fix this(which is not required for this) so I am asking here that should I fix this or leave it as it is and remove the test case because fixing is not required for this.

jackgerrits:
`__repr__` should be consistent. Can you please find a way to make it reproducible between versions?

RituRajSingh878:
@jackgerrits please review it.

PR #2311, opened 2020-03-03
Skip Malformed Examples 

### LINKED ISSUE #2255: Parser should skip examples in cmdline where it encountered an error
When parsing, if an error is encountered then there is a message printed out and VW actually continues on. It is treated like a warning. This means that a malformed example is used as part of learning/inference.
If you wish to fail on malformed examples you can pass `--strict_parse`. This will cause the warning to become a throw and VW will exit on malformed examples. This is useful but not the whole solution.
 
I believe VW’s normal parsing mode for command line needs to be updated, so that if an error is encountered VW actually skips that entire example and moves on. Then in strict mode, instead of skipping it exits the entire program.

If using this in library mode then the error needs to propagate out of the call and be handled by the library user.

TheLethalCode:
While parsing examples, by default, when the parser encounters examples that are malformed, it just prints out a warning and continues using it for further learning. This PR, makes the parser skip malformed examples by default, by adding a variable to monitor the status of the example.

Fixes #2255 .


jackgerrits:
I believe the decision was the we would not need to add another mode, we would change the default to be skip entire example on malformed. This is because most people we asked actually already assumed this to be the case.

TheLethalCode:
@peterychang According to #2255 , what needed to be changed was the default behavior, as using malformed examples for learning doesn't make a lot of sense. Though I could try adding a separate mode, where the parser does what the default behavior does now.
As for the rest, I will try to incorporate the changes into all the workflows after taking care of the issue of multiline examples.

vowpalwabbit/parse_example.cc
```diff
@@ -538,14 +546,20 @@ void read_lines(vw* all, char* line, size_t /*len*/, v_array<example*>& examples
...
+  for (size_t i = 0, j=0; i < lines.size(); i++)
   {
     // Check if a new empty example needs to be added.
-    if (examples.size() < i + 1)
+    while (examples.size() < j + 1)
     {
       examples.push_back(&VW::get_unused_example(all));
     }
-    read_line(*all, examples[i], lines[i]);
+    read_line(*all, examples[j], lines[i]);
```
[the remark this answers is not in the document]
TheLethalCode: Understood.

peterychang:
@TheLethalCode Apologies, my information was outdated. It sounds like a new decision to change the default behavior was made

TheLethalCode:
@peterychang I went through the different workflows, and found that `read_features_string`, `read_features_json` and `line_to_examples_json` ultimately depend on the function `substring_to_example` to convert the string_view to actual examples, which is where the actual detection of whether an example is malformed or not takes place. Other than these, I found another workflow `read_cached_features`, where the situation of malformed examples doesn't arise (or it handles it by itself, because it doesn't throw an exception on `--strict_parse` anywhere as far as I could see). 
I have also modified the edits such that multiline examples are discarded if any of the lines are malformed, and single line examples are discarded if they are found to be malformed.

TheLethalCode:
@peterychang @TheLethalCode any suggestions?

peterychang:
>  read_features_string, read_features_json and line_to_examples_json ultimately depend on the function substring_to_example to convert the string_view to actual examples, which is where the actual detection of whether an example is malformed or not takes place

This is actually not the case. I can see why this was confusing though; `substring_to_example` attempts to call the VW text parser, `TC_parser`. But that object is only used in `read_features_string` (see the outer if condition). In the json workflow, `substring_to_example` is called with an empty string. Essentially, that just creates a default label (The function probably shouldn't even be called in the JSON parser). The actual line parsing uses a totally different workflow. [This context](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/parse_example_json.h#L1257) is the parser's state machine

Within the json parser, there are actually 2 distinct workflows, standard JSON and DecisionServiceJson(DSJson). The main difference between these two is the starting state the JSON parser begins in.
Standard JSON: DefaultState
DSJson: DecisionServiceState

TheLethalCode:
If that's the case, then this implies that there doesn't arise a case where the JSON parser throws an exception on `--strict_parse` (I might be wrong), because as far as I can see only the `TC_parser` does that. So does that mean json examples cannot be malformed?

peterychang:
I believe the json parser always throws an exception on malformed data (regardless of the `--strict_parse` flag. I know the state machine returns a nullptr in the case of badly formatted inputs, but I'm not sure what happens after that. I _think_ it eventually ends up here.
https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/parse_example_json.h#L1461

I'm not particularly familiar with the json parser though, so you should dig into the code there to verify.

TheLethalCode:
Okay, sure. I'll look into it. Thanks!

TheLethalCode:
> In the json workflow, `substring_to_example` is called with an empty string

Sorry, my mistake. I failed to see that the function is called upon an empty string. `substring_to_example` has little to do with json parser.


> I believe the json parser always throws an exception on malformed data (regardless of the `--strict_parse` flag. I know the state machine returns a nullptr in the case of badly formatted inputs, but I'm not sure what happens after that

Yes, that seems to be the case. When the parser encounters malformed data, it sets the corresponding error state, and before returning the example, it checks for the error state and if there is an error, it throws an exception.

TheLethalCode:
But, the main point of the issue is to ignore malformed data instead of including them in the learning process. I believe this PR addresses that.

peterychang:
There are 2 different cases we're handling here. The default behavior and the `--strict_parse` behavior. Under `--strict_parse`, the process will end if a malformed example is encountered. Under the default behavior, it should simply skip the malformed example (possibly after logging) and continue processing.
I believe throwing an exception in the json parser will bring the process down.

On another note, this is still leaking memory, as mentioned in a previous comment

TheLethalCode:
I understood that point, but since the default behavior of the JSON parser is throwing an exception unlike the text parser where just a warning is given out, I assumed that was how it was supposed to behave. I thought there might be a reason for excluding `--strict_parse` from all other workflows other than the text parser. So, do I have to incorporate it into the JSON parser as well?

Moreover, I have removed that part of code which was leaking memory which you pointed out previously. Are there any other cases?

vowpalwabbit/learner.cc
```diff
@@ -207,16 +209,24 @@ class multi_example_handler
...
+    if(ec->malformed)
+    {
+      malformed = true;
+    }
     if (try_complete_multi_ex(ec))
     {
-      _context.template process<multi_ex, learn_multi_ex>(ec_seq);
+      if(!malformed)
+      {
+        _context.template process<multi_ex, learn_multi_ex>(ec_seq);
```
peterychang: malformed examples will never be processed (which is what we want). But a part of the process is calling `finish_example` which, among other things, returns the example back to the pool for reuse. We need to return the discarded examples back to the pool

peterychang:
@jackgerrits @lokitoth Do we want to change the default behavior of the json parser? I think we did, but I wasn't in the last couple of conversations regarding the requirements here

jackgerrits:
Perhaps we do. I think we ultimately want to aim for consistency across all parsers. By default they skip examples/mult_ex that are malformed, and in strict_parse mode they will exit. Currently the JSON parser throws on parse errors determined by RapidJson, as well as our own errors based on incorrect user input. The reason why the default is to throw is because RapidJson throws on malformed input.

vowpalwabbit/learner.cc
TheLethalCode: Okay, understood. I will take care of it.

TheLethalCode:
So, I would like to confirm. We need the default behavior of all workflows to ignore malformed examples and on strict_parse we need them to exit. But, then what about the throws by RapidJson? Should we leave it as it is now and only handle errors determined by us in the JSON parser?

peterychang:
The behavior of all parsers should be the same. So we will need to change the default behavior for the JSON parser.

For exceptions thrown by the JSON parser, I think the best you can do is catch the exception and conditionally rethrow it. It will slow the process down quite a bit, but I think the only options rapidjson gives for error handling is throwing an exception and aborting.

TheLethalCode:
I get the idea. Making the parser uniform across all workflows does make sense. But already parsing is the bottleneck, wouldn't this make the situation even worse? I understand this only acts up in the presence of malformed inputs, but in my opinion these are not uncommon situations.

peterychang:
If the requirements are to continue processing data in the event of a malformed error, then the performance impact of catching exceptions from rapidjson are unavoidable. If a client would rather stop the process in the event of a malformed example (as what happens now), they have the option by setting the strict_parse flag. Either way, this change simply gives a user the option of continuing the parsing (with some perf impact).

TheLethalCode:
Okay, understood. I will start working on it.

jackgerrits:
Hi @TheLethalCode. This has been dormant for a while, are you still interested in working on this? If I don't hear from you I am going to go ahead and close it in 7 days.

PR #2313, opened 2020-03-03
added modified_hubber loss

RituRajSingh878:
Added` modified hubber` loss in loss_function to extend the support of loss function.
But I don't understand few things from loss function so anyone can help with this.

`hubber loss` -
https://en.wikipedia.org/wiki/Huber_loss

I will add the test case in later commits.
Thanks


vowpalwabbit/loss_functions.cc
```diff
@@ -358,6 +358,58 @@ class poisson_loss : public loss_function
...
+      std::cout << "You are using label " << label << " not -1 or 1 as loss function expects!" << std::endl;
+     float e = max(0, 1-prediction*label);
+    return (prediction*label >= -1.f) ? e*e : -4*label*prediction;
+  }
+
+  float getUpdate(float prediction, float label, float update_scale, float pred_per_update)
+  {
+
+  }
+
```
RituRajSingh878: I don't understand how updates work so anyone can explain then I will update it accordingly.

vowpalwabbit/loss_functions.cc
```diff
@@ -358,6 +358,58 @@ class poisson_loss : public loss_function
   }
 };
 
+class modified_hubber_loss : public loss_function
```
gramhagen: typo: huber

vowpalwabbit/loss_functions.cc
```diff
@@ -358,6 +358,58 @@ class poisson_loss : public loss_function
...
+{
+ public:
+  std::string getType() { return "modified_hubber"; }
```
gramhagen: typo huber

vowpalwabbit/loss_functions.cc
gramhagen: i believe the update functions represent how to update the parameters.
e.g. squared loss is (label - pred)^2, and the parameter (weight) update is 2*(label-pred)

@sharatsc can you comment on the other functions, particularly the safe update?

RituRajSingh878: @gramhagen , @sharatsc 
I am implementing this loss so that we can use it for `SGD` in core VW. As [sklearn `SGD`](https://github.com/scikit-learn/scikit-learn/blob/95d4f0841d57e8b5f6b2a570312e9d832e69debc/sklearn/linear_model/_stochastic_gradient.py#L1071)  has this loss also so I think we can implement it for the VW.

JohnLangford:
I believe Huber loss is effectively already there in VW.

Predictions are clipped to the observed label range here: https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/scorer.cc#L20 .   Given that clipping, the gradients of particularly large or particularly small predictions become constant, corresponding to a linear loss function far from the operating range, just as in Huber loss.  

Given that we are effectively doing Huber loss already, I'm hesitant about incorporating this patch.

jackgerrits:
@RituRajSingh878, based on John's comments it seems as though this is already covered in the library. I'm going to close this. If you disagree, please feel free to reopen.

Issue #2317, opened 2020-03-04
Memory leak in vw_dll_parse_escaped unit test


jackgerrits:
This unit test ensures that an exception is thrown with this line:
```C++
  BOOST_CHECK_THROW(VW_InitializeA("-d test\\ --nonexistent_option --quiet"), VW::vw_unrecognised_option_exception);
```

However, it seems that this operation is not exception safe causing Valgrind to report leaks in this test.

Once this is fixed we should turn on unit tests in the Valgrind CI.

Related to #2122

lokitoth:
I am concerned that we are throwing an exception across an `extern "C"` boundary. 

Is it even valid to do this?

This may not just be a test-issue.

jackgerrits:
Yep it's a **big** issue with the C API right now. But we can't fix it without breaking it because we don't actually have a mechanism to communicate errors. It is valid to throw in this case IFF the function is used in a C++ program. If this were linked and called as a C function it would bring down the program (which we have seen reports of before: #2121).

jackgerrits:
Technically we could mitigate this function with nullptr - which we probably should do. But there are plenty of other functions in the C API which cannot communicate errors because of `void` return values.

PR #2318, opened 2020-03-04
Fix most Valgrind errors in unit tests

jackgerrits:
There is one test still with issues and this is due to exception safety. I've created an issue for that here: #2317


test/unit_test/options_boost_po_test.cc
```diff
@@ -41,7 +41,7 @@ BOOST_AUTO_TEST_CASE(typed_options_parsing) {
   char command_line[] = "exe --str_opt test_str -i 5 --bool_opt yes --char_opt f --float_opt 4.3";
   int argc;
   // Only the returned char* needs to be deleted as the individual pointers simply point into command_line.
-  std::unique_ptr<char*> argv(convert_to_command_args(command_line, argc));
+  std::unique_ptr<char*[]> argv(convert_to_command_args(command_line, argc));
```
lokitoth: It worries me that the type system did not catch this issue. Is there something we can do going forward to avoid cases like this?

jackgerrits: It didn't catch it because `convert_to_command_args` returns `char**`, so still a failure of the type system. Or rather, caused by C++'s eager implicit conversion semantics... Not using raw pointers and raw pointers to raw pointers is a good solution for avoiding things like this. Argv is a bit of a special case though

jackgerrits: By the way this caused a mismatched delete/delete[], it least it wasn't a leak!

PR #2321, opened 2020-03-04
Implement Slates label parser

jackgerrits:
- Implements new label type for slates
- Implements label parser for VW text format
- Adds tests for new label parser
- Generify check_collections testing utility, so that it can compare both std::vector and v_array


rajan-chari:
Is there a description of the format somewhere on the wiki perhaps?  It would be good to be able to refer to it.  Please add a link in the parser itself.

jackgerrits:
Here is the wiki page: https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Slates
I still need to add an example for the DSJSON format.

test/unit_test/ccb_test.cc
```diff
@@ -24,7 +24,7 @@ BOOST_AUTO_TEST_CASE(ccb_generate_interactions)
   CCB::calculate_and_insert_interactions(shared_ex, actions, interactions);
   std::sort(compare_set.begin(), compare_set.end());
   std::sort(interactions.begin(), interactions.end());
-  check_vectors(interactions, compare_set);
+  check_collections(interactions, compare_set);
```
peterychang: since we have a `check_collections_with_tolerance`, maybe rename to `check_collections_exact` for clarity?

vowpalwabbit/slates_label.cc
```diff
@@ -0,0 +1,208 @@
...
+#include "parser.h"
+#include "vw_string_view.h"
+
+#include <numeric>
+
+namespace slates
+{
+void default_label(void* v);
+
+#define READ_CACHED_VALUE(DEST, TYPE)                            \
```
peterychang: Is there a reason these are macros and not inlined functions? Because of the early return?

jackgerrits: Yep the early return was the main reason. Then for the other one, since it is so small, I thought keeping it consistent with this macro made sense

jackgerrits: Yep the early return was the main reason. Then for the other one, since it is so small, I thought keeping it consistent with this macro made sense

vowpalwabbit/slates_label.cc
```diff
@@ -0,0 +1,208 @@
...
+
+  if (words.size() < 2)
+    THROW("slates labels may not be empty");
+  if (!(words[0] == "slates"))
+  {
+    THROW("slates labels require the first word to be slates");
+  }
+
+  const auto& type = words[1];
+  if (type == "shared")
```
peterychang: can we move these magic strings into a single place.. like constants.h or something? I think "shared" is used in CCB and here now, so its probably worth doing

vowpalwabbit/slates_label.cc
```diff
@@ -0,0 +1,208 @@
...
+  ldDst.slot_id = ldSrc.slot_id;
+  copy_array(ldDst.probabilities, ldSrc.probabilities);
+}
+
+void parse_label(parser* p, shared_data*, void* v, v_array<VW::string_view>& words)
+{
+  auto& ld = static_cast<polylabel*>(v)->slates;
+  ld.weight = 1;
+
+  if (words.size() < 2)
```
peterychang: It looks like the only valid size of words is 2 or 3, is that correct? Can we make that check here instead of inside all the other if statements?

test/unit_test/test_common.h
```diff
@@ -5,16 +5,31 @@
 
 #include <vector>
 
+#include "action_score.h"
+
 constexpr float FLOAT_TOL = 0.0001f;
```
lokitoth: [Minor] Should this be a default value for `float_tolerance` on `check_collections_with_tol`?

vowpalwabbit/slates_label.cc
jackgerrits: Keeping the check inside the inner statements allows us to have a better error message based on which label type is being parsed

vowpalwabbit/slates_label.cc
```diff
@@ -0,0 +1,208 @@
...
+    THROW("slates labels may not be empty");
```
lokitoth: [Minor] Consider also adding something to explain the `words.size() == 1` case, as otherwise people who add "slates" but nothing else would get confused.

vowpalwabbit/slates_label.cc
```diff
@@ -0,0 +1,209 @@
...
+
+#define READ_CACHED_VALUE(DEST, TYPE)                            \
+  next_read_size = sizeof(TYPE);                                 \
+  if (cache.buf_read(read_ptr, next_read_size) < next_read_size) \
+    return 0;                                                    \
+  DEST = *(TYPE*)read_ptr;                                       \
+  read_count += sizeof(TYPE);
+
+#define WRITE_CACHED_VALUE(VALUE, TYPE)           \
+  *(TYPE*)c = VALUE; \
```
lokitoth: I worry a bit these macros require specific stack state that is configured outside of them (and uses slightly different names for things: `c` vs `read_ptr`). Consider creating a pair of macros that can be used to set it up, e.g. `READ_CACHE(cache) / WRITE_CACHE(cache, size)`.

vowpalwabbit/slates_label.cc
```diff
@@ -0,0 +1,209 @@
...
+
+      // If a full distribution has been given, check if it sums to 1, otherwise throw.
+      if (ld.probabilities.size() > 1)
+      {
+        float total_pred = std::accumulate(ld.probabilities.begin(), ld.probabilities.end(), 0.f,
+            [](float result_so_far, const ACTION_SCORE::action_score& action_pred) {
+              return result_so_far + action_pred.score;
+            });
+
+        // TODO do a proper comparison here.
```
lokitoth: You left yourself a TODO here

lokitoth: Also, should we normalize in these cases?

vowpalwabbit/slates_label.cc
```diff
@@ -0,0 +1,209 @@
...
+        if (total_pred > 1.1f || total_pred < 0.9f)
+        {
+          THROW("When providing all predicition probabilties they must add up to 1.f");
+        }
+      }
+    }
+    else if (words.size() != 2)
```
lokitoth: Did you mean `== 2` or `< 3` here?

vowpalwabbit/slates_label.cc
```diff
@@ -0,0 +1,209 @@
...
+    {
+      THROW("Slates shared labels must be of the form: slates slot [chosen_action_id:probability[,action_id:probability...]]");
+    }
+    ld.type = example_type::slot;
+  }
+  else
+  {
+    THROW("unknown label type: " << type);
```
lokitoth: [Minor] For consistency, consider "unknown slates label type" in parallel to "Slates [type] label"
[Nit] Consider unifying capitalization across the error strings

vowpalwabbit/slates_label.cc
```diff
@@ -0,0 +1,209 @@
...
+  ldDst.weight = ldSrc.weight;
+  ldDst.labeled = ldSrc.labeled;
+  ldDst.cost = ldSrc.cost;
+  ldDst.slot_id = ldSrc.slot_id;
+  copy_array(ldDst.probabilities, ldSrc.probabilities);
+}
+
+void parse_label(parser* p, shared_data*, void* v, v_array<VW::string_view>& words)
+{
+  auto& ld = static_cast<polylabel*>(v)->slates;
```
lokitoth: Consider adding a comment describing the grammar of the label

vowpalwabbit/slates_label.h
```diff
@@ -0,0 +1,43 @@
...
+  action = 2,
+  slot = 3
+};
+
+struct label
+{
+  // General data
+  example_type type;
+  float weight;
+  bool labeled;
```
lokitoth: Could you add a comment describing what this means?

vowpalwabbit/constant.h
```diff
@@ -34,3 +34,9 @@ constexpr unsigned char ccb_id_namespace = 140;
 typedef float weight;
 
 constexpr size_t NUM_NAMESPACES = 256;
+
+constexpr const char* CCB_LABEL = "ccb";
+constexpr const char* SLATES_LABEL = "slates";
+constexpr const char* SHARED_TYPE = "shared";
+constexpr const char* ACTION_TYPE = "action";
+constexpr const char* SLOT_TYPE = "slot";
```
lokitoth: [Nit] GitHub is complaining about no newline at EOF.

vowpalwabbit/slates_label.cc
jackgerrits: Yeah I understand that. But given they are defined in a cc file, I am not worried about them leaking and they are just used to not repeat myself (and cause bugs, like I did when I first did this -_-)

vowpalwabbit/slates_label.cc
jackgerrits: That's a copy paste from the CCB code that does this too

vowpalwabbit/slates_label.cc
jackgerrits: The slot label can have either 2 or 3 tokens. Since at this point we know it doesn't have 3 because of the if statement, and outer checks mean it doesnt have 0 or 1. So the check should probably be > 3

PR #2323, opened 2020-03-05
[Feature: label split] Support learn/predict with label split out of example

jackgerrits:
Reopening this PR against the feature branch. See #2314 for previous comments.

This PR brings in support for a reduction to use a new signature for the learn/predict function:
```C++
void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples, std::vector<CCB::label>& labels)
```
As well as to call the next reduction by passing a reference to labels instead of using the union.
```C++
base.learn_with_label(ec, simple_label);
```
Migrating reductions can be done one at a time because of adapting machinery that is in place in `learner.h`. This involves converting between union based and stack based label passing. There is a performance cost to this, ranging from ~0% to ~13%, depending on the relative cost/frequency of calling learn/predict compared to other operations. In CB workloads I see a hit of ~1.5% when using cache files. This cost is **temporary** and introduced by the adapting machinery. Once we get to stack based label passing in the entire system we will be able to remove the machinery and remove that cost. All in all I think this is something we have been striving for for a long time and very worth this temporary cost. 

Other changes:
- Label type specification has moved to the learner and out of all to allow for each learner to have its own specification
- The binary reduction has been migrated to stack based as a proof of concept
- The CCB reduction has been migrated to stack based as a proof of concept

Note:
- Multipredict is not yet implemented, but I don't think that is a blocker here
- Label as an input to learn/predict should be marked const but that cannot be done until the machinery is done

Tracking issue: #2324


rajan-chari:
Somewhat of a general comment: predict() should not be aware of labels only learn().  So logically we want something like learn(features,label) and prediction = predict(features)

jackgerrits:
We can't do that until #2207 is done though?

vowpalwabbit/learner.h
```diff
@@ -56,13 +56,19 @@ inline func_data tuple_dbf(void* data, base_learner* base, void (*func)(void*))
...
   using multi_fn = void (*)(void* data, base_learner& base, void* ex, size_t count, size_t step, polyprediction* pred,
       bool finalize_predictions);
 
   void* data;
   base_learner* base;
   fn learn_f;
   fn predict_f;
+
+  fn_with_label learn_with_label_f;
+  fn_with_label predict_with_label_f;
```
rajan-chari: This is not making sense to me.  predict should not need label.

jackgerrits: Eventually yes, but this isn't possible until #2207 is complete

rajan-chari: Perhaps we should keep this change forward looking.  i.e. if a predict() needs access to label, it can get it from example.  Maybe even introduce some encapsulation in example to get the label.  Eventually label will disappear from example and all reductions will have to conform to this.   

---
In reply to: [388969924](https://github.com/VowpalWabbit/vowpal_wabbit/pull/2323#discussion_r388969924) [](ancestors = 388969924)

jackgerrits: To have different machinery for predict and learn it pretty difficult especially since most reductions right now use an templated function for both. I see what you're saying but this project is already massive and I really don't want to take on more responsibilities for it. Especially since removing labels from predict is orthogonal to just splitting out the object in the signature.

vowpalwabbit/learner.h
```diff
@@ -128,6 +134,173 @@ inline void decrement_offset(multi_ex& ec_seq, const size_t increment, const siz
...
+    increment_offset(ec, increment, i);                                                                                \
+    if (learn_fd.learn_with_label_f != nullptr)                                                                        \
+    {                                                                                                                  \
+      learn_fd.learn_with_label_f(learn_fd.data, *learn_fd.base, static_cast<void*>(&ec), static_cast<void*>(&label)); \
+    }                                                                                                                  \
+    else                                                                                                               \
+    {                                                                                                                  \
+      polylabel saved = std::move(ec.l);                                                                               \
+      ec.l.FIELD_NAME = std::move(label);                                                                              \
```
rajan-chari: >FIELD_NAME [](start = 11, length = 10)

These macros mostly seem to have to do with the fact that the name of the variable in label is tied to the label type.  Is that right?
If that is the case, I am wondering if this can be accomplished through some form of template specialization around just this bit of code

jackgerrits: The union must be accessed using the named field, as reinterpret casting is breaks if it becomes a struct. I'd like to avoid that debt as I'm all too familiar with it. (Even though this code will be removed by then). Given that you'd need to overload on every type and at that point your back to the same amount of duplication here

vowpalwabbit/learner.h
```diff
@@ -128,6 +135,174 @@ inline void decrement_offset(multi_ex& ec_seq, const size_t increment, const siz
...
+      const auto& example = ec[i];                                                                                   \
+      example->l.FIELD_NAME = std::move(saved[i]);                                                                   \
+    }                                                                                                                \
+  }                                                                                                                  \
+  break;
+
+#define TYPED_EXAMPLE_PREDICT_CASE(TYPE, FIELD_NAME)                                                                   \
+  case label_type_t::FIELD_NAME:                                                                                       \
+  {                                                                                                                    \
+    TYPE saved = std::move(ec.l.FIELD_NAME);                                                                           \
```
rajan-chari: >move [](start = 22, length = 4)

Why is this needed during a predict?

jackgerrits: In the current state of the codebase label information is required in predict. This will be tackled in this feature branch before we are done.

vowpalwabbit/learner.h
```diff
@@ -128,6 +135,174 @@ inline void decrement_offset(multi_ex& ec_seq, const size_t increment, const siz
...
+#define DEFINE_TYPED_PREDICT_AND_LEARN_FUNCS(TYPE, FIELD_NAME) \
+  TYPED_EXAMPLE_LEARN_FN(TYPE, FIELD_NAME)                     \
+  TYPED_EXAMPLE_PREDICT_FN(TYPE, FIELD_NAME)                   \
+  TYPED_MULTI_EXAMPLE_LEARN_FN(TYPE, FIELD_NAME)               \
+  TYPED_MULTI_EXAMPLE_PREDICT_FN(TYPE, FIELD_NAME)
+
+#define TYPED_EXAMPLE_LEARN_CASE(TYPE, FIELD_NAME)                                                                   \
+  case label_type_t::FIELD_NAME:                                                                                     \
+  {                                                                                                                  \
+    TYPE saved = std::move(ec.l.FIELD_NAME);                                                                         \
```
rajan-chari: >TYPE saved = std::move(ec.l.FIELD_NAME) [](start = 4, length = 39)

what's the expectation with this code?

for example if label is a vector, will it pass an empty vector to learn?

jackgerrits: We will be passing a reference to the object that was in `FIELD_NAME` of the polylabel. It must  be stored in a temporary object because we must be able to restore state and the underlying reduction may overwrite the union

rajan-chari: If the reduction does not use the label but calls base.learn(), would this work correctly?

---
In reply to: [390379556](https://github.com/VowpalWabbit/vowpal_wabbit/pull/2323#discussion_r390379556) [](ancestors = 390379556)

jackgerrits: In that case the reduction must forward the label to its own learn call

rajan-chari: So once a reduction starts using label, all reductions invoked by it have to use label and so on down the chain?

---
In reply to: [390393147](https://github.com/VowpalWabbit/vowpal_wabbit/pull/2323#discussion_r390393147) [](ancestors = 390393147)

jackgerrits: Once we are done with this overall feature, every learn function must have a label as input and must call learn with a label.

As of right now, if a reduction opts in to the new signature they must accept a label as argument and call base.learn with a label. If they have not opted in yet they do not need to.

> So once a reduction starts using label, all reductions invoked by it have to use label and so on down the chain?
No, each reduction is independent on if they use or old style and it doesn't enforce things down the chain at the moment.

rajan-chari: I was trying to think about the logical downstream implications of modifying ec.l.  Since not all reductions know about label.   I am wondering if its better to not modify ec.l.

---
In reply to: [390379556](https://github.com/VowpalWabbit/vowpal_wabbit/pull/2323#discussion_r390379556) [](ancestors = 390379556)

jackgerrits: So the point of these macros (and this overall project) is to enable us to not worry about downstream implications. From practice I have not yet found a reduction that this does not work for (I am through about 60% of them)

If you do not modify ec.l then there is no way for a new style reduction to call an old style reduction. Because old style is expecting ec.l to be a certain value where as new style does not touch ec.l at all by design.

This saving and restoring will **go away once the project is done,** I just need it to be able to make reasonable progress in the feature branch.

jackgerrits:
This is going to require so much refactoring that I'll just start fresh.

Issue #2322, opened 2020-03-05
cb_explore and cb_explore_adf are taking invalide epsilon arguments


adityauser:
#### Describe the bug
The command `vw -d train_adf.dat --cb_explore_adf  --epsilon -1` is running flawlessly but clearly it is invalid since the epsilon is beyond its bound 

#### To Reproduce
Run any of these commands:
`vw -d train_adf.dat --cb_explore_adf  --epsilon -1`
`vw -d train_adf.dat --cb_explore_adf  --epsilon 2`
`vw -d train.dat --cb_explore 4 --epsilon 4`
`vw -d train_adf.dat --cb_explore_adf --first 2 --epsilon -4`

#### Expected behaviour
Throw an invalid argument error.

#### Environment

`VW 8.8.0 on Linux with the Python bindings`

#### Status

I'm currently working on this bug

jackgerrits:
Yes, I agree epsilon should be between 0 and 1 inclusive.

mritunjaysharma394:
I have tried to fix this bug by prompting an error :
`ERROR: Invalid argument error, epsilon should be between 0 and 1 (both inclusive).`
and it is working for all the cases, except 
`vw -d train.dat --cb_explore 4 --epsilon 4`.
Maybe @jackgerrits you can help me out in cracking this case. Also, do we need to terminate the entire execution or just prompt an error?

jackgerrits:
Is `vw -d train.dat --cb_explore 4 --epsilon 4` valid input or invalid?

An invalid epsilon is fatal.

mritunjaysharma394:
In the fix, I tried to implement, I am producing this kind of output for all other invalid epsilon cases:

![Screenshot from 2020-03-06 00-46-53](https://user-images.githubusercontent.com/45476902/76017183-3338d980-5f44-11ea-9c0f-a3ae8b4db118.png)

Is this the right way? Do we have to use abort() also?
Also, I don't think `vw -d train.dat --cb_explore 4 --epsilon 4` is a valid input. Maybe @adityauser can clarify more on that?

mritunjaysharma394:
After following the suggested instructions, output for respective wrong entries of epsilon is something like this:

```
vw -d train.txt --cb_explore_adf --epsilon 2
Error: cb_explore_adf_greedy_epsilon must be in [0,1]

finished run
number of examples = 0
weighted example sum = 0.000000
weighted label sum = 0.000000
average loss = n.a.
[...]

PR #2325, opened 2020-03-06
Fix the issue #2322 to throw error for invalid epsilon argument passed.

mritunjaysharma394:
Fix the issue #2322 .
Implements producing an error for epsilon passed not in the range [0,1].

### DIFF
--- a/vowpalwabbit/cb_explore.cc
+++ b/vowpalwabbit/cb_explore.cc
@@ -302,6 +302,11 @@ base_learner* cb_explore_setup(options_i& options, vw& all)
     ss << data->cbcs.num_actions;
     options.insert("cb", ss.str());
   }
+  
+  if (data->epsilon < 0.0 || data->epsilon > 1.0)
+  {
+    THROW("cb_explore_epsilon must be in [0,1]");
+  }
 
   all.delete_prediction = delete_action_scores;
   data->cbcs.cb_type = CB_TYPE_DR;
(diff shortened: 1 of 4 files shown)


mritunjaysharma394:
Please @jackgerrits review the PR of the issue #2322 and also do tell what could be the reason behind Windows CI Build failing?

adityauser:
I don't think its the correct way to solve this issue.
You should check the epsilon conditions in these function: 
https://github.com/VowpalWabbit/vowpal_wabbit/blob/e1a05f483b0c9e35f6a039ed848433d24b6843c6/explore/explore_internal.h#L38
https://github.com/VowpalWabbit/vowpal_wabbit/blob/e1a05f483b0c9e35f6a039ed848433d24b6843c6/explore/explore_internal.h#L158
@jackgerrits what is your opinion?

jackgerrits:
Enforcing epsilon in those functions may be a good thing. But I think this is something that should be caught early, and we can do so while validating command line options.

vowpalwabbit/cb_explore.cc
```diff
@@ -302,6 +302,11 @@ base_learner* cb_explore_setup(options_i& options, vw& all)
     ss << data->cbcs.num_actions;
     options.insert("cb", ss.str());
   }
+  
+  if (data->epsilon < 0.0 || data->epsilon > 1.0)
+  {
+    THROW("cb_explore_epsilon must be in [0,1]");
```
jackgerrits: The use may not know what `cb_explore_epsilon` is, it doesn't correspond to an option they inputted. Can you make the error message clearer with that in mind?

jackgerrits:
The windows failure is a flaky test, I'll restart

vowpalwabbit/cb_explore.cc
mritunjaysharma394: Yes, I have modified the error messages. Please review them, thanks!

mritunjaysharma394:
I think now it looks good to merge. What say @jackgerrits ?

PR #2327, opened 2020-03-07
Fix clang tidy issues in CCB

jackgerrits:


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -270,7 +271,7 @@ void remove_slot_features(example* shared, example* slot)
       continue;
     }
 
-    else if (index == default_namespace)  // slot default namespace has a special namespace in shared
```
peterychang: I guess it doesn't matter either way, but why remove this?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -424,9 +434,13 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
     build_cb_example<is_learn>(data.cb_ex, slot, data);
 
     if (data.all->audit)
-      inject_slot_id<true>(data, data.shared, slot_id);
+    {
+      inject_slot_id<true>(data, data.shared, slot_id);\
```
peterychang: remove the \

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: The specific clang-tidy check here is to not follow a `continue` statement with an `else if` as it is kind of misleading

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: Oops! Thanks

jackgerrits: Oops! Thanks

PR #2330, opened 2020-03-08
Speed optimization by unrolling the loop into two loops

adityauser:
Earlier in every iteration bitwise `&` was being calculated but, after some iterations, the value of `fi` again starts repeating.
So, I counted the number of iteration after which the value of `fi` again starts repeating. And then unrolled the loop accordingly.

### DIFF
--- a/vowpalwabbit/gd.h
+++ b/vowpalwabbit/gd.h
@@ -53,11 +53,10 @@ inline void vec_add_multipredict(multipredict_info<T>& mp, const float fx, uint6
   {
     uint64_t count = 0;
     uint64_t copy_fi = (fi + (uint64_t)mp.step) & mask;
-    for (; copy_fi != fi && count < mp.count; copy_fi += (uint64_t)mp.step)
-    {
+    // number of iteration after which the value of fi again starts repeating
+    for (; copy_fi != fi && count < mp.count; copy_fi += (uint64_t)mp.step, ++count)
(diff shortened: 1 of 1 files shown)


vowpalwabbit/gd.h
```diff
@@ -49,12 +49,22 @@ inline void vec_add_multipredict(multipredict_info<T>& mp, const float fx, uint6
...
-    for (size_t c = 0; c < mp.count; ++c, fi += (uint64_t)mp.step, ++p)
+  else
+  {
+    uint64_t count = 0;
+    uint64_t copy_fi = (fi + (uint64_t)mp.step) & mask;
+    for (; copy_fi != fi && count < mp.count; copy_fi += (uint64_t)mp.step)
+    {
+      copy_fi &= mask;
+      ++count;
+    }
```
thealphadollar: Including line number 59 in line number 56 as below would be better since that is the convention being followed in neighbour for loops.

`for (; copy_fi != fi && count < mp.count; copy_fi += (uint64_t)mp.step, ++count)`

jackgerrits:
Do you have results on whether this improves performance? I am not familiar in this piece of code, @JohnLangford would you be able to take a quick look?

adityauser:
@jackgerrits I just cross-checked, my PR is not improving performance.
You can close the PR.

jackgerrits:
Okay no problem

PR #2332, opened 2020-03-10
added multiclassifier in vw_sklearn

RituRajSingh878:
I am starting to add the VWMultiClassifier code but I have many doubts so I am opening this pr in a draft.


RituRajSingh878:
During implementation, I have gone through the sklearn_vw code and I have few questions.
1- sklearn_vw is made to match the core VW classifier with sklearn. But in VWClassifier class  

https://github.com/VowpalWabbit/vowpal_wabbit/blob/bb68aec721dc95490a94f4b17f263a7c17295f0c/python/vowpalwabbit/sklearn_vw.py#L510

If I am creating its object and calling fit and then I am calling `predict` then it is calling this
https://github.com/VowpalWabbit/vowpal_wabbit/blob/bb68aec721dc95490a94f4b17f263a7c17295f0c/python/vowpalwabbit/sklearn_vw.py#L489
And this function is calling decision function which is this one-
https://github.com/VowpalWabbit/vowpal_wabbit/blob/bb68aec721dc95490a94f4b17f263a7c17295f0c/python/vowpalwabbit/sklearn_vw.py#L539 and again it is calling core VW to predict. 

So my question is, How we are using sklearn and its decision functions and other properties because it is using the core properties of VW.

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -560,6 +561,34 @@ class VWRegressor(VW, RegressorMixin):
...
+            Multi-class targets. An indicator matrix turns on multilabel
+            classification.
+
+        Returns
+        -------
+        self
+        """
+        pass
+
+
```
RituRajSingh878: Here, we have to pass an estimator to the `VWMultiClassifier` due to the `OneVsRestClassifier`.` OneVsRestClassifier` takes an estimator as an argument.So How should I override the decision function?

RituRajSingh878: the fit function of sklearn.
```julia

def fit(self, X, y):
--
  | """Fit underlying estimators.
  |  
  | Parameters
  | ----------
  | X : (sparse) array-like of shape (n_samples, n_features)
  | Data.
  |  
  | y : (sparse) array-like of shape (n_samples,) or (n_samples, n_classes)
  | Multi-class targets. An indicator matrix turns on multilabel
  | classification.
  |  
  | Returns
  | -------
  | self
  | """
  | # A sparse LabelBinarizer, with sparse_output=True, has been shown to
  | # outperform or match a dense label binarizer in all cases and has also
  | # resulted in less or equal memory consumption in the fit_ovr function
  | # overall.
  | self.label_binarizer_ = LabelBinarizer(sparse_output=True)
  | Y = self.label_binarizer_.fit_transform(y)
  | Y = Y.tocsc()
  | self.classes_ = self.label_binarizer_.classes_
  | columns = (col.toarray().ravel() for col in Y.T)
  | # In cases where individual estimators are very fast to train setting
  | # n_jobs > 1 in can results in slower performance due to the overhead
  | # of spawning threads.  See joblib issue #112.
  | self.estimators_ = Parallel(n_jobs=self.n_jobs)(delayed(_fit_binary)(
  | self.estimator, X, column, classes=[
  | "not %s" % self.label_binarizer_.classes_[i],
  | self.label_binarizer_.classes_[i]])
  | for i, column in enumerate(columns))
  |  
  | return self


```

RituRajSingh878:
I will add the test case later.

RituRajSingh878:
@gramhagen have a look.

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -8,8 +8,9 @@
 
 from scipy.sparse import csr_matrix
 from sklearn.exceptions import NotFittedError
-from sklearn.base import BaseEstimator, RegressorMixin
+from sklearn.base import BaseEstimator, RegressorMixin, ClassifierMixin
 from sklearn.linear_model.base import LinearClassifierMixin, SparseCoefMixin
+from sklearn.multiclass import OneVsRestClassifier
```
RituRajSingh878: I will remove it.

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -672,6 +673,73 @@ class VWRegressor(VW, RegressorMixin):
 
     pass
 
+class ThresholdingMultiClassifierMixin(ClassifierMixin):
```
gramhagen: I wouldn't bother creating a separate class, you can roll the predict method into VWMultiClassifier

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -672,6 +673,73 @@ class VWRegressor(VW, RegressorMixin):
...
+
+        Returns
+        -------
+        C : array, shape = [n_samples]
+            Predicted class label per sample.
+        """
+
+        return ThresholdingMultiClassifierMixin.predict(self, X=X)
+
+    def decision_function(self, X):
```
gramhagen: it would be more accurate to use predict_proba instead of decision_function.  my understanding is predict_proba provides probabilities for each class and decision_function provides the distance to the decision boundary. So since vw is providing probabilities that should be the right one here.

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -672,6 +673,73 @@ class VWRegressor(VW, RegressorMixin):
...
+        -------
+        C : array, shape = [n_samples]
+            Predicted class label per sample.
+        """
+        scores = self.decision_function(X)
+        indices = scores.argmax(axis=1)+1
+        return indices
+
+
+class VWMultiClassifier(SparseCoefMixin, ThresholdingMultiClassifierMixin, VW):
```
gramhagen: this can then change to 
```
class VWMultiClassifier(ClassifierMixin, VW):
```

I don't think we're actually using SparseCoefMixin, and since that is moving location within sklearn we should just avoid it if possible..

python/vowpalwabbit/sklearn_vw.py
RituRajSingh878: okay

python/vowpalwabbit/sklearn_vw.py
RituRajSingh878: okay

RituRajSingh878:
> Awesome work. Can you update the VWClassifier docstring to point people to VWMultiClassifier now?
> https://github.com/VowpalWabbit/vowpal_wabbit/blob/267e2468d58efd11839881209fc633ebb96a1215/python/vowpalwabbit/sklearn_vw.py#L618

sure

RituRajSingh878:
I think It is good to go.

RituRajSingh878:
@gramhagen , @jackgerrits  Can we merge it?

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -673,6 +673,74 @@ class VWRegressor(VW, RegressorMixin):
...
+        >>> X = np.array([ [10, 10], [8, 10], [-5, 5.5], [-5.4, 5.5], [-20, -20],  [-15, -20] ])
+        >>> y = np.array([1, 1, 2, 2, 3, 3])
+        >>> from vowpalwabbit.sklearn_vw import VWMultiClassifier
+        >>> model = VWMultiClassifier(oaa=3, loss_function='logistic')
+        >>> model.fit(X, y)
+        >>> model.predict(X)
+        """
+
+        scores = self.predict_proba(X)
+        indices = scores.argmax(axis=1)+1
```
jackgerrits: Can you explain why this is needed?

jackgerrits: I see, this selects the largest score in axis 1 for each list if probabilities then adds 1 to make it start from 1.

RituRajSingh878:
@jackgerrits @gramhagen   thanks

PR #2334, opened 2020-03-12
LDA / minibatch bug fix

lalo:


lalo:
Going to wait until all tests pass, still want to verify more and think about it. This is just a simple patch to see where it gets us.

lalo:
Not the best solution, but 'a' solution. Not helpful that the batching behavior is ingrained in LDA seems a bit forced in. This need a deeper look.

jackgerrits:
So it looks like this PR is making so that in the default case (when minibatch is 1) then it is holding off on finishing until the finish method. Otherwise it is doing the old behavior of doing the learning when the batch is processed. Does that seem right?

PR #2336, opened 2020-03-14
Documentation correction of pyvw.py

### LINKED ISSUE #2267: Documentation of pyvw.py
#### Describe the bug
The documentation of `pyvw.py` can be corrected i.e to add essential details like Parameters, Return type, Examples etc. The documentation of `sklearn_vw.py` provides with this essential details and becomes easy to use at the user end. We can also follow the specific patter of documenation in all `.py `files as in https://numpydoc.readthedocs.io/en/latest/format.html#sections. Is it a good idea?

Smit-create:
Fixes #2267 

**Description**
After changing the documentation style to numpy docstyle in #2305, this PR aims at changing the docs of `pyvw.py` accordingly.


Smit-create:
@jackgerrits @gramhagen Please review the changes. The updated doc of pyvw.py with this diff is attached here.
[pyvwdoc1.pdf](https://github.com/VowpalWabbit/vowpal_wabbit/files/4336367/pyvwdoc1.pdf)

python/vowpalwabbit/pyvw.py
```diff
@@ -8,6 +8,22 @@
 class SearchTask():
     """Search task class"""
     def __init__(self, vw, sch, num_actions):
+        """
+        Parameters
+        ----------
+
+        vw : vw object
```
jackgerrits: Is this able to link to the docs for the vw object?

python/vowpalwabbit/pyvw.py
```diff
@@ -40,27 +56,103 @@ def run():
...
+        - pDECISION_SCORES : Decision scores prediction type
+
+    Examples
+    --------
+
+    >>> from vowpalwabbit import pyvw
+    >>> import pylibvw
+    >>> vw = pyvw.vw(quiet=True)
+    >>> ex = vw.example('1 |a two features |b more features here')
+    >>> pyvw.get_prediction(ex, pylibvw.vw.pSCALAR)
```
jackgerrits: You should not use get_prediction for more than one prediction type as it is essentially a reinterpret_cast. Only use the correct one, which in this case is pSCALAR

python/vowpalwabbit/pyvw.py
```diff
@@ -80,15 +172,31 @@ class vw(pylibvw.vw):
...
-        will be translated into passing two -q keys"""
+        """Initialize the vw object.
+
+        Parameters
+        ----------
+
+        arg_str : str
+            The arg_str is the same as the command line arguments, by default is
+            None. You'd use to run vw (eg,"--audit").
+
```
jackgerrits:
```suggestion
            The command line arguments to initialize VW with, for example "--audit". By default is None.
```

python/vowpalwabbit/pyvw.py
```diff
@@ -121,7 +229,35 @@ def format_input(key, val):
...
     def parse(self, str_ex, labelType=pylibvw.vw.lDefault):
         """Returns a collection of examples for a multiline example learner or a single
-        example for a single example learner."""
+        example for a single example learner.
+
+        Parameters
+        ----------
+
+        str_ex : str/list of str
+            string representing examples
```
jackgerrits: This requires more explanation

python/vowpalwabbit/pyvw.py
```diff
@@ -121,7 +229,35 @@ def format_input(key, val):
...
+        labelType : integer
```
jackgerrits: This is not really an integer. but a kind of fake enum. Can we be more specific here?

python/vowpalwabbit/pyvw.py
Smit-create: That looks better. Thanks!

python/vowpalwabbit/pyvw.py
Smit-create: Sure!

python/vowpalwabbit/pyvw.py
Smit-create: I will provide the list of integers and the corresponding labelType that it maps

python/vowpalwabbit/pyvw.py
Smit-create: This doesn't link to the vw object. We can use "See Also" sub-section of  sphinx docs which is used to recommend the user to have a look at some particular doc related to this.

Smit-create:
I have made the suggested changes. Please find the updated doc attached here.
[pyvwdoc2.pdf](https://github.com/VowpalWabbit/vowpal_wabbit/files/4341659/pyvwdoc2.pdf)

Smit-create:
@jackgerrits Please review this. Thanks!

RituRajSingh878:
@Smit-create later maybe we will follow `doc8` style. one of the checks is-


   ` - lines should not be longer than 79 characters - D001`

We can modify this length but Can you please follow this check in the doc.
@jackgerrits We should follow the above check for the line or line length will be greater than this in our doc maybe about 100 char.

Smit-create:
I will have a look at that and wrap the lines within 72 characters following the PEP8 style.

Smit-create:
@jackgerrits @gramhagen I have used numpy doc format of documentation, and PEP8 style to wrap the lines. Please review the changes.

Smit-create:
Please review this.

Smit-create:
> have you been able to run any python tests

It passes all the tests locally.

Smit-create:
This diff fixes all the code quality warnings apart from a few line lengths which can be ignored.
Please find the documentation of `pyvw.py`
[pyvw1.pdf](https://github.com/VowpalWabbit/vowpal_wabbit/files/4384904/pyvw1.pdf)

gramhagen:
cool thanks!

Smit-create:

jackgerrits:
What is the reason for the `repr` change? I'm not sure I understand what that is doing

Smit-create:
> What is the reason for the `repr` change?

It will provide the information of line breaks which will be replace with a space in `dedent`

jackgerrits:
Using repr with return the string wrapped in quotes, as it is the representation of that. Passing this to dedent will break the intended functionality of dedent. I really don't see a reason to combine repr and dedent

Smit-create:
Yes, I checked that it didn't work in all cases. I can fix it another way

Smit-create:
Hi @jackgerrits, I have added a new function to resolve this. It works in the following way:
```
>>> from vowpalwabbit import pyvw
>>> s = """
...     Hello
...             World
... """
>>> pyvw.fillindents(s)
'Hello   World'
```

jackgerrits:
What is the issue you are trying to solve here?

Smit-create:
> What is the issue you are trying to solve here?

In order to display long lines(which here are broken into multiple lines because of length bounds, mostly error messages) as the single line at the user end instead to directly showing these multiple lines with indentation

jackgerrits:
Okay, all you need to do is have string literals broken over several lines. They are automatically concatenated. Check out the link.

python/vowpalwabbit/pyvw.py
```diff
@@ -2,12 +2,35 @@
 """Python binding for pylibvw class"""
 
 from __future__ import division
-import pylibvw
+from pylibvw import example as pylibvw_example, vw as pylibvw_vw
```
jackgerrits: You must also import predictor and search.

This is technically changing the public interface of the package, so I am hesitant to merge this atm.

jackgerrits:
This PR seems to have turned into 2 things:
- Documentation updates
- Code/lint fixes

It looks like the original PR was to do with documentation. Can you please open a separate PR for the code changes? It's already a big one and I don't want to complicate this review further

Smit-create:
Sure! I will revert the changes related to code. And keep this limited to documentation. I will wait for this to merge,  and then open related to code/lint

python/vowpalwabbit/pyvw.py
Smit-create: I suggest to import it only when needed. It is because it creates a significant time difference in importing the module

jackgerrits: Does this break the public interface? Or is it still possible to do the below?
```
from vowpalwabbit import pyvw
pyvw.pylibvw.search...
```

Smit-create: It doesn't break the public interface. Just I meant to import required classes from pylibvw. Other functionalities remains the same as they are. Yes, it will work

Smit-create: Ahha, I wonder how will `pyvw. pylibvw.search` work. As there is no such class/function named `pylibvw`. So, I think this should work presently too. I am not aware if python allows this

Smit-create: > Does this break the public interface? Or is it still possible to do the below?

This won't work even in current master.
```

>>> from vowpalwabbit import pyvw
>>> pyvw.pylibvw.vw
AttributeError: module 'vowpalwabbit.pyvw' has no attribute 'pylibvw'
```

Smit-create:
I have limited this to doc fixes. I will be waiting for this to be merged for code fixes

Smit-create:
@jackgerrits Can you please restart Travis CI?

jackgerrits:
Travis seems to be having some issues. Can you try pushing an empty commit to retrigger it?

python/vowpalwabbit/pyvw.py
```diff
@@ -40,27 +61,109 @@ def run():
...
+
+    Examples
+    --------
+
+    >>> from vowpalwabbit import pyvw
+    >>> import pylibvw
+    >>> vw = pyvw.vw(quiet=True)
+    >>> ex = vw.example('1 |a two features |b more features here')
+    >>> pyvw.get_prediction(ex, pylibvw.vw.pSCALAR)
+    0.0
```
gramhagen: what is this `0.0` ?

Smit-create: Actually, I ran the above code snippet and got the output as `0.0`, I'm sorry if it was wrong

python/vowpalwabbit/pyvw.py
```diff
@@ -40,27 +61,109 @@ def run():
...
+        - pSCALAR : Scalar prediction type
+        - pSCALARS : Multiple scalar-valued prediction type
+        - pACTION_SCORES : Multiple action scores prediction type
+        - pACTION_PROBS : Multiple action probabilities prediction type
+        - pMULTICLASS : Multiclass prediction type
+        - pMULTILABELS : Multilabel prediction type
+        - pPROB : Probability prediction type
+        - pMULTICLASSPROBS : Multiclass probabilities prediction type
+        - pDECISION_SCORES : Decision scores prediction type
+
```
gramhagen: this is inconsistent with how we define label_type inputs, should it be integer type and then list the integers and corresponding enum names? the short descriptions probably aren't needed

Smit-create: Okay, I will change it.

python/vowpalwabbit/pyvw.py
gramhagen: np, that's what i thought, i just wasn't sure if that was supposed to be output. not sure it's needed though

Smit-create: It is needed to show the user how to use this.

python/vowpalwabbit/pyvw.py
```diff
@@ -155,7 +155,7 @@ def get_prediction(ec, prediction_type):
     >>> import pylibvw
     >>> vw = pyvw.vw(quiet=True)
     >>> ex = vw.example('1 |a two features |b more features here')
-    >>> pyvw.get_prediction(ex, pylibvw.vw.pSCALAR)
+    >>> pyvw.get_prediction(ex, 0)
```
jackgerrits, line 158: Why are you using an integer here? Please use the enum

python/vowpalwabbit/pyvw.py
```diff
@@ -137,7 +137,7 @@ def get_prediction(ec, prediction_type):
     ----------
 
     ec : Example
-    prediction_type : integer
+    prediction_type : enumerate
```
jackgerrits, line 140: I meant to use the enum in the example

Smit-create: I didn't understand how can we provide enumerate as an argument as it will produce an object and also needs an iterable argument to be passed in enumerate.

jackgerrits: My original comment was on:
```python
>>> pyvw.get_prediction(ex, pylibvw.vw.pSCALAR)
vs 
>>> pyvw.get_prediction(ex, 0)
```

I believe it should be:
```python
>>> pyvw.get_prediction(ex, pylibvw.vw.pSCALAR)
```

Smit-create: Thanks! I misunderstood it.

Smit-create:
Thanks @jackgerrits

Issue #2123, opened 2019-10-22
--ring_size should be a hard limit for the number of examples


jackgerrits:
Currently --ring_size specifies the initial pool size, but it will grow if it becomes empty. This should not be the default behavior, but it should be fixed size.

The issue is that tests rely on being able to wrap around the ring ("effectively infinite"), and break when there is a hard failure. Needs design: Possibly the right solution here is to make the limit optional, and have an "unlimited" default.

jackgerrits:
If ring size is set explicitly then we use the capped behavior

jackgerrits:
We have no way of knowing how large a multi_ex will be until it is done parsing. Because of this imposing a hard limit on the queue is a potential deadlock situation, and one we cannot detect and avoid. We would hit this if the multi_ex is larger than the allocated ring_size. Having an undetectable deadlock is not okay, so I want to push this change out until we have infrastructure to be able to know the multi_ex size and not get into this situation.

Memory usage is still bounded as it will level out as examples are parsed. The memory consumption is essentially a factor of the largest multi_ex that exists and how long the learning algo takes to drain examples from the queue.

peterychang:
We currently have 2 very different workflows, which complicates this issue. Regardless of what the final design is, we will need to unify them
1. JSON parser
This parser reads in a single line of JSON and outputs a vector of examples. In the case of a multiline example, the vector contains all the examples. Example allocation occurs within the parser itself via a factory object
2. Text (maybe cached?) parser
This parser reads in a single line of input text and outputs a single example. In the case of a multiline example, each line is parsed individually and the full example is reassembled on the driver (learner.cc)

The most logical design would be to make the text/cached parser mimic the JSON parser's behavior, and have it output fully formed multiline examples at a time. As a longer-term goal, we should simply have an example, which can be a multiline example

peterychang:
If we take futureproofing into consideration (ie: multiple parser threads) there are no small changes to fix the ring_size problem.

For example:
Assume we have the following setup
- 2 json parser threads
- ring-size == 4

Now, assume each thread is parsing a multi example with 4 actions each. Each thread has at least 1 example reserved, so neither thread will be able to complete its parsing; resulting in a deadlock.

We could add back-out logic; where if a parser fails to get an example from the pool and it doesn't have all of the available examples, then it would abort and return all examples back to the pool. This would work, but it would be incredibly inefficient since it can't account for parsed and in-flight examples into account.

[...]

PR #2337, opened 2020-03-14
[FIX] Added warning for "--ring_size" with 0 as an argument

### LINKED ISSUE #2335: Passing "--ring_size" with 0 as an argument, stucks the execution
#### Describe the bug
Passing `--ring_size` with 0 as an argument, stucks the execution

#### To Reproduce
Run command:
`vw -d train_adf.txt --cb_explore_adf --ring_size 0`

#### Expected behavior
`--ring_size` shouldn't take 0 as an argument.
I think we need to switch `--ring_size` to its default value, and a relevent warning should be displayed, whenever 0 is passed as an argument.

#### Observed Behavior
execution got stuck for an undefined time

#### Environment
8.8.0
Linux with python

adityauser:
Fixes #2335.

Switched  `--ring_size` to its default value, and a relevent warning is displayed, whenever 0 is passed as an argument.


mritunjaysharma394:
Hi, I suggest you to use `git fetch upstream` and then pull it in your forked branch so that your branch is in sync with the base branch. @jackgerrits your thoughts on it?

jackgerrits:
Checking for 0 makes sense. I had a check and if you pass -1, VW crashes. Would you be able to include handling that issue in this PR?

adityauser:
Actually datatype of ring_size is size_t [(see this line](https:github.com/VowpalWabbit/vowpal_wabbit/blob/0eb1ffdd21525a3003e526106915549f95f0a5f4/vowpalwabbit/parse_args.cc#L1327)). So, passing -1 does overflow, on my machine passing 100000 crashes VW.
What should be the upper bound of ring_size? 
Is it related to issue #2123?

jackgerrits:
It is unrelated to #2123. Allowing the input to be negative and them implicitly overflowing is surprising behavior. I think this is the issue that should be fixed. It is my opinion that there should not be a ceiling on the number.

adityauser:
@jackgerrits passing 0.1 throws error: `the argument ('0.1') for option '--ring_size' is invalid`.
It would be better if for -1 also it should throw the same error, but I couldn't make it.
Can you please guide me? I'm not much fimilar with boost options
The solution I could come up with is:
   ```
    bool strict_parse = false;
    int ring_size_int;
    size_t ring_size;
    option_group_definition vw_args("VW options");
    vw_args.add(make_option("ring_size", ring_size_int).default_value(256).help("size of example ring"))
        .add(make_option("strict_parse", strict_parse).help("throw on malformed examples"));
    options.add_and_parse(vw_args);
    if (ring_size_int<0)
    {
        THROW("ring_size should be positive");
    }
    else
    {
        ring_size = ring_size_int;
    }
    if (ring_size == 0)
    {
      all.trace_message << "warning: --ring_size: ring_size can't be 0, switching to the default value of 256"
                          << endl;
      ring_size = 256;
    }
```
In the above code if I change datatype of `ring_size_int` to `long int`, then passing any value of ring_size will thow [this](https://github.com/VowpalWabbit/vowpal_wabbit/blob/a428d8f7954e29e5a5fba76517f47895c2c7528b/vowpalwabbit/options_boost_po.cc#L185) error. It seems to be a bug.

**Comment by jackgerrits** [comment]
The error you showed is because you're passing a non whole number to an option accepting an integer, so boost rightly gives an error. Granted, the error message isn't great. I'm not sure what options boost po gives there.

Your proposed solution is inconsistent. One way throws and the other sets to a default value and prints something. 

To use a type as an option it must be specified here: https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/options_types.h#L33

**Comment by adityauser** [comment]
Why checks are failing?

**Comment by jackgerrits** [comment]
It fails to compile:
> /Users/runner/runners/2.165.2/work/1/s/vowpalwabbit/parse_args.cc:1339:24: error: non-constant-expression cannot be narrowed from type 'int' to 'size_t' (aka 'unsigned long') in initializer list [-Wc++11-narrowing]
    all.p = new parser{ring_size, strict_parse};
                       ^~~~~~~~~

File: vowpalwabbit/parse_args.cc [hunk 1]
```diff
@@ -1325,10 +1325,20 @@ vw& parse_args(options_i& options, trace_message_t trace_listener, void* trace_c
 
     bool strict_parse = false;
     size_t ring_size;
+    int tmp;
```
jackgerrits: Please use a more descriptive name

vowpalwabbit/parse_args.cc
```diff
@@ -1325,10 +1325,20 @@ vw& parse_args(options_i& options, trace_message_t trace_listener, void* trace_c
...
         .add(make_option("strict_parse", strict_parse).help("throw on malformed examples"));
     options.add_and_parse(vw_args);
+    
+    if (tmp <= 0)
+    {
+      THROW("ring_size should be positive");
+    }
+    else
+    {
+        ring_size = tmp;
```
jackgerrits: please use a static cast to avoid the warning

vowpalwabbit/parse_args.cc
```diff
@@ -1325,10 +1325,20 @@ vw& parse_args(options_i& options, trace_message_t trace_listener, void* trace_c
...
-    vw_args.add(make_option("ring_size", ring_size).default_value(256).help("size of example ring"))
+    vw_args.add(make_option("ring_size", tmp).default_value(256).help("size of example ring"))
         .add(make_option("strict_parse", strict_parse).help("throw on malformed examples"));
     options.add_and_parse(vw_args);
+    
+    if (tmp <= 0)
+    {
+      THROW("ring_size should be positive");
+    }
+    else
```
jackgerrits: If you remove the else then you don't need to declare ring_size separate to its initialization.

adityauser:
Is it okay now @jackgerrits ?

jackgerrits:
Looks good, will merge when tests pass.

adityauser:
Negative numbers are also passing flawlessly through parameters (such as: `ftrl_alpha,  ftrl_beta, mem, termination, initial_pass_length, l1, l2, decay_learning_rate arg` ).

Issue #2340, opened 2020-03-15
vw taking invalid arguments of power on learning rate decay and learning rate


mritunjaysharma394:
#### Describe the bug
The commands `vw house_dataset.txt -l 8.9` and `vw house_dataset.txt --power_t 1.5
` should produce invalid argument error since both learning rate and power on learning rate decay should be in the range [0,1]. However, the commands are working with these values
#### To Reproduce
1. `vw house_dataset.txt -l 15`
2. `vw house_dataset.txt -l -6`
3. `vw house_dataset.txt --power_t 13`
4. `vw house_dataset.txt --power_t -3`

Using any of the above commands can reproduce this behavior.
#### Expected behavior
1. Throw an error like this when learning rate entered is -4. : 
```
vw house_dataset.txt --power_t 13

finished run
number of examples = 0
weighted example sum = 0.000000
weighted label sum = 0.000000
average loss = n.a.
total feature number = 0
the value of power_on_learning_rate_decay must be in [0,1]
vw (parse_args.cc:1403): the value of power_on_learning_rate_decay must be in [0,1]

```
2. Or the other thing can be done is prompting and resetting it to a default value of learning rate like 0.5.
```
 vw house_dataset.txt -l 3.4
the value of learning_rate must be in [0,1]. resetting from: 3.4 to 0.5
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = house_dataset.txt
can't open 'house_dataset.txt', sailing on!
num sources = 0
average  since         example        example  current  current  current
loss     last          counter         weight    label  predict features

finished run
number of examples = 0
weighted example sum = 0.000000
weighted label sum = 0.000000
average loss = n.a.
total feature number = 0
```
#### Observed Behavior
1. Running flawlessly with:
```
vw house_dataset.txt --power_t 13
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 13
using no cache
Reading datafile = house_dataset.txt
num sources = 1
average  since         example        example  current  current  current
loss     last          counter         weight    label  predict features
0.000000 0.000000            1            1.0   0.0000   0.0000        5
0.666667 1.000000            2            3.0   1.0000   0.0000        5

finished run
number of examples = 3
weighted example sum = 4.000000
weighted label sum = 2.000000
average loss = 0.750000
best constant = 0.500000
best constant's loss = 0.250000
total feature number = 15

```

#### Environment
[...]

PR #2341, opened 2020-03-15
[FIX] Resolves issue #2340 to prompt an error for invalid power_t and -l arguments passed 

mritunjaysharma394:
Fixes VowpalWabbit/vowpal_wabbit#2340

The commented part is another approach to fix the issue by resetting the invalid arguments to the default value. Please review @jackgerrits  and tell which will be a better option?


jackgerrits:
@mritunjaysharma394, as discussed in the issue these two parameters are unbounded and therefore should not have checks.

mritunjaysharma394:
Thank you so much for the clarification. I am going to close the PR as well as the issue.

PR #2344, opened 2020-03-17
Change RelWithDebInfo to use O3 instead of O2 and set it as default

jackgerrits:


CMakeLists.txt
```diff
@@ -8,7 +8,9 @@ file(READ version.txt PACKAGE_VERSION)
 string(STRIP ${PACKAGE_VERSION} PACKAGE_VERSION)
 message(STATUS "VowpalWabbit Version: ${PACKAGE_VERSION}")
 
-set(DEFAULT_BUILD_TYPE "Release")
+set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O3 -g -DNDEBUG")
+set(CMAKE_C_FLAGS_RELWITHDEBINFO "-O3 -g -DNDEBUG")
+set(DEFAULT_BUILD_TYPE "RelWithDebInfo")
```
peterychang: Do we want the default build to be RelWithDebInfo? It'll bloat the binaries

jackgerrits: Hmm yeah that is true, I thought that the default having debug info might be good? But bigger binaries is maybe not worth it

PR #2346, opened 2020-03-18
Fixes AttributeError in multiclass_probabilities_label

### LINKED ISSUE #2345: AttributeError in multiclass_probabilities_label
#### Describe the bug
Produces AttributeError while passing an example as an argument in multiclass_probabilities_label.

#### To Reproduce
```
>>> n = 3
>>> model = pyvw.vw(loss_function='logistic', oaa=n, probabilities=True, quiet=True)
>>> ex = model.example('1 | a b c', 2)
>>> model.learn(ex)
>>> mpl = pyvw.multiclass_probabilities_label(ex)
Traceback (most recent call last)
  ...
AttributeError: 'example' object has no attribute 'get_multiclass_probabilities'

```
#### Observed Behavior
```

AttributeError                            Traceback (most recent call last)
<ipython-input-4-0bd465ff351b> in <module>
----> 1 mpl = pyvw.multiclass_probabilities_label(ex)

~/.local/lib/python3.7/site-packages/vowpalwabbit-8.8.0-py3.7-linux-x86_64.egg/vowpalwabbit/pyvw.py in __init__(self, label, prediction)
    540         abstract_label.__init__(self)
    541         if isinstance(label, example):
--> 542             self.from_example(label)
    543         else:
    544             self.prediction = prediction

~/.local/lib/python3.7/site-packages/vowpalwabbit-8.8.0-py3.7-linux-x86_64.egg/vowpalwabbit/pyvw.py in from_example(self, ex)
    545 
    546     def from_example(self, ex):
--> 547         self.prediction = ex.get_multiclass_probabilities()
    548 
    549     def __str__(self):

AttributeError: 'example' object has no attribute 'get_multiclass_probabilities'
```

#### Environment

VW 8.8.0 on Linux with Python bindings
.

Smit-create:
Fixes #2345
### Changes made
The `example` object has no attribute `get_multiclass_probabilities`, therefore, this was changed to `get_scalars` (`pMULTICLASSPROBS`)

### DIFF
--- a/python/vowpalwabbit/pyvw.py
+++ b/python/vowpalwabbit/pyvw.py
@@ -540,7 +540,7 @@ class multiclass_probabilities_label(abstract_label):
             self.prediction = prediction
 
     def from_example(self, ex):
-        self.prediction = ex.get_multiclass_probabilities()
+        self.prediction = ex.get_scalars()
 
     def __str__(self):
         s = []
--- a/python/vowpalwabbit/pyvw.py
+++ b/python/vowpalwabbit/pyvw.py
@@ -540,7 +540,7 @@ class multiclass_probabilities_label(abstract_label):
             self.prediction = prediction
 
     def from_example(self, ex):
-        self.prediction = ex.get_scalars()
+        self.prediction = get_prediction(ex, pylibvw.vw.pMULTICLASSPROBS)
 
     def __str__(self):
         s = []
--- a/python/tests/test_pyvw.py
+++ b/python/tests/test_pyvw.py
@@ -143,16 +143,12 @@ def test_cost_sensitive_label():
 
 
 def test_multiclass_probabilities_label():
-    n = 3
+    n = 4
     model = pyvw.vw(loss_function='logistic', oaa=n, probabilities=True, quiet=True)
-    ex = model.example('1 | a b c', 2)
+    ex = model.example('1 | a b c d', 2)
     model.learn(ex)
     mpl = pyvw.multiclass_probabilities_label(ex)
-    assert str(mpl) == '1:0.3333333432674408 2:0.3333333432674408 3:0.3333333432674408'
-    ex = model.example('1 | a b', 2)
-    model.learn(ex)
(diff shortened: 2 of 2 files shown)


python/vowpalwabbit/pyvw.py
```diff
@@ -540,7 +540,7 @@ def __init__(self, label, prediction=None):
             self.prediction = prediction
 
     def from_example(self, ex):
-        self.prediction = ex.get_multiclass_probabilities()
```
peterychang: Instead of changing this call, I think we should add `get_multiclass_probabilities()` to the attributes. After this line https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/pylibvw.cc#L870

@jackgerrits thoughts?

jackgerrits: Multiclass probs isn't a prediction type in example.h: https://github.com/VowpalWabbit/vowpal_wabbit/blob/2d00e4514f2ca72fb64b18075100d62c81395e73/vowpalwabbit/example.h#L43

But for some reason in python multiclass probs is it's own prediction type. I don't know how that came to be. But you can see here that it maps to get_scalars:
https://github.com/VowpalWabbit/vowpal_wabbit/blob/2d00e4514f2ca72fb64b18075100d62c81395e73/python/vowpalwabbit/pyvw.py#L72

We should probably use the get_prediction function with the type pMULTICLASSPROBS. That way intention is clearer and the conversion logic is all in one place.

Smit-create: As, it points to `ec.get_scalars()`, we could directly call it using get_scalars, instead of calling `get_prediction function` to avoid unnecessary function call. Is it a good idea?

jackgerrits: I think we want to opt for using get_prediction here for maintainability. (the python bindings are likely to be refactored soonish as well)

Smit-create: Okay, sure. I will change that.

Smit-create:
Thanks! @jackgerrits

PR #2347, opened 2020-03-18
Implement the core reduction for slates

jackgerrits:
- Tested using a mock base reduction and asserting the required transformation takes place
    - This is a new way of testing we haven't done before, and I think it could be a great way to build robust and localized tests going forward
- Small refactor to the way CCB prints its decision scores as slates needed to reuse that logic
- Extracted out decision scores into its own header for reuse
- First implementation of reduction, there are opportunities for optimization but with refactors looming it's questionable how valuable it is to do that now vs later


vowpalwabbit/slates.cc
```diff
@@ -0,0 +1,262 @@
...
+      if (slates_label.labeled)
+      {
+        global_cost_found = true;
+        global_cost = slates_label.cost;
+      }
+    }
+    else if (slates_label.type == slates::example_type::action)
+    {
+      ccb_label.type = CCB::example_type::action;
+      slot_action_pools[slates_label.slot_id].push_back(action_index);
```
peterychang, line 59: slot_id is always [0, num_slots)?

vowpalwabbit/slates.h
```diff
@@ -0,0 +1,53 @@
...
+
+namespace slates
+{
+struct slates_data
+{
+ private:
+  std::vector<slates::label> _stashed_labels;
+
+  /*
+  The primary job of this reduction is to convert slate labels to a form CCB can process.
```
peterychang, line 23: Can you add a comment stating shared must come before action, which must come before slot in the example vector?

vowpalwabbit/slates.cc
```diff
@@ -0,0 +1,262 @@
...
+#include "ccb_label.h"
+
+using namespace VW::config;
+
+namespace slates
+{
+template <bool is_learn>
+void slates_data::learn_or_predict(LEARNER::multi_learner& base, multi_ex& examples)
+{
+  _stashed_labels.reserve(examples.size());
```
peterychang, line 27: is `_stashed_labels` assumed to be empty here? Weird things can happen if its not. Maybe call `clear()` just to enforce the assumption?

vowpalwabbit/slates.cc
```diff
@@ -0,0 +1,262 @@
...
+      }
+
+      float l = CB_ALGS::get_cost_estimate(
+          label_probabilties[TOP_ACTION_INDEX], cost, predictions[slot_index][TOP_ACTION_INDEX].action);
+      loss += l * predictions[slot_index][TOP_ACTION_INDEX].score;
+    }
+  }
+
+  bool holdout_example = is_labelled;
+  for (const auto& example : ec_seq)
```
peterychang, line 195: nit: you could bypass this entire loop if holdout_example is false.

vowpalwabbit/slates.cc
jackgerrits: Yes, it is meant to be. But this might not be guaranteed...

jackgerrits: Adding a throw here to enforce this, until I get frontload this into parsers

vowpalwabbit/slates.cc
jackgerrits: `_stashed_labels` is a vector and not a v_array. But I can still add a clear

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -490,92 +491,40 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
   examples[0]->pred.decision_scores = decision_scores;
 }
 
-void print_decision_scores(int f, decision_scores_t& decision_scores)
+std::string generate_ccb_label_printout(const std::vector<example*>& slots)
 {
-  if (f >= 0)
+  size_t counter = 0;
+  std::string label_str;
```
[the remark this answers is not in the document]
jackgerrits: Good point

vowpalwabbit/slates.cc
```diff
@@ -0,0 +1,270 @@
...
+      ccb_label.type = CCB::example_type::shared;
+      if (slates_label.labeled)
+      {
+        global_cost_found = true;
+        global_cost = slates_label.cost;
+      }
+    }
+    else if (slates_label.type == slates::example_type::action)
+    {
+      if (slates_label.slot_id >= num_slots)
```
peterychang, line 59: Documentation says that `std::count_if()` returns a signed int. I'm not sure if its possible to actually get a negative in this case, but something to be aware of

jackgerrits: It's assigned into a size_t above. While in the docs it is a difference type it also says:
> "number of elements satisfying the condition.

That's a pretty annoying API...

PR #2350, opened 2020-03-18
Enable automatic building and hosting of documentation for every new commit

jackgerrits:
This PR enables a GitHub action to run on every commit in master which will build the documentation for C++ and Python, and then commit the updated documentation to the docs repo. These docs will be hosted at https://vowpalwabbit.org/docs. Current docs there are out of date as it is a manual process. 

I've tested this in my fork, so you can see how it works here: https://github.com/jackgerrits/docs

Documentation is crucially important for VW. I think having these docs automatically update is a good first step, and will be a major convenience in us improving documentation.

It is using GitHub Actions as I originally used that for experimentation due to the GitHub token being present, but it turns out I had to use a separate access token anyway. I would say that GitHub Actions worked quite well for this though.


lalo:
Should it be every commit or every release? If we do every commit and we don't update the version referenced the docs might be misleading.

jackgerrits:
I think it makes sense to have latest built docs. As an extension to this I plan on adding an action that builds tags. You'll notice that as part of the path/url it includes latest. We would then have a directory per tagged version

PR #2356, opened 2020-03-20
Auto-read version in conf.py for docs from version.txt

### LINKED ISSUE #2349: Python documentation version should be automatically fetched
Python documentation version is currently outdated and hardcoded in this file: https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/docs/source/conf.py

It should be automatically read from https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/version.txt which is the central location to determine current version number.

### LINKED ISSUE #2357: Version attribute of vowpalwabbit in python
#### Describe the bug
Version attribute of vowpalwabbit in python

#### To Reproduce
```
>>> import vowpalwabbit
>>> vowpalwabbit.__version__
AttributeError: module 'vowpalwabbit' has no attribute '__version__'
```

#### Environment
VW 8.8.0 on Linux with the Python bindings

We should add this attribute, for knowing the version of vowpalwabbit. I found this useful while working on #2349 
This will also make the `conf.py` look clean in #2356 and also provide an attribute to provide live updates of version while running VW in python

Smit-create:
Fixes #2349 
Fixes #2357 
* python
  * Auto-read version from version.txt while building docs for python
  * Add an attribute `__version__`

### Changes Made
Added an attribute __version__ to provide the version of the currently used vowpalwabbit. The same will be used in `conf.py` for auto-update wihle creating docs.
Example:
```
>>> import vowpalwabbit
>>> vowpalwabbit.__version__
'8.8.0'
```

### DIFF
--- a/python/docs/source/conf.py
+++ b/python/docs/source/conf.py
@@ -24,10 +24,22 @@ project = u'VowpalWabbit'
 copyright = u'2019, John langford et al'
 author = u'John langford et al'
 
+
+# Read version automatically from the version.txt.----------------------------
+def rev_replace(st, old, new, occurrence):
+    li = st.rsplit(old, occurrence)
+    return new.join(li)
+
+path = os.getcwd() # path of present directory
+
+# This makes sure that the path is correct after replacing only the last
+# occurrence of 'python/docs/source' in path by 'version.txt'
+ver_path = rev_replace(path, 'python/docs/source', 'version.txt', 1)
+
 # The short X.Y version
-version = u''
+version = open(ver_path).readlines()[0].strip()
 # The full version, including alpha/beta/rc tags
-release = u'8.6.1'
+release = version
 
 
 # -- General configuration ---------------------------------------------------
--- a/python/docs/source/conf.py
+++ b/python/docs/source/conf.py
@@ -14,6 +14,8 @@
 #
 import os
 import sys
+import vowpalwabbit
+
 sys.path.insert(0, os.path.abspath('.'))
 sys.path.insert(0, os.path.abspath('../../'))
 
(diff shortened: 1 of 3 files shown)


Smit-create:
Screenshot from 2020-03-20 08-54-26
@jackgerrits Please find the attached screenshot of html doc which shows the version `8.8.0` with this diff

python/docs/source/conf.py
```diff
@@ -24,10 +24,22 @@
...
+# Read version automatically from the version.txt.----------------------------
+def rev_replace(st, old, new, occurrence):
+    li = st.rsplit(old, occurrence)
+    return new.join(li)
+
+path = os.getcwd() # path of present directory
+
+# This makes sure that the path is correct after replacing only the last
+# occurrence of 'python/docs/source' in path by 'version.txt'
+ver_path = rev_replace(path, 'python/docs/source', 'version.txt', 1)
```
jackgerrits: Does this work on windows too? Perhaps Pathlib is a cross platform way to do this?

Smit-create: I think this will work for windows too. As it is just dependent on python (`os library`) and `string` operations. Still, I will have a look at this and provide you with an update on this.

Smit-create:
@jackgerrits Does this look good?

python/docs/source/conf.py
Smit-create: The new change will work for windows too.

python/docs/source/conf.py
```diff
@@ -14,6 +14,8 @@
 #
 import os
 import sys
+import vowpalwabbit
```
jackgerrits: This assumes the installed version of vowpalwabbit is the one you're generating documentation right? I think that's okay though

python/vowpalwabbit/version.py
```diff
@@ -0,0 +1,5 @@
+# Provides the present version of VowpalWabbit
+
+import pkg_resources
+
+__version__ = pkg_resources.require('vowpalwabbit')[0].version
```
jackgerrits: Why do you need to get the 0th item here?

python/docs/source/conf.py
Smit-create: I have changed this by keeping in view that the user wouldn't use the doc file. It is at the development end which is synced with the current master and is always updates with the new versions.

python/vowpalwabbit/version.py
Smit-create: It will produce the list whose first element is the instance of `'pkg_resources.Distribution'` of which `version` is an attribute created in `setup.py`

python/docs/source/conf.py
Smit-create: This will also help the users to get updated with the newer versions. As it will allow it to compare their current version in their local computers to the version of the docs provided online

Smit-create:
@jackgerrits  @peterychang  Thanks!

PR #2364, opened 2020-03-23
Prevent overwriting passes when data is passed in VW at object creation

Smit-create:
Partial fix for #1469
* Prevents overwriting  the `passes` when data is passed at object creation

### DIFF
--- a/python/vowpalwabbit/sklearn_vw.py
+++ b/python/vowpalwabbit/sklearn_vw.py
@@ -343,13 +343,10 @@ class VW(BaseEstimator):
         ext_file_args = ['data', 'd']
         if any(x in self.params for x in ext_file_args):
             # fitting will be handled by vw directly
-            # with the given number of passes and initialize the vw model here
-            # itself and store it in self.vw_
             self.fit_ = True
             if not self.params.get('passes', None): # To prevent overwriting of passes
-                self.params['passes'] = 2   # default 2 if not given
+                self.params['passes'] = 1   # default 1 if not given
             self.passes_ = self.params['passes']
-            self.vw_ = self.get_vw()
         else:
             # store passes separately to be used in fit
             self.passes_ = self.params.pop('passes', 1)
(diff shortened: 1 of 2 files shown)


RituRajSingh878:
@Smit-create In this [comment ](https://github.com/VowpalWabbit/vowpal_wabbit/issues/1469#issuecomment-591191032)
I think @gramhagen is not allowing us to do this. Also, We should put` init` as clean as possible because I was [trying to pass](https://github.com/VowpalWabbit/vowpal_wabbit/issues/2333) the ` check_estimator`, and It is failing due to the extra variables created in the `__init__`. Also, I am pretty sure that the sklearn check_estimator will not allow doing this if we wish to pass the check_estimaot of sklearn So I think we should not add anything in `init.`

Let's see what others think?

Smit-create:
So, going through the comment, I think that the case when the data file is passed as the parameters needs attention, to make it clearer that when to initialize a model, whether during the fit(here it is not required though), or directly during predict, as an option to this PR.

Sharad24:
Comments over [here](https://github.com/VowpalWabbit/vowpal_wabbit/pull/2285#discussion_r383425664) might be a bit relevant :)

Smit-create:
Sure, Thanks for the update on this.

Smit-create:
I have updated the PR accordingly and removed model creation in __init__.

Smit-create:
@jackgerrits Please review this.

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -344,7 +344,9 @@ def __init__(self,
         if any(x in self.params for x in ext_file_args):
             # fitting will be handled by vw directly
             self.fit_ = True
-            self.passes_ = 1
+            if not self.params.get('passes', None): # To prevent overwriting of passes
+                self.params['passes'] = 1   # default 1 if not given
+            self.passes_ = self.params['passes']
```
gramhagen:
```suggestion
            self.passes_ = self.params.get('passes', 1)
```

Smit-create: Oh, I am so sorry, it went completely out of my mind. Thanks :)

Smit-create:
Does this look good to merge?

Smit-create:
Thanks @gramhagen!

Smit-create:

PR #2366, opened 2020-03-23
Multiple passes of cleanup to iobuf

kumpera:
This cleanup achieves the following:

- make `head` and `space` private
- move `currentname` and `finalname` to parser and change their type to `std::string`
- moved `isbinary` and `readto` to be member functions of `iobuf`
- removed some dead cleanup code that was no longer needed.


vowpalwabbit/io_buf.cc
```diff
@@ -158,3 +158,26 @@ void io_buf::close_file_or_socket(int f)
   close(f);
 #endif
 }
+
+//make all data that was writen available for reading
+void io_buf::flip_from_write_to_read()
```
peterychang: It looks like this is only used in tests? I'm not sure this operation actually makes any sense in real usage (at least according to the documentation in io_buf.h)

```
/* The i/o buffer can be conceptualized as an array below:
**  _______________________________________________________________________________________
** |__________|__________|__________|__________|__________|__________|__________|__________|   **
** space.begin           space.head             space.end                       space.endarray **
**
** space.begin     = the beginning of the loaded values in the buffer
** space.head      = the end of the last-read point in the buffer
** space.end       = the end of the loaded values from file
** space.endarray  = the end of the allocated space for the array
**
** The values are ordered so that:
** space.begin <= space.head <= space.end <= space.endarray
```

vowpalwabbit/io_buf.cc
```diff
@@ -158,3 +158,26 @@ void io_buf::close_file_or_socket(int f)
...
+  memcpy(dst, space.begin(), to_copy);
+  return to_copy;
+}
+
+void io_buf::replace_buffer(char *buff, size_t capacity)
+{
+  // TODO the following should be moved to v_array
+  space.delete_v();
+  space.begin() = buff;
+  space.end_array = space.end() = buff + capacity;
```
peterychang: space.end() should be set to space.begin(), right? Or is the assumtion that the replacement buffer is filled to capacity when this function is called?

vowpalwabbit/io_buf.h
```diff
@@ -131,23 +133,25 @@ class io_buf
...
+  }
+
   virtual void reset_file(int f)
   {
 #ifdef _WIN32
     _lseek(f, 0, SEEK_SET);
 #else
     lseek(f, 0, SEEK_SET);
 #endif
-    space.end() = space.begin();
```
peterychang: Do we need to call `reset_buffer`?

vowpalwabbit/io_buf.h
```diff
@@ -55,20 +55,22 @@ class io_buf
...
-  char* head;
-  v_array<char> currentname;
-  v_array<char> finalname;
 
   static constexpr int READ = 1;
   static constexpr int WRITE = 2;
 
+  size_t available_to_read() const
+  {
+    return head - space.begin();
```
peterychang: This is the starting index of the first available byte to read, not the number of bytes available to read

vowpalwabbit/io_buf.cc
```diff
@@ -158,3 +158,26 @@ void io_buf::close_file_or_socket(int f)
...
+void io_buf::flip_from_write_to_read()
+{
+  space.end() = head;
+  head = space.begin();
+}
+
+size_t io_buf::copy_to(void *dst, size_t max_size)
+{
+  size_t to_copy = std::min(available_to_read(), max_size);
+  memcpy(dst, space.begin(), to_copy);
```
peterychang: What is this copying? The data thats already been read?

vowpalwabbit/io_buf.cc
jackgerrits: Also it's not clear whether this operation is needed. If I remember I think I added it to hack around the way I was using io_buf in the test

kumpera: This method is only used by tests. I don't see the issue with that by itself.

There's no other way through the current API to flip an io_buf from "write mode" to "read mode".

Peter, it makes sense given that, in "write mode", `end() == begin()`. The documentation you mention only explain how io_buf works when reading.

kumpera: For example, say you write 10 bytes, you end up with the following set of values (assume data is the v_array storage:
```
end == begin == &data[0]
head == &data[10]
```

To read that we need the following set of pointers:

```
head == begin == &data[0]
end == &data[10]
```

This could potentially be used kernel_svm.

vowpalwabbit/io_buf.cc
kumpera: No, `end() == begin()` only makes sense for writing (actually, end is never used).
When reading, end() must point one after the last readable byte.

vowpalwabbit/io_buf.h
kumpera: good catch.

vowpalwabbit/io_buf.h
kumpera: Good point, the naming is wrong, this method is used to tell how many written bytes are in the buffer.  All written data is kept between `being()` and `head`.

kumpera: Renamed to unflushed_bytes_count.

vowpalwabbit/io_buf.cc
peterychang: Ok, so the assumption is that the replacement buffer is filled with data. Can you add a comment making that clear?

vowpalwabbit/io_buf.cc
peterychang: If the documentation is only valid for read-mode, can you add another piece of documentation describing how write-mode works?

lalo: the name of the function could be a bit misleading, feels like its a permissions switch were you are not allowed to write before calling this function.

jackgerrits: I don't think we should support flipping from write/read especially since it isn't needed right now either....

vowpalwabbit/io_buf.cc
jackgerrits: Looks like it is copying out of the internal buffer

vowpalwabbit/io_buf.cc
jackgerrits: I've removed this in the latest iteration

peterychang:
LGTM. The only issue I have with this change is the function name `unflushed_bytes_count`. The function's meaning is different between read and write mode, `unflushed_bytes_count` only makes sense in write mode. In read mode, its the starting index of the unread bytes. I think its fine as long as that distinction is documented somewhere

jackgerrits:
I agree, I'm hoping that io_buf can eventually be split into buffered reader and writer. But for now, I've added a comment to help here.

PR #2367, opened 2020-03-24
Ignore tmp.model

Smit-create:
Added `tmp.model` to `.gitignore` created while running python tests


thealphadollar:
We should add *.model rather.

Smit-create:

Issue #2333, opened 2020-03-10
Write Test For New SKLearn Based Classifier


thealphadollar:
@gramhagen [gave](https://github.com/VowpalWabbit/vowpal_wabbit/issues/2300#issuecomment-596636994) a set of helpful links for things to keep in mind when creating new estimators which we want to be compatible with SKLearn.

Here are the two links:
- http://danielhnyk.cz/creating-your-own-estimator-scikit-learn/
- https://ploomber.io/posts/sklearn-custom/

Going through both the links I found out about the `check_estimator` method present in SKLearn which we can use to test if the estimator implements all required methods or not.

I was wondering if we should add it to our contribution guidelines and also write a python test file that will run all the required checks on all the models we have present in the python folder. This will ensure that new developers can easily start with writing estimators following a particular strict style and also that no code changes break the compatibility of our models with SKLearn methods.

RituRajSingh878:
@thealphadollar  I am working on the implementation/integration of classifier and it will be helpful if we have such a feature and also, I can help here in the implementations of the `check_estmator`.
 During the implementation, we need to add or take care of lots of things, so I think a discussion will be helpful.

thealphadollar:
@RituRajSingh878 Thanks for the hand :) ALso, can you please elaborate what you mean by implement `check_estimator`? In my opinion, we just need to call it and write tests that call it automatically on each of the custom estimators we are writing.

RituRajSingh878:
I tried to do the same, but there are many problems with calling `checkestimator`. The problem is with the way we have implemented our wrapper. 
We can test by `checkestmator` by passing directly class or we can pass its object.

1- By Passing its object and it will show the following
```julia
In [1]: from sklearn.utils.estimator_checks import check_estimator                                                                                                                                          

In [2]: from vowpalwabbit.sklearn_vw import VW                                                                                                                                                              

In [3]: check_estimator(VW())                                                                                                                                                                               
Segmentation fault (core dumped)

```
I tried to find the reason behind this, but I am not able to find it.

2-By Passing class.
```julia
In [2]: from vowpalwabbit.sklearn_vw import VW                                                                                                                                                              

In [3]: check_estimator(VW)                                                                                                                                                                                 
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-3-1749e939b444> in <module>
----> 1 check_estimator(VW)

[...]

PR #2368, opened 2020-03-24
Update python slearn_vw.VW estimator to match SKLearn specs

gramhagen:
Overhaul of base VW class in sklearn_vw to match SKLearn specs.

1. update \_\_init\_\_ params to match `vw --help` (excluding options which are not supported in python or don't make sense in the sklearn setting)
2. removed fit_ attribute in favor of having vw_ != None to determine prior fit
3. dropped transform function, this is a predictor class not a transformer, plus it wasn't doing anything except allowing segfaults by forcing vw_.finish()
4. removed params attribute in favor of directly setting attributes, these get filtered before constructing the vw instances in fit()
5. enabled and ensured check_estimator test passes
6. updated tests/test_sklearn.py TestVW() tests should all pass now, the other classes for children of VW() still need work both in the test and in sklearn_vw
7. added support for pickling the VW() instance (enforced by check_estimator)
8. updated \_\_repr\_\_ method slightly

Related issue: #2333


RituRajSingh878:
I can work on point 6 if you are not working on it.

python/tests/test_sklearn_vw.py
```diff
@@ -228,96 +263,27 @@ def test_predict(self, data):
...
-
-        model = VWMultiClassifier(oaa = 2, loss_function = 'logistic')
+        model = VWMultiClassifier(oaa=2, loss_function='logistic')
         model.fit(data.x, data.y)
-
-        assert np.allclose(raw_model.predict(data.x), model.predict_proba(data.x))
-        # ensure model can make multiple calls to predict
-        assert np.allclose(raw_model.predict(data.x), model.predict_proba(data.x))
+        actual = model.predict_proba(data.x)
+        expected = None
```
RituRajSingh878: why expected is `None`?

gramhagen: i should have put todo's here, sorry. I removed the comparison against the raw model because I was scared that if something broke this might not catch it. so what I wanted to do is determine what the correct expected values are and hard code them here. I just put None now so the test would fail until it's fixed.

gramhagen:
> I can work on point 6 if you are not working on it.

yes please! if you want to push prs to my branch that should work, or if you want to do something different let me know if I need to change anything

RituRajSingh878:
> > I can work on point 6 if you are not working on it.
> 
> yes please! if you want to push prs to my branch that should work, or if you want to do something different let me know if I need to change anything

I will push my work into your branch.

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -94,59 +86,104 @@ def __init__(self,
...
                  save_resume=None,
+                 preserve_performance_counters=None,
                  output_feature_regularizer_binary=None,
                  output_feature_regularizer_text=None,
                  oaa=None,
                  ect=None,
                  csoaa=None,
                  wap=None,
+                 probabilities=None,
+                 score=None,
```
RituRajSingh878: We should not use score here as in` VWRegressor`  we have score function which is inherited from its parent class so It will cause a problem.

RituRajSingh878:
I tried for `VWRegressor` and I made a few changes in code have a look. Now `VWRegressor` is passing the` check_estimator` test.
```julia
index be298ec32..8adea1dcd 100644
--- a/python/vowpalwabbit/sklearn_vw.py
+++ b/python/vowpalwabbit/sklearn_vw.py
@@ -115,7 +115,6 @@ class VW(BaseEstimator):
                  csoaa=None,
                  wap=None,
                  probabilities=None,
-                 score=None,
                  nn=None,
                  inpass=None,
                  multitask=None,
@@ -581,6 +580,9 @@ class VW(BaseEstimator):
                    "appropriate arguments before using this method.")
             raise NotFittedError(msg % {'name': self.__class__.__name__})
 
+        if self.convert_to_vw:
+            X = check_array(X, accept_sparse=True)
+
         try:
             num_samples = X.shape[0] if X.ndim > 1 else len(X)
         except AttributeError:


 class VWRegressor(VW, RegressorMixin):
     """Vowpal Wabbit Regressor model """
 
-    pass
-
+    def _more_tags(self):
+        return {'poor_score': True}
```

python/vowpalwabbit/sklearn_vw.py
gramhagen: oh, good catch

gramhagen:
sorry, where are the regressor changes pushed? I can't seem to find them

RituRajSingh878:
> sorry, where are the regressor changes pushed? I can't seem to find them

I made these changes to pass the `check_estimator(VWRegressor).`

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -398,45 +520,42 @@ def fit(self, X, y=None, sample_weight=None):
...
+                else:
+                    examples = X
+                for idx, example in enumerate(examples):
+                    self.vw_.learn(example)
 
-        X : {array-like, sparse matrix}
-            To be passed into next estimator in pipeline
-        """
-        if not self.get_vw().finished:
-            self.get_vw().finish()
```
RituRajSingh878: Are we going to restore this or not? I think It may cause problems.

I am getting this -
```julia
Exception ignored in: <bound method vw.__del__ of <vowpalwabbit.pyvw.vw object at 0x7f29b9b02d50>>
  File "/home/rituraj/.local/lib/python3.6/site-packages/vowpalwabbit-8.8.0-py3.6-linux-x86_64.egg/vowpalwabbit/pyvw.py", line 258, in __del__
    self.finish()
  File "/home/rituraj/.local/lib/python3.6/site-packages/vowpalwabbit-8.8.0-py3.6-linux-x86_64.egg/vowpalwabbit/pyvw.py", line 249, in finish
    if not self.finished:
AttributeError: 'vw' object has no attribute 'finished'
```

python/tests/test_sklearn_vw.py
```diff
@@ -25,41 +27,52 @@ def data():
...
-        At some point it would probably be worth while figuring out the problem  this and getting the two estimators to
-        pass sklearn's validation check
-        """
 
-        # check_estimator(VW)
-        # check_estimator(VWClassifier)
+class TestVW:
+    def test_check_estimator(self):
+        # run VW through the sklearn estimator validation check
+        check_estimator(VW)
```
thealphadollar: I believe what we should do here is that create a ABC from which all tests need to be derived. This way we can ensure that all tests implement the `test_check_estimator` and we can have more automation. THis will also make sure that in future we can enforce more such tests such as `repr_test` which will ensure two things:

- `__repr__`  methods are written.
- Tests for the same are done.

What do you think? @gramhagen

python/vowpalwabbit/sklearn_vw.py
gramhagen: we should not finish the vw instance until we no longer want to use it (i.e. it's ready for deletion) as any further calls to save will cause a segfault.

I'm not sure why this is used in a transformer if it's not transforming anything? we might want to test out use of the estimator in a pipeline, particularly chained with a downstream estimator to make sure the behavior is as expected.

Until we can determine if this is needed I would exclude it, if we do in fact need the transform method then we need to make sure it works without vw.finish() otherwise check_estimator will crash.

python/tests/test_sklearn_vw.py
gramhagen: hehe, yes sorry, thank _you_ for the test implementation!

i'm okay with having a base test class. it would be preferable to have the check_estimator and repr tests concretely defined in that class so no one needs to write them unless they are changing the behavior of the subclass somehow. is that what you're thinking?

thealphadollar: Yes. That is exactly how I see it happening.

Even if they are changing the behaviour, they can override the function but cannot omit it in their class.

gramhagen:
pulling in suggestions from #2328 by @thealphadollar

gramhagen:
I was able to get tests passing, except for one case in VWMultiClassifier: check_classifiers_predictions fails because we are not predicting both classes. I think this is too stringent of a test given the dataset provided.

and there is a second issues around respecting binary inputs when running check_sparsify_coefficients in the check_estimator code. I'll file both of those against scikit-learn. But for now the rest seems functional.

You do need to run the tests with sklearn >= 0.22 due to https://github.com/scikit-learn/scikit-learn/issues/6981

I also dropped support for the threshold parameter in order to make the code more consistent with sklearn and simplify the inheritance. I had already started leveraging LogisticRegression for mixin support instead of LinearSVC as used in #2328 so there might be some extra hoops I'm jumping through which could be simplified?

python/tests/test_sklearn_vw.py
gramhagen: how's this look?

gramhagen:
cross-referencing SKLearn issues filed from this effort:
https://github.com/scikit-learn/scikit-learn/issues/16798
https://github.com/scikit-learn/scikit-learn/issues/16799

jackgerrits:
Hey @gramhagen, we're getting ready for a release soon. Would you like to try and get this in?

gramhagen:
Yes this would be good. I'll see what I can do to get this wrapped up over there weekend

gramhagen:
@jackgerrits this should be good now, there are some issues using sklearn's internal validation tools for our tests, we need to be eventually we need to be using scikit-learn version >= 0.22 in the build system, but there's also a bug in sklearn (https://github.com/scikit-learn/scikit-learn/issues/16799) so I'm skipping that test for now. We can turn it back on once the PR addressing it is resolved.

jackgerrits:
Thanks @gramhagen, is the original description still up to date as far as describing what this does?

gramhagen:
This PR expands available VW parameters that can be used with the sklearn wrapper, improves the sklearn wrapper by more data validation, moving training to only occur during fit(), and aligns the wrapper with sklearn specifications by using estimator base classes and tests.

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -324,56 +477,24 @@ def __init__(self,
...
-        if self.vw_ is None:
-            self.vw_ = pyvw.vw(**self.params)
-
         return self.vw_
 
-    def fit(self, X, y=None, sample_weight=None):
+    def fit(self, X=None, y=None, sample_weight=None):
         """Fit the model according to the given training data
 
         TODO: for first pass create and store example objects.
```
lalo, line 500: is this todo item stale?

python/vowpalwabbit/sklearn_vw.py
```diff
@@ -742,14 +980,24 @@ class would be predicted.
         >>> X = np.array([ [10, 10], [8, 10], [-5, 5.5], [-5.4, 5.5], [-20, -20],  [-15, -20] ])
         >>> y = np.array([1, 1, 2, 2, 3, 3])
         >>> from vowpalwabbit.sklearn_vw import VWMultiClassifier
-        >>> model = VWMultiClassifier(oaa=3, loss_function='logistic')
+        >>> model = VWMultiClassifier(oaa=2, loss_function='logistic')
```
lalo, line 983: what's the reason of changing this value?

gramhagen: ah good catch, that should be reverted to 3 (or the y values should be updated to only 2). I was changing tests to use 2, because there was a bug with sklearn tests that had > 2 labels and this probably got caught in that change.

jackgerrits: Looks like I jumped the gun with merging! I can submit a PR to fix this if you'd like?

gramhagen: sure, it's just a docstring, but would be good to put in a pr quickly to flip this to 3

jackgerrits: No prob, worth fixing also so we don't forget later. PR here: https://github.com/VowpalWabbit/vowpal_wabbit/pull/2482

PR #2373, opened 2020-03-30
Fix return reference to temporary local bug

jackgerrits:
`options_serializer_boost_po::data()` previously returned a reference to the temporary local object `m_output_stream.str()`. This required a change to the interface of the `options_serializer_i` class to signal the caller that they must own this memory. This function is not being used anywhere, so the other option was simply to remove.

This fixes a warning introduced in Clang 10.

### DIFF
--- a/vowpalwabbit/options_serializer_boost_po.cc
+++ b/vowpalwabbit/options_serializer_boost_po.cc
@@ -4,11 +4,19 @@
 
 #include "options_serializer_boost_po.h"
 
+#include <memory>
+
 using namespace VW::config;
 
 std::string options_serializer_boost_po::str() { return m_output_stream.str(); }
 
-const char* options_serializer_boost_po::data() { return m_output_stream.str().c_str(); }
+std::unique_ptr<const char> options_serializer_boost_po::data() {
+  auto str = m_output_stream.str();
+  char* buffer = new char[str.size() + 1];
+  str.copy(buffer, str.size());
+  buffer[str.size()] = '\0';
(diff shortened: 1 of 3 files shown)


vowpalwabbit/options_serializer_boost_po.cc
```diff
@@ -4,11 +4,19 @@
...
 
 std::string options_serializer_boost_po::str() { return m_output_stream.str(); }
 
-const char* options_serializer_boost_po::data() { return m_output_stream.str().c_str(); }
+std::unique_ptr<const char> options_serializer_boost_po::data() {
+  auto str = m_output_stream.str();
+  char* buffer = new char[str.size() + 1];
+  str.copy(buffer, str.size());
+  buffer[str.size()] = '\0';
+  return std::unique_ptr<const char>(buffer);
```
lokitoth, line 18: Since we are doing a copy here anyways, can we change the interface to be a `std::string`?

jackgerrits: Potentially, but I think the idea is to get a raw binary representation of the object? Although this function is actually not used anywhere.. Should we instead just remove it?

lokitoth: If it is unused, I would remove it. Encapsulation, particularly around configuration is good - it could make it easier to pull boost::po into just the executable, rather than the lib.

jackgerrits: Great, I agree. Will remove.

PR #2375, opened 2020-03-30
Only override RelWithDebInfo flags for non MSVC situations

### LINKED ISSUE #2371: RelWithDebInfo defaults do not work on Windows
This PR (#2344) was recently merged changing the default flags for RelWithDebInfo. This works when the compiler is compatible with GCC/Clang style flags, but does not work for other generators including visual studio.

The cmake script should be updated to either:
1. Good: Determine if build scripts are being generated for GCC/Clang and only set the defaults then
2. Better: Have sensible defaults for each of the common compilers. This means adding the check in 1. and adding a further one for MSVC

jackgerrits:
- Also remove C flags as we have no C code.

Fixes #2371

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,8 +8,14 @@ file(READ version.txt PACKAGE_VERSION)
 string(STRIP ${PACKAGE_VERSION} PACKAGE_VERSION)
 message(STATUS "VowpalWabbit Version: ${PACKAGE_VERSION}")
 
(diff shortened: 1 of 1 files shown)


jackgerrits:
Yeah generally GCC and Clang are compatiable, they should be fine here.

You're right there are actually spots where we misuse things, it would be good to do a pass and fix them.
- `WIN32` is true if the Platform is Windows
- `MSVC` is true if the compiler is MSVC
- `MSVC_IDE` is true if the compiler is MSVC for Visual Studio (as opposed to command line)

PR #2378, opened 2020-03-31
Remove -q: option entirely 

jackgerrits:
This can be safely removed as passing `-q:` will always be correctly interpreted as the standard `-q :` command. There, since it is inaccessible and deprecated we should remove it.

### DIFF
--- a/vowpalwabbit/parse_args.cc
+++ b/vowpalwabbit/parse_args.cc
@@ -612,7 +612,6 @@ void parse_feature_tweaks(options_i& options, vw& all, std::vector<std::string>&
   bool noconstant;
   bool leave_duplicate_interactions;
   std::string affix;
-  std::string q_colon;
 
   option_group_definition feature_options("Feature options");
   feature_options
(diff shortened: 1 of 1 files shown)


vowpalwabbit/parse_args.cc
```diff
@@ -661,8 +660,6 @@ void parse_feature_tweaks(options_i& options, vw& all, std::vector<std::string>&
                .help("Don't remove interactions with duplicate combinations of namespaces. For ex. this is a "
                      "duplicate: '-q ab -q ba' and a lot more in '-q ::'."))
       .add(make_option("quadratic", quadratics).short_name("q").keep().help("Create and use quadratic features"))
-      // TODO this option is unused - remove?
-      .add(make_option("q:", q_colon).help("DEPRECATED ':' corresponds to a wildcard for all printable characters"))
```
lokitoth, line 665: Have we had a major release with this deprecated? If no, I would prefer it if we did not remove it outright, yet.

jackgerrits: There is no way to actually call it, since `-q:` gets interpreted as `-q :`. If you can find a way to hit this code then I will hold off on removing it

jackgerrits: Nevermind, it is possible to access it. Will close this PR.

PR #2380, opened 2020-04-01
Reorder path in RunTests so that local build takes priority over installed VW

jackgerrits:


jackgerrits:
Sounds useful, I would like to avoid deeper changes to the Perl script though... I'd like to port this to Python ideally

PR #2381, opened 2020-04-03
Build and upload documentation for each release

jackgerrits:


.github/workflows/build_docs.yml
```diff
@@ -60,6 +65,12 @@ jobs:
     # The upload step should only be run on the main repository.
     if: github.repository == 'VowpalWabbit/vowpal_wabbit'
     steps:
+      - name: Set folder name to latest if push
+        if: github.event_name == 'push'
+        run: echo "::set-env name=FOLDER_NAME::latest"
+      - name: Set folder name to version if release
+        if: github.event_name == 'release'
+        run: echo "::set-env name=FOLDER_NAME::$(echo ${GITHUB_REF:10})"
```
lalo, line 73: would be nice to get the vw version instead of the github hash, what would this imply in terms of extra work? might not be worth it

lalo: maybe we could have the vw version in a version.h file (maybe we already do this) and get it using bash and be diligent on release process to manage that version

lalo: some grep or awk

jackgerrits: This will extract the version from the tag, it wont be a hash. Seems counter intuitive I agree

jackgerrits: The reason this works is because it cuts out the first 10 chars of the ref "refs/tags/"

jackgerrits: This may be clearer though:
```yaml
- name: Get the version
  id: get_version
  run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
```

https://github.community/t5/GitHub-Actions/How-to-get-just-the-tag-name/td-p/31595

jackgerrits: This is the same way we get the version in the docker-images repo

lalo: got it, got fooled by the code

PR #2385, opened 2020-04-04
Migrate to CodeCov for code coverage

jackgerrits:
- Add GitHub workflow to build and upload coverage
- Add Codecov config file
- Update coverage badge


Smit-create:
@jackgerrits Can we have a code coverage bot commenting on the thread after every commit that passes the test to show the author of the PR about the coverage of diff?

jackgerrits:
That's the plan. CodeCov is having issues on Azure DevOps though: https://community.codecov.io/t/codecov-uploads-from-azure-pipelines-are-failing-with-build-numbers-do-not-match/1100/14

Smit-create:
```ruby
coverage:
  status:
    project:
      default:
        target: 0%
        threshold: 100%

    patch: false

  precision: 3

comment:
  layout: "diff"
  behavior: default
  require_changes: true

```
I am not sure as I haven't work on this before. But can we just add a file with this diff?

jackgerrits:
There is an issue on the CodeCov side, not us

PR #2394, opened 2020-04-08
[Slates] Move the slates namespace to be nested under VW

jackgerrits:


vowpalwabbit/slates_label.cc
```diff
@@ -58,12 +59,8 @@ void cache_label(void* v, io_buf& cache)
 {
   char* c;
   slates::label* ld = static_cast<slates::label*>(v);
-  size_t size = sizeof(ld->type)
-      + sizeof(ld->weight)
-      + sizeof(ld->labeled)
-      + sizeof(ld->cost)
-      + sizeof(ld->slot_id)
-      + sizeof(uint32_t) // Size of probabilities
+  size_t size = sizeof(ld->type) + sizeof(ld->weight) + sizeof(ld->labeled) + sizeof(ld->cost) + sizeof(ld->slot_id) +
```
lokitoth, line 62: [Minor][Formatting] Is this the result of a Clang-Format? Cutting it there seems a bit weird, plus there is some oddness between spacing between `+` and `sizeof(unit32_t)`. Is it possible to compute the size of the label based on the label struct?

jackgerrits: Probably clang-format.. No you cannot use the size of the label struct because of the v_array<action_score>

Issue #2369, opened 2020-03-25
Introducing flake8 in python


Smit-create:
Flake8 is described as a tool for style guide enforcement, to help us find errors in our code or warnings about style.

Running the `flake8` on `pyvw` and `sklearn_vw`, I got the following warnings:
```
vowpalwabbit/pyvw.py:8:1: E302 expected 2 blank lines, found 1
vowpalwabbit/pyvw.py:34:13: E731 do not assign a lambda expression, use a def
vowpalwabbit/pyvw.py:36:13: E731 do not assign a lambda expression, use a def
vowpalwabbit/pyvw.py:49:80: E501 line too long (88 > 79 characters)
vowpalwabbit/pyvw.py:103:80: E501 line too long (95 > 79 characters)
vowpalwabbit/pyvw.py:109:9: E741 ambiguous variable name 'l'
vowpalwabbit/pyvw.py:111:13: E741 ambiguous variable name 'l'
vowpalwabbit/pyvw.py:151:80: E501 line too long (125 > 79 characters)
vowpalwabbit/pyvw.py:159:80: E501 line too long (121 > 79 characters)
vowpalwabbit/pyvw.py:161:31: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
vowpalwabbit/pyvw.py:193:80: E501 line too long (85 > 79 characters)
vowpalwabbit/pyvw.py:204:80: E501 line too long (81 > 79 characters)
vowpalwabbit/pyvw.py:219:80: E501 line too long (113 > 79 characters)
vowpalwabbit/pyvw.py:305:80: E501 line too long (100 > 79 characters)
vowpalwabbit/pyvw.py:310:80: E501 line too long (86 > 79 characters)
vowpalwabbit/pyvw.py:321:80: E501 line too long (89 > 79 characters)
vowpalwabbit/pyvw.py:327:80: E501 line too long (94 > 79 characters)
vowpalwabbit/pyvw.py:337:80: E501 line too long (94 > 79 characters)
vowpalwabbit/pyvw.py:339:80: E501 line too long (139 > 79 characters)
vowpalwabbit/pyvw.py:340:80: E501 line too long (88 > 79 characters)
vowpalwabbit/pyvw.py:347:30: E701 multiple statements on one line (colon)
vowpalwabbit/pyvw.py:349:80: E501 line too long (124 > 79 characters)
[...]

PR #2395, opened 2020-04-09
Use black in python

Smit-create:
Related to discussions in #2369
Relevant comments [#2369(comment)](https://github.com/VowpalWabbit/vowpal_wabbit/issues/2369#issuecomment-603838869), [#2369(comment)](https://github.com/VowpalWabbit/vowpal_wabbit/issues/2369#issuecomment-603809724)

* python
  * Code-quality related


Smit-create:
Should I use the same for `sklearn_vw.py`?

Smit-create:
@jackgerrits This looks good to merge as I see work in progress in #2368 which may lead to conlficts so I will be keeping `sklearn_vw.py` for later

jackgerrits:
- Why choose black over flake8? What are the reasons either way?
- What is the chosen line width requirement? I think we should follow PEP-8 and use 79 (https://www.python.org/dev/peps/pep-0008/#id19)

jackgerrits:
Hey @Smit-create, I think enforcing style and adding linting is helpful but I think it's important to include tooling both to run locally and in CI that makes this not too much of a burden on contributors and to make sure things stay green.

Sharad24:
> Hey @Smit-create, I think enforcing style and adding linting is helpful but I think it's important to include tooling both to run locally and in CI that makes this not too much of a burden on contributors and to make sure things stay green.

A [pre-commit hook](https://pre-commit.com/) can be added so that whenever `git commit` is run, it would run commands provided in `.pre-commit-config.yaml`. This could act as a tool locally. Alternatively, `pre-commit run --all-files` can also be run. Here, we could add flake8 or black or both, whatever's decided.

And for the CI, a link-check stage can be added. Where it checks the files for the same commands that have to be run in the pre-commit stage.

python/vowpalwabbit/pyvw.py
```diff
@@ -2,11 +2,13 @@
 """Python binding for pylibvw class"""
 
 from __future__ import division
-import pylibvw
+from pylibvw import vw as pylibvw_vw, example as pylibvw_example
```
jackgerrits, line 5: This is a breaking change - any script that uses `pyvw.pylibvw.<>` will no longer work, and since this is the only way to get at label and prediction types right now we cannot change this.

jackgerrits:
Pre-commit hook is an option but I don't want to require that to run all of the time. A lot of the time the Python code is not even touched.

@Smit-create if you revert the import changes then I think we can merge the formatting fixes here.

Smit-create:
> if you revert the import changes then I think we can merge the formatting fixes here.

Sorry for the delayed response, I will make this change as soon as possible..

jackgerrits:
Great, thanks @Smit-create! I'll merge as soon as the CI is done.

PR #2398, opened 2020-04-10
Update master branch to be consistent with 8.8.1 release

jackgerrits:


lokitoth:
Since it is likely that 8.8.1 will be reporting (incorrectly) 8.8.0 as the version when queried, we should consider bumping to 8.8.2 and doing another point release.

jackgerrits:
Why would 8.8.1 be reporting incorrectly?

lokitoth:
Oh, I see. I thought we pushed the release branch with version being 8.8.0, just marked it as 8.8.1. Should we file a bug on the release script about making sure we update master, too?

jackgerrits:
We don't have a release script. There was a basic one but it failed in a surprising way when John used it so we haven't used it since

PR #2399, opened 2020-04-10
Move deletion logic to destructor of vw

jackgerrits:


vowpalwabbit/global_data.cc
```diff
@@ -437,4 +438,45 @@ vw::vw()
...
+  {
+    if (sd->ldict)
+    {
+      sd->ldict->~namedlabels();
+      free(sd->ldict);
+    }
+    free(sd);
+  }
+  for (size_t i = 0; i < final_prediction_sink.size(); i++)
+    if (final_prediction_sink[i] != 1)
```
lalo, line 476: nit: What's our bracket style, feel like this is unnecessary might as well change it up in this pr? also for each loop?

vowpalwabbit/global_data.cc
```diff
@@ -437,4 +438,45 @@ vw::vw()
   sd->multiclass_log_loss = 0;
   sd->holdout_multiclass_log_loss = 0;
 }
-IGNORE_DEPRECATED_USAGE_END
\ No newline at end of file
+IGNORE_DEPRECATED_USAGE_END
+
+vw::~vw()
```
lalo, line 443: did you give this a pass to check if we are missing something that got added to the class but not to the old parse args destructor logic? feel that causes room for mismatch

vowpalwabbit/global_data.cc
jackgerrits: Kind of depends who you ask. I prefer brackets always though personally. Good point, can definitely be a for each.

vowpalwabbit/global_data.cc
jackgerrits: Yeah I did a pass and deprecated an unused variable. `searchstr` seems fishy to me but don't want to deal with it in this PR

PR #2400, opened 2020-04-10
Overhaul initialize function interface

jackgerrits:
- Migrate `initialize` and friends to static members on `vw` object
- Old `VW::initialize` moved to compat mode
- Existing code should still work as evidenced by the fact I did not update the unit tests


vowpalwabbit/parse_args.cc
```diff
@@ -1897,8 +1762,12 @@ void finish(vw& all, bool delete_all)
     finalize_regressor_exception_thrown = true;
   }
 
-  if (delete_all)
+  // If this object is managed with a unique pointer then we should not delete.
+  // This can happen if a user calls the wrong API or an exception is thrown and the user is using the new api.
+  if (delete_all && !all.is_managed_by_unique)
```
lalo, line 1767: Need to think it over, but do we actually need is_managed_by_unique? If the user is using unique_ptr impl then they will call into the correct finish which we control fully by sending delete_all as false and if they aren't is_managed_by_unique will be set to false BUT this leaves room for someone distracted to grab the unique_ptr, get the raw pointer (forget to set is_managed_by_unique to false) and call into this function. Not exactly sure if we could do anything about it but just thinking out loud.

jackgerrits: Yeah, this situation is kind of annoying. But I think it's just too easy to get it wrong without a check like this.

Consider:
Someone gets their unique_ptr, gets the raw pointer and goes about using it for all of the different APIs. Because, well, the APIs take raw pointers. Then they call finish just like they have always done, and (without this check) it frees the vw object. Then the unique_ptr goes out of scope and delete is called again and we have a double free.

lalo: But in your scenario the user only gets protected if they are using your API that is the old api wrapping the new api (the one that sets is_managed_by_unique to false) and if they use the new api that returns unique_ptr but later do get() and forget about calling the correct finish it ends up being the same situation (no protection i think?).

jackgerrits: If they use the old then they must call `finish(vw*)`, there are no other options. They don't have a unique_ptr to call the other variant. 

> if they use the new api that returns unique_ptr but later do get() and forget about calling the correct finish it ends up being the same situation (no protection i think?).

This is the situation I am trying to guard against. They should be protected against a double free here.

Because if `is_managed_by_unique` is true, then they cannot get into the block where the object is deleted.

lalo: whoops, you are correct since it's the default value 👍

lalo:
@jackgerrits is it worthwhile to update a test or somewhere to use the new api style?

jackgerrits:
I want to update all of the tests to use it. Then perhaps we can have a set of compatibility tests that use old behavior until they are removed.

peterychang:
Do we want to wrap vw in a unique_ptr if the interface is going to be in C? It seems like create/destroy with raw pointers should be the default behavior

peterychang:
This change looks ok, but it feels like its going in the wrong direction to me. We should expose a C++-style initialize that builds on top of the current initialize, this would mimic the behavior we'll eventually want where
`C++ library -> C API -> C++ wrapper`, where the current initialize is a placeholder for the C API and the unique_ptr initialize is a placeholder for the C++ wrapper.

Also, I don't think the vw object should be aware of its own managed status. If a user wants to use the C++ interface, they have to make a deliberate effort to use it. In this case, they should be aware of the implications of using a unique pointer, and ensure they are not calling incompatible finish functions.

PR #2401, opened 2020-04-11
Fixes passing list of strings as examples in predict and learn

### LINKED ISSUE #2397: Passing a list of strings to pyvw.predict fails in an unclear way
#### Describe the bug
Passing a list of string to predict or learn will cause a cryptic crash. The string needs to be joined with newlines for it to work. This should either work as expected or fail in a clear way.

#### To Reproduce
```python
from vowpalwabbit import pyvw
examples = ["| a", "| b"]
vw = pyvw.vw()

vw.predict(examples)
```

#### Expected behavior
This should either work or a clear error should be raised.

#### Observed Behavior

Cryptic error message:
```
  File "/home/jackgerrits/.local/lib/python3.6/site-packages/vowpalwabbit-8.8.0-py3.6-linux-x86_64.egg/vowpalwabbit/pyvw
.py", line 424, in predict
    pylibvw.vw.predict_multi(self, ec)
TypeError: No registered converter was able to produce a C++ rvalue of type boost::shared_ptr<example> from this Python
object of type str
```

#### Environment
What version of VW did you use? Latest master
What OS or language did you use? WSL, Ubuntu

Smit-create:
Fixes #2397 
After fixing:
```python
>>> from vowpalwabbit import pyvw
>>> examples = ["| a", "| b"]
>>> vw = pyvw.vw(quiet=True)
>>> vw.learn(examples)
  ...
TypeError: Expecting a mutiline Learner.
>>> vw.predict(examples)
  ...
TypeError: Expecting a multiline Learner.
>>> vw = pyvw.vw(quiet=True, cb_adf=True)
>>> vw.learn(examples)
>>> vw.predict(examples)
[0.0, 0.0]
```


python/tests/test_pyvw.py
```diff
@@ -20,6 +20,9 @@ def test_learn_predict(self):
         assert init == 0
         self.model.learn(ex)
         assert self.model.predict(ex) > init
+        ex = ["| a", "| b"]
+        assert self.model.predict(ex) == [self.model.predict(ex[0]), self.model.predict(ex[1])]
+        self.model.learn(ex)
```
jackgerrits, line 25: This is incorrect, perhaps the issue was somewhat unclear. I believe allowing a list to be processed in this way because of multiline examples. If it is a list of strings then we should assume it is a multiline example

python/vowpalwabbit/pyvw.py
```diff
@@ -279,6 +279,12 @@ def parse(self, str_ex, labelType=pylibvw.vw.lDefault):
         ec : list
             list of examples parsed
         """
+        if isinstance(str_ex, example) and getattr(str_ex, 'setup_done', None):
+            return str_ex
+
+        elif isinstance(str_ex, list):
+            if all([getattr(ex, 'setup_done', None) for ex in str_ex]):
+                return str_ex
```
jackgerrits, line 287: I think passing a non-string to parse should be a type error

python/vowpalwabbit/pyvw.py
```diff
@@ -366,19 +372,34 @@ def learn(self, ec):
             ec = self.parse(ec)
             new_example = True
 
+        elif isinstance(ec, list):
+            if self._is_multiline():
+                ec = self.parse(ec)
+            else:
+                ec = list(map(self.parse, ec))
```
jackgerrits, line 379: I think this should be a type error as per my first comment about lists of strings in single-line mode

jackgerrits:
Thanks for working on this! I think allowing lists of examples for singleline examples is troublesome. So I think we should scope this as follows for now:

If list is passed to learn/predict:
- Check if it is a list of strings and learner is multiline. If so, either type error or parse. Probably parse
- If it is a list of examples, type error if singleline learner otherwise proceed
- If list of strings and single line, type error

python/vowpalwabbit/pyvw.py
```diff
@@ -406,7 +424,13 @@ def predict(self, ec, prediction_type=None):
             new_example = True
 
         # If a string was given, parse it before passing to learner.
-        if isinstance(ec, str):
+        elif isinstance(ec, str):
```
jackgerrits, line 427: Shouldn't this just be if not elif?
```suggestion
        if isinstance(ec, str):
```

Smit-create: IMO, it is good to use elif as we are sequentially checking whether it is dict or str or  list

jackgerrits:
Thanks for the patch @Smit-create!

PR #2407, opened 2020-04-15
Cb explore adf rnd

pmineiro:
This is RND style contextual bandit exploration.  tl;dr: predict a fake target whose true value is zero, and use MSE(prediction) as a confidence interval.

Works well on banditized covertype.


vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -0,0 +1,307 @@
...
+  // Should be called through cb_explore_adf_base for pre/post-processing
+  void predict(VW::LEARNER::multi_learner& base, multi_ex& examples) { predict_or_learn_impl<false>(base, examples); }
+  void learn(VW::LEARNER::multi_learner& base, multi_ex& examples) { predict_or_learn_impl<true>(base, examples); }
+
+ private:
+  template <bool is_learn>
+  void predict_or_learn_impl(VW::LEARNER::multi_learner& base, multi_ex& examples);
+
+  std::vector<float> bonuses;
+  std::vector<float> initials;
```
jackgerrits, line 48: It looks like you've structured the class as:
- private members
- public methods
- private methods

These private members should probably be in the first block right?

vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -0,0 +1,307 @@
...
+  template<bool> void make_fake_rnd_labels(multi_ex&);
+  template<bool> void restore_labels(multi_ex&);
+  template<bool> void base_learn_or_predict(VW::LEARNER::multi_learner&, multi_ex&, uint32_t);
+};
+
+void cb_explore_adf_rnd::zero_bonuses(multi_ex& examples)
+{
+  bonuses.clear();
+  bonuses.reserve(examples.size());
+  for (auto* ec: examples) { (void) ec; bonuses.push_back(0.f); }
```
jackgerrits, line 67: I think if you aren't going to do anything with examples then you should just use an old style for loop

vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -0,0 +1,307 @@
...
+}
+
+void cb_explore_adf_rnd::accumulate_bonuses(multi_ex& examples)
+{
+  v_array<ACTION_SCORE::action_score>& preds = examples[0]->pred.a_s;
```
jackgerrits, line 72:
```suggestion
const auto& preds = examples[0]->pred.a_s;
```

vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -0,0 +1,307 @@
...
+  for (const auto& p : preds)
+    {
```
jackgerrits, line 74: Throughout this PR the indentation is a bit off:

```suggestion
  for (const auto& p : preds)
  {
```

vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -0,0 +1,307 @@
...
+      float score = p.score - initials[p.action];
+      bonuses[p.action] += score * score;
+    }
+}
+
+void cb_explore_adf_rnd::finish_bonuses()
+{
+  for (auto& b: bonuses) { b = sqrt(b / numrnd); }
```
jackgerrits, line 82:
```suggestion
  for (auto& b: bonuses) { b = std::sqrt(b / numrnd); }
```

vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -0,0 +1,307 @@
...
+}
+
+void cb_explore_adf_rnd::compute_ci(v_array<ACTION_SCORE::action_score>& preds,
+                                    float maxbonus)
+{
+  const float eulergamma = 0.57721566490153286;
```
jackgerrits, line 88:
```suggestion
  constexpr float eulergamma = 0.57721566490153286;
```

vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -0,0 +1,307 @@
...
+    }
+}
+
+void cb_explore_adf_rnd::finish_bonuses()
+{
+  for (auto& b: bonuses) { b = sqrt(b / numrnd); }
+}
+
+void cb_explore_adf_rnd::compute_ci(v_array<ACTION_SCORE::action_score>& preds,
+                                    float maxbonus)
```
jackgerrits, line 86:
```suggestion
                                    float max_bonus)
```

vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -0,0 +1,307 @@
...
+      return (   ec->l.cb.costs.size() == 1
+              && ec->l.cb.costs[0].cost != FLT_MAX
+              && ec->l.cb.costs[0].probability > 0);
+    }
+}
+
+template <bool is_learn>
+void cb_explore_adf_rnd::save_labels(multi_ex& examples)
+{
+  if (is_learn)
```
jackgerrits, line 105: Given that several of these functions do nothing in the is_learn = false case does it make sense to uplift this check before calling. It's a little misleading to have the function be a noop in my mind

vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -0,0 +1,307 @@
...
+            {
+              save_class.cost = ec->l.cb.costs[0].cost;
+              save_class.probability = ec->l.cb.costs[0].probability;
+              break;
+            }
+        }
+    }
+}
+
+namespace
```
jackgerrits, line 119: What's the rationale behind the anonymous namespaces?

vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -0,0 +1,307 @@
...
+        LazyGaussianWeight(float _sigma) : count(0), sigma(_sigma) { }
+        float operator[](uint64_t index) const
+          {
+            ++count;
+            return sigma * merand48_boxmuller(index);
+          }
+
+        uint32_t get_fc() const
+          {
+            return std::max((uint32_t) 1, count);
```
jackgerrits, line 137:
```suggestion
            return std::max(static_cast<uint32_t>(1), count);
```

vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -0,0 +1,307 @@
...
+          }
+    };
+}
+
+float cb_explore_adf_rnd::get_initial_prediction(example* ec)
+{
+  LazyGaussianWeight w(sqrtinvlambda);
+  float rv = GD::inline_predict(w,
```
jackgerrits, line 145: Clang-format is probably going to format this differently when run

vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -0,0 +1,307 @@
...
+      get_initial_predictions(examples, 1 + id);
+      make_fake_rnd_labels<is_learn>(examples);
+      base_learn_or_predict<is_learn>(base, examples, 1 + id);
+      accumulate_bonuses(examples);
+    }
+  finish_bonuses();
+  restore_labels<is_learn>(examples);
+  base_learn_or_predict<is_learn>(base, examples, 0);
+
+  v_array<ACTION_SCORE::action_score>& preds = examples[0]->pred.a_s;
```
jackgerrits, line 232:
```suggestion
  auto& preds = examples[0]->pred.a_s;
```

vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -0,0 +1,307 @@
...
+  float alpha = 0.;
+  float invlambda = 0.;
+  uint32_t numrnd = 1;
+  config::option_group_definition new_options("Contextual Bandit Exploration with Action Dependent Features");
+  new_options
+      .add(make_option("cb_explore_adf", cb_explore_adf_option)
+               .keep()
+               .help("Online explore-exploit for a contextual bandit problem with multiline action dependent features"))
+      .add(make_option("epsilon", epsilon).keep().allow_override().help("minimum exploration probability"))
+      .add(make_option("rnd", numrnd).keep().help("rnd based exploration"))
```
jackgerrits, line 254: Does `rnd` have special meaning? If not, I would say just expand it. Is it rand or round?

vowpalwabbit/rand48.cc
```diff
@@ -26,3 +27,18 @@ float merand48(uint64_t& initial)
 }
 
 float merand48_noadvance(uint64_t v) { return merand48(v); }
+
+float merand48_boxmuller(uint64_t& index)
+{
+  static float x1 = 0.0;
```
jackgerrits, line 33: These being null makes this non-threadsafe. They are local state required for this right? I'm wondering if there is another way to do this to avoid that constraint

vowpalwabbit/rand48.cc
```diff
@@ -26,3 +27,18 @@ float merand48(uint64_t& initial)
...
+  static float x2 = 0.0;
+  static float temp = 0.0;
+  do
+  {
+    x1 = 2.0f * merand48(index) - 1.0f;
+    x2 = 2.0f * merand48(index) - 1.0f;
+    temp = x1 * x1 + x2 * x2;
+  } while ((temp >= 1.0) || (temp == 0.0));
+  temp = sqrtf((-2.0f * logf(temp)) / temp);
```
jackgerrits, line 42:
```suggestion
  temp = std::sqrt((-2.0f * logf(temp)) / temp);
```

vowpalwabbit/cb_explore_adf_rnd.cc
pmineiro: Whoops.  I hit this with the clang formatter in VS to get consistency.

vowpalwabbit/cb_explore_adf_rnd.cc
pmineiro: In the RL literature RND stands for "Random Network Distillation" ...

vowpalwabbit/rand48.cc
pmineiro: They don't have to be static.  I copied this from another file into here to reuse the code, and I didn't notice that.

vowpalwabbit/cb_explore_adf_rnd.cc
pmineiro: Done

vowpalwabbit/cb_explore_adf_rnd.cc
pmineiro: Switched to std::assign

vowpalwabbit/cb_explore_adf_rnd.cc
pmineiro: This way made the logic from 223-234 more readable.  Other ways involved sprinkling conditionals and/or duplicating code blocks.

vowpalwabbit/cb_explore_adf_rnd.cc
pmineiro: I want the equivalent of a static function in C, i.e., I don't want anybody outside this file using LazyGaussianWeight.

vowpalwabbit/cb_explore_adf_rnd.cc
pmineiro: Done

vowpalwabbit/cb_explore_adf_rnd.cc
pmineiro: Done

vowpalwabbit/cb_explore_adf_rnd.cc
jackgerrits: Yeah I see what you mean. I thought about that when I saw it too. Makes sense.

vowpalwabbit/cb_explore_adf_rnd.cc
jackgerrits: Awesome sounds good.

vowpalwabbit/cb_explore_adf_rnd.cc
jackgerrits: Okay cool. Definitely put that in the file somewhere

PR #2410, opened 2020-04-16
Implement 1-slot continuous CB

ajay0:
Includes
- A new label (and parser) suitable for 1-slot scenario
- New base learner implementing the core algorithm
- Unit tests for label parser and e2e tests (using Runtests)


vowpalwabbit/contcb.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include "reductions.h"
+
+namespace CONTCB
```
rajan-chari, line 5: >CONTCB [](start = 10, length = 6)

VW format for namespaces is getting standardized.  

I would suggest the following:
namespace VW { namespace continuous_cb { }

vowpalwabbit/gd.cc
```diff
@@ -21,6 +21,7 @@
 #include "accumulate.h"
 #include "reductions.h"
 #include "vw.h"
+#include "contcb.h"
```
rajan-chari, line 24: >contcb [](start = 10, length = 6)

needed here?

ajay0: Oops, was accidental. Not needed.

ajay0:
### Doubts

1. If you notice the output when vw is run as an executable, all the stats are `0/na`. This is because I don't use any of that info (present mostly in `struct shared_data`). Considering that this algorithm would be run in library mode for almost every case I can think of, would it still be a good idea to handle this?

2. Currently the prediction is composed of two scalars: one is an `action` and the other is a `radius` essentially defining the exploration distribution around that `action`. In the two cases where predictions are required to be printed: `--predictions` and `--audit`, I print both these scalars. Is this behavior fine?

3. What's the right way to delete predictions (in this case a vector of scalars). Do it in the `learner`'s `finish_example()`?

### Fixes

1. I just realized that I wrongly assumed
	- `--audit` must print only for test examples
	- `--predictions` must output predictions only for test examples.

Is it okay to push a fix for these in this PR itself?

2. The `--invert_hash` option isn't supported currently as I was confused regarding something which is resolved now. Should I update the PR with changes or keep feature additions for a later PR?

ajay0:
### Windows CI Failure

Windows CI fails due to 2 tests in `RunTests` using bash commands which the C# tests have trouble with. I could either
- Skip the C# tests for these (by editing `RunTests.tt`)
- Add a data file to get around using these bash commands.

From a glance at `RunTests.tt`, some tests using bash commands are skipped (even those where a data file could have helped get around using bash commands). Should I do the same?

jackgerrits:
It's up to you if you would rather skip the test or make it so that it can run with the C# wrapper.

jackgerrits:
> 
> ### Doubts
> 
>1. If you notice the output when vw is run as an executable, all the stats are `0/na`. This is because I don't use any of that info (present mostly in `struct shared_data`). Considering that this algorithm would be run in library mode for almost every case I can think of, would it still be a good idea to handle this?

I think it is important to support command line for each reduction we add. Is that possible?
 
>2. Currently the prediction is composed of two scalars: one is an `action` and the other is a `radius` essentially defining the exploration distribution around that `action`. In the two cases where predictions are required to be printed: `--predictions` and `--audit`, I print both these scalars. Is this behavior fine?

Yes, but it is important to add a wiki page describing how to use the reduction, labels, and how to interpret its output.


>3. What's the right way to delete predictions (in this case a vector of scalars). Do it in the `learner`'s `finish_example()`?

You should be able to set `delete_prediction` on the `vw` object to handle cleaning that up.

> ### Fixes
> 
> 1. I just realized that I wrongly assumed
>        
>     * `--audit` must print only for test examples
>     * `--predictions` must output predictions only for test examples.
> 
> 
> Is it okay to push a fix for these in this PR itself?

If you change those it is a drastic change of behavior from the current implementation, which I am wary of. @rajan-chari , @JohnLangford thoughts?

>     1. The `--invert_hash` option isn't supported currently as I was confused regarding something which is resolved now. Should I update the PR with changes or keep feature additions for a later PR?

It would be good to support it in this PR.

ajay0:
>If you change those it is a drastic change of behavior from the current implementation, which I am wary of. @rajan-chari , @JohnLangford thoughts?

@jackgerrits  I don't wish to change from the existing behavior. Existing behavior looks like `--predictions` and `--audit` is actionable for both train and test examples while what I've implemented makes it actionable only for test examples. Thus I wanted to push an update making the behavior conform to the existing implementation.

jackgerrits:
Oh of course, sounds good.

ajay0:
> I think it is important to support command line for each reduction we add. Is that possible?

@jackgerrits from what I understand, supporting command line is equivalent to supporting an offline learning setting where an algorithm must learn from a static dataset file (like a .dat used for CB). CB algorithms get this by working with estimators like IPS, but for the continuous setting, IPS type scoring is not feasible. So its not clear how such an offline dataset file can be prepared and made use of, for a continuous problem.

I've been able to fix all other issues we've discussed. Windows tests pass now as well. Could you let me know if there's anything else I've missed out?

vowpalwabbit/cbzo.cc
```diff
@@ -200,17 +171,42 @@ void print_audit_features(vw& all, example& ec)
...
+inline float close_greater_value(float x) {
+  if (x != 0.f)
+    return nextafter(x, numeric_limits<float>::infinity());
+  return 1e-5;
+}
+
+// Returns a value close to x and lesser than it
+inline float close_lesser_value(float x) {
+  if (x != 0.f)
+    return nextafter(x, -numeric_limits<float>::infinity());
```
jackgerrits:
```suggestion
    return std::nextafter(x, -std::numeric_limits<float>::infinity());
```

vowpalwabbit/cbzo.cc
```diff
@@ -200,17 +171,42 @@ void print_audit_features(vw& all, example& ec)
   GD::print_features(all, ec);
 }
 
+// Returns a value close to x and greater than it
+inline float close_greater_value(float x) {
+  if (x != 0.f)
+    return nextafter(x, numeric_limits<float>::infinity());
```
jackgerrits:
```suggestion
    return std::nextafter(x, std::numeric_limits<float>::infinity());
```

ajay0: Fixed.

vowpalwabbit/cbzo.cc
ajay0: Fixed

vowpalwabbit/cbzo.cc
```diff
@@ -0,0 +1,373 @@
...
+
+inline std::string get_pred_repr(example& ec)
+{
+  return continuous_actions::to_string(ec.pred.pdf, false, std::numeric_limits<float>::max_digits10);
+}
+
+void print_audit_features(vw& all, example& ec)
+{
+  if (all.audit) all.print_text_by_ref(all.stdout_adapter.get(), get_pred_repr(ec), ec.tag);
+  fflush(stdout);
```
olgavrou, line 167: is this `fflush` here needed?

vowpalwabbit/cbzo.cc
```diff
@@ -0,0 +1,373 @@
...
+  // Note: print_features() declaration was brought to gd.h so it can be used here. If it's
```
olgavrou, line 168: is this note still relevant?

olgavrou (commented):
Apart from a couple of questions that I have this looks good to me

vowpalwabbit/cbzo.cc
ajay0: Ah, doesn't look like it is. Fixed.

jackgerrits:
Have you had a chance to work on a wiki page for this reduction?

vowpalwabbit/cbzo.cc
ajay0: It's insignificant. Removed.

ajay0:
> Have you had a chance to work on a wiki page for this reduction?

@jackgerrits I'm expecting it to be ready tomorrow.

olgavrou (approved):

PR #2411, opened 2020-04-20
Added `--local` option (for loading active learning/damon models)

ZechyW:
When a model is saved in active learning mode, it resumes active learning when loaded (i.e., it starts listening for input on a port and ignoring local sources).
This option allows the user to change this behaviour (so that the Python library can interact with the loaded model, for example).

### DIFF
--- a/vowpalwabbit/global_data.h
+++ b/vowpalwabbit/global_data.h
@@ -527,7 +527,7 @@ struct vw
 
   bool stdin_off;
 
-  bool local;  // If a model was saved in daemon or active learning mode, start it in local mode instead
+  bool no_daemon = false;  // If a model was saved in daemon or active learning mode, force it to accept local input when loaded instead.
 
   // runtime accounting variables.
   float initial_t;
--- a/vowpalwabbit/global_data.h
+++ b/vowpalwabbit/global_data.h
@@ -527,6 +527,8 @@ struct vw
 
   bool stdin_off;
 
+  bool local;  // If a model was saved in daemon or active learning mode, start it in local mode instead
+
   // runtime accounting variables.
   float initial_t;
   float eta;  // learning rate control.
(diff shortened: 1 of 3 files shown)


vowpalwabbit/global_data.h
```diff
@@ -525,7 +525,9 @@ struct vw
   VW_DEPRECATED("This is unused and will be removed")
   char* program_name;
 
-  bool stdin_off;
+  bool stdin_off;
+
+  bool local;  // If a model was saved in daemon or active learning mode, start it in local mode instead
```
jackgerrits, line 530: This needs to be initialized with a default value

vowpalwabbit/parse_args.cc
```diff
@@ -438,6 +438,7 @@ input_options parse_source(vw& all, options_i& options)
                   "use gzip format whenever possible. If a cache file is being created, this option creates a "
                   "compressed cache file. A mixture of raw-text & compressed inputs are supported with autodetection."))
       .add(make_option("no_stdin", all.stdin_off).help("do not default to reading from stdin"))
+      .add(make_option("local", all.local).help("Do not start in daemon or active learning mode (for loaded models)"))
```
jackgerrits, line 441: Can you please make the option name more specific, so that it's very clear to users it only affects daemon/active

ZechyW: Got it, how do you feel about `--force-local-input` or `--no-daemon`?

jackgerrits: I'd be okay with `--no-daemon`

ZechyW:
Have made the requested changes -- Thanks for checking, @jackgerrits, and let me know if you need me to modify it any further!

ZechyW:
@peterychang: Didn't see any other way to revert to local input for an active learning model once it's saved, but let me know if there was already an option for it or if there's a better way of doing it!

peterychang:
I agree. I just find it silly that we actually need this change. IIRC @lokitoth you were doing some work to allow alteration of the model-encoded options, right? Would that work help with this?

ZechyW:
A general way to change the saved model options sounds very useful indeed -- Will keep an eye out for it.  Thanks for the update!

PR #2421, opened 2020-04-24
Fix Slates slot assignment

jackgerrits:
Slot assignment should occur whether or not the example is labelled.


vowpalwabbit/slates.cc
```diff
@@ -70,16 +70,18 @@ void slates_data::learn_or_predict(VW::LEARNER::multi_learner& base, multi_ex& e
...
+      for (const auto index : slot_action_pools[slot_index])
+      {
+        ccb_label.explicit_included_actions.push_back(index);
+      }
+
       if (global_cost_found)
       {
         ccb_label.outcome = new CCB::conditional_contextual_bandit_outcome();
         ccb_label.outcome->cost = global_cost;
         ccb_label.outcome->probabilities = v_init<ACTION_SCORE::action_score>();
```
lalo, line 83: what about this line? how does that relate to global_cost_found

jackgerrits: Probabilities are associated with the global cost

PR #2430, opened 2020-05-06
Fix text parser in Python

jackgerrits:
The text parser does not need a newline example appended to its end.


peterychang:
Both of the functions you touched are for the json parser. Did you mean the python json parser doesn't need the newline example? Why would that be?

jackgerrits:
Discussed offline but adding here for record:
This affects the `text_reader` which is only used by Python to enable parsing directly from a string.

jackgerrits:
Done

PR #2431, opened 2020-05-07
Implement two new RAII utilities and migrate nn

jackgerrits:
- `VW::scope_exit` - for exception safe resource releasing based on scopes
- `VW::swap_guard` - scope based swapping of variables for restoring object state


vowpalwabbit/nn.cc
```diff
@@ -243,7 +230,7 @@ void predict_or_learn_multi(nn& n, single_learner& base, example& ec)
     float save_final_prediction = 0;
     float save_ec_loss = 0;
 
-CONVERSE:  // That's right, I'm using goto.  So sue me.
+  CONVERSE:  // That's right, I'm using goto.  So sue me.
```
peterychang, line 233: IIRC labels need to be at the beginning of the line

jackgerrits: This was clang-format being "helpful" will revert it

vowpalwabbit/nn.cc
```diff
@@ -169,7 +152,11 @@ void predict_or_learn_multi(nn& n, single_learner& base, example& ec)
   shared_data sd;
   memcpy(&sd, n.all->sd, sizeof(shared_data));
   {
-    sd_guard(n.all, &sd);
+    auto* saved_all = n.all;
+    auto* saved_sd = n.all->sd;
+    // guard for all.sd as it is modified - this will restore the state at the end of the scope.
+    auto sd_guard = VW::scope_guard([saved_all, saved_sd]() { saved_all->sd = saved_sd; });
```
peterychang, line 158: I'm not sure if I like this or if I hate it... On one hand its a generic guard, on the other it still requires a lot of manual setup that leaves a lot of room for mistakes

jackgerrits: I think this is because the old guard wasn't actually just a guard but it also handed setting up the state of the all->sd pointer to a new value. This separates the concerns to /just/ the cleanup action

peterychang: Guards typically handle the setup and teardown of the objects being passed in; think of a lock guard. The part I dislike is that you need to manually do the setup while the teardown is automated. It feels like you could get into a lot of trouble if they're not paired together

jackgerrits: Yeah I guess it is kind of a taste thing. If there is a pattern that enables reuse of a guard then a specific object may make sense - but for one offs it seems excessive to create lots of objects with custom behavior. Renaming it to scope_exit may be better?

This pattern is inspired by the WIL (Windows internal library) scope_exit functionality: https://github.com/microsoft/wil/blob/cd51fa9d3c24e9e9f5d4cdf1a9768d68f441fc48/include/wil/resource.h#L543

vowpalwabbit/guard.h
```diff
@@ -0,0 +1,136 @@
...
+  swap_guard_impl(const swap_guard_impl&) = delete;
+  swap_guard_impl& operator=(const swap_guard_impl&) = delete;
+
+  swap_guard_impl(swap_guard_impl&& other) noexcept
+      : _original_location(other._original_location), _value_to_swap(other._value_to_swap), _will_swap_back(other._will_swap_back)
+  {
+    other._will_swap_back = false;
+  }
+
+  swap_guard_impl& operator=(swap_guard_impl&& other) noexcept
```
peterychang, line 34: Do we need the assignment operator? Seems safer to delete this too

peterychang: Oh, I see, this is a move assignment.. Still though, do we need it?

jackgerrits: Good point - will remove

vowpalwabbit/guard.h
```diff
@@ -0,0 +1,136 @@
+// Copyright (c) by respective owners including Yahoo!, Microsoft, and
+// individual contributors. All rights reserved. Released under a BSD (revised)
+// license as described in the file LICENSE.
+#pragma once
+
+#include <utility>
+#include "future_compat.h"
+
+// While it is a general concept this specific implemenation of scope based cleanup was inspired by
+// https://github.com/microsoft/wil/blob/cd51fa9d3c24e9e9f5d4cdf1a9768d68f441fc48/include/wil/resource.h#L543
```
peterychang, line 10: This is in the wrong file

vowpalwabbit/guard.h
```diff
@@ -0,0 +1,136 @@
...
+ public:
+  swap_guard_impl(T& original_location, T& value_to_swap) noexcept : _original_location(original_location), _value_to_swap(value_to_swap)
+  {
+    std::swap(_original_location, _value_to_swap);
+  }
+
+  swap_guard_impl(const swap_guard_impl&) = delete;
+  swap_guard_impl& operator=(const swap_guard_impl&) = delete;
+
+  swap_guard_impl(swap_guard_impl&& other) noexcept
```
peterychang, line 28: the move constructors aren't necessarily safe for all of our objects. If they're not necessary, I'd rather delete them to be safe

jackgerrits: These are definitely needed as they permit the make functions.

PR #2435, opened 2020-05-13
Fix some warnings, improve example lifecycle management

jackgerrits:
- Move example cleanup to be part of the class
- Fix double free issue in object_pool
- Fix memset of non-trivially copyable types in object_pool
- Fix parser deletion logic around example pool


vowpalwabbit/example.cc
```diff
@@ -4,8 +4,39 @@
...
 #include "gd.h"
 
+example::example()
+{
+  memset(&l, 0, sizeof(polylabel));
+  memset(&pred, 0, sizeof(polyprediction));
+  tag = v_init<char>();
+}
+
+example::~example()
```
peterychang, line 17: You'll need to add an r-value constructor too, otherwise you can end up getting double frees and the like

vowpalwabbit/example.h
```diff
@@ -55,31 +56,41 @@ typedef union
...
   // input fields
   polylabel l;
 
   // output prediction
   polyprediction pred;
 
-  float weight;       // a relative importance weight for the example, default = 1
-  v_array<char> tag;  // An identifier for the example.
-  size_t example_counter;
+  float weight = 1.f;  // a relative importance weight for the example, default = 1
```
peterychang, line 73: Are all these default values sane? Do we ever have to make a distinction between "value doesn't exist" vs "value isn't set" vs "value is set to default"?

vowpalwabbit/example_predict.cc
```diff
@@ -4,18 +4,48 @@
 
 #include "example_predict.h"
 
-safe_example_predict::safe_example_predict()
+example_predict::iterator::iterator(features* feature_space, namespace_index* index)
```
peterychang, line 7: whats the difference between `example_predict` and `safe_example_predict`?

peterychang: nit: We should add a `using iterator = T*` to v_array. Then you can change `namespace_index*` to `v_array<T>::iterator` to make it easier to swap out with other standard container types in the future. Really just a minor nit though

vowpalwabbit/example_predict.h
```diff
@@ -19,36 +20,33 @@ struct example_predict
...
-    {
-      _index++;
-      return *this;
-    }
-
-    namespace_index index() { return *_index; }
-
-    bool operator==(const iterator& rhs) { return _index == rhs._index; }
-    bool operator!=(const iterator& rhs) { return _index != rhs._index; }
+    iterator(features* feature_space, namespace_index* index);
```
peterychang, line 23: nit: maybe add a quick comment that modifying the indices can invalidate these iterators

vowpalwabbit/example.h
jackgerrits: These are the values that are used in `setup_example` right now which is effectively like the constructor

vowpalwabbit/example_predict.cc
jackgerrits: `safe_example_predict` is just `example_predict` with a destructor that properly releases resources. Now that `example_predict` does this `safe_example_predict` is now marked deprecated.

vowpalwabbit/object_pool.h
```diff
@@ -112,7 +113,6 @@ struct no_lock_object_pool
 
     for (size_t i = 0; i < size; i++)
     {
-      memset(&chunk[i], 0, sizeof(T));
       new (&chunk[i]) T{};
```
peterychang, line 116: You don't need to call new on each chunk here, you're doing it a few lines earlier

vowpalwabbit/parser.cc
```diff
@@ -1048,17 +1028,21 @@ void free_parser(vw& all)
     output->currentname.delete_v();
   }
 
+  std::vector<example*> drain_pool;
```
peterychang, line 1031: It feels like there should be a way to call `for_each` on the example pool so that you don't need to do all this pushing and popping. Then again, that functionality doesn't really make sense in a resource pool data structure, so maybe this is ok...

vowpalwabbit/stagewise_poly.cc
```diff
@@ -76,9 +76,11 @@ struct stagewise_poly
 #endif  // DEBUG
 
     //synth_ec.feature_space[tree_atomics].delete_v();
-    synth_ec.indices.delete_v();
     free(sd);
     free(depthsbits);
+
+    // Intentionally do not clear the unions here.
+    synth_ec.delete_unions(nullptr, nullptr);
```
peterychang, line 83: Does this call actually do anything or is it only here to make it clear that nothing needs to be done?

vowpalwabbit/object_pool.h
jackgerrits: Right, I had added this before I realized unique_ptr<T[]> calls the constructor, thanks for the catch!

vowpalwabbit/parser.cc
jackgerrits: Yeah I thought the same thing... This will eventually go away too when the destructor is able to delete all of the contents (*cough* unions...)

vowpalwabbit/stagewise_poly.cc
jackgerrits: So originally i had debug assertions verifying that for every example the unions were deleted, or at least considered through a call like this. However, due to an issue in memory_tree it's not immediately feasible. I left this in like you said to make it clear that it is not necessary.

PR #2440, opened 2020-05-17
Probabilistic Label Tree

mwydmuch:
This PR adds Probabilistic Label Tree for logarithmic time multilabel classification. The implementation supports two types of prediction, with a marginal probability threshold as well as top-k prediction. The default prediction mode is with threshold = 0.5 for the best hamming loss. PR also includes examples of usage under `demo/plt`.


jackgerrits:
Hi @mwydmuch, thanks for open this PR! Would you be able to please add a test exercising this reduction to `test/RunTests`?

mwydmuch:
Hi @jackgerrits, I've added tests for training and both types of prediction using a multilabel dataset already included in `train-sets`. If it's not enough, I can add some more.

jackgerrits:
Thanks @mwydmuch! Once there are a couple of sign offs then we can get this in

vowpalwabbit/plt.cc
```diff
@@ -0,0 +1,382 @@
...
+#include <unordered_set>
+#include <algorithm>
+#include <queue>
+#include "reductions.h"
+#include "vw.h"
+
+using namespace VW::LEARNER;
+using namespace VW::config;
+
+namespace plt_ns
```
jackgerrits, line 19: Does this `_ns` stand for namespace? If so, then I think keeping it as just `plt` is more consistent with elsewhere,

vowpalwabbit/plt.cc
```diff
@@ -0,0 +1,382 @@
...
+{
+  uint32_t n;  // node number
+  float p;     // node probability
+
+  bool operator<(const node &r) const { return p < r.p; }
+};
+
+struct plt
+{
+  vw *all;
```
jackgerrits, line 31: Nit: pointer should be with type.

```suggestion
  vw* all;
```

vowpalwabbit/plt.cc
```diff
@@ -0,0 +1,382 @@
...
+
+  uint32_t k;     // number of labels
+  uint32_t t;     // number of tree nodes
+  uint32_t ti;    // number of internal nodes
+  uint32_t kary;  // kary tree
+
+  v_array<float> nodes_t;  // in case of sgd, this keeps stores individual t for each node
```
jackgerrits, line 38: Nit: `_t` implies this is a type and not a member variable.

vowpalwabbit/plt.cc
```diff
@@ -0,0 +1,382 @@
...
+    p.all->sd->t = p.nodes_t[n];
+    p.nodes_t[n] += ec.weight;
+  }
+  base.learn(ec, n);
+}
+
+void learn(plt &p, single_learner &base, example &ec)
+{
+  MULTILABEL::labels multilabels = ec.l.multilabels;
+  MULTILABEL::labels preds = ec.pred.multilabels;
```
jackgerrits, line 85: For futureproofing these should like be moves?

vowpalwabbit/plt.cc
mwydmuch: Fixed, note: clang-format does not fix it since `DerivePointerAlignment: true` option.

vowpalwabbit/plt.cc
mwydmuch: Yes, I think `_ns` stand for namespace, I followed convention used in two other tree-based methods: `memory_tree.cc` and `recall_tree.cc`, they also use `_ns` suffix for their namespaces.

vowpalwabbit/plt.cc
mwydmuch: I changed the name to omit the confusion.

vowpalwabbit/plt.cc
mwydmuch: Like this?
```suggestion
  MULTILABEL::labels multilabels = std::move(ec.l.multilabels);
  MULTILABEL::labels preds = std::move(ec.pred.multilabels);
```

mwydmuch:
Thank you for the review @jackgerrits!

vowpalwabbit/plt.cc
jackgerrits: Looks good!

vowpalwabbit/plt.cc
jackgerrits: Thanks for letting me know!

vowpalwabbit/plt.cc
mwydmuch: Added in all relevant places

demo/plt/rcv1x_2K.py
```diff
@@ -0,0 +1,45 @@
...
+
+# Dataset filenames
+train_data = "rcv1x_train.vw"
+test_data = "rcv1x_test.vw"
+
+print("Vowpal Wabbit PLT example on RCV1X-2K dataset with {} labels".format(k))
+
+# Download dataset
+if not os.path.exists(train_data):
+    os.system("wget http://www.cs.put.poznan.pl/mwydmuch/data/{}".format(train_data))
```
rajan-chari, line 22: Perhaps a comment here with a link to the original source of data and how it was converted to vw format?
Specific command would be helpful for conversion would be helpful.

demo/plt/amazonCat_13K.py
```diff
@@ -0,0 +1,45 @@
...
+output_model = "amazonCat_model"
+
+# Dataset filenames
+train_data = "amazonCat_train.vw"
+test_data = "amazonCat_test.vw"
+
+print("Vowpal Wabbit PLT example on AmazonCat-13K dataset with {} labels".format(k))
+
+# Download dataset
+if not os.path.exists(train_data):
```
rajan-chari, line 21: Perhaps a comment here with a link to the original source of data and how it was converted to vw format?
Specific command would be helpful for conversion would be helpful

demo/plt/wiki10_31K.py
```diff
@@ -0,0 +1,45 @@
...
+b = 30
+output_model = "wiki10_model"
+
+# Dataset filenames
+train_data = "wiki10_train.vw"
+test_data = "wiki10_test.vw"
+
+print("Vowpal Wabbit PLT example on Wiki10-31K dataset with {} labels".format(k))
+
+# Download dataset
```
rajan-chari, line 20: Perhaps a comment here with a link to the original source of data and how it was converted to vw format?
Specific command would be helpful for conversion would be helpful

rajan-chari:
Thanks for adding Demos with 3 datasets.

vowpalwabbit/plt.cc
```diff
@@ -0,0 +1,382 @@
...
+               .help("predict top-<k> labels instead of labels above threshold"));
+  options.add_and_parse(new_options);
+
+  if (!options.was_supplied("plt"))
+    return nullptr;
+
+  tree->all = &all;
+
+  // calculate number of tree nodes
+  double a = std::pow(tree->kary, std::floor(std::log(tree->k) / std::log(tree->kary)));
```
rajan-chari, line 336: all these variables could be const

rajan-chari, line 336: Brief comments describing what the intermediate (a,b,...e) values represent would be helpful

vowpalwabbit/plt.cc
```diff
@@ -0,0 +1,382 @@
...
+{
+struct node
+{
+  uint32_t n;  // node number
+  float p;     // node probability
+
+  bool operator<(const node& r) const { return p < r.p; }
+};
+
+struct plt
```
rajan-chari, line 29: Just an FYI for the future.  We started adding learn, predict methods to these structs and creating unit tests to test the basics of learn/predict.  Perhaps in a future PR?

vowpalwabbit/plt.cc
```diff
@@ -0,0 +1,382 @@
...
+template <bool threshold>
+void predict(plt& p, single_learner& base, example& ec)
+{
+  MULTILABEL::labels multilabels = std::move(ec.l.multilabels);
+  MULTILABEL::labels preds = std::move(ec.pred.multilabels);
+  preds.label_v.clear();
+
+  // split labels into true and skip (those > max. label num)
+  std::unordered_set<uint32_t> true_labels;
+  std::unordered_set<uint32_t> skip_labels;
```
rajan-chari, line 158: >skip_labels [](start = 31, length = 11)

These don't seem to be used. Is it needed?

vowpalwabbit/plt.cc
```diff
@@ -0,0 +1,382 @@
...
+  for (auto label : multilabels.label_v)
+  {
+    if (label < p.k)
+      true_labels.insert(label);
+    else
+      skip_labels.insert(label - p.k);
```
rajan-chari, line 164: >skip_labels [](start = 6, length = 11)

Should this be just a warning in non-quiet mode?

vowpalwabbit/plt.cc
```diff
@@ -0,0 +1,382 @@
...
+
+template <bool threshold>
+void predict(plt& p, single_learner& base, example& ec)
+{
+  MULTILABEL::labels multilabels = std::move(ec.l.multilabels);
+  MULTILABEL::labels preds = std::move(ec.pred.multilabels);
+  preds.label_v.clear();
+
+  // split labels into true and skip (those > max. label num)
+  std::unordered_set<uint32_t> true_labels;
```
rajan-chari, line 157: >unordered_set [](start = 7, length = 13)

This will cause allocations for every predict.  Typically we would add a container to "struct plt" here and have the container be something that would not change past some steady state.

vowpalwabbit/plt.cc
```diff
@@ -0,0 +1,382 @@
...
+    if (label < p.k)
+      true_labels.insert(label);
+    else
+      skip_labels.insert(label - p.k);
+  }
+
+  // prediction with threshold
+  if (threshold)
+  {
+    std::queue<node> node_queue;
```
rajan-chari, line 170: >queue [](start = 9, length = 5)

This will also cause allocations so would need similar treatment.  i.e. create a container in plt that reaches steady state size and does not cause allocations after that.

vowpalwabbit/plt.cc
```diff
@@ -0,0 +1,382 @@
...
+  p.all->sd->t = t;
+  p.all->sd->weighted_holdout_examples = weighted_holdout_examples;
+
+  ec.pred.multilabels = std::move(preds);
+  ec.l.multilabels = std::move(multilabels);
+}
+
+inline float predict_node(uint32_t n, single_learner& base, example& ec)
+{
+  ec.l.simple = {FLT_MAX, 1.f, 0.f};
```
rajan-chari, line 144: >l [](start = 5, length = 1)

Is there a reason a label is being updated during predict()?  Ideally predict should only deal with predictions and not labels.

vowpalwabbit/plt.cc
```diff
@@ -0,0 +1,382 @@
...
+  base.predict(ec, n);
+  return 1.0f / (1.0f + exp(-ec.partial_prediction));
```
rajan-chari, line 146: >partial_prediction [](start = 32, length = 18)

Is this meant to be partial_prediction or ec.p.<something>?

demo/plt/rcv1x_2K.py
lalo: unless this is optional, we should limit pulling in info during build or normal setup. This should be an opt-in. Haven't read if demo dir is built or just a basic demo directory.

jackgerrits: I agree with you, but I think it is okay in this directory. Just a demo dir that isn't built.

lalo:
Hey @mwydmuch thanks for the PR, any update on the questions of the CR? I might be able to help out shape this up to fast track it to be merged.

mwydmuch:
Hello @rajan-chari, thank you very much for reviewing and hello @lalo, thank you for offering your help! I'm sorry, I was quite busy last two weeks. I will address the rest of the comments by next Monday.

demo/plt/amazonCat_13K.py
mwydmuch: I've added links to a data source in each demo script + in `demo/plt/README.md` I've added note about Github repository with scripts that transform these datasets to VW format.

vowpalwabbit/plt.cc
mwydmuch: If I leave `ec.l` as multilabel and do the binary prediction it doesn't work correctly and throw a lot of warnings, so I find it necessary to change it.

vowpalwabbit/plt.cc
mwydmuch: I want to have a value of dot product here, it seemed to me that `ec.pred` behaves differently based on example data and link function.

vowpalwabbit/plt.cc
mwydmuch: Removed

vowpalwabbit/plt.cc
mwydmuch: Added warning as std::cout as in `oaa.cc`

vowpalwabbit/plt.cc
mwydmuch: Change it to a member of `plt` structure

vowpalwabbit/plt.cc
mwydmuch: Change it to a simple vector and member of `plt` structure.

vowpalwabbit/plt.cc
mwydmuch: I can change it in feature PR, can you provide an example of already modified learner?

vowpalwabbit/plt.cc
mwydmuch: Changed to consts

mwydmuch: They are used to divide the longer equation, it's hard to comment on them

lalo: is this the equation to calculate number of total nodes (or just leaves) in a complete k-ary tree? or how would you describe it exactly, mostly curious rather than related to the PR

vowpalwabbit/plt.cc
lalo: not sure if there's a better example regarding the learn/predict as part of struct https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/cb_adf.cc#L473

Regarding the tests, not sure which unit tests Rajan is referencing.

vowpalwabbit/plt.cc
mwydmuch: Yes, the number of leaves (each leaf corresponds to one label) (`plt.k`) and tree arity (`plt.kary`) are arguments. Then this part calculates the total number of nodes and number of inner nodes (nodes without labels) in a complete k-ary tree for given arity and number of leaves. The tree is complete in the sense that each level of the tree is filled with nodes from left to right. As the results, it is possible to address parent and child nodes of a given node easily:
parent = floor(node_id - 1 / tree_arity)
first_child = node_id * tree_arity + 1
second_child = node_id * tree_arity + 2 ...

lalo:
Thanks Marek @mwydmuch for this contribution and for iterating over all the nits and feedback.

mwydmuch:
Thank you for accepting it and for all the comments @jackgerrits, @rajan-chari, @lalo, I'm very excited that I was able to contribute to VW!

PR #2455, opened 2020-05-27
Separate IO

cassmarcussen:
Task: "Separate I/O threads from parse threads" for Parallelized parsing project - [RLOS](https://vowpalwabbit.org/rlos_projects.html)

Changes made:
- Instead of calling readto directly in read_features (parse_example.cc), the input processing is now separated into its own thread, which is created in parse_dispatch_loop.h. This thread adds the input lines to an input queue. In parallel, once a line of input is available, read_features pops a line off of the input queue.
- Created the file io_to_queue.h, which contains functions to add the input lines to the input queue, and to pop input from the input queue.
- Each input line is encapsulated in a class called IO_Item (io_item.h), which contains the line (message) and numCharsInit (used in read_features).
- parser.h now contains a struct called IO_State, which contains the queue of input lines as well as an atomic control variable, done_with_io, which indicates that all of the input has been read.


cassmarcussen:
_Note:_ This PR is intended for the RLOS Parallelized parsing project. A few shortcuts were made that need to be addressed before merging to master.

Namely, the implementation of IO_State may need revision before merging to master. Currently, IO_State is a member variable of the parser object. However, so that the io thread owns and encapsulates all of the details of the input source, it is worthwhile to consider making IO_State inherit from a pure virtual interface, and making a thread object which owns the IO_State.

vowpalwabbit/io_to_queue.h
```diff
@@ -0,0 +1,76 @@
...
+
+#include <thread>
+#include <future>
+#include <mutex>  
+#include <chrono>   
+
+#include "io_item.h"
+
+//Adds a line of input to the input queue
+inline bool add_to_queue(vw& all, char *& line){
```
jackgerrits, line 12: nit: bracket should be on the next line. You can use `clang-format` and the `.clang-format` file in the root of the repo to quickly resolve these little formatting things.

vowpalwabbit/io_to_queue.h
```diff
@@ -0,0 +1,76 @@
...
+
+  bool finish = false;
+
+  size_t num_chars_initial = readto(*(all.p->input), line, '\n');
+  
+  if(num_chars_initial < 1 || strlen(line) < 1){
+      finish = true;
+  }
+
+  IO_Item line_item(std::string(line), num_chars_initial);
```
jackgerrits, line 22: Rather than constructing this and then copying it on line 26, you can either do a move or better yet just emplace this item directly on line 26.

vowpalwabbit/io_to_queue.h
```diff
@@ -0,0 +1,76 @@
...
+
+  bool should_finish = false;
+
+  while(!should_finish)
+  {    
+    should_finish = add_to_queue(all, line);
+
+  }
+
+  all.p = original_p;
```
jackgerrits, line 48: Does all.p get changed?

vowpalwabbit/parse_dispatch_loop.h
```diff
@@ -5,19 +5,38 @@
...
 
 using dispatch_fptr = std::function<void(vw&, const v_array<example*>&)>;
+struct IO_State;
 
 inline void parse_dispatch(vw& all, dispatch_fptr dispatch)
 {
   v_array<example*> examples = v_init<example*>();
   size_t example_number = 0;  // for variable-size batch learning algorithms
 
+  std::queue<IO_Item> *io_lines = new std::queue<IO_Item>;
```
jackgerrits, line 21: Can this be made a `std::unique_ptr`?

vowpalwabbit/parse_dispatch_loop.h
```diff
@@ -5,19 +5,38 @@
...
+  IO_State io_state(io_lines);
+
+  all.p->_io_state = io_state;
```
jackgerrits, line 24: Is there a way we can make this passing of state more explicit? Right now it is stashed in the parser object and then moved over to the other thread by passing a reference to the all object. Generally if we can constrain what the other thread has access to we can avoid misuse and bugs.

vowpalwabbit/parse_example.cc
```diff
@@ -4,17 +4,45 @@
...
 size_t read_features(vw* all, char*& line, size_t& num_chars)
 {
-  line = nullptr;
-  size_t num_chars_initial = readto(*(all->p->input), line, '\n');
+ 
+  IO_Item result;
+
+  while(true){
+
+    if((*all).p->_io_state.done_with_io && (*all).p->_io_state.io_lines->size() == 0){
```
jackgerrits, line 27:
```suggestion
    if(all->p->_io_state.done_with_io && all->.p->_io_state.io_lines->size() == 0){
```

vowpalwabbit/parse_example.cc
```diff
@@ -4,17 +4,45 @@
...
+class IO_Item;
 
 size_t read_features(vw* all, char*& line, size_t& num_chars)
 {
-  line = nullptr;
-  size_t num_chars_initial = readto(*(all->p->input), line, '\n');
+ 
+  IO_Item result;
+
+  while(true){
```
jackgerrits, line 25: Can you explain why this needs to be in a loop?

vowpalwabbit/parse_example.cc
```diff
@@ -4,17 +4,45 @@
...
+
+    result = pop_io_queue(all);
+
+    if(result.numCharsInit > 0){
+      break;
+    }
+
+  }
+
+  std::string result_string =  result.message;
```
jackgerrits, line 39: To remove a copy you can change this to:
```suggestion
  const auto& result_string =  result.message;
```

vowpalwabbit/parser.h
```diff
@@ -35,6 +37,46 @@ struct example_initializer
...
+      inline void set_done_io(bool done_io){
+        done_with_io.store(done_io);
+      }
+
+      inline bool get_done_io(){
+        return done_with_io;
+      }
+
+
+      ~IO_State() {}
```
jackgerrits, line 76: Currently this is leaking the `io_lines` object. Using a `std::unique_ptr` is a quick and easy way to resolve this.

vowpalwabbit/parser.h
```diff
@@ -35,6 +37,46 @@ struct example_initializer
...
+         done_with_io.store(false);
+      }
+
+      IO_State(std::queue<IO_Item> *new_input_lines){
+          //input_lines_copy now points to new_input_lines
+          io_lines = new_input_lines;
+          done_with_io.store(false);
+      }
+
+      IO_State operator=(const IO_State &toCopy){
```
jackgerrits, line 56: Needs to return a reference and do a self reference check

jackgerrits, line 56: Given that you are implementing the copy constructor you should either implement or delete the move constructors.

jackgerrits:
Do you have any preliminary measurements about what this does for perf?

vowpalwabbit/parser.h
```diff
@@ -35,6 +37,46 @@ struct example_initializer
   example* operator()(example* ex);
 };
 
+struct IO_State {
+
+      std::queue<IO_Item> *io_lines = nullptr;
+      std::atomic<bool> done_with_io;
+
+      IO_State(){
```
peterychang, line 45: One of these two constructors should be removed. Otherwise you'll end up with 2 different situations to clean up
1. we call `new` on io_lines. IO_State now owns the memory and needs to call `delete`.
2. the io_lines pointer is passed in. IO_State does not own the memory, and should not call `delete`.

This will cause either a memory leak or a double free exception.

If IO_State is being shared between the threads (which looks to be the case) then the first scenario is the one we want to use and the queue should never be passed in as a parameter or even be accessible outside of the class.

vowpalwabbit/parser.h
```diff
@@ -35,6 +37,46 @@ struct example_initializer
   example* operator()(example* ex);
 };
 
+struct IO_State {
```
peterychang, line 40: IO_State isn't encapsulating an entire concept here. Ideally it wrap all the functionality of a queue behind a thread-safe interface, along with any state information you need.

You should add
- A mutex member variable
- `push()` and `pop()` member functions

vowpalwabbit/io_to_queue.h
peterychang, line 12: Most of these should be moved into IO_State (more details in the comments there)

vowpalwabbit/io_item.h
```diff
@@ -0,0 +1,51 @@
+#ifndef _IO_ITEM_
+#define _IO_ITEM_
+class IO_Item {
+
+  public:
+
+      std::string message;
+      int numCharsInit;
```
peterychang, line 8: I think numCharsInit is unnecessary now. IIRC during our call we figured out that you could detect this while reading the file and just set done to true instead of saving it.

vowpalwabbit/io_item.h
```diff
@@ -0,0 +1,51 @@
...
+          numCharsInit = 0;
+      }
+
+      IO_Item(std::string myMsg, int myNumCharsInit){
+
+          message.assign(myMsg);
+          numCharsInit = myNumCharsInit;
+      }
+
+      IO_Item operator=(const IO_Item &toCopy){
```
peterychang, line 20: You don't need to specify the copy constructor or assignment operator since the default behavior will do what you want.

vowpalwabbit/io_item.h
```diff
@@ -0,0 +1,51 @@
...
+      }
+
+      IO_Item(const IO_Item &toCopy){
+          message.assign(toCopy.message);
+          numCharsInit = toCopy.numCharsInit;
+      }
+
+      ~IO_Item() {}
+
+      inline std::string getString(){
```
peterychang, line 33:
```suggestion
      inline const std::string& getString(){
```
avoid copying if you don't need to

vowpalwabbit/io_to_queue.h
```diff
@@ -0,0 +1,76 @@
...
+#include "io_item.h"
+
+//Adds a line of input to the input queue
+inline bool add_to_queue(vw& all, char *& line){
+
+  bool finish = false;
+
+  size_t num_chars_initial = readto(*(all.p->input), line, '\n');
+  
+  if(num_chars_initial < 1 || strlen(line) < 1){
```
peterychang, line 18: readto will strip characters from whatever was read in the file. I think that can cause a situation where `strlen(line)` == 0 but we haven't finished reading the file yet

vowpalwabbit/parse_example.cc
cassmarcussen: Since the I/O thread and parser thread are running in parallel, there are moments in the execution of the program when the io queue stored in IO_State is empty, but read_features and the parser should still be expecting lines of code to parse. For example, this occurs occasionally when the I/O thread and parsing of lines first begins, when read_features must wait until an input line is added to the io queue before it can move on to parse and process it. The while loop enforces this waiting when a line is still expected from the io queue. 

Is it ok to keep this while loop, or is there another design I should be considering instead?

peterychang: This should be done via condition variable instead of a busy loop

vowpalwabbit/parse_example.cc
```diff
@@ -4,17 +4,45 @@
...
+
+  line = new char[result_string.size() + 1];
```
peterychang, line 41: This is not ideal since we'd have to deal with the memory cleanup in multiple places (text parser + json parser + maybe cache parser). Ideally we'd convert line to a string instead of a char*, but the json parser requires a mutable c-string as its input.

I'm checking what our options are, at worst we may be able to make a small (but really awful) change as a **_temporary_** band-aid

peterychang: I spoke with the guy who wrote the json parser and he was working on a change that makes this a lot nicer. He's trying to get it into a place where its usable, once thats done I'll pass it over to you so we can start building on top of it.

Essentially we'll be changing `char*& line` to `std::string& line` where the caller of this function will pass in a reference to the string to have it filled. Keep this the way it is for now since we'll have to refactor it anyways once the new interface is determined

cassmarcussen:
In new commits, applyied CR changes, including:
    * io_item and io_state constructors
    * Moving push/pop io queue into io_state
    * In read_features, CV instead of busy loop
    * In read_features, io_item - char * line converted to vector<char> (for compatibility with cache reader/features)
    * Formatting, style guidelines, adding copyright headers

vowpalwabbit/parser.h
olgavrou: Could the use of `ptr_queue` from `queue.h` be used in `io_state` instead of having new queue/mutex/cv implemented, and mirror the `ready_parsed_examples` queue usage? `ptr_queue` only takes pointers to objects so `io_state::add_to_queue` could create an `io_item` object on heap and `io_state::pop_io_queue` could return a `unique_ptr` to that object, transferring the ownership.

peterychang: Since this PR is intended for a summer project, there are several reasons I chose not to suggest using the queue implementation we have in `queue.h`. Some of these are educational as the queue is an easy way to get your feet wet when learning about threading. I wouldn't recommend using the `ptr_queue` in either case because there workflows where early termination would cause additional complexity to the cleanup.

Both the author and I understand this PR cannot be merged as-is. My hope is that we'll have time to revisit this change once the main focus of the project is completed. Or if we time is too short for that, I hope to continue working with her on the change afterward to get it into a merge-ready state

JohnLangford:
I understand this was opened by mistake, so closing for now (reopen if I'm wrong).

PR #2457, opened 2020-05-27
Drain the queue if it is non empty due to possible early termination exits

### LINKED ISSUE #2456: Flaky debug assert hit in BFGS
A debug assert that was recently added is hitting in the BFGS test with low probability (~1/35) due to the early termination.
```
RunTests: test 16: '/usr/bin/timeout 80 ../build/vowpalwabbit/vw -k -c -d train-sets/rcv1_small.dat --loss_function=logistic --bfgs --mem 7 --passes 20 --termination 0.001 --l2 1.0 --holdout_off' failed (exitcode=134)
using l2 regularization = 1
enabling BFGS based optimization **without** curvature calculation
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
decay_learning_rate = 1
m = 7
Allocated 18M for weights and mem
## avg. loss     der. mag.     d. m. cond.     wolfe1        wolfe2        mix fraction    curvature     dir. magnitude    step size
creating cache_file = train-sets/rcv1_small.dat.cache
Reading datafile = train-sets/rcv1_small.dat
num sources = 1
 1 0.69315       0.00266       0.87764                                                 2.24708       776.93237     0.39057
 3 0.51357       0.00493       4.93046        0.523903      0.088793                                  76.25748      1.00000
 4 0.65936       0.04915       49.15202       -0.910622     -2.480116                                 (revise x 0.5)    0.50000
 5 0.51658       0.00876       8.76105        -0.037665     -0.999616                                 (revise x 0.5)    0.25000
 6 0.49499       0.00028       0.28254        0.463963      -0.056952                                 0.51262       1.00000
 7 0.49354       0.00006       0.05641        0.619867      0.244153                                  0.08545       1.00000
 8 0.49287       0.00005       0.05434        0.870687      0.741762                                  0.91640       1.00000
 9 0.48978       0.00014       0.13750        0.772761      0.546930                                  2.01229       1.00000
10 0.48472       0.00027       0.27437        0.750340      0.501776                                  3.21400       1.00000
11 0.47920       0.00017       0.16868        0.671044      0.340515                                  1.40136       1.00000
12 0.47707       0.00001       0.00760        0.593375      0.181238                                  0.09201       1.00000
13 0.47691       0.00000       0.00168        0.593278      0.185019                                  0.00955       1.00000
finished run
number of examples per pass = 1000
passes used = 13
weighted example sum = 13000.000000
weighted label sum = -1066.000000
average loss = 0.441700
best constant = -0.164369
best constant's loss = 0.689781
total feature number = 1023607
vw: /home/jack/w/repos/vowpal_wabbit/vowpalwabbit/parser.cc:1034: void free_parser(vw&): Assertion `all.p->ready_parsed_examples.size() == 0' failed.
/usr/bin/timeout: the monitored command dumped core
RunTests: exiting with status=134
```

jackgerrits:
fixes #2456

### DIFF
--- a/vowpalwabbit/parser.cc
+++ b/vowpalwabbit/parser.cc
@@ -229,7 +229,7 @@ void make_write_cache(vw& all, std::string& newname, bool quiet)
     all.trace_message << "can't create cache file !" << temp << endl;
     return;
   }
-  
+
   size_t v_length = (uint64_t)VW::version.to_string().length() + 1;
 
   output->bin_write_fixed(reinterpret_cast<const char*>(&v_length), sizeof(v_length));
@@ -1017,6 +1017,18 @@ void start_parser(vw& all) { all.parse_thread = std::thread(main_parse_loop, &al
 
 void free_parser(vw& all)
 {
+  // It is possible to exit early when the queue is not yet empty.
+
+  while(all.p->ready_parsed_examples.size() > 0)
+  {
+    auto* current  = all.p->ready_parsed_examples.pop();
+    // this function also handles examples that were not from the pool.
+    VW::finish_example(all, *current);
+  }
+
+  // There should be no examples in flight at this point.
+  assert(all.p->ready_parsed_examples.size() == 0);
+
   std::vector<example*> drain_pool;
   drain_pool.reserve(all.p->example_pool.size());
   while (!all.p->example_pool.empty())
@@ -1030,8 +1042,6 @@ void free_parser(vw& all)
     all.p->example_pool.return_object(example_ptr);
   }
 
-  // There should be no examples in flight at this point.
-  assert(all.p->ready_parsed_examples.size() == 0);
 }
 
 namespace VW


peterychang:
I think draining the queue in free_parser can hide bugs.

Its more complex, but I think the correct thing to do would be to drain the queue/start rejecting new entries when done is set, and assert that the queue is empty when we call free_parser

jackgerrits:
The problem with that is the `done` state also affects the `pop` behavior of the queue. Right now, if you are waiting for something to be pushed to the queue and the `done` state is entered you will get back nullptr.

If we extract the concept of `done` out, then how ill we deal with this case where a caller is waiting on something to be added that never will be?

peterychang:
The callers of the queue should be handling the concept of `done`. The queue implementation should just be a thread-safe queue with a few common utilities, the concept of done-ness is something that belongs entirely to VW. In this case we have 1 place inserting and 1 place removing elements so it would be pretty straightforward to implement.

peterychang:
As a quick fix, this looks ok to me. Maybe add an issue to do some minor refactoring to pull `done` out of the queue though?

jackgerrits:
Discussed offline and the real fix to this race condition is non-trivial and unclear at this time. This mitigation will be merged and an issue has been created (#2462) to track the race condition.

PR #2460, opened 2020-06-01
Revert API break

### LINKED ISSUE #2459: Positional args refactor change introduced API break
#2418 changed the processing of positional arguments so that if more than 1 were passed it changed from being ignored to an error. This is a breaking change and should not being introduced without a deprecation and major release.

In this case, it was done accidentally and was not intended to be a breaking change.

jackgerrits:
Fixes #2459


rajan-chari:
Looks good.  Glad to have the unit test.

PR #2461, opened 2020-06-01
Fix build break for older versions of ZLib

jackgerrits:


JohnLangford:
Looks good, merge at will.

PR #2463, opened 2020-06-02
Simplify clang-format and small fixes and changes

jackgerrits:
The previous clang-format file was very large and hard to reason about. I've updated it to base on the Google style guide (which we pretty much follow already).

- Simplify format
- Base it on Google style
- Fix: pointer and reference alignment
    - `thing *a;` -> `thing* a;`
- Change: short loops and functions on single line from allowed to disallowed

Running this new file results in **very few changes**. The few changes that do occur are either:
- An existing formatting inconsistency
- Fixing the pointer alignment
- Changing single line loops or functions to be over several lines
    - This is a personal preference thing but as it is inconsistent in the code base I figured I would set a guide in the format file. Preferably all blocks would use braces (yet again, opinion)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,10 +2,10 @@ BasedOnStyle: Google
 AccessModifierOffset: -2
 AlignAfterOpenBracket: DontAlign
 AlignOperands: false
-AllowShortBlocksOnASingleLine: false
+AllowShortBlocksOnASingleLine: Always
 AllowShortCaseLabelsOnASingleLine: false
-AllowShortFunctionsOnASingleLine: true
-AllowShortIfStatementsOnASingleLine: true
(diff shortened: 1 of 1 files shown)


.clang-format
```diff
@@ -1,89 +1,13 @@
...
-  AfterFunction:   false
-  AfterNamespace:  false
-  AfterObjCDeclaration: false
-  AfterStruct:     false
-  AfterUnion:      false
-  BeforeCatch:     false
-  BeforeElse:      false
-  IndentBraces:    false
-BreakBeforeBinaryOperators: None
+AllowShortLoopsOnASingleLine: false
```
jackgerrits, line 9:
```suggestion
AllowShortLoopsOnASingleLine: true
```

Issue #2458, opened 2020-06-01
Improve DFtoVW user experience


etiennekintzler:
I have been thinking (and playing) with the `DFtoVW` and I think **the class API could be improved**. Except for the convenience function `DFtoVW.from_colnames` which is quite clear (maybe except its name) the default `DFtoVW` constructor could be further simplified:

1. The user shouldn't have to use `Col` (although this class could be kept for internal use).

2. A parameter `features`, that build a naked namespace (no name nor scaling factor) with supplied `Feature` object(s), could be introduced.

I detail these two points below. If you think these two points are valid, I can of course contribute on this issue. Suggestions are welcomed !

-----

## 1. User shouldn't have to use `Col`

The main problem with `Col` is that it's tedious write and not really legible. For instance if we want to create the resulting line " | feat_x:x feat_y:y", (with dataframe df with two colmn a and b), the user has to write: 

```python
DFtoVW(namespaces=Namespace([Feature(Col("a"), "feat_a"), Feature(Col("b"), "feat_b")]))
```

I might be wrong here but **for most of the use case, the *value* of the feature is a column and the *name* of the feature is a constant**. If it is the case, then default behavior of `Feature` should reflect that. 

Instead of using `Col` we could just add two boolean parameters that specify whether the argument passed is a column or not. In doing so, we can: 
- handle the edge cases (name from a column and/or constant feature value) 
- keep the API simple for most of the usages.

The new signature of Feature would then be : 
```python
Feature(value, name=None, value_from_df=True, name_from_df=False)
```
Hence the current formulation:
```python
Feature(Col("a"), "feat_a")
Feature("a", "feat_a")
Feature("a", Col("b"))
```
would become:
```python
Feature("a", "feat_a")
Feature("a", "feat_a", value_from_df=False)
Feature("a", "b", name_from_df=True)
```

## 2. Parameter `features` in the constructor

To my knowledge, Namespace is a term highly specific to Vowpal Wabbit input format. It is not found elsewhere in the other python ML packages that the user might use (scikit-learn, statsmodels, keras, pytorch or tensorflow). Since multiple namespaces are for advanced usage, I suggest to create a parameter `features` (along with `namespaces`) that build a naked namespace.

In the current API formulation, if the user want to create the " | x" (for prediction purpose) he has to write (for column "a" of dataframe "df"):
```python
DFtoVW(namespaces=Namespace(Feature(Col("a"))))
```
[...]

Issue #2452, opened 2020-05-27
Support all label types in DFToVW converter


jackgerrits:
Currently only SimpleLabel is supported, it should eventually support all label types.

- [ ] multiclass
- [ ] mutlilabels
- [ ] cost sensitive (multiline)
- [ ] contextual bandit (multiline)
- [ ] ccb (multiline)
- [ ] slates (multiline)
- [ ] cb_eval (multiline)

etiennekintzler:
Is the **multiclass** what's explained in [One Against All (oaa) multi class example](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/One-Against-All-(oaa)-multi-class-example)? If yes, I think the multiclass label type is quite straightforward to implement. The main difference with the `SimpleLabel` approach is that it required the label to be in the natural number set (between 1 and K). In fact `SimpleLabel` work just fine if the user convert the target column to natural number beforehand (but introducing `Multiclass` type might allow type checking and/or automatic conversion to natural numbers).

However I saw in the very wiki page that weights could be added to the target as well : 
> You may add weights to the example and the features, use name-spaces, etc.

Are weights the same thing as `Importance` in the [Input format wiki](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Input-format) ? Since you said that the latter was under change should it be implemented ? Is the weight set by adding a space along with the desired weight (as in `1 0.3`) ?

Regarding **multilabels** is it what's explained in the [Cost Sensitive One Against All multi class example](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Cost-Sensitive-One-Against-All-%28csoaa%29-multi-class-example) ? If yes the main difficulty is that all the multiple labels can be present but not necessary, for instance : 
> 1:1.0 a1_expect_1| a
2:1.0 b1_expect_2| b
3:1.0 c1_expect_3| c
1:2.0 2:1.0 ab1_expect_2| a b
2:1.0 3:3.0 bc1_expect_2| b c
1:3.0 3:1.0 ac1_expect_3| a c
2:3.0 d1_expect_2| d

In your opinion, how would the input dataframe be like ?

jackgerrits:
Yes, OAA uses multiclass labels. Weight is a part of this label as well as the class, it's effectively the same thing as importantance.
> Is the weight set by adding a space along with the desired weight (as in 1 0.3) ?

Yes

[...]

PR #2471, opened 2020-06-14
simplify api, introduce AttributeDescriptor for type and value checki…

etiennekintzler:
This PR goal is to simplify the API of `DFtoVW` (issue #2458) 

The simplifications made are the following:
- The user does not have to use the `Col` type anymore, although the class is still used internally (under the name `_Col`). Whether the argument passed refers to a dataframe column or is aconstant depends on an `{attribute_name}_from_df` attribute. The latter attribute has default value for the most common use case.
- A `features` parameter is added to the `DFtoVW` constructor. This parameter create a naked `Namespace` (namespace without any name or scaling factor). Of course, the user can still pass `Namespace` object(s) to the parameter `namespaces` for more advanced usage (but he can't use both `namespaces` and `features` in the same time).
- changed the main method name from `DFtoVW.process_df` to `DFtoVW.convert_df` which is more explicit in my opinion.

Now the `DFtoVW` can be used as follow:
```python
df = pd.DataFrame({"y": [1], "a": [2], "b": [3]})
conv = DFtoVW(df=df, label=SimpleLabel("y"), features=[Feature(col) for col in ["a", "b"]])
conv.convert_df()
```

#### Using descriptor class
Along with the API simplifications I rewrite the internals so type and/or value checking can be performed directly on the attribute (or on the underlying column in the dataframe if the attribute refers to a column). 

This type and value checking is performed using a **descriptor class** named `AttributeDesciptor`. The class will _manage_ the setting of the attributes of the following classes: SimpleLabel, Feature, and Tag. When an attribute refers to a column (i.e when `{attribute_name}_from_df=True`) the expected type and value are passed to the `_Col` instance and checking is done when initializing `DFtoVW` (at this time the dataframe and thus the column is available).

#### Minor change
Some minor changes:
- the `Namespace` does not rely on `AttributeDescriptor` since none of its attributes is either a value or a column.
- descriptor and the managed class inherit from `object` to comply with python 2.7 (see [Descriptor HowTo Guide](https://docs.python.org/2/howto/descriptor.html#id4))
- I create a `_Tag` type, but the parameter `tag_from_df` is available in `DFtoVW` and not in `_Tag` constructor since `_Tag` won't be used directly by the user.
- use black for style formatting with 79 line length.

### DIFF
--- a/python/tests/test_pyvw.py
+++ b/python/tests/test_pyvw.py
@@ -2,7 +2,13 @@ import os
 
 from vowpalwabbit import pyvw
 from vowpalwabbit.pyvw import vw
-from vowpalwabbit.pyvw import DFtoVW, SimpleLabel, Feature, Namespace, Col
+from vowpalwabbit.pyvw import (
+    DFtoVW,
+    SimpleLabel,
+    MultiClass,
+    Feature,
+    Namespace,
+)
 import pytest
 import pandas as pd
 
@@ -17,7 +23,7 @@ class TestVW:
         assert isinstance(self.model, vw)
 
     def test_learn_predict(self):
-        ex = self.model.example('1 | a b c')
+        ex = self.model.example("1 | a b c")
         init = self.model.predict(ex)
         assert init == 0
         self.model.learn(ex)
@@ -28,11 +34,11 @@ class TestVW:
 
     def test_get_tag(self):
         ex = self.model.example("1 foo| a b c")
-        assert ex.get_tag() == 'foo'
+        assert ex.get_tag() == "foo"
         ex = self.model.example("1 1.0 bar| a b c")
-        assert ex.get_tag() == 'bar'
+        assert ex.get_tag() == "bar"
         ex = self.model.example("1 'baz | a b c")
-        assert ex.get_tag() == 'baz'
+        assert ex.get_tag() == "baz"
 
     def test_num_weights(self):
         assert self.model.num_weights() == 2 ** BIT_SIZE
@@ -48,28 +54,29 @@ class TestVW:
 
 def test_delete():
     model = vw(quiet=True, b=BIT_SIZE)
-    assert 'model' in locals()
+    assert "model" in locals()
     del model
-    assert 'model' not in locals()
+    assert "model" not in locals()
 
 
 # Test prediction types
 
+
 def test_scalar_prediction_type():
     model = vw(quiet=True)
-    model.learn('1 | a b c')
+    model.learn("1 | a b c")
     assert model.get_prediction_type() == model.pSCALAR
-    prediction = model.predict(' | a b c')
+    prediction = model.predict(" | a b c")
     assert isinstance(prediction, float)
     del model
 
 
 def test_scalars_prediction_type():
     n = 3
-    model = vw(loss_function='logistic', oaa=n, probabilities=True, quiet=True)
-    model.learn('1 | a b c')
+    model = vw(loss_function="logistic", oaa=n, probabilities=True, quiet=True)
+    model.learn("1 | a b c")
     assert model.get_prediction_type() == model.pSCALARS
-    prediction = model.predict(' | a b c')
+    prediction = model.predict(" | a b c")
     assert isinstance(prediction, list)
     assert len(prediction) == n
     del model
@@ -77,31 +84,39 @@ def test_scalars_prediction_type():
 
 def test_multiclass_prediction_type():
     n = 3
-    model = vw(loss_function='logistic', oaa=n, quiet=True)
-    model.learn('1 | a b c')
+    model = vw(loss_function="logistic", oaa=n, quiet=True)
+    model.learn("1 | a b c")
     assert model.get_prediction_type() == model.pMULTICLASS
-    prediction = model.predict(' | a b c')
+    prediction = model.predict(" | a b c")
     assert isinstance(prediction, int)
     del model
 
 
 def test_prob_prediction_type():
-    model = vw(loss_function='logistic', csoaa_ldf='mc', probabilities=True, quiet=True)
-    multi_ex = [model.example('1:0.2 | a b c'), model.example('2:0.8  | a b c')]
+    model = vw(
+        loss_function="logistic",
+        csoaa_ldf="mc",
+        probabilities=True,
+        quiet=True,
+    )
+    multi_ex = [
+        model.example("1:0.2 | a b c"),
+        model.example("2:0.8  | a b c"),
+    ]
     model.learn(multi_ex)
     assert model.get_prediction_type() == model.pPROB
-    multi_ex = [model.example('1 | a b c'), model.example('2 | a b c')]
+    multi_ex = [model.example("1 | a b c"), model.example("2 | a b c")]
     prediction = model.predict(multi_ex)
     assert isinstance(prediction, float)
     del model
 
 
 def test_action_scores_prediction_type():
-    model = vw(loss_function='logistic', csoaa_ldf='m', quiet=True)
-    multi_ex = [model.example('1:1 | a b c'), model.example('2:-1  | a b c')]
+    model = vw(loss_function="logistic", csoaa_ldf="m", quiet=True)
+    multi_ex = [model.example("1:1 | a b c"), model.example("2:-1  | a b c")]
     model.learn(multi_ex)
     assert model.get_prediction_type() == model.pMULTICLASS
-    multi_ex = [model.example('1 | a b c'), model.example('2 | a b c')]
+    multi_ex = [model.example("1 | a b c"), model.example("2 | a b c")]
     prediction = model.predict(multi_ex)
     assert isinstance(prediction, int)
     del model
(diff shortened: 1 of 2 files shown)


jackgerrits:
Hi @etiennekintzler,

Thanks for your continued work on this! 

>     simplify the API of DFtoVW (issue #2458)
>     add a multi class label (part of issue #2452).

Would you be able to split this PR into these two things? Generally a PR should address only a single thing if possible. Smaller PRs will make it much easier to review and generally help the turnaround time.

etiennekintzler:
Hello @jackgerrits ! 

Ok sure, if you prefer I can split this into 2 different PR. 

The multiclass label part is quite short and is based on the descriptor pattern which the simplification introduces. So I might do a 2nd PR for the multiclass __after__ the 1st one (simplification of the API) is accepted.

Should I close this PR and create a new one ? Otherwise I can make a new commit to remove the part related to multiclass label but the name of the branch will still be wrong (simplify_api_and_add_multiclass).

jackgerrits:
Thanks! Updating this PR is perfectly fine I think.

python/tests/test_pyvw.py
```diff
@@ -347,94 +381,158 @@ def check_error_raises(type, argument):
...
         argument()
 
+
 def test_from_colnames_constructor():
     df = pd.DataFrame({"y": [1], "x": [2]})
     conv = DFtoVW.from_colnames(y="y", x=["x"], df=df)
-    lines_list = conv.process_df()
+    lines_list = conv.convert_df()
     first_line = lines_list[0]
     assert first_line == "1 | 2"
```
lalo, line 390: I just noticed that there isn't a test that verifies past one line, can you modify one of the tests and an assert on the second_line just to have more coverage for future PRs.

python/vowpalwabbit/pyvw.py
```diff
@@ -1398,386 +1432,413 @@ def get_label(self, label_class=simple_label):
...
                 )
             )
 
 
-class SimpleLabel:
-    """The SimpleLabel class is used to build a simple label for the
-    constructor of DFtoVW.
-    """
+class SimpleLabel(object):
+    """The simple label type for the constructor of DFtoVW."""
```
lalo, line 1700: Not related to this PR, but this naming convention might be confusing since this is only helpful for DFtoVW. What do you think @jackgerrits

python/tests/test_pyvw.py
lalo: also, would there be any value on having a test of a big DF being converted? Maybe not in this file, but for CI purposes

python/tests/test_pyvw.py
```diff
@@ -347,94 +381,158 @@ def check_error_raises(type, argument):
...
+    conv = DFtoVW(df=df, features=[Feature(col) for col in ["a", "b"]])
+    first_line = conv.convert_df()[0]
+    assert first_line == "| 2 3"
+
+
+def test_multiclass():
+    df = pd.DataFrame({"a": [1], "b": [0.5], "c": ["x"]})
     conv = DFtoVW(
-        df=df, namespaces=Namespace([Feature(Col("a")), Feature(Col("b"))])
+        df=df, label=MultiClass(name="a", weight="b"), features=Feature("c")
```
lalo, line 459: had a hard time finding this test. This makes me realize this single test file might be too much. Not related with this PR.

python/tests/test_pyvw.py
etiennekintzler: @lalo 

Regarding your **1st** comment : a test with multiple lines is performed line 428 of the _new_ file (`test_multiple_lines`).

Regarding your **2nd** comment : of course, a test to see how it scales could be included.

I personally tested the class on [avazu click through rate prediction dataset from kaggle](https://www.kaggle.com/c/avazu-ctr-prediction/data). The class takes about 10 seconds to convert 1 millions rows with about 10 columns. The **main bottleneck is the conversion of float/int columns to string columns** (almost 1 sec for 1 million rows). You can see the related question I post on [StackOverflow](https://stackoverflow.com/questions/61445646/concatenate-the-values-of-two-pandas-series-that-are-not-of-str-type-is-too-sl). 

I can add a test using random data with 100k, 1million or more rows.

python/vowpalwabbit/pyvw.py
etiennekintzler: Yep, I saw too that in pyvw there are classes like `simple_label`. In my opinion the following 2 things could be done : 
- prepending all the classes used in `DFtoVW` with `DF` 
- since `simple_label` and friends does not respect [python's CapsWords convention for class names](https://www.python.org/dev/peps/pep-0008/#class-names) it could be renamed with the appropriate CapsWords. Also, it won't clash with the classes used in DFtoVW since we preppend them with "DF"

python/tests/test_pyvw.py
lalo: There's 2 things at play here
1) we shouldn't be converting into the string to send it off to vw but this is just room for improvement on the python api side, not related to vw. So in the long term, there might be performance gains.
2) I wonder if we can improve performance by switching things up in the functions that create the string. For example if we use df.iterrows to iterate over each row and create the line on one go, instead of accessing each value from different places. Not sure if there would be a benefit, though. Another way would be by converting the df to a numpy.array and operate directly from those values. This might come into play if we had what i mentioned on the previous point available.

etiennekintzler: 1. If not string, what should format sent ?

2. Iterate through the rows takes much more times, the main benefit of using columns operation is we take advantage of vectorization. Also converting to `numpy.array` won't result in huge gains since it's what pandas already use in the backend. Indeed you can try the following in ipython console : 

```python
import pandas as pd
import numpy as np

x = np.random.normal(size=int(1e6))
s = pd.Series(x)

%timeit s.astype(str) # 1.42 s ± 79.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit  s.to_numpy().astype(str) # 1.27 s ± 83.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

lalo: 1) Binary data and try to avoid converting to strings just to pass it down elsewhere
2) Good point on the column. I was going for minimizing string concat operations and building the string in one go. Anyways, its best to measure before going on possible dead ends.

Could you share the python script to load the avazu dataset? Or maybe add it as a python example (there is a dir for these scripts) if you want to

etiennekintzler: 1. Okay I see
2. Yes when I did the original PR 1 month ago I tried various methods too (line per line, converting to numpy and doing ops in numpy, doing operation directly on pandas columns).

Of course I can share the script to load the avazu dataset ! You examples foldder is the following https://github.com/VowpalWabbit/vowpal_wabbit/tree/master/python/examples ? 

Please note that the dataset is huge (the training part is about 44 millions lines and 6 Gib uncompressed), so not sure if this is to be uploaded :smile: I can do sampling too. As the dataset does not fit in memory I used pandas generator to load it in batch of 1 million rows and learn on each batch. I think it can be a good showcase on how to use the DFtoVW API and to learn on data that wouldn't fit in memory (within Python session).

lalo: Yes, that's the folder. I think we could just reference the dataset on a python comment on the script. It would still be helpful to see an actual example of using DFtoVW. Later, if someone wants to run it they have to download the dataset themselves. Feel free to open another PR.

python/vowpalwabbit/pyvw.py
```diff
@@ -1857,243 +1929,410 @@ def process(self, df=None):
...
+
     The class also provided a convenience constructor to initialize the class
     based on the target/features column names only.
     """
 
-    def __init__(self, df, namespaces, label=None, tag=None):
+    def __init__(
+        self,
+        df,
+        features=None,
```
lalo, line 1942: what happens when we set both the features and the namespaces, that would always imply an error?

lalo: or is there a valid scenario when we would use the init and set both features and namespaces when calling into this function?

etiennekintzler: Error will always be raised if both `features` and `namespaces` are set. 
@lalo 

I see two distinct cases: 
- `features` is used to create naked namespace, 
- `namespaces` is used to create a single named/scaled namespace or multiples named/scaled namespaces.

Let's imagine an user that want to build a naked namespace with `features` and 2 named namespaces with `namespaces`. This operation could be performed using just the `namespaces` parameter, by passing a list of 3 namespaces to it (the first element of the list being the unnamed namespace for example).

 I see `features` as shortcut/convenience parameter for new/rushed users, and `namespaces` for more advanced usages. But allowing both parameters to be passed might be a bit confusing (especially regarding the location where the naked namespace should be). What do you think ?

etiennekintzler:
> Hi @etiennekintzler,
> 
> Thanks for your continued work on this!
> 
> > ```
> > simplify the API of DFtoVW (issue #2458)
> > add a multi class label (part of issue #2452).
> > ```
> 
> Would you be able to split this PR into these two things? Generally a PR should address only a single thing if possible. Smaller PRs will make it much easier to review and generally help the turnaround time.

Done since commit 4a37de8 for code. I have updated the PR description as well.

python/vowpalwabbit/pyvw.py
lalo: Yes, your last point was what I was going with but at least we error out right now if we supply both. I feel it is just a matter of coming up with better naming and get feedback from more people. It just might be misleading to have both parameters. For now, maybe just add on the comment of this function that it only takes one or the other. Later we can shift things around more.

lalo:
@etiennekintzler Thanks for submitting this PR! I think the previous points discussed are related more to future improvements or things to consider on how to evolve the python side of the world instead of things to be addressed in this PR.

lalo:
Also @etiennekintzler , let us know if you have found the api in general funky to work with or pain point we can address. This is in general to vw (both vw and pyvw), not only to this DFtoVW which you've been pushing for. Just tag me on that post, to be aware of it.

etiennekintzler:
Thanks for asking for my feedbacks @lalo 

I will mostly talk about the `vw.py` since what I've used so far. Also I am fairly new to the library so I might encounter new issue/ideas of improvements over time.

### In pyvw

I'd say the main issue is that **a lot of Python errors refers to C++ errors**. For example : 
```python
ArgumentError: Python argument types in
    vw.audit_example(vw, str)
did not match C++ signature:
    audit_example(class boost::shared_ptr<struct vw>, class boost::shared_ptr<struct example>)
```

or

```python
TypeError: No registered converter was able to produce a C++ rvalue of type class boost::shared_ptr<struct example> from this Python object of type str
```

The following points are more improvements than issues:
- I found that the performance of the predictor depends highly on the initial learning rate. The learning rate default value of 0.5 is quite high and a helper function to cross validate its initial value could be helpful.
- It'd be nice if we could use `pyvw.vw.learn` for multiple lines not only line per line. Training is otherwise quite slow in Python compare to CLI.
- more helpers function for diagnostic: see for instance my issue #2474 . Also helper function to get loss and learning rate value could be added. Of course one can activate diagnostic using `--audit` in the console but it'd be nicer to have it directly in the Python session. I 've discovered recently the script [vw-varinfo](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Using-vw-varinfo); this is quite nice and could be the standard output for the model (defining for instance `pyvw.vw.summary` method).


## In general in vw

The work done on tutorials on http://vowpalwabbit.org is great for newcomers. The contextual bandit tutorial is insightful but based on simulated data. I think a tutorial on offline data could be interesting. Also for CB label, it's not clear how to obtain the probability when it is not a direct output of the algorithm that generates the data. 

For inspiration there is a recent R package that deals with contextual bandit and that is quite neat in terms of tutorials about usage of CB on offline data. See for example : [Offline evaluation: Replication of Li et al 2010](https://cran.r-project.org/web/packages/contextual/vignettes/replication.html) or [Demo: Offline cMAB: CarsKit DePaul Movie Dataset](https://cran.r-project.org/web/packages/contextual/vignettes/offline_depaul_movies.html). The other tutorial of the package are quite good; for full list you can check the entry "Vignettes" in https://cran.r-project.org/web/packages/contextual/index.html

PR #2480, opened 2020-06-25
Make count_label thread safe locklessly

peterychang:


vowpalwabbit/best_constant.h
```diff
@@ -6,26 +6,19 @@
...
-      if (sd->second_observed_label != FLT_MAX)
-      {
-        if (sd->second_observed_label != l)
-          sd->is_more_than_two_labels_observed = true;
-      }
-      else
-        sd->second_observed_label = l;
+  float l1 = FLT_MAX;
+  float l2 = FLT_MAX;
+  if (!sd->first_observed_label.compare_exchange_strong(l1, l) && l1 != l) {
```
peterychang, line 17: Explanation of this line:
```
compare_exchange_strong(&expected, value) {
    if(sd->first_observed_label == expected) {
        sd->first_observed_label = value;
        return true;
    }
    expected = sd->first_observed_label
    return false;
}
```
If `first_observed_label == FLT_MAX`, do `first_observed_label = l`
If `first_observed_label != FLT_MAX`, and the value in `first_observed_label != l`, enter the if block

PR #2483, opened 2020-06-25
Proposal "half-step to remov*all": Modeling vw with types

lalo:
**Disclaimer:** at the early stage of this PR, ignore all files edited. The file you want to be reviewing and thinking about is vowpalwabbit/global_data.h. **Second disclaimer:** ignore the comments added with this format //** **//, those will end up being removed but they are currently being used as live notes. You should still read them though, since it represents a bit of the thought process on how to represent everything.

Not ready for merge in the sense that variable names my change until we reach consensus. Also, want to get some support before I start tackling the dependencies (i.e. java or whatever reason linux build fails).

----------------------------------

This is mostly a semantic change, since we're just shifting things around between structs.

The long term goal is to get rid of the infamous "all" struct. The purpose of this change is to represent with our types what "all" does at a clear glance.

So as a first step, this PR breaks down members from the vw struct into 8 other structs (this might change with time). 

-   InputConfig: Mostly consists of strings that represents file names to be read at some point in execution. This might be a code smell, in reality we shouldn't care where things are coming from.
-   OutputConfig: configuration related to anything that impacts how things are outputted by vw.
-   WeightConfig
-   FeatureConfig
-   ExampleConfig
-   UpdateConfig: config related on how things are mutated turn by turn by vw, i.e. learning rate
-   VWRuntimeConfig: config related on how vw is executed, i.e. is deamon mode on
-   GlobalState: all seems to be like a dumping ground on some sort of global state, i.e. current_pass, normalized_sum_norm_x, etc

This also provides a side effect benefit that vw struct starts to represent the domain in an easy way to grasp. Also, it's nice to know quickly if a variable is holding an i/o config, mutating state that reductions depend on, state used by vw to orchestrate itself or config that needs to be bubbled up to the core users of vw.

Feel free to point out if I'm missing more natural representations or want to talk about my reasoning.

-------------------

There's the open question if this constitutes as a breaking change or not, is this part of our contract with our users and if not we should start formalizing what is and what isn't.

Next steps after this merges (proposals and still designed):
- address reduction specific configs still in all
- address function pointers in all
- address the basis of all (learner, etc)
- refactor GlobalState to config where possible
- refactor code smells
- remove impossible states
- start building a contract and seperate CLI with implementation (one thing is how we represent things when we parse the command line arguments from a human vs how we represent things when we're actually running the reduction)
- refactor to decide which structs are accessible and required during setup (of reductions mostly) VS structs accessible during actual runtime: we should enforce with code what's possible instead of giving free reign thru all everywhere. This will be helpful to get more contributions of new reductions, and communicate clearly what someone can or can't do during each step


lalo:
Not expecting it to have successful builds, just worried about happy path for now.

vowpalwabbit/global_data.h
```diff
@@ -572,6 +580,17 @@ struct GlobalState
...
   uint64_t parse_mask;  // 1 << num_bits -1
+
+  /*
+   * following 7 are feature related settings?
+   * not so sure but feels like its state, related to features though
+   */
+  // Referenced by examples as their set of interactions. Can be overriden by reductions.
+  std::vector<std::vector<namespace_index>> interactions;
+
+  /* highly related to feature.ignore but this is just evil state*/
```
lalo, line 591: could be a code smell, in reality we care about ignoring if we have things to ignore

vowpalwabbit/global_data.h
```diff
@@ -541,6 +560,18 @@ struct GlobalState
    * not directly set by user
    */
   uint32_t wpp;
+
+  /*
+   * not input, used as state
+   * related to examples
+   */
+  size_t passes_complete;
```
lalo, line 576: this is also state but very related to ExampleConfig.numpasses and .pass_length

cs/unittest/TestMarshalNumeric.cs
```diff
@@ -1,4 +1,4 @@
-
+﻿
```
jackgerrits, line 1: Change intended?

vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -166,10 +166,10 @@ float cb_explore_adf_rnd::get_initial_prediction(example* ec)
...
   GD::foreach_feature<std::pair<float, float>, const float&, vec_add_with_norm, LazyGaussian>(
     w,
-    all->ignore_some_linear,
-    all->ignore_linear,
-    all->interactions,
-    all->permutations,
+    all->gs.ignore_some_linear,
+    all->fc.ignore_linear,
+    all->gs.interactions,
+    all->fc.permutations,
```
jackgerrits, line 172: When you see two different structures used like this, it brings up interesting questions

vowpalwabbit/cost_sensitive.cc
```diff
@@ -226,13 +226,13 @@ void print_update(vw& all, bool is_test, example& ec, multi_ex* ec_seq, bool act
         pred_buf << ec.pred.a_s[0].action;
       if (action_scores)
         pred_buf << ".....";
-      all.sd->print_update(all.holdout_set_off, all.current_pass, label_buf, pred_buf.str(), num_current_features,
-          all.progress_add, all.progress_arg);
+      all.sd->print_update(all.ec.holdout_set_off, all.gs.current_pass, label_buf, pred_buf.str(), num_current_features,
```
jackgerrits, line 229: I know you're planning on changing the names but when you see `ec` in VW, by default you usually think examples collection which is misleading here

vowpalwabbit/global_data.h
```diff
@@ -358,6 +358,328 @@ struct vw_logger
...
+   */
+  double normalized_sum_norm_x;
+
+  /*
+   * wpp = weights per problem
+   * feature related?
+   * parameters related? (see array_parameters.h)
+   * not directly set by user
+   */
+  uint32_t wpp;
```
jackgerrits, line 614: This is set by reduction setup functions

vowpalwabbit/cb_explore_adf_rnd.cc
lalo: feels like a halfway between "hey we support turning off and on ignoring features" but not really. Is there any case when we want to have something on the collections (len of list of things to ignore_linear>= 1) and NOT ignore (bool ignore_some_linear= false)? if so, the representation of these structs should represent what is valid and avoid invalid states.

cs/unittest/TestMarshalNumeric.cs
lalo: not intented

vowpalwabbit/multiclass.cc
```diff
@@ -129,14 +129,14 @@ void print_score(vw& all, example& ec, uint32_t prediction)
   std::stringstream label_ss;
   label_ss << ec.l.multi.label;
 
-  all.sd->print_update(all.holdout_set_off, all.current_pass, label_ss.str(), pred_ss.str(), ec.num_features,
-      all.progress_add, all.progress_arg);
+  all.sd->print_update(all.ec.holdout_set_off, all.gs.current_pass, label_ss.str(), pred_ss.str(), ec.num_features,
```
lalo, line 132: @jackgerrits feels like we could implement a different version of print which already has access to all.ec and all.oc to the very least. I think we could even get rid of the reference to all.gs. Instead of sending things back and forth just set once and call from everywhere else.

vowpalwabbit/cb_adf.cc
```diff
@@ -567,7 +567,7 @@ base_learner* cb_adf_setup(options_i& options, vw& all)
 
   auto base = as_multiline(setup_base(options, all));
   all.p->lp = CB::cb_label;
-  all.label_type = label_type_t::cb;
+  all.gs.label_type = label_type_t::cb;
```
lalo, line 570: @jackgerrits should this be encapsulated on it's own function? seems like in a lot of places we do the same operation of p->lp and label_type. Could even be like a state machine in case there's more to do when changing this.

lalo: see this commit for all cases https://github.com/VowpalWabbit/vowpal_wabbit/pull/2483/commits/529752434cb171175caad38b765a52e9f21ecaac

jackgerrits: Yeah encapsulating this could be good! I agree

vowpalwabbit/multiclass.cc
jackgerrits: To add to this it would be interesting to think how this functionality should work for a library - are we just printing to a different output device? Or we updating some state and printing is one possible thing we do with it?

vowpalwabbit/cost_sensitive.cc
lalo: changed to example_config, to avoid confusion while doing pr. We can hammer down on the final names later.

cs/unittest/TestMarshalNumeric.cs
lalo: i think this file shouldn't even be part of the repo

lalo: https://github.com/VowpalWabbit/vowpal_wabbit/pull/2492

vowpalwabbit/global_data.h
```diff
@@ -359,6 +359,342 @@ struct vw_logger
...
+   * Should be part of learner?
+   * not input, used as state
+   */
+  uint64_t current_pass;
+
+  /*
+   * not input, used as state
+   * related to examples
+   */
+  size_t passes_complete;
```
peterychang, line 606: I _think_ `current_pass` and `passes_complete` are almost the same thing. AFAIK the only difference is that `passes_complete` counts based on the input file (incremented once an input is fully read and reset), and `current_pass` is based on the example itself (incremented after the last example in a file is processed).

Maybe add a todo to ensure these two variables are being used in a consistent way, If my understanding is correct, `passes_complete` should only ever be used in the parser thread while `current_pass` should only ever be used in the learner thread

vowpalwabbit/global_data.h
```diff
@@ -373,8 +709,17 @@ struct vw
 
   AllReduceType all_reduce_type;
   AllReduce* all_reduce;
-
-  bool chain_hash = false;
+
+  InputConfig ic;
```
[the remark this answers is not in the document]
lalo: Yep, as outlined in the description of the PR I didn't bother to name properly bc feel that can be decided once and then replace instead of iterating over and over. Did not know how people would react to the word config, etc

vowpalwabbit/global_data.h
```diff
@@ -359,6 +359,342 @@ struct vw_logger
...
+
+  /* related to example/features bc namespace, this could be refactored */
+  // feature_dict can be created in either loaded_dictionaries or namespace_dictionaries.
+  // use shared pointers to avoid the question of ownership
+  std::vector<dictionary_info> loaded_dictionaries;  // which dictionaries have we loaded from a file to memory?
+  // This array is required to be value initialized so that the std::vectors are constructed.
+  std::array<std::vector<std::shared_ptr<feature_dict>>, NUM_NAMESPACES>
+      namespace_dictionaries{};  // each namespace has a list of dictionaries attached to it
+
+  label_type_t label_type;
```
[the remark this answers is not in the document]
lalo: well I meant 'state' in the sense that it was getting mutated in multiple places, but this makes sense projecting them as details of the reduction stack itself.

vowpalwabbit/global_data.cc
```diff
@@ -314,95 +314,96 @@ vw::vw()
   p = nullptr;
 
   reg_mode = 0;
-  current_pass = 0;
 
-  data_filename = "";
+  gs.current_pass = 0;
```
peterychang, line 318: These should be initialized in each subcomponents' constructor if at all possible. Dunno if that belongs in this PR or in a followup

lalo: follow-up, feel free to unblock yourself if you're on top of this branch. We should get this in, to start chopping all those next tasks if not the PR will be to crazy to parse

lalo:
This PR is out of date and a draft, I'll re-open later.

PR #2484, opened 2020-06-25
Fix warnings to do with memset of non-trivial types

jackgerrits:
Still have search to do, will address in a separate PR.


vowpalwabbit/cb.cc
```diff
@@ -281,11 +285,11 @@ void parse_label(parser* p, shared_data* sd, void* v, v_array<VW::string_view>&
 
   ld->action = (uint32_t)hashstring(words[0].begin(), words[0].length(), 0);
 
-  words.begin()++;
-
+  // Removing the first element of a vector is not efficient at all, every element must be copied/moved.
```
olgavrou, line 288: I am not familiar with the `words` vector/array, but from a quick search it doesn't seem to have index lookups, or if it does they don't go past the 3rd index. Would it be worth making it a linked list instead of a vector?

jackgerrits: The words vector (like you said) is often accessed randomly, even if it is only to around the 3rd index. It looks to me like `std::list` doesn't allow for random access?

jackgerrits: Discussed offline and std::list doesn't make sense because the container is always small in length.

lalo: Doesn't the previous code remove the first one, then call parse_label? or am i missing something

jackgerrits:
```
[xxxx]
 ^begin
 
>> words.begin()++;

[xxxx]
  ^begin

>> CB::parse_label(p, sd, &(ld->event), words);
>> words.begin()--;

[xxxx]
 ^begin
```

It doesn't remove it, it moves the begin pointer of the buffer so that to `parse_label` it doesn't exist. This is of course very unsafe and not something other containers really let you do. It was able to happen because currently (to be fixed) `v_array` returns a reference to it's own begin pointer from `begin`

jackgerrits:
I did some investigation into the performance impact of this change.

Key takeaways:
- Newer toolchains do a better job at optimizing usage of `std::vector`
- GCC produces the fastest code by quite a large margin
- Some toolchains result in faster execution from `std::vector`, some have a minimal (maybe in the range of variance) drop in performance

Based on this I think we should move forward with this change.

Toolchain version | Code built | Avg 10 runs (s) | Difference from best
-- | -- | -- | --
Clang 10 | Master | 3.42026 | 14.19%
Clang 10 | PR | 3.4257 | 14.37%
GCC 9.3 | Master | 2.99523 | 0.00%
GCC 9.3 | PR | 3.03684 | 1.39%
MSVC v141 | Master | 3.72218 | 24.27%
MSVC v141 | PR | 4.06887 | 35.84%
MSVC v142 | Master | 4.29236 | 43.31%
MSVC v142 | PR | 3.87345 | 29.32%

**Note:** MSVC v141 was built using checked in vcxproj, and v142 was generated using CMake and so the flags may be different. So direct comparison between them is not recommended.

PR #2490, opened 2020-06-30
Proposal: Use a power of 10 lookup table to improve VW text parser perf by 10%

jackgerrits:
I'd like to get people's thoughts about if we should do this or not.

By noticing that a specific `powf` call in the float parser function always raises 10 to some integer power, we can apply a lookup table. Float parsing is the majority of time spent in VW text parsing. This change **reduces overall runtime** in an RCV1 parsing benchmark by about **8%**. 

Running a benchmark of the `parseFloat` function on the rcv1 dataset, which heavily uses the float exponent notation (e.g. `2.5941234e-02`) that this optimizes shows a speedup of **47%** for the function.
```
2020-06-30 17:01:55
Running ./perf_tests
Run on (12 X 3800 MHz CPU s)
CPU Caches:
  L1 Data 32K (x6)
  L1 Instruction 32K (x6)
  L2 Unified 512K (x6)
  L3 Unified 16384K (x2)
Load Average: 1.03, 1.18, 1.26
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
-----------------------------------------------------------
Benchmark                 Time             CPU   Iterations
-----------------------------------------------------------
parse_float            3200 ns         3199 ns       222906
parse_float_fast       1693 ns         1692 ns       408178
```

`parse_float` is the current `parseFloat` implementation
`parse_float_fast` is the proposed implementation of `parseFloat`.


lalo:
I like the idea, if it's controversial a middle point would be to have a flag to disable/enable and drive that process with cmake.

peterychang:
A bit fragile having to handle positive and negative like that, but I think its fine since its all encapsulated in the function calls

jackgerrits:
I believe this should catch all cases? To the caller it should be practically indistinguishable compared to calling `std::pow`, do you agree?

peterychang:
Since the min and max are outside of the guaranteed precision limits you could get some rounding errors, so its not guaranteed to be exactly the same every time (probably will be though?)
Also, where did you get the limits of [-44, 38]?

jackgerrits:
`+38` is the max value of a float: https://en.cppreference.com/w/c/types/limits
`-44` was through testing. Beyond that value I simply get 0 back during testing. But this is almost overkill because the float epsilon is usually around `1.192093e-07`, so values become indistinguishable pretty quickly.

My concern about this change is since this stuff can be platform dependent, what am I forgetting or overlooking here?

peterychang:
You should use the built-in limits instead of arbitrary values
`FLT_MIN_10_EXP` and `FLT_MAX_10_EXP`
https://en.cppreference.com/w/c/types/limits

jackgerrits:
Ah excellent point! Now I just need to work out how to dynamically generate that array...

PR #2492, opened 2020-06-30
Remove autogen file

lalo:
This makes sense, but let me know if I'm wrong.


lalo:
for the curious https://github.com/VowpalWabbit/vowpal_wabbit/blob/9137a493369677a82120479fb7e86ac4b25360e2/cs/unittest/cs_unittest.csproj#L186

PR #2494, opened 2020-07-01
Migrate away from deprecated strerror function

jackgerrits:
`strerror` is deprecated and not threadsafe. This PR replaces usage with the platform specific safe versions:
- `strerror_s` on Windows
- `strerror_l` on Linux

### DIFF
--- a/vowpalwabbit/allreduce_sockets.cc
+++ b/vowpalwabbit/allreduce_sockets.cc
@@ -67,7 +67,6 @@ socket_t AllReduceSockets::sock_connect(const uint32_t ip, const int port)
     std::stringstream msg;
     if (!quiet)
     {
-
       msg << "connect attempt " << count << " failed: " << VW::strerror_to_string(errno);
       cerr << msg.str() << endl;
     }
@@ -214,7 +213,8 @@ void AllReduceSockets::all_reduce_init()
       {
         if (listen(sock, kid_count) < 0)
         {
-          if (!quiet) cerr << "listen: " << VW::strerror_to_string(errno) << endl;
+          if (!quiet)
+            cerr << "listen: " << VW::strerror_to_string(errno) << endl;
           CLOSESOCK(sock);
           sock = getsock();
         }
@@ -339,7 +339,7 @@ void AllReduceSockets::broadcast(char* buffer, const size_t n)
       int read_size = recv(socks.parent, buffer + parent_read_pos, (int)count, 0);
       if (read_size == -1)
       {
-        THROW(" recv from parent: " << VW::strerror_to_string(errno));
+        THROW("recv from parent: " << VW::strerror_to_string(errno));
       }
       parent_read_pos += read_size;
     }
--- a/vowpalwabbit/allreduce_sockets.cc
+++ b/vowpalwabbit/allreduce_sockets.cc
@@ -67,7 +67,8 @@ socket_t AllReduceSockets::sock_connect(const uint32_t ip, const int port)
     std::stringstream msg;
     if (!quiet)
     {
-      msg << "connect attempt " << count << " failed: " << strerror(errno);
+
+      msg << "connect attempt " << count << " failed: " << VW::strerror_to_string(errno);
       cerr << msg.str() << endl;
     }
 #ifdef _WIN32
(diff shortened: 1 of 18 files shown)


jackgerrits:
Looks like MacOS needs some special attention here - I'll have to give it a test.

vowpalwabbit/vw_exception.h
```diff
@@ -104,6 +111,31 @@ class strict_parse_exception : public vw_exception
...
+  if (result != 0) { return "unknown"; }
+
+  auto length = std::strlen(error_message_buffer.data());
+  return std::string(error_message_buffer.data(), length);
+#else
+  // Passing "" for the locale means use the default system locale
+  locale_t locale = newlocale(LC_ALL_MASK, "", static_cast<locale_t>(0));
+
+  if (locale == static_cast<locale_t>(0)) {
+    return "Failed to create locale";
```
lalo, line 129: on failure newlocale sets errorno to specific error cases. Not sure if we want to incorporate that for beefier error messages. Not sure if this path is worthy of it though

vowpalwabbit/vw_exception.h
```diff
@@ -104,6 +111,31 @@ class strict_parse_exception : public vw_exception
...
+  }
+
+  // Even if error_number is unknown, will return a "Unknown error nnn" message.
+  std::string message = strerror_l(error_number, locale);
```
lalo, line 133: actually might be worth it over here

vowpalwabbit/vw_exception.h
jackgerrits: Yeah, I thought about that. But then to get the error message you need to create a locale - the exact thing that failed... It seems like a sort of fail loop?

lalo: would it be worth it to report the number only?

lalo: aren't those lookable or fixed? I actually don't remember

vowpalwabbit/vw_exception.h
```diff
@@ -104,6 +111,31 @@ class strict_parse_exception : public vw_exception
   ~strict_parse_exception() noexcept override = default;
 };
 
+inline std::string strerror_to_string(int error_number)
+{
+#ifdef _WIN32
+  constexpr auto BUFFER_SIZE = 256;
+  std::array<char, BUFFER_SIZE> error_message_buffer;
+  auto result = strerror_s(error_message_buffer.data(), error_message_buffer.size(), error_number);
+  if (result != 0) { return "unknown"; }
```
lalo, line 120: do we want to return just "unknown"?

vowpalwabbit/vw_exception.h
jackgerrits: Yeah we could just report out the number so the user can look it up. Seems like a reasonable backup

vowpalwabbit/allreduce_sockets.cc
```diff
@@ -67,7 +67,8 @@ socket_t AllReduceSockets::sock_connect(const uint32_t ip, const int port)
     std::stringstream msg;
     if (!quiet)
     {
-      msg << "connect attempt " << count << " failed: " << strerror(errno);
+
```
lalo, line 70: nit extra space

vowpalwabbit/allreduce_sockets.cc
```diff
@@ -213,8 +214,7 @@ void AllReduceSockets::all_reduce_init()
       {
         if (listen(sock, kid_count) < 0)
         {
-          if (!quiet)
-            cerr << "listen: " << strerror(errno) << endl;
+          if (!quiet) cerr << "listen: " << VW::strerror_to_string(errno) << endl;
```
lalo, line 217: should we stick to one format of if quiet in this file?

vowpalwabbit/allreduce_sockets.cc
```diff
@@ -339,7 +339,7 @@ void AllReduceSockets::broadcast(char* buffer, const size_t n)
       int read_size = recv(socks.parent, buffer + parent_read_pos, (int)count, 0);
       if (read_size == -1)
       {
-        THROW(" recv from parent: " << strerror(errno));
+        THROW(" recv from parent: " << VW::strerror_to_string(errno));
```
lalo, line 342: nit: since we're here, is there a purpose on that first space? feel free to ignore this minimal

vowpalwabbit/allreduce_sockets.cc
jackgerrits: This the clang format file at work - short blocks get collapsed to one line.

jackgerrits: Ill revert this line for now - I don't want to edit each one in this file

lalo: got it, your call

lalo:
Did you figure out a way to test this?

jackgerrits:
Perhaps an acceptable way to test is to call it with a certain error message and confirm that the returned message is not the "unknown" message implying it successfully retrieved the message.

PR #2495, opened 2020-07-01
Improve the handling of ignoring warnings across platforms and toolch…

jackgerrits:
…ain versions

### DIFF
--- a/vowpalwabbit/parser.cc
+++ b/vowpalwabbit/parser.cc
@@ -180,19 +180,19 @@ void reset_source(vw& all, size_t numbits)
       if (all.p->input->isbinary())
       {
         all.p->reader = read_cached_features;
-VW_DISABLE_WARNING_PUSH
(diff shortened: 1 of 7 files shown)


vowpalwabbit/parser.cc
```diff
@@ -480,9 +482,10 @@ void enable_sources(vw& all, bool quiet, size_t passes, input_options& input_opt
     int one = 1;
```
lalo, line 482: what about line 388 on this file? is that meant to be migrated to this new style

lalo:
Seems like there's a pragma push on lda_core.cc, not sure if that should be part of this PR. There might be more didn't do a full search.

vowpalwabbit/parser.cc
jackgerrits: I plan on following this up and migrating thing to this system. There will be a few to do - so I'll do it in a separate PR

PR #2496, opened 2020-07-02
Flatbuffers

Sharad24:
Part of the _VW support for FlatBuff and/or Protobuf_ Project

This PR introduces :-
- Flatbuffer Example Schema (`vowpalwabbit/parser/flatbuffer/schema/example.fbs`)
- Flatbuffer Example parsing (in `vowpalwabbit/parser/flatbuffer/`)
- Flatbuffer Example Creation for an existing dataset in formats supported by VW (in `utl/flatbuffer`)
- Flatbuffer Inspection through a python file (in `utl/flatbuffer/tools`)
- Runtests for label types (in `test/RunTests`)
- All Runtests (in `test/flatbuffer`)
- Unit-tests for flatbuffer creation and flatbuffer parsing

There are some things that I'm still working on :-
1. Flatbuffer RunTests (`test/flatbuffer/Runtests`) does not pass all tests
2. A few RunTests in `test/Runtests` are missing (e.g. Slates)


test/flatbuffer/README.md
```diff
@@ -0,0 +1,17 @@
+#### Running flatbuffer tests
+
+Note: **Work in Progress** (Not all tests run as of now.)
+
+To run all tests in RunTests for flatbuffers, first all flatbuffer data files need to be created. 
+For this, just replacing `../../build/vowpalwabbit/vw` with `../../build/utl/flatbuffer/to_flatbuff` 
+would work, provided some flags are removed. This includes `skips`, `ngrams`, etc. A list of 
+these is present in `create_bash_script.py` which is used to create the file `create_datafiles.sh`.
```
olgavrou, line 8: if `create_datafiles.sh` is a generated file, should it also be checked in?

utl/flatbuffer/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+set(CMAKE_CXX_STANDARD 11)
```
olgavrou, line 1: It looks like `utl/flatbuffer` directory should be under `vowpalwabbit/parser/flatbuffer/utl` with the other flatbuffer files. In that case you can generate the below executables in the `parser/flatbuffer`'s CMakeLists.txt.

utl/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+add_subdirectory(flatbuffer)
```
olgavrou, line 1: this could be removed if `utl/flatbuffer` moves under `vowpalwabbit/parser/flatbuffer`

utl/flatbuffer/txt_to_flat.cc
```diff
@@ -0,0 +1,97 @@
...
+// license as described in the file LICENSE.
+
+#ifdef _WIN32
+#define NOMINMAX
+#include <WinSock2.h>
+#else
+#include <sys/socket.h>
+#include <arpa/inet.h>
+#endif
+#include <sys/timeb.h>
```
olgavrou, line 12: should be able to remove all the relative paths and just do : `#include "parse_args.h"`, etc

utl/flatbuffer/txt_to_flat.cc
```diff
@@ -0,0 +1,97 @@
...
+#include "../../vowpalwabbit/parse_args.h"
+#include "../../vowpalwabbit/parse_regressor.h"
+#include "../../vowpalwabbit/accumulate.h"
+#include "../../vowpalwabbit/best_constant.h"
+#include "../../vowpalwabbit/vw_exception.h"
+#include <fstream>
+
+#include "../../vowpalwabbit/vw.h"
```
olgavrou, line 20: Any files that have already been included in `vw_to_flat.h` which is in turn included here don't need to be included again (e.g. `vw.h`)

utl/flatbuffer/vw_to_flat.cc
```diff
@@ -0,0 +1,264 @@
+// Copyright (c) by respective owners including Yahoo!, Microsoft, and
+// individual contributors. All rights reserved. Released under a BSD (revised)
+// license as described in the file LICENSE.
+
+#include <sys/timeb.h>
+#include <fstream>
+#include <vector>
+
+#include "vw_to_flat.h"
```
olgavrou, line 9: same for these include paths

vowpalwabbit/parser.cc
```diff
@@ -586,6 +588,14 @@ IGNORE_DEPRECATED_USAGE_END
 
         all.p->decision_service_json = input_options.dsjson;
       }
+      else if (input_options.flatbuffer)
+      {
+        all.flat_converter = new VW::parsers::flatbuffer::parser(all.data_filename);
+        all.flat_converter->init();
+
+        all.p->reader = VW::parsers::flatbuffer::flatbuffer_to_examples;
+        all.p->text_reader = VW::parsers::flatbuffer::read_flatbuffer;
```
olgavrou, line 597: You could omit setting `all.p->text_reader` here which is set by default above to `VW::read_lines` since you don't need it. You should then be able to remove the `read_flatbuffer` function from `parse_flat_example` which doesn't do anything

vowpalwabbit/parser/flatbuffer/parse_label.cc
```diff
@@ -0,0 +1,138 @@
...
+void parser::parse_ccb_label(polylabel* l, const CCBLabel* label)
+{
+  l->conditional_contextual_bandit.weight = label->weight();
+  if (label->example_type() == 1) l->conditional_contextual_bandit.type = CCB::example_type::shared;
+  else if (label->example_type() == 2) l->conditional_contextual_bandit.type = CCB::example_type::action;
+  else if (label->example_type() == 3) {
+    l->conditional_contextual_bandit.type = CCB::example_type::slot;
+
+    if (label->explicit_included_actions() != nullptr){
+      for (auto i=0;i<label->explicit_included_actions()->Length();i++){
```
olgavrou, line 54:
```suggestion
      for (size_t i = 0; i < label->explicit_included_actions()->Length(); i++){
```

vowpalwabbit/parser/flatbuffer/parse_label.cc
```diff
@@ -0,0 +1,138 @@
...
+  else l->conditional_contextual_bandit.type = CCB::example_type::unset;
+    // l->conditional_contextual_bandit.outcome->cost = label->outcome()->cost();
+  }
+}
+
+void parser::parse_cb_eval_label(polylabel* l, const CB_EVAL_Label* label)
+{
+  l->cb_eval.action = label->action();
+  l->cb_eval.event.weight = label->event()->weight();
+  for (auto i=0;i<label->event()->costs()->Length();i++){
```
olgavrou, line 77:
```suggestion
  for (size_t i = 0; i < label->event()->costs()->Length(); i++){
```

vowpalwabbit/parser/flatbuffer/parse_label.cc
```diff
@@ -0,0 +1,138 @@
...
+    f.action = label->event()->costs()->Get(i)->action();
+    f.probability = label->event()->costs()->Get(i)->probability();
+    f.partial_prediction = label->event()->costs()->Get(i)->partial_pred();
+    l->cb_eval.event.costs.push_back(f);
+  }
+}
+
+void parser::parse_cs_label(polylabel* l, const CS_Label* label)
+{
+  for (auto i=0;i<label->costs()->Length();i++){
```
olgavrou, line 89:
```suggestion
  for (size_t i = 0; i < label->costs()->Length(); i++){
```

vowpalwabbit/parser/flatbuffer/parse_label.cc
```diff
@@ -0,0 +1,138 @@
...
+void parser::parse_mc_label(shared_data* sd, polylabel* l, const MultiClass* label)
+{
+  uint32_t word = label->label();
+  l->multi.label = sd->ldict ? (uint32_t)sd->ldict->get(std::to_string(word)) : word;
+  l->multi.weight = label->weight();
+}
+
+void parser::parse_multi_label(polylabel* l, const MultiLabel* label)
+{
+  for (auto i=0;i<label->labels()->Length();i++)
```
olgavrou, line 108:
```suggestion
  for (size_t i = 0; i < label->labels()->Length(); i++)
```

vowpalwabbit/parser/flatbuffer/parse_flat_example.cc
```diff
@@ -0,0 +1,183 @@
...
+  all->p->lp.default_label(&ae->l);
+  parse_flat_label(all->sd, ae, eg);
+
+  if (flatbuffers::IsFieldPresent(eg, Example::VT_TAG)){
+    ae->tag = v_init<char>();
+    VW::string_view tag(eg->tag()->c_str());
+    push_many(ae->tag, tag.begin(), tag.size());
+  }
+
+  for (int i=0; i < eg->namespaces()->size(); i++){
```
olgavrou, line 91:
```suggestion
  for (size_t i = 0; i < eg->namespaces()->size(); i++){
```

vowpalwabbit/parser/flatbuffer/parse_flat_example.cc
```diff
@@ -0,0 +1,183 @@
...
+    _c_hash = all->p->hasher(ns->name()->c_str(), ns->name()->Length(), all->hash_seed);
+  }
+  else {
+    ae->indices.push_back(ns->hash());
+    temp_index = ns->hash();
+  }
+
+  features& fs = ae->feature_space[temp_index];
+
+  for (int j=0; j<ns->features()->size(); j++){
```
olgavrou, line 111:
```suggestion
  for (size_t j = 0; j < ns->features()->size(); j++){
```

vowpalwabbit/parser/flatbuffer/parse_flat_example.h
```diff
@@ -0,0 +1,51 @@
...
+    void init();
+    parser(std::string filename);
+    parser(uint8_t *buffer_pointer);
+    const VW::parsers::flatbuffer::ExampleCollection* data();
+
+private:
+    const VW::parsers::flatbuffer::ExampleCollection* _data;
+    std::string _filename;
+    uint8_t* _flatbuffer_pointer;
+    int _example_index;
```
olgavrou, line 34:
```suggestion
    size_t _example_index;
```

vowpalwabbit/parser/flatbuffer/parse_flat_example.cc
```diff
@@ -0,0 +1,183 @@
+// Copyright (c) by respective owners including Yahoo!, Microsoft, and
```
olgavrou, line 1: So far we have `parse_example.h` and `parse_example_json.h`. Maybe rename this to `parse_example_flatbuffer` for consistency?

vowpalwabbit/parser/flatbuffer/parse_flat_example.cc
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  features& fs = ae->feature_space[temp_index];
+
+  for (int j=0; j<ns->features()->size(); j++){
+    parse_features(all, ae, fs, ns->features()->Get(j));
+  }
+}
+
+void parser::parse_features(vw* all, example* ae, features& fs, const Feature* feature)
```
olgavrou, line 116: It looks like `example* ae` is not used here so it could be removed from the function's parameters

vowpalwabbit/parser/flatbuffer/parse_flat_example.cc
olgavrou: all these int->size_t suggestions stem from the fact that any size returned from flatbuffers will be an `unsigned int` and so should use the same type to access those sizes

vowpalwabbit/parser/flatbuffer/parse_flat_example.cc
```diff
@@ -0,0 +1,183 @@
...
+    temp_index = (uint8_t)ns->name()->c_str()[0];
+    ae->indices.push_back(temp_index);
+    _c_hash = all->p->hasher(ns->name()->c_str(), ns->name()->Length(), all->hash_seed);
+  }
+  else {
+    ae->indices.push_back(ns->hash());
+    temp_index = ns->hash();
+  }
+
+  features& fs = ae->feature_space[temp_index];
```
olgavrou, line 109:
```suggestion
  auto& fs = ae->feature_space[temp_index];
```

test/flatbuffer/README.md
Sharad24: Yes. It has some extra commands for copying reference files (stderr, predict, etc). Although, I could generate these from the Python file as well.

vowpalwabbit/parser/flatbuffer/parse_flat_example.cc
Sharad24: Oh okay, thanks for pointing it out. I'll change these

vowpalwabbit/parser/flatbuffer/parse_flat_example.cc
```diff
@@ -0,0 +1,183 @@
...
+    push_many(ae->tag, tag.begin(), tag.size());
+  }
+
+  for (int i=0; i < eg->namespaces()->size(); i++){
+    parse_namespaces(all, ae, eg->namespaces()->Get(i));
+  }
+}
+
+void parser::parse_namespaces(vw* all, example* ae, const Namespace* ns)
+{
```
olgavrou, line 97: `ae->indices` is a `std::array` of `namespace_index`'s. You could declare `temp_index` as a `namespace_index` here (and there will be no need for the `uint8_t` cast of `ns->name()`). Also `ae->indices.push_back(tmp_index)` could be done once outside of the if/else clause.

utl/flatbuffer/vw_to_flat.cc
```diff
@@ -0,0 +1,264 @@
...
+
+void to_flat::create_no_label(example* v, flatbuffers::Offset<void>& label, VW::parsers::flatbuffer::Label& label_type)
+{
+  label = VW::parsers::flatbuffer::Createno_label(_builder, (uint8_t)'\000').Union();
+}
+
+void to_flat::convert_txt_to_flat(vw& all)
+{
+  std::vector<flatbuffers::Offset<VW::parsers::flatbuffer::Example>> examplecollection;
+  example* v = all.p->ready_parsed_examples.pop();
```
jackgerrits, line 176: Please use a more descriptive name than `v`

utl/flatbuffer/vw_to_flat.cc
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    uint64_t multiplier = (uint64_t)all.wpp << all.weights.stride_shift();
+    if (multiplier != 1) {
+      for (features& fs : *v) {
+        for (auto& j : fs.indicies) {j /= multiplier;}
+      }
+    }
+    std::vector<flatbuffers::Offset<VW::parsers::flatbuffer::Namespace>> namespaces;
+    for (namespace_index& ns : v->indices)
```
jackgerrits, line 224: const?

vowpalwabbit/parser/flatbuffer/parse_label.cc
```diff
@@ -0,0 +1,138 @@
...
+#include "../../best_constant.h"
+#include "parse_flat_example.h"
+#include "generated/example_generated.h"
+
+
+namespace VW {
+namespace parsers {
+namespace flatbuffer {
+
+void parser::parse_simple_label(shared_data* sd, polylabel* l, const SimpleLabel* label)
```
jackgerrits, line 23: Rather than needing to implement all of the label types here? Can this become part of the label parser definition?

vowpalwabbit/parser/flatbuffer/schema/example.fbs
```diff
@@ -0,0 +1,101 @@
+namespace VW.parsers.flatbuffer;
+
+table Feature {
+  name:string;
```
jackgerrits, line 4: Does this mean that strings are always required? Is it possible to produce a file that only specifies the hashes?

Sharad24: Nope, the name is not always required. Can use a file with just hashes as well. In case, the name is provided, the hashes are produced.

test/flatbuffer/RunTests
```diff
@@ -0,0 +1,1693 @@
+#!/usr/bin/env perl
```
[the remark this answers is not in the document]
Sharad24: These tests are not to be run every time, they're there as a sanity check only. There are some tests added to the existing Runtests file (which will be run everytime) for checking the flatbuffer code.

test/flatbuffer/create_bash_script.py
```diff
@@ -0,0 +1,131 @@
...
+
+    return newc
+
+tags = ['--audit', '-c ','--bfgs']
+td = ['--passes', '--ngram', '--skips', '-q', '-b']
+
+names, _c, _d, datafiles = extract_commands(lines, tags, td)
+
+commands = change_datafile_names(_d, datafiles)
+to_skip = [60, 61, 92, 96, 212, 213, 214, 215, 216, 217, 218, 219, 83, 88, 89, 210, 198, 202, 209, 83, 176, 177, 203, 206, 207, 208, 193, 194, 195, 155, 156, 152, 149, 137, 125, 111, 112]
```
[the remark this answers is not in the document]
Sharad24: Some of them don't need to be tested, some of them are failing (which we're tracking and continuously resolving)

For not to be tested ones an e.g. is the daemon test - 60

Sharad24:
PR for Flatbuffers installation in docker images: https://github.com/VowpalWabbit/docker-images/pull/24

utl/flatbuffer/vw_to_flat.cc
```diff
@@ -0,0 +1,264 @@
...
+#include "../../vowpalwabbit/options_boost_po.h"
+#include "../../vowpalwabbit/parser/flatbuffer/generated/example_generated.h"
+
+to_flat::to_flat() : _builder(1) {};
+
+std::string to_flat::get_label_string(label_type_t label_type)
+{
+  switch (label_type)
+  {
+  case label_type_t::nolabel:
```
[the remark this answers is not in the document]
Sharad24: Just to clarify, do you mean moving the function into `global_data.h` or a similar file?

utl/flatbuffer/vw_to_flat.cc
```diff
@@ -0,0 +1,264 @@
...
+#include "../../vowpalwabbit/options.h"
+#include "../../vowpalwabbit/parse_args.h"
+#include "../../vowpalwabbit/parse_regressor.h"
+#include "../../vowpalwabbit/accumulate.h"
+#include "../../vowpalwabbit/best_constant.h"
+#include "../../vowpalwabbit/vw_exception.h"
+#include "../../vowpalwabbit/options_boost_po.h"
+#include "../../vowpalwabbit/parser/flatbuffer/generated/example_generated.h"
+
+to_flat::to_flat() : _builder(1) {};
```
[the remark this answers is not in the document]
Sharad24: Removed the constructor

vowpalwabbit/parser/flatbuffer/parse_flat_example.cc
Sharad24: Changed

cmake/BuildFlatBuffers.cmake
```diff
@@ -0,0 +1,148 @@
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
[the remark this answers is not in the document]
Sharad24: Yes, I think so. More about that here: https://www.gnu.org/licenses/license-compatibility.en.html

vowpalwabbit/parser/flatbuffer/parse_label.cc
Sharad24: It could, but doing this would require including the flatbuffer schema-generated file (`parser/flatbuffer/generated/example_generated.h`) in every label parser definition. 

Another reasoning behind this was to keep parser definitions separate altogether, which VW aims to do in the long run as mentioned by @rajan-chari

vowpalwabbit/parser/flatbuffer/parse_label.cc
```diff
@@ -0,0 +1,138 @@
...
+      ccb_outcome.cost = label->outcome()->cost();
+      ccb_outcome.probabilities = v_init<ACTION_SCORE::action_score>();
+
+      for (auto const& as : *(label->outcome()->probabilities()))
+        ccb_outcome.probabilities.push_back({as->action(), as->score()});
+
+      l->conditional_contextual_bandit.outcome = &ccb_outcome;
+    }
+  else l->conditional_contextual_bandit.type = CCB::example_type::unset;
+    // l->conditional_contextual_bandit.outcome->cost = label->outcome()->cost();
```
Sharad24, line 69: Remove comment

vowpalwabbit/CMakeLists.txt
```diff
@@ -275,6 +278,7 @@ set(vw_all_sources
 )
 
 add_library(vw STATIC ${vw_all_sources} ${vw_all_headers})
+add_dependencies(vw fbschemas)
```
jackgerrits, line 281: the add_subdirectory(parser/flatbuffer) should be in this file to avoid the file above this one bringing in a dependency this file needs

vowpalwabbit/CMakeLists.txt
```diff
@@ -295,7 +299,9 @@ endif()
 target_include_directories(vw PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
-  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
+  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+  ${FLATBUFFERS_INCLUDE_DIR}
```
jackgerrits, line 303: This should be brought in with a flatbuffers target not like this

vowpalwabbit/cbify.cc
```diff
@@ -435,21 +435,29 @@ base_learner* cbify_setup(options_i& options, vw& all)
   if (data->use_adf)
   {
     multi_learner* base = as_multiline(setup_base(options, all));
-    if (use_cs)
+    if (use_cs){
```
jackgerrits, line 438: please run clang-format

vowpalwabbit/parse_args.cc
```diff
@@ -445,8 +445,8 @@ input_options parse_source(vw& all, options_i& options)
       .add(make_option("no_stdin", all.stdin_off).help("do not default to reading from stdin"))
       .add(make_option("no_daemon", all.no_daemon).help("Force a loaded daemon or active learning model to accept local input instead of starting in daemon mode"))
       .add(make_option("chain_hash", parsed_options.chain_hash)
-               .help("enable chain hash for feature name and string feature value. e.g. {'A': {'B': 'C'}} is hashed as A^B^C"));
-
+               .help("enable chain hash for feature name and string feature value. e.g. {'A': {'B': 'C'}} is hashed as A^B^C"))
+      .add(make_option("flatbuffer", parsed_options.flatbuffer).help("Input a FlatBuffer File"));
```
jackgerrits, line 449: Suggestion: change description to "data file will be interpreted as a flatbuffer file"

vowpalwabbit/global_data.h
```diff
@@ -345,7 +348,8 @@ struct vw
 
   uint32_t hash_seed;
 
-  std::string data_filename;  // was vm["data"]
+  VW::parsers::flatbuffer::parser* flat_converter;
```
jackgerrits, line 351: Please change this to a unique_ptr to fix the current leak

vowpalwabbit/CMakeLists.txt
olgavrou: The target created `fbschemas` is a custom target and afaik you can't use `target_include_directories` on custom targets. However the flatbuffers include dir needs to be added to vw. Is there a better way?

jackgerrits: `FLATBUFFERS_INCLUDE_DIR` comes from 
`find_package(flatbuffers REQUIRED)` right?

If it is header only we should do something similar to what we do for RapidJson.

olgavrou: Yup I see, addressed [here](https://github.com/VowpalWabbit/vowpal_wabbit/pull/2496/commits/d2952344eda31d3ee5f5275d64aa04670c83e2ad)

test/unit_test/flatbuffer_parser_test.cc
```diff
@@ -71,7 +71,7 @@ BOOST_AUTO_TEST_CASE(check_flatbuffer)
   uint8_t* buf = build._builder.GetBufferPointer();
   int size = build._builder.GetSize();
 
-  all->flat_converter = new VW::parsers::flatbuffer::parser(buf);
+  all->flat_converter = std::unique_ptr<VW::parsers::flatbuffer::parser>(new VW::parsers::flatbuffer::parser(buf));
```
jackgerrits, line 74: You can use `VW::make_unique` to eliminate the need for new

```suggestion
  all->flat_converter = VW::make_unique<VW::parsers::flatbuffer::parser>(buf);
```

.github/workflows/build_docs.yml
```diff
@@ -14,7 +14,7 @@ on:
 jobs:
   CXX:
     container:
-      image: vowpalwabbit/ubuntu1604-build:0.6.0
+      image: vowpalwabbit/ubuntu1604-build:latest
```
jackgerrits, line 17: Prior to merging these will need to be moved to a tagged version

.scripts/linux/install-flatbuffers.sh
```diff
@@ -0,0 +1,14 @@
...
+git clone https://github.com/google/flatbuffers.git flatbuffers
+cd ./flatbuffers
+# 1.10.0 release commit
+# git checkout 925c1d77fcc72636924c3c13428a34180c30f96f
+mkdir build
+cd build
+cmake .. -DFLATBUFFERS_BUILD_TESTS=Off -DFLATBUFFERS_INSTALL=On -DCMAKE_BUILD_TYPE=Release -DFLATBUFFERS_BUILD_FLATHASH=Off
+make -j `nproc`
+echo ~
+make DESTDIR=~ install
```
jackgerrits, line 12: How does this differ compared to using `-DCMAKE_INSTALL_PREFIX=/path/to/install` in the cmake command?

cmake/BuildFlatBuffers.cmake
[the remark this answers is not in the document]
jackgerrits: > Any GNU license, version N, subsumes the Apache 2.0 license provided N is at least 3.
> 
> The GNU GPL, version N, subsumes all versions of the Mozilla Public License that are compatible with it.
> 
> The Apache 2.0 license subsumes the BSD, Expat, X11, ISC and CC-0 licenses. BSD 3 clause subsumes BSD 2 clause. The BSD licenses subsume the Expat, X11 and ISC licenses and CC-0.
> 
> This is not meant to be a complete list, but if we are informed of other cases worth mentioning, we will add them.
> 
> When some license is subsumed, you still need to include a copy of it with all distribution of the combined program.

Does this notice in the file constitute including a copy of it? Given that Apache 2.0 subsumes BSD-3 clause does that not mean a kind of license change.

@lokitoth can you comment?

jackgerrits:
I am not sure but I wanted to bring it up - is committing files > 1mb a good idea?

.github/workflows/build_docs.yml
olgavrou: We were actually using different versions in different builds (i.e. some images were `0.6.0`, others `0.5.0` etc). I think it's nice to pick up latest in our CI builds just so we stay up to date without having to update the builds/forget to update some of them, and pickup braking changes early. We could still create another tag that includes the flatbuffers change.

jackgerrits: The problem is when a new build accidentally breaks builds - locking on a tagged build means we selectively move forward when ready. But you're right, on the other hand being latest is nice and knowing about the failures sooner is helpful too.

.scripts/linux/install-flatbuffers.sh
olgavrou: https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html

`DESTDIR` relocates the whole installation vs pre-pending the path to onto all install directories.

What this actually looks like is:

`CMAKE_INSTALL_PREFIX=~` will install flatbuffers under `/home/<username>/include` and `/home/<username>/bin`
whereas `DESTIR=~` will install flatbuffers under `/home/<username>/usr/local/include` and `/home/<username>/usr/local/bin`

jackgerrits: Thanks for the explanation, I didn't know that!

olgavrou:
> I am not sure but I wanted to bring it up - is committing files > 1mb a good idea?

@jackgerrits I agree and we can make these files smaller since they are test files. I don't think it hurts (we already have files over 1mb e.g. `rcv1_multiclass.dat` is  1.4M) but still, nice to change

jackgerrits:
Yeah, we shouldn't block on this but I think it's something we should consider addressing. A fresh clone of the repo is currently ~170MiB, and the test dir is ~17MiB. The .git folder is ~128MiB. So there's some decently large files spread through the git history.

One approach I've seen used [elsewhere](https://github.com/azure/mmlspark) is to place datasets into an archive that is uploaded to blob storage. The repo can then download and uncompress the artifact if they are needed for testing, with the added benefit that if you don't need to test the code you can avoid that cose.

jackgerrits:
I've created an issue to track #2566

olgavrou:
> I've created an issue to track #2566

Cool thanks! In the mean time I reduced all test files in this PR to < 1MB

explore/future_compat.h
```diff
@@ -54,20 +54,23 @@
...
     // This warning was added in GCC 8
-    #if __GNUC__ >= 8
-        #define VW_WARNING_DISABLE_CLASS_MEMACCESS    VW_WARNING_DISABLE(-Wclass-memaccess)
-        #define VW_WARNING_DISABLE_CAST_FUNC_TYPE    VW_WARNING_DISABLE(-Wcast-function-type)
-    #else
-        #define VW_WARNING_DISABLE_CLASS_MEMACCESS
-        #define VW_WARNING_DISABLE_CAST_FUNC_TYPE
-    #endif
+#  if __GNUC__ >= 8
+#    define VW_WARNING_DISABLE_CLASS_MEMACCESS VW_WARNING_DISABLE(-Wclass - memaccess)
```
jackgerrits, line 66: Does this still work?

utl/flatbuffer/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_executable(to_flatbuff txt_to_flat.cc vw_to_flat.cc vw_to_flat.h)
+target_link_libraries(to_flatbuff vw explore FlatbuffersTarget)
```
jackgerrits, line 2: CMake vs vcxproj... Wow

test/flatbuffer/create_bash_script.py
```diff
@@ -0,0 +1,122 @@
+import os
+import re
+
+with open('../RunTests', 'r') as file:
```
jackgerrits, line 4: This should be relative to the script dir and not cwd

test/unit_test/flatbuffer_parser_test.cc
```diff
@@ -0,0 +1,106 @@
...
+  all->flat_converter = VW::make_unique<VW::parsers::flatbuffer::parser>(buf);
+
+  BOOST_CHECK_EQUAL(all->flat_converter->data()->examples()->Length(), 1);
+  BOOST_CHECK_EQUAL(all->flat_converter->data()->examples()->Get(0)->namespaces()->Length(), 1);
+  BOOST_CHECK_EQUAL(all->flat_converter->data()->examples()->Get(0)->namespaces()->Get(0)->features()->size(), 1);
+  BOOST_CHECK_CLOSE(all->flat_converter->data()->examples()->Get(0)->label_as_SimpleLabel()->label(), 0.0, FLOAT_TOL);
+  BOOST_CHECK_CLOSE(all->flat_converter->data()->examples()->Get(0)->label_as_SimpleLabel()->weight(), 1.0, FLOAT_TOL);
+  BOOST_CHECK_EQUAL(all->flat_converter->data()->examples()->Get(0)->namespaces()->Get(0)->hash(), constant_namespace);
+  BOOST_CHECK_EQUAL(
+      all->flat_converter->data()->examples()->Get(0)->namespaces()->Get(0)->features()->Get(0)->name()->c_str(),
```
jackgerrits, line 75: Woah...

utl/flatbuffer/vw_to_flat.cc
```diff
@@ -0,0 +1,294 @@
...
+  label_type = VW::parsers::flatbuffer::Label_CBLabel;
+}
+
+void to_flat::create_ccb_label(example* v, flatbuffers::Offset<void>& label, VW::parsers::flatbuffer::Label& label_type)
+{
+  auto weight = v->l.conditional_contextual_bandit.weight;
+  auto e_type = v->l.conditional_contextual_bandit.type;
+  label_type = VW::parsers::flatbuffer::Label_CCBLabel;
+
+  if (e_type == 1)
```
jackgerrits, line 71: What is 1? Can this be a constant?

test/unit_test/flatbuffer_parser_test.cc
olgavrou: welcome to nested flatbuffers :P

vowpalwabbit/CMakeLists.txt
```diff
@@ -21,6 +21,11 @@ add_library(vw_io STATIC io/io_adapter.h io/io_adapter.cc)
 target_link_libraries(vw_io PRIVATE ZLIB::ZLIB)
 add_library(VowpalWabbit::io ALIAS vw_io)
 
+add_subdirectory(parser/flatbuffer)
+add_library(FlatbuffersTarget INTERFACE)
```
jackgerrits, line 25: Does this need to exported? Does it affect the install target? (does `cmake -DVW_INSTALL=On` still work?)

vowpalwabbit/parser/flatbuffer/parse_example_flatbuffer.cc
```diff
@@ -0,0 +1,177 @@
...
+}
+
+parser::parser(uint8_t* buffer_pointer)
+    : _filename("empty"), _flatbuffer_pointer(buffer_pointer), _example_index(0), _c_hash(0)
+{
+  _data = VW::parsers::flatbuffer::GetExampleCollection(_flatbuffer_pointer);  // Shift as a separate call
+}
+
+void parser::init()
+{
```
jackgerrits, line 38: Why make init separate? Not it is possible to call parse_examples without it being in a valid state?

utl/flatbuffer/vw_to_flat.cc
```diff
@@ -0,0 +1,294 @@
...
+#include "vw_to_flat.h"
+#include "options.h"
+#include "parse_args.h"
+#include "parse_regressor.h"
+#include "accumulate.h"
+#include "best_constant.h"
+#include "vw_exception.h"
+#include "options_boost_po.h"
+
+std::string to_flat::get_label_string(label_type_t label_type)
```
jackgerrits, line 18: This should be a utility exposed in vw itself, no reason to keep it in here. It may infact already exist?

vowpalwabbit/parser/flatbuffer/parse_example_flatbuffer.cc
```diff
@@ -0,0 +1,177 @@
...
+  std::ifstream infile;
+  infile.open(_filename, std::ios::binary | std::ios::in);
```
jackgerrits, line 40: Shouldn't this make use of io_buf?

vowpalwabbit/parser/flatbuffer/parse_example_flatbuffer.cc
```diff
@@ -0,0 +1,177 @@
...
+void parser::parse_flat_label(shared_data* sd, example* ae, const Example* eg)
+{
+  Label label_type = eg->label_type();
+
+  switch (label_type)
+  {
+    case Label_SimpleLabel:
+    {
+      const SimpleLabel* simple_lbl = static_cast<const SimpleLabel*>(eg->label());
+      parse_simple_label(sd, &(ae->l), simple_lbl);
```
jackgerrits, line 121: This breaks the label parser modularity, I think it would be worth looking at if these label specific functions should exist alongside the rest of the label functions?

vowpalwabbit/parser/flatbuffer/parse_example_flatbuffer.cc
olgavrou: Yes. During the benchmarking work that was done this was refactored to use io_buf, and some more cleanups have happened so this will change quite a bit in upcoming PRs

vowpalwabbit/parser/flatbuffer/parse_example_flatbuffer.cc
olgavrou: The way we parse examples will change a bit. The work has already been done along with all the benchmarking work and might be better if it comes in the codebase in smaller PRs to make reviewing easier. What do you think?

explore/future_compat.h
olgavrou: I believe it does, preprocessor should remove whitespace

vowpalwabbit/parser/flatbuffer/parse_example_flatbuffer.cc
jackgerrits: If it's addressed in upcoming PRs I'm happy

vowpalwabbit/parser/flatbuffer/parse_example_flatbuffer.cc
olgavrou: I think the original design for this was made so that flatbuffers related code can live on its own and not pollute the rest of the code base, so that it can easily be replaced with (e.g.) protobufs or something else if need be. Is that right @rajan-chari ?

utl/flatbuffer/vw_to_flat.cc
olgavrou: From what I can see it doesn't already exist. I am tempted to remove this altogether as it is only used in the to_flatbuff utility for printing out the label type when we convert from txt to flatbuffer, rather than adding a method to every label parser.

utl/flatbuffer/vw_to_flat.cc
olgavrou: Yes, using ccb_label enum instead

vowpalwabbit/CMakeLists.txt
olgavrou: `cmake -DVW_INSTALL=On` still works

test/flatbuffer/RunTests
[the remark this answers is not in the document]
olgavrou: Tracked on issue: https://github.com/VowpalWabbit/vowpal_wabbit/issues/2576

test/flatbuffer/create_bash_script.py
olgavrou: The files under `test/flatbuffer/` are not currently used anywhere. I think they should be checked in so that they are not lost, but created a issue to track them (https://github.com/VowpalWabbit/vowpal_wabbit/issues/2576) and use them in the CI or delete them if they can't be used.

rajan-chari (approved):

jackgerrits (approved):

PR #2498, opened 2020-07-07
Resolve second batch of v_array based memset warnings

jackgerrits:
- Migrate from `v_array` to `std::vector` for non _trivially copyable_ types
- Implement simple print for `std::vector`
- Deprecate two unsafe search APIs, (there were already `std::vector` equivalents)
    ```c++
    predictor& add_allowed(v_array<std::pair<action, float> >& a);
    predictor& set_allowed(v_array<std::pair<action, float> >& a);
    ````


vowpalwabbit/search.cc
```diff
@@ -1730,11 +1725,11 @@ action search_predict(search_private& priv, example* ecs, size_t ec_cnt, ptag my
         size_t label_size = priv.is_ldf ? sizeof(CS::label) : sizeof(MC::label_t);
         void (*label_copy_fn)(void*, void*) = priv.is_ldf ? CS::cs_label.copy_label : nullptr;
 
-        ensure_size(priv.learn_ec_copy, ec_cnt);
+        priv.learn_ec_copy.resize(ec_cnt);
         for (size_t i = 0; i < ec_cnt; i++)
-          VW::copy_example_data(priv.all->audit, priv.learn_ec_copy.begin() + i, ecs + i, label_size, label_copy_fn);
+          VW::copy_example_data(priv.all->audit, priv.learn_ec_copy.data() + i, ecs + i, label_size, label_copy_fn);
```
olgavrou, line 1730: since `priv.learn_ec_copy` is now a vector we could potentially use normal vector access here instead of pointer arithmetic (i.e. `&priv.learn_ec_copy.at(i)` instead)

vowpalwabbit/search_meta.cc
```diff
@@ -164,7 +158,7 @@ void run(Search::search& sch, multi_ex& ec)
     d.total_cost = 0.;
     d.output_string = nullptr;
 
-    cdbg << "*** BRANCH " << i << " *** " << d.branches[i].first << " : " << d.branches[i].second << std::endl;
+    //cdbg << "*** BRANCH " << i << " *** " << d.branches[i].first << " : " << d.branches[i].second << std::endl;
```
olgavrou, line 161: should this line be un-commented or deleted?

PR #2500, opened 2020-07-07
add multiclass label type

etiennekintzler:
The PR added a multiclass label type (called `MultiClass`) that can be used in the class `DFtoVW` that converts a pandas dataframe into VW input format (see #2452)

### Implementation detail
- The type is called `MultiClass` but I hesitate with `MultiClassLabel` or `MulticlassLabel`. The last two have the benefit of a name that is uniform with `SimpleLabel` and are a hint to the user that it is a label and not something else. I can change if necessary.
- The class has two attributes: `name` and `weight`. Value checking if performed on both attributes (using `AttributeDescriptor` introduced in commit eae8e8e ): `name` should be an int >= 1 and `weight` should be an int/float >= 0. I called the attribute `name` as in `SimpleLabel` but thinking of it a better might be just `value` (for both class). I can change that too if necessary.
- This label is also available for the convenience method `from_colnames` using `label_type='multi_class'`.
- 5 tests for this classs have been added to `test_pyvw.py`


jackgerrits:
My vote goes to: `MulticlassLabel`

> This label is also available for the convenience method from_colnames using label_type='multi_class'
I think 'multiclass' is more consistent with documentation.

Apart from that, looks great. Thanks for your work!

etiennekintzler:
> My vote goes to: `MulticlassLabel`
> > This label is also available for the convenience method from_colnames using label_type='multi_class'
> > I think 'multiclass' is more consistent with documentation.

OK, done !

> Apart from that, looks great. Thanks for your work!

Glad to contribute !

PR #2501, opened 2020-07-09
Error if multilabel given to multiclass

### LINKED ISSUE #2264: Multiclass labels do not error when given a multilabel

olgavrou:
Resolves #2264 

- `multiclass` parser to throw error if label is malformed
- Also added check for `slot_id`


vowpalwabbit/multiclass.cc
```diff
@@ -79,16 +79,33 @@ void parse_label(parser*, shared_data* sd, void* v, std::vector<VW::string_view>
...
+        {
+          THROW("malformed example: label has trailing character(s): " << *char_after_int);
+        }
+      }
       ld->weight = float_of_string(words[1]);
       break;
     default:
-      std::cerr << "malformed example!\n";
-      std::cerr << "words.size() = " << words.size() << std::endl;
+      THROW("malformed example, words.size() = " << words.size());
```
olgavrou, line 108: Not sure if these should throw or write to `cerr`, but other label parser implementations seem to throw

test/unit_test/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ add_executable(vw-unit-test.out
   test_common.h
   tokenize_tests.cc
   chain_hashing.cc
+  multiclass_label_parser_test.cc
```
jackgerrits, line 28: Did you forget to add this file?

olgavrou: Yes :) thanks!

PR #2507, opened 2020-07-13
Resolve C4244 MSVC warnings

olgavrou:
This PR resolves the conversion warnings generated by MSVC by explicitly casting from one type to the other.

More MSVC warnings to be resolved in next PR.


vowpalwabbit/ezexample.h
```diff
@@ -119,7 +119,7 @@ class ezexample
     if (current_ns != 0)
     {
       str[0] = current_ns;
-      current_seed = VW::hash_space(*vw_ref, str);
+      current_seed = static_cast<fid>(VW::hash_space(*vw_ref, str));
```
jackgerrits, line 122: what is `fid`?

vowpalwabbit/lda_core.cc
```diff
@@ -923,7 +923,7 @@ void learn_batch(lda &l)
   float additional = (float)(l.all->length()) * l.lda_rho;
   for (size_t i = 0; i < l.all->lda; i++) l.digammas.push_back(l.digamma(l.total_lambda[i] + additional));
 
-  uint64_t last_weight_index = -1;
+  auto last_weight_index = static_cast<uint64_t>(-1);
```
jackgerrits, line 926: Does it make sense to use the max value here instead like in cs_active?

vowpalwabbit/network.cc
```diff
@@ -34,7 +34,7 @@ int open_socket(const char* host)
   hostent* he;
   if (colon != nullptr)
   {
-    port = atoi(colon + 1);
+    port = (short unsigned int)atoi(colon + 1);
```
jackgerrits, line 37: Can this be made static cast?

test/unit_test/power_test.cc
```diff
@@ -13,15 +13,16 @@ bool are_same(float a, float b) { return std::abs(a - b) < std::numeric_limits<f
...
-  BOOST_CHECK(are_same(VW::fast_pow10(-127), std::pow(10, -127)));
-  BOOST_CHECK(are_same(VW::fast_pow10(-46), std::pow(10, -46)));
-  BOOST_CHECK(are_same(VW::fast_pow10(-45), std::pow(10, -45)));
-  BOOST_CHECK(are_same(VW::fast_pow10(-44), std::pow(10, -44)));
-  BOOST_CHECK(are_same(VW::fast_pow10(-40), std::pow(10, -40)));
-  BOOST_CHECK(are_same(VW::fast_pow10(-38), std::pow(10, -38)));
-  BOOST_CHECK(are_same(VW::fast_pow10(-37), std::pow(10, -37)));
-  BOOST_CHECK(are_same(VW::fast_pow10(-10), std::pow(10, -10)));
-  BOOST_CHECK(are_same(VW::fast_pow10(-5), std::pow(10, -5)));
+  float base = 10;
```
jackgerrits, line 16: nit: const?

vowpalwabbit/lda_core.cc
olgavrou: I was unsure since the [cppreference](https://en.cppreference.com/w/cpp/types/numeric_limits/max) doesn't specify the `uint64_t` type but I think it shouldn't be a problem to use the numeric_limits

vowpalwabbit/ezexample.h
olgavrou: `fid` is a `uint32_t`. Along this file there are casts from `uint64_t` to `uint32_t` wherever a hash occurs.

PR #2511, opened 2020-07-15
resolve MSVC C4456 and C4459

olgavrou:
This PR resolves MSVC warnings related to local variables hiding other local or global variables with the same variable name.

Also resolves 3 small warnings related to parentheses, using a deprecated method, and using the wrong `int` type as index.

The CI should not display any more VW related warnings. However compiling with VS2019 I get more warnings locally so will work towards resolving those too.

### DIFF
--- a/explore/explore_internal.h
+++ b/explore/explore_internal.h
@@ -14,10 +14,10 @@
 
 namespace exploration
 {
-  const uint64_t CONSTANT_A = 0xeece66d5deece66dULL;
-  const uint64_t CONSTANT_C = 2147483647;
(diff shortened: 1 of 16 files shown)


explore/explore_internal.h
```diff
@@ -14,10 +14,10 @@
 
 namespace exploration
 {
-  const uint64_t a = 0xeece66d5deece66dULL;
-  const uint64_t c = 2147483647;
+  const uint64_t CONSTANT_A = 0xeece66d5deece66dULL;
```
jackgerrits, line 17: These could be marked constexpr

PR #2515, opened 2020-07-16
Resolving various MSVC warnings

olgavrou:
This addresses MSVC warnings for `fopen`, `strtok`, `strcpy`, `fscanf`, `fprintf`, `getenv` being unsafe, and potentially uninitialized variable warnings.

The only warnings left are `C4127: conditional expression is constant` and `C4996: 'gethostbyname': Use getaddrinfo() instead`

### DIFF
--- a/library/recommend.cc
+++ b/library/recommend.cc
@@ -141,30 +141,24 @@ int main(int argc, char* argv[])
   }
 
   FILE* fB;
-  int errnofb;
   FILE* fU;
-  int errnofu;
   FILE* fI;
-  int errnofi;
 
-  VW::file_open(&fB, blacklistfilename.c_str(), "r", &errnofb);
-  if (errnofb != 0)
+  if (VW::file_open(&fB, blacklistfilename.c_str(), "r") != 0)
   {
-    fprintf(stderr, "can't open %s: %s\n", blacklistfilename.c_str(), strerror(errno));
+    fprintf(stderr, "can't open %s: %s\n", blacklistfilename.c_str(), VW::strerror_to_string(errno).c_str());
     cerr << desc << endl;
     exit(2);
   }
-  VW::file_open(&fU, userfilename.c_str(), "r", &errnofu);
-  if (errnofu != 0)
+  if (VW::file_open(&fU, userfilename.c_str(), "r") != 0)
   {
-    fprintf(stderr, "can't open %s: %s\n", userfilename.c_str(), strerror(errno));
+    fprintf(stderr, "can't open %s: %s\n", userfilename.c_str(), VW::strerror_to_string(errno).c_str());
     cerr << desc << endl;
     exit(2);
   }
-  VW::file_open(&fI, itemfilename.c_str(), "r", &errnofi);
-  if (errnofi != 0)
+  if (VW::file_open(&fI, itemfilename.c_str(), "r") != 0)
   {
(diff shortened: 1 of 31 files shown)


library/recommend.cc
```diff
@@ -141,22 +141,28 @@ int main(int argc, char* argv[])
...
   FILE* fU;
+  int errnofu;
   FILE* fI;
+  int errnofi;
 
-  if ((fB = fopen(blacklistfilename.c_str(), "r")) == NULL)
+  VW::file_open(&fB, blacklistfilename.c_str(), "r", &errnofb);
+  if (errnofb != 0)
   {
     fprintf(stderr, "can't open %s: %s\n", blacklistfilename.c_str(), strerror(errno));
```
lalo, line 153: @jackgerrits should this be using VW::strerror_to_string?

lalo: not relevant to this pr though

olgavrou: Replaced.

vowpalwabbit/memory_tree.cc
```diff
@@ -376,6 +377,8 @@ inline int random_sample_example_pop(memory_tree& b, uint64_t& cn)
 
 // train the node with id cn, using the statistics stored in the node to
 // formulate a binary classificaiton example.
+VW_WARNING_STATE_PUSH
+VW_WARNING_DISABLE_POTENTIALLY_UNINITIALIZED
```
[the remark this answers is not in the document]
olgavrou:
```
MULTICLASS::label_t mc;
MULTILABEL::labels multilabels;
MULTILABEL::labels preds;
```

the last two are structs with a `v_array` member and are used in if/else loops which causes the compiler to complain.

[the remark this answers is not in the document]
jackgerrits: Seems fishy to me. They are relying on being initialized with zero memory for it to function. We can come back to this though, the good thing about `VW_WARNING_DISABLE_POTENTIALLY_UNINITIALIZED` being in the code is we can audit it.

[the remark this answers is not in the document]
olgavrou: discussed offline, will call `v_init` instead and remove `VW_WARNING_DISABLE_POTENTIALLY_UNINITIALIZED `

vowpalwabbit/crossplat_compat.h
```diff
@@ -4,14 +4,50 @@
...
+  // strcpy_s returns an errno_t
+  return strcpy_s(dest, dest_size, src);
+#else
+  (void)dest_size; // unused here
+  strcpy(dest, src);
+  return 0;
+#endif
+}
+
+inline int file_open(FILE **pf, const char *filename, const char *mode)
```
jackgerrits, line 39: Should this new function be used in the file io adapters?
https://github.com/VowpalWabbit/vowpal_wabbit/blob/d4476ff327dc9416915b456b78d17da032c66a44/vowpalwabbit/io/io_adapter.cc#L241

olgavrou: It looks like the io_adapter uses `_sopen_s/open` to get a file descriptor and do a lower level read/write of the underlying file vs `fopen_s/fopen` which provide for example buffering, so I am not sure they are suitable here?

jackgerrits: Got it, no prob

PR #2517, opened 2020-07-19
Add test for spoof_hex_encoded_namespaces

jackgerrits:
EDIT: Originally I thought there was a bug - but I was interpreting the code incorrectly. The purpose of this function is to be able to add a hex encoded single character. So I ended up with a little cleanup and some tests.

### DIFF
--- a/vowpalwabbit/parse_args.cc
+++ b/vowpalwabbit/parse_args.cc
@@ -580,28 +580,28 @@ std::string spoof_hex_encoded_namespaces(const std::string& arg)
     {
       std::string substr = arg.substr(pos + NUMBER_OF_HEX_CHARS, NUMBER_OF_HEX_CHARS);
       char* p;
-      auto c = std::strtoul(substr.c_str(), &p, HEX_BASE);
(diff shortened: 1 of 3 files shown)


vowpalwabbit/parse_args.cc
```diff
@@ -557,22 +557,34 @@ const char* are_features_compatible(vw& vw1, vw& vw2)
...
 std::string spoof_hex_encoded_namespaces(const std::string& arg)
 {
+  constexpr size_t NUMBER_OF_HEX_CHARS = 2;
+  // "\x" + hex chars
+  constexpr size_t LENGTH_OF_HEX_TOKEN = 2 + NUMBER_OF_HEX_CHARS;
+  constexpr size_t HEX_BASE = 16;
+
+  // Too short to be hex encoded.
+  if (arg.size() < LENGTH_OF_HEX_TOKEN)
+  {
```
olgavrou, line 571: do we want to print a warning here?

jackgerrits: This isn't an issue, most namespace inputs are passed through this function so it's entirely valid to have no hex strings in here

PR #2518, opened 2020-07-20
WIP: impl of generator of DFtoVW

lalo:


lalo:
Hey @etiennekintzler if you have some free time check out this generator driven function for converting the DF to VW format. This might be helpful depending on the scenario and to avoid having in memory a huge list of strings in case the dataframe is big. Let me know if you test it out and perceive any performance change.

etiennekintzler:
Hello @lalo,

A streaming approach is indeed necessary for big dataframe ! Another way to do is to process the dataframe by chunk, using `chunksize` parameter from the `pd.read_*` method (but it's another usecase).

I haven't test performance so far but it's going to be slower for sure that the implementation based on columns. I can say that from a previous implementation where processing row by row was slower than processing column by column (the former doesn't make use of vectorization). 

Having said that, the row by row approach is to my knowledge the only way to use generator in our case. Might be it's possible to buffer more than one line but I've never tried. 

Regarding your implementation I can do a proper review is you want. For now, my main point is the following :  the method `iterrows` might be preferabler to `itertuples`. Each element of `iterrows` is that it (a tuple that contains) a dataframe of one column where the **index corresponds to the colname**. In this case you can directly use the syntax `row[colname]` and there is no need to introduce an `_Col.get_attr` method nor handle this special case in `_get_col_or_value`. Below is a toy example of how it works:

```python
import pandas as pd
df = pd.DataFrame({"a": [1, 2], "b": [3, 4]})
gen = df.df.iterrows()
 _, row = next(gen)  # the first element is index, which is not needed..   
row["a"]   # return 1
```

I will do some tests soon to compare the performances of the streaming vs whole dataframe conversion.

lalo:
The main benefit is a lower peak of memory usage, as long as you do the learning/predicting on the loop instead of collecting them in a list. So, it wouldn't shock me if it's overall slower but you avoid being impacted by a big wait time right at the beginning (as you mentioned you saw a 10 second convert time).

Also the docs mentioned that itertuples is faster than iterrows, and it seemed that at one point it had some overhead: https://stackoverflow.com/questions/24870953/does-pandas-iterrows-have-performance-issues

Anyways, I think we should offer both apis in case the performance trade offs are justified. Let me know if you have thoughts on this tradeoff.

etiennekintzler:
> Also the docs mentioned that itertuples is faster than iterrows, and it seemed that at one point it had some overhead: https://stackoverflow.com/questions/24870953/does-pandas-iterrows-have-performance-issues
https://github.com/dimgold/Datathon-TAU/blob/master/7.6%20Pandas%20iterations%20test/Pandas_Iterations.ipynb

I ignored this fact ! I will check this out.


> The main benefit is a lower peak of memory usage, as long as you do the learning/predicting on the loop instead of collecting them in a list. 

Yep the lower peak of memory usage is clearly a benefit !

> So, it wouldn't shock me if it's overall slower but you avoid being impacted by a big wait time right at the beginning (as you mentioned you saw a 10 second convert time).

Yep you avoid being impact by the wait time in the beginning but you still pay the cost through the examples. 


----


Regarding performance I've just tried and the generator implementation is 3 times slower. Below is the script I used (in IPython): 

```python
import pandas as pd
from vowpalwabbit.pyvw import DFtoVW
import numpy as np

def make_df(N=1000):
    y = np.random.choice(a=[-1, 1], size=N)
    x = np.random.normal(size=N)
    df = pd.DataFrame(dict(x5=x, x4=x, x3=x, x2=x, x1=x, y=y))
    return df

def consume_gen(gen):
    while True:
        try:
            next(gen)
        except StopIteration:
            break

def test_perf_stream(conv):
    gen = conv.convert_next_df()
    consume_gen(gen)


def test_perf_full(conv):
    conv.convert_df()

N = 1_000_000
df = make_df(N)
conv = DFtoVW.from_colnames(x=["x{}".format(i) for i in range(1, 5)], y="y", df=df)

%time test_perf_stream(conv) # 14.8 s
%time test_perf_full(conv) # 5.1 s
```

etiennekintzler:
I've just tried `iterrows`, it's so much slower it's not even funny :

```python
%time pd.Series("{}".format(x[1].x1) for x in df.iterrows())
%time pd.Series("{}".format(x.x1) for x in df.itertuples())
```

About 0.25 s for itertuples and 9 s for iterrows.

lalo:
can you test if theres a difference with this one?

```
def alt_test_perf_stream(conv):
    for estr in conv.convert_df():
        continue
```

I feel like using exception mechanism as the control of flow should be slower.

etiennekintzler:
You miss the `next` in the method name ;-) 

```python
def test_perf_stream(conv):
    gen = conv.convert_next_df()
    consume_gen(gen)
```

Take the same time for me

etiennekintzler:
I've make another test **without using generator** barely gain 1 sec (13.8s). So the main bottleneck is that we process row by row more than using generator.

lalo:
I'll update the PR with some changes, and then we can test further. There's room for improvement if I diverge from the current implementation but it was a good first start to get the ball rolling.

etiennekintzler:
Ok great

I think we create the simplest function to _manually_ build the examples from the simulated dataframe. It will provided us with a lower bound of what speed/performance can be attained if we optimized code/reduce method dispatch.

From my remembrance even the simplest implementation (of processing row by row) is at least 2 times slower than proceeding column by column.

etiennekintzler:
> I think we could create the simplest function to manually build the examples from the simulated dataframe. It will provided us with a lower bound of what speed/performance can be attained if we optimize code/reduce method dispatch.

I tried that and actually the simplest generator (with few method dispatch and the logic hard-coded) took 5.82 s (± 12 ms per loop) for the dataset specified earlier. That's about 3 times faster than the current streaming implementation. 

Below is the implementation that I used : 

```python
class SimpleGen:
    def __init__(self, df):
        self.df = df
        self.x_cols = set(df.columns) - set("y")
    
    def get_gen(self):
        map_fields = {field:i for (i, field) in enumerate(self.df.columns)}
        index_y = map_fields["y"]
        index_cols = [map_fields[col] for col in self.x_cols]

        for row in self.df.itertuples(index=False, name=None):
            yield self.process_row(row, index_y, index_cols)
        
    def process_row(self, row, index_y, index_cols):
        y = str(row[index_y])
        features = " ".join(str(row[index]) for index in index_cols)
        return y + " | " +  features
```

lalo:
Btw, @etiennekintzler are you measuring release performance or debug?

I have some pending work that I wanted to compare also:

1. When we build DataToVW object we can build a map between colName -> "position on the string array" (or something like that since the format is relatively simple and we know at that point in time how many strings are going to be involved (DataToVW itself is the schema).
2. When parsing via streaming we use that map to iterate quickly row by row and we place every item on the correct place.
3. Some light cleanup / logic - in the case that some columns are null on that specific row
4. Convert array of strings into a single string, AKA just concat

I believe it's sort of similar to your code since the added cost seems to be process_features when used in this PR's streaming implementation and you are also removing that logic.

etiennekintzler:
> Btw, @etiennekintzler are you measuring release performance or debug?

I am not sure what it means in the context of Python. My tests where on a simple script and not on a built version of the package.

> I believe it's sort of similar to your code (that you just shared) since the added cost seems to be process_features when used in this PR's streaming implementation and you are also removing that logic.

Yes it's similar to my implementation (using map from colname to array position, at the end joining list into a single string) !

By the way, I found that using `str` to convert float to string is not the fastest. Since we know it's a float we can use the function `format` with the parameter `format_spec='f'` (see https://docs.python.org/3/library/string.html#format-specification-mini-language) which seems to be around 40% faster than converting using `str`. Although using 'f' yield float with default precision of 6; **do you know what is the maximal precision allowed for float in VW** ?

etiennekintzler:
I confirm that the bottleneck is the float to str conversion 

```

 
         11001039 function calls (11001022 primitive calls) in 9.804 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  6000000    5.072    0.000    5.072    0.000 <ipython-input-59-54098edb0523>:18(<genexpr>)
  1000001    1.507    0.000    9.314    0.000 <ipython-input-59-54098edb0523>:6(get_gen)
  2000000    1.403    0.000    6.475    0.000 {method 'join' of 'str' objects}
  1000000    1.331    0.000    7.806    0.000 <ipython-input-59-54098edb0523>:15(process_row)
  1000001    0.277    0.000    9.591    0.000 {built-in method builtins.next}
        1    0.213    0.213    9.804    9.804 <ipython-input-1-6fc163689a6c>:11(consume_gen)
        1    0.000    0.000    0.000    0.000 managers.py:238(_rebuild_blknos_and_blklocs)
      135    0.000    0.000    0.000    0.000 {built-in method builtins.isinstance}
```

where the genexpr is related to the following expression 
```python
(str(row[index]) for index in index_cols)
```
By the way using `repr` instead of `str` yield slightly faster conversion (7.7 on average to 6.77 , average on 7 runs for 1m rows ).

Also, I have seen a thread about fast float to string conversion in Python (https://discuss.python.org/t/faster-float-string-conversion-ryu/2466) and they suggest on algorithm call Ryu (https://github.com/ulfjack/ryu). However there is no Python package and I doubt it can be easily integrated in the project...

PR #2522, opened 2020-07-21
Fix incorrect type casting in weight default initialization

jackgerrits:
Previously values were being illegally cast to implement the function based weight initialization. This has been refactored to make this simpler and more generic. It uses the lambda capture to allow passing extra state.

### DIFF
--- a/vowpalwabbit/array_parameters.h
+++ b/vowpalwabbit/array_parameters.h
@@ -6,6 +6,7 @@
 
 #include <unordered_map>
 #include <cstddef>
+#include <functional>
 
 #ifndef _WIN32
 #define NOMINMAX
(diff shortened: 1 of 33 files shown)


vowpalwabbit/array_parameters.h
```diff
@@ -146,7 +142,15 @@ class sparse_parameters
     uint64_t index = i & _weight_mask;
     weight_map::const_iterator iter = _map.find(index);
     if (iter == _map.end())
-      return *default_value;
+    {
```
olgavrou, line 145: is there a reason this used to not mutate the internal map but now we want it to do so?

jackgerrits: Previously there was this stashed default_value that was always returned here - the problem with that was that it didn't take into account the fact that the default value is dependent on the index.

Issue #2497, opened 2020-07-06
First slot loss report for ccb_explore_adf


ataymano:
## Short description
Add capability to output loss from first slot only for ccb_explore_adf.

## **How this suggestion will help you/others**
Idea is to start using loss from the first slot in CFE pipeline.

## **Possible solution/implementation details**

Additional option that switching the meaning of reported average loss for ccb_explore_adf between loss from first slot only and sum 
of losses over all slots (what is implemented now)

## **Example/links if any**
Exact name of the option - TBD.
vw --ccb_explore_adf --epsilon 0.2 -d data.txt   -  average loss from first slot only is reported as Average Loss
vw --ccb_explore_adf --epsilon 0.2 -d data.txt --all_slots_loss    -  average loss from all slots is reported (current implementation)

olgavrou:
@ataymano should we close this issue now?

ataymano:
yes

PR #2525, opened 2020-07-23
"First slot"/"all slots" loss report for ccb

ataymano:


ataymano:
#2497

jackgerrits:
This changes the default, is that intended?

jackgerrits:
@ataymano Looks like RunTests file merged without a conflict - there are duplicate IDs now.

PR #2526, opened 2020-07-23
Disable flaky test

olgavrou:
Disabling `TestJsonDictClass` as it fails randomly. Issue: #2527


jackgerrits:
I think you'll also need to comment out the test itself because it is discovered and run

PR #2529, opened 2020-07-28
refactor option interface / dependency checking

lalo:
If we do some refactoring around how we interact with options, we might later be able to enhance the functionality of options to be able to programmatically build a reduction stack (as part of pyatom) and keep track of the dynamic requirements that are currently being added as command line options.

There's still some options.insert missing in this refactor, they don't exactly follow this pattern so I'm still undecided on how to simplify.

This PR is on top of this other PR https://github.com/VowpalWabbit/vowpal_wabbit/pull/2531. So 2531 has to go in before this one.

### DIFF
--- a/vowpalwabbit/cb_adf.cc
+++ b/vowpalwabbit/cb_adf.cc
@@ -499,18 +499,12 @@ base_learner* cb_adf_setup(options_i& options, vw& all)
                .help("Clipping probability in importance weight. Default: 0.f (no clipping)."))
       .add(make_option("cb_type", type_string)
                .keep()
+               .default_value(type_string)
                .help("contextual bandit method to use in {ips, dm, dr, mtr, sm}. Default: mtr"));
   options.add_and_parse(new_options);
 
   if (!cb_adf_option)
     return nullptr;
-
-  // Ensure serialization of this option in all cases.
-  if (!options.was_supplied("cb_type"))
-  {
-    options.insert("cb_type", type_string);
-    options.add_and_parse(new_options);
-  }
 
   // number of weight vectors needed
   size_t problem_multiplier = 1;  // default for IPS
@@ -547,15 +541,8 @@ base_learner* cb_adf_setup(options_i& options, vw& all)
   if ((!options.was_supplied("csoaa_ldf") && !options.was_supplied("wap_ldf")) || rank_all ||
       !options.was_supplied("csoaa_rank"))
   {
-    if (!options.was_supplied("csoaa_ldf"))
-    {
-      options.insert("csoaa_ldf", "multiline");
-    }
-
-    if (!options.was_supplied("csoaa_rank"))
-    {
-      options.insert("csoaa_rank", "");
-    }
+    options.ensure_default_dependency("csoaa_ldf", "multiline");
+    options.ensure_default_dependency("csoaa_rank");
   }
 
   if (options.was_supplied("baseline") && check_baseline_enabled)
--- a/vowpalwabbit/cb_algs.cc
+++ b/vowpalwabbit/cb_algs.cc
@@ -139,19 +139,12 @@ base_learner* cb_algs_setup(options_i& options, vw& all)
   option_group_definition new_options("Contextual Bandit Options");
   new_options
       .add(make_option("cb", data->cbcs.num_actions).keep().help("Use contextual bandit learning with <k> costs"))
-      .add(make_option("cb_type", type_string).keep().help("contextual bandit method to use in {ips,dm,dr}"))
+      .add(make_option("cb_type", type_string).keep().default_value(type_string).help("contextual bandit method to use in {ips,dm,dr} default dr"))
       .add(make_option("eval", eval).help("Evaluate a policy rather than optimizing."));
   options.add_and_parse(new_options);
 
   if (!options.was_supplied("cb"))
     return nullptr;
-
-  // Ensure serialization of this option in all cases.
-  if (!options.was_supplied("cb_type"))
-  {
-    options.insert("cb_type", type_string);
-    options.add_and_parse(new_options);
-  }
 
   cb_to_cs& c = data->cbcs;
 
@@ -176,12 +169,9 @@ base_learner* cb_algs_setup(options_i& options, vw& all)
     c.cb_type = CB_TYPE_DR;
   }
 
-  if (!options.was_supplied("csoaa"))
-  {
-    std::stringstream ss;
-    ss << data->cbcs.num_actions;
-    options.insert("csoaa", ss.str());
-  }
+  std::stringstream ss;
+  ss << data->cbcs.num_actions;
+  options.ensure_default_dependency("csoaa", ss.str());
 
   auto base = as_singleline(setup_base(options, all));
   if (eval)
--- a/vowpalwabbit/cb_explore.cc
+++ b/vowpalwabbit/cb_explore.cc
@@ -299,12 +299,9 @@ base_learner* cb_explore_setup(options_i& options, vw& all)
   data->_random_state = all.get_random_state();
   uint32_t num_actions = data->cbcs.num_actions;
 
-  if (!options.was_supplied("cb"))
-  {
-    std::stringstream ss;
-    ss << data->cbcs.num_actions;
-    options.insert("cb", ss.str());
-  }
+  std::stringstream ss;
+  ss << data->cbcs.num_actions;
+  options.ensure_default_dependency("cb", ss.str());
 
   if (data->epsilon < 0.0 || data->epsilon > 1.0)
   {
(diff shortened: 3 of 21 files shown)


vowpalwabbit/cb_algs.cc
```diff
@@ -147,9 +147,8 @@ base_learner* cb_algs_setup(options_i& options, vw& all)
     return nullptr;
 
   // Ensure serialization of this option in all cases.
-  if (!options.was_supplied("cb_type"))
+  if (options.ensure_default_dependency("cb_type", type_string))
```
olgavrou, line 150: I find the prefix `ensure` on the method name a bit semantically overloaded. Combined with the method returning `true` or `false` depending on whether the key existed or not is a bit confusing. My first thought was that if `ensure_default_dependency` returns false then we failed ensure that key is set. Maybe just `set_default_if_missing` or something along those lines?

vowpalwabbit/cb_adf.cc
```diff
@@ -506,9 +506,8 @@ base_learner* cb_adf_setup(options_i& options, vw& all)
     return nullptr;
 
   // Ensure serialization of this option in all cases.
-  if (!options.was_supplied("cb_type"))
+  if (options.ensure_default_dependency("cb_type", type_string))
```
lalo, line 509: This is weird, shouldn't cb_type be set with a default value above and we could remove this line

vowpalwabbit/cb_algs.cc
lalo: The name is clunky so I wouldn't mind changing it. The boolean was mostly tacked on to be able to keep almost the same structure on the calling side

vowpalwabbit/cb_adf.cc
```diff
@@ -506,9 +506,8 @@ base_learner* cb_adf_setup(options_i& options, vw& all)
...
   {
-    options.insert("cb_type", type_string);
     options.add_and_parse(new_options);
```
lalo, line 511: why don't we avoid parsing if we can modify everything in memory. We would just need a separate api that skips all the communication via string.

vowpalwabbit/cb_adf.cc
```diff
@@ -490,6 +490,7 @@ base_learner* cb_adf_setup(options_i& options, vw& all)
   new_options
       .add(make_option("cb_adf", cb_adf_option)
                .keep()
+               .necessary()
```
lalo, line 493: haven't implemented this, but I'm thinking in ways of elevating that if check into a something more concrete, since effectively this is how to turn on or off any reduction (which is helpful for pyatom). The 505 line change is optional, the boolean could make sense with a better name maybe?

jackgerrits: I really like this, I've been thinking we need some way to specify the main option that turns the reduction "on" in a more structured way

vowpalwabbit/cb_adf.cc
jackgerrits: Yes, technically type_string could be set and we could avoid the parse and we can then make `ensure_default_dependency` be a void func

vowpalwabbit/cb_adf.cc
jackgerrits: Yeah we could actually just set `type_string` instead of running the parse again - then we can also make `ensure_default_dependency` void

vowpalwabbit/cb_explore.cc
```diff
@@ -299,12 +299,9 @@ base_learner* cb_explore_setup(options_i& options, vw& all)
...
   uint32_t num_actions = data->cbcs.num_actions;
 
-  if (!options.was_supplied("cb"))
-  {
-    std::stringstream ss;
-    ss << data->cbcs.num_actions;
-    options.insert("cb", ss.str());
-  }
+  std::stringstream ss;
+  ss << data->cbcs.num_actions;
```
jackgerrits, line 303: maybe we can just use `std::to_string` here?

vowpalwabbit/cb_adf.cc
```diff
@@ -500,15 +501,14 @@ base_learner* cb_adf_setup(options_i& options, vw& all)
       .add(make_option("cb_type", type_string)
                .keep()
                .help("contextual bandit method to use in {ips, dm, dr, mtr, sm}. Default: mtr"));
-  options.add_and_parse(new_options);
 
-  if (!cb_adf_option)
-    return nullptr;
+  if (!options.add_and_parse(new_options))
```
lalo, line 505: rename to `add_parse_and_check_necessary`

vowpalwabbit/cb_adf.cc
lalo: opened this pr for this https://github.com/VowpalWabbit/vowpal_wabbit/pull/2531

vowpalwabbit/cb_adf.cc
```diff
@@ -499,18 +499,13 @@ base_learner* cb_adf_setup(options_i& options, vw& all)
                .help("Clipping probability in importance weight. Default: 0.f (no clipping)."))
       .add(make_option("cb_type", type_string)
                .keep()
+               .default_value(type_string)
```
lalo, line 502: @jackgerrits this would be equivalent right?

peterychang:
It looks like this PR specifically is just formalizing the way we add dependency reductions and adds a (currently no-op) new `necessary()` options modifier? How would that modifier work with `cb_explore_adf` since each of the different explore types has its own setup function?

lalo:
https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/shared_feature_merger.cc#L22

vowpalwabbit/options_boost_po.cc
```diff
@@ -150,6 +155,21 @@ bool options_boost_po::was_supplied(const std::string& key) const
...
 
+bool options_boost_po::ensure_default_dependency(const std::string& key) {
+  return this->ensure_default_dependency(key, "");
+}
+
+bool options_boost_po::ensure_default_dependency(const std::string& key, const std::string& value)
+{
+  if (!this->was_supplied(key))
+  {
+    this->insert(key, value);
```
lalo, line 166: if it's not been defined it should be a .require

vowpalwabbit/parse_args.cc
```diff
@@ -455,16 +455,19 @@ input_options parse_source(vw& all, options_i& options)
...
-  // an empty list.
-  const auto positional_tokens = options.get_positional_tokens();
-  if (positional_tokens.size() == 1)
-  {
-    all.data_filename = positional_tokens[0];
-  }
-  else if (positional_tokens.size() > 1)
-  {
-    all.trace_message << "Warning: Multiple data files passed as positional parameters, only the first one will be read and the rest will be ignored." << endl;
+  if (!options.was_supplied("data")) {
```
lalo, line 458: what happens if users does both -d and positional?

vowpalwabbit/options_boost_po.cc
```diff
@@ -130,6 +135,69 @@ void options_boost_po::add_and_parse(const option_group_definition& group)
     THROW(ex.what());
   }
 }
+  std::vector<std::string> options_boost_po::get_data_values()
```
jackgerrits, line 138: What's the thinking with the rename?

vowpalwabbit/cbify.cc
```diff
@@ -474,18 +474,16 @@ base_learner* cbifyldf_setup(options_i& options, vw& all)
...
-  if (!options.was_supplied("cb_explore_adf"))
-  {
-    options.insert("cb_explore_adf", "");
-  }
-  options.insert("cb_min_cost", std::to_string(data->loss0));
-  options.insert("cb_max_cost", std::to_string(data->loss1));
+  options.ensure_default_dependency("cb_explore_adf");
+
+  options.require("cb_min_cost", std::to_string(data->loss0));
+  options.require("cb_max_cost", std::to_string(data->loss1));
```
jackgerrits, line 480: Whats the difference between `ensure_default_dependency` and `require` here?

lalo: after doing the refactor, its becoming obvious that it was the same thing after all. I didn't do the full merging of both because I wanted to think about it for sure.

vowpalwabbit/options_boost_po.cc
lalo: remnant of me wanting to use https://www.boost.org/doc/libs/1_37_0/doc/html/boost/program_options/positional_options_desc_id2344626.html but I couldn't get it to work.

lalo: since I noticed that only data was the valid positional one, there's a way to describe that on boost options api instead of trying to get a single one on parse_args and then checking again if there's any positional and return that as "data" argument. Also it only makes sense to have one single interpretation of all positional arguments, no?

vowpalwabbit/cbify.cc
lalo: I guess it depends if we need to build custom stacks that require this functionality if not they can be merged in a single one.

PR #2530, opened 2020-07-29
Fix python tests on Windows

peterychang:
Fixes a few things.
1. deleting the vw object in python did not result in its destruction
1. `finish_example` cannot be called on an example that has not been used. Doing so on Windows will cause it to crash (this behavior appears to have existed for a while, an incorrect test simply exposed it)
1. Creating a DataFrame with int values will generate a DataFrame containing `int64`s. However, the default conversion of `int`->`pandas_dtype` on Windows is `int`->`int32`. This causes type mismatch errors when performing our type checks in DFtoVW.

I'm not completely sure of the ramifications of the vw deletion change though; on the surface it makes sense and all of the tests pass. However there may be some subtle interactions with how we implement the python wrappers or how they're being used that could cause issues.

### DIFF
--- a/python/pylibvw.cc
+++ b/python/pylibvw.cc
@@ -59,7 +59,7 @@ vw_ptr my_initialize(std::string args)
 { if (args.find_first_of("--no_stdin") == std::string::npos)
     args += " --no_stdin";
   vw*foo = VW::initialize(args);
-  return boost::shared_ptr<vw>(foo, dont_delete_me);
+  return boost::shared_ptr<vw>(foo);
 }
 
 void my_run_parser(vw_ptr all)
--- a/python/vowpalwabbit/pyvw.py
+++ b/python/vowpalwabbit/pyvw.py
@@ -1513,6 +1513,9 @@ class _Col:
         expected_type = list(self.expected_type)
         if str in expected_type:
             expected_type.append(object)
+        if int in expected_type:
+            import numpy as np
+            expected_type.append(np.int64)
 
         col_type = df[self.colname].dtype
         if not (col_type in expected_type):
--- a/python/vowpalwabbit/pyvw.py
+++ b/python/vowpalwabbit/pyvw.py
@@ -5,6 +5,7 @@ from __future__ import division
 import pylibvw
 import warnings
 import pandas as pd
+import numpy as np
 
 
 class SearchTask:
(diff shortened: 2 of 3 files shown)


python/vowpalwabbit/pyvw.py
```diff
@@ -1513,6 +1513,9 @@ def check_col_type(self, df):
         expected_type = list(self.expected_type)
         if str in expected_type:
             expected_type.append(object)
+        if int in expected_type:
+            import numpy as np
```
jackgerrits, line 1517: Why not put this include at the top of the file?

peterychang: Whoops, I added that there to test because I was too lazy to scroll back up to the top. Let me move that haha

python/pylibvw.cc
```diff
@@ -59,7 +59,7 @@ vw_ptr my_initialize(std::string args)
 { if (args.find_first_of("--no_stdin") == std::string::npos)
     args += " --no_stdin";
   vw*foo = VW::initialize(args);
-  return boost::shared_ptr<vw>(foo, dont_delete_me);
```
lalo, line 62: should we leave a comment in case things go downhill in the future?

lalo: particularly bc we're not sure of the impact and we also had some memory leak reported so there's something quirky going on with the python bindings. maybe a comment would be helpful as a breadcrumb in a future investigation

lalo: https://github.com/VowpalWabbit/vowpal_wabbit/issues/2516

peterychang: I think leaving the rationale here would be ok. The code makes sense, and if a problem comes up, the commit history will eventually lead back to this PR. Sound reasonable?

peterychang:
Commenting on the change to the vw shared_ptr deleter function:
It looks like the original no-op deleter was added as part of the initial commit for the python wrapper. It appears to have been added either as a mistake or possibly to deal with some peculiarity in the way vw cleanup was being done at the time. One piece of evidence pointing to the former case is the comment left here

```
void my_finish(vw_ptr all)
{ VW::finish(*all, false);  // don't delete all because python will do that for us!
}
```

This line incorrectly comments that python will delete the object; in actuality the object would be leaked because the default shared pointer deleter was replaced with the no-op deleter.

PR #2531, opened 2020-07-29
--dry_run & elevate "option gates" into a native feature of options

lalo:
A pattern we use a lot is checking if a flag was set to continue. This information is now signaled via  a new "necessary()" function. This will enable tasks related to pyatom.

**Interesting files to review first**:
- parse_args.cc
- options.h
- options_boost_po.*
- related unit tests

if you run this command:
`vw --cs_active 3 -d \test\train-sets\cs_test --cost_max 2 --mellowness 1.0 --simulation --adax  --dry-run`
emphasis on --dry_run, which makes vw skip running the driver. Enabled reductions is now always being printed.

output:
```
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = \test\train-sets\cs_test
num sources = 1
Enabled reductions: gd, scorer, cs_active
finished run
number of examples = 0
weighted label sum = 0.000000
average loss = n.a.
total feature number = 0
```


vowpalwabbit/options.h
```diff
@@ -158,6 +148,54 @@ struct options_i
...
+    if (op.m_necessary) { m_necessary_flags.insert(op.m_name); }
+    
+    return *this;
+  }
+
+  bool check_necessary_enabled(const options_i& options) const
+  {
+    if (m_necessary_flags.size() == 0) return false;
+
+    bool at_least_one = true;
```
lalo, line 178: rename this var

vowpalwabbit/cbify.cc
```diff
@@ -393,11 +393,13 @@ base_learner* cbify_setup(options_i& options, vw& all)
...
       .add(make_option("cbify", num_actions)
                .keep()
+               .necessary()
                .help("Convert multiclass on <k> classes into a contextual bandit problem"))
       .add(make_option("cbify_cs", use_cs).help("consume cost-sensitive classification examples instead of multiclass"))
       .add(make_option("loss0", data->loss0).default_value(0.f).help("loss for correct label"))
       .add(make_option("loss1", data->loss1).default_value(1.f).help("loss for incorrect label"));
-  options.add_and_parse(new_options);
+  //options.add_and_parse(new_options);
+  options.add_parse_and_check_necessary(new_options);
```
lalo, line 402: missing

rajan-chari:
Are all keep() options necessary() and are necessary() options keep()?

vowpalwabbit/cbify.cc
```diff
@@ -393,11 +393,13 @@ base_learner* cbify_setup(options_i& options, vw& all)
...
 
   if (!options.was_supplied("cbify"))
```
rajan-chari, line 404: delete?

vowpalwabbit/memory_tree.cc
```diff
@@ -1259,8 +1260,9 @@ base_learner* memory_tree_setup(options_i& options, vw& all)
                .default_value(0)
                .help("turn on dream operations at reward based update as well"))
       .add(make_option("online", tree->online).help("turn on dream operations at reward based update as well"));
-  options.add_and_parse(new_options);
-  if (!tree->max_nodes)
+
+  // TODO: check this one
+  if (!options.add_parse_and_check_necessary(new_options))
```
rajan-chari, line 1265: Should there be an additional check for 0?

lalo: I don't think so, but left a comment to verify that the behavior holds up.

vowpalwabbit/bfgs.cc
```diff
@@ -1090,24 +1090,18 @@ base_learner* bfgs_setup(options_i& options, vw& all)
   bool bfgs_option = false;
   option_group_definition bfgs_outer_options("LBFGS and Conjugate Gradient options");
   bfgs_outer_options.add(
-      make_option("conjugate_gradient", conjugate_gradient).keep().help("use conjugate gradient based optimization"));
+      make_option("conjugate_gradient", conjugate_gradient).keep().necessary().help("use conjugate gradient based optimization"));
```
peterychang, line 1093: this reduction is created when `conjugate_gradient || bfgs`. Only one is required, not both. Is there a way to differentiate that here? Maybe that gets handled by the outer/inner_options, it just seems a little confusing

vowpalwabbit/cb_explore_adf_bag.cc
```diff
@@ -127,14 +127,14 @@ VW::LEARNER::base_learner* setup(VW::config::options_i& options, vw& all)
   new_options
       .add(make_option("cb_explore_adf", cb_explore_adf_option)
                .keep()
+               .necessary()
```
peterychang, line 130: cb_explore_adf is a weird one, in that `cb_explore_adf` is the reduction type; the other options marked as required is the explore type, which is a sub-option, only one of which can(?) be used (I'm not sure exactly what'll happen if multiple explore types are specified.. undefined behavior probably)

vowpalwabbit/parse_args.cc
```diff
@@ -1325,7 +1325,16 @@ void parse_reductions(options_i& options, vw& all)
   all.reduction_stack.push(explore_eval_setup);
   all.reduction_stack.push(ExpReplay::expreplay_setup<'c', COST_SENSITIVE::cs_label>);
   all.reduction_stack.push(Search::setup);
-  all.reduction_stack.push(audit_regressor_setup);
+  // this string can be a constant under audit_regressor.h
+  all.reduction_stack.push([](options_i& options, vw& all) { return audit_regressor_setup("audit_regressor", options, all); });
```
lalo, line 1329: if we had an id for each setup function we could have a dictionary and then we could keep track of setup functions that actually did something and are active.

vowpalwabbit/audit_regressor.cc
```diff
@@ -241,18 +241,18 @@ void init_driver(audit_regressor_data& dat)
...
+                      .necessary()
                       .help("stores feature names and their regressor values. Same dataset must be used for both "
                             "regressor training and this mode."));
-  options.add_and_parse(new_options);
 
-  if (!options.was_supplied("audit_regressor"))
+  if (!options.add_parse_and_check_necessary(new_options))
     return nullptr;
 
   if (out_file.empty())
```
lalo, line 258: make_base can keep track if this reduction got called or not

vowpalwabbit/parse_args.cc
lalo, line 1329: this could be  stack of tuples<string, setup_function> and setup_base can keep track if a nullptr got returned or not (which mean this function did something or not)

vowpalwabbit/global_data.h
```diff
@@ -455,7 +455,8 @@ struct vw
 
   size_t length() { return ((size_t)1) << num_bits; };
 
-  std::stack<VW::LEARNER::base_learner* (*)(VW::config::options_i&, vw&)> reduction_stack;
+  std::stack<std::tuple<std::string, VW::LEARNER::base_learner* (*)(VW::config::options_i&, vw&)> > reduction_stack;
+  std::set<std::string> enabled_reductions;
```
lalo, line 459: new one

vowpalwabbit/options_boost_po.cc
```diff
@@ -131,6 +131,12 @@ void options_boost_po::add_and_parse(const option_group_definition& group)
   }
 }
 
+bool options_boost_po::add_parse_and_check_necessary(const option_group_definition& group)
```
lalo, line 134: what is the code formatting complain in this file? @jackgerrits, I checked the log and just the header is referenced

lalo:
Since I'm changing the default printing, tests will fail until I update the ref sets. Don't want to do that until we decide on the final format.

vowpalwabbit/options_boost_po.h
```diff
@@ -66,6 +67,11 @@ struct options_boost_po : public options_i
     {
       m_command_line.push_back(value);
     }
+
+    // If .insert() call gets made, it means that VW itself is adding a new option
+    // therefore we know it exists and should be marked as a defined_option.
+    // User input should not be added via insert() but rather via the constructor
+    m_defined_options.insert(key);
```
lalo, line 74: related to pyatom

test/train-sets/ref/0002.autolink.stderr
```diff
@@ -6,6 +6,7 @@ power_t = 0.5
 using no cache
 Reading datafile = train-sets/0002.dat
 num sources = 1
+Enabled reductions: gd, autolink, scorer
```
olgavrou, line 9: Printing which reductions are used is really nice ✨

vowpalwabbit/parse_args.cc
olgavrou: nit: I personally prefer structs with named member vars vs tuples, I find the `std::get<0>()` ugly

lalo:
change to --dry_run

vowpalwabbit/options.h
```diff
@@ -231,6 +232,89 @@ struct option_group_definition
   std::vector<std::shared_ptr<base_option>> m_options;
 };
 
+struct options_name_extractor : options_i
```
lalo, line 235: this should live in its own file probs

vowpalwabbit/parse_args.cc
```diff
@@ -1290,124 +1290,126 @@ VW::LEARNER::base_learner* setup_base(options_i& options, vw& all)
 
 }
 
-/*
-void register_reduction(vw& all, std::tuple<std::string, VW::LEARNER::base_learner* (*)(VW::config::options_i&, vw&)> tup)
+void register_reductions(vw& all, std::vector<VW::LEARNER::base_learner* (*)(VW::config::options_i&, vw&)>& reductions)
```
lalo, line 1293: this could be a method under vw but avoiding moving too much code for now

vowpalwabbit/parse_args.cc
```diff
@@ -1290,124 +1290,126 @@ VW::LEARNER::base_learner* setup_base(options_i& options, vw& all)
...
 {
+  std::map<VW::LEARNER::base_learner* (*)(VW::config::options_i&, vw&), std::string> allowlist = {
```
lalo, line 1295: some members on this list could be refactored to avoid them being exceptions to the general rule

vowpalwabbit/cbify.cc
lalo: done

vowpalwabbit/memory_tree.cc
lalo: done and verified

vowpalwabbit/bfgs.cc
lalo: they are two separate option_group_definition. It is differentiated in that way, by being tracked by different instantiations. The new if statement is symmetrical to the past if statement, let me know if you have a suggestion to make it less confusing? It seemed the same as the previous impl to me.

vowpalwabbit/cb_explore_adf_bag.cc
lalo: Indeed cb_explore_adf is a weird one. If a user specifies more than one who knows what will happen, unless we test for it and add the safeguards to the reductions themselves. That could be a separate task unrelated to this refactor.

vowpalwabbit/parse_args.cc
lalo: I'll probably move away from tuples and into a dictionary. I agree they are ugly in cpp.

test/unit_test/options_boost_po_test.cc
```diff
@@ -338,4 +338,111 @@ BOOST_AUTO_TEST_CASE(unregistered_options)
...
+
+  bool result;
+  BOOST_CHECK_NO_THROW(result = options->add_parse_and_check_necessary(arg_group));
+  // should be true since both necessary options are present
+  BOOST_CHECK_EQUAL(result, true);
+  BOOST_CHECK_EQUAL(int_opt, 3);
+  BOOST_CHECK_EQUAL(str_opt, "test");
+  BOOST_CHECK_EQUAL(bool_opt, false);
+  BOOST_CHECK_EQUAL(other_bool_opt, true);
+
```
olgavrou, line 437: nit: this test case could be split up into two test cases to be consistent with the rest

vowpalwabbit/parse_args.cc
```diff
@@ -1272,98 +1275,130 @@ VW::LEARNER::base_learner* setup_base(options_i& options, vw& all)
 {
   auto setup_func = all.reduction_stack.top();
   all.reduction_stack.pop();
-  auto base = setup_func(options, all);
+  auto base = std::get<1>(setup_func)(options, all);
 
-  if (base == nullptr)
-    return setup_base(options, all);
+  // this mean that setup_func did not do any setup since it didnt add itself to the chain of learners
```
olgavrou, line 1280: nit:
```suggestion
  // this means that the current (base) reduction was not enabled (didn't do any setup) since it didn't add itself to the chain of learners
```

test/unit_test/options_boost_po_test.cc
lalo: done

vowpalwabbit/parse_args.cc
lalo: I rephrased it once more, since the (base) is extra and added phrased it more explicitly in case you want to read it again

vowpalwabbit/ftrl.cc
```diff
@@ -336,8 +336,10 @@ base_learner* ftrl_setup(options_i& options, vw& all)
       .add(make_option("pistol", pistol).keep().help("PiSTOL: Parameter-free STOchastic Learning"))
       .add(make_option("ftrl_alpha", b->ftrl_alpha).help("Learning rate for FTRL optimization"))
       .add(make_option("ftrl_beta", b->ftrl_beta).help("Learning rate for FTRL optimization"));
+
   options.add_and_parse(new_options);
 
+  // TODO custom any necessary instead of all
```
olgavrou, line 342: is this TODO done/forgotten or needs to become a ticket?

vowpalwabbit/options.h
```diff
@@ -193,6 +183,122 @@ struct options_i
...
+  option_group_definition& add(typed_option<T>& op)
+  {
+    m_options.push_back(std::make_shared<typename std::decay<typed_option<T>>::type>(op));
+    if (op.m_necessary) { m_necessary_flags.insert(op.m_name); }
+
+    return *this;
+  }
+
+  // will check if ALL of 'necessary' options were suplied
+  bool check_necessary_enabled(const options_i& options) const
```
olgavrou, line 208: nit: should this be named `check_necessary_supplied`?

vowpalwabbit/options_boost_po.h
```diff
@@ -51,6 +51,7 @@ struct options_boost_po : public options_i
   options_boost_po& operator=(options_boost_po&) = delete;
 
   void add_and_parse(const option_group_definition& group) override;
+  bool add_parse_and_check_necessary(const option_group_definition& group) override;
```
olgavrou, line 54: another naming concern:

`add_parse_and_check_necessary` can do more than one thing depending on the implementation
it doesn't only add, parse, and check necessary. Could it be better represented under a more generic name, like `options.process_and_check_enabled` or `options.process_and_check_supplied`

That way when you do : `if(!options.process_and_check_enabled(new_options)) return nullptr`
 it is a bit clearer that you are returning with `nullptr` here because the reduction was not enabled. 

Also process is more generic because in the general case we are adding and parsing the options and checking whether they were all supplied, but in the name extractor case we are generating the reduction name, not really adding and parsing

vowpalwabbit/parse_args.cc
olgavrou: lgtm :)

vowpalwabbit/parse_args.cc
```diff
@@ -1272,98 +1275,132 @@ VW::LEARNER::base_learner* setup_base(options_i& options, vw& all)
...
+  reductions.push_back(ftrl_setup);
+  reductions.push_back(svrg_setup);
+  reductions.push_back(sender_setup);
+  reductions.push_back(gd_mf_setup);
+  reductions.push_back(print_setup);
+  reductions.push_back(noop_setup);
+  reductions.push_back(lda_setup);
+  reductions.push_back(bfgs_setup);
+  reductions.push_back(OjaNewton_setup);
+  // reductions.push_back(VW_CNTK::setup);
```
olgavrou, line 1333: Not related to this PR but do we know why this reduction here is commented out? Should we remove it?

lalo: that's a good question, no idea

vowpalwabbit/options_boost_po.h
lalo: since we cannot remove (for now) add_and_parse() since it's being used to add general vw options all over the place (like in parse_args.cc). I wanted to keep the symmetry with that name to signal that it's just an augmented version of that fn call that adds a step to check if necessary was enabled. Also, I wanted the name to send a clear signal that it depends on .necessary(). Given that context, what are your thoughts?

vowpalwabbit/options.h
lalo: I see both equally the same, what's making you opt for supplied? the comment above or you think it's more ad-hoc for what's going on?

vowpalwabbit/ftrl.cc
lalo: it should be removed.

vowpalwabbit/options.h
olgavrou: The comment above got me thinking. This is a super super nit but semantically we could say that the user supplied an option and we enable the reduction. I am happy with either naming

vowpalwabbit/options_boost_po.h
olgavrou: For me it's just not super clear when/why `options.add_parse_and_check_necessary(new_options))`  will return `false` or `true` (based on the name alone). We have the context now, but if we come back in a couple of months then we might need to go into the `add_parse_and_check_necessary` method to remember what it does and why/when it is returning `true` or `false`.

Maybe `add_parse_check_necessary_enabled`?

lalo: That's a reasonable name, true if everything was successful (parsing, adding, enabled) false for anything else. If there's errors on parsing or adding that would be fatal errors.

olgavrou (approved):

rajan-chari:
Most users are focused on (a) learn performance and (b) getting predictions and not the internals of VW.  From that perspective, I am wondering if we should print out enabled reductions.  I personally find this interesting because I debug VW.  However from the perspective of most users, this is noise.  So I am wondering if reductions should be printed out only in the dry run case

This would also reduce the changes to the stderr files.

Thoughts?

vowpalwabbit/parse_args.cc
```diff
@@ -1272,98 +1275,132 @@ VW::LEARNER::base_learner* setup_base(options_i& options, vw& all)
...
   else
+  {
+    all.enabled_reductions.push_back(std::get<0>(setup_func));
     return base;
+  }
+}
+
+void register_reductions(vw& all, std::vector<VW::LEARNER::base_learner* (*)(VW::config::options_i&, vw&)>& reductions)
+{
+  std::map<VW::LEARNER::base_learner* (*)(VW::config::options_i&, vw&), std::string> allowlist = {{GD::setup, "gd"},
```
rajan-chari, line 1293: A typedef for the function pointer would make this code easier to read.  Can also be reused in the method signature above.

test/unit_test/options_test.cc
```diff
@@ -83,3 +83,106 @@ BOOST_AUTO_TEST_CASE(create_argument_group) {
...
+{
+  char loc;
+  std::vector<std::string> loc2;
+  option_group_definition ag("g1");
+  ag(make_option("im_necessary", loc).keep().necessary());
+  ag(make_option("opt2", loc));
+  ag.add(make_option("opt3", loc2));
+  ag.add(make_option("opt4", loc2).keep());
+
+  auto name_extractor = options_name_extractor();
```
rajan-chari, line 97: Nit: One of the things we discussed in the past was that auto would be avoided unless the deduced type is super obvious.

test/unit_test/options_test.cc
```diff
@@ -83,3 +83,106 @@ BOOST_AUTO_TEST_CASE(create_argument_group) {
...
+  // result should always be false
+  auto result = name_extractor.add_parse_and_check_necessary(ag);
```
rajan-chari, line 99: Nit: usage of auto

test/unit_test/options_test.cc
```diff
@@ -83,3 +83,106 @@ BOOST_AUTO_TEST_CASE(create_argument_group) {
...
+  char loc;
+  std::vector<std::string> loc2;
+  option_group_definition ag("g1");
+  ag(make_option("im_necessary", loc).keep().necessary());
+  ag(make_option("opt2", loc));
+  ag.add(make_option("opt3", loc2));
+  ag.add(make_option("opt4", loc2).keep());
+
+  auto name_extractor = options_name_extractor();
+  // result should always be false
```
rajan-chari, line 98: Why should this be false even if there is a 'necessary' option added?

rajan-chari (approved):

test/unit_test/options_test.cc
lalo: options_name_extractor() returns an instance of that class, isn't this a good use of var?

test/unit_test/options_test.cc
lalo: options_name_extractor.add_parse_and_check_necessary will always return false since it's sole purpose is to extract the string name from an options_group_definition and is not to be confused with other implementations of options_i

test/unit_test/options_test.cc
lalo: I'll update this to bool to be more clear

test/unit_test/options_test.cc
lalo: it could also throw asides from returning false

vowpalwabbit/parse_args.cc
lalo: good idea

test/unit_test/options_test.cc
lalo: done

vowpalwabbit/parse_args.cc
lalo: done

test/unit_test/options_test.cc
rajan-chari: This is more of a nit.  Switched to my IDE to figure out the type.   If the goal with avoiding auto is to make types obvious, then this should be changed to the concrete type.

vowpalwabbit/parse_args.cc
rajan-chari: I believe there was desire to integrate CNTK into VW at one point before I joined.  I think this can be safely removed.

PR #2534, opened 2020-08-03
Make valgrind build fail on error

olgavrou:
This should make the valgrind build fail if valgrind reports any problems

Need to fix the valgrind reported failures before merging


test/RunTests
```diff
@@ -669,6 +669,7 @@ sub run_tests() {
                 else {
                     my $errfile = valgrind_errfile($TestNo);
                     warn "$0: test $TestNo: FAILED: valgrind errors in $errfile\n";
+                    next;
```
jackgerrits, line 672: What does the next do?

olgavrou: It continues running the other tests instead of stopping here

jackgerrits:
Can you point me to where the issue was for not reporting failure?

olgavrou:
> Can you point me to where the issue was for not reporting failure?

Sure, here is an example pipeline run where valgrind reports an error but the CI doesn't fail: https://dev.azure.com/vowpalwabbit/Vowpal%20Wabbit/_build/results?buildId=12017&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=dee4e6ec-be1b-5671-4678-ab293b4004c5

jackgerrits:
Oh interesting, so the script would fail but it wouldn't exit with a non-zero status? So the CI wouldn't see it as a failure?

.scripts/linux/build-minimal.sh
```diff
@@ -11,6 +11,6 @@ BUILD_CONFIGURATION=${1:-Release}
 
 mkdir -p build
 cd build
-cmake .. -DCMAKE_BUILD_TYPE=${BUILD_CONFIGURATION} -DWARNINGS=Off -DBUILD_JAVA=Off -DBUILD_PYTHON=Off -DBUILD_TESTS=Off
+cmake .. -DCMAKE_BUILD_TYPE=${BUILD_CONFIGURATION} -DBUILD_VALGRIND_TESTS=On -DWARNINGS=Off -DBUILD_JAVA=Off -DBUILD_PYTHON=Off -DBUILD_TESTS=Off
```
lalo, line 14: so how do DBUILD_VALGRIND_TESTS and DBUILD_TESTS relate?

olgavrou:
> Oh interesting, so the script would fail but it wouldn't exit with a non-zero status? So the CI wouldn't see it as a failure?

Actually the script _would_ exit with a non-zero status but the CI would still not fail. Just ran it as a ctest to check how the CI will respond. But actually I think that just by setting `failOnStderr` to `true` on the `build-linux-valgrind.yml` file will also work and will avoid the cmake changes.

jackgerrits:
Ah okay I see, so when the perl script uses the "warn" print then it goes to stderr?

olgavrou:
> Ah okay I see, so when the perl script uses the "warn" print then it goes to stderr?
> 
> EDIT: yep, just checked https://perldoc.perl.org/functions/warn.html

Yes and looks like `FailOnStderr` defaults to `false` so it was being ignored

https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/bash?view=azure-devops

.scripts/linux/build-minimal.sh
olgavrou: removed

.pipelines/build-linux-valgrind.yml
```diff
@@ -17,3 +17,4 @@ steps:
   displayName: Build C++ VW binary
 - bash: ./.scripts/linux/test-valgrind.sh
   displayName: Run test suite with Valgrind
+  failOnStderr: true
```
lalo, line 20: why don't we need this on the other files, like the java one? is it because this one returns a 0 error code even if it failed

olgavrou:
So looking at this more, it seems we don't even need the `failOnStderr` change, we only need the check what we do inside this if statement here: https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/test/RunTests#L658

If there is a warning the test exits with `$full_status` which (locally at least) is `25600` which in turn will get reported as `exit code 0` from `test-valgrind.sh`. Exit codes are in the range 0-255 and so `25600( = 256*100)` fits perfectly and gets reported as zero.

I would suggest to not exit from that if statement at all and continue running the other tests, letting the RunTests fail what it reaches the end by reporting the `ErrorCount`

PR #2535, opened 2020-08-03
Define cpack variables and include CPack

jackgerrits:
This is a first step towards better binary distribution for VW. Note: this doesn't address dependencies at all, which would need to happen for this to be useful. You will be able to do the following:

```sh
mkdir build
cd build
cmake ..
make all -j N
cpack -G TGZ 
# or
cpack -G DEB
```

This produces:
- A .tar.gz file which contains what would be installed if you had run `make install`.  
    - Example file: [vowpal_wabbit-8.8.1-Linux.tar.gz](https://github.com/VowpalWabbit/vowpal_wabbit/files/5016966/vowpal_wabbit-8.8.1-Linux.tar.gz)
- A .deb file which can be installed on Debian based distros, and is preferable as it can be easily uninstalled too.


jackgerrits:
For this to be useful the deb and targz should be standalone. The ones above depend on boost po:
```
linux-vdso.so.1 (0x00007ffd8673d000)
    libboost_program_options.so.1.71.0 => /lib/x86_64-linux-gnu/libboost_program_options.so.1.71.0 (0x00007f5c2db27000)
    libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f5c2db0b000)
    libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f5c2d92a000)
    libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f5c2d7db000)
    libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f5c2d7c0000)
    libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f5c2d79d000)
    libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f5c2d5a9000)
    /lib64/ld-linux-x86-64.so.2 (0x00007f5c2df8d000)
```

olgavrou (approved):

PR #2538, opened 2020-08-05
Overhaul formatting CI

jackgerrits:
- Change to older bool versions of  `AllowShortBlocksOnASingleLine` and `AllowShortIfStatementsOnASingleLine` for better compatibility
- Migrate formatting check to Github Actions
    - PR check does a format check for the PR diff
    - Non fail PR/push check does a global codebase check and reports warnings

This PR should mean that the formatting check if both visible and actually useful now.


.clang-format
```diff
@@ -2,10 +2,10 @@ BasedOnStyle: Google
 AccessModifierOffset: -2
 AlignAfterOpenBracket: DontAlign
 AlignOperands: false
-AllowShortBlocksOnASingleLine: Always
```
lalo, line 5: Alt Text

utl/clang-format
```diff
@@ -4,18 +4,31 @@
...
     if [ "$1" = check ]; then
-       diff $VW_FILE <(clang-format $VW_FILE);
-       if [ $? -ne 0 ]; then
-       	  echo "Run \"./utl/clang-format fix\""
-       fi
+        diff $VW_FILE <(clang-format $VW_FILE);
+        if [ $? -ne 0 ]; then
+            ISSUE_FOUND="true"
+            if [[ -v GH_WORKFLOW_LOGGING ]]; then
+                echo "::warning file=$VW_FILE:: Formatting issues found in $VW_FILE"
```
lalo, line 17: @jackgerrits do we have access to git here? it be interesting to filter the files by the ones only being affected by this change

lalo: might be too complicated but worthwhile to consider or discuss

jackgerrits: That's exactly what the workflow does :) The file you're looking at here is the general purpose "run clang-format on the whole codebase". The following is how to do it just for the PR diff:

```
git diff origin/master HEAD -U0 --no-color | clang-format-diff -r '^.*\.(cc|h)$' -p1 > clang_format_diff.txt
          if [ -s clang_format_diff.txt ]; then
            cat clang_format_diff.txt
            echo "::error:: Formatting issues found"
            exit 1
          else
            echo "No formatting issues found in the PR diff."
          fi 
```

lalo: Nice, got misled by scrolling down

jackgerrits: It will be misleading until we fix the warnings :)

PR #2540, opened 2020-08-06
Enable unit test run with valgrind

olgavrou:
This PR

- adds a script to run the unit tests with valgrind in the CI
- resolves various erros that valgrind was reporting when running the unit tests


test/unit_test/vwdll_test.cc
```diff
@@ -73,6 +75,14 @@ BOOST_AUTO_TEST_CASE(vw_dll_parsed_and_constructed_example_parity)
 
 BOOST_AUTO_TEST_CASE(vw_dll_parse_escaped)
 {
+  for (size_t i = 0; i < framework::master_test_suite().argc; i++)
```
jackgerrits, line 78: What's the reason we have to skip it?

olgavrou: This specific test explicitly throws and valgrind reports memory blocks lost from the `std::set<std::string> m_defined_options` set (defined in `options_boost_po.h`) which I can only account to aborting with an exception causing destructors not being called rather than an actual memory leak

jackgerrits: Seems odd that boost test doesn't let the destructors run though. Oh well

olgavrou: Discussed offline, this is actually due to the fact the the `options_i` pointer created when the `vw` object is initialized leaks since the test throws on `initialize` and `options` is not cleaned up. Will leave as-is for now and look into making the `options_i options` pointer a `unique_ptr`

PR #2541, opened 2020-08-06
Add devcontainer configuration

lokitoth:
This enables automated container setup when opening the repo in a VS Code Remote Container or VS CodeSpaces (and hopefully will just work out of the box in GitHub CodeSpaces)


lalo:
who triggers the `git clone`? asking to see if we can force the --recursive flag

lokitoth:
> who triggers the `git clone`? asking to see if we can force the --recursive flag

Unfortunately, this is built into the system, as far as I can see. With that said, our cmake script triggers recursive expansion, so as soon as you configure it, the submodules get expanded.

PR #2542, opened 2020-08-07
Make code formatting check easier to fix

jackgerrits:
- Provide guidance about how to fix
- Change to triple dot notation for diff

### DIFF
--- a/.github/workflows/format_check.yml
+++ b/.github/workflows/format_check.yml
@@ -30,10 +30,14 @@ jobs:
         if: github.event_name == 'pull_request'
         shell: bash
         run: |
-          git diff origin/master HEAD -U0 --no-color | clang-format-diff -r '^.*\.(cc|h)$' -p1 > clang_format_diff.txt
+          git diff origin/master...HEAD -U0 --no-color | clang-format-diff -r '^.*\.(cc|h)$' -p1 > clang_format_diff.txt
           if [ -s clang_format_diff.txt ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/format_check.yml
```diff
@@ -30,10 +30,14 @@ jobs:
...
         run: |
-          git diff origin/master HEAD -U0 --no-color | clang-format-diff -r '^.*\.(cc|h)$' -p1 > clang_format_diff.txt
+          git diff origin/master...HEAD -U0 --no-color | clang-format-diff -r '^.*\.(cc|h)$' -p1 > clang_format_diff.txt
           if [ -s clang_format_diff.txt ]; then
             cat clang_format_diff.txt
             echo "::error:: Formatting issues found"
+            echo "To fix:"
+            echo -e "\tEnsure there is an upstream remote that points to: https://github.com/VowpalWabbit/vowpal_wabbit.git"
+            echo -e "\tRun: \"git fetch upstream\""
+            echo -e "\tRun: \"git diff upstream/master...HEAD -U0 --no-color | clang-format-diff -r '^.*\.(cc|h)$' -p1 -i\""
```
olgavrou, line 40: Ah does this format only the changes you have made that caused the check to complain vs formatting the whole file? Will this work on a new file being reported though?

jackgerrits: Yep this only formats the diff, so it'll match what the check complains about. Works for new files too.

PR #2544, opened 2020-08-11
Fix C4296 and C2065

SchuylerGoodman:
This PR has two changes:
1. Check socket == INVALID_SOCKET instead of socket < 0, based on the pattern defined in the [winsock2 docs](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-socket)
2. Fix cout and flush to std::cout and std::flush in stagewise_poly.cc (fix for C2065, undeclared identifier cout and flush)


vowpalwabbit/allreduce_sockets.cc
```diff
@@ -84,7 +84,7 @@ socket_t AllReduceSockets::sock_connect(const uint32_t ip, const int port)
 socket_t AllReduceSockets::getsock()
 {
   socket_t sock = socket(PF_INET, SOCK_STREAM, 0);
-  if (sock < 0)
+  if (sock == INVALID_SOCKET)
```
jackgerrits, line 87: It looks as though `INVALID_SOCKET` is a (Windows return value) [https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-socket], whereas on Linux this function explicitly returns [`-1` on error](https://man7.org/linux/man-pages/man2/socket.2.html).

`INVALID_SOCKET` looks as though it is defined as:
```C
#define INVALID_SOCKET  (SOCKET)(~0)
```

Do you agree with this? Can we abstract this to deal with the platform differences?

SchuylerGoodman: Ah, thanks for the clarification. So how can I mitigate this error when I try to build on Windows?

jackgerrits: Adding a function such as `is_valid_socket` or something like that and then using the `#ifdef ...` checks for platform should work here.

PR #2547, opened 2020-08-18
Add multilabels class for DFtoVW

etiennekintzler:
The PR add multilabels support for DFtoVW class (see #2452).

### Usage
The class `MultiLabels` can take a column name or a list of column names:
```python
from vowpalwabbit.pyvw import MultiLabels, DFtoVW, Feature
import pandas as pd

df = pd.DataFrame({"y1": [1], "y2": [2], "x": [3]})
conv = DFtoVW(
        df=df, label=MultiLabels(["y1", "y2"]), features=Feature("x")
    )
conv.convert_df() # returns: ['1,2 | 3']
```

### Implementation details

- Changes to `AttributeDescriptor` were made so it can handle a list of column names. 
- Changes to the method `DFtoVW.check_instance_columns` were made so it can handle a list of column names. An error will be raised at the first missing column and that column only will appear in the error even though other columns are also missing. Improvement could probably be made here.
- The class main attribute is `name`. Value checking if performed to ensure that `name` is an int >= 1. 
- As with the other label classes:
    - a parameter `name_from_df=False` can be used to give the values directly (and not refers to the column names from the dataframe).
    - the convenience method `from_colnames`  has be changed a bit to handle this new class.
- 3 tests were added.


python/vowpalwabbit/pyvw.py
```diff
@@ -1638,6 +1637,47 @@ def __set__(self, instance, arg):
                 else:
                     self.generate_error(instance, valid_type, valid_value)
 
+    def validate_value(self, arg):
+        """Validate value for litterals (not for _Col).
```
olgavrou, line 1641: nit:
```suggestion
        """Validate value for literals (not for _Col).
```

python/tests/test_pyvw.py
```diff
@@ -6,6 +6,7 @@
     DFtoVW,
     SimpleLabel,
     MulticlassLabel,
+    MultiLabels,
```
lalo, line 9: rename to MultiLabel, basically drop the s everywhere please 👍

python/vowpalwabbit/pyvw.py
```diff
@@ -2117,28 +2215,28 @@ def from_colnames(cls, y, x, df, label_type="simple_label"):
...
+
+        if isinstance(y, list) and label_type != "multilabels":
             if len(y) == 1:
                 y = y[0]
             else:
                 raise ValueError(
                     "Argument 'y' should a list of one string (or a string)."
                 )
-        if not isinstance(y, str):
+        if not isinstance(y, str) and label_type != "multilabels":
```
lalo, line 2235: on a first glance, this `if` 'wants' to be an else if, but I noticed that it shouldn't. Maybe add a space for readability before line 2235.

lalo: do we need to add another check to see if label_type is multilabel and the list doesn't contain strings?

etiennekintzler: You're right this part is awkward: you could provided a list of non string for multilabels and no error would be raised. 

I think the testing for string should happen right after testing for appropriate label type. What do you think ?

lalo:
BTW @etiennekintzler be sure to check the updated Python wiki page https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Python for more information on how to get the latest bits via pip.

etiennekintzler:
> Great work, thanks for addressing all the previous comments.

thanks, you're welcome !

> BTW @etiennekintzler be sure to check the updated Python wiki page https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Python for more information on how to get the latest bits via pip.

Ok great ! Usually I would just download the latest dev version and run the files directly but might be cleaner to install from the wheel.

PR #2548, opened 2020-08-20
Handle boost exception being raised from add_and_parse

jackgerrits:


vowpalwabbit/options_boost_po.cc
```diff
@@ -11,6 +11,9 @@
 #include <iterator>
 #include <utility>
 
+#include <boost/exception/exception.hpp>
```
lalo, line 14: I'm assuming none of these files are 1.69 specific?

vowpalwabbit/options_boost_po.cc
```diff
@@ -111,6 +114,13 @@ void options_boost_po::add_and_parse(const option_group_definition& group)
     po::store(parsed_options, vm);
     po::notify(vm);
   }
+// It seems as though Boost 1.69 was introduced in 1.69 and is later started to be thrown out of Boost PO.
```
olgavrou, line 117: the comment is a bit confusing

olgavrou (approved):

vowpalwabbit/options_boost_po.cc
jackgerrits: wow yeah I messed that up, thanks for pointing it out

vowpalwabbit/options_boost_po.cc
jackgerrits: No I don't believe so

PR #2549, opened 2020-08-20
Resolve more MSVC warnings

olgavrou:
This should resolve the last warnings. Checked build with Visual Studio 17 and Visual Studio 19.

### DIFF
--- a/vowpalwabbit/array_parameters.h
+++ b/vowpalwabbit/array_parameters.h
@@ -160,7 +160,8 @@ public:
     {
       for (auto& iter : _map)
       {
-        free(iter.second);
+        auto weight_ptr = iter.second;
+        free(weight_ptr);
       }
     }
     _map = input._map;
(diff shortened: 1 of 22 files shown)


vowpalwabbit/array_parameters.h
```diff
@@ -160,7 +160,8 @@ class sparse_parameters
     {
       for (auto& iter : _map)
       {
-        free(iter.second);
+        auto weight_ptr = iter.second;
```
jackgerrits, line 163: Nit: `auto*`

vowpalwabbit/array_parameters.h
```diff
@@ -206,7 +207,8 @@ class sparse_parameters
...
```
jackgerrits, line 210: Nit: `auto*`

olgavrou: is this for styling? because you don't need to declare pointers as `auto*`

jackgerrits: Just a preference thing, to me it makes it clear that this is a pointer so I don't need to worry about a potential copy here.

PR #2550, opened 2020-08-25
[WIP] Reduction stack manipulation 1

peterychang:
- Pushing and popping reductions are functional
- This change does not take changes to the all object into consideration
- Initialize function needs to be split up, this will require work on the API to hold the setup context until the stack manipulation is completed
- 2 major hacks that will need to be fixed
  1. Hard coded list of reductions that use multi examples
  1. options_boost_po command line needs to be accessible; just made public for now

### DIFF
--- a/python/vowpalwabbit/pyvw.py
+++ b/python/vowpalwabbit/pyvw.py
@@ -290,7 +290,8 @@ class vw(pylibvw.vw):
         if any(x in kw for x in ext_file_args):
             self.needs_parser = True
 
-        # TODO: clean this up
+        # TODO: Having to call complete_initialize() based on this flag
+        # is unintuitive. Clean up the workflow around this
         if not partial_initialize and self.needs_parser:
             pylibvw.vw.run_parser(self)
 
--- a/python/vowpalwabbit/pyvw.py
+++ b/python/vowpalwabbit/pyvw.py
@@ -2,11 +2,56 @@
 """Python binding for pylibvw class"""
 
 from __future__ import division
+import abc
 import pylibvw
 import warnings
 import pandas as pd
 import numpy as np
 
+class Learner:
+    def __init__(self, vwCppBridge):
+        self.vwCppBridge = vwCppBridge
+
+    def learn(self, ec):
+        self.vwCppBridge.call_base_learner(ec, True)
+
+    def predict(self, ec):
+        self.vwCppBridge.call_base_learner(ec, False)
+
+# compatible with Python 2 *and* 3
+ABC = abc.ABCMeta('ABC', (object,), {'__slots__': ()})
+
+def no_impl(method):
+  method.no_impl = True
+  return method
+
+# End user (pyvw library consumer) will have to create a class that
+# inherits from Copperhead and implements the custom _predict() and
+# _learn(). Optionally the user can also implement _finish_example()
+# if needed. See test_pyreduction.py for examples
+"""Copperhead class"""
+class Copperhead(ABC):
+    @abc.abstractmethod
+    def _predict(self, ec, learner):
+        pass
+
+    @abc.abstractmethod
+    def _learn(self, ec, learner):
+        pass
+
+    # this method should be implemented only if needed
+    @no_impl
+    def _finish_example(self, ec):
+        pass
+
+    def _learn_convenience(self, ec, vwbridge):
+        self._learn(ec, Learner(vwbridge))
+
+    def _predict_convenience(self, ec, vwbridge):
+        self._predict(ec, Learner(vwbridge))
+
+    def _is_finish_example_implemented(self):
+        return not hasattr(self._finish_example, 'no_impl')
 
 class SearchTask:
     """Search task class"""
@@ -187,7 +232,7 @@ class vw(pylibvw.vw):
     object; you're probably best off using this directly and ignoring
     the pylibvw.vw structure entirely."""
 
-    def __init__(self, arg_str=None, **kw):
+    def __init__(self, arg_str=None, partial_initialize=False, **kw):
         """Initialize the vw object.
 
         Parameters
@@ -237,14 +282,25 @@ class vw(pylibvw.vw):
         if arg_str is not None:
             l = [arg_str] + l
 
-        pylibvw.vw.__init__(self, " ".join(l))
+        pylibvw.vw.__init__(self, " ".join(l), partial_initialize)
 
         # check to see if native parser needs to run
         ext_file_args = ["d", "data", "passes"]
+        self.needs_parser = False
         if any(x in kw for x in ext_file_args):
+            self.needs_parser = True
+
+        # TODO: clean this up
+        if not partial_initialize and self.needs_parser:
             pylibvw.vw.run_parser(self)
 
         self.finished = False
+        self.custom_reduction = None
+
+    def complete_initialize(self):
+        pylibvw.vw.complete_initialize(self)
+        if(self.needs_parser):
+            pylibvw.vw.run_parser(self)
 
     def parse(self, str_ex, labelType=pylibvw.vw.lDefault):
         """Returns a collection of examples for a multiline example learner or
@@ -690,6 +746,14 @@ class vw(pylibvw.vw):
             else search_task(self, sch, num_actions, task_data)
         )
 
+    def create_and_push_custom_reduction(self, name, custom_reduction):
+        if issubclass(custom_reduction, Copperhead):
+            #self.custom_reduction = custom_reduction()
+            print("copying thing to self")
+            pylibvw.vw.create_and_push_custom_reduction(self, name, custom_reduction())
+        else:
+            raise TypeError("The python_reduction argument must be a class that inherits from Copperhead")
+    
 
 class namespace_id:
     """The namespace_id class is simply a wrapper to convert between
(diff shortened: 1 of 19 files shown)


vowpalwabbit/learner.h
```diff
@@ -82,13 +93,33 @@ struct learn_data
...
 
 struct sensitivity_data
 {
   using fn = float (*)(void* data, base_learner& base, example& ex);
   void* data;
   fn sensitivity_f;
+
+  bool operator==(const sensitivity_data& other) {
+    if (sensitivity_f != other.sensitivity_f) return false;
+    return true;
```
jackgerrits, line 118:
```suggestion
    return sensitivity_f == other.sensitivity_f;
```
For the others too. It removes the branching,

vowpalwabbit/learner.h
```diff
@@ -465,9 +515,51 @@ VW_WARNING_STATE_POP
     ret.learn_fd.multipredict_f = nullptr;
     ret.pred_type = pred_type;
     ret.is_multiline = std::is_same<multi_ex, E>::value;
+    ret._hash_index = std::type_index(typeid(T));
```
jackgerrits, line 518: So this means the hash of a learner is defined by the contained data object type?

vowpalwabbit/learner.h
```diff
@@ -465,9 +515,51 @@ VW_WARNING_STATE_POP
...
+  inline VW::LEARNER::base_learner* get_base_reduction() { return learn_fd.base; }
+
+  void apply_from(
+    const VW::LEARNER::base_learner* base,
+    const std::unordered_map<std::type_index, VW::LEARNER::base_learner*>& reduction_template_map) {
+    if (base == nullptr || get_base_reduction() == nullptr) {
+      // currently working on base type. Nothing to do
+      return;
+    }
+    auto src = *this;
```
jackgerrits, line 534: Copy intentional? If so can you add a comment

vowpalwabbit/learner.h
```diff
@@ -465,9 +515,51 @@ VW_WARNING_STATE_POP
...
+    ret._hash_index = std::type_index(typeid(T));
 
     return ret;
   }
+
+  std::type_index hash_index() { return _hash_index; }
+
+  inline VW::LEARNER::base_learner* get_base_reduction() { return learn_fd.base; }
+
+  void apply_from(
```
jackgerrits, line 527: Can you add a comment explaining what this function is doing?

vowpalwabbit/options_boost_po.h
```diff
@@ -156,7 +158,7 @@ struct options_boost_po : public options_i
  private:
   std::map<std::string, std::shared_ptr<base_option>> m_options;
 
-  std::vector<std::string> m_command_line;
+  //std::vector<std::string> m_command_line;
```
jackgerrits, line 161: Is this move required?

vowpalwabbit/reduction_stack.cc
```diff
@@ -0,0 +1,63 @@
...
+
+  
+  struct noop {};
+
+  void noop_single(noop&, VW::LEARNER::single_learner&, example&) {}
+  void noop_multi(noop&, VW::LEARNER::multi_learner&, multi_ex&) {}
+
+  template<bool is_learn>
+  void passthrough_single(noop&, VW::LEARNER::single_learner& base, example& ex){
+    if(is_learn) base.learn(ex);
```
jackgerrits, line 28: nit: braces required

vowpalwabbit/parse_args.cc
```diff
@@ -1272,82 +1274,134 @@ VW::LEARNER::base_learner* setup_base(options_i& options, vw& all)
...
+    , cb_dro_setup
+    , VW::cb_explore_adf::greedy::setup
+    , VW::cb_explore_adf::softmax::setup
+    , VW::cb_explore_adf::rnd::setup
+    , VW::cb_explore_adf::regcb::setup
+    , VW::cb_explore_adf::first::setup
+    , VW::cb_explore_adf::cover::setup
+    , VW::cb_explore_adf::bag::setup
+    , cb_adf_setup
+    , cb_adf_setup
```
jackgerrits, line 1297: Why are only a subset of the reductions here?

vowpalwabbit/learner.h
```diff
@@ -465,9 +515,51 @@ VW_WARNING_STATE_POP
...
+    if (noop_template->sensitivity_fd != src_template->sensitivity_fd) sensitivity_fd = src.sensitivity_fd;
+    if (noop_template->finish_example_fd != src_template->finish_example_fd) finish_example_fd = src.finish_example_fd;
+    if (noop_template->save_load_fd != src_template->save_load_fd) save_load_fd = src.save_load_fd;
+    if (noop_template->end_pass_fd != src_template->end_pass_fd) end_pass_fd = src.end_pass_fd;
+    if (noop_template->end_examples_fd != src_template->end_examples_fd) end_examples_fd = src.end_examples_fd;
+    if (noop_template->pred_type != src_template->pred_type) pred_type = src.pred_type;
+    weights = src.weights;
+    increment = base->increment * src.weights;
+
+    _hash_index = src._hash_index;
```
lalo, line 553: are these the things that change when some _setup functions get called?

vowpalwabbit/parse_args.cc
lalo: might be related to the function apply_from

peterychang: These are the reductions that require a multilearner, its used to determine which base reduction needs to be used (single vs multi)

vowpalwabbit/learner.h
peterychang: This function does the following:
1. copies the base learner
1. overwrites everything that is set in `setup_base`
1. overwrites anything that was modified in the setup function (on the learner only)

vowpalwabbit/learner.h
peterychang: This was mostly for expediency. I started with an enum class in the beginning, but realized it would require changing every reduction. Since no reductions currently share the same types (with the exception of the noop ones I created), this works as a temporary solution

vowpalwabbit/learner.h
peterychang: It is intentional, I need to keep a clean copy in order to re-write bits of it further down

vowpalwabbit/options_boost_po.h
peterychang: Not required, as explained in the description this is a hack. I need to get the original command line so I can copy the options object; currently you can't get any unprocessed options out to use in the serializer, so I resorted to pulling out the original command line

vowpalwabbit/learner.h
peterychang: I did this for consistency and readability. I imagine we'd switch all the equivalency operators to
`return x == other.x && y == other.y && z == other.z;` in the final implementation

vowpalwabbit/parse_args.cc
```diff
@@ -1272,82 +1274,135 @@ VW::LEARNER::base_learner* setup_base(options_i& options, vw& all)
...
+  };
+  std::deque<VW::LEARNER::base_learner* (*)(VW::config::options_i&, vw&)> tmp_reduction_stack;
+  std::swap(tmp_reduction_stack, all.reduction_stack);
+
+  auto *cast_options = dynamic_cast<options_boost_po*>(&options);
+  auto options_str = cast_options->m_command_line;
+  options_i* tmp_options = new options_boost_po(options_str);
```
olgavrou, line 1306: make `tmp_options` a `unique_ptr`?

vowpalwabbit/parse_args.cc
```diff
@@ -1272,82 +1274,135 @@ VW::LEARNER::base_learner* setup_base(options_i& options, vw& all)
...
+
+  for (auto reduction_it = tmp_reduction_stack.rbegin(); reduction_it != tmp_reduction_stack.rend() - 11; ++reduction_it) {
```
olgavrou, line 1308: 11? could we give 11 a name?

python/vowpalwabbit/pyvw.py
```diff
@@ -237,14 +282,25 @@ def format_input(key, val):
...
-        pylibvw.vw.__init__(self, " ".join(l))
+        pylibvw.vw.__init__(self, " ".join(l), partial_initialize)
 
         # check to see if native parser needs to run
         ext_file_args = ["d", "data", "passes"]
+        self.needs_parser = False
         if any(x in kw for x in ext_file_args):
+            self.needs_parser = True
+
+        # TODO: clean this up
```
rajan-chari, line 293: What is the clean up required?

vowpalwabbit/parse_args.cc
peterychang: The `-11` is only there to avoid calling the base reduction setup functions here. And the only reason we need to do that is because some of them do things that will break if the setup is called multiple times. Final implementation should not have that at all (I'll add a comment)

python/vowpalwabbit/pyvw.py
peterychang: Mostly the workflow. This is a little weird and unintuitive, where you either complete the initialization or not based on some flag that you pass in. I'll clarify this comment

PR #2551, opened 2020-08-25
[CAPI] Implement buildsystem for C bindings

jackgerrits:
First step in C API changes - adding in placeholder and skeleton build system


jackgerrits:
Yeah, exactly. The placeholders make sure the CMake targets work

jackgerrits:
The LGTM issue seems to be firing for unrelated code

PR #2553, opened 2020-08-26
[CAPI] Add exports and types headers

jackgerrits:
- Includes export macros
- Includes typedef for errorcode type
    - This should be unified with the existing single source of truth for error codes
    - I am thinking we have a single copy of the header in some common directory and then we use cmake to place that file into the build dir for both internal C++ and the public API. Then of course when it is installed this would be included.

Blocked by #2560


bindings/c/include/vw/experimental/types.h
```diff
@@ -0,0 +1,26 @@
...
+// uint32_t, uint64_t, etc
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+  typedef uint32_t VWStatus;
+  static const VWStatus VW_success = 0;
```
jackgerrits, line 15: The definitions in `error_data.h` are all lower case, so we are forced to make them lower case here. I think it might be work using upper case names though?

peterychang: are ints preferable over enums here?

jackgerrits: I don't have a strong preference either way. Keep in mind they are c enums here so the typing isn't really there

bindings/c/include/vw/experimental/exports.h
```diff
@@ -0,0 +1,35 @@
...
+  #if __GNUC__ >= 4
+    #define VW_DLL_PUBLIC __attribute__((visibility("default")))
+    #define VW_DLL_LOCAL __attribute__((visibility("hidden")))
+  #else
+    #define VW_DLL_PUBLIC
+    #define VW_DLL_LOCAL
+  #endif
+#endif
+
+#ifdef __cplusplus
```
olgavrou, line 31: apart from lines 31-35 the rest of the file is the same as `vwvis.h` right? Would it make sense to add the lines 31-35 to the `vwvis.h` file and include it where needed, or copy the file like we do `error_data.h` in order to avoid code duplication?

bindings/c/CMakeLists.txt
```diff
@@ -6,8 +6,11 @@ set(vw_c_impl_files
   src/placeholder.cc
 )
 
+# In order to reuse the error definition data in the C API we must make the error_data.h header available here.
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/../../vowpalwabbit/error_data.h include/vw/experimental/error_data.h COPYONLY)
+
 add_library(vw_c_api ${vw_c_public_headers} ${vw_c_impl_files})
-target_include_directories(vw_c_api PUBLIC include/)
+target_include_directories(vw_c_api PUBLIC include/ ${CMAKE_CURRENT_BINARY_DIR}/include/)
```
olgavrou, line 13: are the `include` and `{CMAKE_CURRENT_BINARY_DIR}/include` dirs different?

bindings/c/include/vw/experimental/exports.h
jackgerrits: Good point, will use the same process as error_data

bindings/c/CMakeLists.txt
jackgerrits: Yes, when we configure file to copy the `error_data.h` file, it goes into the binary dir and not the source dir.

PR #2558, opened 2020-08-31
Fix formatting CI - use version 9

jackgerrits:


jackgerrits:
Replaced by #2559, clang 9 is too recent to depend on.

PR #2563, opened 2020-09-01
[CAPI] Add stubs for VW workspace

jackgerrits:


bindings/c/include/vw/experimental/types.h
```diff
@@ -13,12 +15,60 @@ extern "C"
...
+  struct VWPrediction_tag;
+  typedef struct VWPrediction_tag VWPrediction;
+
+  struct VWExample_tag;
+  typedef struct VWExample_tag VWExample;
+
+  struct VWOptions_tag;
+  typedef struct VWOptions_tag VWOptions;
+
+  typedef VWStatus(VWReadFunc)(void*, char*, size_t, size_t*);
```
lalo, line 66: should we use typedefs to provide more insight on those args?

lalo, line 66: Does it make sense to use typedefs in the arguments?

jackgerrits: So that they are named? We could add names in here too

bindings/c/include/vw/experimental/workspace.h
```diff
@@ -0,0 +1,71 @@
...
+#include "vw/experimental/types.h"
+#include "vw/experimental/vwvis.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+  VW_DLL_PUBLIC VWStatus vw_create_workspace(VWOptions* options_handle, VWTraceMessageFunc* trace_listener,
+      void* trace_context, VWWorkspace** output_handle, VWErrorString* err_str_container) VW_API_NOEXCEPT;
```
jackgerrits, line 15: `VWErrorString*` do we want to turn this into a more fully featured `VWErrorInfo` object which has been discussed including things such as a call stack?

bindings/c/include/vw/experimental/workspace.h
```diff
@@ -0,0 +1,71 @@
...
+      VWWorkspace* workspace_handle, VWErrorString* err_str_container) VW_API_NOEXCEPT;
+
+  VW_DLL_PUBLIC VWStatus vw_workspace_get_prediction_type(const VWWorkspace* workspace_handle,
+      VWPredictionType* prediction_type, VWErrorString* err_str_container) VW_API_NOEXCEPT;
+  VW_DLL_PUBLIC VWStatus vw_workspace_get_label_type(
+      const VWWorkspace* workspace_handle, VWLabelType* label_type, VWErrorString* err_str_container) VW_API_NOEXCEPT;
+
+  // TODO add multipredict
+
+  // These are "Legacy" variants because predictions and labels are in the example object
```
olgavrou, line 43: are these post-fixed with `legacy` to indicate that they will be changed in the near future?

jackgerrits: Yes, the goal is to introduce variants where label and prediction exist externally to the example object and as separate parameters. I used legacy so we don't have to break the API to move to that other interface. "legacy" is a bit icky though...

olgavrou: I was also wondering about the fact that we use `workspace` everywhere here and if it is adding any value. Can there be a different `vw_<x>_learn_multiline` or something like that, if we pass in a different handle that is not `VWWorkspace`?

bindings/c/src/workspace.cc
```diff
@@ -0,0 +1,112 @@
...
+{
+  return VW_not_implemented;
+}
+
+VW_DLL_PUBLIC VWStatus vw_workspace_finish(VWWorkspace* workspace_handle, VWErrorString* err_str_container) noexcept
+{
+  return VW_not_implemented;
+}
+
+VW_DLL_PUBLIC VWStatus vw_destroy_workspace(VWWorkspace* workspace_handle, VWErrorString* err_str_container) noexcept
```
olgavrou, line 50: and `vw_workspace_destroy` again for consistency?

bindings/c/include/vw/experimental/types.h
```diff
@@ -13,12 +15,60 @@ extern "C"
...
+  static const VWPredictionType VW_PRED_ACTION_SCORES = 2;
+  static const VWPredictionType VW_PRED_ACTION_PROBS = 3;
+  static const VWPredictionType VW_PRED_ACTION_MULTICLASS = 4;
+  static const VWPredictionType VW_PRED_ACTION_MULTILABELS = 5;
+  static const VWPredictionType VW_PRED_PROB = 6;
+  static const VWPredictionType VW_PRED_MUTLICLASS_PROBS = 7;
+  static const VWPredictionType VW_PRED_DECISION_PROBS = 8;
+
+  struct VWWorkspace_tag;
+  typedef struct VWWorkspace_tag VWWorkspace;
```
olgavrou, line 46: Just to make sure, `VWWorkspace` is the equivalent of the `all` object, right?

bindings/c/src/workspace.cc
jackgerrits: I think the idea was for create and destroy to be `vw_create_X` and `vw_destroy_X`, but for all of the "member" functions to be `VW_X_VERB`. Do you think we should just opt for the consistency of `VW_ X_VERB` for everything?

bindings/c/include/vw/experimental/workspace.h
jackgerrits: I think we want the function names to be verbose. They are all operating as members of the workspace object. If a different object offered the ability to perform the learn operation it would offer a similar function

bindings/c/src/workspace.cc
olgavrou: oh I see it now. I personally like `VW_X_VERB` better but I don't mind `vw_create_X` and `vw_destroy_X`

bindings/c/include/vw/experimental/workspace.h
olgavrou: Even if `VWErrorInfo` only initially contains an error string I think it offers more flexibility as the error code design evolves

olgavrou (approved):

bindings/c/src/workspace.cc
peterychang: The `create` and `destroy` functions are a bit of a special case in my mind. Where `vw_X_VERB` is performing an operation on `X`, create and destroy are kind of performing operations on the VW environment itself (creating or destroying an object).

I also think its a little more natural to read
`vw_create_example` and `vw_example_create_feature_space`
vs
`vw_example_create` and `vw_example_feature_space_create`

PR #2564, opened 2020-09-01
[Build] Make vw_io an object library 

jackgerrits:
The core reason is that we don't want to create or install a libvw_io.a file, this CMake target is only intended to make internal lib organization easier.

### DIFF
--- a/vowpalwabbit/CMakeLists.txt
+++ b/vowpalwabbit/CMakeLists.txt
@@ -17,7 +17,16 @@ else()
   target_compile_options(allreduce PUBLIC ${linux_flags})
 endif()
 
-add_library(vw_io STATIC io/io_adapter.h io/io_adapter.cc)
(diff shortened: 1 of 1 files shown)


vowpalwabbit/CMakeLists.txt
```diff
@@ -288,7 +298,7 @@ target_link_libraries(vw
   PUBLIC
     VowpalWabbit::explore VowpalWabbit::allreduce Boost::boost
   PRIVATE
-    Boost::program_options ${CMAKE_DL_LIBS} ${LINK_THREADS} vw_io
+    Boost::program_options ${CMAKE_DL_LIBS} ${LINK_THREADS} vw_io ZLIB::ZLIB
```
olgavrou, line 301: Just reminding you that you said you probably can remove `zlib` from here

olgavrou (approved):

vowpalwabbit/CMakeLists.txt
jackgerrits: Thanks! I may have forgotten... :P

PR #2565, opened 2020-09-01
[CAPI] Implement reader and writer to be used with raw c api

jackgerrits:
Reader and writer to be used with the CAPI for things like model loading from across the API boundary


bindings/c/src/c_io_adapter.cc
```diff
@@ -0,0 +1,35 @@
...
+    : VW::io::reader(false /* is_resettable */), _context(context), _read_func(read)
+{
+}
+
+ssize_t c_reader::read(char* buffer, size_t num_bytes)
+{
+  if (_read_func == nullptr) { THROW("Read function is not set. Should not be possible.") }
+
+  size_t bytes_read = 0;
+  auto result = _read_func(_context, buffer, num_bytes, &bytes_read);
```
olgavrou, line 21: what is `context`? Should we also check if it is `nullptr`?

jackgerrits: It's an opaque pointer passed to the user provided function, it allows it to have state. It's valid for it to be a nullptr

olgavrou (approved):

PR #2571, opened 2020-09-07
Fail on warnings

olgavrou:
- Enable build fail on warnings and fix cmake-related warnings for msvc (C4127: conditional expression is constant)

- Did not force the experimental cmake msvc build to fail with warnings since that doesn't run on the CI afaik


vowpalwabbit/allreduce_sockets.cc
```diff
@@ -16,10 +16,11 @@ Alekh Agarwal and John Langford, with help Olivier Chapelle.
 #include <stdlib.h>
 #ifdef _WIN32
 #define NOMINMAX
-#include <WinSock2.h>
-#include <Windows.h>
-#include <WS2tcpip.h>
-#include <io.h>
+#  define _WINSOCK_DEPRECATED_NO_WARNINGS
```
olgavrou, line 19: `_WINSOCK_DEPRECATED_NO_WARNINGS` removed from `vcxproj` files and added here so that when building with `cmake` the warning is also silenced

jackgerrits:
I'm hesitant to make this a default. Since we don't control what people build with, this can (and will) result in "failures" because of warnings. I think we 100% should have it on in CI, and we should add a CMake option to turn on/off treat warnings as errors. So there would be two cmake options, warnings on/off and warnings_as_errors on/off

olgavrou:
> I'm hesitant to make this a default. Since we don't control what people build with, this can (and will) result in "failures" because of warnings. I think we 100% should have it on in CI, and we should add a CMake option to turn on/off treat warnings as errors (and encourage people to use it). So there would be two cmake options, warnings on/off and warnings_as_errors on/off

@jackgerrits that's a good point, I'll add an extra cmake option. Do you think that we might run into the same problem regarding the windows build? I have not enabled warnings as errors for windows cmake since it doesn't even run in the CI, but do you think that the windows CI is enough to cover our users ways of building on windows?

jackgerrits:
Yeah I see what you mean about the Windows build. There is currently both CMake and manual vcxproj's and we only run the manual vcxproj version in CI. But those are being maintained less and are deprioritized as the goal is to converge on CMake (it generates vcxproj's anyway). I think if we can get CMake based Windows CI running then it would provide better coverage.

jackgerrits (approved):

PR #2573, opened 2020-09-08
[WIP] Create common lib

jackgerrits:
Testing how warning CI reacts to deprecations


common/include/vowpalwabbit/common/rand48.h
```diff
@@ -0,0 +1,63 @@
...
+
+constexpr int BIAS = 127 << 23;
+
+namespace VW
+{
+namespace common
+{
+inline float merand48(uint64_t& initial)
+{
+  static_assert(sizeof(int) == sizeof(float), "Floats and ints are converted between, they must be the same size.");
```
peterychang, line 30: I don't see `int` playing any part in this. We're taking the lower 32 bits of a `uint64_t`, shift/masking it and casting to a float. Really, this should be checking `sizeof(uint32_t) == sizeof(float)`

jackgerrits: Yes, that's true this really should be the actual type we are manipulating `int32_t`

jackgerrits: I'll fix this in master since this is a draft

jackgerrits:
Closing as this is only feasible when we are converged on CMake. Will readdress at that time.

PR #2579, opened 2020-09-10
Update Python build script to allow for wheel generation

jackgerrits:
- Uses newer dependency resolution that allows for a more specific dependency to be found
- Version check to use old behavior before this is supported
- This CI will test the artifacts in a clean docker image

Note: requires new container before this will work.


olgavrou (approved):
So we are not supporting python2?

jackgerrits:
We no longer officially support Python2, so we aren't providing binary wheels. The build still works I believe, but we aren't actively removing it for now.

PR #2592, opened 2020-09-23
Support using "_p" in DSJSON parser.

zwd-ms:
Change the DSJSON parser to support both "p" and "_p":
- When the "_p" key is inside "c", its parsing is ignored.
- When duplicate "p" or "_p" are present, the latest one is used.

### DIFF
--- a/vowpalwabbit/parse_example_json.h
+++ b/vowpalwabbit/parse_example_json.h
@@ -852,8 +852,8 @@ class DefaultState : public BaseState<audit>
 
       if (ctx.key_length == 2 && ctx.key[1] == 'p')
       {
-        // Ignore "_p" when it is under the "c" key in decision service state.
-        if (ctx.root_state == &ctx.decision_service_state) Ignore(ctx, length);
+        // Ignore "_p" when it is inside the "c" key in decision service state.
+        if (ctx.root_state == &ctx.decision_service_state) { Ignore(ctx, length); }
 
         ctx.array_float_state.output_array = &ctx.label_object_state.probs;
         ctx.array_float_state.return_state = this;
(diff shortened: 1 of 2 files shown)


olgavrou:
Do we ever have:
```
{
 "c":
     {
           "p": [...]
     }
}
```

or 

```
{
 "c":
     {
           "p": [...],
            "_p": [...]
     }
}
```
?

zwd-ms:
I discussed this with @lokitoth. When PassthroughPDF is used and later joined by the trainer, the probability array will be copied inside "c". Currently there is not a mechanism to remove it when logging the original context.

zwd-ms:
> Do we ever have:
> 
> ```
> {
>  "c":
>      {
>            "p": [...]
>      }
> }
> ```
> 
> or
> 
> ```
> {
>  "c":
>      {
>            "p": [...],
>             "_p": [...]
>      }
> }
> ```
> 
> ?
I discussed this with @lokitoth. When PassthroughPDF is used and later joined by the trainer, the probability array will be copied inside "c". Currently there is not a mechanism to remove it when logging the original context.

vowpalwabbit/parse_example_json.h
```diff
@@ -852,6 +852,9 @@ class DefaultState : public BaseState<audit>
 
       if (ctx.key_length == 2 && ctx.key[1] == 'p')
       {
+        // Ignore "_p" when it is under the "c" key in decision service state.
+        if (ctx.root_state == &ctx.decision_service_state) Ignore(ctx, length);
```
jackgerrits, line 856: Could you add braces here?

jackgerrits (approved):

vowpalwabbit/parse_example_json.h
zwd-ms: Done.

rajan-chari (approved):

PR #2595, opened 2020-10-01
Create test for programmatic access to Search SequenceSpanTask in DotNet

lokitoth:
This will help ensure that the APIs remain sufficient for using Search programatically in the DotNet bindings.


lokitoth:
Not really: The expectation is that in C# you will be using the Object APIs, or getting the input as JSON strings, for the most part. Explicitly driving the parser on top of an input VW Text file is not a common practice, since most would be better served using the built-in driver.

cs/unittest/TestSearch.cs
```diff
@@ -0,0 +1,159 @@
...
+
+    [TestClass]
+    public class TestSearch : TestBase
+    {
+        public sealed class FileDriver : IDisposable
+        {
+            private string file;
+            private StreamReader streamReader;
+
+            public FileDriver(string file)
```
rajan-chari, line 24: Seems like a nice reusable utility class for testing that can live on it's own

Issue #2600, opened 2020-10-09
REMOVE the possibility in DFtoVW to provide constant values (i.e values that are not from the dataframe)


etiennekintzler:
Using DFtoVW types we can provide constant values (i.e values not from the dataframe).

For instance : 
```python
Feature(value=1.2, name="x", value_from_df=False)
```
would return a column of `x:1.2` when converting the dataframe.

The feature could be remove since : 
- **it doesn't seem really useful**: if the user really want to add constant value he could always create constant column in his dataframe (e.g `df['x'] = 1.2`)
- **Simplify the API**: no `name_from_df` or `value_from_df` arguments.
- **Simplify the underlying code**: the class `AttributeDescriptor` is much shorter + no need to handle the uncertainty about the argument being either a column or a constant value.

I'm working on the PR already.

lalo:
Closing since it's been addressed in PR #2601

PR #2601, opened 2020-10-09
Remove the possibility in DFtoVW to provide constant values

etiennekintzler:
This PR fix issue #2600.

### Changes

- The feature that allow the user to supply constant value instead of column names as argument in classes DFtoVW (and friends) is removed.
- In the `Feature` class, the name provided can only be a constant string now and not a column name.  Indeed, since now the user cannot choose whether the argument `name` is a constant or a column name, it's by default a constant.

### DIFF
--- a/python/vowpalwabbit/pyvw.py
+++ b/python/vowpalwabbit/pyvw.py
@@ -1433,23 +1433,6 @@ class example(pylibvw.example):
         return label_class(self)
 
 
-def _get_col_or_value(x, df):
-    """If x is of type _Col returns the column to which x refers to, else
-    returns the string value of x.
-
-    Parameters
-    ----------
-    x: object
-        The value.
-    df: pandas.DataFrame
-        The dataframe from which to select the column (if x is of type _Col).
-    """
-    if isinstance(x, _Col):
-        return x.get_col(df)
-    else:
-        return str(x)
-
-
 class _Col:
     """_Col class. It refers to a column of a dataframe."""
 
@@ -1559,10 +1542,6 @@ class AttributeDescriptor(object):
     performed (since the dataframe and the subsequent column is not known yet)
     but the type and value requirements are passed to the _Col instance for
     future usage in the DFtoVW constructor.
-
-    In the managed class, the optional presence of a column for an attribute is
-    indicated using *_from_df attribute. The *_from_df attribute should be set
-    before the actual attribute to take effect.
     """
 
     def __init__(self, attribute_name, expected_type, min_value=None):
@@ -1598,8 +1577,8 @@ class AttributeDescriptor(object):
         ----------
         instance: object
             The managed instance.
-        arg: str/int/float
-            The argument to set (if the checks are passed).
+        arg: str
+            The argument to set.
         """
         # initialize empty set that register the column names (for _Col arg)
         if "columns" not in instance.__dict__:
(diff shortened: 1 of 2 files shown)


lalo (approved):

python/vowpalwabbit/pyvw.py
```diff
@@ -1458,7 +1441,7 @@ def __init__(self, colname, expected_type, min_value=None):
 
         Parameters
         ----------
-        colname: str/int/float
+        colname: any hashable type (str/int/float/tuple/etc.)
```
etiennekintzler, line 1444: not actually link to the current PR. It's just that pandas accept all these types as colname (same as allowed keys for dict type I think).

lalo: do you think it's worthwhile adding that as a comment in the code itself to avoid confusion with vw format, etc

etiennekintzler: I think I could specify the type directly as  `colname: any valid pandas's column name (str/int/float/etc.)`

python/vowpalwabbit/pyvw.py
```diff
@@ -1898,35 +1745,23 @@ def process(self, df):
...
-        value_from_df: bool (default: True)
-            True if the value refers to a column in the dataframe and False if
-            the value is a constant.
-        name_from_df: bool (default: False)
-            True if the name refers to a column in the dataframe and False if
-            the name is a constant.
+        value : str
+            The column name with the value of the feature.
+        name : str
+            The column name with the name of the feature.
```
etiennekintzler, line 1759: The explanation is a bit misleading here, it's actually not a column name. It is just a simple string that is used to name the feature (with a different name than its colname). Maybe this arg should be called rename instead.

lalo: what about something like name_override? and with some explanation like "set a string name to use instead of using colname of value" (feel free to paraphrase or improve). Also, I'm assuming that value is required always to build a feature and name is optional right? maybe adding that to the description is helpful. Thoughts?

etiennekintzler: Ok I change it to `rename_feature` which is quite straightforward IMO, I change the docstring accordingly

python/vowpalwabbit/pyvw.py
```diff
@@ -1705,71 +1614,24 @@ def initialize_col_type(self, instance, colname):
...
-                    attribute_name=self.attribute_name,
-                    min_value=self.min_value,
-                )
-            )
-
 
 class SimpleLabel(object):
     """The simple label type for the constructor of DFtoVW."""
 
     name = AttributeDescriptor("name", expected_type=(int, float))
```
etiennekintzler, line 1621: semirelated to this PR but @lalo what do you think of changing the `name` arg into `label` (and same for MulticlassLabel, etc.)

lalo: I think that stands with our current docs, if you refer to https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Input-format you can see that the terminology uses 'label' even for multiclass.

python/vowpalwabbit/pyvw.py
lalo: agree with being more explicit mentioning panda's here

python/vowpalwabbit/pyvw.py
etiennekintzler: Ok i've just change it !

lalo:
Great work and congrats on your fifth pull request to VW @etiennekintzler

etiennekintzler:
Thanks ! :+1:

PR #2603, opened 2020-10-09
Ensure CCB labels are restored if an exception is throw. Add temporary logging so we can validate the fix.

kumpera:
This is a fix against master.

### DIFF
--- a/vowpalwabbit/conditional_contextual_bandit.cc
+++ b/vowpalwabbit/conditional_contextual_bandit.cc
@@ -414,7 +414,22 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
 
   // This will overwrite the labels with CB.
   create_cb_labels(data);
+  auto restore_guard = VW::scope_exit(
+    [&data, &examples]
+    {
+      delete_cb_labels(data);
 
+      // Restore ccb labels to the example objects.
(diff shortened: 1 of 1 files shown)


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -414,7 +414,22 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
...
+  auto restore_guard = VW::scope_exit(
+    [&data, &examples]
+    {
+      delete_cb_labels(data);
+
+      // Restore ccb labels to the example objects.
+      for (size_t i = 0; i < examples.size(); i++)
+      {
+        examples[i]->l.conditional_contextual_bandit = {
+            data.stored_labels[i].type, data.stored_labels[i].outcome, data.stored_labels[i].explicit_included_actions, 0.};
```
jackgerrits, line 426: Stored labels should be cleared after this is done right?

jackgerrits (approved):

rajan-chari (approved):

vowpalwabbit/conditional_contextual_bandit.cc
kumpera: Yes, this runs during frame unwind and resets the labels to a good state.

PR #2604, opened 2020-10-11
remove throw

JohnLangford:
Potential throw removal patch.   

The key issue here is numerical stability.   I've worked through and caught several places where a 0*inf creates a nan.


vowpalwabbit/gd.cc
```diff
@@ -496,7 +496,10 @@ inline void pred_per_update_feature(norm_data& nd, float x, float& fw)
       x2 = x2_min;
     }
     if (x2 > x2_max)
-      THROW("your features have too much magnitude");
+      {
+        x2 = x2_max;
```
ataymano, line 500: x2_max is FLT_MAX which is relatively easily can be converted into inf (i.e. by multiplying by 2). So it seems like now we have good chances to propagate inf into weights (i.e. line 511) which seems problematic.

ataymano: it was always the case for just a little bit smaller numbers though.

JohnLangford:
I changed the test location slightly and added an isnan(update) check.   

Incidentally, I've been exercising this code with: 
echo -e "1 |a b:1000000000\n1 |a b:1000000000\n-1 |a b:1" | ../build/vowpalwabbit/vw --interactions aaaaa --noconstant --power_t 0
echo -e "1 |a b:1000000000\n1 |a b:1000000000\n-1 |a b:1" | ../build/vowpalwabbit/vw --interactions aaaaa --noconstant

lalo:
is it worthwhile to add tests that exercise those edge cases? maybe not in this pr if it's urgent

JohnLangford:
I added a test case.

vowpalwabbit/gd.cc
```diff
@@ -645,6 +649,12 @@ float compute_update(gd& g, example& ec)
   if (sparse_l2)
     update -= g.sparse_l2 * ec.pred.scalar;
 
+  if (std::isnan(update))
+    {
+      std::cout << "update is NAN, replacing with 0" << std::endl;
```
lalo, line 654: this will ignore --quiet probably, not sure if on purpose. using all.trace_message << should respect it

lalo: but don't actually understand when we use std::cerr vs std::cout since you also use std::cerr on line 533

JohnLangford: I shifted to cerr.    

The reporting issue is nontrivial here.  When we are deep inside an optimization loop and we discover some sort of fatal flaw, how should we report?   Not by throwing because that leads to bad states.  On the comandline I can print.  Used via library?   (What if tracemessage is unavailable?)

rajan-chari:
> I shifted to cerr.    

The reporting issue is nontrivial here.  When we are deep inside an optimization loop and we discover some sort of fatal flaw, how should we report?   Not by throwing because that leads to bad states.  On the comandline I can print.  Used via library?   (What if tracemessage is unavailable?) 

Being able to return an error code would have been useful.

JohnLangford:
Yeah, an error code would be great here, but the plumbing for that is not in place and will take a more substantial amount of work.   Do you (or does someone else) want to take that on?

vowpalwabbit/gd.cc
lalo: Yeap, just pointing that out in case it was intentional to use cout, but cerr I understand.

Just a thought, but we could also augment 'example' to be able to be marked as faulty since that is the root cause of the failure. That way it could be communicated outside the hot path that the root cause of the failure comes from that specific example.

JohnLangford: Use the example to carry error info seems pretty reasonable to me.  It offers a nice synchronous path.

rajan-chari:
> Yeah, an error code would be great here, but the plumbing for that is not in place and will take a more substantial amount of work.   Do you (or does someone else) want to take that on?  
 
Examples as error channel between reductions is a good idea.  Within a reduction, we can shift to proper error codes.  Once we have most of the code migrated to using error codes, we can shift the framework to propogate error codes throughout and perhaps remove it from examples.  Lends itself to smaller PRs.

Yes.  I will add this to the backlog and schedule.  Should be  it in managable pieces.  Happy to take on stories around this once I am done with predict/learn changes.

rajan-chari (approved):

vowpalwabbit/gd.cc
```diff
@@ -495,8 +495,6 @@ inline void pred_per_update_feature(norm_data& nd, float x, float& fw)
       x = (x > 0) ? x_min : -x_min;
       x2 = x2_min;
     }
-    if (x2 > x2_max)
-      THROW("your features have too much magnitude");
```
ataymano, line 499: don't we have decent chance to turn w[adaptive] to inf on line 506 now?

JohnLangford: I suspect 'yes', but I'm not seeing how that is a problem.  Do you?

ataymano: Once it is there, it stays forever => we are not going to learn on next occurences of this feature forever. Or am I missing something?

I tried that test:

Python code to generate data:
n = 1000
passes = 10

#healthy data that mimicking y=sign(x1^4 - x2)
def write_good_data(f, n, passes):
    for i in range(n):
        for k in range(passes):
            label = -1 if i % 2 == 0 else 1
            x = i / n
            delta = (x**4) * 0.1 * label
            f.write(f'{label} |a b:{x} |c b:{x**4 + delta}'+'\n')


good_path = rf'good-{n}.txt'
bad_path = rf'bad-{n}.txt'

#bad data: one poisoned example + long tail of healthy y = sign(x1^4-x2) data
with open(bad_path, 'w') as f:
    f.write('1 |a b:10000000000')
    write_good_data(f, n, passes)

#good data: only long tail of healthy y = sign(x1^4-x2) data
with open(good_path, 'w') as f:
    write_good_data(f, n, passes)


Execution:
.\vw --interactions aaaa --noconstant -d bad-100.txt --invert_hash readable_bad.txt
.\vw --interactions aaaa --noconstant -d good-100.txt --invert_hash readable_good.txt

Result:
readable_good.txt:
a^b*a^b*a^b*a^b:68928:-1.10958
a^b:144016:-0.67623
c^b:196862:2.12328

readable_bad.txt:
:0
a^b:144016:-2.17053e-19
c^b:196862:0.322327

JohnLangford: This seems like an appropriate failure mode.

PR #2607, opened 2020-10-14
[py] move DFtoVW to its own file

lalo:
This would be a **breaking change** but we can afford it since it's fairly recent work, unless others have stronger opinions.

pyvw.py is huge and could be split further but this is an easy win. Also, later we could not depend on numpy and pandas since DFtoVW is the only piece that needs those bits for now.

Let me know what you think @etiennekintzler 

Hopefully I didn't forget to move something and CI should work and pick up the new test file.


python/vowpalwabbit/DFtoVW.py
```diff
@@ -0,0 +1,939 @@
+import numpy as np
+import pandas as pd
+
+class _Col:
```
lalo, line 4: we could reorganize this file to have the most important class on top

jackgerrits:
DFtoVW hasn't been released so it isn't a breaking change. Now is definitely the best time to do this

jackgerrits (approved):

Issue #2505, opened 2020-07-11
Better handling of string/categorical feature vs float feature in DFtoVW


etiennekintzler:
The default behavior for the type `Feature` (used in DFtoVW) is to build feature as follow  `{feature_name}:{feature_value}`. If no name is supply the feature will be build as in `{feature_value}`.

This approach has 2 problems : 
- For a _float_ feature : if **no name is provided**, the float feature will be incorrectly built as a string/categorical feature. The correct approach is to pass a empty string to the `name` parameter of `Feature`, e.g `Feature(value="colA", name='')`. It will then return `:{feature_value}`. However it doesn't seem totally obvious for newcomers.
- For a _string/categorical_ feature : if **a name is provided**, the feature will be built `{feature_name}:{feature_value}` that is it will be considered as a float. It will raises the following message in the console : `malformed example! Float expected after :`. The proper way to deal with it would be to use the `=` instead of `:` when a name is provided for a categorical/string feature.

The most natural way to handle this would be to **base the feature construction on the pandas column type** : 
- If the column is float/integer and no name is provided, then prepend `:` to the feature value. 
- If the column is a string and a name is provided, then use `=` instead of `:` between the feature name and the feature value.

Along with it, we could add a parameter `type` in `Feature` to specify the type in case the pandas column default type is not appropriate (e.g an column that is made of integers but that is related to categories).

etiennekintzler:
To @lalo and @jackgerrits what do you think this issue? Seems to pose an hindrance to the use of the DFtoVW convertor for newcomers as the output format might yield underoptimal performances (since the variables might not be of the appropriate type).

lalo:
I don't believe feature value can be anything else than a float/integer. A feature name can be a string or a number. For semi-related context: under the covers id is always an integer, generated by hashing that string name also known as the hashing trick https://booking.ai/dont-be-tricked-by-the-hashing-trick-192a6aae3087).

On the classic vw input format, if a feature name is present without a value it means value = 1, omitting a feature name is the same as adding it with value of 0. Any other numbers are allowed. This might simplify the approach you propose, right?

olgavrou:
[...]

PR #2609, opened 2020-10-14
Default feature name based on pandas colname, default VW separator based on type

etiennekintzler:
This PR fix issue #2505 

4 modifications:
- Add default feature name based on the related colname. The method `_Col.get_colname` has been created, it also slugify the colname (removing " " and ":" if present)
- Default separator is based on the pandas column type. 
- The possibility to enforce a type has also been add in `Feature` using `as_type` argument (as suggest by @olgavrou). Note that if the user select `as_type="numerical"` for a column of strings that do not represent number _underneath_ (e.g "a", "b" instead of "1", "1.5", etc.) no warning is raised. If necessary I can add warning.
- changes some tests to match the default colname behavior. Add new tests (for instance related to `as_type`).

@lalo feel free to add any change if you want to contribute on this PR :)

### DIFF
--- a/python/tests/test_DFtoVW.py
+++ b/python/tests/test_DFtoVW.py
@@ -11,12 +11,13 @@ from vowpalwabbit.DFtoVW import (
     Namespace,
 )
 
+
 def test_from_colnames_constructor():
     df = pd.DataFrame({"y": [1], "x": [2]})
     conv = DFtoVW.from_colnames(y="y", x=["x"], df=df)
     lines_list = conv.convert_df()
     first_line = lines_list[0]
-    assert first_line == "1 | 2"
+    assert first_line == "1 | x:2"
 
 
 def test_feature_column_renaming_and_tag():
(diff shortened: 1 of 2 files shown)


python/tests/test_DFtoVW.py
```diff
@@ -31,23 +32,50 @@ def test_feature_column_renaming_and_tag():
...
+def test_mixed_type_features():
+    df = pd.DataFrame({"y": [1], "x1": ["a"], "x2": [2]})
+    conv = DFtoVW(label=SimpleLabel("y"),
+                  features=[Feature(value=colname) for colname in ["x1", "x2"]],
+                  df=df)
+    first_line = conv.convert_df()[0]
+    assert first_line == "1 | x1=a x2:2"
+
+
+def test_as_type_in_features():
```
olgavrou, line 57: Can we still determine a different column to hold the feature values?
i.e. what if we want `"a"` to be categorical but also have different values other than 1, `a=2:4` for example

jackgerrits (approved):
Awesome stuff!

python/tests/test_DFtoVW.py
etiennekintzler: Unfortunately it was disabled in #2601. The title of the PR is "Remove the possibility in DFtoVW to provide constant values" but actually it's more "Allow just one type of value (const or _Col) in DFtoVW classes". What was possible before is to give in the `name` arg the column name and set arg `name_from_df=True` to use the values of the column instead of constant string name.

Could you give an example of why you would do this (setting specific feature values to a categorical column), the use case is not really clear for me ? Thanks

etiennekintzler:
Thanks a lot @jackgerrits :smiley:

lalo (approved):

python/tests/test_DFtoVW.py
olgavrou: I synced with @lalo off line, it seems that the real world use case for categorical features having values other than `1` is probably not worth supporting and if someone really wants to do that explicitly then they would probably specify `a=2` as the column name

olgavrou (approved):

PR #2610, opened 2020-10-16
Enable automatic CB and CCB equivalence for single slot examples

jackgerrits:
When using a single slot with CCB it does not make sense nor add value to add the interactions and slot features. Therefore, this stops those being added in that case. This allows for smaller models in this case and for more seamless upgrades from a CB setup to CCB (this is the primary use case). Once more than 1 slot is seen the normal behavior is used and this fact is stored in the model.

:rotating_light: WARNING: This PR MUST be merged alongside a version bump as we must be able to test whether this new information is in the model or not and version is the only mechanism at the moment. I've tentatively marked this as 8.9.0


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -616,6 +625,21 @@ void finish_multiline_example(vw& all, ccb& data, multi_ex& ec_seq)
   VW::finish_example(all, ec_seq);
 }
 
+void save_load(ccb& sm, io_buf& io, bool read, bool text)
+{
+  if (io.num_files() == 0) { return; }
+
+  // We want to enter this block if either we are writing, or reading a model file after the version in which this was
+  // added.
+  // if (!read || sm.model_file_version >= VERSION_FILE_WITH_CCB_MULTI_SLOTS_SEEN_FLAG)
```
lalo, line 634: is this commented on purpose?

jackgerrits: Not on purpose, good catch thankyou!

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -616,6 +625,21 @@ void finish_multiline_example(vw& all, ccb& data, multi_ex& ec_seq)
...
+void save_load(ccb& sm, io_buf& io, bool read, bool text)
+{
+  if (io.num_files() == 0) { return; }
+
+  // We want to enter this block if either we are writing, or reading a model file after the version in which this was
+  // added.
+  if (!read || sm.model_file_version >= VERSION_FILE_WITH_CCB_MULTI_SLOTS_SEEN_FLAG)
+  {
+    std::stringstream msg;
+    if (!read) { msg << "CCB: has_seen_multi_slot_example = " << sm.has_seen_multi_slot_example << "\n"; }
```
lalo, line 637: why do we need this exactly?

jackgerrits: The feature or saving and loading from models?

lalo: saving and loading, wouldn't version number be enough? o why do we need to create that string on 637

jackgerrits: That string is for the readable_model output

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -420,6 +420,17 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
...
   data.has_seen_multi_slot_example = data.has_seen_multi_slot_example || data.slots.size() > 1;
+
+  // If we have not seen more than one slot, we need to check if the user has supplied slot features.
+  // In that case we will turn on CCB slot id/interactions.
+  if (!data.has_seen_multi_slot_example)
+  {
+    // We decide that user defined features exist if there is at least one feature space which is not the constant
+    // namespace.
+    const bool user_defined_slot_features_exist =
+        !data.slots[0]->indices.empty() && data.slots[0]->indices[0] != constant_namespace;
```
lokitoth, line 431: Does there need to be a check for data.slots.size() > 0?

jackgerrits: Oof if slots.size() is 0 there are going to be more issues

lokitoth: Do we want to support slots.size() == 0 as syntactic sugar for 1 slot, no features? That will make CB = (CCB without Slots) more explicit.

jackgerrits: I would rather avoid that, it introduces allocation requirement and extra management. If you parse CB dsjson logs right now with CCB turned on you already get automatic upgrading/sugar. (It adds the slot)

python/tests/test_pyvw.py
```diff
@@ -353,6 +353,58 @@ def test_example_features():
     assert ex.pop_namespace()
 
 
+def count_weights_from_readable_model_file(file_name):
```
olgavrou, line 356: Can we add a test that checks the no-slot-features case too?

olgavrou: And maybe also validate the model version in the file if it's not too difficult to get

olgavrou (commented):
I really like the python test. Do you think that adding some unit tests around this would add some value?

jackgerrits:
I wasn't too sure what I could unit test here which is why I went for the more integration style Python test. Do you have any suggestions?

olgavrou:
> I wasn't too sure what I could unit test here which is why I went for the more integration style Python test. Do you have any suggestions?

I was thinking of calling `learn_or_predict` with the `ccb data` struct and then verify it, but looking at this more closely there doesn't seem to be a straightforward way to do that.

python/tests/test_pyvw.py
lalo, line 356: I also like the use of python for testing, good precedence. BTW, I recently added a new python test file called test_cb so it might be worth migrating this to over there? your call

lalo: this and the other test

jackgerrits: The current case is with no slot features, I can check the case where this does not apply when there are slot features. I'm not sure if checking the version makes sense though as that will change on each release

jackgerrits: Yeah easily done, moving to a test_ccb file

lalo (approved):

PR #2611, opened 2020-10-16
Add exit scope for predictions

olgavrou:


rajan-chari (approved):

vowpalwabbit/csoaa.cc
```diff
@@ -462,10 +462,52 @@ void do_actual_learning(ldf& data, single_learner& base, multi_ex& ec_seq_all)
 
   /////////////////////// add headers
   uint32_t K = (uint32_t)ec_seq.size();
+  uint32_t predicted_K = 0;
```
rajan-chari, line 465: It seems like K and predicted_K can be moved inside the restore_guard.

rajan-chari (approved):

vowpalwabbit/csoaa.cc
olgavrou: They are both used outside of the restore guard and `predicted_K` is also set outside of the restore guard (line 538)

vowpalwabbit/csoaa.cc
```diff
@@ -462,10 +462,52 @@ void do_actual_learning(ldf& data, single_learner& base, multi_ex& ec_seq_all)
...
+        ec_seq[0]->pred.a_s.push_back(data.a_s[k]);
+      }
+    }
+    else
+    {
+      // Mark the predicted subexample with its class_index, all other with 0
+      for (size_t k = 0; k < K; k++)
+      {
+        if (k == predicted_K)
+          ec_seq[k]->pred.multiclass = ec_seq[k]->l.cs.costs[0].class_index;
```
kumpera, line 485: Could we run into the situation that we throw with a label that is not a cost sensitive one?

olgavrou: Good point - I have added a restore guard for one of the potential paths (if we throw through `make_single_prediction` line 536)

For the other potential paths `do_actual_learning_wap` and `do_actual_learning_oaa` they already have restore guards for the label

kumpera (approved):
LGTM

PR #2616, opened 2020-10-23
Change loss function to a unique_ptr

jackgerrits:
- Removes several raw deletes in the codebase
- Additionally this change moves to scope based control in nn


vowpalwabbit/nn.cc
```diff
@@ -262,7 +260,7 @@ void predict_or_learn_multi(nn& n, single_learner& base, example& ec)
       }
     }
 
-    n.all->loss = save_loss;
+    loss_function_swap_guard_converse_block.do_swap();
```
lalo, line 263: do we need all the do_swaps technically? since they will go out of scope and trigger eventually, this might be more explicit though

lalo:
might be worthwhile to add a test that uses swap guard and unique_ptr in case impl of swap_guard changes? not sure how probable though https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/test/unit_test/guard_test.cc

vowpalwabbit/nn.cc
jackgerrits: Perhaps not, but this code is pretty complicated and makes use of a `goto` so I didn't want to mess with the logic. This should be equivalent to the original code.

jackgerrits:
Yep good thinking, can do!

lalo (approved):

olgavrou (approved):

Issue #2605, opened 2020-10-12
Segmentation fault


leoguelman:
I get a "Segmentation fault" error when running the Getting Started example below. Any hints?

```python
import pandas as pd
import sklearn as sk
import numpy as np
from vowpalwabbit import pyvw

train_data = [{'action': 1, 'cost': 2, 'probability': 0.4, 'feature1': 'a', 'feature2': 'c', 'feature3': ''},
              {'action': 3, 'cost': 0, 'probability': 0.2, 'feature1': 'b', 'feature2': 'd', 'feature3': ''},
              {'action': 4, 'cost': 1, 'probability': 0.5, 'feature1': 'a', 'feature2': 'b', 'feature3': ''},
              {'action': 2, 'cost': 1, 'probability': 0.3, 'feature1': 'a', 'feature2': 'b', 'feature3': 'c'},
              {'action': 3, 'cost': 1, 'probability': 0.7, 'feature1': 'a', 'feature2': 'd', 'feature3': ''}]

train_df = pd.DataFrame(train_data)

train_df['index'] = range(1, len(train_df) + 1)
train_df = train_df.set_index("index")


test_data = [{'feature1': 'b', 'feature2': 'c', 'feature3': ''},
            {'feature1': 'a', 'feature2': '', 'feature3': 'b'},
            {'feature1': 'b', 'feature2': 'b', 'feature3': ''},
            {'feature1': 'a', 'feature2': '', 'feature3': 'b'}]

test_df = pd.DataFrame(test_data)

# Add index to data frame
test_df['index'] = range(1, len(test_df) + 1)
test_df = test_df.set_index("index")

vw = pyvw.vw("--cb 4")

vw

for i in train_df.index:
      action = train_df.loc[i, "action"]
      cost = train_df.loc[i, "cost"]
      probability = train_df.loc[i, "probability"]
      feature1 = train_df.loc[i, "feature1"]
      feature2 = train_df.loc[i, "feature2"]
      feature3 = train_df.loc[i, "feature3"]
    
      learn_example = str(action) + ":" + str(cost) + ":" + str(probability) + " | " + str(feature1) + " " + str(feature2) + " " + str(feature3)
      vw.learn(learn_example)
```

#### Environment
MacOS Catalina 10.15.6
Python 3.8.3 
vowpalwabbit 8.8.1

lalo:
[...]

PR #2617, opened 2020-10-27
[py] add test based on example

lalo:
adding this test based on the 'getting started' example we have on python.

there's a reported seg fault based on issue #2605

### DIFF
--- a/python/tests/test_cb.py
+++ b/python/tests/test_cb.py
@@ -42,7 +42,6 @@ def test_getting_started_example():
         learn_example = str(action) + ":" + str(cost) + ":" + str(probability) + " | " + str(feature1) + " " + str(feature2) + " " + str(feature3)
         vw.learn(learn_example)
 
-    print(vw.get_prediction_type())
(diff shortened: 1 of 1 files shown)


olgavrou (approved):

lalo:
Feel free to suggest more asserts, I just added a couple of obvious ones. We should add later one that uses DFtoVW

python/tests/test_cb.py
```diff
@@ -0,0 +1,56 @@
...
+        cost = train_df.loc[i, "cost"]
+        probability = train_df.loc[i, "probability"]
+        feature1 = train_df.loc[i, "feature1"]
+        feature2 = train_df.loc[i, "feature2"]
+        feature3 = train_df.loc[i, "feature3"]
+        
+        learn_example = str(action) + ":" + str(cost) + ":" + str(probability) + " | " + str(feature1) + " " + str(feature2) + " " + str(feature3)
+        vw.learn(learn_example)
+
+    print(vw.get_prediction_type())
```
lalo, line 45:
```suggestion
```

PR #2620, opened 2020-10-30
swap_buffer implementation for comments

peterychang:
All names are tentative
Usage:
```
writer:
    while(!finished) {
      swap_buffer.write_data(buffer, len);
    }
    swap_buffer.finalize_input();

reader:
    auto buffer_data = swap_buffer.get_data();
    uint8_t* buffer = buffer_data.data();
    size_t length = buffer_data.size();
    if(buffer == nullptr) {
      return;
    }
    // Parse buffer
    swap_buffer.return_buffer(buffer_data);
```

The number of buckets per buffer should be a multiple of the number of parser threads. The number of buffers should be >= 2. My testing used `num_buckets = num_parse_threads * 5` and `num_buffers = 3`, which seemed to work well

Tentative results:
997760 examples, no command-line options (other than input file and buffer sizes). Times averaged over 5 runs
| Parse Threads | Time(s)  |
| :------------ | :---- |
| Baseline | 3.811 |
| 1             | 3.350 |
| 2             | 2.636 |
| 4             | 2.601 |
| 8             | 2.594 |
| 16            | 2.487 |

49888 examples, `--quiet -q:: --leave_duplicate_interactions`, 100 runs performed
| Parse Threads | Mean  | Std.Dev. | Min   | Median | Max   |
| :------------ | :---- | :------- | :---- | :----- | :---- |
| Baseline      | 4.876 | 0.577    | 4.009 | 4.756  | 6.438 |
| 1             | 4.053 | 0.282    | 3.667 | 4.055  | 5.302 |
| 2             | 4.203 | 0.207    | 3.791 | 4.172  | 4.669 |
| 4             | 4.330 | 0.332    | 3.747 | 4.256  | 5.965 |
| 8             | 4.984 | 0.425    | 3.946 | 4.987  | 6.868 |

I'm not entirely sure why the times increase with the number of threads but I suspect it has to do with lock contention. The gain we see from 1 thread over baseline likely has to do with the fact that I/O runs on a separate thread from parsing (1 parser thread = 1 io thread, 1 parser thread, 1 ML thread)

### DIFF
--- a/vowpalwabbit/swap_buffer.h
+++ b/vowpalwabbit/swap_buffer.h
@@ -1,0 +1,271 @@
+#pragma once
+
+#include <cstdint>
+#include <cstddef>
+#include <cstring>
+#include <cassert>
+
+#include <algorithm>
+#include <deque>
+#include <list>
+#include <memory>
+
+#include <condition_variable>
+#include <mutex>
+
+#include <iostream>
+
+namespace VW {
+namespace buffer {
+
+class bucket_buffer
+{
+ private:
+  // Use v_array instead? The usage is different enough that it feels wrong though
+  uint8_t* _buffer_begin = nullptr;
+  uint8_t* _buffer_end = nullptr;
+  size_t _num_buckets = 0;
+
+  // size is [1, n+1]
+  // Data range for each bucket is [_bucket_begin[i], _bucket_begin[i+1])
+  std::deque<uint8_t*> _bucket_begin;
+
+  unsigned int _lock_count = 0;
+ private:
+  size_t buffer_capacity() { return _buffer_end - _buffer_begin; }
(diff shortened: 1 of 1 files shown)


peterychang:
The testing implementation was built on top of Cassandra's original design. There are a lot of limitations that need to be addressed (no JSON support, cache is slow and can only run single-threaded) so I didn't use it for this PR

kumpera:
Hey Peter, who's the reader and writer in this context? 

The naming is confusing since your benchmark is read-only.

peterychang:
The writer is an I/O thread thats reading from input
The reader(s) are the parse threads

jackgerrits:
Closing this PR until there is need for this class in the parallel parsing work.

PR #2623, opened 2020-11-04
cats pdf edge cases

olgavrou:
Generating the pdf from pmf had problems for the below cases:

if `bandwidth > 1` then predicted actions around the continuous actions limits (i.e. `action == 1` or `action == num_actions - 1`) lead to problems:
- action smaller than bandwidth (e.g. action = 1 and bandwidth = 3) means that we would overflow when checking `action - bandwidth` (action in this case is `uint32_t`) resulting either in wrong results or segfaults
- if predicted action = num_actions - 1 then resulting pdf would look something like:

|left|right|pdf_value|
| ---|-----|-----------|
|min_value|max_value + bandwidth|pdf_value|
|max_value + bandwidth|max_value|pdf_value|

e.g. running `cats_pdf` with `num_actions=16 bandwidth=3 min_value=185 max_value=23959` would result in

|left|right|pdf_value|
| ---|-----|-----------|
|185|16034.3|2.10314e-06|
|16034.3|25543.9|0.000102002|
|25543.9|23959|2.10314e-06|


test/unit_test/pmf_to_pdf_test.cc
```diff
@@ -32,18 +32,20 @@ struct reduction_test_harness
     for (uint32_t i = 0; i < _predictions.size(); i++)
     { ec.pred.a_s.push_back(ACTION_SCORE::action_score{_predictions[i].first, _predictions[i].second}); }
 
-    cout << "\nec.pred.a_s (PMF): " << endl;
-    for (uint32_t i = 0; i < _predictions.size(); i++)
-    { cout << "(" << ec.pred.a_s[i].action << " : " << ec.pred.a_s[i].score << "), " << endl; } }
+    // cout << "\nec.pred.a_s (PMF): " << endl;
```
olgavrou, line 35: I am commenting these out as I think unit test should not print in general, but leaving in them in in case they are useful for debuggin

akshaykr (approved):
LGTM

PR #2624, opened 2020-11-04
Add name field to learner and add learner callstack print statements

jackgerrits:
- Removed duplicate stack tracking variables in `example_predict`
- Removed double counting of stack depth in `learner`
- Fixed uninitialized memory for stack depth counter
- Fixed invalid learner state due to no construction
- Only ported debug statements to do with learner call stack printing


rajan-chari (approved):

vowpalwabbit/example_predict.h
```diff
@@ -48,7 +48,8 @@ struct example_predict
   // and each char is a namespace.
   std::vector<std::vector<namespace_index>>* interactions;
 
-  uint32_t _current_reduction_depth;  // Used for debugging reductions.  Keeps track of current reduction level
+  // Used for debugging reductions.  Keeps track of current reduction level.
+  uint32_t _current_reduction_depth;
```
olgavrou, line 52: it would be nice if the name included the word `debug` so that it is clear

vowpalwabbit/example_predict.h
```diff
@@ -62,3 +63,7 @@ class VW_DEPRECATED("example_predict is now RAII based. That class can be used i
 
   void clear();
 };
+
+std::string features_to_string(const example_predict& ec);
+std::string depth_indent_string(const example_predict& ec);
```
olgavrou, line 68: same for these methods, they should also include `debug` or `dbg` and/or be capitalized (even though they aren't macros)

vowpalwabbit/learner.h
```diff
@@ -198,6 +218,13 @@ struct learner
   using end_fptr_type = void (*)(vw&, void*, void*);
   using finish_fptr_type = void (*)(void*);
 
+  void log_message(example& ec, const std::string& msg) { VW_DBG(ec) << "[" << name << "." << msg << "]" << std::endl; }
```
olgavrou, line 221: same here

olgavrou:
In order to avoid creating another name from the reduction, and since we know the name of the reduction when we call setup (see [here](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/parse_args.cc#L1202)), we could pass in the name to `setup` and the `setup` methods could forward the name to `init_learner`

vowpalwabbit/example_predict.h
rajan-chari: Done.

vowpalwabbit/example_predict.h
rajan-chari: Not sure about capitalized since that's not following convention in our code.  Adding debug.

vowpalwabbit/learner.h
rajan-chari: added debug.

rajan-chari:
> In order to avoid creating another name from the reduction, and since we know the name of the reduction when we call setup (see [here](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/parse_args.cc#L1202)), we could pass in the name to `setup` and the `setup` methods could forward the name to `init_learner`

Done.

rajan-chari:
All questions raised have been addressed, including deriving learner name from setup_fn name.

vowpalwabbit/kernel_svm.cc
```diff
@@ -899,7 +899,7 @@ VW::LEARNER::base_learner* kernel_svm_setup(options_i& options, vw& all)
 
   params->all->weights.stride_shift(0);
 
-  learner<svm_params, example>& l = init_learner(params, learn, predict, 1);
+  learner<svm_params, example>& l = init_learner(params, learn, predict, 1, all.get_setup_name(kernel_svm_setup));
```
lalo, line 902: not a fan of this solution but not sure what other options are there. it's a bit absurd that the information to get this name is right above it (options_group) but now everything is wired in a circular manner by again adding more state to the all object.

vowpalwabbit/global_data.cc
```diff
@@ -85,6 +85,18 @@ int print_tag_by_ref(std::stringstream& ss, const v_array<char>& tag)
 
 int print_tag(std::stringstream& ss, v_array<char> tag) { return print_tag_by_ref(ss, tag); }
 
+std::string vw::get_setup_name(reduction_setup_fn setup_fn)
```
lalo, line 88: can both these functions have better names?

vowpalwabbit/parse_args.cc
```diff
@@ -1230,17 +1230,20 @@ void register_reductions(vw& all, std::vector<reduction_setup_fn>& reductions)
...
 
       if (base == nullptr)
-        all.reduction_stack.push(std::make_tuple(name_extractor.generated_name, setup_fn));
+        all.reduction_stack.push_back(std::make_tuple(name_extractor.generated_name, setup_fn));
       else
         THROW("fatal: under register_reduction() all setup functions must return nullptr");
     }
   }
+
+  // populate setup_fn -> name map to be used to lookup names in setup_base
```
lalo, line 1245: not a fan of now having two way dictionary for the same information... not sure of other options

vowpalwabbit/kernel_svm.cc
rajan-chari: The name returned by setup can be overridden ridden by parse_args.  So the options_group by itself is not enough.

vowpalwabbit/parse_args.cc
rajan-chari: _setup_name_map  is not identical to reduction_stack.  _setup_name_map has two purposes.  One is to map names to setup functions.   The other is to hold onto the mapping during the recursive call to setup_base that drains the reduction stack.  By the time this information is used in the reduction setup function to pass to init_learner, it is no longer in the reduction stack..

lalo (commented):
- adds more state to the vw/all struct
- no instructions on how to use this
- no documentation
- not running on ci (I think?)
- no example of how the output looks for actual real input and not a simple test
- output not machine parseable (unless someone does a parser), so destined to be consumed by humans eyes only
- no explanation of a success story of this scaffolding, which I've asked for before but I'm sure there are multiple theoretical future use cases.

These are just suggestions to improve this PR, feel free to pick and choose whatever you want. I still don't understand the need of all this permanent scaffolding so I can't weigh the pros and cons of this PR in an unbiased way, or if it meets X design requirements.

vowpalwabbit/parse_args.cc
lalo: The state of this code is temporary, setup_base was going to evolve to use a dictionary holding string->setup_fn to enable the python reductions. This is not on master yet, but on my branch but explained this when justifying that change. I got confused and thought that was on master, so I'm wary on how that's going to look in the future (holding 2 different dicts with the same info).

rajan-chari: Got it.  A couple of possible solutions.  1) Encapsulate both maps in one class that provides both look ups.  2) Reduce the efficiency of one of the lookups in favor or the other look up.  We can cross that bridge at the appropriate time though.

olgavrou (approved):
I agree with some of @lalo 's points. This could use some documentation and we should evolve/think about the name map as setup evolves

rajan-chari:
> * adds more state to the vw/all struct
> * no instructions on how to use this
Good Idea.  I will add these.
> * no documentation
Good Idea.  I will add these.
> * not running on ci (I think?)
Not sure I get this. 
> * no example of how the output looks for actual real input and not a simple test
Not sure I get this. 
> * output not machine parseable (unless someone does a parser), so destined to be consumed by humans eyes only
Traces and logs are not typically machine parseable. 
> * no explanation of a success story of this scaffolding, which I've asked for before but I'm sure there are multiple theoretical future use cases.
I will add documentation to help understand the usage.  I don't want to expand the scope of tracing further at this point.
> 
> These are just suggestions to improve this PR, feel free to pick and choose whatever you want. I still don't understand the need of all this permanent scaffolding so I can't weigh the pros and cons of this PR in an unbiased way, or if it meets X design requirements.

The requirement is to have traces that are helpful in a) comparing before and after reduction changes that modify predict learn path b) tracing execution of reductions in stack-tree form.

vowpalwabbit/global_data.cc
rajan-chari: Do have any suggestions for better names?

lalo: nit: for `map_setup_name` maybe something like `build_setupfn_name_dict`? for `get_setup_name` maybe just adding setupfn i.e. `get_setupfn_name`

rajan-chari: Changed.

lalo:
not running on ci (I think?)
RC> Not sure I get this.
ES> I believe no CI build is exercising this codepath or even building it since I assume is OFF by default. Is this correct?
no example of how the output looks for actual real input and not a simple test
RC> Not sure I get this.
ES> this is a corollary to CI build, best case scenario would be to have a diff test like the rest to make sure no regressions happen. To the minimum it would be good to see a file to see if this code is producing what we actually want. Haven't seen a full example of a vw instance (since its using std::out and other things are using std::out)
output not machine parseable (unless someone does a parser), so destined to be consumed by humans eyes only
RC> Traces and logs are not typically machine parseable.
ES> In my experience they are, matter fact most systems permit searching and linking and filtering but I understand the limits of this. My main concern was that there's a lot of info being printed as tabs that could be useful for bigger things, but probably for the future.
no explanation of a success story of this scaffolding, which I've asked for before but I'm sure there are multiple theoretical future use cases.
RC> I will add documentation to help understand the usage. I don't want to expand the scope of tracing further at this point.
ES> it would be good to at least have documentation on instructions on how to use this (change this bool, build debug, etc) and if theres time you can extend to talk about use cases.

@rajan-chari answered under ES

rajan-chari:
Added documentation explaining how to use reduction trace logging in a private location.  I will move this to the Wiki when the PR is merged.

PR #2625, opened 2020-11-05
Experimental benchmarking pipeline

olgavrou:
This pipeline is meant to perform benchmarks between pull request branches and master using google benchmarks. For now it will just print the comparison in the pipeline. We can decide how to report large differences as a second step.

This is pulling in some of the benchmarks created during the flatbuffer benchmarking, so pulling in those related to text and cache parsing.

Once this is merged to master need to update the first step to actually checkout `master`


.scripts/linux/compare-benchmarks.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+set -x
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_DIR=$SCRIPT_DIR/../../
+cd $REPO_DIR
+
+benchmark/tools/compare.py benchmarks "$1" "$2"
```
jackgerrits, line 9: Looks like this file is missing from the pr?

olgavrou: compare.py? it's part of the google benchmarks repo which is checked out

jackgerrits: Oh I see

jackgerrits (approved):

.github/workflows/run_benchmarks.yml
```diff
@@ -0,0 +1,56 @@
...
+    container:
+      image: vowpalwabbit/ubuntu1604-build:latest
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+        # with:
+          # ref: master # master doesn't have benchmark scripts right now, change once master is merged
+      - name: Install google benchmarks
+        shell: bash
+        run: ./.scripts/linux/install-benchmark.sh # repo dir
```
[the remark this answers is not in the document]
olgavrou: I'll remove that it was for me to remember where it is checking out the benchmarks

PR #2629, opened 2020-11-05
[py] add native python vw config

lalo:
This might be a good first start, the API returns a complicated struct but we should iterate more later.

See test for idea of why it's complicated. The surface level API is simple, one general call to get ALL the options we define and another that gets that VW instances options but filtered by if the reduction was enabled or not.


python/pylibvw.cc
```diff
@@ -54,6 +57,113 @@ const size_t pDECISION_SCORES = 8;
 
 void dont_delete_me(void* arg) {}
 
+class OptionManager
+{
+  std::map<std::string, std::vector<VW::config::option_group_definition>> m_option_group_dic;
+  py::object m_cla;
```
olgavrou, line 63: what does `cla` stand for?

vowpalwabbit/parse_args.cc
```diff
@@ -1197,17 +1197,23 @@ void load_input_model(vw& all, io_buf& io_temp)
...
+  reduction_setup_fn setup_func = std::get<1>(all.reduction_stack.top());
+  std::string setup_func_name = std::get<0>(all.reduction_stack.top());
   all.reduction_stack.pop();
-  auto base = std::get<1>(setup_func)(options, all);
+
+  // 'hacky' way of keeping track of the option group created by the setup_func about to be created
+  options.tint(setup_func_name);
+  auto base = setup_func(options, all);
+  // reset back to general
+  options.tint("general");
```
olgavrou, line 1208: nit: maybe `options.un_tint` could set `"general"` in `options_boost_po` so that we encapsulate the `general` string in one place

python/pylibvw.cc
```diff
@@ -54,6 +57,113 @@ const size_t pDECISION_SCORES = 8;
...
+    auto it = m_option_group_dic.begin();
+
+    auto enabled_reductions = all->enabled_reductions;
+
+    while (it != m_option_group_dic.end())
+    {
+      auto reduction_enabled =
+          std::find(enabled_reductions.begin(), enabled_reductions.end(), it->first) != enabled_reductions.end();
+
+      if (((it->first).compare("general") != 0) && enabled_only && !reduction_enabled)
```
olgavrou, line 133: same here, is there a way to hide the "general" string?

vowpalwabbit/parse_args.cc
lalo: yes, I'm not that happy with this tint solution but haven't figured out anything better. untint or reset_tint could also work though

python/pylibvw.cc
lalo: class, this is where I store the python class that is used to contain each vw option. It should be renamed to something more descriptive

python/pylibvw.cc
```diff
@@ -1058,6 +1174,10 @@ BOOST_PYTHON_MODULE(pylibvw)
       .def("set_tag", &my_set_tag, "change the tag of this prediction")
       .def("predict", &Search::predictor::predict, "make a prediction");
 
+  py::class_<OptionManager, op_manager_ptr>("option_manager", py::no_init)
+      .def("get_options", &OptionManager::get_vw_option_pyobjects, py::return_value_policy<py::return_by_value>(),
+          "do something");
```
lalo, line 1179: update string

vowpalwabbit/parse_args.cc
olgavrou: hm would something similar to the scope_guards logic work? create a tint object which resets to "general" on destruction? (might be an overkill)

lalo: that feels like the using keyword on c#

python/pylibvw.cc
lalo: done

vowpalwabbit/parse_args.cc
lalo: semi-done

python/pylibvw.cc
lalo: done

python/pylibvw.cc
lalo: done

vowpalwabbit/options_boost_po.h
```diff
@@ -142,7 +153,15 @@ struct options_boost_po : public options_i
   template <typename T>
   void add_to_description(std::shared_ptr<typed_option<T>> opt, po::options_description& options_description);
 
+  void add_to_option_group_collection(const option_group_definition& group);
+
 private:
+  // Collection that tracks for now
+  // setup_function_id (str) -> list of option_group_definition
+  std::map<std::string, std::vector<option_group_definition>> m_option_group_dic;
```
olgavrou, line 161: is there a reason we are using `map` instead of `unordered_map`? Do we want these to be ordered?

python/pylibvw.cc
```diff
@@ -56,6 +57,117 @@ const size_t pACTION_PDF_VALUE = 9;
...
+    return nullptr;
+  }
+
+  template <typename TTypes>
+  py::object base_option_to_pyobject(VW::config::base_option& options)
+  {
+    py::object* temp = transform_if_t<typename TTypes::head>(options);
+    if (temp != nullptr)
+    {
+      auto repack = py::object(*temp);
```
olgavrou, line 119: why does this need to be repacked?

python/pylibvw.cc
```diff
@@ -56,6 +57,117 @@ const size_t pACTION_PDF_VALUE = 9;
...
+      it++;
+    }
+    return dres;
+  }
+};
+
+template <>
+py::object OptionManager::base_option_to_pyobject<VW::config::typelist<>>(VW::config::base_option& options)
+{
+  return py::object();
```
olgavrou, line 168: we are just returning an empty object here and not using `options` that is passed in?

vowpalwabbit/options_boost_po.h
lalo: no specific reason, but having them ordered has benefits when outputting them for now. i.e. showing up in almost alphabet order https://github.com/VowpalWabbit/vowpal_wabbit/wiki/VW-arguments-as-of-8.9.0-in-json-format

python/pylibvw.cc
lalo: not a fan of this chunk, but it is to unify all the `py::objects` on the same memory mgmt scheme. If you see `line 97`, this object is being allocated with `new` expression, which would imply that someone is responsible for destroying it later. Boost python api provides `manage_new_object` for these cases, but it gets a bit complicated since I'm also using a `py::dict` not using new down in `line 129`. So my workaround was to instantiate here another copy and deleting the original. Let me know if you have more thoughts about this.

python/pylibvw.cc
lalo: this part is for completeness, a.k.a. make the compiler happy but no one will ever call into `base_option_to_pyobject` with an empty typelist, see `line 150` for more context. It is possible that maybe throwing is better, but I just opted for returning a python None object.

python/pylibvw.cc
olgavrou: I see. Is there a way to avoid the `new` in line 97 or do you have to create the object using `new`?

python/pylibvw.cc
olgavrou: ok sounds good, maybe just add a comment?

olgavrou (approved):

lalo (commented):
a

python/pylibvw.cc
```diff
@@ -57,6 +58,117 @@ const size_t pPDF = 10;
...
+
+      dres[it->first] = option_groups;
+
+      it++;
+    }
+    return dres;
+  }
+};
+
+template <>
```
lalo, line 166:
```suggestion
// specialization needed to compile, this should never be reached since we always use VW::config::supported_options_types
template <>
```

python/pylibvw.cc
lalo: done

python/pylibvw.cc
lalo: I can't think of another way of doing it

PR #2630, opened 2020-11-05
[py] add docs for DFtoVW

lalo:


python/docs/source/vowpalwabbit.dftovw.rst
```diff
@@ -0,0 +1,7 @@
+vowpalwabbit.dftovw
+====================
+
+.. automodule:: vowpalwabbit.dftovw
```
lalo, line 4: i hope its not case sensitive

jackgerrits: I think it is?

lalo: inconclusive , but 'it worked on my machine'

jackgerrits (approved):

PR #2631, opened 2020-11-05
bug fix: action index is inconsistent in explore_eval

cheng-tan:
The input format within the reductions seem to be inconsistent. Some of the reductions assuming "shared" is required, but parser doesn't require example to have "shared". 

So the action index between `get_observed_cost` and the `example sequence` are different if shared feature exists. For `explore_eval` we skip the shared features to keep the action index consistent.

### DIFF
--- a/vowpalwabbit/explore_eval.cc
+++ b/vowpalwabbit/explore_eval.cc
@@ -133,7 +133,7 @@ void do_actual_learning(explore_eval& data, multi_learner& base, multi_ex& ec_se
   if (label_example != nullptr)  // restore label
     label_example->l.cb = data.action_label;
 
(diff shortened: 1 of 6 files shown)


vowpalwabbit/explore_eval.cc
```diff
@@ -68,7 +68,7 @@ void output_example(vw& all, explore_eval& c, example& ec, multi_ex* ec_seq)
   {
     for (uint32_t i = 0; i < preds.size(); i++)
     {
-      float l = get_cost_estimate(&c.known_cost, preds[i].action);
+      float l = get_cost_estimate(&c.known_cost, preds[i].action + 1);
```
jackgerrits, line 71: Why do we need to add 1 here?

jackgerrits (approved):

PR #2634, opened 2020-11-06
Make chain hash the default with text format and warn with json formats

jackgerrits:


peterychang:
I guess there's no reason to allow somebody to disable chain hashing on the text parser? Why can't we make it default behavior for json as well?

jackgerrits:
I don't think disabling it makes sense for the text parser because this is essentially an expansion of the format. Making it default for json is a breaking change we need to deprecate first.

vowpalwabbit/parse_args.cc
```diff
@@ -435,7 +435,7 @@ input_options parse_source(vw& all, options_i& options)
       .add(make_option("no_daemon", all.no_daemon)
                .help("Force a loaded daemon or active learning model to accept local input instead of starting in "
                      "daemon mode"))
-      .add(make_option("chain_hash", parsed_options.chain_hash)
+      .add(make_option("chain_hash", parsed_options.chain_hash_json)
                .help("enable chain hash for feature name and string feature value. e.g. {'A': {'B': 'C'}} is hashed as "
```
lalo, line 439: should the help message mention this is default and only option for json?

jackgerrits: Good catch thankyou

test/unit_test/dsjson_parser_test.cc
```diff
@@ -266,8 +267,9 @@ BOOST_AUTO_TEST_CASE(parse_dsjson_cats_no_label)
   }
 }
 )";
-  auto vw = VW::initialize("--dsjson -t --cats 4 --min_value=185 --max_value=23959 --bandwidth 1 --no_stdin --quiet",
-      nullptr, false, nullptr, nullptr);
+  auto vw = VW::initialize(
```
olgavrou, line 270: I find this confusing, we need to turn chain hash on for json, even if the feature values are numerical?

jackgerrits: The alternative is keeping track of if a text feature was or wasn't seen and warn only if one was seen. Given we are migrating to this behavior it seems simpler to just enforce it on an input type level?

olgavrou: yeah..

olgavrou (approved):

PR #2636, opened 2020-11-09
fixed duplicate quadratic interactions

rajan-chari:
VW models with multiple save/resume cycles added many duplicate quadratic interactions to saved command line.


vowpalwabbit/parse_args.cc
```diff
@@ -1424,7 +1424,7 @@ vw& parse_args(options_i& options, trace_message_t trace_listener, void* trace_c
 
 bool check_interaction_settings_collision(options_i& options, std::string file_options)
 {
-  bool command_line_has_interaction = options.was_supplied("q") || options.was_supplied("quadratic") ||
+  bool command_line_has_interaction = options.was_supplied("-q") || options.was_supplied("quadratic") ||
```
lalo, line 1427: why would the previous one not work? the code that parses seems to do the 'correct' thing regarding short names see line 90, 92, 95 and 139

https://github.com/VowpalWabbit/vowpal_wabbit/blob/2536011d2b1aca53e080862aa48d769c8cf697e4/vowpalwabbit/options_boost_po.cc#L95

maybe a unit test would be helpful?

lalo: unless its not using that options_boost_po options_i impl, didn't verify that

jackgerrits, line 1427: This will explicitly not work for option providers such as json which do not use `-` or `--` like command lines do. This interface should operate on `q` or `quadratic` like it was before. The bug is in how this code handles the value of `-q`

jackgerrits (changes requested):

vowpalwabbit/parse_args.cc
rajan-chari: q was not registered with options object before calling this function.  Fixed it upstream to register -q and reverted this.

rajan-chari: q was not registered with options object before calling this function.  Fixed it upstream to register -q and reverted this.

rajan-chari: Had to roll back registering quadratic since there are many other side effects that interact with loading the model.  Reverting back to testing for '-q' since quadratic option is not yet registered

rajan-chari: Had to roll back registering quadratic since there are many other side effects that interact with loading the model.  Reverting back to testing for '-q' since quadratic option is not yet registered

lalo: this ends up checking if '---q' is present, is that the desired behavior?

rajan-chari:
From what I can see in debugger, it checks m_supplied_options for '-q'.      On Wednesday, November 11, 2020, 01:28:54 PM EST, Eduardo Salinas <notifications@github.com> wrote:  
 
 


@lalo commented on this pull request.

In vowpalwabbit/parse_args.cc:
> @@ -1424,7 +1424,7 @@ vw& parse_args(options_i& options, trace_message_t trace_listener, void* trace_c
 
 bool check_interaction_settings_collision(options_i& options, std::string file_options)
 {
-  bool command_line_has_interaction = options.was_supplied("q") || options.was_supplied("quadratic") ||
+  bool command_line_has_interaction = options.was_supplied("-q") || options.was_supplied("quadratic") ||

this ends up checking if '---q' is present, is that the desired behavior?

vowpalwabbit/parse_args.cc
lalo: responding to your answer, m_supplied_options gets populated after registering the options so I was assuming that to always return false and then try to search '---q'

rajan-chari: Ok changed was_supplied to look for '-' in addition to '--' instead of trying to register quadratic before calling this function.

rajan-chari: Ok changed was_supplied to look for '-' in addition to '--' instead of trying to register quadratic before calling this function.

PR #2638, opened 2020-11-11
Add metadata info to example

olgavrou:
- `metadata_info struct` added to example
- passed to txt label parsers so that anything before `|` that might need to go into `meta_info` can
- needed for cats prediction hint to be passed through the reduction stack (in upcoming PR)
- benchmarked locally, there doesn't seem to be a difference in performance. Will also check via the new experimental benchmark pipeline
- potential next step: move `tag` into `metadata_info`

thanks @jackgerrits for helping with the design of this

### DIFF
--- a/vowpalwabbit/example.cc
+++ b/vowpalwabbit/example.cc
@@ -14,6 +14,7 @@ example::example()
   memset(&l, 0, sizeof(polylabel));
   memset(&pred, 0, sizeof(polyprediction));
   tag = v_init<char>();
+  pred_info.cats_pdf = v_init<VW::continuous_actions::pdf_segment>();
 }
 VW_WARNING_STATE_POP
 
@@ -22,6 +23,7 @@ VW_WARNING_DISABLE_DEPRECATED_USAGE
 example::~example()
 {
   tag.delete_v();
+  pred_info.cats_pdf.delete_v();
   if (passthrough)
   {
     delete passthrough;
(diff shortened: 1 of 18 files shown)


vowpalwabbit/example.cc
```diff
@@ -52,6 +55,10 @@ example::example(example&& other) noexcept
     , in_use(other.in_use)
 {
   other.weight = 1.f;
+  auto& other_cats_pdf = other.meta_info.cats_pdf_hint;
```
peterychang, line 58: It would be more useful to overload the move operator in v_array and just `std::move()` this

olgavrou: good point

olgavrou: actually I found this comment in v_array and I am guessing it holds for move constructors too

```
// v_array cannot have a user-defined constructor, because it participates in various unions.
  // union members cannot have user-defined constructors.
  // v_array() : _begin(nullptr), _end(nullptr), end_array(nullptr), erase_count(0) {}
```

olgavrou: ah you said move operator, never mind

olgavrou: actually, can't add move assignment operator either

copy constructors are not auto generated if you specify a move assignment operator
if copy constructor is non-trivial then you can't use the class in a union

jackgerrits: One day we can add them... :/ Need those unions gone first...

PR #2643, opened 2020-11-13
Cats response during inference based on user provided hints

olgavrou:
During inference and if there is no model (determined by a command line flag) the user can provide prediction hints via new `"pdf"` field in `dsjson`

- Full pdf: 
  - `"pdf": [{"left":<left>, "right":<right>, "pdf_value":<pdf_value>}, {{"left":<left>, "right":<right>, "pdf_value":<pdf_value>}]`
  - validate user provided pdf, if OK use it, if not OK default to uniform random
- Chosen action only: 
  - `"pdf": [{"chosen_action" : <action>}]`
  - create an epsilon-greedy pdf around that action
- Nothing provided but flag set: default to uniform random

if both chosen action and a full pdf is provided, we chose to ignore the chosen action

Used `--first_only` as that is what we use for contextual bandits too, but I am not 100% sure about the name, maybe `--pdf_hints` might be better


vowpalwabbit/cb_explore_pdf.cc
```diff
@@ -34,6 +34,7 @@ struct cb_explore_pdf
   float epsilon;
   float min_value;
   float max_value;
+  bool first_only;
```
olgavrou, line 37: not sure I like `first_only`, maybe `pdf_hints` or something similar..

jackgerrits (approved):

Issue #2646, opened 2020-11-16
Cover (adf and non-adf) to use fixed epsilon when defined


olgavrou:
Right now cover (adf and non-adf) don't make use of passed epsilon, but default to `1.f` and then decay the epsilon during learning.
If epsilon is specified it should be fixed epsilon that does not decay (like `bag` does)

Issue #2645, opened 2020-11-16
Cover (adf and non-adf) to store example counter (for decaying epsilon)


olgavrou:
Cover behavior now (adf and non-adf):
- default epsilon to `1.f`
- decay epsilon while learning (by using example counter) but doesn't store example counter in model file resulting in
  - (a) example counter being reset every time (even when we want to resume learning from the same file) and
  - (b) not used during prediction and hence resulting in uniform exploration during predict

We want to (when epsilon not defined)
- default epsilon to `1.f`
- decay epsilon while learning, and store the counter in the model file to be reused for prediction and when we resume learning form a file

When epsilon is defined:
- use fixed epsilon and don't decay (https://github.com/VowpalWabbit/vowpal_wabbit/issues/2646)

PR #2647, opened 2020-11-16
Fix cover bug

olgavrou:
Issues #2645 and #2646


vowpalwabbit/cb_explore_adf_common.h
```diff
@@ -194,5 +195,13 @@ void cb_explore_adf_base<ExploreType>::finish_multiline_example(
 
   VW::finish_example(all, ec_seq);
 }
+
+template <typename ExploreType>
+inline void cb_explore_adf_base<ExploreType>::save_load(
+    cb_explore_adf_base<ExploreType>& data, io_buf& io, bool read, bool text)
+{
+  data.explore.save_load(io, read, text);
```
pmineiro, line 203: Looks like I'll need exactly this change for another exploration algorithm I'm working on (save_load hook for cb_explore_adf), so I'll wait for this to merge.

Question: if the ExploreType doesn't call set_save_load ... this is safe?

vowpalwabbit/cb_explore_adf_cover.cc
```diff
@@ -239,14 +278,27 @@ VW::LEARNER::base_learner* setup(config::options_i& options, vw& all)
...
-  auto data = scoped_calloc_or_throw<explore_type>(
-      cover_size, psi, nounif, first_only, as_multiline(all.cost_sensitive), all.scorer, cb_type_enum);
+  auto data = scoped_calloc_or_throw<explore_type>(cover_size, psi, nounif, epsilon, epsilon_decay, first_only,
+      as_multiline(all.cost_sensitive), all.scorer, cb_type_enum, all.model_file_ver);
 
   VW::LEARNER::learner<explore_type, multi_ex>& l = init_learner(
       data, base, explore_type::learn, explore_type::predict, problem_multiplier, prediction_type_t::action_probs);
 
   l.set_finish_example(explore_type::finish_multiline_example);
+  l.set_save_load(explore_type::save_load);
```
pmineiro, line 301: Oh I see, this sets up the hook, so if you call this, you better have defined a class method.

jackgerrits (approved):

vowpalwabbit/cb_explore_adf_cover.cc
olgavrou: Exactly :D You would get a compiler error otherwise so it should be safe as it is

PR #2651, opened 2020-11-17
[test] runtests parser

lalo:


runtests_parser.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def commit_parsed_test(self):
+        if self.temp_test is not None:
+            self.temp_test.clean()
+            self.results.append(self.temp_test)
+            self.temp_test = None
+
+    def process_line(self, line):
+        if Parser.is_perl_comment(line):
+            result = Parser.process_perl_comment(line)
```
jackgerrits, line 89: Does this handle multiline comments?

runtests_parser.py
```diff
@@ -0,0 +1,121 @@
...
+        self.temp_test = None
+        self.results = []
+
+    @staticmethod
+    def is_perl_comment(line):
+        return line[0] == "#"
+
+    @staticmethod
+    def is_vw_command(line):
+        return line.split()[0] == "{VW}"
```
jackgerrits, line 57: Does this handle multiline commands with continuations? "\"

lalo: Yes, but its more of a state machine rather than trying to foresee. It's handled via more_vw flag, basically checking if line ends with `\` and then parsing as such on line 98

runtests_parser.py
lalo: Yes, they all get merged if they are defined on the same 'test' chunk

lalo: see line 95 and method add_more_comments

jackgerrits (approved):

jackgerrits:
Would be cool if this could be extended to define which files are consumed and which are produced by the command? This would make running tests in isolation possible

lalo:
@jackgerrits I believe all the files listed under files are output only. If by 'isolation' you mean running them parallel, I think you can achieve that by checking if it has an element on depends_on and queuing that test before actually running the test you are on. Granted, that would run some tests more than once but it should be minimal. What convenience do you see by adding a list of input files?

test/runtests_parser.py
```diff
@@ -0,0 +1,222 @@
...
+        elif Parser.is_filename_of_testset(line):
+            self.curr_test.add_file(line)
+        else:
+            self.curr_test.add_bash_command(line)
+    
+    def get_results(self):
+        self.commit_parsed_test()
+        return self.results
+
+def file_to_obj(filename):
```
lalo, line 201: @jackgerrits added this for in memory python usage

jackgerrits (approved):

PR #2657, opened 2020-11-19
add learn benchmarks for learn/predict

olgavrou:


rajan-chari (approved):
Looks good.

rajan-chari:
An example with more features ~100+  would help test the benchmarking more.

PR #2659, opened 2020-11-21
Do not add visibility flag on MacOS to silence warning

jackgerrits:
Using this flag causes many warnings about differing visibility settings:
```
ld: warning: direct access in function 'void predict_or_learn_bandit_adf<false>(warm_cb&, VW::LEARNER::learner<char, std::__1::vector<example*, std::__1::allocator<example*> > >&, example&, int)' from file 'libvw.a(warm_cb.cc.o)' to global weak symbol 'typeinfo for VW::vw_exception' from file 'CMakeFiles/vw-bin.dir/main.cc.o' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
```


olgavrou (commented):
Does this mean that visibility isn't hidden in macos or that it will be hidden by default and we don't need the flag?

jackgerrits:
I am not sure if the root cause here, and this patch probably isn't the fix. Not sure if the issue is only present on MacOS or if the warnings only occur there. But with the warnings it kind of clobbers the terminal output since there are so many...

peterychang:
I wonder if it has to do with the fact that main.cc uses several header-only un-exported symbols (like vw_exception)

jackgerrits:
It could be but those flags should apply to vw-bin as well as the library because they are PUBLIC cmake options.

jackgerrits:
Closing this as I think it needs more thought. Transferred this over to issue #2678 to track.

PR #2660, opened 2020-11-21
Fix options warnings on macos

jackgerrits:


olgavrou (commented):
I think the problem here is that we are using `auto` along with the reference operator for `std::vector<bool>` which returns a proxy object instead of your classical reference to the vector's underlying object. It think it might be possible to avoid this specialization by replacing `auto` with `T` in the for loops

olgavrou (approved):

PR #2663, opened 2020-11-23
Syntheticcover

pmineiro:
This simple exploration strategy only uses a single predictor yet compares favorably with `--bag 5` on circa 500 cbify-d openml classification datasets.


vowpalwabbit/cb_explore_adf_synthcover.cc
```diff
@@ -0,0 +1,223 @@
...
+{
+private:
+  float _epsilon;
+  float _psi;
+  size_t _synthcoversize;
+  std::shared_ptr<rand_state> _random_state;
+
+  VW::version_struct _model_file_version;
+
+  v_array<ACTION_SCORE::action_score> _action_probs;
```
jackgerrits, line 40: This needs to be initialized in the constructor.

vowpalwabbit/cb_explore_adf_synthcover.cc
```diff
@@ -0,0 +1,223 @@
...
+  for (uint32_t i = 0; i < num_actions; i++) _action_probs.push_back({i, 0.});
+
+  policyaction.clear();
+  for (uint32_t i = 0; i < _synthcoversize;)
+  {
+    std::pop_heap(
+        preds.begin(), preds.end(), [](const ACTION_SCORE::action_score& a, const ACTION_SCORE::action_score& b) {
+          return ACTION_SCORE::score_comp(&a, &b) > 0;
+        });
+    // NB: what STL calls pop_back(), v_array calls pop().  facepalm.
```
jackgerrits, line 117: Yeah, we've been planning on fixing that for a while

jackgerrits (approved):

vowpalwabbit/cb_explore_adf_synthcover.cc
pmineiro: Done.

vowpalwabbit/vw_core.vcxproj
```diff
@@ -145,6 +145,7 @@
     <ClInclude Include="cb_algs.h" />
     <ClInclude Include="cb_dro.h" />
     <ClInclude Include="cb_explore_adf_bag.h" />
+    <ClInclude Include="cb_explore_adf_synthcover.h" />
```
olgavrou, line 148: I think you also need to add the `.cc` file (after line 270 in this file)

vowpalwabbit/cb_explore_adf_synthcover.cc
```diff
@@ -0,0 +1,224 @@
...
+private:
+  float _epsilon;
+  float _psi;
+  size_t _synthcoversize;
+  std::shared_ptr<rand_state> _random_state;
+
+  VW::version_struct _model_file_version;
+
+  v_array<ACTION_SCORE::action_score> _action_probs;
+  std::vector<int> policyaction;
```
olgavrou, line 41: nit: these three private vars should probably be pre-pended with "_" too

vowpalwabbit/cb_explore_adf_synthcover.cc
```diff
@@ -0,0 +1,224 @@
...
+
+cb_explore_adf_synthcover::cb_explore_adf_synthcover(float epsilon, float psi, size_t synthcoversize,
+    std::shared_ptr<rand_state> random_state, VW::version_struct model_file_version)
+    : _epsilon(epsilon)
+    , _psi(psi)
+    , _synthcoversize(synthcoversize)
+    , _random_state(random_state)
+    , _model_file_version(model_file_version)
+    , _action_probs(v_init<ACTION_SCORE::action_score>())
+    , min_cost(0.0)
```
olgavrou, line 68: are costs ever negative? If not, doesn't `min_cost` need to be initialized to a max numeric limit instead of `0.0` in the constructor? Otherwise it seems it will never get replaced.

vowpalwabbit/cb_explore_adf_synthcover.cc
```diff
@@ -0,0 +1,224 @@
...
+        preds.begin(), preds.end(), [](const ACTION_SCORE::action_score& a, const ACTION_SCORE::action_score& b) {
+          return ACTION_SCORE::score_comp(&a, &b) > 0;
+        });
+    // NB: what STL calls pop_back(), v_array calls pop().  facepalm.
+    auto minpred = preds.pop();
+
+    auto secondminpred = preds[0];
+    for (; secondminpred.score >= minpred.score && i < _synthcoversize; i++)
+    {
+      policyaction.push_back(minpred.action);
```
olgavrou, line 124: I don't understand what `policyaction` is used for. I can only see it being cleared or having the `minipred.action` pushed to it but I can't see it being used anywhere else.

pmineiro: This is vestigial, I'll get rid of it.  I was analytically centering but playing the uniform policy performed similarly and removed a hyperparameter.

vowpalwabbit/cb_explore_adf_synthcover.cc
pmineiro: It's ok to assume that a cost of 0 was observed at the beginning of time. I just need to clip the underlying prediction to something reasonable, otherwise the exploration bonus can get overwhelmed by large predictions.

vowpalwabbit/vw_core.vcxproj
lalo: thanks!

PR #2680, opened 2020-12-04
feat: Add cb_to_cb_adf reduction

lalo:
Purpose: run before cb, cb_explore, cbify and cb_adf related reductions
    This will 'translate' cb (non adf) input commands into their cb_adf counterparts
    Except when:
        - the model file loaded is from a version older or including 8.9.0
        - user bypasses this translation step using '--cb_force_legacy'
        - user specifies the cb_type to 'dm' (not implemented in adf)
    Related files: cb_algs.cc, cb_explore.cc, cbify.cc

'translation' layer from cb -> cb_adf

- [x] probably needs more testing
- [x] migrate other runtests to this one
- [x] figure out naming/release
- [x] figure out model version in case this becomes "--cb"
- [x] cb_explore integration

questions / pending
- [x] how is cb_eval label parser different than standard cb label parser (?)

Collab with @pmineiro


vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,115 @@
...
+
+VW::LEARNER::base_learner* cb_to_cb_adf_setup(options_i& options, vw& all)
+{
+  bool eval = false;
+  size_t num_actions;
+
+  option_group_definition new_options("Contextual Bandit Options");
+  new_options
+      .add(make_option("new_cb", num_actions).keep().necessary().help("Use contextual bandit learning with <k> costs"))
+      .add(make_option("eval", eval).help("Evaluate a policy rather than optimizing."));
```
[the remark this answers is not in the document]
pmineiro: Let's "take over" `--cb`, map the old code onto `--oldcb`, and then check the model version when loading in order to preserve backwards compatability.

pmineiro:
> 'translation' layer from cb -> cb_adf
> 
> * [ ]  probably needs more testing

The backwards compatibility stuff needs to be tested, so we should make a `--cb` file with the old version in order to facilitate a RunTests.

For Q/A on the reduction, we can test the code using coba.

pmineiro:
> > 'translation' layer from cb -> cb_adf
> > 
> > * [ ]   probably needs more testing
> 
> The backwards compatibility stuff needs to be tested, so we should make a `--cb` file with the old version in order to facilitate a RunTests.
> 
> For Q/A on the reduction, we can test the code using coba.

Also, it's unclear whether or not `--cb_explore` "just works" ... we should test this, we might need to make some modifications.

vowpalwabbit/cb_to_cb_adf.cc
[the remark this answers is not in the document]
lalo: did an initial takeover, hardcoded everything to old_cb for now. backward compat will come in a later commit in this pr

vowpalwabbit/cbify.cc
```diff
@@ -145,8 +139,7 @@ void copy_example_to_adf(cbify& data, example& ec)
     // offset indices for given action
     for (features& fs : eca)
     {
-      for (feature_index& idx : fs.indicies)
-      { idx = ((((idx >> ss) * 28904713) + 4832917 * (uint64_t)a) << ss) & mask; }
+      for (feature_index& idx : fs.indicies) { idx = (((idx >> ss) * adf_data.num_actions + a) << ss) & mask; }
```
lalo, line 142: @pmineiro wouldnt this change regress models using the previous way since the weight is located elsewhere now (different hash)?

pmineiro: Yes and no.  

**No**: `--cbify` is  used for testing implementations against a supervised dataset, and the details of how the features were induced were private already, so nobody was saving and reusing the model.  Therefore ignore this issue.

**Yes**: Your comment does apply to the (previous implementation of) `--cb`, which is why the model load of older model versions needs to be modified.

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,117 @@
...
+      .add(make_option("eval", eval).help("Evaluate a policy rather than optimizing."))
+      .add(make_option("force_legacy", force_legacy).help("Default to old cb implementation"));
+
+  if (!options.add_parse_and_check_necessary(new_options)) return nullptr;
+
+  // if user specified both old_cb and cb, we default to old_cb
+  if (options.was_supplied("old_cb")) return nullptr;
+
+  // models created with older version should default to --old_cb
+  if (all.model_file_ver != EMPTY_VERSION_FILE) compat_old_cb = !(all.model_file_ver >= VERSION_FILE_WITH_CB_TO_CBADF);
```
pmineiro, line 90: Hey this looks hot but I'd love a test.  Please create a small --cb file with a legacy version of vw, save it, and make a RunTests that loads that file and gets the output it expects.

java/src/test/java/vowpalWabbit/learner/VWLearnersTest.java
```diff
@@ -32,7 +32,7 @@
     @Test
     public void testWrongType() {
         thrown.expect(ClassCastException.class);
-        VWScalarsLearner learner = VWLearners.create("--cb 4 --quiet");
+        VWScalarsLearner learner = VWLearners.create("--cb 4 --quiet --force_legacy");
```
pmineiro, line 35: Rename flag to "--cb-force-legacy" since it is cb specific.

lalo: done

test/train-sets/ref/rcv1_raw_cb_explore.stderr
```diff
@@ -8,23 +8,23 @@ power_t = 0.5
...
-0.975000 0.000000            2            2.0  1:0:0.5 2:0.975000      598
-0.487500 0.000000            4            4.0  1:0:0.5 2:0.975000      784
-0.250000 0.012500            8            8.0  1:0:0.5 2:0.975000      850
-0.618750 0.987500           16           16.0  1:1:0.5 2:0.975000      118
-0.318750 0.018750           32           32.0  1:0:0.5 1:0.975000      166
-0.321875 0.325000           64           64.0  2:1:0.5 1:0.975000      340
-0.244922 0.167969          128          128.0  1:1:0.5 2:0.975000      610
-0.231250 0.217578          256          256.0  2:0:0.5 2:0.975000      712
-0.216992 0.202734          512          512.0  2:1:0.5 1:0.975000      424
+2.000000 2.000000            1            1.0    known        1      280
```
lalo, line 14: @pmineiro I guess we need a cb_explore specific print function

vowpalwabbit/offset_tree.cc
```diff
@@ -260,6 +260,9 @@ base_learner* setup(VW::config::options_i& options, vw& all)
   // Ensure that cb_explore will be the base reduction
   if (!options.was_supplied("cb_explore")) { options.insert("cb_explore", "2"); }
 
+  // default to usual implementation
+  options.insert("cb_force_legacy", "");
```
pmineiro, line 264: Todo: revisit after supporting cb_explore.

vowpalwabbit/cb_explore.cc
```diff
@@ -255,15 +256,13 @@ void print_update_cb_explore(vw& all, bool is_test, example& ec, std::stringstre
...
   float loss = 0.;
+  auto known_cost = get_observed_cost(ld);
 
-  cb_to_cs& c = data.cbcs;
-
-  if ((c.known_cost = get_observed_cost(ld)) != nullptr)
-    for (uint32_t i = 0; i < ec.pred.a_s.size(); i++)
-      loss += get_cost_estimate(c.known_cost, c.pred_scores, i + 1) * ec.pred.a_s[i].score;
+  if (known_cost != nullptr)
+    loss = CB_ALGS::get_cost_estimate(known_cost, known_cost->action) * ec.pred.a_s[known_cost->action - 1].score;
```
lalo, line 265: this loss change changes legacy cb_explore, @pmineiro that ok?

test/train-sets/ref/rcv1_raw_cb_dm_test.stderr
```diff
@@ -8,23 +8,23 @@ power_t = 0.5
...
--0.213881 -0.589583            2            2.0    known        1      598
--0.257713 -0.301545            4            4.0    known        1      784
--0.170628 -0.083542            8            8.0    known        1      850
-0.049008 0.268643           16           16.0    known        2      118
-0.029790 0.010573           32           32.0    known        1      166
-0.108944 0.188098           64           64.0    known        1      340
-0.079725 0.050506          128          128.0    known        2      610
-0.094953 0.110182          256          256.0    known        2      712
-0.114991 0.135028          512          512.0    known        1      424
+0.000000 0.000000            1            1.0    known        2      280
```
lalo, line 14: we're missing something here

vowpalwabbit/cb_explore.cc
pmineiro: Hmm.  Now we are computing an IPS estimate.  We'll need to dig into that other overload of `get_cost_estimate` to see what it was doing, perhaps it was doing a doubly robust estimate.

test/train-sets/ref/rcv1_raw_cb_dm_test.stderr
pmineiro: There is no direct method (`dm`) in the adf branch.  It's probably defaulted to `mtr`.  We should decide what to do ... maybe treat `dm` similarly to `--cb-force-legacy` ... but with a big fat warning because it is far inferior to `mtr`.

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,179 @@
...
+  bool eval = false;
+  std::string type_string = "mtr";
+  uint32_t num_actions;
+  uint32_t cbx_num_actions;
+  uint32_t cbi_num_actions;
+
+  option_group_definition new_options("Contextual Bandit Options");
+  new_options
+      .add(make_option("cb_to_cbadf", num_actions)
+               .keep()
```
lalo, line 92: not needed

vowpalwabbit/cb_to_cb_adf.cc
lalo: pending

lalo: done: Test 276 in this PR

test/train-sets/ref/help.stdout
```diff
@@ -214,6 +214,16 @@ Experience Replay:
                               example be played (default: 1 = permuting)
 Offset tree Options:
   --ot arg              Offset tree with <k> labels
+Contextual Bandit Options:
```
lalo, line 217: not happy with --help output

python/tests/test_cb.py
```diff
@@ -1,11 +1,18 @@
 import pandas as pd
 import sklearn as sk
 import numpy as np
+# import os
```
olgavrou, line 4: nit: remove un-used import

python/tests/test_cb.py
```diff
@@ -50,6 +57,9 @@ def test_getting_started_example():
         feature3 = test_df.loc[j, "feature3"]
         choice = vw.predict("| "+str(feature1)+" "+str(feature2)+" "+str(feature3))
         assert isinstance(choice, int), "choice should be int"
-        assert choice == 3, "predicted action should be 3"
+        assert choice == 3, "predicted action should be 3 instead of " + choice
 
     vw.finish()
+
```
olgavrou, line 63: nit: remove commented out code

test/train-sets/ref/cbe_cover_e_dec.stderr
```diff
@@ -1,22 +1,23 @@
...
+Enabled reductions: gd, scorer, csoaa_ldf, cb_adf, cb_explore_adf_cover, shared_feature_merger, cb_to_cbadf
 average  since         example        example  current  current  current
 loss     last          counter         weight    label  predict features
 0.000000 0.000000            1            1.0  1:0:0.5 1:0.500000        3
-0.323483 0.646967            2            2.0  2:1:0.5 1:0.500000        3
-0.456686 0.589888            4            4.0  1:2:0.5 2:0.591752        3
-0.542755 0.628824            8            8.0  2:1:0.5 2:0.732739        3
+0.500000 1.000000            2            2.0  2:1:0.5 1:0.500000        3
+0.841752 1.183503            4            4.0  1:2:0.5 1:0.591752        3
+0.670876 0.500000            8            8.0  2:1:0.5 1:0.666667        3
```
olgavrou, line 16: why do these change?

test/train-sets/ref/help.stdout
olgavrou: nit, suggestions: 
old->legacy
cbexploreadf -> cb_explore_adf

Maps cb_adf to cb -> shouldn't that be the other way around, i.e. cb to cb_adf?

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,186 @@
...
+};
+
+template <bool is_learn>
+void predict_or_learn(cb_to_cb_adf& data, multi_learner& base, example& ec)
+{
+  data.adf_data.copy_example_to_adf(*data.weights, ec);
+
+  if (is_learn && !CB::is_test_label(&ec.l.cb))
+  {
+    uint32_t chosen_action = ec.l.cb.costs[0].action - 1;
```
olgavrou, line 33: can `ec.l.cb.costs[0].action` here ever be `0` causing an overflow in `chosen_action`?

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,186 @@
...
+      CB::label ld = data.adf_data.ecs[chosen_action]->l.cb;
+      data.adf_data.ecs[chosen_action]->l.cb = ec.l.cb;
+      base.learn(data.adf_data.ecs);
+      data.adf_data.ecs[chosen_action]->l.cb = ld;
+
+      CB::default_label(&data.adf_data.ecs[chosen_action]->l.cb);
+    }
+    else
+    {
+      std::cerr << "warning: malformed label blah outside of range bluh. skipping." << std::endl;
```
olgavrou, line 46: should we be calling `predict` here if the label is outside of range (we are still in the `learn` loop)

vowpalwabbit/cbify.cc
```diff
@@ -145,8 +157,7 @@ void copy_example_to_adf(cbify& data, example& ec)
     // offset indices for given action
     for (features& fs : eca)
     {
-      for (feature_index& idx : fs.indicies)
-      { idx = ((((idx >> ss) * 28904713) + 4832917 * (uint64_t)a) << ss) & mask; }
+      for (feature_index& idx : fs.indicies) { idx = (((idx >> ss) * num_actions + a) << ss) & mask; }
```
olgavrou, line 160: do we know what these constants were?

test/train-sets/ref/help.stdout
```diff
@@ -424,8 +434,9 @@ Contextual Bandit with Action Dependent Features:
                         (no clipping).
   --cb_type arg         contextual bandit method to use in {ips, dm, dr, mtr, 
                         sm}. Default: mtr
-Contextual Bandit Options:
-  --cb arg              Use contextual bandit learning with <k> costs
+Deprecated: Contextual Bandit Options:
```
olgavrou, line 437: Do we also want to mark the setup function for this reduction with `VW_DEPRECATED` to discourage it's usage?

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,186 @@
...
+    uint32_t chosen_action = ec.l.cb.costs[0].action - 1;
+    if (chosen_action < data.adf_data.num_actions)
+    {
+      // TODO: do i need a guard here?
+      CB::label ld = data.adf_data.ecs[chosen_action]->l.cb;
+      data.adf_data.ecs[chosen_action]->l.cb = ec.l.cb;
+      base.learn(data.adf_data.ecs);
+      data.adf_data.ecs[chosen_action]->l.cb = ld;
+
+      CB::default_label(&data.adf_data.ecs[chosen_action]->l.cb);
```
olgavrou, line 42: we are setting `data.adf_data.ecs[chosen_action]->l.cb` twice? one on line 40 and then on line 42 to feault label?

vowpalwabbit/cb_explore.cc
```diff
@@ -286,9 +285,9 @@ void output_example(vw& all, cb_explore& data, example& ec, CB::label& ld)
   print_update_cb_explore(all, CB::cb_label.test_label(&ld), ec, sso);
 }
 
-void finish_example(vw& all, cb_explore& c, example& ec)
+void finish_example(vw& all, cb_explore&, example& ec)
```
olgavrou, line 288: will the copied examples in `adf_data.ecs` be cleared for re-use here too?

vowpalwabbit/cb_to_cb_adf.cc
lalo: I believe so, this is user input so it could be anything. chosen_action is an unsigned int and then right after on the next line we compare it against num_actions. If it overflows this check should fail.

python/tests/test_cb.py
lalo: I'll remove the comments on this file once I'm done addressing most comments/feedback. I've been using them as a debug entrypoint.

vowpalwabbit/cb_to_cb_adf.cc
lalo: Yes, good q. We figured that it was an inoffensive thing to do if the user gave us an example without a proper label. The other options were failing/throwing but those seemed too harsh. We should think about possible side-effects of doing this though.

test/train-sets/ref/help.stdout
lalo: VW_DEPRECATED might be for API public facing functions, but not entirely sure. Maybe @jackgerrits can chime in. The current strategy was going to be:
- figure out how to message this change via --help
- add legacy on cb on ` Enabled reductions: gd, scorer, cb (legacy), ...`

Since those two items are the way end users interact with VW. A possible third item would be to hide completely from --help the legacy cb but might be too radical. Same hide() from #2710

vowpalwabbit/cb_explore.cc
lalo: I'm actually blindly trusting in that calling copy again on the next run would bring those examples to a good state from the get go. I'll verify if this is the case.

vowpalwabbit/cb_to_cb_adf.cc
lalo: I'll take a look on these 4 lines again.

test/train-sets/ref/cbe_cover_e_dec.stderr
lalo: Ok, this changes for two reasons.

1) The left side of the diff should be compared with the new file right below this one called `test/train-sets/ref/cbe_cover_e_dec_legacy.stderr`. If we compare those, there is a difference due to a change introduced in this PR where we change the loss function to use IPS estimate instead of doubly robust. See the change in `cb_explore.cc` under `generic_output_example` fn.

2) If you are asking why the left side and the right side are different, it uses a different stack (cb vs cb_adf) + the loss function change mentioned on the previous point.

@pmineiro feel free to chime in in case I'm misremembering anything

vowpalwabbit/cbify.cc
lalo: Not sure how those constants were concocted in the past, but this change is to avoid collision between the actions on the same example. @pmineiro feel free to chime in this one too.

vowpalwabbit/cb_explore.cc
```diff
@@ -341,6 +342,8 @@ base_learner* cb_explore_setup(options_i& options, vw& all)
     std::stringstream ss;
     ss << data->cbcs.num_actions;
     options.insert("cb", ss.str());
+    // --cb_explore only works with legacy cb
+    options.insert("cb_force_legacy", "");
```
lalo, line 346: not needed anymore

test/train-sets/ref/inv_hash_load_model.invert.txt
```diff
@@ -1,4 +1,4 @@
-Version 8.9.0
```
jackgerrits, line 1: Are you planning on checking in a rev here?

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,186 @@
...
+  options.add_parse_and_check_necessary(new_options);
+
+  // ANY model created with older version should default to --cb_force_legacy
+  if (all.model_file_ver != EMPTY_VERSION_FILE) compat_old_cb = !(all.model_file_ver >= VERSION_FILE_WITH_CB_TO_CBADF);
+
+  // not compatible with adf
+  if (options.was_supplied("cbify_reg")) compat_old_cb = true;
+
+  // dm not implemented in cb_adf
+  if (type_string == "dm") compat_old_cb = true;
```
jackgerrits, line 122:
```suggestion
  compat_old_cb = type_string == "dm";
```

vowpalwabbit/cbify.cc
```diff
@@ -126,15 +115,38 @@ void finish_cbify_reg(cbify_reg& data, std::ostream* trace_stream)
...
   {
-    auto& eca = *adf_data.ecs[a];
+    ecs[a]->pred.a_s.delete_v();
+    VW::dealloc_example(CB::cb_label.delete_label, *ecs[a]);
+    free_it(ecs[a]);
+  }
+}
+
+void cbify_adf_data::copy_example_to_adf(parameters& weights, example& ec)
+{
```
jackgerrits, line 143: Rather than depending on the entire `parameters` object, can this instead just depend on the mask and stride shift?

vowpalwabbit/cbify.cc
jackgerrits: Are we sure it is okay to remove these constants? This would break existing users of cbify no?

test/train-sets/ref/inv_hash_load_model.invert.txt
```diff
@@ -1,4 +1,4 @@
...
+Version 8.9.1
```
lalo, line 1: 8.10.0

vowpalwabbit/cbify.cc
pmineiro: cbify is only used for research, and the details of how it creates contextual bandit example features are not in any public contract.

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,186 @@
...
+      std::cerr << "warning: malformed label blah outside of range bluh. skipping." << std::endl;
+      base.predict(data.adf_data.ecs);
+    }
+  }
+  else
+  {
+    base.predict(data.adf_data.ecs);
+  }
+
+  if (data.explore_mode) { ec.pred.a_s = data.adf_data.ecs[0]->pred.a_s; }
```
lalo, line 55: I believe this is the same as #2730

vowpalwabbit/cb_to_cb_adf.cc
lalo: they can't since if type_string != "dm" it would end up putting a false on compat_old_cb and that would revert the previous case in line 119.

vowpalwabbit/cbify.cc
lalo: this might be good to do on its own PR, since I see the pattern of extracting mask and stride shift in multiple places

test/train-sets/ref/inv_hash_load_model.invert.txt
lalo: @jackgerrits we chatted about this offline, but using .10. is a bad idea!

lalo: see #2774

lalo:
blocked on https://github.com/VowpalWabbit/vowpal_wabbit/pull/2774 unless we go straight to 9.0.0

vowpalwabbit/cb_to_cb_adf.cc
lalo: done

test/train-sets/ref/inv_hash_load_model.invert.txt
lalo: yes

vowpalwabbit/cb_to_cb_adf.cc
lalo: good catch, the second one should not be needed

vowpalwabbit/cb_explore.cc
lalo: not needed, the copy fn calls into default_label which calls into clear. Let me know if this is what you meant.

python/tests/test_cb.py
```diff
@@ -59,7 +66,7 @@ def test_getting_started_example():
         feature3 = test_df.loc[j, "feature3"]
         choice = vw.predict("| "+str(feature1)+" "+str(feature2)+" "+str(feature3))
         assert isinstance(choice, int), "choice should be int"
-        assert choice == 3, "predicted action should be 3"
+        assert choice == 3, "predicted action should be 3 instead of " + choice
```
lalo, line 69:
```suggestion
        assert choice == 3, "predicted action should be 3 instead of " + str(choice)
```

python/tests/test_cb.py
lalo: done

python/tests/test_cb.py
lalo: done

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,183 @@
...
+               .help("Translate cb explore to cbexploreadf. Disable with cb_force_legacy."))
+      .add(make_option("cbify", cbi_num_actions)
+               .keep()
+               .help("Translate cbify to cbexploreadf. Disable with cb_force_legacy."))
+      .add(make_option("cb_type", type_string).keep().help("contextual bandit method to use in {}"))
+      .add(make_option("cb_force_legacy", force_legacy).keep().help("Default to old cb implementation"));
+
+  options.add_parse_and_check_necessary(new_options);
+
+  if (options.was_supplied("eval")) return nullptr;
```
lalo, line 112: need to double check this

olgavrou (approved):

test/train-sets/ref/cbe_cover_e_dec.stderr
```diff
@@ -1,23 +1,24 @@
 final_regressor = models/cover_e_dec.model
 predictions = cover_e_dec_train.pred
+warning: currently, mtr is only used for the first policy in cover, other policies use dr
```
lalo, line 3: @olgavrou you worked recently on some change regarding this error message right? am I missing something?

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,212 @@
...
+  }
+  else
+  {
+    l = &init_learner(
+        data, base, predict_or_learn<true>, predict_or_learn<false>, 1, prediction_type_t::multiclass, "cb_to_cb_adf");
+  }
+
+  l->set_finish_example(finish_example);
+
+  all.delete_prediction = nullptr;
```
lalo, line 209: deprecated?

test/train-sets/ref/cbe_cover_e_dec.stderr
olgavrou: No that was the interactions warning message, this one looks like it has been around for a while

olgavrou: it looks like it got added since the `cb_explore_adf_cover` reduction got enabled

vowpalwabbit/cb_explore.cc
```diff
@@ -278,7 +284,7 @@ void output_example(vw& all, cb_explore& data, example& ec, CB::label& ld)
     if (ec.pred.a_s[i].score > maxprob)
     {
       maxprob = ec.pred.a_s[i].score;
-      maxid = i + 1;
+      maxid = ec.pred.a_s[i].action + 1;
```
lalo, line 287: suspect

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,212 @@
...
+  VW::finish_example(all, ec);
+}
+
+/*
+    Purpose: run before cb, cb_explore, cbify and cb_adf related reductions
+    This will 'translate' cb (non adf) input commands into their cb_adf counterparts
+    Except when:
+        - the model file loaded is from a version older or including 8.9.0
+        - user bypasses this translation step using '--cb_force_legacy'
+        - user specifies the cb_type to 'dm' (not implemented in adf)
```
lalo, line 108: not true

test/train-sets/ref/help.stdout
```diff
@@ -242,8 +242,9 @@ CATS Tree Options:
                               #actions
   --link arg                  Specify the link function: identity, logistic, 
                               glf1 or poisson
-Contextual Bandit Options:
-  --cb arg              Use contextual bandit learning with <k> costs
+Deprecated: Contextual Bandit Options:
+  --cb arg              Deprecated: Use contextual bandit learning with <k>
```
lalo, line 246: not technically true, this will redirect to cb_adf

vowpalwabbit/cb_algs.cc
```diff
@@ -143,17 +143,19 @@ base_learner* cb_algs_setup(options_i& options, vw& all)
   std::string type_string = "dr";
   bool eval = false;
 
-  option_group_definition new_options("Contextual Bandit Options");
+  option_group_definition new_options("Deprecated: Contextual Bandit Options");
```
pmineiro, line 146: Not depricated?

vowpalwabbit/cb_explore.cc
```diff
@@ -356,6 +358,8 @@ base_learner* cb_explore_setup(options_i& options, vw& all)
     std::stringstream ss;
     ss << data->cbcs.num_actions;
     options.insert("cb", ss.str());
+    // --cb_explore only works with legacy cb
```
pmineiro, line 361: just assert that cb_force_legacy was supplied rather than inserting it

pmineiro: even worse, given line 348, this is totally redundant, should be deleted

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,237 @@
...
+
+  if (data.explore_mode) { ec.pred.a_s = std::move(data.adf_data.ecs[0]->pred.a_s); }
+  else
+  {
+    // cb_adf => first action is a greedy action TODO: is this a contract?
+    ec.pred.multiclass = data.adf_data.ecs[0]->pred.a_s[0].action + 1;
+  }
+}
+
+// float calc_loss(example& ec, CB::label& ld)
```
pmineiro, line 67: no more commented stuff!

vowpalwabbit/gen_cs_example.cc
```diff
@@ -41,7 +41,7 @@ void gen_cs_example_ips(multi_ex& examples, COST_SENSITIVE::label& cs_labels, fl
 
     COST_SENSITIVE::wclass wc = {0., i, 0., 0.};
     if (ld.costs.size() == 1 && ld.costs[0].cost != FLT_MAX)
-      wc.x = ld.costs[0].cost / safe_probability(std::max(ld.costs[0].probability, clip_p));
+    { wc.x = ld.costs[0].cost / safe_probability(std::max(ld.costs[0].probability, clip_p)); }
```
pmineiro, line 44: revert to be exactly the same as before

lalo: done

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  VW::finish_example(all, ec);
+}
+
+/*
+    Purpose: run before cb, cb_explore, cbify and cb_adf related reductions
+    This will 'translate' cb (non adf) input commands into their cb_adf counterparts
+    Except when:
+        - the model file loaded is from a version older or including 8.9.0
```
lalo, line 93: update vw version here

vowpalwabbit/cb_explore_adf_cover.cc
```diff
@@ -282,7 +282,10 @@ VW::LEARNER::base_learner* setup(config::options_i& options, vw& all)
   }
 
   // Set explore_type
-  const size_t problem_multiplier = cover_size + 1;
+  size_t problem_multiplier = cover_size + 1;
+
+  // Cover is using doubly robust without the cooperation of the base reduction
+  if (cb_type_enum == CB_TYPE_MTR) { problem_multiplier *= 2; }
```
jackgerrits, line 288: Did this come from somewhere?

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,200 @@
...
+void predict_or_learn(cb_to_cb_adf& data, multi_learner& base, example& ec)
+{
+  data.adf_data.copy_example_to_adf(*data.weights, ec);
+
+  auto ld = get_label(data, ec);
+  CB::label saved_ld;
+
+  if (ld != nullptr)
+  {
+    saved_ld = *ld;
```
jackgerrits, line 46: Should these be moves?

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,200 @@
...
+    *ld = ec.l.cb;
+  }
+
+  auto restore_guard = VW::scope_exit([&ld, &saved_ld] {
+    if (ld != nullptr) *ld = saved_ld;
+  });
+
+  if (!base.learn_returns_prediction || !is_learn) { base.predict(data.adf_data.ecs); }
+
+  if (is_learn) base.learn(data.adf_data.ecs);
```
jackgerrits, line 56:
```suggestion
  if (is_learn) { base.learn(data.adf_data.ecs); }
```

vowpalwabbit/cbify.cc
```diff
@@ -145,7 +147,13 @@ void cbify_adf_data::copy_example_to_adf(parameters& weights, example& ec)
...
     {
       for (feature_index& idx : fs.indicies)
-      { idx = ((((idx >> ss) * 28904713) + 4832917 * (uint64_t)a) << ss) & mask; }
+      {
+        auto rawidx = idx >> ss;
+        rawidx -= rawidx % (num_actions * increment);
+        rawidx += a * increment;
+        rawidx <<= ss;
+        idx = rawidx & mask;
+      }
```
jackgerrits, line 156: Can you explain this change?

jackgerrits (approved):

vowpalwabbit/cb_explore_adf_cover.cc
lalo: what do you mean with that question, can you rephrase?

jackgerrits: Sorry, it was unclear. This is a new line. Can you explain why it was added?

lalo: https://github.com/VowpalWabbit/vowpal_wabbit/blob/70d4863d382bf2dda53918db7e17b069498b86b4/vowpalwabbit/cb_explore_adf_cover.cc#L273

lalo: https://github.com/VowpalWabbit/vowpal_wabbit/blob/70d4863d382bf2dda53918db7e17b069498b86b4/vowpalwabbit/cb_adf.cc#L500

lalo: mtr is using DR for some policies. cb_adf does the correct thing on adjusting when using dr but at that point cb_adf registers as using "mtr" skipping the dr adjustment. This syncs that up. See both files linked above this comment for more context.

vowpalwabbit/learner.h
```diff
@@ -472,6 +473,30 @@ struct learner
     VW_WARNING_STATE_POP
   }
 
+  // never called, convienience method in case reduction has a seperate
```
olgavrou, line 476: never called?

vowpalwabbit/cb_explore_adf_bag.cc
```diff
@@ -135,6 +135,13 @@ void finish_bag_example(vw &all, cb_explore_adf_base<cb_explore_adf_bag> &data,
   cb_explore_adf_base<cb_explore_adf_bag>::finish_multiline_example(all, data, ec_seq);
 }
 
+void print_bag_example(vw& all, cb_explore_adf_base<cb_explore_adf_bag>& data, multi_ex& ec_seq)
+{
```
olgavrou, line 139: setting the prediction of `ec_seq[0]` seems important? Could the call to `print_bag_example` ever be omitted?

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -0,0 +1,200 @@
...
+
+  auto ld = get_label(c, ec);
+  if (ld != nullptr)
+  {
+    auto saved_ld = *ld;
+    *ld = ec.l.cb;
+
+    auto restore_guard = VW::scope_exit([&ld, &saved_ld] { *ld = saved_ld; });
+
+    c.adf_learner->print_example(all, c.adf_data.ecs);
```
olgavrou, line 78: is `adf_learner` used only to call `print_example`? (relevant I guess with my comment on bag print)

lalo: yes

vowpalwabbit/cb_explore_adf_bag.cc
lalo: omitted by whom? not sure i understand the phrasing of the question

vowpalwabbit/learner.h
lalo: good point, the comment might be nonsensical. It's a comment on how this method is never called by the VW machinery, some learner fns get called recursively, others just the top one, this one never gets called by it. It does get called by cb_to_cb_adf so yea, that's a bit weird.

vowpalwabbit/learner.h
```diff
@@ -535,6 +560,18 @@ struct learner
 
     return ret;
   }
+
+  base_learner* get_learner_by_name_prefix(std::string reduction_name)
```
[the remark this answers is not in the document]
lalo: Yes

vowpalwabbit/cb_explore_adf_bag.cc
olgavrou: as in is there a scenario where we might forget to set/call `print_bag_example` and then forget to set the prediction? I guess I am wondering if the prediction setting should be done in a print method

lalo: I see, if you compare finish_bag_example (same file) and print_bag_example, you'll see that they both mirror each other. Think of print as a finish without deleting. I can't call delete bc that multi_ex I'm sending down, is owned by cb_to_cb_adf and it is not owned by the vw driver / parser logic.

olgavrou (approved):

vowpalwabbit/cbify.cc
lalo: to avoid collisions between same name features in an example translated to multi_ex, each action is getting unique features even if they come from the same feature name (the input)

vowpalwabbit/cb_to_cb_adf.cc
```diff
@@ -22,33 +22,27 @@ struct cb_to_cb_adf
...
-
-  if (ld != nullptr)
+  CB::label ld;
+  uint32_t chosen_action = ec.l.cb.costs[0].action - 1;
+  bool is_test_label = CB::is_test_label(ec.l.cb);
+  if (!is_test_label)
   {
-    saved_ld = *ld;
-    *ld = ec.l.cb;
+    if (chosen_action < data.adf_data.num_actions) { ld = std::move(data.adf_data.ecs[chosen_action]->l.cb); }
```
lalo, line 35: chosen_action is input from user

test/train-sets/ref/cbe_cover_e_fixed_predict.stderr
```diff
@@ -1,23 +1,24 @@
...
 num sources = 1
-Enabled reductions: gd, scorer, csoaa, cb, cb_explore
+Enabled reductions: gd, scorer, csoaa_ldf, cb_adf, cb_explore_adf_cover, shared_feature_merger, cb_to_cbadf
 average  since         example        example  current  current  current
 loss     last          counter         weight    label  predict features
-0.000000 0.000000            1            1.0  1:0:0.5 2:0.750000        3
-0.250000 0.500000            2            2.0  2:1:0.5 1:0.750000        3
-0.375000 0.500000            4            4.0  1:2:0.5 2:0.750000        3
-0.437500 0.500000            8            8.0  2:1:0.5 1:0.750000        3
+0.000000 0.000000            1            1.0    known        1:0.75...        6
```
lalo, line 14: --cb doesnt have 0 as input

PR #2692, opened 2020-12-08
Allow port to be selected in spanning tree tool

jackgerrits:
In order to parallelize tests, they need to run on different ports. This allows us to supply the spanning tree tool with a custom port.

### DIFF
--- a/cluster/spanning_tree_main.cc
+++ b/cluster/spanning_tree_main.cc
@@ -10,8 +10,6 @@ This creates a binary tree topology over a set of n nodes that connect.
 #include "spanning_tree.h"
 #include "vw_exception.h"
 
-#include "options_boost_po.h"
-
 #ifdef _WIN32
 int daemon(int a, int b) { return 0; }
 int getpid() { return (int)::GetCurrentProcessId(); }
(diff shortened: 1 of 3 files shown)


cluster/spanning_tree_main.cc
```diff
@@ -17,33 +17,70 @@ int getpid() { return (int)::GetCurrentProcessId(); }
...
-#include <errno.h>
 
 using namespace VW;
 
+#include <boost/program_options.hpp>
+namespace po = boost::program_options;
+
+void usage(const po::options_description& desc)
+{
+  std::cout << "usage: spanning_tree [--port,-p number] [--nondaemon] [--help,-h] [pid_file]" << std::endl;
```
lalo, line 28: is the boost print help message easy to incorporate?

jackgerrits: Thats what line 29 is. It doesn't give you one in this form though and it is important to signal the final positional arg

cluster/spanning_tree_main.cc
```diff
@@ -17,33 +17,70 @@ int getpid() { return (int)::GetCurrentProcessId(); }
...
+  std::cout << desc << std::endl;
+}
+
 int main(int argc, char* argv[])
 {
-  if (argc > 2)
+  int port = 26543;
```
olgavrou, line 34: has this always been the default port?

jackgerrits: yep, its in the spanning tree constructor. its also hard coded into vw

cluster/spanning_tree_main.cc
```diff
@@ -17,33 +17,70 @@ int getpid() { return (int)::GetCurrentProcessId(); }
...
+  try
+  {
+    po::store(po::command_line_parser(argc, argv).options(all_options).positional(pos).run(), vm);
+    po::notify(vm);
+  }
+  catch (std::exception& e)
+  {
+    std::cout << argv[0] << ": " << e.what() << std::endl << std::endl;
+    usage(desc);
+    exit(1);
```
olgavrou, line 62: does this need to exit? can it `return 1`?

jackgerrits: Return makes sense

cluster/spanning_tree_main.cc
olgavrou: should it live as a constexpr somewhere then and be re-used?

jackgerrits: This binary does not depend on VW at all. Maybe in future. Common lib constants or something?

olgavrou: mmm ok I see

olgavrou (approved):

PR #2695, opened 2020-12-09
Use unique ports in each test

jackgerrits:


olgavrou (approved):

test/same-model-test.sh
```diff
@@ -9,15 +9,15 @@ MODEL_0=models/${TEST_NAME}.0.txt
 MODEL_1=models/${TEST_NAME}.1.txt
 
 # train on cluster mode
-../cluster/spanning_tree
+../cluster/spanning_tree -p 26544
```
lalo, line 12: move to variable

jackgerrits: Gotcha

PR #2696, opened 2020-12-09
Make flatbuffers optional

olgavrou:
Making flatbuffers optional. ON by default.

Will update documentation but when building via cmake can be turned off by setting `-DBUILD_FLATBUFFERS=OFF` and on windows via visual studio will need to remove the pre-processor definition `BUILD_FLATBUFFERS`


test/unit_test/CMakeLists.txt
```diff
@@ -35,13 +33,14 @@ add_executable(vw-unit-test.out
   test_common.cc
   test_common.h
   tokenize_tests.cc
```
olgavrou, line 35: the removed files here are duplicates

test/unit_test/unit_test.vcxproj.filters
```diff
@@ -81,9 +78,6 @@
     <ClCompile Include="options_boost_po_test.cc">
       <Filter>Source Files</Filter>
     </ClCompile>
-    <ClCompile Include="options_test.cc">
```
olgavrou, line 84: duplicate

CMakeLists.txt
```diff
@@ -197,7 +198,10 @@ endif()
 # Align and foreach are also required, for some reason they can't be specified as components though.
 find_package(Boost REQUIRED COMPONENTS program_options system)
 find_package(ZLIB REQUIRED)
-find_package(flatbuffers REQUIRED)
+if(BUILD_FLATBUFFERS)
+  find_package(flatbuffers REQUIRED)
+  add_definitions(-DBUILD_FLATBUFFERS)
```
jackgerrits, line 203: Use `target_compile_definitions` instead

test/unit_test/unit_test.vcxproj
```diff
@@ -35,7 +35,7 @@
     <ClCompile Include="error_test.cc" />
     <ClCompile Include="example_header_test.cc" />
     <ClCompile Include="explore_test.cc" />
-    <ClCompile Include="flatbuffer_parser_test.cc" />
+    <ClCompile Include="flatbuffer_parser_test.cc" Condition="Exists('$(VcpkgIntegration)')" />
```
jackgerrits, line 38: Is it possible to condition this on `BUILD_FLATBUFFERS` or some env var?

CMakeLists.txt
olgavrou: Done

test/unit_test/unit_test.vcxproj
olgavrou: I tried it out

CMakeLists.txt
```diff
@@ -250,7 +253,7 @@ add_subdirectory(cluster)
 add_subdirectory(library)
 add_subdirectory(vowpalwabbit)
 
-if (BUILD_FLATBUFFER_UTILS)
+if (BUILD_FLATBUFFER_UTILS AND BUILD_FLATBUFFERS)
```
jackgerrits, line 256: Should we collapse both of these options into just `BUILD_FLATBUFFERS`? The main reason we have the option is to control whether we require the flatbuffer dependency to be present or not. If it's present I feel like having the target present for fb utils is not an issue

jackgerrits (approved):

CMakeLists.txt
olgavrou: Yes good point, will do

PR #2698, opened 2020-12-09
Turn CCB sanity checks into reported errors

### LINKED ISSUE #2675: Execution should fail gracefully when # slots > # actions in CCB call in debug mode
https://github.com/VowpalWabbit/vowpal_wabbit/blob/3f5793d9920fdff44f674cd4252180411da1cff5/vowpalwabbit/conditional_contextual_bandit.cc#L113-L120

In debug mode, when the number of slots input to CCB exceeds the number of actions, execution logs the issue to `cerr` in and exits `learn_and_predict`. This results in an unallocated `action_ids` array later in [rl::fb_serializer](https://github.com/VowpalWabbit/reinforcement_learning/blob/82e74e122c6ff44046dbe1e94333894278738b19/rlclientlib/serialization/fb_serializer.h#L134), which causes a crash.

jackgerrits:
Fixes #2675


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -373,9 +348,23 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
...
 
-#ifndef NDEBUG
-  if (!sanity_checks<is_learn>(data)) { return; }
-#endif
+  if (data.slots.size() > data.actions.size())
+  {
+    std::stringstream msg;
+    msg << "ccb_adf_explore: badly formatted example - number of actions " << data.actions.size()
+        << " must be greater than the number of slots " << data.slots.size();
+    THROW(msg.str())
```
olgavrou, line 356: is throwing the right way to go here since we are trying to reduce throw's in vw? Should we just cleanup (if there is anything to cleanup) and return?

jackgerrits: The problem is that whoever is calling this expects predictions etc to be filled out and by just returning that is not the case. This was an issue reported by a user that they were hitting in the wild. Error code would be better so we aren't adding a throw here but we don't have the infra in place to do it yet.

olgavrou: The problem was only in debug mode, which means that we are were doing something different in release mode

jackgerrits: They would crash in release mode AFAIK so the old sanity check just did nothing to help them

olgavrou (approved):

PR #2702, opened 2020-12-09
[py] Run parser automatically based on str

### LINKED ISSUE #2694: Pyvw fails to run driver if data file given as command line
Pyvw should run the driver if a data file is given, but fails to do so if given as a command line like so:

```Python
from vowpalwabbit import pyvw
vw = pyvw.vw("--data ./test/train-sets/rcv1_small.dat")
```

The output shows it sees the file but the parser is not run.
```
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = ./test/train-sets/rcv1_small.dat
num sources = 1

finished run
number of examples = 0
weighted label sum = 0.000000
average loss = n.a.
total feature number = 0
```

Expected output is as follows:
```
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = rcv1_small.dat
num sources = 1
1.000000 1.000000            1            1.0  -1.0000   0.0000      128
0.893728 0.787455            2            2.0  -1.0000  -0.1126       44
0.905122 0.916517            4            4.0  -1.0000  -0.1701      190
0.924790 0.944458            8            8.0   1.0000  -0.0231       34
0.894742 0.864695           16           16.0   1.0000   0.0065       43
0.875196 0.855650           32           32.0  -1.0000   0.0423       47
0.838072 0.800947           64           64.0   1.0000   0.0619       54
0.754589 0.671106          128          128.0  -1.0000  -0.2779       67
0.662491 0.570393          256          256.0   1.0000   0.6543       86
0.564488 0.466484          512          512.0  -1.0000  -0.8828      104

finished run
number of examples = 1000
weighted example sum = 1000.000000
weighted label sum = -82.000000
average loss = 0.502732
best constant = -0.082000
best constant's loss = 0.993276
total feature number = 78739
```

This expected output can be achieved by:
```Python
from vowpalwabbit import pyvw
vw = pyvw.vw(data="./test/train-sets/rcv1_small.dat")
```

So in the Python layer the issue is it not checking the command line itself to determine whether to run the parser.
https://github.com/VowpalWabbit/vowpal_wabbit/blob/956a76fcf67563dd9bcd7c56fac1ef92a13c05be/python/vowpalwabbit/pyvw.py#L246

lalo:
Fixes #2694


python/tests/test_pyvw.py
```diff
@@ -352,6 +352,10 @@ def test_example_features():
     ex.push_namespace(ns2)
     assert ex.pop_namespace()
 
+def test_runparser_cmd_string():
+    vw = pyvw.vw("--data ./test/train-sets/rcv1_small.dat")
+    assert vw.parser_ran == True, "vw should set parser_ran to true if --data present"
```
jackgerrits, line 357: Can you add a test asserting that parser did not run if it wasn't present?

jackgerrits (approved):

python/vowpalwabbit/pyvw.py
```diff
@@ -241,10 +241,19 @@ def format_input(key, val):
...
+
         # check to see if native parser needs to run
         ext_file_args = ["d", "data", "passes"]
         if any(x in kw for x in ext_file_args):
             pylibvw.vw.run_parser(self)
+            self.parser_ran = True
+        elif arg_str:
+            # space after -d to avoid matching with other substrings
+            ext_file_cmd_str = ["-d ", "--data", "--passes"]
+            if [cmd for cmd in ext_file_cmd_str if(cmd in arg_str)]:
```
jackgerrits, line 254: should this be `any([cmd for cmd in ext_file_cmd_str if(cmd in arg_str)]` ?

olgavrou: Doesn't need to, the expression after `if` will create an array that contains the args that are present, so if any of them are present the array will not be empty, so we will enter the `if` statement once

olgavrou (approved):

PR #2703, opened 2020-12-09
Fix issue with index in DFtoVW.empty_col

etiennekintzler:
The conversion of pandas's DataFrame to VW format is **incorrect when the index of the supplied dataframe is different from the default one** (i.e different from [0, ..., n_rows]. 

The issue is that the `DFtoVW.empty_col` method always use the default index and thus the concatenation of the empty_col with the label (and/or features) column(s) fail. For example:

```python
import pandas as pd
empty_col = pd.Series(["", ""], index=[0, 1])
df = pd.Series(["a", "b"], index=[1, 2])

empty_col + df

# returns
# 0    NaN
# 1      a
# 2    NaN
# dtype: object
```

The fix is simply to use the DataFrame index in the `DFtoVW.empty_col` method. I also add a test related to this issue.

For those using the last release, the workaround is to supply a dataframe with default index using `df.reset_index(drop=True)`.


jackgerrits (approved):
Thanks for the fix!

PR #2705, opened 2020-12-09
make options a unique_ptr

### LINKED ISSUE #2317: Memory leak in vw_dll_parse_escaped unit test

olgavrou:
Resolves #2317


vowpalwabbit/global_data.h
```diff
@@ -382,8 +382,8 @@ struct vw
   vw_ostream trace_message;
 
   // Flag used when VW internally manages lifetime of options object.
-  bool should_delete_options = false;
-  VW::config::options_i* options;
+  bool should_release_options = false;
```
jackgerrits, line 385: This could probably be tracked at creation time of the unique_ptr with the deleter function?

olgavrou: Not sure what you mean. This is in case we want to delete the options pointer outside of `all` object, so all should just release it.

vowpalwabbit/global_data.cc
```diff
@@ -366,7 +366,7 @@ vw::~vw()
   }
 
   // Check if options object lifetime is managed internally.
-  if (should_delete_options) delete options;
+  if (should_release_options) options.release();
```
jackgerrits, line 369: This will leak I think? Release removes the ptr from being managed

vowpalwabbit/parse_args.cc
```diff
@@ -1672,9 +1674,19 @@ void free_args(int argc, char* argv[])
 }
 
 vw* initialize(
-    options_i& options, io_buf* model, bool skipModelLoad, trace_message_t trace_listener, void* trace_context)
+    config::options_i& options, io_buf* model, bool skipModelLoad, trace_message_t trace_listener, void* trace_context)
 {
-  vw& all = parse_args(options, trace_listener, trace_context);
+  std::unique_ptr<options_i> opts;
```
jackgerrits, line 1679: Since the options object is managed externally we can add a custom deleter which is a noop for the unique_ptr, rather than needing to manage the external state of `should_release_options`

olgavrou: ah good point

vowpalwabbit/global_data.h
jackgerrits: Added a comment at the spot where we'd need to do it.

Instead of using the `should_release_options` state, we can just make the deleter of the unique_ptr a noop in the cases where we would have called release. This also mitigates the case where an exception would have caused the object we're meant to release to be deleted.

jackgerrits (approved):

PR #2713, opened 2020-12-11
feat: [py] Add contextual bandit label to DFtoVW

etiennekintzler:
This PR implements the contextual bandit label class (see #2452)

## Usage

```python
import pandas as pd
from vowpalwabbit.DFtoVW import DFtoVW, Feature, ContextualBanditLabel

df_simple = pd.DataFrame({
        "a": [1, 2, 3],
        "c": [1.2, 2.3, 3.4],
        "p": [0.2, 0.8, .01],
        "x": [1, 1, 2],
})


conv = DFtoVW(df=df_simple, label=ContextualBanditLabel(["a", "c", "p"]), features=Feature("x"))
conv.convert_df()
```

returns 

```
[
 '1:1.2:0.2 | x:1', 
 '2:2.3:0.8 | x:1',
 '3:3.4:0.01 | x:2'
]
```

## Implementation details
- Multiple contextual bandit labels are also possible (e.g `ContextualBanditLabel([["a1", "c1", "p1"], ["a2", "c2", "p2"]])`)
- 5 tests have been added. I also reorganize the whole test file to sort the tests by label.
-  Value checking is made to be sure that action is an non negative integer and that the proba is between 0 and 1. For this last check an attribute `max_value` has been added.
- this new label can be used from the convenient method `DFtoVW.from_colnames`
- the class is called `ContextualBanditLabel` but it can be renamed `ContextualbanditLabel` if it seems more consistent with the other class names.

Also not directly related to this PR but doesn't justify a PR in iteself:
- I remove the type `str` in the `Returns` statement of the docstring for some classes (since it always returns a pd.Series, not a str).
- Any hashtable type is allowed as colname in the `DFtoVW.from_colnames` (instead of just string).


jackgerrits:
Rather than using an array, it seems like it might be clearer to use named params? What if someone provides 2 or 4 values in the array?
```Python
ContextualBanditLabel(["a", "c", "p"])
# Changes to:
ContextualBanditLabel(action_col="a", cost_col="c", prob_col="p")
```

etiennekintzler:
If  **just 1** CB label was to be passed, I agree that something like `ContextualBanditLabel(action="a", cost="c", prob="p")` would be the most forward way to do it. 

However if multiple CB labels need to be passed, the natural extension of this API feels a bit weird to me : `ContextualBanditLabel(action=["a1", "a2"], cost=["c1", "c2"], prob=["p1", "p2"])`.  What do you think @jackgerrits ?

Another way to make it more explicit (regarding the named params) would be to use dictionary as in : 

```python
ContextualBanditLabel({"action": "a", "cost": "c", "prob": "p"})
ContextualBanditLabel([{"action": "a1", "cost": "c1", "prob": "p1"), {"action": "a1", "cost": "c1", "prob": "p1")])
```

However that's more writing for the user (with the risk of wrongly named dict keys)...

lalo:
Another option would be the following 
```

labels.append(ContextualBanditLabel({"action": "a1", "cost": "c1", "prob": "p1"}))
labels.append(ContextualBanditLabel({"action": "a2", "cost": "c2", "prob": "p2"}))
conv = DFtoVW(df=df_simple, label=labels, features=Feature("x"))
```

etiennekintzler:
Hey @lalo ,

I think you forget to define `labels` as a list in your example (since you use `.append` method). If that's what you meant it will amount to:

```python
labels = [
    ContextualBanditLabel({"action": "a1", "cost": "c1", "prob": "p1"}),
    ContextualBanditLabel({"action": "a2", "cost": "c2", "prob": "p2"})
]
```

Is that right ?

----

I would tend to favour a more succinct formulation:

```python
ContextualBanditLabel([
    {"action": "a1", "cost": "c1", "prob": "p1"), 
    {"action": "a1", "cost": "c1", "prob": "p1")
])
```

or 

```python
ContextualBanditLabel(action=["a1", "a2"], cost=["c1", "c2"], prob=["p1", "p2"])
```

lalo:
Yes, correct I was implying a list. My reasoning with this more verbose proposal is more about what the type actually represents. So if you see a type/class called `ContextualBanditLabel` it sort of implies it has all the necessary info to contain a single label and if you want multiple ones you just use a container (a list in this case). The concise option implies that `ContextualBanditLabel` itself is a container containing multiple labels which by just naming convention is not communicated properly. This is a matter of taste/style, I tend to be more verbose.

I might be biased since I'm visualizing in the long term having this sort of classes for the regular vw python api, i.e. instantiating a label directly by stating

`ContextualBanditLabel({"action": 1, "cost": 2, "prob": 0.4})`

and this would keep compatibility with that. 

Let's see if @olgavrou or @jackgerrits chime in later, so don't rush in changing this impl before we gain consensus.

olgavrou:
I agree with @lalo here. `ContextualBanditLabel` seems to be representing one label, same as the other labels available so it would provide consistency from that aspect. If one object corresponds to one label it will ease the cognitive load when we add multi-line examples.

Also, when constructing examples, it seems to be clearer if we have a container (say array) of `ContextualBanditLabel`s and an array of features, that an index in the feature array would correspond to an index in the label array, instead of having it correspond to an internal index of an object's array. 

Regarding the name of the class I think `ContextualbanditLabel` is consistent with the internal naming (although I prefer looking at `ContextualBanditLabel` :) )

Thanks for adding all the extra tests and re-organising the test file!

etiennekintzler:
@lalo and @olgavrou I agree with you. It's also consistent with what we do with Feature that is  `[Feature("a"), "Feature("b")]` and not `Feature(["a", "b"])` which would be awkward.

One of the reason I wanted to use container inside the function is to be consistent with the API of Multilabels, but I can see how it's different in the case of ContextualBanditLabel (Multilabels can't have just one label can't it ?). 

Not directly related to this PR but there might be some confusion between the usage of `*` or `+` in https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Input-format ; shouldn't `*` mean '0 or more' and `+`  '1 or more' ? If it's the case, it should `(<action>:<cost>:<probability>)+` for CB label (and not `(<action>:<cost>:<probability>)*`), and same for  `Features=(String[:Value] )*` (unless VW accept namespace without feature).

olgavrou:
> @lalo and @olgavrou I agree with you. It's also consistent with what we do with Feature that is `[Feature("a"), "Feature("b")]` and not `Feature(["a", "b"])` which would be awkward.
> 
> One of the reason I wanted to use container inside the function is to be consistent with the API of Multilabels, but I can see how it's different in the case of ContextualBanditLabel (Multilabels can't have just one label can't it ?).

Yeah I can see that. A mutlilabel example can have one label, but it can have more than one. In the context of a dataframe it makes sense for the Multilabel label to be an array of "labels".
 
> Not directly related to this PR but there might be some confusion between the usage of `*` or `+` in https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Input-format ; shouldn't `*` mean '0 or more' and `+` '1 or more' ? If it's the case, it should `(<action>:<cost>:<probability>)+` for CB label (and not `(<action>:<cost>:<probability>)*`), and same for `Features=(String[:Value] )*` (unless VW accept namespace without feature).


You seem to be correct about our usage of `*` and `+` in the documentation. A vw example doesn't _need_ to have a label, hence the `*` in `(<action>:<cost>:<probability>)*`. Same for namespace. You can pass in a namespace without features and it will use the constant feature for that namespace

etiennekintzler:
> Yeah I can see that. A mutlilabel example can have one label, but it can have more than one. 

OK !

> In the context of a dataframe it makes sense for the Multilabel label to be an array of "labels".

Ok, however it's not what have been implemented so far :/ 

>  A vw example doesn't _need_ to have a label, hence the `*` in `(<action>:<cost>:<probability>)*`. 

Of course, but isn't it the case of _every_ label available (e.g `Simple`, `Multiclass`, `Multilabels` etc.) ? 

Yet these labels don't have any `*` in the current version.  If it's the case that every label is optional in a VW example, wouldn't it be more straightforward to mention it somewhere in https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Input-format#labels and drop the `*` in the labels definition (unless some parts of the label definition are _really_ optionnal) ?

olgavrou:
> Yet these labels don't have any `*` in the current version. If it's the case that every label is optional in a VW example, wouldn't it be more straightforward to mention it somewhere in https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Input-format#labels and drop the `*` in the labels definition (unless some parts of the label definition are _really_ optionnal) ?

The documentation says that `If the label is omitted, then no training will be performed with the corresponding example, although VW will still compute a prediction.` that is what I mean when I say that a label isn't strictly necessary. The `*` is there on labels where you can pass in more than one e.g. `action:cost:probability` tuple.

etiennekintzler:
@lalo @olgavrou I've been thinking about how to allow for list of labels in DFtoVW class and **it requires quite much changes in the code**. Indeed, the label attribute (in DFtoVW the class) is always assumed to be a label class (SimpleLabel, MulticlassLabel, etc.) and **not a list of label class**.

I see 2 possibilities : 

1. make the change in DFtoVW code to allow for a list of labels.
2. allow `ContextualBanditLabel` to have list for each of its parameter and parse the list of ContextualBanditLabel in `DFtoVW.__init__` accordingly

What do you think ?

---

To clarify point 2, let's say the following signature is allowed : ```ContextualBanditLabel(action=["a1", "a2"], cost=["c1", "c2"], proba=["p1", "p2"])```.

The user could provide the following list as the `label` in `DFtoVW` : 
```python
[ContextualBanditLabel(action="a1", cost="c1", proba="p1"), ContextualBanditLabel(action="a2", cost="c2", proba="p2")
```

Then the constructor `DFtoVW.__init__` would get back the attributes of each of the instance in the list and then use the following ContextualBanditLabel's constructor : 
```python
ContextualBanditLabel(action=["a1", "a2"], cost=["c1", "c2"], proba=["p1", "p2"])
```
Then the rest of the methods in `DFtoVW` can be called without change since now `self.label` is indeed a label class (and not a list of label).

lalo:
Ok, I understand. Is it feasible to rename one to ContextualBanditLabel and the other to ContextualBanditLabels/ContextualBanditLabelList and have the same step of adjusting List[ContextualBanditLabel] to the other type? This might make it easier to document or keep track of which is which instead of overloading the same class. What do you think @etiennekintzler ?

etiennekintzler:
@lalo thank you for your suggestion. 

Based on what you said I've thought of an alternative that doesn't require adding a new ContextualBanditLabel(s/List) class. 

What could be done instead is to delegate the parsing of the list of labels to a (private) class `_ListLabelManager`. This class would have the following properties:
- can be initialized with list of pre-defined label class such as `ContextualBanditLabel`. The constructor would also unpack from the list the elements required in the `DFtoVW` methods (for instance the columns)
- has a `process` method so it integrates seamlessly in `DFtoVW` class. This method would just iterate through the list and call the `process` method of each element of the list (that is `ContextualBanditLabel.process` in this example).

What do you think of this approach ?

etiennekintzler:
I end up creating a private helper class `_ListLabel` that is used to parse the list of labels that the user could provide to  `DFtoVW` class. For now, this class accepts the labels `ContextualbanditLabel` but also  `MultiLabel` (as @olgavrou suggested). Also, I am not sure if the old way of building list of MultiLabel (`MultiLabel(["y1", "y2"])` should be kept at all (I guess not).

### Usage

Below is an example of how the user could use a list of labels : 

```python
import pandas as pd
from vowpalwabbit.DFtoVW import DFtoVW, MultiLabel, ContextualbanditLabel, Feature

df = pd.DataFrame({
    'a1': [1, 2, 3], 'c1': [-3, -4, -5], 'p1': [1.0, 0.0, 0.5],
    'a2': [3, 2, 1], 'c2': [-1, -2, -3], 'p2': [0.75, 0.6, 0.2],
    'x': [2, 2.4, 2.6]
})

conv_cb = DFtoVW(df=df, label=[ContextualbanditLabel("a1", "c1", "p1"), ContextualbanditLabel("a2", "c2", "p2")], features=Feature("x"))
print(conv_cb.convert_df())
#['1:-3:1.0 3:-1:0.75 | x:2.0', '2:-4:0.0 2:-2:0.6 | x:2.4', '3:-5:0.5 1:-3:0.2 | x:2.6']

conv_multilabel = DFtoVW(df=df, label=[MultiLabel("a1"), MultiLabel("a2")],features=Feature("x"))
print(conv_multilabel.convert_df())
# ['1,3 | x:2.0', '2,2 | x:2.4', '3,1 | x:2.6']
```

### Implementation

Almost no changes have been made to the class `DFtoVW` as most of the logic in the `_ListLabel` class itself. I had to dynamically name the class so that the error triggered (if any) has the _class name of the objects in the list_ instead of _ListLabel, which the user doesn't know about. I also added tests that check if errors are correctly raised.

The PR contains some refactoring (in particular the method `DFtoVW.check_instance_columns`) and the file is formatted using black (with line length 100). I realized that doing the latter introduces quite much formatting modifications which make the review a bit more tedious (apologies for that).

python/vowpalwabbit/DFtoVW.py
```diff
@@ -513,6 +569,94 @@ def process(self):
...
+                "The only labels that can be used with list are {accepted}.".format(
+                    accepted=repr([cls.__name__ for cls in self.available_labels])[1:-1]
+                )
+            )
+
+        # Unpack columns
+        columns = set()
+        for label in label_list:
+            columns.update(label.columns)
+        self.columns = columns
```
etiennekintzler, line 612: This step is necessary since the attribute `columns` is used in `DFtoVW.check_missing_columns_df`

python/vowpalwabbit/DFtoVW.py
```diff
@@ -513,6 +569,94 @@ def process(self):
...
+
+        self.label_list = label_list
+
+        # Unpack attributes that are of type _Col
+        label_cols = []
+        for label_instance in label_list:
+            label_cols += [
+                value for key, value in vars(label_instance).items() if isinstance(value, _Col)
+            ]
+        self.label_cols = label_cols
```
etiennekintzler, line 622: Same here, the method `DFtoVW.check_instance_columns` run through the instance attributes and check types and values for attributes that are _Col or list of _Col. If no unpacking is made, this verification step would be skipped for _ListLabel

python/vowpalwabbit/DFtoVW.py
```diff
@@ -513,6 +569,94 @@ def process(self):
...
+
+        self.sep = self.sep_by_label[list(instance_classes)[0]]
+
+        # Dynamically change class_name
+        _ListLabel.__name__ = "List[{typename}]".format(typename=type(label_list[0]).__name__)
```
etiennekintzler, line 627: Necessary so that the errors raised has the class name of the instances in the list (instead of _ListLabel)

python/vowpalwabbit/DFtoVW.py
```diff
@@ -738,19 +876,21 @@ def check_label_type(self):
...
-            if not isinstance(self.label, available_labels):
-                raise TypeError(
-                    "Argument 'label' should be either of the following type: {label_types}.".format(
-                        label_types=repr(
-                            [x.__name__ for x in available_labels]
-                        )[1:-1]
-                    )
+        label = self.label[0] if isinstance(self.label, _ListLabel) else self.label
+
+        if not isinstance(label, available_labels):
```
etiennekintzler, line 888: When self.label is a _ListLabel, the underlying type checked is the one of the _instances_ in the list (by default the first one).

python/vowpalwabbit/DFtoVW.py
```diff
@@ -789,55 +926,48 @@ def check_missing_columns_df(self):
...
-            if len(value) > 0
+            key: sorted(list(value)) for (key, value) in missing_cols.items() if len(value) > 0
         }
-        self.generate_missing_col_error(missing_cols)
+        if missing_cols:
+            self.raise_missing_col_error(missing_cols)
 
-    def generate_missing_col_error(self, absent_cols_dict):
-        """Generate error if some columns are missing
+    def raise_missing_col_error(self, missing_cols_dict):
```
etiennekintzler, line 942: use **raise** instead of **generate**. Also this function is now called only if there are missing columns (see method `check_missing_columns_df` ).

PR #2715, opened 2020-12-11
Runtests driver in Python

jackgerrits:
- Created new Python driver for testing
- Consumes @lalo's RunTests parser work
- Runs tests in parallel and provides colorized, easier to understand output
- Added to run in Linux and MacOS CI

## TODO in another PR
- `-O` flag not yet supported
- `-V` flag not yet supported

## Args
```
usage: run_tests.py [-h] [-t TEST [TEST ...]] [-E EPSILON] [-e] [-o] [-f] [--ignore_dirty] [--working_dir WORKING_DIR]
                    [--ref_dir REF_DIR] [-j JOBS] [--vw_bin_path VW_BIN_PATH] [--spanning_tree_bin_path SPANNING_TREE_BIN_PATH]
                    [--test_spec TEST_SPEC]

optional arguments:
  -h, --help            show this help message and exit
  -t TEST [TEST ...], --test TEST [TEST ...]
                        Run specific tests and ignore all others (default: None)
  -E EPSILON, --epsilon EPSILON
                        Tolerance used when comparing floats. Only used if --fuzzy_compare is also supplied (default: 0.001)
  -e, --exit_first_fail
                        If supplied, will exit after the first failure (default: False)
  -o, --overwrite       If test output differs from the reference file, overwrite the contents (default: False)
  -f, --fuzzy_compare   Allow for some tolerance when comparing floats (default: False)
  --ignore_dirty        The test ref dir is checked for dirty files which may cause false negatives. Pass this flag to skip this check.
                        (default: False)
  --working_dir WORKING_DIR
                        Directory to save test outputs to (default: /Users/jagerrit/.vw_runtests_working_dir)
  --ref_dir REF_DIR     Directory to read test input files from (default: /Users/jagerrit/w/repos/vowpal_wabbit/test)
  -j JOBS, --jobs JOBS  Number of tests to run in parallel (default: 4)
  --vw_bin_path VW_BIN_PATH
                        Specify VW binary to use. Otherwise, binary will be searched for in build directory (default: None)
  --spanning_tree_bin_path SPANNING_TREE_BIN_PATH
                        Specify spanning tree binary to use. Otherwise, binary will be searched for in build directory (default: None)
  --test_spec TEST_SPEC
                        Optional. If passed the given JSON test spec will be used, otherwise a test spec will be autogenerated from the
                        RunTests test definitions (default: None)
```


lalo:
is easy to add a flag to disable the color print? in case somewhere or some terminal doesnt support it (and avoid printing junk)?

test/run_tests.py
```diff
@@ -0,0 +1,642 @@
...
+    FAIL = 2
+    SKIPPED = 3
+
+
+def try_decode(binary_object):
+    return binary_object.decode("utf-8") if binary_object is not None else ""
+
+
+# Returns true if they are close enough to be considered equal.
+def fuzzy_float_compare(float_one, float_two, epsilon):
```
lalo, line 44: you should organize everything related to diffing in its seperate file

.scripts/macos/test.sh
```diff
@@ -2,10 +2,12 @@
...
-# NUM_PROCESSORS=$(cat nprocs.txt)
-make help
-make test_with_output -j ${NUM_PROCESSORS}
\ No newline at end of file
+cd test
+NUM_PROCESSORS=$(cat ../build/nprocs.txt)
+python3 run_tests.py -f -j ${NUM_PROCESSORS}
+
+cd ../build
+make test_with_output
```
lalo, line 13: why did you drop the -j here

lalo:
I think the best way to test this is to just merge it and run it on CI along side the current perl script. Just don't disable the old way until we have a good chunk of time running both.

.scripts/macos/test.sh
jackgerrits: `NUM_PROCESSORS` was undefined before resulting in `-j` with no max jobs, which you never really want. Besides it is redundant because the tests are all built at this point and the test running component which this is driving can't be parallelized.

jackgerrits:
100% agree, that's why I added it to Linux and MacOS CI in this PR. Let's run it side by side and see how it goes.

test/run_tests.py
```diff
@@ -0,0 +1,662 @@
...
+        test_dest_file = Path(test_working_dir).joinpath(file)
+        Path(test_dest_file.parent).mkdir(parents=True, exist_ok=True)
+        # We always want to replace this file in case it is the output of another test
+        if test_dest_file.exists():
+            test_dest_file.unlink()
+        shutil.copyfile(str(file_to_copy), str(test_dest_file))
+    return test_working_dir
+
+
+def main():
```
lalo, line 390: would be nice if you could split main in to functions to have some more semantics i.e.

setup_arguments()
find_vw_binary()
...

lalo (approved):

PR #2716, opened 2020-12-13
Remove unused label_size from label_parser struct

jackgerrits:
`label_size` has been unused in `label_parser` for quite some time. This change removes the fields and deprecates the public functions that used it. These functions have also ignored the parameter for some time too.


vowpalwabbit/example.cc
```diff
@@ -401,10 +405,13 @@ std::string prob_dist_pred_to_string(const example& ec)
...
 namespace VW
 {
-example* alloc_examples(size_t, size_t count = 1)
+example* alloc_examples(size_t count = 1)
 {
   example* ec = calloc_or_throw<example>(count);
   if (ec == nullptr) return nullptr;
+
+  new (ec) example;
+
```
olgavrou, line 414: why is this `new` needed here? Shouldn't the example be allocated on line 408?

jackgerrits: We were never actually running the constructor before, I figured I should fix this up here. Since it isn't exactly related to the change, I can split out into an independent change?

olgavrou (approved):

PR #2720, opened 2020-12-16
Vcpkg do not link everything

olgavrou:


ataymano (approved):

jackgerrits (approved):
Should this be added to cmake too? As in will this affect the cmake generated solution?

olgavrou:
Updated to check of vcpkg has `VcpkgAutoLink` variable, if not then we are running an old version of vcpkg and the build will with a message.

@jackgerrits I tried it out with cmake and the flag gets ignored, not sure how to fix that, will take a look

PR #2722, opened 2020-12-16
Rcv1 benchmarks

jackgerrits:


test/benchmarks/rcv1_benchmarks.cc
```diff
@@ -0,0 +1,278 @@
...
+-1 |f 10:7.0070818e-02 13:3.0282617e-02 24:1.5686616e-02 26:4.1215312e-02 42:6.1233964e-02 55:6.4943276e-02 74:4.0680200e-02 90:8.7173775e-02 98:4.9359940e-02 125:1.0314921e-01 131:5.5968020e-02 133:5.5713881e-02 147:4.0348452e-02 152:5.8046080e-02 161:3.9000869e-02 166:1.0284418e-01 179:3.4781855e-02 188:2.1661038e-02 189:4.0627711e-02 210:4.2160925e-02 219:8.0608368e-02 228:6.6707455e-02 229:3.3767268e-02 231:4.2642169e-02 240:3.6020506e-02 25
```
olgavrou, line 265: does the memory allocation need to be in the state loop?

olgavrou (approved):

test/benchmarks/rcv1_benchmarks.cc
jackgerrits: This needs to be done for each instance of vw, doing it outside is not really possible. This benchmark's goal is to represent an end to end run with the disk IO removed

PR #2730, opened 2020-12-18
[cbify] Fix dangling pointer in example due prediction array caching.

kumpera:
When memory reuse is done, we must ensure the invariant at any time, there's
only one valid "owning" pointer to that memory.

Owning, in this case, is a bit loose and doesn't mean smart pointer. It's
any pointer that could possibly be used to free the underlying memory.

In this case, cbify doesn't zero `example::pred` after it reclaims that v_array.


jackgerrits (approved):

vowpalwabbit/cbify.cc
```diff
@@ -45,6 +45,13 @@ struct cbify_reg
   float max_cost = std::numeric_limits<float>::lowest();
 };
 
+template<typename T>
+void v_move(v_array<T> &dst, v_array<T> &src)
+{
+  dst = src;
+  src = v_init<T>();
```
lalo, line 52: does v_init alloc?

jackgerrits: Nope, just sets all four fields to 0

Issue #2731, opened 2020-12-21
Ignoring duplicated interactions is causing problems


ataymano:
#### Describe the bug
[UPD] More accurate description in first comment

Ignoring duplicated interaction leads to dropped previous option

#### To Reproduce
vw --cb_explore_adf --epsilon 0.2 --dsjson --cb_type ips --cubic ::: -d cb_as_ccb.json -f mymodel.bin   - Model is saved
vw --cb_explore_adf --epsilon 0.2 --dsjson --cb_type ips --cubic ::: -d cb_as_ccb.json -i mymodel.bin   - Crash

#### What is happening:
Here:
https://github.com/VowpalWabbit/vowpal_wabbit/blob/6bf73450c1446b3829feb5cb0d2d8936e26c3e8b/vowpalwabbit/parse_args.cc#L1499
we are ignoring duplicated interaction, while if previous option was key-only we still have it in saved_key variable and it supposed to be saved later, so it got dropped from saving:
https://github.com/VowpalWabbit/vowpal_wabbit/blob/6bf73450c1446b3829feb5cb0d2d8936e26c3e8b/vowpalwabbit/parse_args.cc#L1512
In --cb_explore_adf -q:: kind of command lines dropped guy is csoaa_rank, so predictions got messed up here:
https://github.com/VowpalWabbit/vowpal_wabbit/blob/6bf73450c1446b3829feb5cb0d2d8936e26c3e8b/vowpalwabbit/csoaa.cc#L449

#### Environment
latest VW binary on Windows

ataymano:
During fixing the problem realized that original description is not 100% accurate.

What happens:
1) When we parsing header of input model, following issues occur
     a) If option immediately before that is key-only, it got dropped
     b) if there are arguments after that, then empty option ('--') is inserted before that.

So, if we have ['--cb_explore_adf',  '--csoaa_rank', '-q', '::', '--epsilon', '0.2'] written in model header,
options that we get out of here (https://github.com/VowpalWabbit/vowpal_wabbit/blob/6bf73450c1446b3829feb5cb0d2d8936e26c3e8b/vowpalwabbit/parse_args.cc#L1450) are the following:
['--cb_explore_adf', '-q', '::', '--', '--epsilon', '0.2']

2) we are adding csoaa_rank during cb_adf call.
[...]

PR #2732, opened 2020-12-21
[draft] Ignoring duplicated interactions issue fix

ataymano:
No fix yet - checking if test is failing on CI


ataymano:
#2731

ataymano:
preparing better patch with more tests

PR #2738, opened 2020-12-23
Add required new option type for constrained option input types

jackgerrits:
Adds `int64_t`, `std::vector<uint64_t>`, `std::vector<int64_t>`


vowpalwabbit/options_types.h
```diff
@@ -30,9 +30,9 @@ struct typelist
 
 // is_scalar detection relies on only accepting std::vector<T> for vector types. If that changes, update
 // is_scalar_impl below.
-using supported_options_types =
-    typelist<unsigned int, int, size_t, uint64_t, float, double, char, std::string, bool, std::vector<int>,
-        std::vector<size_t>, std::vector<float>, std::vector<double>, std::vector<char>, std::vector<std::string>>;
+using supported_options_types = typelist<unsigned int, int, size_t, uint64_t, int64_t, float, double, char, std::string,
+    bool, std::vector<int>, std::vector<uint64_t>, std::vector<int64_t>, std::vector<size_t>, std::vector<float>,
```
olgavrou, line 34: do we also want `std::vector<unsigned int>`?

olgavrou (approved):

vowpalwabbit/options_types.h
jackgerrits: unsigned is already covered by `std::vector<uint64_t>`. Once I've gone through and aligned usage to either `int64_t` or `uint64_t`, then the `unsigned int`, `int` and `size_t` variants will be removed.

The diff is unclear but all this does is add `int64_t`, `std::vector<uint64_t>`, `std::vector<int64_t>`

vowpalwabbit/options_types.h
```diff
@@ -30,9 +30,9 @@ struct typelist
 
 // is_scalar detection relies on only accepting std::vector<T> for vector types. If that changes, update
 // is_scalar_impl below.
-using supported_options_types =
-    typelist<unsigned int, int, size_t, uint64_t, float, double, char, std::string, bool, std::vector<int>,
-        std::vector<size_t>, std::vector<float>, std::vector<double>, std::vector<char>, std::vector<std::string>>;
+using supported_options_types = typelist<unsigned int, int, size_t, uint64_t, int64_t, float, double, char, std::string,
```
lalo, line 33: what is the purpose of adding these 3 new types? or why would be a better question

lalo: oh, I just noticed the other prs that depend on this one

jackgerrits: Yep, the plan is to constrain options to only a few essential types compared to the several duplicate types there are now

PR #2741, opened 2020-12-23
Full RunTests for flatbuffers

olgavrou:
Using the recent python runtest work.

The goal is to run all of the tests in RunTests with `--flatbuffer`

running `python3 run_tests.py -f --for_flatbuffers` will first transform all of the input files into flatbuffer ones and then run the RunTest using the new flatbuffer input


test/run_tests.py
```diff
@@ -476,6 +510,46 @@ def get_deps(test_number, tests):
...
+        if 'dictionary' in test['vw_command']:
+            print("{}Skipping test {} for flatbuffers, currently dictionaries are not supported{}".format(color_enum.LIGHT_CYAN, test_id, color_enum.ENDC))
+            continue
+        if 'invert_hash' in test['vw_command']:
+            print("{}Skipping test {} for flatbuffers, invert_hash not supported on converted files{}".format(color_enum.LIGHT_CYAN, test_id, color_enum.ENDC))
+            continue
+        if 'audit' in test['vw_command']:
+            print("{}Skipping test {} for flatbuffers, audit not supported{}".format(color_enum.LIGHT_CYAN, test_id, color_enum.ENDC))
+            continue
+        if 'dsjson' in test['vw_command']:
```
olgavrou, line 541: `cats` and `dsjson` will not need to be skipped once the 2 related PR's (linked in description) are merged

test/run_tests.py
```diff
@@ -476,6 +510,46 @@ def get_deps(test_number, tests):
...
+            print("{}Skipping test {} for flatbuffers, dsjson coming soon{}".format(color_enum.LIGHT_CYAN, test_id, color_enum.ENDC))
+            continue
+        if 'cats' in test['vw_command']:
+            print("{}Skipping test {} for flatbuffers, continuous actions coming soon{}".format(color_enum.LIGHT_CYAN, test_id, color_enum.ENDC))
+            continue
+
+        fb_test_converter = fb_converter.FlatbufferTest(test, working_dir)
+        fb_test_converter.to_flatbuffer(to_flatbuff, color_enum)
```
lalo, line 549: if you want to avoid pushing color_enum around you can define it before importing runtests_flatbuffer_converter and then you can import it as usual on runtests_flatbuffer_converter.py. Another option would be to have a runtest_util.py but that might be overkill. Another option would be to move the import of runtests_flatbuffer_converter either to line 514 or 547 for it to be imported right where it is needed (I prefer this one, but its a matter of style/taste).

lalo (commented):
still haven't finished reviewing but left a comment for now

cs/unittest/RunTests.tt
```diff
@@ -31,8 +31,8 @@ var skipList = new[] { 13, 32, 39, 40, 41, 59, 60, 61, 66, 68, 90,
     220, // deamon mode shell script test not supported
```
lalo, line 31: on line 27: you want to 260 to become 258 based on your changes on Runtests

test/RunTests
```diff
@@ -1436,7 +1436,7 @@ printf '3 |f a b c |e x y z\n2 |f a y c |e x\n' | \
     pred-sets/ref/cb_eval1.preds
 
 # Test 124: cb_eval save/load #2
-{VW} -d train-sets/cb_eval2 -i mwt.model -p cb_eval2.preds
+{VW} -d train-sets/cb_eval2 -i mwt.model -p cb_eval2.preds --multiworld_test f
```
lalo, line 1439: why is this necessary?

test/RunTests
```diff
@@ -1492,7 +1492,7 @@ printf '3 |f a b c |e x y z\n2 |f a y c |e x\n' | \
     train-sets/ref/cb_adf_crash1.stderr
 
 # Test 136: cb explore adf predict
-{VW} -d train-sets/cb_adf_crash_2.data -i models/cb_adf_crash.model -t
+{VW} -d train-sets/cb_adf_crash_2.data -i models/cb_adf_crash.model -t --cb_explore_adf
```
lalo, line 1495: applies here too

test/run_tests.py
```diff
@@ -452,11 +475,22 @@ def do_dirty_check(test_base_ref_dir):
...
+def clean_dirty(test_base_ref_dir):
+    result = subprocess.run(
+        "git clean -f -d -x -e __pycache__".split(),
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        cwd=test_base_ref_dir,
+        timeout=10)
+    return_code = result.returncode
+    if return_code != 0:
+        print("Failed to run 'git clean -f -d -x -e __pycache__'")
```
lalo, line 492: move this string to a var since its being used twice

test/run_tests.py
```diff
@@ -452,11 +475,22 @@ def do_dirty_check(test_base_ref_dir):
...
         sys.exit(1)
 
+def clean_dirty(test_base_ref_dir):
+    result = subprocess.run(
+        "git clean -f -d -x -e __pycache__".split(),
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        cwd=test_base_ref_dir,
+        timeout=10)
+    return_code = result.returncode
```
lalo, line 490: merge with if to avoid variable

cs/unittest/RunTests.tt
olgavrou: Thanks

test/run_tests.py
olgavrou: `color_enum` is dependent on input arguments (it is set to NoColor or Color depending on args.no_color) and is set in `main`. I could still define it somewhere else but if I don't pass it as a var it's value will not change with the input arg preference

test/run_tests.py
olgavrou: done

test/run_tests.py
olgavrou: done

test/RunTests
lalo: not a fan of changing the test, since the thing being tested is changed a bit. Would it be reasonable to infer that info from the test it depends on? see: https://github.com/VowpalWabbit/vowpal_wabbit/blob/07961e3674648a66dd048191be5e8dbec0f6c31d/test/runtests.AUTOGEN.json#L1524

olgavrou: Yeah good point, trying it out

olgavrou: Restored tests. To get around it I am stashing the original vw command in each test object, and I am having tests use the `depends_on` arguments for file conversions

test/run_tests.py
```diff
@@ -452,11 +475,24 @@ def do_dirty_check(test_base_ref_dir):
...
     stdout = try_decode(result.stdout)
     if len(stdout) != 0:
-        print("Error: Test dir is not clean, this can result in false negatives. To ignore this and continue anyway pass --ignore_dirty")
+        print("Error: Test dir is not clean, this can result in false negatives. To ignore this and continue anyway pass --ignore_dirty or pass --clean_dirty to clean")
         print("'git clean --dry-run -d -x -e __pycache__' output:\n---")
         print(stdout)
         sys.exit(1)
 
+def clean_dirty(test_base_ref_dir):
+    git_command = "git clean -f -d -x -e __pycache__"
```
jackgerrits, line 484: Can you use the long form of the args?

test/run_tests.py
```diff
@@ -452,11 +475,24 @@ def do_dirty_check(test_base_ref_dir):
...
+        git_command.split(),
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        cwd=test_base_ref_dir,
+        timeout=10)
+
+    if result.returncode != 0:
+        print("Failed to run {}".format(git_command))
+
+    stdout = try_decode(result.stdout)
```
jackgerrits, line 495: This looks unused?

test/run_tests.py
lalo: you could also use `git clean -fedex`

jackgerrits (approved):
run_tests.py is coming to life!!

test/run_tests.py
olgavrou: removed

test/run_tests.py
olgavrou: used `--force` and `--exclude`

olgavrou:
> run_tests.py is coming to life!!

Yes this would be extremely difficult without all the python support, thanks guys :)

PR #2744, opened 2020-12-24
Add python bindings for pdf prediction type

ajay0:
The pdf prediction type (`VW::continuous_actions::probability_density_function`) is the one used by `cats_pdf`.

It is also the prediction type used by #2410 and thus having Python bindings would be helpful.

### DIFF
--- a/python/pylibvw.cc
+++ b/python/pylibvw.cc
@@ -53,6 +53,7 @@ const size_t pPROB = 6;
 const size_t pMULTICLASSPROBS = 7;
 const size_t pDECISION_SCORES = 8;
 const size_t pACTION_PDF_VALUE = 9;
+const size_t pPDF = 10;
 
 void dont_delete_me(void* arg) {}
 
@@ -186,6 +187,8 @@ size_t my_get_prediction_type(vw_ptr all)
       return pDECISION_SCORES;
     case prediction_type_t::action_pdf_value:
       return pACTION_PDF_VALUE;
+    case prediction_type_t::pdf:
+      return pPDF;
     default:
       THROW("unsupported prediction type used");
   }
@@ -559,6 +562,13 @@ py::list ex_get_decision_scores(example_ptr ec)
 py::tuple ex_get_action_pdf_value(example_ptr ec)
 {
   return py::make_tuple(ec->pred.pdf_value.action, ec->pred.pdf_value.pdf_value);
+}
+
+py::list ex_get_pdf(example_ptr ec)
+{
+  py::list values;
+  for (auto const& segment : ec->pred.pdf) { values.append(py::make_tuple(segment.left, segment.right, segment.pdf_value)); }
+  return values;
 }
 
 py::list ex_get_multilabel_predictions(example_ptr ec)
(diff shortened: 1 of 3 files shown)


python/pylibvw.cc
```diff
@@ -53,6 +53,7 @@ const size_t pPROB = 6;
 const size_t pMULTICLASSPROBS = 7;
 const size_t pDECISION_SCORES = 8;
 const size_t pACTION_PDF_VALUE = 9;
+const size_t pPDF = 10;
```
lalo, line 56: For completeness, can you add a test exercising this? maybe under test_cats.py

jackgerrits (approved):
Looks good. I agree with @lalo, once there's a test should be good to go.

ajay0:
Added a pytest, let me know if that works.

python/tests/test_cats.py
```diff
@@ -17,3 +17,23 @@ def test_cats():
...
+    vw = pyvw.vw("--cats_pdf 4 --min_value " + str(min_value) + " --max_value " + str(max_value) + " --bandwidth 1")
+    vw_example = vw.parse("ca 15:0.657567:6.20426e-05 | f1 f2 f3 f4", pyvw.vw.lContinuous)
+    vw.learn(vw_example)
+    vw.finish_example(vw_example)
+
+    assert vw.get_prediction_type() == vw.pPDF, "prediction_type should be pdf"
+
+    pdf_segments = vw.predict("| f1 f2 f3 f4")
+    for segment in pdf_segments:
+        assert len(segment) == 3
```
jackgerrits, line 34: What is the 3rd segment? Should it be asserted too?

python/tests/test_cats.py
```diff
@@ -17,3 +17,23 @@ def test_cats():
     assert action >= 10
     assert action <= 20
     vw.finish()
+
+def test_cats_pdf():
+    min_value = 10
+    max_value = 20
```
jackgerrits, line 23: Can you add a comment explaining how you chose these values?

python/tests/test_cats.py
ajay0: The 3rd value in a segment is the [pdf_value](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/CATS,-CATS-pdf-for-Continuous-Actions#prediction-type) and the only constraint on it is that is should be non-negative. 

I wasn't sure what level of assertions I should add. For instance, I could have asserted the exact segment values themselves. I chose to follow what the existing `test_cats` function in the same file did -- only test if the action is in the expected range.

python/tests/test_cats.py
ajay0: I used the same setup used by the existing `test_cats`  function in the file but changed `--cats` to `--cats_pdf`. But I'm not sure how those values were picked.

olgavrou: It doesn't really matter what these values are, they were picked randomly in the existing cats test. Cats needs min and max to determine the action range and then we can use them to test the prediction is not out of the supplied range.

python/tests/test_cats.py
olgavrou: One test we could add here would be to check that the returned pdf is valid, see [here](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/prob_dist_cont.cc#L51)

olgavrou (approved):

python/tests/test_cats.py
ajay0: Makes sense. Added an assertion for checking pdf mass ~= 1. (while at it, added one for pdf value being non-negative)

jackgerrits (approved):

PR #2746, opened 2020-12-28
Duplicated interactions issue mitigation

ataymano:
Better to do more solid refactoring, but this is straightforward mitigation


ataymano:
#2731

vowpalwabbit/parse_args.cc
```diff
@@ -1447,21 +1446,12 @@ bool check_interaction_settings_collision(options_i& options, std::string file_o
   return file_options_has_interaction;
 }
 
-options_i& load_header_merge_options(options_i& options, vw& all, io_buf& model)
+void merge_options_from_header_strings(const std::vector<std::string>& strings, bool skip_interactions, VW::config::options_i& options)
```
lalo, line 1449: Going off on your description of this pr about a possible refactor: this function screams to be moved to options related files, you can push back on this in case you want to keep the pr very contained but it would be ideal to have

`options_i` interface `merge(...)`
and then copy this function to `options_boost` as an impl of merge(...)

or maybe a more saner approach would be to have separate options instances

1) model_options = options(args from model file)
2) cmdline_options = options(args from command line)

and then just impl a a merge method on options and merge both of them. Options would just simply parse the string and separate into string keys and string values. Then you just manipulate those dictionaries of key, values when merging.

ataymano: yeah that's a good idea - let me check if something like that is doable in contained way.

lalo: i would expect it wouldn't be that easy, since right now all the parsing occurs on add_and_parse and this would effectively shift the string only parsing on the constructor of options.

Not sure if the benefit is big in just moving this function over to option land to have all the low level parsing of strings under options.

jackgerrits: Yeah an abstract `merge` operations would really help clean this code up

jackgerrits (approved):

jackgerrits:
This is good reasoning to move to something like JSON for options in model files.

vowpalwabbit/parse_args.cc
lalo: do we even need to merge? why not have both input strings inside options (one string for cmd line, one string from model) and then just change add_and_parse to parse first from one and then parse from another and then it should just work (hopefully).

jackgerrits: What happens when we need a third? Like an options file?

lalo: the logic of precedence has to be baked in somewhere, so you would have to explicitly enumerate that third place wherever you want it to i.e. model first -> options file -> cmd line

vowpalwabbit/parse_args.cc
```diff
@@ -1538,18 +1529,32 @@ options_i& load_header_merge_options(options_i& options, vw& all, io_buf& model)
...
 
   if (count == 0 && saved_key != "") { options.insert(saved_key, ""); }
+}
+
+options_i& load_header_merge_options(options_i& options, vw& all, io_buf& model, bool& interactions_settings_doubled)
+{
+  std::string file_options;
+  save_load_header(all, model, true, false, file_options, options);
+
+  interactions_settings_doubled = check_interaction_settings_collision(options, file_options);
```
olgavrou, line 1539: doubled -> duplicated?

doubled and interactions makes me think that we want to double the interactions for some reason

olgavrou (approved):

vowpalwabbit/parse_args.cc
ataymano: fixed

PR #2750, opened 2020-12-30
Fix missing progressive validation header

lalo:
At least when using the python vw api, the progressive validation header does not print. This will fix it.

### DIFF
--- a/cs/unittest/TestTracing.cs
+++ b/cs/unittest/TestTracing.cs
@@ -30,7 +30,7 @@ namespace cs_unittest
             var trace = string.Join("\n", messages);
             var expected = 19;
             Assert.AreEqual(expected, messages.Count, $"Expected {expected} lines. Found {messages.Count}. '{trace}'");
-            Assert.AreEqual("total feature number = 4", messages[15]);
+            Assert.AreEqual("total feature number = 4", messages[18]);
         }
     }
 }
--- a/cs/unittest/TestTracing.cs
+++ b/cs/unittest/TestTracing.cs
@@ -28,7 +28,7 @@ namespace cs_unittest
             }
 
             var trace = string.Join("\n", messages);
-            var expected = 19
+            var expected = 19;
             Assert.AreEqual(expected, messages.Count, $"Expected {expected} lines. Found {messages.Count}. '{trace}'");
             Assert.AreEqual("total feature number = 4", messages[15]);
         }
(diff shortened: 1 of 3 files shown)


vowpalwabbit/parse_args.cc
```diff
@@ -1688,6 +1688,38 @@ void free_args(int argc, char* argv[])
   free(argv);
 }
 
+void print_progressive_validation_header(vw& all)
```
jackgerrits, line 1691: Can this be const?

lalo: don't think so since we're using trace_message

cs/unittest/TestTracing.cs
```diff
@@ -28,7 +28,8 @@ public void TestTraceListener()
             }
 
             var trace = string.Join("\n", messages);
-            Assert.AreEqual(16, messages.Count, $"Expected 16 lines. Found {messages.Count}. '{trace}'");
+            var expected = 19
```
lalo, line 31:
```suggestion
            var expected = 19;
```

cs/unittest/TestTracing.cs
```diff
@@ -28,7 +28,8 @@ public void TestTraceListener()
             }
 
             var trace = string.Join("\n", messages);
-            Assert.AreEqual(16, messages.Count, $"Expected 16 lines. Found {messages.Count}. '{trace}'");
+            var expected = 19;
+            Assert.AreEqual(expected, messages.Count, $"Expected {expected} lines. Found {messages.Count}. '{trace}'");
             Assert.AreEqual("total feature number = 4", messages[15]);
```
lalo, line 33:
```suggestion
            Assert.AreEqual("total feature number = 4", messages[18]);
```

cs/unittest/TestTracing.cs
```diff
@@ -28,8 +28,9 @@ public void TestTraceListener()
             }
 
             var trace = string.Join("\n", messages);
-            Assert.AreEqual(16, messages.Count, $"Expected 16 lines. Found {messages.Count}. '{trace}'");
-            Assert.AreEqual("total feature number = 4", messages[15]);
+            var expected = 19;
```
lalo, line 31: @lokitoth would this break something on the c# world?

lokitoth: I do not believe so. While there may be someone depending on the specific exact output, it seems unreasonable to expect there to never be new trace messages as the version increments.

It does not seem reasonable for consumers of this API to assume that the contents of the trace logging are an API contract.

jackgerrits (approved):

PR #2751, opened 2020-12-30
[py] with syntax for vw

lalo:
proposal, don't think this is required


jackgerrits (approved):

python/tests/test_cb.py
```diff
@@ -29,6 +28,11 @@ def test_getting_started_example():
     test_df['index'] = range(1, len(test_df) + 1)
     test_df = test_df.set_index("index")
 
+    return train_df, test_df
+
+def test_getting_started_example():
+    train_data, test_df = helper_get_data()
```
lalo, line 34:
```suggestion
    train_df, test_df = helper_get_data()
```

olgavrou (approved):

PR #2752, opened 2020-12-30
Add Python 3.9 wheel for Linux

jackgerrits:
- Windows is pending compatible vcpkg packages being available
- MacOS is pending the required dependencies to be available for 3.9 on conda forge


.github/workflows/build_python_wheels_macos.yml
```diff
@@ -43,6 +44,7 @@ jobs:
         - { version: 3.6, include_dir_name: python3.6m/}
         - { version: 3.7, include_dir_name: python3.7m/}
```
lalo, line 45: why do some end on m and others dont?

jackgerrits: Not 100% sure but it changed over the course of the releases

jackgerrits:
Looks like there are dependency issues with the conda activate

.github/workflows/build_python_wheels_macos.yml
peterychang: the `m` suffix indicates its built using a pymalloc allocator. In most cases i've seen its the same thing (3.6 symlinks to 3.6m), but technically the libraries are not guaranteed to be binary compatible

peterychang:
LGTM

PR #2753, opened 2020-12-30
Typed option update

jackgerrits:
- Location is optional in typed option
- Location is written to as part of value call
- Provide callback for inheriting classes of typed_option


vowpalwabbit/options.h
```diff
@@ -88,25 +91,36 @@ struct typed_option : base_option
   typed_option& value(T value)
   {
     m_value = std::make_shared<T>(value);
+    if (m_location != nullptr) { *m_location = value; }
+    value_set_callback(value);
```
lalo, line 95: so the callback gets called always, it is not an either or situation depending on if m_location is set?

vowpalwabbit/options.h
```diff
@@ -88,25 +91,36 @@ struct typed_option : base_option
...
     return *this;
   }
 
   T value() const { return m_value ? *m_value : T(); }
 
-  T& m_location;
+protected:
+  // Allows inheriting classes to handle set values. Noop by default.
+  virtual void value_set_callback(T /*value*/) {}
```
lalo, line 103: should T be const, in case its a list? or that doesnt make sense here

jackgerrits: Const ref makes sense yeah

vowpalwabbit/options.h
jackgerrits: The callback is not related to the location, so it is always called.

vowpalwabbit/options.h
```diff
@@ -88,25 +91,36 @@ struct typed_option : base_option
   typed_option& value(T value)
   {
     m_value = std::make_shared<T>(value);
+    if (m_location != nullptr) { *m_location = value; }
+    value_set_callback(value);
     return *this;
   }
 
   T value() const { return m_value ? *m_value : T(); }
```
lalo, line 99: semi related since we're here: isn't this code wrong? shouldn't it put the default value after calling into T()?

vowpalwabbit/options.h
```diff
@@ -88,25 +91,36 @@ struct typed_option : base_option
   typed_option& value(T value)
```
lalo, line 91: is this the convention? or should value be called set_value?

jackgerrits: Looking back set and get would have been better. I'd like to refactor it, but in a separate change

vowpalwabbit/options.h
jackgerrits: `T()` is a fallback for if there is no value provided, it's the default for `T` but necessarily a default value in the sense of the option

vowpalwabbit/options.h
lalo: yea, i just wanted to see if we're missing somewhere else where we call into value(...) but we didn't use to update the location

vowpalwabbit/options.h
lalo: oh yea, typed option doesn't even have knowledge of default value

vowpalwabbit/options.h
jackgerrits: oh that's a good point

jackgerrits:
Requires #2759

vowpalwabbit/options.h
```diff
@@ -121,23 +121,51 @@ struct typed_option : base_option
...
+struct typed_option_with_location : typed_option<T>
 {
-  return option_builder<typed_option<T>>(name, location);
+  typed_option_with_location(const std::string& name, T& location) : typed_option<T>(name), m_location{&location} {}
+  virtual void value_set_callback(const T& value) override
+  {
+    if (m_location != nullptr)
+    {
+      *m_location = value;
+      m_location = nullptr;
```
olgavrou, line 151: why are we setting as nullptr?

jackgerrits: Since the location often points to a stack local you can only safely set the location in the first add and parse call. By setting it to null here we say only the first set gets written to location. 

However, I can see this being a problem when we set the value of an option before the add and parse call (such as when insert is replaced with set). Honestly, the whole location to write to thing is rather unsafe and makes the options framework more complicated.

jackgerrits: Reworked it so that add and parse signals through to here instead

olgavrou (approved):

PR #2756, opened 2020-12-31
[py] fix test

### LINKED ISSUE #2755: python test fails but only sometimes
https://github.com/VowpalWabbit/vowpal_wabbit/blob/f4e6b0fc0a53952fa079e5909240a4c188db3681/python/tests/test_pyvw.py#L210

test is missing finish_example since we're using parse but also for some reason finish_example only works if you call into learn before and not by itself

also not sure why CI doesn't fail, but it consistently fails when I run it on my env

lalo:
try to fix #2755


.scripts/linux/python36.sh
```diff
@@ -16,5 +16,5 @@ rm -rf build
 source activate test-python36
 python setup.py check -mrs
 python setup.py install --user
-py.test ./python/tests/
+py.test -rsx ./python/tests/
```
lalo, line 19:
```suggestion
py.test ./python/tests/
```

python/tests/test_pyvw.py
```diff
@@ -207,28 +207,47 @@ def test_keys_with_list_of_values():
...
     | a:1 b:0.5
     0:0.1:0.75 | a:0.5 b:1 c:2
     """
     )
+
+    helper_parse(["| a:1 b:0.5", "0:0.1:0.75 | a:0.5 b:1 c:2"])
+
+
+def test_parse_2():
+    model = vw(quiet=True, cb_adf=True)
```
olgavrou, line 238: can these also call `helper_parse`?

lalo: i wanted a second test not using helper_parse, in case theres funny business with the python gc

lalo:
can someone run the python tests locally, on the main branch to see if they can repro any failure
`python -m pytest ./python/tests/ -v`

olgavrou:
I can repro: tests/test_pyvw.py::test_parse Fatal Python error: Aborted

python/tests/test_pyvw.py
```diff
@@ -207,28 +207,47 @@ def test_keys_with_list_of_values():
     del model
 
 
-def test_parse():
+def helper_parse(examples):
     model = vw(quiet=True, cb_adf=True)
-    ex = model.parse("| a:1 b:0.5\n0:0.1:0.75 | a:0.5 b:1 c:2")
+    ex = model.parse(examples)
     assert len(ex) == 2
+    model.learn(ex)
```
lalo, line 214: try using setup instead of learn?

lalo:
line 408 of pyvw.py should check if its a list of examples vs list of strings

jackgerrits (approved):

PR #2759, opened 2021-01-05
Decouple concrete option type from the building interface

jackgerrits:
Currently, new options are directly built as typed_option<T>. This decouples the concrete option type and just requires it to conform to some interface.

The purpose of this is to allow higher level/custom option types.

### DIFF
--- a/vowpalwabbit/options.h
+++ b/vowpalwabbit/options.h
@@ -20,6 +20,70 @@ namespace VW
 {
 namespace config
 {
+// option_builder decouples the specific type of the option and the interface
+// for building it. It handles generating a shared_ptr at the end of the
+// base_option type which the options framework takes as input.
+// Therefore, T must satisfy:
+// - Inherit from base_option
+// - Have a shared_ptr field called m_default_value
+// - Statically expose a type value_type
+// - For nearly all purposes T should be typed_option<value_type> or a subclass
+//   of this.
+template <typename T>
+struct option_builder
+{
+  template <typename... Args>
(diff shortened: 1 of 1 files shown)


vowpalwabbit/options.h
```diff
@@ -20,69 +20,103 @@ namespace VW
...
   {
-    static_assert(is_scalar_option_type<T>::value, "allow_override can only apply to scalar option types.");
-    m_allow_override = allow_override;
+    if (!is_scalar_option_type<typename T::value_type>::value)
+    { THROW("allow_override can only apply to scalar option types.") }
+    m_option_obj.m_allow_override = allow_override;
     return *this;
   }
 
+  static std::shared_ptr<base_option> finalize(option_builder&& option)
```
olgavrou, line 80: I don't understand what happens to `option` after it is moved into `finalize()` since we only care about `option.m_option_obj`

jackgerrits: It is moved into a `shared_ptr` and returned. This preserves the concrete type while giving back the base object to be passed to be saved.

olgavrou: but we could just pass option as a const ref and then move the internal object?

jackgerrits: We could but the `option_builder` object is invalid after that point and calling any of the builder functions would result in undefined behavior

olgavrou: fair

olgavrou (approved):

PR #2761, opened 2021-01-05
Make prediction a struct.

kumpera:
Fix issue in ccb and disable bad unit test under OSX.


jackgerrits:
This is for prediction not label right? Label is done here: https://github.com/VowpalWabbit/vowpal_wabbit/pull/2275

kumpera:
> This is for prediction not label right? Label is done here: #2275

Yes, bad description.

kumpera:
It.. worked? 😮

vowpalwabbit/example.cc
```diff
@@ -12,7 +12,13 @@ VW_WARNING_DISABLE_DEPRECATED_USAGE
 example::example()
 {
   memset(&l, 0, sizeof(polylabel));
-  memset(&pred, 0, sizeof(polyprediction));
+  // init predictions
```
peterychang, line 15: Would it be possible to put these in the polypred constructor/destructor and just call placement new/the destructor here? Or even better, initialize everything properly in the constructor instead of relying on memset

kumpera: Let me fix the memory leaks find and get this PR green, then I'll address this issue.

Let's keep cleanups and improvements in the context of this PR minimal to simplify any bisection we need to do it the future.

vowpalwabbit/example.cc
```diff
@@ -134,7 +148,8 @@ void example::delete_unions(void (*)(polylabel*), void (*delete_prediction)(void
   CB_EVAL::cb_eval.delete_label(&l);
   MULTILABEL::multilabel.delete_label(&l);
 
-  if (delete_prediction) { delete_prediction(&pred); }
+  // if (delete_prediction) { delete_prediction(&pred); }
```
olgavrou, line 151: nit: delete commented out code

olgavrou (approved):

jackgerrits (approved):

PR #2765, opened 2021-01-06
Fail instead of returning default for value and default_value

jackgerrits:


vowpalwabbit/options.h
```diff
@@ -127,7 +131,11 @@ struct typed_option : base_option
     return *this;
   }
 
-  T value() const { return m_value ? *m_value : T(); }
+  T value() const
+  {
+    if (m_value) { return *m_value; }
```
lalo, line 136: throwing it out there, shouldn't this one return default_value (if set) if m_value does not exist?

jackgerrits: Hmm good point, it probably should. It has worked so far since default value is forwarded to boost PO and it is always set as part of the parsing process

jackgerrits: If someone calls `value_supplied` and there is no value but there _is_ a default value that would be returned does `value_supplied` return true or false?

lalo: I think value_supplied should return true in that case, but I believe that would break some current usage.

jackgerrits: value_supplied is not actually used anywhere right now, so it should be fine

lalo: let me think about it for a bit though

lalo (approved):

PR #2769, opened 2021-01-10
Fix cbzo test dependency mention

ajay0:
Probably just a nit but the first time I wrote these tests it seemed necessary to reference the previous test whose output file is being used in the current test. Not sure how they passed without that being correct. Nevertheless fixing it.


olgavrou (approved):
I believe it is because it is in the test description which does not affect the actual test results

PR #2771, opened 2021-01-11
[v_array] Make v_array an RAII type

jackgerrits:
Adds copy and move constructors for v_array


jackgerrits:
This is now ready for review

vowpalwabbit/cb_explore.cc
```diff
@@ -375,8 +376,7 @@ base_learner* cb_explore_setup(options_i& options, vw& all)
       data->epsilon_decay = true;
     }
     data->cs = (learner<cb_explore, example>*)(as_singleline(all.cost_sensitive));
-    data->second_cs_label.costs.resize(num_actions);
-    data->second_cs_label.costs.end() = data->second_cs_label.costs.begin() + num_actions;
+    for (uint32_t j = 0; j < num_actions; j++) { data->second_cs_label.costs.push_back(COST_SENSITIVE::wclass{}); }
```
peterychang, line 379: make an `actual_resize()` function?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -434,8 +435,10 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
         for (auto* ex : data.actions) { ex->interactions = &data.generated_interactions; }
       }
 
+      // shared, action, action, slot
       data.include_list.clear();
-      build_cb_example<is_learn>(data.cb_ex, slot, data);
+      build_cb_example<is_learn>(
+          data.cb_ex, slot, data.stored_labels[1 /* shared */ + data.actions.size() + slot_id], data);
```
peterychang, line 441: can the label's index go out of bounds somehow?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -456,7 +459,8 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
       else
       {
         // the cb example contains no action => cannot decide
-        decision_scores.push_back(data.action_score_pool.get_object());
+        decision_scores.push_back(v_array<ACTION_SCORE::action_score>());
+        data.action_score_pool.acquire_object(*decision_scores.end());
```
peterychang, line 463: `end()` is not a valid destination

vowpalwabbit/object_pool.h
```diff
@@ -121,30 +121,23 @@ struct no_lock_object_pool
...
+template <typename T>
+struct moved_object_pool
 {
-  value_object_pool() = default;
+  moved_object_pool() = default;
+
+  void reclaim_object(T&& obj) { m_pool.push(std::move(obj)); }
 
-  ~value_object_pool()
+  void acquire_object(T& dest)
```
peterychang, line 131: I'm not sure I'm a fan of this syntax. It would probably be cleaner to use `unique_ptr` with a custom deleter that returns the objects to the pool, but thats a much more involved change

vowpalwabbit/search.cc
```diff
@@ -3187,30 +3180,44 @@ predictor& predictor::add_to(v_array<T>& A, bool& A_is_ptr, T* a, size_t count,
...
 {
-  return add_to(oracle_actions, oracle_is_pointer, a, action_count, false);
+  for (size_t i = 0; i < action_count; i++) { oracle_actions.push_back(*(a + i)); }
+  return *this;
 }
+
 predictor& predictor::add_oracle(v_array<action>& a)
 {
-  return add_to(oracle_actions, oracle_is_pointer, a.begin(), a.size(), false);
+  for (const auto& item : a) { oracle_actions.push_back(item); }
```
peterychang, line 3201: TODO: range-based insertion

vowpalwabbit/v_array.h
```diff
@@ -39,24 +54,71 @@ struct v_array
...
+
+    std::swap(_begin, other._begin);
+    std::swap(_end, other._end);
+    std::swap(end_array, other.end_array);
+    std::swap(erase_count, other.erase_count);
+  }
+
+  v_array<T>& operator=(v_array<T>&& other) noexcept
+  {
+    delete_v_array();
```
peterychang, line 93: Is the deletion necessary here? The source object would theoretically clean everything up once it goes out of scope

vowpalwabbit/cb_explore.cc
jackgerrits: I'd like to do that in a follow up PR to avoid this one doing too much.

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: It should not as that vector is produced based on those other values.

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: Good point. Since we have just pushed back then end -1 should be valid

vowpalwabbit/object_pool.h
jackgerrits: It is a much more involved change thats for sure. I would like to avoid this for now

vowpalwabbit/search.cc
jackgerrits: I agree. Once v_array supports it

vowpalwabbit/v_array.h
jackgerrits: The moved out of object doesn't necessarily go out of scope though. Someone could use it. It seems better to leave it in this safe and well defined state.

jackgerrits: Will change to just a swap

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: Ill add an assert

olgavrou (approved):

PR #2772, opened 2021-01-11
[v_array] Give v_array constructor, destructor, copy and move special functions

jackgerrits:
This will not pass CI, the purpose is to review small changes visually until we get to a point where we are back to working end to end.

### DIFF
--- a/vowpalwabbit/v_array.h
+++ b/vowpalwabbit/v_array.h
@@ -72,10 +72,10 @@ public:
   inline const_iterator cbegin() const { return _begin; }
   inline const_iterator cend() const { return _end; }
 
-  v_array() : _begin(nullptr), _end(nullptr), end_array(nullptr), erase_count(0) {}
+  v_array() noexcept : _begin(nullptr), _end(nullptr), end_array(nullptr), erase_count(0) {}
   ~v_array() { delete_v_array(); }
 
-  v_array(v_array<T>&& other)
+  v_array(v_array<T>&& other) noexcept
   {
     erase_count = 0;
     _begin = nullptr;
@@ -88,7 +88,7 @@ public:
     std::swap(erase_count, other.erase_count);
   }
 
-  v_array<T>& operator=(v_array<T>&& other)
+  v_array<T>& operator=(v_array<T>&& other) noexcept
   {
     delete_v_array();
     std::swap(_begin, other._begin);
--- a/vowpalwabbit/v_array.h
+++ b/vowpalwabbit/v_array.h
@@ -114,6 +114,7 @@ public:
     if (this == &other) return *this;
 
     delete_v_array();
+    // TODO this should use the other version when T is trivially copyable and this otherwise.
     copy_array_no_memcpy(*this, other);
     return *this;
   }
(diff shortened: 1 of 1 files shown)


vowpalwabbit/v_array.h
```diff
@@ -39,25 +54,70 @@ struct v_array
...
+  inline const_iterator cbegin() const { return _begin; }
+  inline const_iterator cend() const { return _end; }
+
+  v_array() noexcept : _begin(nullptr), _end(nullptr), end_array(nullptr), erase_count(0) {}
+  ~v_array() { delete_v_array(); }
+
+  v_array(v_array<T>&& other) noexcept
+  {
+    erase_count = 0;
+    _begin = nullptr;
```
peterychang, line 81: this might leak memory. Do we care?

vowpalwabbit/v_array.h
```diff
@@ -31,6 +31,21 @@ const size_t erase_point = ~((1u << 10u) - 1u);
 template <class T>
 struct v_array
 {
+private:
+  void delete_v_array()
+  {
+    if (_begin != nullptr)
+    {
+      for (iterator item = _begin; item != _end; ++item) { item->~T(); }
```
peterychang, line 39: The rest of v_array assumes it only contains trivial types, do we want to maintain that assumption (for now)? We can add non-trivial type support in one shot later

vowpalwabbit/v_array.h
```diff
@@ -39,25 +54,70 @@ struct v_array
...
+    // TODO this should use the other version when T is trivially copyable and this otherwise.
+    copy_array_no_memcpy(*this, other);
+  }
+
+  v_array<T>& operator=(const v_array<T>& other)
+  {
+    if (this == &other) return *this;
+
+    delete_v_array();
+    copy_array_no_memcpy(*this, other);
```
olgavrou, line 117: does the above TODO apply here too?

vowpalwabbit/v_array.h
jackgerrits: This is the move constructor, so there is not object that we could be leaking.

vowpalwabbit/v_array.h
jackgerrits: There are places where it does not contain trivial types at the moment, such as decision_scores. So I think we need this

vowpalwabbit/v_array.h
jackgerrits: Yes it does, I'll add it in.

vowpalwabbit/v_array.h
```diff
@@ -39,25 +54,70 @@ struct v_array
...
+
+    std::swap(_begin, other._begin);
+    std::swap(_end, other._end);
+    std::swap(end_array, other.end_array);
+    std::swap(erase_count, other.erase_count);
+  }
+
+  v_array<T>& operator=(v_array<T>&& other) noexcept
+  {
+    delete_v_array();
```
olgavrou, line 93: shouldn't we check for
`if (this == &other)` here too?

jackgerrits: I believe the answer to this is no https://stackoverflow.com/questions/9322174/move-assignment-operator-and-if-this-rhs

olgavrou (approved):

PR #2774, opened 2021-01-11
[test] add vw version test

lalo:
this test is going to fail because we used 10 which is being compared as 1

### DIFF
--- a/test/unit_test/vw_versions_test.cc
+++ b/test/unit_test/vw_versions_test.cc
@@ -15,5 +15,5 @@ BOOST_AUTO_TEST_CASE(verify_vw_versions)
   version_struct temp;
   temp.from_string(VERSION_FILE_WITH_RANK_IN_HEADER);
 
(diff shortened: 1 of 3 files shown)


test/unit_test/vw_versions_test.cc
```diff
@@ -0,0 +1,19 @@
...
+#include "vw_versions.h"
+
+using namespace VW;
+
+BOOST_AUTO_TEST_CASE(verify_vw_versions)
+{
+  version_struct temp;
+  temp.from_string(VERSION_FILE_WITH_RANK_IN_HEADER);
+
+  BOOST_CHECK_EQUAL(true, temp < VERSION_FILE_WITH_INTERACTIONS);
```
jackgerrits, line 18: BOOST_CHECK requires the arg to be true

jackgerrits (approved):

test/unit_test/vw_versions_test.cc
lalo: done

PR #2775, opened 2021-01-12
[Prediction] Modernize polypred a little by introducing ctor.

kumpera:
In order to introduce a ctor, and pave the way for a dtor, we need to make the
type copy/assignment safe.

Since v_array is not copy safe, so is polyprediction. So I decided to make it
not copyable/assignable, only moveable.

This change introduce a new guard, stash_guard, that borrows the name from the same name
git command. It's used all over the place to replace code that manually save/restore example::pred.

With this change in, it should be possible to move cleanup from ~example() to ~polyprediction.


vowpalwabbit/nn.cc
```diff
@@ -394,7 +394,8 @@ void multipredict(nn& n, single_learner& base, example& ec, size_t count, size_t
     else
       predict_or_learn_multi<false, false>(n, base, ec);
     if (finalize_predictions)
-      pred[c] = ec.pred;
+      pred[c] = std::move(ec.pred);  // TODO: this breaks for complex labels because = doesn't do deep copy! (XXX we
+                                     // "fix" this by moving)
```
olgavrou, line 398: nit: can we remove the TODO now?

vowpalwabbit/learner.h
```diff
@@ -243,7 +243,8 @@ struct learner
       {
         learn_fd.predict_f(learn_fd.data, *learn_fd.base, (void*)&ec);
         if (finalize_predictions)
-          pred[c] = ec.pred;  // TODO: this breaks for complex labels because = doesn't do deep copy!
+          pred[c] = std::move(ec.pred);  // TODO: this breaks for complex labels because = doesn't do deep copy! (XXX we
```
olgavrou, line 246: nit: same here?

vowpalwabbit/example.h
```diff
@@ -50,6 +50,15 @@ inline void delete_scalars(void* v)
...
 {
+  polyprediction();
+
+  // disable those two since v_array is not cleanly copy-able
+  polyprediction(polyprediction&&) = default;
+  polyprediction& operator=(polyprediction&&) = default;
+
+  polyprediction(const polyprediction&) = delete;
+  polyprediction& operator=(const polyprediction&) = delete;
+  // ~polyprediction();
```
olgavrou, line 61: nit: remove commented out destructor

olgavrou (approved):

jackgerrits (approved):

vowpalwabbit/nn.cc
kumpera: Not really, the livecycle of non-scalar predictions remains a mystery in this case.

vowpalwabbit/example.h
kumpera: done

PR #2779, opened 2021-01-14
[v_array] CCB v_array fixes

jackgerrits:
Main change here is the value based object pool has changed to being rvalue based. Thanks @olgavrou for your help on that

### DIFF
--- a/vowpalwabbit/conditional_contextual_bandit.cc
+++ b/vowpalwabbit/conditional_contextual_bandit.cc
@@ -27,10 +27,10 @@ using namespace VW;
 using namespace VW::config;
 
 template <typename T>
-void return_v_array(v_array<T>& array, VW::v_array_pool<T>& pool)
+void return_v_array(v_array<T>&& array, VW::v_array_pool<T>& pool)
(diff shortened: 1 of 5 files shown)


olgavrou (approved):

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -113,18 +113,23 @@ bool split_multi_example_and_stash_labels(const multi_ex& examples, ccb& data)
 // create empty/default cb labels
 void create_cb_labels(ccb& data)
 {
-  data.shared->l.cb.costs = data.cb_label_pool.get_object();
+  data.cb_label_pool.get_object(data.shared->l.cb.costs);
```
olgavrou, line 116: I am struggling with a good name for `get_object` since one would expect it to return a value and not place the returned value on an out argument. Maybe `aquire_object`, `reclaim_object?`

jackgerrits: I like that naming

PR #2784, opened 2021-01-20
Make label_type a property of label_parser

olgavrou:
Making `label_type` a member of the `label_parser` struct so that it doesn't have to be redefined all over the place with the risk of it becoming out of sync with the label parser

### DIFF
--- a/vowpalwabbit/mwt.cc
+++ b/vowpalwabbit/mwt.cc
@@ -92,28 +92,27 @@ void predict_or_learn(mwt& c, single_learner& base, example& ec)
   }
   VW_WARNING_STATE_PUSH
   VW_WARNING_DISABLE_CPP_17_LANG_EXT
-  if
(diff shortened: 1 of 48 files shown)


vowpalwabbit/mwt.cc
```diff
@@ -92,27 +92,28 @@ void predict_or_learn(mwt& c, single_learner& base, example& ec)
...
-  {
-    c.indices.clear();
-    uint32_t stride_shift = c.all->weights.stride_shift();
-    uint64_t weight_mask = c.all->weights.mask();
-    for (unsigned char ns : ec.indices)
-      if (c.namespaces[ns])
-      {
-        c.indices.push_back(ns);
-        if (learn)
+  if
```
olgavrou, line 95: formatter picked up some random formatting here

jackgerrits (approved):
:ship:

PR #2791, opened 2021-01-29
[CATS ] report average loss instead of label cost

olgavrou:
to be merged after #2788 

cats was parsing the label and reporting the label cost as the average loss
this resulted in the same average loss being reported for the same file, no matter what the arguments passed


rajan-chari:
Reviewing ...

vowpalwabbit/cats.cc
```diff
@@ -146,8 +172,17 @@ void reduction_output::print_update_cb_cont(vw& all, const example& ec)
...
+  float bandwidth = 0;
+  float min_value = 0;
+  float max_value = 0;
+  new_options.add(make_option("cats", num_actions).keep().necessary().help("number of tree labels <k> for cats"))
+      .add(make_option("min_value", min_value).keep().help("Minimum continuous value"))
+      .add(make_option("max_value", max_value).keep().help("Maximum continuous value"))
+      .add(make_option("bandwidth", bandwidth)
+               .default_value(1)
+               .keep()
+               .help("Bandwidth (radius) of randomization around discrete actions in terms of continuous range."));
```
lalo, line 185: would it be beneficial to add default value to help message?

olgavrou: Actually thanks for pointing it out, should probably change that. `1` is not a good default anymore

vowpalwabbit/cats.cc
```diff
@@ -112,13 +117,34 @@ void reduction_output::output_predictions(std::vector<std::unique_ptr<VW::io::wr
   for (auto& f : predict_file_descriptors) f->write(str.c_str(), str.size());
 }
 
-// "average loss" "since last" "example counter" "example weight"
-// "current label" "current predict" "current features"
-void reduction_output::report_progress(vw& all, const cats&, const example& ec)
+void reduction_output::report_progress(vw& all, const cats& data, const example& ec)
```
rajan-chari, line 120: A unit test around this with  +/-data.min_value, +/-data.max_value, +/-actions would be helpful.

vowpalwabbit/cats.cc
olgavrou: Changed default in #2788 to be to be half the unit range (i.e. leaf width)

vowpalwabbit/cats.cc
olgavrou: Added unit tests

test/train-sets/ref/help.stdout
```diff
@@ -224,6 +224,15 @@ Bootstrap:
...
 Continuous actions tree with smoothing:
   --cats arg            number of tree labels <k> for cats
+  --min_value arg       Minimum continuous value
+  --max_value arg       Maximum continuous value
+  --bandwidth arg       Bandwidth (radius) of randomization around discrete 
+                        actions in terms of continuous range. By default will 
+                        be set to half of the continuous action unit-range 
+                        resulting in smoothing that stays inside the action 
+                        space unit-range:
+                        unit_range = (max_value - min_value) / num_actions
```
akshaykr, line 234: Is num_actions equal to the "--cats" argument? Maybe we should avoid using both "actions" and "tree labels" to refer to the same thing? (If it is the same thing)

olgavrou: yes it is good point, I'll use one term throughout

vowpalwabbit/cats.cc
```diff
@@ -61,6 +48,32 @@ int cats::learn(example& ec, experimental::api_status* status = nullptr)
...
+    const float unit_range = continuous_range / num_actions;
+
+    const float ac = (predicted_action - min_value) / unit_range;
+    int discretized_action = static_cast<int>(floor(ac));
+    // centre of predicted action
+    const float centre = min_value + discretized_action * unit_range + unit_range / 2.0f;
+
+    // is centre close to action from label
+    auto logged_action = cb_cont_costs.costs[0].action;
+    if ((logged_action - bandwidth <= centre) && (centre <= logged_action + bandwidth))
```
akshaykr, line 66: I think it's equivalent, but more intuitive to write this like
(logged_action - center <= bandwidth) and (center - logged_action <= bandwidth).
In a sense we are checking | logged_action - center | <= bandwidth so it is nice to keep these two together.

vowpalwabbit/cats.cc
```diff
@@ -144,8 +155,19 @@ void reduction_output::print_update_cb_cont(vw& all, const example& ec)
...
+  float max_value = 0;
+  new_options.add(make_option("cats", num_actions).keep().necessary().help("number of tree labels <k> for cats"))
+      .add(make_option("min_value", min_value).keep().help("Minimum continuous value"))
+      .add(make_option("max_value", max_value).keep().help("Maximum continuous value"))
+      .add(make_option("bandwidth", bandwidth)
+               .keep()
+               .help("Bandwidth (radius) of randomization around discrete actions in terms of continuous range. By "
+                     "default will be set to half of the continuous action unit-range resulting in smoothing that "
+                     "stays inside the action space unit-range:\nunit_range = (max_value - min_value) / "
+                     "num_actions\ndefault bandwidth = unit_range / 2.0"));
```
akshaykr, line 170: Same comment as above here. Better to use the same terminology, "tree labels" or "actions"

vowpalwabbit/cats.cc
```diff
@@ -61,6 +48,32 @@ int cats::learn(example& ec, experimental::api_status* status = nullptr)
...
+    const float ac = (predicted_action - min_value) / unit_range;
+    int discretized_action = static_cast<int>(floor(ac));
+    // centre of predicted action
+    const float centre = min_value + discretized_action * unit_range + unit_range / 2.0f;
+
+    // is centre close to action from label
+    auto logged_action = cb_cont_costs.costs[0].action;
+    if ((logged_action - bandwidth <= centre) && (centre <= logged_action + bandwidth))
+    {
+      float actual_b = std::min(max_value, logged_action + bandwidth) - std::max(min_value, logged_action - bandwidth);
```
akshaykr, line 68: I think we should replace logged_action with centre when computing actual_b. (I think this does make a small difference!)

akshaykr (approved):

PR #2792, opened 2021-01-29
Consolidate get_observed_cost implementations

jackgerrits:
This removes several parallel implementations and settles on just one for cb and one for cb_adf. It removes a pointer return in favor of a value. However, this value /really/ should be an optional so I am not sure if this API is totally correct...


vowpalwabbit/cb.h
```diff
@@ -5,23 +5,35 @@
...
-  float partial_prediction;  // essentially a return value
-  bool operator==(cb_class j) { return action == j.action; }
+  float cost = FLT_MAX;      // the cost of this class
+  uint32_t action = 0;       // the index of this class
+  float probability = -1.f;  // new for bandit setting, specifies the probability the data collection policy chose this
+                             // class for importance weighting
+  float partial_prediction = 0.f;  // essentially a return value
+
+  cb_class() = default;
+  cb_class(float cost, uint32_t action, float probability)
```
peterychang, line 29: It seems dangerous having both using list initialization and having a constructor like this. If the member variables aren't ordered the same way, then their meanings will be different, which can lead to some hard to debug issues

vowpalwabbit/mwt.cc
```diff
@@ -24,7 +25,7 @@ struct mwt
 {
   bool namespaces[256];        // the set of namespaces to evaluate.
   v_array<policy_data> evals;  // accrued losses of features.
-  CB::cb_class* observation;
+  std::pair<bool, CB::cb_class> optional_observation;
```
peterychang, line 28: pairs default to 0 initialization, right? If not, then we should probably initialize `first` to false

vowpalwabbit/cb.h
jackgerrits: Where is the list initialization?

vowpalwabbit/mwt.cc
jackgerrits: Pairs will value initialize[0] which will result in a false value for the bool[1].

- [\[0\] - https://stackoverflow.com/questions/9025792/does-the-default-constructor-of-stdpair-set-basic-types-int-etc-to-zero](https://stackoverflow.com/questions/9025792/does-the-default-constructor-of-stdpair-set-basic-types-int-etc-to-zero)
- [\[1\] - https://stackoverflow.com/questions/3803153/what-are-primitive-types-default-initialized-to-in-c](https://stackoverflow.com/questions/3803153/what-are-primitive-types-default-initialized-to-in-c)

vowpalwabbit/cb_algs.cc
```diff
@@ -49,9 +49,14 @@ void predict_or_learn(cb& data, single_learner& base, example& ec)
 {
   CB::label ld = ec.l.cb;
   cb_to_cs& c = data.cbcs;
-  c.known_cost = get_observed_cost(ld);
-  if (c.known_cost != nullptr && (c.known_cost->action < 1 || c.known_cost->action > c.num_actions))
-    std::cerr << "invalid action: " << c.known_cost->action << std::endl;
+  auto optional_cost = get_observed_cost_cb(ld);
```
olgavrou, line 52: it looks like we could skip the check here and just do `c.known_cost = optional_cost.second` since it will hold the default `CB::cb_class` if the cost is not known.

vowpalwabbit/cb_algs.cc
```diff
@@ -49,9 +49,14 @@ void predict_or_learn(cb& data, single_learner& base, example& ec)
...
+  if (optional_cost.first) { c.known_cost = optional_cost.second; }
+  else
+  {
+    c.known_cost = CB::cb_class{};
+  }
+  if (optional_cost.first && (optional_cost.second.action < 1 || optional_cost.second.action > c.num_actions))
```
olgavrou, line 58: maybe here instead of using `optional_cost.second.action` using `c.known_cost.action` would be easier to understand?

vowpalwabbit/cb_algs.cc
```diff
@@ -83,7 +88,12 @@ void learn_eval(cb& data, single_learner&, example& ec)
   CB_EVAL::label ld = ec.l.cb_eval;
 
   cb_to_cs& c = data.cbcs;
-  c.known_cost = get_observed_cost(ld.event);
+  auto optional_cost = get_observed_cost_cb(ld.event);
```
olgavrou, line 91: same here? just assign `optional_cost.second` to known cost? Maybe instead of `optional_cost` the name `observed_or_default_cost` would also work

vowpalwabbit/cb.cc
```diff
@@ -17,6 +17,15 @@ using namespace VW::LEARNER;
 
 namespace CB
 {
+std::pair<bool, cb_class> get_observed_cost_cb(const label& ld)
```
olgavrou, line 20: maybe this should also be renamed to `get_observed_cost_or_default_cb` to match the `adf` case

olgavrou (approved):
Apart from some small comments lgtm. Optionally might be nice to add a comment over all the `optional_cost.first` statements clarifying that it means: `// cost observed, not default`, or something like that

vowpalwabbit/cb_algs.cc
jackgerrits: I intentionally did not do this as if this were a true optional then the `.second` value would not be accessible

vowpalwabbit/cb.cc
jackgerrits: The contract of this function is if the first value is false that the second value should not be read which is why it is named like that compared to the other one

vowpalwabbit/cb_algs.cc
jackgerrits: Same explanation here too

vowpalwabbit/cb.h
peterychang: every usage of cb_class is done via list initialization..  the first diff on the PR for example `ec.l.cb.costs.push_back(CB::cb_class{3.5f, 2, 0.5f});`

jackgerrits: `CB::cb_class{3.5f, 2, 0.5f}` is calling this constructor. List initialization is not available for this type now. (It used to be but this removes that)

vowpalwabbit/cb_algs.cc
olgavrou: True true

PR #2807, opened 2021-02-12
Make -q :: faster by calculating interactions on the fly instead of pre calculating them

olgavrou:
This PR affects only `-q ::` for now, could be extended to also work for `--cubic ::` in different PR.

We don't hold all possible interactions anymore but calculate them on the fly by keeping namespaces when they are found.
This should not affect explicit quadratics when passed

Could potentially improve more. Reduces run time for a `ccb adf` run with a file with `347k` examples and 3 interactions (Action,Slot,User) from ~ `11m41s` to ~ `1m50s`


vowpalwabbit/example_predict.h
```diff
@@ -12,7 +12,33 @@ typedef unsigned char namespace_index;
...
+#  pragma managed(push, off)
+#  undef _M_CEE
+#  include <mutex>
+#  define _M_CEE 001
+#  pragma managed(pop)
+#else
+#  include <mutex>
+#endif
+
+struct namsepace_interactions
```
ataymano, line 30: typo? namsepace -> namespace

vowpalwabbit/interactions.cc
```diff
@@ -80,6 +80,61 @@ std::vector<std::vector<namespace_index>> expand_interactions(
...
+  for (auto it = set_interactions.begin(); it != set_interactions.end(); ++it)
+  {
+    for (auto jt = it; jt != set_interactions.end(); ++jt)
+    {
+      if (interactions.active_interactions.find({*it, *jt}) == interactions.active_interactions.end())
+      {
+        active_interactions.push_back({*it, *jt});
+        interactions.active_interactions.insert({*it, *jt});
+      }
+      if (interactions.active_interactions.find({*it, *it}) == interactions.active_interactions.end())
```
ataymano, line 102: seems jt-independent - can we extract it from jt loop?

vowpalwabbit/interactions.cc
```diff
@@ -80,6 +80,61 @@ std::vector<std::vector<namespace_index>> expand_interactions(
...
+      {
+        active_interactions.push_back({*it, *it});
+        interactions.active_interactions.insert({*it, *it});
+      }
+      if (interactions.active_interactions.find({*jt, *jt}) == interactions.active_interactions.end())
+      {
+        active_interactions.push_back({*jt, *jt});
+        interactions.active_interactions.insert({*jt, *jt});
+      }
+      if (interactions.active_interactions.find({*jt, *it}) == interactions.active_interactions.end() &&
```
ataymano, line 112: Do we need this?

vowpalwabbit/interactions.cc
olgavrou: Yes good point

vowpalwabbit/example_predict.h
olgavrou: Yes thanks

vowpalwabbit/interactions.cc
olgavrou: I believe so. This allows for duplicate interactions (i.e. if namespace `A` and namespace `T` was found it will add `A*T` and `T*A` as interactions). If check is removed then `-q :: --leave_duplicate_interactions` and `-q ::` are identical
I reverted the order and check for `leave_duplicate_interactions` first

vowpalwabbit/example_predict.h
```diff
@@ -12,7 +12,33 @@ typedef unsigned char namespace_index;
...
+#  pragma managed(push, off)
+#  undef _M_CEE
+#  include <mutex>
+#  define _M_CEE 001
+#  pragma managed(pop)
+#else
+#  include <mutex>
+#endif
+
+struct namespace_interactions
```
ataymano, line 30: Not sure if we want to do it in this iteration, but is there a reason for not making new code more oop-ish?
It looks like it can be done as class without exposing all internals and expand_quadratics_wildcard_interactions as potential method.
I.e. size variable looks a bit confusing here - it is not clear which size it is actually + as far as I can see it is actually used as isUpdated flag which could be set to false on every add(namespace) call or something like that.

olgavrou: There is no particular reason apart from the fact that, `interactions.interactions` vector will be used independently from everything else if we do not do `-q ::` and all the rest is only needed/used if we do `-q ::` and I opted for not disturbing the rest of the code if not

`size` does need a better name for sure and we could look into refactoring to make it more object oriented

ataymano (approved):

vowpalwabbit/example_predict.cc
```diff
@@ -81,3 +81,21 @@ void safe_example_predict::clear()
   for (auto ns : indices) feature_space[ns].clear();
   indices.clear();
 }
+
+void namespace_interactions::clear()
+{
+  active_interactions.clear();
+  all_seen_namespaces.clear();
+  interactions.clear();
+  quadraditcs_wildcard_expansion = false;
```
jackgerrits, line 90: Typo:
```suggestion
  quadratics_wildcard_expansion = false;
```

vowpalwabbit/interactions.cc
```diff
@@ -80,6 +80,62 @@ std::vector<std::vector<namespace_index>> expand_interactions(
...
+        active_interactions.push_back({*jt, *it});
+        interactions.active_interactions.insert({*jt, *it});
+      }
+    }
+  }
+
+  interactions.interactions.insert(
+      interactions.interactions.end(), active_interactions.begin(), active_interactions.end());
+  std::sort(interactions.interactions.begin(), interactions.interactions.end(),
+      [](std::vector<namespace_index>& a, std::vector<namespace_index>& b) {
```
jackgerrits, line 125: Can these be const?

jackgerrits (approved):
Looks great! Just a thought, should we test combining `-q ::` with explicit `-q AB` and other interactions `--interaction AAA --interaction :::`?

olgavrou:
> Looks great! Just a thought, should we test combining `-q ::` with explicit `-q AB` and other interactions `--interaction AAA --interaction :::`?

Good idea about a mixed quadratics test, will add

vowpalwabbit/interactions.cc
olgavrou: yes

lalo:
Relevant https://github.com/VowpalWabbit/vowpal_wabbit/issues/1527

lalo:
We will need to update these docs (see bottom) https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Feature-interactions

olgavrou:
> We will need to update these docs (see bottom) https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Feature-interactions

Awesome thanks, I would not have noticed that

Added more tests to this, leaving it broken for now (tests timeout needs to be increased) since I am experiencing something strange for the latest tests: `-q :: -interactions AAA --interactions :::`. The interactions generated match exactly with those when run from master (seen from `--invert_hash` output) but their weights are ever so slightly different and I don't know why and if it should be a concern or not

rajan-chari:
Looks good.

rajan-chari (approved):

PR #2813, opened 2021-02-19
Refactoring VW logging

### LINKED ISSUE #1937: Unify warning and error reporting
There are several methods used to output warnings and errors:

Warning:
- vw_ostream -> all.trace_message
- std::cerr
- VW_TRACE

Error:
- THROW
- VW_ASSERT

They should all be unified under one interface/macro so that they can be searched for easily and their behavior can be controlled centrally as well.

Perhaps two macros as the top level API:
VW_WARN
VW_ERROR

peterychang:
The biggest question will be, do we leave logging as standalone functions or do we create a wrapper object that we pass through the system? The latter gives us flexibility, but will require a lot of refactoring to allow for the object to be passed around

Closes #1937


vowpalwabbit/io/logger.h
```diff
@@ -0,0 +1,79 @@
...
+  
+  template<typename FormatString, typename... Args>
+    void log_error(const FormatString &fmt, Args&&...args)
+  {
+    spdlog::default_logger_raw()->error(fmt, std::forward<Args>(args)...);
+  }
+
+
+  template<typename FormatString, typename... Args>
+    void log_critical(const FormatString &fmt, Args&&...args)
```
jackgerrits, line 71: Do we want both error and critical? I feel like just error is sufficient

vowpalwabbit/io/logger.h
```diff
@@ -0,0 +1,79 @@
...
+  {
+    _internal_logger->critical(fmt, std::forward<Args>(args)...);
+  }
+
+};
+  */
+
+namespace logger
+{
+  enum class log_level { trace, debug, info, warn, error, critical, off };
```
jackgerrits, line 48: If we only support a subset of these we should remove the ones we dont use

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "rapidjson"]
 	path = rapidjson
 	url = https://github.com/Tencent/rapidjson.git
+[submodule "ext_libs/spdlog"]
+	path = ext_libs/spdlog
```
jackgerrits, line 5: We should target a tagged version of spdlog

jackgerrits (commented):
Looks good. In my opinion we should use an object based logger to allow for us to swap it out during testing. But for now we can start with this first step and consider if an object to be passed around is helpful?

There is already an object called `vw_logger` which we should perhaps deal with to avoid confusion with this.

vowpalwabbit/io/logger.h
peterychang: I see error as separate and distinct from critical. Errors are localized problems that will, for example, cause an example to be unusable. Critical errors are problems that will cause the entire state to be unusable. We may not use it now, but I think its something we should start using going forward

jackgerrits: Sounds like a good distinction

vowpalwabbit/io/logger.h
jackgerrits: Are we going to support trace and debug? It seems as though they may be too verbose and add too many runtime log checks?

peterychang: I'm not sure, it depends it can be useful for debugging purposes, but I agree it adds a lot of branching statements. We could always use a compile-time flag to no-op some of the more verbose logging, but I'm not really a fan of having 2 ways to turn off logging

jackgerrits: The multiple ways of turning off logging already exists...

peterychang:
There are several places I had to leave TODO's, mostly because they're logging partial lines, which spdlog cannot do (see `bfgs` and `ect` for examples). We might be able to handle some instances with custom ostream operators, but others not so much.

vowpalwabbit/cs_active.cc
```diff
@@ -10,14 +10,20 @@
...
+#include "io/logger.h"
 //#define B_SEARCH_MAX_ITER 50
 #define B_SEARCH_MAX_ITER 20
 
 using namespace VW::LEARNER;
 using namespace COST_SENSITIVE;
 using namespace VW::config;
 
-using std::cerr;
+// TODO: cs_active should have its own logger instance (since it uses its own debug flag)
```
jackgerrits, line 22: Should we take this opportunity to converge it on the global log levels?

jackgerrits:
Looking at the test failures - it looks like in quite a few cases output has been moved to the logging interface that should have been moved to the driver printing interface

peterychang:
Some of them I can see being logged under the wrong logger. Others though, seem like the tests themselves are wrong (since they just do an output comparison). For example
```
 Enabled reductions: gd, scorer
-NAN prediction in example 1, forcing 0
-your features have too much magnitude
-update is NAN, replacing with 0
 1.000000 1.000000            1            1.0   1.0000   0.0000        2
-NAN prediction in example 2, forcing 0.000000
-your features have too much magnitude
-update is NAN, replacing with 0
 1.000000 1.000000            2            2.0   1.0000   0.0000        2
```
Those `NAN prediction in example` logs are warnings/errors, should they really be logged like that?

jackgerrits:
You're right those are warnings/errors. I am referring to:
```diff
-Enabling FTRL based optimization
-Algorithm used: Coin Betting
-ftrl_alpha = 4
-ftrl_beta = 1
 Num weight bits = 18
 learning rate = 0.5
 initial_t = 0

```

peterychang:
Yeah, I'm moving the setup-based logs under the trace logger. I'll fix the test for the error ones

peterychang:
Most usages of cout/cerr have been converted to using one of the two loggers. Several TODO's littered throughout, much of it regarding questions around which logger is the correct one to use

peterychang:
Some benchmarking
```
-----------------------------------------------------
Benchmark              Time           CPU Iterations
-----------------------------------------------------
direct_stderr       9321 ns       9417 ns      74667
direct_spdlog       2842 ns       2783 ns     235789
direct_fmt          1911 ns       1904 ns     344615
ss_to_stderr        4792 ns       4671 ns     167253
ss_to_spdlog        3282 ns       3296 ns     213333
ss_to_fmt           2349 ns       2354 ns     298667
fmt_to_stderr       3850 ns       3836 ns     179200
fmt_to_spdlog       2859 ns       2849 ns     235789
fmt_to_fmt          2048 ns       2040 ns     344615
```

The outputs were printed to stderr, which was redirected to /dev/null
The output consisted of the string
`"this is a long string to be a long string: " << x++ << " " << y--` (where x and y are ints) plus a newline.

Three output styles were used:
`cerr`, `spdlog`, and `fmt`. `spdlog` appended additionally appended the log level to the line (current behavior in this PR)

Additionally, 3 string construction methods were used:
natively, constructing a stringstream and printing `ss.str()`, and constructing a string via `fmt::format()`

I'm completely baffled as to why printing directly to stderr is slower than constructing a stringstream and printing that to stderr...

kumpera:
> 
> 
> Most usages of cout/cerr have been converted to using one of the two loggers. Several TODO's littered throughout, much of it regarding questions around which logger is the correct one to use

I think the easiest thing is to enable it at high log levels and later fix those that are too verbose.

peterychang:
@kumpera One problem is that the new logger outputs with a log level prepended onto the line (which I think is the feature we need), whereas `trace_message` does not (again, behavior we want to keep)

lalo:
Looks like you can customize the format to remove the log level at the beginning, but yes if it's useful we should keep it. 

https://github.com/gabime/spdlog/wiki/3.-Custom-formatting

peterychang:
One thing to watch out for with this change is expensive function calls in the logger's parameter list. Since the quiet flag is checked within the function call, all parameters are evaluated regardless of the quiet flag. Currently the only one that looks relatively expensive is `strerror_to_string()`, which is constructing a string. `fmt::join()` looks to be pretty well optimized, doing the heavy lifting through the ostream operator instead of at the function's evaluation

vowpalwabbit/CMakeLists.txt
```diff
@@ -345,9 +352,9 @@ add_library(vw STATIC ${vw_all_sources} ${vw_all_headers})
 
 target_link_libraries(vw
   PUBLIC
-    VowpalWabbit::explore VowpalWabbit::allreduce Boost::boost
+    VowpalWabbit::explore VowpalWabbit::allreduce Boost::boost vw_io
```
jackgerrits, line 355: I think we should leave vw_io as a private dep but add spdlog and fmt here as public deps

jackgerrits:
Can you fix the LGTM issue: "2 for Inconsistent definition of copy constructor and assignment ('Rule of Two')"

jackgerrits (approved):

peterychang:
The LGTM warning for the rule of 2 is inside the spdlog library

PR #2821, opened 2021-02-28
Remove code that can now be default

jackgerrits:
When defaulting move constructors the noexcept must be omitted and it is up to the to compiler to add it if it can. https://akrzemi1.wordpress.com/2015/09/11/declaring-the-move-constructor/


vowpalwabbit/feature_group.cc
```diff
@@ -17,57 +17,37 @@ struct feature_slice  // a helper struct for functions using the set {v,i,space_
...
-  v._end = nullptr;
-  v.end_array = nullptr;
-  auto& i = other.indicies;
-  i._begin = nullptr;
-  i._end = nullptr;
-  i.end_array = nullptr;
-  other.sum_feat_sq = 0;
-  return *this;
-}
+// features::features()
```
peterychang, line 20: can we just delete this?

peterychang:
lgtm

vowpalwabbit/feature_group.cc
jackgerrits: Good catch thanks!

PR #2823, opened 2021-03-01
Add reserve, resize_actual, make erase_count private

jackgerrits:
Migrated from @peterychang's work here: https://github.com/peterychang/vowpal_wabbit/pull/2

### DIFF
--- a/vowpalwabbit/v_array.h
+++ b/vowpalwabbit/v_array.h
@@ -174,7 +174,7 @@ public:
 
   // maintain the original (deprecated) interface for compatibility. To be removed in VW 10
   //   VW_DEPRECATED(
-  //       "v_array::resize() is deprecated. Use reserve() instead. \
+  //       "v_array::resize() is deprecated. Use reserve() instead.
   // For standard resize behavior, use actual_resize(). The function names will be re-aligned in VW 10")
   void resize(size_t length) { reserve_nocheck(length); }
 
--- a/vowpalwabbit/v_array.h
+++ b/vowpalwabbit/v_array.h
@@ -50,12 +50,11 @@ private:
 
   void reserve_nocheck(size_t length)
   {
-    if(capacity() == length || length == 0) { return; }
+    if (capacity() == length || length == 0) { return; }
     const size_t old_len = size();
 
     T* temp = reinterpret_cast<T*>(std::realloc(_begin, sizeof(T) * length));
-    if (temp == nullptr)
-    { THROW_OR_RETURN("realloc of " << length << " failed in resize().  out of memory?"); }
+    if (temp == nullptr) { THROW_OR_RETURN("realloc of " << length << " failed in resize().  out of memory?"); }
     else
     {
       _begin = temp;
--- a/vowpalwabbit/v_array.h
+++ b/vowpalwabbit/v_array.h
@@ -53,9 +53,7 @@ private:
 
     T* temp = reinterpret_cast<T*>(std::realloc(_begin, sizeof(T) * length));
     if ((temp == nullptr) && ((sizeof(T) * length) > 0))
-    {
-      THROW_OR_RETURN("realloc of " << length << " failed in resize().  out of memory?");
-    }
+    { THROW_OR_RETURN("realloc of " << length << " failed in resize().  out of memory?"); }
     else
     {
       _begin = temp;
--- a/vowpalwabbit/v_array.h
+++ b/vowpalwabbit/v_array.h
@@ -32,6 +32,8 @@ const size_t erase_point = ~((1u << 10u) - 1u);
 template <class T>
 struct v_array
 {
+  static_assert(sizeof(T) > 0, "The sizeof v_array's element type T cannot be 0.");
+
 private:
   void delete_v_array()
   {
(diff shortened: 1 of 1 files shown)


vowpalwabbit/v_array.h
```diff
@@ -43,7 +43,25 @@ struct v_array
     _begin = nullptr;
     _end = nullptr;
     end_array = nullptr;
-    erase_count = 0;
+    _erase_count = 0;
+  }
+
+  void reserve_nocheck(size_t length)
+  {
+    if (capacity() == length) { return; }
```
peterychang, line 51: `if(capacity() == length || length == 0)`, and remove the length check on line 55 (realloc for size 0 is undefined behavior)

vowpalwabbit/v_array.h
```diff
@@ -151,41 +172,69 @@ struct v_array
   inline size_t size() const { return _end - _begin; }
   inline size_t capacity() const { return end_array - _begin; }
 
-  void resize(size_t length)
+  // maintain the original (deprecated) interface for compatibility. To be removed in VW 10
+  //   VW_DEPRECATED(
+  //       "v_array::resize() is deprecated. Use reserve() instead.
+  // For standard resize behavior, use actual_resize(). The function names will be re-aligned in VW 10")
+  void resize(size_t length) { reserve_nocheck(length); }
```
peterychang, line 179: We'll need to destroy objects if the v_array is shrinking. Alternatively we could just call `reserve()` here? Not sure what the expected behavior is

vowpalwabbit/v_array.h
```diff
@@ -151,41 +172,69 @@ struct v_array
   inline size_t size() const { return _end - _begin; }
   inline size_t capacity() const { return end_array - _begin; }
 
-  void resize(size_t length)
+  // maintain the original (deprecated) interface for compatibility. To be removed in VW 10
+  //   VW_DEPRECATED(
```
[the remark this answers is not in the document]
jackgerrits: Resize is still used in lots of places and I haven't removed them yet. Since warnings are error this will break Windows CI. But it is commented out to show the intention.

vowpalwabbit/v_array.h
```diff
@@ -43,7 +43,25 @@ struct v_array
...
+    if ((temp == nullptr) && ((sizeof(T) * length) > 0))
+    { THROW_OR_RETURN("realloc of " << length << " failed in resize().  out of memory?"); }
+    else
+    {
+      _begin = temp;
+    }
+
+    _end = _begin + std::min(old_len, length);
+    end_array = _begin + length;
+    memset(_end, 0, (end_array - _end) * sizeof(T));
```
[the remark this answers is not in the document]
jackgerrits: we probably don't but it may break things if removed at the moment. VW tends to rely on 0'd memory a bit still

vowpalwabbit/v_array.h
```diff
@@ -53,8 +71,11 @@ struct v_array
 
 public:
   T* end_array;
```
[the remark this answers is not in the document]
jackgerrits: Not yet, it is still used quite a lot

vowpalwabbit/v_array.h
jackgerrits: I'm not sure the best way to handle this because the old resize did not destroy. We will also not need to destroy once constrained to trivially copyable only. The new actual_resize destroys too.

I think we should keep as is because it is bug for bug compatible with the old versioj

vowpalwabbit/v_array.h
```diff
@@ -151,41 +172,69 @@ struct v_array
...
+  void resize(size_t length) { reserve_nocheck(length); }
+
+  // change the number of elements in the vector
+  // to be renamed to resize() in VW 10
+  void actual_resize(size_t length)
+  {
+    auto old_size = size();
+    // if new length is smaller than current size destroy the excess elements
+    for (auto idx = length; idx < old_size; ++idx) { _begin[idx].~T(); }
+    if (capacity() < length) { reserve_nocheck(length); }
```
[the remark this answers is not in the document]
jackgerrits: This was a copy paste from your code :) Will update

vowpalwabbit/v_array.h
jackgerrits: Should we add a static assert for sizeof(T) > 0? Since the realloc function is undefined for 0?

vowpalwabbit/v_array.h
```diff
@@ -151,41 +173,69 @@ struct v_array
...
+  // change the number of elements in the vector
+  // to be renamed to resize() in VW 10
+  void actual_resize(size_t length)
   {
-    if (capacity() != length)
+    auto old_size = size();
+    // if new length is smaller than current size destroy the excess elements
+    for (auto idx = length; idx < old_size; ++idx) { _begin[idx].~T(); }
+    reserve(length);
+    _end = _begin + length;
```
olgavrou, line 190: won't `_end` already be set if `reserve_nocheck` is called via `reserve`?

olgavrou: oh never mind, you are filling in the rest of the array on line 194

olgavrou (approved):
What do you think about some unit tests for v_array?

jackgerrits:
Yeah unit tests would be very good. I'll do some in a separate PR

PR #2824, opened 2021-03-01
v_array cleanup

jackgerrits:
Make v_array.h more lightweight of a file to bring in.

Remove dead code:
- pop
- vstring
- v_move, v_extract

Move printing functions to debug_print.h

### DIFF
--- a/vowpalwabbit/debug_print.h
+++ b/vowpalwabbit/debug_print.h
@@ -27,21 +27,3 @@ std::ostream& operator<<(std::ostream& os, const std::vector<T>& vec)
   return os;
 }
 }  // namespace std
-
-template <class T>
-std::ostream& operator<<(std::ostream& os, const v_array<T>& v)
-{
-  os << '[';
-  for (T* i = v._begin; i != v._end; ++i) os << ' ' << *i;
-  os << " ]";
-  return os;
-}
-
-template <class T, class U>
-std::ostream& operator<<(std::ostream& os, const v_array<std::pair<T, U> >& v)
-{
-  os << '[';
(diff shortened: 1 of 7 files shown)


vowpalwabbit/debug_print.h
```diff
@@ -24,4 +26,22 @@ std::ostream& operator<<(std::ostream& os, const std::vector<T>& vec)
   os << " ]";
   return os;
 }
-}  // namespace std
\ No newline at end of file
+}  // namespace std
+
+template <class T>
```
peterychang, line 31: nit: Personally I like to keep object-specific functionality with the object (that way I can just include `v_array.h` when I want to print one instead of needing both that and `debug_print.h`)

vowpalwabbit/v_array.h
```diff
@@ -310,68 +301,10 @@ void calloc_reserve(v_array<T>& v, size_t length)
...
-}
-
-inline std::string v_string2string(const v_string& v_s)
-{
-  std::string res;
-  for (unsigned char* i = v_s._begin; i != v_s._end; ++i) res.push_back(*i);
-  return res;
-}
-
-template <typename T>
```
peterychang, line 364: deprecate these lines functions we remove them at a later point?

vowpalwabbit/debug_print.h
jackgerrits: That requires iostream to be in the v_array header which is relatively heavy. That's the motivating factor here

vowpalwabbit/v_array.h
jackgerrits: We haven't used them for some time. Are they actually part of the "supported" surface of VW?

peterychang:
lgtm

olgavrou (approved):

Issue #2820, opened 2021-02-27
AttributeError: 'vw' object has no attribute 'finished'


JuiP:
#### Traceback:
```Exception ignored in: <bound method vw.__del__ of <vowpalwabbit.pyvw.vw object at 0x7fbd0e2caea0>>
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/vowpalwabbit/pyvw.py", line 516, in __del__
    self.finish()
  File "/usr/local/lib/python3.6/dist-packages/vowpalwabbit/pyvw.py", line 482, in finish
    if not self.finished:
AttributeError: 'vw' object has no attribute 'finished'
```

#### How to Reproduce
I am not sure what caused it, it just happened once. I tried investigating the cause; looked at the source code [here](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/vowpalwabbit/pyvw.py) but no progress.
My guess is it is a race condition.

#### Environment
Version of vw: `8.9.0`
I was using the python wrapper in jupyter notebook (OS: Ubuntu 18.04)

jackgerrits:
What code did you run to cause this?

JuiP:
This is the code I was repeatedly executing with different exploration algorithms and parameters. Like I mentioned before, it happened once and on re-executing it worked with no tracebacks.

```# Instantiate learner in VW
vw = pyvw.vw("--cb_explore_adf -q UA --quiet --epsilon 0.1")

num_iterations = 5000
ctr = run_simulation(vw, num_iterations, users, times_of_day, actions, get_cost)

plot_ctr(num_iterations, ctr)

**Comment by ataymano** [comment]
Got similar message on this yesterday. Way to reproduce.
```
try:
    _vw = pyvw.vw('--cb_explore_adf --epsilon 0.2 --some_unrecognized_option')
except Exception as e:
    print("Ignore exception")
```
Can take a look.

ataymano:
It is clear why is it reproduced in my case, but the one reported seems different.
As long as init is successful, finished should be there.
If you see it next time, can you please:
1) check if it is reproducible without run_simulation / plot_ctr and with constructor only. By the way, is the snippet you shared wrapped into try/except?
2) maybe double check if there is any pyvw object construction/copying inside run_simulation?

JuiP:
Hi @ataymano, thanks for looking at this issue.
Checked #2826, I think that should probably fix this.

[...]

PR #2826, opened 2021-03-02
Ataymano/python safe constructor

ataymano:
do not throw on destructor if constructor was not successful

### DIFF
--- a/python/vowpalwabbit/pyvw.py
+++ b/python/vowpalwabbit/pyvw.py
@@ -277,6 +277,10 @@ class vw(pylibvw.vw):
     """The pyvw.vw object is a (trivial) wrapper around the pylibvw.vw
     object; you're probably best off using this directly and ignoring
     the pylibvw.vw structure entirely."""
+    log_wrapper = None
+    parser_ran = False
+    init = False
+    finished = False
 
     def __init__(self, arg_str=None, enable_logging=False, **kw):
         """Initialize the vw object.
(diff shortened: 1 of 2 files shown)


ataymano:
#2820

python/vowpalwabbit/pyvw.py
```diff
@@ -277,6 +277,10 @@ class vw(pylibvw.vw):
     """The pyvw.vw object is a (trivial) wrapper around the pylibvw.vw
     object; you're probably best off using this directly and ignoring
     the pylibvw.vw structure entirely."""
+    log_wrapper = None
```
jackgerrits, line 280: Does this work outside of the `__init__` function?

ataymano: yes, it is called class attributes

ataymano: actually I see that I haven't moved all of them here (at least forgot log_fwd). If approach is good, will send update and make it more consistent

jackgerrits: Which version were that added in?

ataymano: I do not see version where it is unavailable. 2.7, 3.5+ all have it.

jackgerrits: Awesome!

jackgerrits (approved):

PR #2828, opened 2021-03-02
Fix confusing API - dealloc_example

jackgerrits:
The old `dealloc_example` call would destruct but not free, was confusing to use and did not match with the alloc_examples call that was right next to it. This fixes that up

### DIFF
--- a/vowpalwabbit/example.cc
+++ b/vowpalwabbit/example.cc
@@ -314,10 +314,7 @@ example* alloc_examples(size_t, size_t count)
 {
   example* ec = calloc_or_throw<example>(count);
   if (ec == nullptr) return nullptr;
-  for (size_t i = 0; i < count; i++)
-  {
-    new (ec + i) example;
-  }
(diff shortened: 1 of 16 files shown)


vowpalwabbit/example.cc
```diff
@@ -314,19 +314,18 @@ example* alloc_examples(size_t, size_t count)
 {
   example* ec = calloc_or_throw<example>(count);
   if (ec == nullptr) return nullptr;
-  for (size_t i = 0; i < count; i++)
-  {
-    ec[i].ft_offset = 0;
-  }
+  for (size_t i = 0; i < count; i++) { new (ec + i) example; }
```
olgavrou, line 317: this doesn't need to be `new (ec + i) example()`?

olgavrou (approved):

vowpalwabbit/example.cc
jackgerrits: I think the parentheses can be omitted when there are no args

PR #2829, opened 2021-03-02
add scaffolding for adding an external (binary) parser

olgavrou:


vowpalwabbit/parser.cc
```diff
@@ -571,6 +575,14 @@ void enable_sources(vw& all, bool quiet, size_t passes, input_options& input_opt
         all.example_parser->reader = VW::parsers::flatbuffer::flatbuffer_to_examples;
       }
 #endif
+
+#ifdef BUILD_EXTERNAL_PARSER
+      else if (input_options.external_parser)
+      {
+        all.external_parser = VW::make_unique<VW::external_parser>(&all);
+        all.example_parser->reader = VW::parse_examples;
```
kumpera, line 583: Don't you need to call into the external parser somewhere?

kumpera (commented):
I love this approach, a few thoughts.

The CLI flag should be something that's more future proof. In particular, in case more than one external parsers is developed or one is moved into VW. For example `--format=binary-thingy`.

This extensions mechanism will need to be eventually documented as it will be part of our public API surface.

It would be interesting if/how this could be used by our bindings. @lalo what's your take on this?

vowpalwabbit/parser.cc
olgavrou: vw calls the parser you provide [here](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/parse_dispatch_loop.h#L22) during parsing

olgavrou:
> The CLI flag should be something that's more future proof. In particular, in case more than one external parsers is developed or one is moved into VW. For example `--format=binary-thingy`.
> 

@kumpera They way this is structured right now VW can accept (compile) only one external parser at a time, and that is why it is generic.
The idea is that if an external parser is provided then the compiler flag needs to be switched on, the cli flag passed, and at least the header file with the corresponding name and parser interface should be provided. It is not tied to the binary parser specifically so it shouldn't have to know that explicitly in the cli args

Not sure how to make this accept multiple external parsers at the same time without adding extra explicit compiler flags and `#ifdefs`

If a parser is moved into VW then it shouldn't be compiled with `BUILD_EXTERNAL_PARSER` and would probably have it's own flag like `--flatbuffers`, `--dsjson`, etc. Unless we want to move all of our parser flags into something like: `--format=[external | dsjson | json | flatbuffers]` otherwise we aren't being consistent

olgavrou:
@kumpera Oh no actually I think I might know how to compile more than one external parser that without extra compile time checks, will be right back with a amendments, thanks for the suggestion!

vowpalwabbit/parser.cc
lalo: the bindings would probably not work with this, bc this is being extended on compile time. You would need to compile another python package with this underlaying forked vw binary.

lalo: Regarding the idea of more extendable args, it could be possible if you have a generic hook for 'plugins' where vw has a call_external_arg_parser() and as long as a plugin fulfills that contract, it could work. This way the new flag would be defined on the other side, and the end result should look the same. That would allow you to have more generic names on flags that you can later migrate to vw itself, without breaking whoever was consuming the forked binary before.

vowpalwabbit/parse_args.cc
```diff
@@ -445,7 +445,9 @@ input_options parse_source(vw& all, options_i& options)
                      "A^B^C. Note: this will become the default in a future version, so enabling this option will "
                      "migrate you to the new behavior and silence the warning."))
       .add(make_option("flatbuffer", parsed_options.flatbuffer)
-               .help("data file will be interpreted as a flatbuffer file"));
+               .help("data file will be interpreted as a flatbuffer file"))
+      .add(make_option("external_parser", parsed_options.external_parser)
```
lalo, line 449: If all this is being extended at compile time, might as well have a cmake flag to enable this forked binary mode and hide this code behind that flag. Sort of like the flatbuffers being compiled out, this would be external parser compiled out.

olgavrou: good point, done

vowpalwabbit/parser.cc
olgavrou: added `VW::external::parser::set_parse_args()` in `parse_args.cc`

external parser adds whatever options it wants see [here](https://github.com/VowpalWabbit/reinforcement_learning/pull/265/files#diff-5617db68ef0406524f96b6a49b8526b756796e20e65a3fb31019947f924826faR23)

the contract that the external parser has to fill (in general, not only for args) can be seen [here](https://github.com/VowpalWabbit/reinforcement_learning/pull/265/files#diff-959fe33f0cd6431dc062b3357fda604b7b3fb4e47aad4fc1d25e31ddf3e42327R1)

kumpera (approved):
LGTM

jackgerrits (approved):

lalo (approved):

PR #2833, opened 2021-03-03
add quiet check before outputting to trace_message

olgavrou:


vowpalwabbit/parse_args.cc
```diff
@@ -775,7 +775,7 @@ void parse_feature_tweaks(
     {
       if (options.was_supplied("leave_duplicate_interactions"))
       { all.interactions.leave_duplicate_interactions = true; }
-      else
+      else if (!all.logger.quiet)
```
lalo, line 778: should warnings be quiet-eable?

olgavrou: all the other ones in parse args are, so this should probably follow along

jackgerrits (approved):

PR #2837, opened 2021-03-05
[WIP] Remove the inner v_array from io_buf

jackgerrits:


peterychang:
There's a lot of duplicated code in here. I think it would be better to wrap the buffer into a private class that has the functionality we need to use. I'll be a lot more readable and maintainable that way

jackgerrits:
Did some reworking - it's not completely encapsulated but for the purposes of this class it seems a little better

vowpalwabbit/io_buf.h
```diff
@@ -10,54 +10,105 @@
...
+      else
+      {
+        _begin = temp;
+      }
+      _end = _begin + old_size;
+      _end_array = _begin + new_capacity;
+      memset(_end, 0, sizeof(char) * (_end_array - _end));
+    }
+
+    void shift_to_front(char* new_begin)
```
peterychang, line 73: `new_begin` is a confusing name. I'd say either rename it to something like `head_ptr` or maybe move the pointer inside the buffer object?

peterychang (approved):

vowpalwabbit/io_buf.h
```diff
@@ -10,54 +10,105 @@
...
+    {
+      // This specific internal buffer should only ever grow.
+      assert(new_capacity >= capacity());
+      const auto old_size = size();
+      char* temp = reinterpret_cast<char*>(std::realloc(_begin, sizeof(char) * new_capacity));
+      if (temp == nullptr)
+      {
+        // _begin still needs to be freed but the destructor will do it.
+        THROW_OR_RETURN("realloc of " << new_capacity << " failed in resize().  out of memory?");
+      }
```
olgavrou, line 63: super nit:

don't really need the `else` since if `temp == nullptr` we will throw or return

vowpalwabbit/io_buf.h
```diff
@@ -10,54 +10,105 @@
...
 **
-** The interval [space.head, space.end] may be shifted down to space.begin
+** The interval [head, _buffer._end] may be shifted down to _buffer._begin
 ** if the requested number of bytes to be read is larger than the interval size.
 ** This is done to avoid reallocating arrays as much as possible.
 */
 
 class io_buf
 {
+  struct internal_buffer
```
olgavrou, line 45: do you think it is worth adding a comment here explaining that:

io_buf re-implements a grow-only variant of `v_array` in this `internal_buffer` since it needs to access the array's internals? Or maybe call `internal_buffer` `expand_only_v_array`?

olgavrou (approved):

vowpalwabbit/io_buf.h
jackgerrits: Good point! This can be fixed in v_array too but ill leave that one for later

vowpalwabbit/io_buf.h
jackgerrits: Added a comment explaining why this is here

PR #2843, opened 2021-03-09
Implement v_array::insert for single elements

jackgerrits:


vowpalwabbit/v_array.h
```diff
@@ -66,6 +66,14 @@ struct v_array<T, typename std::enable_if<std::is_trivially_copyable<T>::value>:
     memset(_end, 0, (end_array - _end) * sizeof(T));
   }
 
+  // This will move all elements after idx by width positions and reallocate the underlying buffer if needed.
+  // The size must be updated prior to calling this
```
olgavrou, line 70: Does the comment: `The size must be updated prior to calling this` correspond to updating `_end` by `1` during insert?

olgavrou (approved):

vowpalwabbit/v_array.h
jackgerrits: Yes

vowpalwabbit/v_array.h
```diff
@@ -66,6 +66,14 @@ struct v_array<T, typename std::enable_if<std::is_trivially_copyable<T>::value>:
...
+  void make_space_at(size_t idx, size_t width)
+  {
+    if (size() + width > capacity()) { reserve(2 * capacity() + width); }
+    memmove(&_begin[idx + width], &_begin[idx], (size() - (idx + width)) * sizeof(T));
```
peterychang, line 74: I think the size calculation is wrong?
If you want to insert 2 elements at the beginning, your memmove would be
`memmove(&_begin[2], &_begin[0], size() - 2)`, so you'd end up overwriting the last 2 elements of your array, right?

peterychang: wait, nevermind.. missed the comment (olga even commented on it!) where the size has to be updated first

jackgerrits: Yeah it is a little weird - you need to extend the end and them shift down into it.

Issue #2834, opened 2021-03-03
--coin --synthcover is not working


ataymano:
#### Describe the bug
--cb_explore_adf --coin --synthcover is not working in many simple cases

#### To Reproduce
```
def generate(fname, count):
    examples = [
            [
                'shared |s f1',
                '0:-3:0.5 |a f1',
                '|a f2'
            ],
            [
                'shared |s f1',
                '|a f1',
                '1:-1:0.5 |a f2'
            ]
        ]
    with open(fname, 'w') as f:
        for i in range(count):
            for l in examples[i % 2]:
                f.write(l + '\n')
            f.write('\n')
```

#### Expected behavior
We should learn that action 0 is the best and average loss should converge to 3 (or 3 - exploration)

#### Observed Behavior
--coin --synthcover stuck in [0.5, 0.5] state
![image](https://user-images.githubusercontent.com/41013086/109837463-3330ef80-7c13-11eb-8e7a-ea3e9245d80c.png)

####  How did VW behave? Please include any stack trace, log messages or crash logs.
Here:
https://github.com/VowpalWabbit/vowpal_wabbit/blob/4ac29b313299900e5eca36e747cfdc49f250761c/vowpalwabbit/cb_explore_adf_synthcover.cc#L102
we are clipping our predicted cost by min/max of logged costs which make sense only under assumption that predicted cost should converge to the real one.
That seems to be not true for --coin - we have some constant factor there. In this specific case it seems to tend to [-9, -3] which is going out of [-3, -1] and causing both costs to be clipped to -3.

```
generate('test.txt, 100000)

vw --cb_explore_adf -d test.txt --invert_hash inv_hash.txt --coin --synthcover


inv_hash.txt:
:0
a^f1:47870:-4.92344
a^f2:91850:1.07656
Constant:116060:-2.03828
s^f1:205995:-2.03828
```

Haven't checked yet what is going on inside coin, but is it fair to rely on that convergence in synthcover (@pmineiro)?

pmineiro:
synthcover has a hyperparameter which is the "reward difference allowed to
explore over".  this might need a setting different than default for this
problem.  let's discuss.

PR #2844, opened 2021-03-09
coin weights fix

ataymano:
If we are using coin for any regression, weights in the model are not giving us predictions that converge to observed costs, but have some constant multiplier (weighted average of all examples squared norms).
Proof:
```
def generate(fname, count):
    with open(fname, 'w') as f:
        for i in range(count):
            f.write('2:0.5 | f1 f2 f3\n')

generate('data.txt', 1000000)

vw -d data.txt --invert_hash model_no_coin.txt
vw -d data.txt --coin --invert_hash model_coin.txt

model_no_coin.txt:
f2:82936:0.5
Constant:116060:0.5
f3:222792:0.5
f1:259761:0.5

0.5 + 0.5 + 0.5 + 0.5 = 2

model_coin.txt:
:82936:2
:116060:2
:222792:2
:259761:2

2 + 2 + 2 + 2 = 8

```

Reason for that is the following:
coin is implemented as mode of ftrl, and there are 2 predict methods are used here:
[ftrl::predict](https://github.com/VowpalWabbit/vowpal_wabbit/blob/f105fbf26c9bcc0025fd2547299db67144201a79/vowpalwabbit/ftrl.cc#L77) (which is pretty much GD::predict)
[coin_betting_predict](https://github.com/VowpalWabbit/vowpal_wabbit/blob/f105fbf26c9bcc0025fd2547299db67144201a79/vowpalwabbit/ftrl.cc#L209) - which is used only inside learn_coin_betting and expected to return unnormalized prediction (according to [this](https://github.com/VowpalWabbit/vowpal_wabbit/blob/f105fbf26c9bcc0025fd2547299db67144201a79/vowpalwabbit/ftrl.cc#L219))

So every stack that is relying on predictions from base::predict has to deal with scaled predictions (and may fail if it relies on them to converge to logged costs, like #2834).

Here is quick patch which is about explicitly normalizing prediction before returning from ftrl::predict.
Still looking how to do it properly, but please let me know if you have suggestions.

@bremen79, it would be amazing if you can take a look and suggest something here (or is there a paper with details of this algorithm variation exists somewhere?).


ataymano:
This diff is completely fixing example reported in #2834.

Results on cbified 500 OpenML classification data sets:

1) Epsilon-greedy + coin
image

2) SquareCB + coin
image

3) Bagging + coin
image

4) Softmax + coin
image

5) Synthcover + coin
image

6) Cover + coin
image


So, cover and synthcover got better, but is not fixed completely. Looking more into it.

bremen79:
Hi, I'll try to take a look this weekend.

bremen79:
I have taken a look at the code: the change looks good to me.
Regarding the source of this algorithm: there is no paper for this variation. The main algorithm is COCOB (that is published), but then there are a bunch of tricks: monitoring the range of the samples, the Lipschitz constant and the normalization you fixed. The normalization actually comes from the code of John of SGD: I added it because empirically it improved the performance, but I don't have a clear reason on why this is a good idea...

ataymano:
> I have taken a look at the code: the change looks good to me.
> Regarding the source of this algorithm: there is no paper for this variation. The main algorithm is COCOB (that is published), but then there are a bunch of tricks: monitoring the range of the samples, the Lipschitz constant and the normalization you fixed. The normalization actually comes from the code of John of SGD: I added it because empirically it improved the performance, but I don't have a clear reason on why this is a good idea...

Thank you for reviewing!
Some tests are failing (because of changed stdout) - still going to double check that there is nothing wrong with them and will merge later this week.

jackgerrits (approved):

ataymano:
Fixes #2834

PR #2846, opened 2021-03-09
Use insert and erase in interact

jackgerrits:


jackgerrits:
I'd like to add an interact test before merging this

PR #2848, opened 2021-03-10
Add json parsing for dedup-ed examples

olgavrou:
Implements parsing for dedup, avoiding extra json parsing, if we are getting what we call deduplicated actions from json.

Example:

First receive deduplication dictionary:

```
[4282062864] -> {"TAction":{"a1":"f1"},"_slot_id":0}
[4199675127] -> {"TAction":{"a2":"f2"},"_slot_id":1}
```

The dictionary's actions can be parsed resulting in example objects, and we can store those example-object pointers in a dedup map `std::unordered_map<uint64_t, example*> dedup_examples`.

We then receive the below payload where the actions have been replaced by their action dedup id's. We can use the `deduped_examples` map to copy the already parsed payloads. e.g.:

```
{
  "GUser":{"id":"a","major":"eng","hobby":"hiking"},
  "_multi":[{"__aid":4282062864},{"__aid":4199675127}],
  "_slots":[{"Slot":{"a1":"f1"}},{"Slot":{"a2":"f2"}}]
}
```

How the examples in the dedup map are managed and/or cleaned is not the json parser's responsibility


vowpalwabbit/parse_slates_example_json.h
```diff
@@ -148,65 +148,106 @@ void handle_features_value(const char* key_namespace, const Value& value, exampl
...
-    ex->l.slates.slot_id = slot_id;
-    handle_features_value(" ", obj, ex, namespaces, all);
-    assert(namespaces.size() == 0);
+    const auto& multi = context["_multi"].GetArray();
+    if (dedup_examples && !dedup_examples->empty())
+    {
+      for (const Value& obj : multi)
+      {
+        auto dedup_id = obj["__aid"].GetUint64();
+        auto* stored_ex = (*dedup_examples)[dedup_id];
```
kumpera, line 178: We need to handle this failing to find.

kumpera (approved):

vowpalwabbit/parse_slates_example_json.h
olgavrou: good point, I made it throw (at least for now) since that's what the json parser does on malformed json input

PR #2850, opened 2021-03-10
Add test for interact reduction

jackgerrits:


olgavrou (commented):
Is it worth checking the interactions by using --invert_hash?

jackgerrits:
This doesn't add interactions in that sense - but it adds new features. So maybe a readable model here is a good idea?

olgavrou:
> This doesn't add interactions in that sense - but it adds new features. So maybe a readable model here is a good idea?

Yeah my understanding of `--invert_hash` is that it's the same as `--readable_model` but instead of hashes and weights it gives use user-readable features and weights. I guess either would do

jackgerrits:
Added readable model

olgavrou (approved):

PR #2855, opened 2021-03-10
Do a fuzzy float compare instead of exact

jackgerrits:


jackgerrits:
I believe the difference emerged when we hit this situation. In the past the `fill_tied` function only ever really worked when  positions in the prediction were exact. Practically speaking this was only ever for `0.0`. But now this case hits.

The question is - for the purpose of `fill_tied` are they first two in this list "equal"?


  Name | Value | Type
 -- | -- | --
 | [0] | {action=1 score=0.399922609 } | ACTION_SCORE::action_score
 | [1] | {action=2 score=0.399999976 } | ACTION_SCORE::action_score
 | [2] | {action=0 score=1.00033188 } | ACTION_SCORE::action_score

olgavrou (approved):

peterychang:
The only real objection I have is that is_same breaks the transitive property. Namely if a==b and b==c, a==c may not hold true.. I'm not sure if this matters in our code, but its something we should remain aware of

peterychang:
Also, i'm not sure if we ever compare containers containing floats (eg: `vector<float> == vector<float>`), but that would also break

jackgerrits:
I'm not aware of us doing that. This isn't a systemic change, just for one place. I don't think the transitive properly is a problem in this place.

peterychang (approved):

jackgerrits:
Went over this with Paul and he said it looks good. We moved from abs to rel float comparison

PR #2856, opened 2021-03-11
fb off by default in linux build

olgavrou:
Since it is off by default on windows build too (since the user has to specify environment variables in order for fb's to be enabled on windows), and it is causing too much trouble when it is ON by default


jackgerrits (approved):

lalo (approved):
do we need to update the wiki? lgtm

olgavrou:
> do we need to update the building wiki? lgtm

Yes, just updated stating that flatbuffers is optional

PR #2859, opened 2021-03-11
add vw_core.lib dependency on unit test build

olgavrou:


olgavrou:
closing for now, will try to solve this later

PR #2861, opened 2021-03-12
[cbify] cbify_adf_data refactor

lalo:
Prompted by https://github.com/VowpalWabbit/vowpal_wabbit/pull/2680

cbify_adf_data can be reused completely and we just need the convenience of all the related functions in the same place.

### DIFF
--- a/vowpalwabbit/cbify.h
+++ b/vowpalwabbit/cbify.h
@@ -4,6 +4,25 @@
 
 #pragma once
 #include "reductions_fwd.h"
+#include "example.h"
+
+template <typename T>
+void v_move(v_array<T>& dst, v_array<T>& src)
+{
+  dst = src;
+  src = v_init<T>();
+}
+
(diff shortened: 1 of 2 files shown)


vowpalwabbit/cbify.h
```diff
@@ -4,6 +4,25 @@
 
 #pragma once
 #include "reductions_fwd.h"
+#include "example.h"
+
+template <typename T>
+void v_move(v_array<T>& dst, v_array<T>& src)
```
jackgerrits, line 10: this is just doing a copy now - it should be removed and either copy or move constructor used

jackgerrits (approved):

vowpalwabbit/cbify.h
lalo: done

jackgerrits (approved):

PR #2868, opened 2021-03-15
[test] RunTests VW options coverage script

lalo:
Script that generated this: https://github.com/VowpalWabbit/vowpal_wabbit/issues/2867


olgavrou (approved):
super duper nit, feel free to ignore: could we group the `BUT has default value` ones from the ones that are not tested at all?

test/runtest_coverage.py
```diff
@@ -0,0 +1,71 @@
+from vowpalwabbit import pyvw
+
+def get_latests_tests():
```
jackgerrits, line 3:
```suggestion
def get_latest_tests():
```

test/runtests_parser.py
```diff
@@ -234,12 +234,17 @@ def file_to_obj(filename):
 
     return results
 
-def main():
+def find_runtest_file():
+    rtfile = None
     possible_paths = ["./RunTests", "./test/RunTests"]
```
jackgerrits, line 239: Consider making this not dependent on cwd?

lalo: done

PR #2869, opened 2021-03-15
Replace strncpy wit memcpy as the former is flagged as unsafe by linters.

kumpera:


olgavrou (commented):
would anything from this cross_plat file work for string copying? https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/crossplat_compat.h

kumpera:
> 
> 
> would anything from this cross_plat file work for string copying? https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/crossplat_compat.h

On this case, we don't need a string copy since we already figured out the size.

And OMG, no, the unix implementation of string_cpy is bad AF. Got to fix that one as well.

PR #2871, opened 2021-03-16
Add nix shell to repo

jackgerrits:


jackgerrits:
Not enough interest in this cool tech - I'll just keep it in my local copy for now then

PR #2873, opened 2021-03-16
Fix calloc_reserve and push_many

jackgerrits:


vowpalwabbit/parse_example.cc
```diff
@@ -292,7 +292,7 @@ class TC_parser
             spelling_v.push_back(_index);
             spelling_v.push_back('_');
           }
-          push_many(spelling_v, spelling_strview.begin(), spelling_strview.size());
+          spelling_v.insert(spelling_v.begin(), spelling_strview.begin(), spelling_strview.end());
```
peterychang, line 295: inserting to the beginning of spelling_v?

jackgerrits: Ah it should be end thanks!

vowpalwabbit/v_array.h
```diff
@@ -429,24 +429,18 @@ void copy_array(v_array<T>& dst, const v_array<T>& src, T (*copy_item)(const T&)
...
 }
 
 template <class T>
+VW_DEPRECATED("calloc_reserve is no longer supported. You should use appropriate constructors instead.")
 void calloc_reserve(v_array<T>& v, size_t length)
 {
-  v._begin = calloc_or_throw<T>(length);
-  v._end = v._begin;
-  v.end_array = v._begin + length;
+  v.reserve(length);
```
peterychang, line 442: We should call clear() before we reserve

jackgerrits: Prior to this change it would leak if there were anything inside so this is kind of relying on a user passing in a default v_array. That being said I see no harm in clearing

vowpalwabbit/parse_example.cc
```diff
@@ -491,8 +491,8 @@ void substring_to_example(vw* all, example* ae, VW::string_view example)
     {
       VW::string_view tag = all->example_parser->words.back();
       all->example_parser->words.pop_back();
-      if (tag.front() == '\'') tag.remove_prefix(1);
-      push_many(ae->tag, tag.begin(), tag.size());
+      if (tag.front() == '\'') { tag.remove_prefix(1); }
+      ae->tag.insert(ae->tag.begin(), tag.begin(), tag.end());
```
peterychang, line 495:
```suggestion
      ae->tag.insert(ae->tag.end(), tag.begin(), tag.end());
```

peterychang (approved):

PR #2874, opened 2021-03-16
Make v_array internals private, don't allow begin/end to be modified

jackgerrits:
Required #2873


vowpalwabbit/feature_group.cc
```diff
@@ -30,16 +30,16 @@ void features::clear()
 void features::truncate_to(const features_value_iterator& pos)
 {
   auto i = pos._begin - values.begin();
```
kumpera, line 32: Could we make this safe? If `pos` didn't come from `values` we have UD.

vowpalwabbit/lrqfa.cc
```diff
@@ -125,14 +125,8 @@ void predict_or_learn(LRQFAstate& lrq, single_learner& base, example& ec)
...
-      rfs.values.end() = rfs.values.begin() + lrq.orig_size[right];
-
-      if (all.audit || all.hash_inv)
-      {
-        for (size_t j = lrq.orig_size[right]; j < rfs.space_names.size(); ++j) rfs.space_names[j].~audit_strings_ptr();
-
-        rfs.space_names.end() = rfs.space_names.begin() + lrq.orig_size[right];
-      }
+      rfs.values.actual_resize(lrq.orig_size[right]);
+      if (all.audit || all.hash_inv) { rfs.space_names.resize(lrq.orig_size[right]); }
```
kumpera, line 129: Why is the dtor gone?

vowpalwabbit/v_array.h
```diff
@@ -75,12 +75,9 @@ struct v_array<T, typename std::enable_if<std::is_trivially_copyable<T>::value>:
     memmove(&_begin[idx + width], &_begin[idx], (size() - (idx + width)) * sizeof(T));
   }
 
-public:
-  // private:
```
kumpera, line 79: <3 but commented code.

kumpera (approved):
A few questions and an out of place review of the previous PR that introduced actual_resize.

That name is pretty bad as it's not obvious what it does. I think a better naming for those placeholders are on the form `xxx_but_yyy` or `xxx_with_yyy` so, for example, it could be `resize_with_stl_behavior`, which still keeps the sense of impropriety of its existence.

Finally, that PR makes VW 10 promises but didn't record it in the VW 10 project board.

vowpalwabbit/lrqfa.cc
jackgerrits: `rfs.space_names` is a std::vector Because the destructor is handled by std::vectors resize. I don't know how this code worked before if at all...

vowpalwabbit/v_array.h
jackgerrits: Welp it is private for real now

vowpalwabbit/feature_group.cc
jackgerrits: features_value_iterator is created with a ptr, so we'd likely need to enforce in its creation. In this scope I do not think we have to information to make this safe. I think this should be handled separately

vowpalwabbit/kernel_svm.cc
```diff
@@ -179,9 +179,7 @@ int svm_example::compute_kernels(svm_params& params)
 
 int svm_example::clear_kernels()
 {
-  int rowsize = (int)krow.size();
-  krow.end() = krow.begin();
-  krow.resize(0);
+  int rowsize = static_cast<int>(krow.size());
```
olgavrou, line 182: does `krow` still need to be cleared here?

vowpalwabbit/v_array.h
```diff
@@ -75,12 +75,9 @@ struct v_array<T, typename std::enable_if<std::is_trivially_copyable<T>::value>:
...
+private:
   T* _begin;
   T* _end;
-
-public:
   T* end_array;
 
 private:
```
olgavrou, line 83: probably don't need this second private anymore

PR #2877, opened 2021-03-16
[py] add enabled_reductions api

lalo:


python/pylibvw.cc
```diff
@@ -298,6 +298,14 @@ std::string get_arguments(vw_ptr all)
   return serializer.str();
 }
 
+py::list get_enabled_reductions(vw_ptr all)
+{
+  py::list enabled_reductions;
+  for (auto ex : all->enabled_reductions) { enabled_reductions.append(ex); }
+
+  return enabled_reductions;
```
jackgerrits, line 306: What is the type in the list?

lalo: std::vector<std::string>>

olgavrou (approved):

jackgerrits (approved):

PR #2879, opened 2021-03-16
label initialization

jackgerrits:


vowpalwabbit/multiclass.cc
```diff
@@ -13,10 +13,19 @@
 
 namespace MULTICLASS
 {
+  label_t::label_t() { reset_to_default(); }
```
olgavrou, line 16: maybe `set_to_default` instead? it's counter intuitive for a constructor to `reset`

olgavrou (approved):

vowpalwabbit/multiclass.cc
jackgerrits: Ah set is much better, thanks!

PR #2882, opened 2021-03-17
[Flatbuffers] actually skip tests that flatbuffers don't convert in flatbuffers run_test

### LINKED ISSUE #2804: [test] skipped flatbuffer test are shown as success in the printout
The test that are skipped using the python runtest driver are shown as successfully ran when they should be reported as skipped. 

See any of the skipped tests and verify with the output: https://github.com/VowpalWabbit/vowpal_wabbit/blob/cb37da4553c475b4d563e22f43194adc8f127cd2/test/run_tests.py#L521

olgavrou:
Resolves #2804


jackgerrits:
Does it make sense for these tests to be reported as skipped in the test result framework? That way any tests with dependencies will behave correctly

olgavrou:
> Does it make sense for these tests to be reported as skipped in the test result framework? That way any tests with dependencies will behave correctly

Yes good point, amended

Issue #2866, opened 2021-03-12
--ccb_explore_adf cannot be initialized with cb model


ataymano:
#### Describe the bug
CB model is not readable by CCB  vw instance (while it should in order to treat CB data as single-slot CCB)

#### To Reproduce
vw --cb_explore_adf --save_resume -d cb_data.txt -f cb_model.bin
vw --ccb_explore_adf --save_resume -d ccb_data.txt -i cb_model.bin

#### Expected behavior
Both executions should be succesfull

#### Observed Behavior
Error: Model content is corrupted, weight vector index 4228632 must be less than total vector length 262144

Originally CB and CCB models were compatible, but recently one extra field was introduced to ccb header: 
https://github.com/VowpalWabbit/vowpal_wabbit/blob/713c5896f58f0186a4510d10b059a413d48af058/vowpalwabbit/conditional_contextual_bandit.cc#L598
This field is read conditionally on model version only, and making it conditional on whether model is cb or ccb should fix the problem.

#### Environment
Current vw master on all platforms

PR #2884, opened 2021-03-17
cb model 2 ccb trainer 

ataymano:
It is super dirty, but unfortunately I do not see better quick solutions here.
Other option is to replace new ccb option with extra field in all which seems worse to me.
Please let me know if you see better ways to solve it.
If we are ok with this approach, I will add more tests.


ataymano:
Fixes #2866

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -621,7 +622,8 @@ base_learner* ccb_explore_adf_setup(options_i& options, vw& all)
                .necessary()
                .help(
                    "EXPERIMENTAL: Do Conditional Contextual Bandit learning with multiline action dependent features."))
-      .add(make_option("all_slots_loss", all_slots_loss_report).help("Report average loss from all slots"));
+      .add(make_option("all_slots_loss", all_slots_loss_report).help("Report average loss from all slots"))
+      .add(make_option("ccb_model_without_has_seen_flag", data->input_model_without_has_seen_multi_slot_flag).help("Legacy ccb model format without HasSeenMultipleSlots flag"));
```
olgavrou, line 626: does this have to be exposed to the end user? Could we do the `is_ccb_run` and `is_ccb_model` checks in conditional_contextual_bandit setup and set the boolean there?

ataymano: we are merging options from command line and model before entering stack of setups, so when we first time entersconditional_contextual_bandit::setup information about which options came from model and which are from command line is already lost (and there is no way to populate is_ccb_run and is_ccb_model here).

olgavrou: I see, good point... would something like `ccb_with_cb_model` work?

jackgerrits: This option could be hidden too

ataymano: what does it mean?

jackgerrits: @lalo I thought we added hidden? I can't seem to find it though

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -68,6 +68,7 @@ struct ccb
   // If the reduction has not yet seen a multi slot example, it will behave the same as if it were CB.
   // This means the interactions aren't added and the slot feature is not added.
   bool has_seen_multi_slot_example = false;
+  bool input_model_without_has_seen_multi_slot_flag = false;
```
jackgerrits, line 71: Can you add a comment explaining this?

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits, line 626: Is this options meant to be stored in the model? Should it be 'keep'?

lalo: I deleted it, but the impl is pretty brief see commit: https://github.com/VowpalWabbit/vowpal_wabbit/commit/4ee50a8ddbea104013f9d8540b6ea2902bb9acfe

options.h and options_boost_po.cc

vowpalwabbit/parse_args.cc
```diff
@@ -1593,7 +1601,9 @@ void merge_options_from_header_strings(
       }
     }
   }
-
+  if (is_ccb_run & !is_ccb_model) {
+    options.insert("ccb_model_without_has_seen_flag", "");
```
jackgerrits, line 1605: This does not verify the has seen flag was not added though

vowpalwabbit/parse_args.cc
```diff
@@ -1593,7 +1601,9 @@ void merge_options_from_header_strings(
       }
     }
   }
-
+  if (is_ccb_run & !is_ccb_model) {
```
lalo, line 1604: is this & the behavior wanted? or typo for &&?

vowpalwabbit/parse_args.cc
lalo: I think its only checking if you have the option ccb_explore_adf on the command line and ccb_explore_adf (or not) on the model file. So I'm assuming a model file that is CB related can be override to be ccb_explore_adf if you add that arg (and behind the scenes it adds ccb_model_without_has_seen_flag to make sure it can parse the model correctly). Is this interpretation correct Alexey?

vowpalwabbit/conditional_contextual_bandit.cc
lalo: *although it was for options groups and not individual options

vowpalwabbit/parse_args.cc
ataymano: Kind of.
Before HasSeenMoreThanOneSlot flag was introduced  there was no difference between cb and ccb models. So basically cb models of vw from any version are equivalent (in terms of format) to ccb models of old vw. So logic which that we need here is the following:
if we are going to make ccb decisions, but input model is cb, it should be loaded as cb (~ legacy ccb) model.
In terms of options logic should be the following:
if model is cb and command line is ccb:
    set ccb_model_without_has_seen_flag

we can check if this flag is set already and not set it in this case, but not sure about benefit of that.

vowpalwabbit/conditional_contextual_bandit.cc
ataymano: I do not think so. Option is used only in scenario: --ccb_explore_adf -i cb_model.bin
Since we are doing --ccb_explore_adf output model is ccb model, and this flag is not applicable there.

vowpalwabbit/conditional_contextual_bandit.cc
ataymano: sure will do this.

vowpalwabbit/conditional_contextual_bandit.cc
ataymano: sorry.
other scenario is backward compatibility one - when we are running new vw in ccb mode using old ccb model.
I still feel that it is better that model should be produced in new format (since it is new vw), but it seems a bit open question actually.

jackgerrits (approved):

Issue #1579, opened 2018-08-27
Java jni including and using native libraries inside jar


ampm78:
The released Java jni currently includes the MacOs native library as a resource inside the jar, but this is not actually used (it will still look for the native library in the "java.library.path"). This forces users to install the native library in their systems if they want to use the vw_jni.

I pushed some code to extract the lib from inside the jar and load it in my clone repo (https://github.com/ampm78/vowpal_wabbit/commit/a957fa7f80ea4bee201619c73713466216ea192b) but this will only work at the moment for MacOs as that's the only native lib currently included in the jar. 

Ideally the native libraries for at least the major platforms (MacOs, Linux, Windows?) should be built and included in the jar when releasing it, as other java libraries do (for example tensorflow).

JohnLangford:
@jmorra @deaktator your thoughts here?

deaktator:
@JohnLangford @jmorra @ampm78: I think this depends on the status of the build process. Previous to 8.4.1, we included the binaries for several OS variants in the jar. To the best of my recollection, the primary reason these were removed was related to two primary factors:  

1. the release process for Java was independent of the C lib release process.  
1. At the time, boost program options was dynamically linked in the C build.

These issues led to version conflicts with the boost program opts library and made it extremely hard to use because the library would be tied to the boost lib against which it was built. You’d have to simlink multiple boost libs and change them to make the JNI lib work.

Some of this detail is in the readme file is the java dir in the VW project.

If we can statically link the boost library in the C build and use a CI/CD service to release everything, then we can run the docker script we used previously to build the library for several OSs. I think that Travis has OS support which would allow us to build for major Linux distros and OS X. Not sure about windows support.

The OS detection code was a little “quirky” (if I remember correctly) and it’s often difficult to distinguish between some Linux distros.

@JohnLangford it’s been a while since I looked at the current status of the build process. Do you think we can build and release via Travis? And can we statically link against boost program opts?

JohnLangford:
I'm unsure, but several more people are becoming involved so it's possible we can engineer a more comprehensive release process.

andrusha:
It would be convenient to only need to include vw as java dep and then it would take care of the rest. This would enable / ease the use of VW on services like Dataflow.

eisber:
[...]

PR #2886, opened 2021-03-17
[Java] Support loading the VW JNI library from JAR file.

zwd-ms:
For Java API, the native library is packed inside the JAR file, but it is not loaded. Instead, the code tries to load the JNI lib from `java.library.path`. This PR adds loading from JAR for the Java API, which is already used in the spark API.

- move the native lib loader to a common package
- add loading from JAR in case the JNI lib is not found in java.library.path

Resolves issue #1579 (for linux_64).


jackgerrits:
So this resolves lib loading for the original java bindings on linux only right? We'll need to add a bit more for macos and windows?

zwd-ms:
That's right, the lib loader hardcodes the directory for linux_64, `natives/linux_64/libvw_jni.so`. MacOS and Win still need to manually specify the `java.library.path` for now.

Also, the published jar only has linux_64 lib. We haven't changed that if there is a plan.

zwd-ms:
Actually the windows build is [unsupported](https://github.com/VowpalWabbit/vowpal_wabbit/blob/e53b3a0d9e354427977a10d190b9b59d0132937b/CMakeLists.txt#L105). (Is there a reason?)

jackgerrits:
It's only unsupported as we havent had the time to make it work.

Some of the tests are not working on MacOS so I think we need to fix them before we can enable it

jackgerrits (approved):

PR #2894, opened 2021-03-18
Create a CMake based Windows CI

jackgerrits:
This only build C++ and only runs the unit tests. But it is MUCH faster than the existing Windows CI so allows for faster issue turnaround time


.github/workflows/build_windows_cmake.yml
```diff
@@ -0,0 +1,60 @@
...
+          key: |
+            ${{ env.VCPKG_REF }}-vcpkg-cache-invalidate-0
+      - uses: ilammy/msvc-dev-cmd@v1
+      - name: Bootstrap vcpkg
+        run: ${{ env.VCPKG_ROOT }}/bootstrap-vcpkg.bat
+      - name: Install vcpkg dependencies
+        run: ${{ env.VCPKG_ROOT }}/vcpkg.exe --triplet x64-windows install zlib boost-system boost-program-options boost-test boost-align boost-foreach boost-math flatbuffers
+      - name: Generate project files
+        run: |
+          cmake -S "${{ env.SOURCE_DIR }}" -B "${{ env.CMAKE_BUILD_DIR }}" -GNinja -DCMAKE_TOOLCHAIN_FILE="${{ env.VCPKG_ROOT }}/scripts/buildsystems/vcpkg.cmake" -DWARNINGS=Off -DUSE_LATEST_STD=On -DBUILD_FLATBUFFERS=On
```
olgavrou, line 53: should we have warnings on?

jackgerrits: For some reason CMake add /W3 and we add /W4 and so it results in a message for every single file that we are overriding warnings

jackgerrits: Ah we just need to set the policy

olgavrou (approved):

lalo (approved):

PR #2896, opened 2021-03-19
[test] draw reduction graph

lalo:


test/runtest_coverage.py
```diff
@@ -81,6 +105,8 @@ def main():
     
     print_non_supplied(allConfig)
 
+    # draw_graph(stacks)
```
lalo, line 108: commented out bc of the extra depedencies, but want to check in the code anyways

jackgerrits (approved):

PR #2898, opened 2021-03-19
add ability to use system installs for fmt, spdlog

peterychang:


ext_libs/ext_libs.cmake
```diff
@@ -0,0 +1,51 @@
+# Use position independent code for all external libraries
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+
+if(FMT_SYS_DEP)
+  # fmt was initially built against 7.1.3. Its possible earlier versions will also work, but that needs to be tested
+  find_package(fmt 7.1.3 REQUIRED)
```
jackgerrits, line 6: Does that mean exactly 7.1.3 or anything newer?

peterychang: That should be >= 7.1.3. The comment under rapidjson explains

jackgerrits (approved):

PR #2900, opened 2021-03-19
fix: fix cost assignment in cats tree

### LINKED ISSUE #2891: CATs fails to converge with fixed context and simple convex reward function
#### Describe the bug
Not sure if this is a bug, so feel free to close if not. I am running a simple test of CATs, with min_value 0 and max_value 1, where the cost is defined as follows:

- if the prediction is exactly 0.0, 0 cost
- if it is not, return the absolute distance/difference to 0.0 squared as cost

The context is fixed for each round, so I would assume this to be a pretty simple learning task. I've made a jupyter notebook that gridsearches a bunch of options (actions, bandwith, epsilon) and can't find a combination that satisfactorily predicts close to 0.5 when exploiting. Some parameter combinations result in wildly different average cost. Changing GD to FTRL doesn't help. Even if I ensure good exploration by changing epsilon, i see no reduction in accrued average cost per round over time.

So my question is, is this behaviour a feature of CATs, some bug, or something I am just missing in my code?

#### To Reproduce
Install a fresh pyvw wheel from master (pypi doesn't yet have CATS support) and run the attached notebook: 
[CATS-test.ipynb.zip](https://github.com/VowpalWabbit/vowpal_wabbit/files/6164450/CATS-test.ipynb.zip)

#### Expected behavior
I would expect predictions for such a simple use case to converge to the optimal prediction of 0.5 (0 cost)

#### Observed Behavior
 Observe the cost plots in the notebook not trending down over time, regardless of gridsearch combination

#### Environment
pyvw, OS X, latest built wheel from the master branch

olgavrou:
Resolves #2891


vowpalwabbit/cats_tree.cc
```diff
@@ -92,15 +92,15 @@ void min_depth_binary_tree::build_tree(uint32_t num_nodes, uint32_t bandwidth)
       if (bandwidth)
       {
         right_only = (id == (_num_leaf_nodes / (2 * bandwidth) - 1));
-        left_only = (id == (_num_leaf_nodes / (bandwidth)-2));
+        left_only = (id == (_num_leaf_nodes / (bandwidth) - 2));
```
lalo, line 95: why isnt there symmetry between right_only and left_only (as in same formula?)

olgavrou: it's derived from the [paper](https://arxiv.org/pdf/2006.06040.pdf) algorithm 9 page 27

jackgerrits (approved):

PR #2903, opened 2021-03-24
Deprecate explicit copy functions and use std::copy as implementation

jackgerrits:
std::copy will use memcpy under the hood

### DIFF
--- a/vowpalwabbit/v_array.h
+++ b/vowpalwabbit/v_array.h
@@ -76,6 +76,21 @@ private:
     memmove(&_begin[idx + width], &_begin[idx], (size() - (idx + width)) * sizeof(T));
   }
 
+  void resize_no_initialize(size_t old_size, size_t length)
+  {
+    // if new length is smaller than current size destroy the excess elements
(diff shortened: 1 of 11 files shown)


vowpalwabbit/v_array.h
```diff
@@ -76,6 +76,21 @@ struct v_array<T, typename std::enable_if<std::is_trivially_copyable<T>::value>:
...
 
+  void resize_no_initialize(size_t old_size, size_t length)
+  {
+    // if new length is smaller than current size destroy the excess elements
+    for (auto idx = length; idx < old_size; ++idx) { _begin[idx].~T(); }
+    reserve(length);
+    _end = _begin + length;
+  }
+
+  void copy_into_this(const v_array<T>& src)
```
olgavrou, line 87: should this just be called `copy`?

olgavrou (approved):

vowpalwabbit/v_array.h
jackgerrits: I went with `copy_into_this` as there is no `dest` argument so it makes it clear where it is ending up

PR #2908, opened 2021-03-25
Fix memory tree demo

### LINKED ISSUE #2907: Contextual Memory Tree fails to load models
#### Describe the bug

Contextual memory tree not working. I think the error is because some parts of VW have been changed API and this part of the code hasn't seen the reflection yet.

#### To Reproduce

Just following instructions on `https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/demo/memory_tree/README.md`

I test two examples. Neither of `python aloi_script_progerror.py` nor `python aloi_script.py` works. I think none of the remaining examples works anymore either.

#### Expected behavior

The model to train and perform eval on test.

#### Observed Behavior

VW just gets stuck and I think it even parses the arguments incorrectly.

#### Environment

`VW 8.9.0 on Ubuntu 18`

jackgerrits:
Fixes #2907


vowpalwabbit/memory_tree.cc
```diff
@@ -1219,7 +1219,7 @@ base_learner* memory_tree_setup(options_i& options, vw& all)
                .default_value(1)
                .help("number of dream operations per example (default = 1)"))
       .add(make_option("top_K", tree->top_K).default_value(1).help("top K prediction error (default 1)"))
-      .add(make_option("learn_at_leaf", tree->learn_at_leaf).help("whether or not learn at leaf (default = True)"))
+      .add(make_option("learn_at_leaf", tree->learn_at_leaf).help("Enable learning at leaf"))
```
lalo, line 1222: .default_value?

jackgerrits: This is a boolean flag. It is either false if it is missing or true if the flag is present

lalo (approved):

Issue #2528, opened 2020-07-28
Python tests fail on Windows


peterychang:
The python tests fail on windows. Most of the failures appear to happen in the DFtoVW converter, specifically when doing type checking.

After some investigation it looks like the problem is that the python `int` type is converted badly on Windows.

Instead of the `int`->`int64` conversion that happens on Linux, Windows does `int`->`int32`

jackgerrits:
Is this fixed now that #2530 is merged?

peterychang:
Fixed in https://github.com/VowpalWabbit/vowpal_wabbit/pull/2530

PR #2909, opened 2021-03-30
refactor: [DFtoVW] Use more generic numpy types, add attribute name to Feature

etiennekintzler:
## 1. Main changes

1. Use more generic numpy type for column type checking
2. Add `name` attribute and systematic checking of name attribute

### 1.1. Use generic numpy type
- Use numpy type that are more generic: `np.floating` and `np.integer`. It should solve the #2528 without the need to add `np.int32` or `np.float32` for Windows system.

### 1.2. Add a `name` attribute to `Feature` class

There is currently no good way to get the feature name. You can use `Feature.value.colname` but it's verbose and just returns the *column name* and not the *feature name*. The column name is different from the feature name if:
- the feature is renamed and/or
- if it has invalid character (in which case it is modified).

Also, using the `name` provides a more consistent API as the `Namespace` class also has a `name` attribute. For instance, it allows the user to use the following listcomp : 

```python
ns_misc = Namespace(name="namespace_misc", features=[Feature(col) for col in ["a", "b"]])
ns_with_fnames = [(ns_misc.name, feature.name) for feature in ns_misc.features ]
```

Also the name is now systematically checked using the static method `_Col.make_valid_name` in classes `Feature` and `Namespace`. It raises a warnings to indicate when the column name is being renamed.

## 2. Minor changes

- Simplify the `AttributeDescriptor.__set__` function
- Simplify the function `DFtoVW.check_columns_existence_in_df` and rename it `check_missing_columns_df`
- Delete faulty line in test `test_multiple_named_namespaces_multiple_features_multiple_lines`
- Raise error when namespace **has value but doesn't have name**. Add a test for that error.

The modifications in PR #2713 that are not related to the `ContextualBanditLabel` class is present in this PR. I will change the former PR as WIP since there is still work to be done.


lalo (approved):

lalo:
Thanks for the PR once again @etiennekintzler

etiennekintzler:
Thanks for your review @lalo  !

PR #2911, opened 2021-03-31
Revert "Migrate example copying to `clone` function and deprecate old methods"

jackgerrits:
Reverts VowpalWabbit/vowpal_wabbit#2906


jackgerrits:
Reverting this change as it did not properly capture some of the nuance of copying examples.

PR #2914, opened 2021-04-02
ci: Enable CI for VW Slim

### LINKED ISSUE #2486: VW Slim unit tests are not running in CI
Ideally slim would be built and tested on each CI platform - MacOS, Windows, Linux. A parallel check to verify this would be good as it should be very fast to execute.

jackgerrits:
Runs on Linux only

Closes #2486


vowpalwabbit/slim/CMakeLists.txt
```diff
@@ -1,8 +1,3 @@
-cmake_minimum_required(VERSION 3.5)
```
lalo, line 1: why are these not needed anymore

lalo (approved):

vowpalwabbit/slim/CMakeLists.txt
jackgerrits: You cannot actually build the slim project "standalone" as it depends on the explore target. Therefore having these here is not useful

PR #2918, opened 2021-04-05
refactor: Refactor learner construction

jackgerrits:
As we continue to add things to the learner struct it is getting harder to use. Furthermore, default values result in a large increase in the number of overloads.

- This PR introduces a fluent interface for constructing a learner and migrates the print reduction to using it.
- Once the builder produces a learner the idea is that it's setup will no longer be able to be changed. This interface trivially supports this.


lalo (approved):

vowpalwabbit/learner.h
```diff
@@ -209,6 +209,13 @@ template <class T, class E>
 struct learner
```
lalo, line 209: add comment explaining both ways? and that its temp?

vowpalwabbit/learner.h
```diff
@@ -533,6 +537,14 @@ struct learner
...
 };
 
+// OLD WAY:
+// The init_learner set of functions is the old way to create a learner. They
+// have been replaced with:
+//   - make_reduction_learner
+//   - make_base_learner
+// They were replaced due to how many different overloads there are and the
+// fact defaults are very hard to express. This problem got worse as more
+// arguments got added.
```
olgavrou, line 547: do we want to deprecate?

jackgerrits: We can't deprecate as using a deprecated function is a build break in the warning as error CIs

vowpalwabbit/learner.h
```diff
@@ -678,5 +690,212 @@ void multiline_learn_or_predict(multi_learner& base, multi_ex& examples, const u
...
+  {
+  }
+
+  FluentBuilderT& set_predict(void (*fn_ptr)(DataT&, BaseLearnerT&, ExampleT&))
+  {
+    this->_learner->learn_fd.predict_f = (learn_data::fn)fn_ptr;
+    return *static_cast<FluentBuilderT*>(this);
+  }
+
+  FluentBuilderT& set_learn(void (*u)(DataT&, BaseLearnerT&, ExampleT&))
```
olgavrou, line 723: nit: `u` -> `fn_ptr`?

vowpalwabbit/learner.h
```diff
@@ -678,5 +690,212 @@ void multiline_learn_or_predict(multi_learner& base, multi_ex& examples, const u
...
+  FluentBuilderT& set_finish_example(void (*f)(vw& all, DataT&, ExampleT&))
+  {
+    _learner->finish_example_fd.data = _learner->learn_fd.data;
+    _learner->finish_example_fd.finish_example_f = (end_fptr_type)(f);
+    return *static_cast<FluentBuilderT*>(this);
+  }
+};
+
+template <class DataT, class ExampleT, class BaseLearnerT>
+struct reduction_learner_builder
```
olgavrou, line 797: I have never done this myself but I think you can use template types of templates to avoid passing in `DataT`, `ExampleT`, and `BaseLearnerT` twice see [here](https://stackoverflow.com/a/213811)

vowpalwabbit/learner.h
```diff
@@ -678,5 +690,212 @@ void multiline_learn_or_predict(multi_learner& base, multi_ex& examples, const u
...
+    this->_learner->learn_fd.data = this->_learner->learner_data.get();
+    this->_learner->sensitivity_fd.sensitivity_f = (sensitivity_data::fn)recur_sensitivity;
+    this->_learner->finisher_fd.data = this->_learner->learner_data.get();
+    this->_learner->finisher_fd.base = make_base(*base);
+    this->_learner->finisher_fd.func = (func_data::fn)noop;
+
+    set_params_per_weight(1);
+
+    this->_learner->pred_type = base->pred_type;
+    // TODO add label type as something learner knows about itself.
```
olgavrou, line 819: is this bit of code commented out on purpose?

vowpalwabbit/learner.h
```diff
@@ -678,5 +690,212 @@ void multiline_learn_or_predict(multi_learner& base, multi_ex& examples, const u
...
+    this->_learner->finisher_fd.func = (func_data::fn)noop;
+    this->_learner->sensitivity_fd.sensitivity_f = (sensitivity_data::fn)noop_sensitivity_base;
+    this->_learner->finish_example_fd.data = this->_learner->learner_data.get();
+    this->_learner->finish_example_fd.finish_example_f = (finish_example_data::fn)return_simple_example;
+
+    this->_learner->learn_fd.data = this->_learner->learner_data.get();
+    this->_learner->pred_type = pred_type;
+    // TODO add label type as something learner knows about itself.
+    std::ignore = label_type;
+    // this->_learner.label_type = label_type;
```
olgavrou, line 858: same here

vowpalwabbit/learner.h
jackgerrits: Yes, learner does not yet support having the label supplied

vowpalwabbit/learner.h
jackgerrits: It will come later, probably once all of the reductions are migrated

vowpalwabbit/learner.h
olgavrou: True just saw that only the print reduction has been migrated

vowpalwabbit/learner.h
olgavrou: yup I was just wondering if here we could do: `:public common_learner_builder<reduction_learner_builder<DataT, ExampleT, BaseLearnerT>>` and then in `common_learner_builder` struct, do e.g. `FluidBuilderT<DataT>` instead of `DataT`

jackgerrits: Ah I see what you're saying. However, base_learner_builder and common_learner_builder have a different number of template arguments so the common one can't just apply the 3:
```C++
FluentBuilderT<DataT, ExampleT, BaseLearnerT>
```
 As that would only work for `reduction_learner_builder` and not `base_learner_builder`

olgavrou: ah right, never mind then

olgavrou (approved):

PR #2920, opened 2021-04-05
refactor: Simple label uses reduction features for initial

jackgerrits:


vowpalwabbit/simple_label.h
```diff
@@ -11,17 +11,16 @@ struct vw;
...
 
   label_data();
-  label_data(float label, float weight, float initial);
+  label_data(float label);
+  void reset_to_default();
+};
+
+struct simple_label_reduction_features
+{
+  float weight = 1.f;
```
peterychang, line 22: can you rename this to something like `initial_weight` with a comment that the field should be considered read-only once its set?

vowpalwabbit/simple_label.h
```diff
@@ -11,17 +11,16 @@ struct vw;
 struct label_data
 {
   float label = 0.f;
-  // only used for serialization and parsing.  example.weight is used for
-  // computation
-  // DeSerialized/Parsed values are copied into example in VW::setup_example()
-  float serialized_weight = 0.f;
```
peterychang, line 17: This will change the ABI. Is that ok? Do we ever pass labels across an API layer?

vowpalwabbit/example.cc
```diff
@@ -185,7 +182,7 @@ flat_example* flatten_example(vw& all, example* ec)
 {
   flat_example& fec = calloc_or_throw<flat_example>();
   fec.l = ec->l;
-  fec.weight = ec->weight;
```
peterychang, line 188: should this be removed? From what I understand, example.weight is still used?

vowpalwabbit/example.h
```diff
@@ -133,7 +119,7 @@ struct vw;
 struct flat_example
 {
   polylabel l;
-  float weight = 1.f;  // a relative importance weight for the example, default = 1
```
peterychang, line 136: Why is weight needed in example, but not in flat_example?

vowpalwabbit/example.cc
jackgerrits: Yes, this was added in the pred/learn PR and so is undone by these changes

vowpalwabbit/example.h
jackgerrits: I am simply migrating back to old usage here. It's potentially a shortcoming in the flat_example type

vowpalwabbit/simple_label.h
jackgerrits: It will change the ABI, not sure if that matters. But this splitting of label into label+reduction_features actually will change the API is a stronger way and should probably be reworked and trigger a major bump as we work out how the external API deals with that change

vowpalwabbit/simple_label.h
jackgerrits: I actually think this is not something I want to tackle in the same change. Weight is used in inconsistent ways and i think require more careful attention. This is pretty much a bug for bug refactor

lalo (approved):

PR #2921, opened 2021-04-05
refactor: Remove copy label

jackgerrits:
Copy label is now implemented by the copy assignment iterators, so this code is no longer needed.


peterychang (approved):

vowpalwabbit/baseline.cc
```diff
@@ -107,7 +107,7 @@ void predict_or_learn(baseline& data, single_learner& base, example& ec)
       init_global(data);
       data.global_initialized = true;
     }
-    VW::copy_example_metadata(/*audit=*/false, data.ec, &ec);
```
lalo, line 110: whats the deal with audit being slowly removed

jackgerrits: Well it wasn't necessary in this function. All audit is copied automatically if it exists. It may be a leftover from how this code used to work

PR #2923, opened 2021-04-06
refactor: Migrate ccb to new make learner structure

jackgerrits:


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -698,12 +698,15 @@ base_learner* ccb_explore_adf_setup(options_i& options, vw& all)
...
-      prediction_type_t::decision_probs, all.get_setupfn_name(ccb_explore_adf_setup), true);
-
-  l.set_finish_example(finish_multiline_example);
-  l.set_save_load(save_load);
-  return make_base(l);
+  auto* l = VW::LEARNER::make_reduction_learner(std::move(data), base, learn_or_predict<true>, learn_or_predict<false>,
+      all.get_setupfn_name(ccb_explore_adf_setup))
+                .set_learn_returns_prediction(true)
+                .set_prediction_type(prediction_type_t::decision_probs)
+                .set_label_type(label_type_t::ccb)
```
lalo, line 705: so label type is a new addition, i'm not seeing on the left side

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -698,12 +698,15 @@ base_learner* ccb_explore_adf_setup(options_i& options, vw& all)
   data->id_namespace_str.append("_id");
   data->id_namespace_hash = VW::hash_space(all, data->id_namespace_str);
 
-  learner<ccb, multi_ex>& l = init_learner(data, base, learn_or_predict<true>, learn_or_predict<false>, 1,
```
lalo, line 701: where is the 1 specified now?

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: Correct, this is something new that learners should specify

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: 1 is the default and so it doesnt need to be added in that case

vowpalwabbit/conditional_contextual_bandit.cc
olgavrou: this information is also specified by the label parser, maybe we should get it from there? (i.e. here do: `all.example_parser->lbl_parser.label_type`)

Also could we use that info from the label parser? where will the learner use it?

jackgerrits: The label parser can only be used for the label type of the topmost reduction and not each one in the stack

The learner is going to use this to enforce type safety between reductions in the long term

lalo: I think we should add ac omment on set_label_type clarifying Olga's point and mnentioning that its not used for now and its for long term design.

olgavrou: maybe in the set function?

olgavrou (approved):

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: There already was a comment in the set function. I added a bit to it stating why it is helpful

PR #2926, opened 2021-04-07
refactor: migrate bfgs to new style learner creation

jackgerrits:


vowpalwabbit/learner.h
```diff
@@ -764,25 +764,25 @@ struct common_learner_builder
 
   FluentBuilderT& set_finish(void (*fn_ptr)(DataT&))
   {
-    _learner->finisher_fd = func_data(_learner->learn_fd.data, _learner->learn_fd.base, (finish_fptr_type)(fn_ptr));
+    _learner->finisher_fd = tuple_dbf(_learner->learn_fd.data, _learner->learn_fd.base, (finish_fptr_type)(fn_ptr));
```
olgavrou, line 767: nit: `tuple_dbf` -> `tuple_data_base_fn`

olgavrou (approved):

vowpalwabbit/learner.h
jackgerrits: This was a bug I introduced in the original PR and was not caught because the template was not instantiated.  The rename would be good but is out of scope for this change

PR #2928, opened 2021-04-07
refactor: move shared_data to its own header 

jackgerrits:


vowpalwabbit/shared_data.cc
```diff
@@ -0,0 +1,317 @@
...
+  weighted_unlabeled_examples = other.weighted_unlabeled_examples;
+  weighted_labels = other.weighted_labels;
+  sum_loss = other.sum_loss;
+  sum_loss_since_last_dump = other.sum_loss_since_last_dump;
+  dump_interval = other.dump_interval;
+  gravity = other.gravity;
+  contraction = other.contraction;
+  min_label = other.min_label;
+  max_label = other.max_label;
+  if (other.ldict) { ldict = VW::make_unique<VW::named_labels>(*other.ldict); }
```
jackgerrits, line 34: Unfortunately we have to define the copy constructor due to the unique ptr and the atomics. Luckily we get move for free!

vowpalwabbit/global_data.cc
```diff
@@ -257,14 +258,7 @@ VW_WARNING_DISABLE_DEPRECATED_USAGE
...
 {
-  sd = &calloc_or_throw<shared_data>();
-  sd->dump_interval = 1.;  // next update progress dump
-  sd->contraction = 1.;
-  sd->first_observed_label = FLT_MAX;
-  sd->is_more_than_two_labels_observed = false;
-  sd->max_label = 0;
-  sd->min_label = 0;
-
+  sd = new shared_data();
```
olgavrou, line 261: could this be made a unique_ptr?

vowpalwabbit/shared_data.cc
olgavrou: can't we do: `ldict = std::move(other.ldict)` here?

olgavrou: nevermind -> copy constructor

vowpalwabbit/global_data.cc
jackgerrits: A seeded copy of the vw struct may point to an existing shared_data object. Because of that a shared_ptr is probably required to express the ownership. I'm open to making it a shared_ptr but think it should be a separate pr

vowpalwabbit/shared_data.cc
jackgerrits: Ah it seems like this is not okay on MacOS :(

olgavrou (approved):

PR #2930, opened 2021-04-08
fix!: cs_active now returns active_multiclass prediction

### LINKED ISSUE #2681: Inconsistent prediction usage in cs_active
cs_active currently uses the prediction in an inconsistent way.
There are 3 elements of the prediction that are used
[scalar](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/cs_active.cc#L169) as an input
[multilabels](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/cs_active.cc#L269) as an output
[multiclass](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/cs_active.cc#L298) as an output

It appears the multilabels usage is incorrect as the union is never initialized as a v_array and would still be a scalar at that point in time

jackgerrits:
This changes the returned prediction from `multiclass` to the new `active_multiclass`. This is technically a breaking change but the old way it was being used was unsafely misusing types. 

There are no differences in command line mode. However, if being used from a bindings the prediction type will now different. 

Python was ported in this PR, but C# and Java still need to be done. Opened issues to address this #2933 #2934

resolves #2681


olgavrou (commented):
The issue linked mentions that `scalar` is used as a prediction input. Should that be moved to the `reduction_features`? (not necessarily part of this PR)

jackgerrits:
It uses the scalar from a predict call internally, so I don't think it fits with reduction_features

Issue #1588, opened 2018-08-31
error "error in demarshal of cost data" provides no tracing information


matanster:
This error message should better include the input line for which it occurs, same as other warnings such as the `malformed example!` warning, at least in daemon mode.

JohnLangford:
This seems like a good idea.

PR #2931, opened 2021-04-08
fix!: Propagate cache reading failure

jackgerrits:
Propagate the reading failure so that the parse loop reports it.

Resolves: #1588


olgavrou (commented):
this looks like it is breaking the previous behavior by throwing? If we get an error here does vw currently continue processing the next example?

jackgerrits:
Hmm good point. This does cause VW to exit.

VW does not currently ignore failed parsing and continue to the next example. I think doing so would be desirable but would be a breaking change right?

jackgerrits:
We should discuss further before if we want to make this change yet since it affects behavior

PR #2936, opened 2021-04-09
feature: enable binary parser for windows build

olgavrou:
`BUILD_EXTERNAL_PARSER_TARGETS` is an env variable


olgavrou:
Un-needed will build with cmake

PR #2950, opened 2021-04-15
fix: Stash benchmark scripts

olgavrou:


jackgerrits (approved):

olgavrou:
benchmark run will fail in this PR since I changed the name of `install-benchmarks.sh` and it tries to find that script in master

PR #2952, opened 2021-04-16
ci: standardize ci names

jackgerrits:


.github/workflows/upload_coverage.yml
```diff
@@ -1,4 +1,4 @@
-name: Upload Code Coverage
```
lalo, line 1: why is this called Upload, shouldnt it just be Code coverage

.github/workflows/format_check.yml
```diff
@@ -1,4 +1,4 @@
-name: Check code formatting
```
lalo, line 1: Code formatting?

.github/workflows/upload_coverage.yml
jackgerrits: The idea is "verb subject", but shorter names do seem preferable

olgavrou (approved):

PR #2953, opened 2021-04-16
fix: cb_adf learn_returns_prediction() incorrect or condition

rajan-chari:
CB_TYPE_DR || CB_TYPE_DM resolves to 0 || 1 which is equal to 1.  This bug results in an extra call to predict() for CB_TYPE_DR.

### DIFF
--- a/vowpalwabbit/cb_adf.cc
+++ b/vowpalwabbit/cb_adf.cc
@@ -107,7 +107,7 @@ public:
   bool learn_returns_prediction()
   {
     return ((_gen_cs.cb_type == CB_TYPE_MTR) && !_no_predict) || _gen_cs.cb_type == CB_TYPE_IPS ||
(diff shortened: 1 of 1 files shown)


jackgerrits:
Since this is a single commit PR (at the moment), the commit message needs to be prefixed with the type also.

Explanation from the [workflow](https://github.com/VowpalWabbit/vowpal_wabbit/blob/d12c1c1a474ac7c120aa94aacb87dfbcd918f3b7/.github/workflows/check_pr_title.yml#L19)
```
          # Reason from action docs:
          # When using "Squash and merge" on a PR with only one commit, GitHub
          # will suggest using that commit message instead of the PR title for the
          # merge commit, and it's easy to commit this by mistake. Enable this option
          # to also validate the commit message for one commit PRs.
```

olgavrou (approved):

PR #2960, opened 2021-04-23
refactor: migrate active_cover learner init

jackgerrits:


vowpalwabbit/active_cover.cc
```diff
@@ -26,18 +26,17 @@ inline float sign(float w)
 struct active_cover
 {
   // active learning algorithm parameters
-  float active_c0;
-  float alpha;
-  float beta_scale;
-  bool oracular;
-  size_t cover_size;
+  float active_c0 = 0.f;
```
lalo, line 29: is this change automated? i've seen it in all your migrations

jackgerrits: As part of this change I am migrating from calloc to new, so I need to set the defaults now

lalo (approved):

PR #2969, opened 2021-04-29
build: Fix clang-format

lokitoth:
PR created by running `./utl/clang-format fix`


jackgerrits:
Since there is only one commit you will need to amend it to include the semantic tag

lokitoth:
I closed this in favour of a new PR, because doing a force push was taking way more effort than it should (in linux container)

PR #2976, opened 2021-05-03
refactor: Calculate number of features created by interactions at base of reductions stack

jackgerrits:


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -580,10 +580,13 @@ void output_example(vw& all, ccb& c, multi_ex& ec_seq)
...
 
-  // Should this be done for shared, action and slot?
   for (auto* ec : ec_seq)
   {
-    num_features += ec->num_features;
+    // Only features in action examples should be counted. This is because we
+    // merged the slot into the shared example and then later down the stack the
+    // shared example is merged into each action.
+    if (ec->l.conditional_contextual_bandit.type == CCB::example_type::action)
+    { num_features += ec->get_num_features(); }
```
olgavrou, line 589: does that mean we never cound the slot/shared example features?

vowpalwabbit/example.h
```diff
@@ -88,11 +88,24 @@ struct example : public example_predict  // core example datatype.
   size_t example_counter = 0;
 
   // helpers
-  size_t num_features = 0;         // precomputed, cause it's fast&easy.
+  // precomputed, cause it's fast&easy. These values are the number of singular/non interacted features.
+  size_t num_features = 0;
+  // precomputed, cause it's fast&easy.
```
olgavrou, line 93: where are these precomputed?

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: The feature counting in CCB is still not quite right, this is a step in the right direction but not enough. It's incorrect to count shared features once, really this needs to be counted for each iteration. This is just a step in the right direction (I think)

vowpalwabbit/example.h
jackgerrits: Because they can be trivially precomputed in the parser, so there's no reason to do it at a later stage. I believe they are always used by the shared_data counters.

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: Would you prefer if this change was omitted until it can be fixed properly?

olgavrou: would
```if (type == ::action)
         num_features += ec->get_num_features() + ec->get_num_features() * num_shared_features
```

be wrong?

I don't have a strong preference if things are being done incrementally that's fine with me, maybe add a TODO?

jackgerrits: I don't think thats right either. I think the tricky thing here is that you need to accommodate for the fact that slot features are merged into shared but then also count taking into account how `cb_adf` merges shared features into each action.

I'll add a todo because I want to come back to it.

vowpalwabbit/count_interactions.cc
```diff
@@ -0,0 +1,105 @@
...
+  {
+    learn_func = transform_single_ex<true, INTERACTIONS::eval_count_of_generated_ft_combinations>;
+    pred_func = transform_single_ex<false, INTERACTIONS::eval_count_of_generated_ft_combinations>;
+    update_func = update<INTERACTIONS::eval_count_of_generated_ft_combinations>;
+    multipredict_func = multipredict<INTERACTIONS::eval_count_of_generated_ft_combinations>;
+  }
+
+  auto data = VW::make_unique<count_interactions>();
+  auto* base = as_singleline(setup_base(options, all));
+  auto* l = VW::LEARNER::make_reduction_learner(
```
lalo, line 99: this reduction might register as learn not returning predict which might not be true

jackgerrits: Yeah youre right. I think I am going to scrap this PR though and take a new approach

test/RunTests
```diff
@@ -33,7 +33,7 @@ my $Valgrind = 'valgrind --quiet --error-exitcode=100 --track-origins=yes --leak
 
 # -- timeout is part of GNU coreutils, some systems may not have it
 my $TimeOut = '';
-my $TimeOutSec = 80;    # max allowed time for single vw command run
+my $TimeOutSec = 120;    # max allowed time for single vw command run
```
olgavrou, line 36: what's timing out?

test/train-sets/ref/help.stdout
```diff
@@ -406,6 +406,9 @@ importance weight classes:
 Confidence:
   --confidence                 Get confidence for binary predictions
   --confidence_after_training  Confidence after training
+Generate interactions:
```
olgavrou, line 409: should this say `Generate permutations` or `Generate permutated interactions`?

test/unit_test/interactions_test.cc
```diff
@@ -0,0 +1,94 @@
...
+
+#include <cstddef>
+#include <cstdint>
+
+#include "test_common.h"
+#include "vw.h"
+#include "gd_predict.h"
+#include "interactions.h"
+
+// the code under DEBUG_EVAL_COUNT_OF_GEN_FT below is an alternative way of implementation of
```
olgavrou, line 16: remove comment and maybe add it as a PR comment for tracking

vowpalwabbit/count_interactions.cc
```diff
@@ -0,0 +1,106 @@
...
+#include <cstddef>
+
+#include "interactions.h"
+#include "example.h"
+#include "memory.h"
+#include "parse_args.h"
+
+using namespace VW::config;
+
+struct count_interactions
```
olgavrou, line 14: why do we need this empty struct?

vowpalwabbit/count_interactions.cc
olgavrou: should this comment be un-resolved since we are keeping this PR?

vowpalwabbit/parser.cc
```diff
@@ -737,6 +737,10 @@ void setup_example(vw& all, example* ae)
     ae->total_sum_feat_sq += fs.sum_feat_sq;
   }
 
+  // Reset any previously counted interacted features.
```
olgavrou, line 740: should this cleanup be done when we are releasing the example back to the pool during example cleanup?

vowpalwabbit/count_interactions.cc
olgavrou: oh nevermind it's the data type of the learner

vowpalwabbit/count_interactions.cc
```diff
@@ -0,0 +1,106 @@
...
+    update_func = update<INTERACTIONS::eval_count_of_generated_ft_combinations>;
+    multipredict_func = multipredict<INTERACTIONS::eval_count_of_generated_ft_combinations>;
+  }
+
+  auto data = VW::make_unique<count_interactions>();
+  auto* base = as_singleline(setup_base(options, all));
+  auto* l = VW::LEARNER::make_reduction_learner(
+      std::move(data), base, learn_func, pred_func, all.get_setupfn_name(count_interactions_setup))
+                .set_learn_returns_prediction(base->learn_returns_prediction)
+                .set_multipredict(multipredict_func)
```
olgavrou, line 102: I guess we don't need to set a label and predict type for this learner because we don't use label/prediction in this reduction, right?

test/RunTests
jackgerrits: A ccb test `-d train-sets/ccb_lots_of_interactions_mini.dat --ccb_explore_adf -q :: -q AB --interactions AAA --interactions ::: --invert_hash ccb_implicit_and_explicit_interactions.inv`

test/train-sets/ref/help.stdout
jackgerrits: Interactions are either permutations or combinations depending on the arguments. This should probably actually say "Count interacted features"

vowpalwabbit/parser.cc
jackgerrits: That's a good idea, I'll do it there too

vowpalwabbit/count_interactions.cc
jackgerrits: By default it will use the base's label and prediction type which is the behavior we want here

olgavrou (approved):

PR #2992, opened 2021-05-11
feat: [metrics] add optional metrics for --dsjson

lalo:


vowpalwabbit/metrics.cc
```diff
@@ -33,7 +33,7 @@ struct metrics_data
   size_t predicted_not_first = 0;
 };
 
-void list_to_json_file(std::string filename, std::vector<std::tuple<std::string, size_t>>& metrics)
+void list_to_json_file(vw& all, std::string filename, std::vector<std::tuple<std::string, size_t>>& metrics)
```
olgavrou, line 36: is it worth passing a pointer to the `metrics` here instead of `all`?

olgavrou (approved):

vowpalwabbit/parse_example_json.h
```diff
@@ -1726,11 +1732,28 @@ bool parse_line_json(vw* all, char* line, size_t num_chars, v_array<example*>& e
     VW::template read_line_decision_service_json<audit>(*all, examples, line, num_chars, false,
         reinterpret_cast<VW::example_factory_t>(&VW::get_unused_example), all, &interaction);
 
+    if (!interaction.eventId.empty())
+    {
+      if (all->example_parser->metrics->FirstEventId.empty())
```
ataymano, line 1737: looks like metrics is initialized conditionally (based on boolean flag), but there is no similar checks here. Or am I missing something?

lalo: yes, right now it is initialized in the same place where the bool of line 1726 of this file gets set. I was expecting a test or some binding to break this contract of not calling into set_json_reader( but doesn't seem like it. Anyways, do you think I should move the initialization over here or just check if its null or not?

vowpalwabbit/parse_example_json.h
```diff
@@ -1726,11 +1732,21 @@ bool parse_line_json(vw* all, char* line, size_t num_chars, v_array<example*>& e
     VW::template read_line_decision_service_json<audit>(*all, examples, line, num_chars, false,
         reinterpret_cast<VW::example_factory_t>(&VW::get_unused_example), all, &interaction);
 
+    if (!interaction.eventId.empty() && all->example_parser->metrics->FirstEventId.empty())
```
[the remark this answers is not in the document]
lalo: yes, in that case the contract would be to keep the first non-empty time and ID and to keep the last non-zero one for the last.

vowpalwabbit/metrics.cc
lalo: done

vowpalwabbit/parse_example_json.h
lalo: addressed this point, see latest diff

PR #2993, opened 2021-05-12
feat: support cached interaction expansion as a reduction

jackgerrits:
99% of the file changes are in test reference files, the line counter is misleading


test/pred-sets/ref/ccb_quad.inv
```diff
@@ -24,128 +24,174 @@ weighted_labeled_examples 8
 weighted_labels 0
 weighted_unlabeled_examples 0
 example_number 8
-total_features 758
+total_features 1151
 total_weight 15.2712
```
olgavrou, line 28: is this calculating more interactions than before?

test/unit_test/ccb_test.cc
```diff
@@ -18,92 +18,6 @@ void inject_slot_features(example* shared, example* slot);
 void remove_slot_features(example* shared, example* slot);
 }  // namespace CCB
 
-BOOST_AUTO_TEST_CASE(ccb_generate_interactions)
```
olgavrou, line 21: are we removing this test?

jackgerrits: Yes, since the code this is testing is no longer needed

test/pred-sets/ref/ccb_quad.inv
jackgerrits: Yes, CCB will interact more namespaces than before. The old code didn't add all the intended interactions. 

The set of interactions for CCB (see https://github.com/VowpalWabbit/vowpal_wabbit/pull/2993/files#diff-bf3dd006bfbdeca21413c272e2a5c433c01659ccd060d2423df5c9c25ca877c8) is:
1. For each existing in interaction insert two new interactions:
    1. The existing one plus ccb_id_namespace
    2. The existing one plus ccb_slot_namespace
2. wildcard +  ccb_id_namespace
3. wildcard +  ccb_slot_namespace

test/train-sets/ref/ccb_test_nonewline.stderr
```diff
@@ -6,18 +6,18 @@ power_t = 0.5
...
-0.000000 0.000000            1            1.0 0:0,1:0,... 0,1,2,...       32
-0.000000 0.000000            2            2.0 3:0,4:0,... 3,4,0,...       32
-0.000000 0.000000            4            4.0 1:0,4:-1,... 1,4,2,...       32
--0.144345 -0.288690            8            8.0 4:-1,1:0,... 4,1,0,...       32
--0.076984 0.000000           16           16.0  ?,?,... 4,2,1,...       32
+0.000000 0.000000            1            1.0 0:0,1:0,... 0,1,2,...       38
+0.000000 0.000000            2            2.0 3:0,4:0,... 3,4,0,...       38
+0.000000 0.000000            4            4.0 1:0,4:-1,... 1,4,2,...       38
+-0.144345 -0.288690            8            8.0 4:-1,1:0,... 4,1,0,...       38
+-0.076984 0.000000           16           16.0  ?,?,... 4,2,1,...       38
```
lalo, line 16: is ? just an edge case in this test file

test/train-sets/ref/ccb_fb.stderr
```diff
@@ -5,18 +5,18 @@ power_t = 0.5
...
-Enabled reductions: gd, scorer, csoaa_ldf, cb_adf, cb_explore_adf_greedy, cb_sample, shared_feature_merger, ccb_explore_adf
+Enabled reductions: gd, generate_interactions, scorer, csoaa_ldf, cb_adf, cb_explore_adf_greedy, cb_sample, shared_feature_merger, ccb_explore_adf
 average  since         example        example  current  current  current
 loss     last          counter         weight    label  predict features
-0.000000 0.000000            1            1.0 0:0,1:0,... 0,1,2,...       32
-0.000000 0.000000            2            2.0 3:0,4:0,... 3,4,0,...       32
-0.000000 0.000000            4            4.0 1:0,4:-1,... 1,4,2,...       32
--0.144345 -0.288690            8            8.0 4:-1,1:0,... 4,1,0,...       32
--0.076984 0.000000           16           16.0  ?,?,... 4,2,1,...       32
+0.000000 0.000000            1            1.0 0:0,1:0,... 0,1,2,...       38
```
lalo, line 11: how does this change affect loading a model generated with the old interaction style?

jackgerrits: As far as I know it is just adding more features. the existing features still exist, so model loading should work fine

test/train-sets/ref/ccb_test_nonewline.stderr
jackgerrits: ? is emitted when the example was unlabelled

test/pred-sets/ref/ccb_quad.inv
lokitoth: Will this change cause difference in behaviour in existing models?

test/train-sets/ref/ccb_quad_save_resume.stderr
```diff
@@ -8,17 +8,17 @@ power_t = 0.5
...
 num sources = 1
-Enabled reductions: gd, scorer, csoaa_ldf, cb_adf, cb_explore_adf_greedy, cb_sample, shared_feature_merger, ccb_explore_adf
+Enabled reductions: gd, generate_interactions, scorer, csoaa_ldf, cb_adf, cb_explore_adf_greedy, cb_sample, shared_feature_merger, ccb_explore_adf
 average  since         example        example  current  current  current
 loss     last          counter         weight    label  predict features
-0.000000 0.000000            1            1.0 0:0,1:0,... 0,1,2,...      135
-0.000000 0.000000            2            2.0 3:0,4:0,... 3,4,0,...       89
-0.000000 0.000000            4            4.0 1:0,4:-1,... 1,4,2,...       89
--0.285714 -0.571429            8            8.0 4:-1,1:0,... 4,1,0,...       89
+0.000000 0.000000            1            1.0 0:0,1:0,... 0,1,2,...      199
```
peterychang, line 14: Any idea why the number of features actually drops here? Mostly just curious about this one..

jackgerrits: The first example vs the second one here? It's just the fact that the first input example has an extra shared feature

vowpalwabbit/interactions.cc
```diff
@@ -4,123 +4,165 @@
...
-  auto& set_interactions = interactions.all_seen_namespaces;
-  for (auto it = set_interactions.begin(); it != set_interactions.end(); ++it)
-  {
-    if (interactions.active_interactions.find({*it, *it}) == interactions.active_interactions.end())
-    {
-      interactions.interactions.push_back({*it, *it});
-      interactions.active_interactions.insert({*it, *it});
-    }
+          // ensure results as big as order_of_inter and empty.
+          for (size_t i = 0; i < results.size(); ++i) results[i] = 0.;
```
peterychang, line 92: nit: `std::fill(results.begin(), results.end(), 0));`

vowpalwabbit/interactions.cc
```diff
@@ -4,123 +4,165 @@
...
+          while (results.size() < order_of_inter) results.push_back(0.);
```
peterychang, line 93: nit:
`if(results.size() < order_of_inter) results.resize_but_with_stl_behavior(order_of_inter);`
The int constructor should initialize the new values to 0 (and if not, just swap the order of lines 92 and 93)`

edit: actually, result should just be the size of order_of_inter. Just ignore the if and always do the resize..

vowpalwabbit/interactions.cc
```diff
@@ -4,123 +4,165 @@
...
+
+          }  // details on http://bit.ly/1Hk9JX1
+
+          num_features_in_inter *= n;
+
+          ns += order_of_inter - 1;  // jump over whole block
+        }
       }
+
+      if (num_features_in_inter == 0) continue;  // signal that values should be ignored (as default value is 1)
```
peterychang, line 160: nit: if line 126 is the only way to reach this state, same as above. Reset both variables and just do the increment?

vowpalwabbit/interactions.cc
```diff
@@ -168,6 +210,27 @@ inline bool must_be_left_sorted(const std::vector<namespace_index>& oi)
   return false;  // 'aaa' or 'abc'
 }
 
+std::vector<std::vector<namespace_index>> expand_quadratics_wildcard_interactions(
```
peterychang, line 213: this function will always remove true duplicates (ie: `{a, b} {a, b}`). I assume thats desirable?

vowpalwabbit/interactions.cc
```diff
@@ -168,6 +210,27 @@ inline bool must_be_left_sorted(const std::vector<namespace_index>& oi)
...
+
+  for (auto it = new_example_indices.begin(); it != new_example_indices.end(); ++it)
+  {
+    auto idx1 = *it;
+    interactions.insert({idx1, idx1});
+
+    for (auto jt = it; jt != new_example_indices.end(); ++jt)
+    {
+      auto idx2 = *jt;
+      interactions.insert({idx1, idx2});
```
peterychang, line 226: is emplace better here? I think it'll save one move call

vowpalwabbit/interactions.cc
```diff
@@ -168,6 +210,27 @@ inline bool must_be_left_sorted(const std::vector<namespace_index>& oi)
...
+      interactions.insert({idx2, idx2});
+      if (leave_duplicate_interactions) { interactions.insert({idx2, idx1}); }
+    }
+  }
+  return std::vector<std::vector<namespace_index>>(interactions.begin(), interactions.end());
```
peterychang, line 231: is there a way to move all these vectors instead of copying them?

vowpalwabbit/interactions.cc
```diff
@@ -227,197 +290,101 @@ void sort_and_filter_duplicate_interactions(
...
 {
-  new_features_cnt = 0;
-  new_features_value = 0.;
-
-  v_array<float> results;
-
-  if (permutations)
+  std::vector<std::vector<namespace_index>> result;
+  // This computation involves factorials and so can only be done with relatively small inputs.
+  if ((namespaces.size() + num_to_pick) <= 21)
```
peterychang, line 326: why 21 in particular?

vowpalwabbit/interactions.cc
```diff
@@ -227,197 +290,101 @@ void sort_and_filter_duplicate_interactions(
...
+      {
+        next_index++;
+        one_based_chosen_indices[next_index - 1] = 1;
       }
 
-      if (num_features_in_inter == 0) continue;  // signal that values should be ignored (as default value is 1)
-
-      new_features_cnt += num_features_in_inter;
-      new_features_value += sum_feat_sq_in_inter;
+      result.push_back(indices_to_values_one_based(one_based_chosen_indices, namespaces));
```
peterychang, line 383: emplace_back? move? does it matter here?

vowpalwabbit/constant.h
```diff
@@ -18,6 +18,7 @@ constexpr float probability_tolerance = 1e-5f;
 constexpr uint32_t FNV_prime = 16777619;
 
 constexpr unsigned char default_namespace = 32;
+constexpr unsigned char wildcard_namespace = 58;  // :
```
peterychang, line 21: nit: It'd be nice to namespace these constants at some point...

vowpalwabbit/interactions.h
```diff
@@ -28,49 +34,113 @@ inline constexpr bool is_printable_namespace(const unsigned char ns)
...
-    generate_interactions<R, S, T, audit, audit_func, dense_parameters>(
-        *ec.interactions, all.permutations, ec, dat, all.weights.dense_weights, num_interacted_features);
+  std::vector<std::vector<namespace_index>> final_interactions;
+
+  for (const auto& inter : interactions)
+  {
+    if (contains_wildcard(inter))
+    {
+      auto compiled = compile_interaction<generate_func, leave_duplicate_interactions>(inter, indices);
+      std::copy(compiled.begin(), compiled.end(), std::back_inserter(final_interactions));
```
peterychang, line 102: any way to move these vectors instead of copying them? Does this fn get called enough for it to matter?

vowpalwabbit/mf.cc
```diff
@@ -59,8 +57,12 @@ void predict(mf& data, single_learner& base, example& ec)
   ec.indices.clear();
   ec.indices.push_back(0);
 
+  auto* saved_interactions = ec.interactions;
```
peterychang, line 60: use a `swap_guard` since this fn is a bit more complex than the other ones where we do this

vowpalwabbit/parse_args.cc
```diff
@@ -1313,14 +1288,17 @@ void parse_reductions(options_i& options, vw& all)
   reductions.push_back(OjaNewton_setup);
   // reductions.push_back(VW_CNTK::setup);
 
+  reductions.push_back(mf_setup);
```
peterychang, line 1291: Is it ok for mf to move here? I assume mf requires the expanded namespaces whereas the others don't?

jackgerrits: From what I can tell it is okay. MF doesnt use the generate_interaction call and does it manually, so it needs the compiled set of interactions

vowpalwabbit/interactions.h
jackgerrits: In these functions the intention is to rely on RVO and NRVO rather than explicitly moving. Often times explicit move calls actually prevents RVO and results in worse performance. The number of times these are called should be infrequent enough to not matter though.

vowpalwabbit/constant.h
jackgerrits: I agree, but a TODO for another day

vowpalwabbit/interactions.cc
jackgerrits: I can add a comment but 21 is the largest factorial that fits in 64 bits, 22 would overflow.

vowpalwabbit/interactions.cc
jackgerrits: Replied in other comment

vowpalwabbit/interactions.cc
jackgerrits: Yes that is desired behavior

vowpalwabbit/interactions.h
```diff
@@ -28,49 +34,113 @@ inline constexpr bool is_printable_namespace(const unsigned char ns)
...
+{
+  return std::find(interaction.begin(), interaction.end(), wildcard_namespace) != interaction.end();
+}
+
+// function estimates how many new features will be generated for example and their sum(value^2).
+void eval_count_of_generated_ft(bool permutations, const std::vector<std::vector<namespace_index>>& interactions,
+    const std::array<features, NUM_NAMESPACES>& feature_spaces, size_t& new_features_cnt, float& new_features_value);
 
-void expand_quadratics_wildcard_interactions(namespace_interactions& interactions);
+std::vector<std::vector<namespace_index>> generate_combinations_with_repetition(
```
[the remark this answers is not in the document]
jackgerrits: Ill update the name

test/unit_test/interactions_test.cc
```diff
@@ -41,28 +46,38 @@ void eval_count_of_generated_ft_naive(vw& all, example_predict& ec, size_t& new_
...
   new_features_value = 0.;
 
+  auto interactions = INTERACTIONS::compile_interactions<generate_func, leave_duplicate_interactions>(
+      all.interactions, std::set<namespace_index>(ec.indices.begin(), ec.indices.end()));
+
   v_array<float> results;
 
   eval_gen_data dat(new_features_cnt, new_features_value);
   size_t ignored = 0;
+  ec.interactions = &interactions;
```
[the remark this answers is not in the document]
jackgerrits: Yes, generate_interactions is pretty big. It's role is to expand out to all feature combinations represented in the compiled interactions. This function could do with work but I think is out of scope of this (already big) change

vowpalwabbit/interactions.cc
olgavrou: `emplace_back` probably helps, `move` wouldn't do anything since returning from a function is already an rvalue

test/pred-sets/ref/ccb_quad.inv
olgavrou: are we sure those were intended to be added in the wildcard interactions and not only as an extra interaction at then end of the existing ones (i.e. just case `1.` described above)?

vowpalwabbit/interactions.cc
```diff
@@ -168,6 +210,27 @@ inline bool must_be_left_sorted(const std::vector<namespace_index>& oi)
...
 }
 
+std::vector<std::vector<namespace_index>> expand_quadratics_wildcard_interactions(
+    bool leave_duplicate_interactions, const std::set<namespace_index>& new_example_indices)
+{
+  std::set<std::vector<namespace_index>> interactions;
+
+  for (auto it = new_example_indices.begin(); it != new_example_indices.end(); ++it)
+  {
+    auto idx1 = *it;
```
olgavrou, line 220: is the copy to `idx1` and `idx2` done for readability here?

olgavrou (commented):
It is hard to make sure that this PR is preserving all of the interaction logic since many bits of the code have shifted and the test files have also changed.

Could we preserve the same amount of interactions in this change (e.g. CCB) to make sure the calculations of interactions stays the same by having the test files stay the same? And if we think we miscalculate the CCB interactions we could add that on a separate PR?

vowpalwabbit/interactions.cc
jackgerrits: Yep just for readability

vowpalwabbit/interactions_predict.h
```diff
@@ -100,8 +97,8 @@ inline void inner_kernel(DataT& dat, features::const_audit_iterator& begin, feat
 template <class DataT, class WeightOrIndexT, void (*FuncT)(DataT&, float, WeightOrIndexT), bool audit,
     void (*audit_func)(DataT&, const audit_strings*),
     class WeightsT>  // nullptr func can't be used as template param in old compilers
-inline void generate_interactions(namespace_interactions& interactions, bool permutations, example_predict& ec,
-    DataT& dat, WeightsT& weights,
+inline void generate_interactions(const std::vector<std::vector<namespace_index>>& interactions, bool permutations,
```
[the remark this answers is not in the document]
jackgerrits: This is just a code move to get slim compiling, I'd like to avoid changing it in this PR

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -43,6 +43,23 @@ void return_v_array(v_array<T>&& array, VW::v_array_pool<T>& pool)
   array = v_init<T>();
 }
 
+void insert_ccb_interactions(std::vector<std::vector<namespace_index>>& interactions_to_add_to)
+{
+  std::vector<std::vector<namespace_index>> new_interactions;
+  for (const auto& inter : interactions_to_add_to)
+  {
+    auto interaction_copy = inter;
```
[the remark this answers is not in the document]
jackgerrits: This was removed from this PR, but when it gets added back in the next one Ill come back to this

vowpalwabbit/interactions.cc
```diff
@@ -4,123 +4,165 @@
...
+      float sum_feat_sq_in_inter = 1.;
+
+      for (namespace_index ns : inter)
+      {
+        num_features_in_inter *= feature_spaces[ns].size();
+        sum_feat_sq_in_inter *= feature_spaces[ns].sum_feat_sq;
+        if (num_features_in_inter == 0) break;
+      }
+
+      if (num_features_in_inter == 0) continue;
```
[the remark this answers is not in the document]
jackgerrits: This is a port of existing code, I do not want to modify it in this PR

vowpalwabbit/interactions.cc
```diff
@@ -4,123 +4,165 @@
...
+    for (const auto& inter : interactions)
     {
-      for (unsigned char j = printable_start; j <= printable_end; ++j)
+      size_t num_features_in_inter = 1;
+      float sum_feat_sq_in_inter = 1.;
+
+      for (auto ns = inter.begin(); ns != inter.end(); ++ns)
       {
-        if (valid_ns(j))
+        if ((ns == inter.end() - 1) || (*ns != *(ns + 1)))  // neighbour namespaces are different
```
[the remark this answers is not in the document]
jackgerrits: This is a port of existing code, I do not want to modify it in this PR

vowpalwabbit/interactions.cc
jackgerrits: This is a port of existing code, I do not want to modify it in this PR.  To all of these I would be happy to come back and address these in another PR

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -455,18 +474,13 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
...
-          std::unique_lock<std::mutex> lock(data.original_interactions->mut);
-          data.generated_interactions.append(*data.original_interactions);
-        }
         calculate_and_insert_interactions(data.shared, data.actions, data.generated_interactions);
-        data.shared->interactions = &data.generated_interactions;
-        for (auto* ex : data.actions) { ex->interactions = &data.generated_interactions; }
       }
+      // Since CCB must interact its own namespaces this operation must always be done to ensure : is expanded.
+      data.shared->interactions = &data.generated_interactions;
+      for (auto* ex : data.actions) { ex->interactions = &data.generated_interactions; }
```
peterychang, line 483: It'd be nice having a guard for this, but its not really necessary since all this is just mucking with member objects

peterychang (approved):
make note of the todos and lgtm

vowpalwabbit/conditional_contextual_bandit.cc
jackgerrits: I'll come back to this as i keep working on this file

vowpalwabbit/parse_args.cc
```diff
@@ -770,88 +771,77 @@ void parse_feature_tweaks(
...
-
-    if (all.interactions.quadratics_wildcard_expansion)
+    if (!all.logger.quiet && !options.was_supplied("leave_duplicate_interactions"))
     {
-      if (options.was_supplied("leave_duplicate_interactions"))
-      { all.interactions.leave_duplicate_interactions = true; }
-      else if (!all.logger.quiet)
+      bool contains_wildcard_quadratic =
+          std::find_if(quadratics.begin(), quadratics.end(), [](const std::string& interaction) {
+            return interaction.find(':') != std::string::npos;
```
olgavrou, line 784: super nit: use wildcard_namespace here instead of `':'`?

vowpalwabbit/parser.cc
```diff
@@ -739,17 +739,6 @@ void setup_example(vw& all, example* ae)
     ae->num_features += fs.size();
   }
 
-  if (all.interactions.quadratics_wildcard_expansion)
-  {
-    // lock while adding interactions since reductions might also be adding their own interactions
```
olgavrou, line 744: yay for lock removal :)

olgavrou (approved):

PR #2994, opened 2021-05-12
fix: bfgs preconditioner

bassmang:


olgavrou (commented):
is this PR related to a specific issue?

PR #2995, opened 2021-05-13
fix: bfgs preconditioner, Resolves #2745

### LINKED ISSUE #2745: LBFGS is broken
Hi!

I believe this change breaks lbfgs:

https://github.com/VowpalWabbit/vowpal_wabbit/commit/ef13913af2f06f7bc4e5c609ae8f3fa96d57cd6b#diff-734f84d3a7f4280dd78e794ac1e9983b6d1b585039f300d4c294adfaadac8702L441

bassmang:


jackgerrits (approved):

test/RunTests
```diff
@@ -2323,4 +2323,13 @@ train-sets/ref/slates_simple_w_interactions.stderr
 {VW} -d train-sets/squarecb_load.dat --cb_explore_adf -q UA --squarecb -i models/sqcb_ld.model -t
     train-sets/ref/squarecb_load.stderr
 
+# Test 316: bfgs preconditioner training
+{VW} -c -k -d train-sets/rcv1_small.dat --bfgs --passes 12 -f models/bfgs_precon.model --save_resume
+    train-sets/ref/bfgs_train.stderr
+    
+# Test 317: bfgs preconditioner testing
```
ataymano, line 2330: What is the purpose of the second test? If the fixed overflowing check is triggered by this dataset, it probably should be already captured by the output from the first test

test/RunTests
```diff
@@ -2323,4 +2323,13 @@ train-sets/ref/slates_simple_w_interactions.stderr
 {VW} -d train-sets/squarecb_load.dat --cb_explore_adf -q UA --squarecb -i models/sqcb_ld.model -t
     train-sets/ref/squarecb_load.stderr
 
+# Test 316: bfgs preconditioner training
```
ataymano, line 2326: [nit] I do not think we can say "preconditioner training". Probably something like "bfgs preconditioner overflowing test" is more accurate description

ataymano (approved):

PR #2997, opened 2021-05-13
fix: cb_explore_adf_first.cc save_load

bassmang:


test/train-sets/ref/squarecb_save.stderr
```diff
@@ -1,11 +1,11 @@
-creating quadratic features for pairs: UA
+creating quadratic features for pairs: UA
```
olgavrou, line 1: do we know why the `squarecb` tests affected by this change?

olgavrou (approved):

jackgerrits (approved):

PR #2998, opened 2021-05-14
refactor: [metrics] cb_explore_adf optional metrics

lalo:


olgavrou (approved):
Do we care about save_resume with metrics?

PR #3001, opened 2021-05-17
build: Update dev environment to use new development image

jackgerrits:


lokitoth (approved):
Good move. Will also incentivize us to start addressing warnings in newer compilers.

PR #3007, opened 2021-05-19
feat: [dsjson] Malformed lines are now skipped instead of being fatal

lalo:
Use --strict_parse to fail instead of skip.


test/train-sets/ccb_data_parse_error.json
```diff
@@ -0,0 +1,59 @@
...
+{"Timestamp":"2021-04-26T17:23:28.1726275,"Version":"1","EventId":"0000006","c":{"FromUrl":[{"weather":"sunny"}],"_multi":[{"_tag":"Mocha","i":{"constant":1,"id":"Mocha"},"j":[{"roast":"light"}]},{"_tag":"Latte","i":{"constant":1,"id":"Latte"},"j":[{"roast":"dark"}]},{"_tag":"Cappuccino","i":{"constant":1,"id":"Cappuccino"},"j":[{"roast":"dark"}]},{"_tag":"Americano","i":{"constant":1,"id":"Americano"},"j":[{"roast":"light"}]}],"_slots":[{"_id":"MainArticle","sj":[{"size":"Large","position":"Center"}]},{"_id":"SideBar","sj":[{"size":"Medium","position":"Right"}]}]},"_outcomes":[{"_label_cost":0.0,"_id":"MainArticle","_a":[1,0,2,3],"_p":[0.25,0.25,0.25,0.25],"_o":[],"_original_label_cost":0.0},{"_label_cost":0.0,"_id":"SideBar","_a":[3,2,0],"_p":[0.33333334,0.33333334,0.33333334],"_o":[],"_original_label_cost":0.0}],"_ba":[1,2],"VWState":{"m":"N/A"}}
```
lalo, line 7: " missing on purpose

test/RunTests
```diff
@@ -2359,4 +2359,10 @@ train-sets/ref/slates_simple_w_interactions.stderr
 {VW} -d train-sets/cb_load.dat --cb_explore_adf -q UA --regcb -i models/regcb_ld.model -t
     train-sets/ref/regcb_load.stderr
 
+# Test 324: test extra_metrics with djson ccb
+# --quiet bc logging is not thread safe, so diff fails randomnly
+{VW} -d train-sets/ccb_data_parse_error.json --dsjson --ccb_explore_adf --extra_metrics metrics_ccb_parse_error.json --quiet
```
olgavrou, line 2364: (super nit, not too fussy): just so that we don't go looking for it if needed, maybe add a comment here that line 7 in data file, Timestamp field is malformed ?

olgavrou (approved):

vowpalwabbit/parse_example_json.h
```diff
@@ -1708,15 +1714,15 @@ void read_line_decision_service_json(vw& all, v_array<example*>& examples, char*
...
   apply_pdrop(all, data->probabilityOfDrop, examples);
 
-  if (!result.IsError()) return;
-
-  BaseState<audit>* current_state = handler.current_state();
+  if (result.IsError())
+  {
+    BaseState<audit>* current_state = handler.current_state();
+    logger::errlog_warn("JSON parser error at {0}: {1}. Handler: {2} State: {3}", result.Offset(),
+        GetParseError_En(result.Code()), handler.error().str(), (current_state ? current_state->name : "null"));
```
cirvine-MSFT, line 1721: Per discussion in other PR, perhaps we should only log this a set number of times, before then just not logging for the rest of this input. e.g. log 10 times and then say "there are a lot of parse errors in this file". The metric will still keep count of how many errors (which is super nice :) )

lalo: I think VW should not be that opinionated about silencing selectively, i.e. library mode or other bindings want to know that information all the time. Let's think about it for a bit.

jackgerrits:
Can you add logic to respect `strict_parse` which is used in parse_example.cc

jackgerrits (approved):

jackgerrits:
Nit: this PR seems like a feature not a refactor. DSJSON parsing now respects the strict_parse flag

ataymano (approved):

vowpalwabbit/parse_example_json.h
lalo: https://github.com/VowpalWabbit/vowpal_wabbit/pull/3021/files

lalo: #3021

PR #3017, opened 2021-05-21
fix: set default learn_returns_prediction to false, refactor cover

bassmang:


vowpalwabbit/learner.h
```diff
@@ -974,12 +974,14 @@ VW_WARNING_STATE_POP
 template <class DataT, class ExampleT, class BaseLearnerT>
 reduction_learner_builder<DataT, ExampleT, BaseLearnerT> make_reduction_learner(std::unique_ptr<DataT>&& data,
     BaseLearnerT* base, void (*learn_fn)(DataT&, BaseLearnerT&, ExampleT&),
-    void (*predict_fn)(DataT&, BaseLearnerT&, ExampleT&), const std::string& name)
+    void (*predict_fn)(DataT&, BaseLearnerT&, ExampleT&), const std::string& name,
+    bool learn_returns_prediction = false)
```
jackgerrits, line 978: This should not be a part of the `make_reduction_learner` call. The point of this interface is to reduce the number of arguments required.

```suggestion
    void (*predict_fn)(DataT&, BaseLearnerT&, ExampleT&), const std::string& name)
```

Line 984 should be: 
```c++
builder.set_learn_returns_prediction(false);
```

vowpalwabbit/learner.h
```diff
@@ -974,12 +974,14 @@ VW_WARNING_STATE_POP
...
 {
   auto builder = reduction_learner_builder<DataT, ExampleT, BaseLearnerT>(std::move(data), base, name);
   builder.set_learn(learn_fn);
   builder.set_update(learn_fn);
   builder.set_predict(predict_fn);
+  builder.set_learn_returns_prediction(learn_returns_prediction);
```
jackgerrits, line 984:
```suggestion
  builder.set_learn_returns_prediction(false);
```

vowpalwabbit/cb_explore_adf_cover.cc
```diff
@@ -305,17 +306,19 @@ VW::LEARNER::base_learner* setup(config::options_i& options, vw& all)
...
-  l.set_finish_example(explore_type::finish_multiline_example);
-  l.set_print_example(explore_type::print_multiline_example);
-  l.set_save_load(explore_type::save_load);
-  l.set_persist_metrics(explore_type::persist_metrics);
-  return make_base(l);
+  auto data = VW::make_unique<explore_type>(with_metrics, cover_size, psi, nounif, epsilon, epsilon_decay, first_only,
+      as_multiline(all.cost_sensitive), all.scorer, cb_type_enum, all.model_file_ver);
+  auto* l = make_reduction_learner(
+      std::move(data), base, explore_type::learn, explore_type::predict, all.get_setupfn_name(setup) + "-cover", true)
+                .set_params_per_weight(problem_multiplier)
```
jackgerrits, line 313: `learn_returns_prediction` should not be in the `make_reduction_learner` call. Set it with the builder instead.

```suggestion
  auto* l = make_reduction_learner(
      std::move(data), base, explore_type::learn, explore_type::predict, all.get_setupfn_name(setup) + "-cover")
                .set_learn_returns_prediction(true)
                .set_params_per_weight(problem_multiplier)
```

jackgerrits (changes requested):
New parameters should not be added to the `make_reduction_learner` call as one of the reasons it exists is to reduce the number of arguments required in the default case and remove the need for overloads.

jackgerrits (approved):

jackgerrits (commented):
resetting my vote until tests are fixed

jackgerrits:
It looks as though GEN_CS::cb_to_cs_adf does not initialize its fields and we are now making use of its default constructor through this change.

bassmang (commented):
Added initialization for cb_type and scorer

jackgerrits (approved):

PR #3027, opened 2021-05-28
feat: [py] Providing Basic support for Tensorboard and Tensorwatch

speedhunter001:
## Usage
- Created `VWtoTensorboard`, `VWtoTensorwatchStreamer`, `VWtoTensorwatchClient`, `DFtoVWtoTBorTW` classes in `python/vowpalwabbit/DFtoVWtoTBorTW.py` 
-  Also created an example (.ipynb) file in `python/examples/iris-df-to-vw-to-tensorboard-tensorwatch`
-  Edited `pyvw.py` to add a method `.get_label` and modified `vw.learn` to support tensorboard and tensorwatch logs
- Added **Iris.csv** to `python/examples/iris-df-to-vw-to-tensorboard-tensorboard`
- Purpose of `DFtoVWtoTBorTW` is to use this class to create tensorboard or tensorwatch logs or both for vw metrics out of a pandas df,  DFtoVW and vw objects
- Purpose of `VWtoTensorboard` is to support Tensorboard logs for vw metrics: **average_loss, since_last**, **graph of vw reductions**,  **arguments as text** for now
- Purpose of `VWtoTensorwatchStreamer` and  `VWtoTensorwatchClient` is to support Tensorwatch logs for vw metrics: **average_loss, since_last, label, prediction** for now
```python
df_simple = pd.DataFrame({
        "f1": [1, 2, 3],
        "f2": [1.2, 2.3, 3.4],
        "f3": [0.2, 0.8, .01],
        "l": [1, 2, 3],
})

label = MulticlassLabel(label="l")
features = [Feature(col) for col in ["f1", "f2", "f3"]]

df_to_vw = DFtoVW(df=df_simple, label=label, features=features)
vw = pyvw.vw('--oaa 3 -P 1')

# For tensorboard
logdir = './logs'
vw_to_tb = VWtoTensorboard(logdir)

# For tensorwatch
logfile = './iris.log'
vw_to_tw = VWtoTensorwatchStreamer(logfile)

df_to_tb_tw = DFtoVWtoTBorTW(df_to_vw.convert_df(), vw)
df_to_tb_tw.fit(vw_to_tensorboard=vw_to_tb, vw_to_tensorwatch=vw_to_tw)  

vw_to_tb.draw_reductions_graph(vw)
vw_to_tb.show_args_as_text(vw)

vw_to_tw_client = VWtoTensorwatchClient(logfile)
vw_to_tw_client.plot_metrics()

# Both vw_to_tb and vw_to_tw objects can be passed to a single DFtoVWtoTBorTW instance
```

Tensorboard logs will be created as `VWtoTensorboard` instance is passed to `.fit` method
Tensorwatch logs will be created as `VWtoTensorwatchStreamer` instance is passed to `.fit` method

To run tensorboard on logs that were created, type command `tensorboard --logdir ./logs` in the directory where logs for tensorboard are present 

For tensorwatch visualization use `.plot_metrics` of Tensorwatch Client as above
## Implementation Details

1. Create a `DFtoVW` object using `pandas Dataframe` object
2. Create a `vw` object, specify `-P 1` too whle initializing object
3. Create a `VWtoTensorboard` instance, or `VWtoTensorwatchStreamer` instance or both
3. Create `DFtoVWtoTBorTW` object by passing examples from `DFtoVW` and `vw` objects as parameters
4. Call `.fit()` method of `DFtoVWtoTBorTW`  (`tensorboard` and `tensorwatch` argument is optional) for the `vw` object to start learning, and to output **Tensorboard** logs (if `VWtoTensorboard` instance is passed, otherwise default value is **None**), and to also output **Tensorwatch** logs (if `VWtoTensorwatchStreamer` instance is passed, otherwise default value is **None**)

`VWtoTensorboard` is used to log metrics for tensrboard
 `.get_label()` method is also added to `pyvw.py` before `pyvw.get_prediction` implementation and `vw.learn` has been modified to support **Tensorboard logs for average_loss and since_last, graph of vw reductions, arguments as text** and **Tensorwatch logs for average_loss, since_last, label, prediction**

## Limitations

- Curretly its only for `vw objects` where `-P 1` is specified
- Currently no `.predict()` method for predicting on data set in `DFtoVWtoTensorboard`
- Methods for these seem to be not in vw currenlty, so couldn't accomodate them in `pyvw.get_label()`						
                                               - 5: lMAX
						- 6: lCONDITIONAL_CONTEXTUAL_BANDIT
						- 7: lSLATES
						- 8: lCONTINUOUS
- Tensorboard logs for metrics: **label, prediction** are not created as line plot for them is not suitable inside Tensorboard but barplot support is available for Tensorwatch so they are supported in Tensorwatch				
- `self.finish_example` is not called in `vw.learn` if an **example instance** is passed as parameter, calling `vw.learn(ec, vw_to_tb)` with **ec as example instance, vw_to_tb as VWtoTensorboard instance** would give an error as metrics are not calculated correctly
- logs of **label, prediction** for Tensorwatch is only done when **string ec** object is passed


python/vowpalwabbit/DFtoVWtoTensorboard.py
```diff
@@ -0,0 +1,208 @@
...
+						Default value is True, this parameter is used to control the logging of metrics for Tensorboard visualization
+						If value is True : metrics are logged for Tensorboard visualization
+						If value is False : metrics are not logged for Tensorboard visualization
+
+		Returns
+		-------
+
+		None
+		"""
+		if tensorboard:
```
lokitoth, line 171: Consider wrapping this bit into a VWtoTensorboard class, and optionally pass it to your .learn/.fit methods, rather than a flag to turn Tensorboard on. This will allow you to specify the log folder rather than hardcoding it, and later configure which metrics are projected into Tensorboard.

python/vowpalwabbit/DFtoVWtoTensorboard.py
```diff
@@ -0,0 +1,208 @@
...
+			pyvw.vw.lContextualBandit: example.get_cbandits_class,
+	#         pyvw.vw.lConditionalContextualBandit: None,
+	#         pyvw.vw.lSlates: None,
+	#         pyvw.vw.lContinuous: None
+		}
+		return switch_label_type[label_type]()	
+
+
+	#------------------------------------------------------------------------------------------------		
+	def fit(self, df, tensorboard=True):
```
lokitoth, line 153: A lot of this method shares code with `vw.learn` - consider unifying them (e.g. pull in the Tensorboard code into `.learn`, and either add a path to take in a DF in the method, or create a `learn_all` or `fit` method on the vw class.

cc @etiennekintzler, @jackgerrits - do you have an opinion on whether we should reuse the `.learn` method vs. a `learn_all`/`fit` method to learn over the entire DFtoVW dataset.

@

model_file_cb_equiv.txt
```diff
@@ -0,0 +1,16 @@
+Version 8.11.0
```
lokitoth, line 1: [Minor] Did you intend to include these .txt files? If not, consider removing them from the PR.

lokitoth (changes requested):

python/vowpalwabbit/DFtoVWtoTensorboard.py
lokitoth: I realized that the above is a bit ambiguous: If switching to a VWtoTensorboard class, you should enable/disable Tensorboard export based on whether an instance of the class was passed in to your `.learn`/`.fit` method. On this class you can set up things like what the appropriate log paths are, run identifiers, etc.

python/vowpalwabbit/DFtoVWtoTensorboard.py
```diff
@@ -0,0 +1,208 @@
...
+		-------
+
+		since_last      : float
+							since_last calculated using sum_loss_since_last and weighted_examples_since_last or if ZeroDivisionError is encountered then 0. is returned 
+		""" 
+		try:
+			return sum_loss_since_last / weighted_examples_since_last
+
+		except ZeroDivisionError:
+			return 0.
```
etiennekintzler, line 79: I think these 2 methods could be removed since :
- they both do the same thing (guard against division by 0)
- they are only used once in the `fit ` method

python/vowpalwabbit/DFtoVWtoTensorboard.py
```diff
@@ -0,0 +1,208 @@
...
+						If value is False : metrics are not logged for Tensorboard visualization
+
+		Returns
+		-------
+
+		None
+		"""
+		if tensorboard:
+			logdir = "logs/scalars/" + datetime.now().strftime("%Y%m%d-%H%M%S")  # logs directory
+			file_writer = tx.SummaryWriter(logdir + "/iris")   # creating file writer
```
etiennekintzler, line 173: `"iris"` shouldn't be the default directory name

python/vowpalwabbit/DFtoVWtoTensorboard.py
```diff
@@ -0,0 +1,208 @@
...
+ 					This would be used for learning using the algorithm specified in the object, and also to compute a few metrics
+
+		Returns
+		-------
+
+		self : DFtoVWtoTensorboard
+		"""
+		self.df_to_vw = df_to_vw
+		self.vw = vw
+		self.vw_formatted_data = df_to_vw.convert_df()
```
etiennekintzler, line 29: The dependency on `DFtoVW` is not necessary in my opinion.

Indeed, the list of examples need not be the result of the  `DFtoVW.convert_df` method. The user could just provide the list of examples instead.

python/vowpalwabbit/DFtoVWtoTensorboard.py
```diff
@@ -0,0 +1,208 @@
...
+		"""
+		if tensorboard:
+			logdir = "logs/scalars/" + datetime.now().strftime("%Y%m%d-%H%M%S")  # logs directory
+			file_writer = tx.SummaryWriter(logdir + "/iris")   # creating file writer
+		
+		sum_loss = 0.
+		weighted_examples = 0.
+
+		for ind, iteration in zip(df.index, range(len(df))):
+			vw_format = self.vw_formatted_data[ind]   # get the string format of specific example
```
etiennekintzler, line 179: Something more pythonic could be used 

```python
for iteration, vw_format in enumerate(self.vw_formatted_data):
```

python/vowpalwabbit/DFtoVWtoTensorboard.py
```diff
@@ -0,0 +1,208 @@
...
+					This would be used to construct examples 
+		vw		 : A vowpalwabbit object
+ 					This would be used for learning using the algorithm specified in the object, and also to compute a few metrics
+
+		Returns
+		-------
+
+		self : DFtoVWtoTensorboard
+		"""
+		self.df_to_vw = df_to_vw
```
etiennekintzler, line 27: No need to set `df_to_vw` has instance attribute since it is not used elsewhere

python/vowpalwabbit/DFtoVWtoTensorboard.py
etiennekintzler, line 179: `vw_formatted_data` is a list but the index of a dataframe can be strings. In the case of strings,  you'll get a `TypeError` (e.g you can't access element of the list `lst["a"]`).

python/vowpalwabbit/DFtoVWtoTensorboard.py
etiennekintzler: @lokitoth In my opinion the whole class is a bit overkill and embedding the fitting process inside a  `fit` method is too restrictive.

I am not super familiar with tensorboard but the writer is usually used inside the fitting loop, for example see the following pytorch reference: https://pytorch.org/tutorials/intermediate/tensorboard_tutorial.html . Also it is up to the user to include what he wants in the tensorboard, which is not allowed here.

Instead of this class, I think what could be great is a notebook that uses Tensorboard in the same fashion as the pytorch tensorboard tutorial. What do you think of this idea @speedhunter001 @lokitoth ?

etiennekintzler: Another thing that could be done is to build a helper class that parse the standard vowpal wabbit output (see below) and write it to files that can be rendered with tensorboard.

```
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = test.csv
num sources = 1
Enabled reductions: gd, scorer
1.000000 1.000000            1            1.0   1.0000   0.0000       15
0.500000 0.000000            2            2.0   1.0000   1.0000       15
0.504930 0.509860            4            4.0   1.0000   0.5378       15
0.323170 0.141410            8            8.0   1.0000   1.0000       15
....
```

It could be useful to anyone who want to have graphical summary of the performances/parameters and would not require a tight coupling with the user code.

speedhunter001: @etiennekintzler To get familiar with Tensorboard, I modified the following tutuorial on **Content Personalization Simulation**  https://vowpalwabbit.org/tutorials/cb_simulation.html  to see the ctr (click through rate) plots on Tensorboard.  By seeing such an example, the user may get an idea to log any scalar metrics for Tensorboard. So what do you suggest on this that should I include this as an example in PR too?   

The current class DFtoVWtoTensorboard **logs the following standard vowpal wabbit output metrics i.e average_loss, since_last, current_features** which one can see in Tensorboard, currently Tensorboard does not have support for bar plots so we thought that logging **current label** and **current_predict** would not be suitable. What is your suggestion on this?

model_file_cb_equiv.txt
speedhunter001: Yes I had not intention for .txt files to be include. Gonna remove them soon

python/vowpalwabbit/DFtoVWtoTensorboard.py
speedhunter001: So should I instead write the code for these inside `.fit` method? At first I though that doing that would make the code look messy or less readable. What would you suggest on this?

python/vowpalwabbit/DFtoVWtoTensorboard.py
speedhunter001: Ok gonna change that soon

python/vowpalwabbit/DFtoVWtoTensorboard.py
speedhunter001: So should the user providing the list of examples be **optional behaviour**? or `DFtoVW ` object passed as `df_to_vw` should not be used at all to get the list of example and the user should provide the list of examples always?

python/vowpalwabbit/DFtoVWtoTensorboard.py
speedhunter001: Ok gonna remove that

python/vowpalwabbit/DFtoVWtoTensorboard.py
speedhunter001: If I use `for iteration, vw_format in enumerate(self.vw_formatted_data):` I think that would solve this bug

python/vowpalwabbit/DFtoVWtoTensorboard.py
```diff
@@ -0,0 +1,208 @@
...
+		"""
+		print( 'average_loss:{:.6f}'.format(average_loss) , end='\t')
+		print('since_last:{:.6f}'.format(since_last), end='\t')
+		print('label:', label, end='\t')    
+		print('prediction:', prediction, end='\t')    
+		print('num_features:', num_features) 
+
+
+	#------------------------------------------------------------------------------------------------		
+	def get_label(self, example, label_type):
```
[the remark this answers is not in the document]
speedhunter001: Okay I would look into it

python/vowpalwabbit/DFtoVWtoTensorboard.py
speedhunter001: Okay I would try that now

python/vowpalwabbit/DFtoVWtoTensorboard.py
speedhunter001: @lokitoth I would try to understand the code of `vw.learn` to consider making the changes

etiennekintzler: > The current class DFtoVWtoTensorboard logs the following standard vowpal wabbit output metrics i.e average loss, since last, current features which one can see in Tensorboard, this currently provides the user to see few of the standard vowpal wabbit output metrics, currently Tensorboard does not have support for bar plots so we thought that logging current label and current predict would not be suitable. What is your suggestion on this?

**Your class computes metrics that are already available in the default vowpal wabbit output** (the output in my last message). 

My suggestion was to **just parse this VW output** and write the parsed result to tensorboard. Below is a notebook that essentially do this: https://github.com/etiennekintzler/vowpal_wabbit/blob/tensorboard_example/python/examples/tensorboard_example/tensorboard_vw.ipynb

Also, this way to do is a bit more generic as you just need the output file and you don't need to redo the whole fitting process inside the class.

Below is the plot of tensorboard I obtained : 

image

lokitoth: @etiennekintzler Rather than parsing the VW output, it seems like it would be good to enable a callback into python directly exposing these metrics (and short-circuiting the printout entirely), or having them be available in a dictionary (like the extra metrics that @lalo added in #3022). That would be a longer-term goal of @speedhunter001's efforts here this summer.

etiennekintzler: If the metrics are directly available from python's VW model it's indeed more straightforward to use it directly to write into tensorboard files

Good luck and thanks to @speedhunter001 for taking care of this :)

python/vowpalwabbit/DFtoVWtoTensorboard.py
```diff
@@ -0,0 +1,182 @@
...
+		
+			self._print_metrics(average_loss, since_last, label, prediction, num_features)
+
+			if isinstance(vw_to_tensorboard, VWtoTensorboard):
+				file_writer.add_scalar('average_loss', average_loss, iteration)  # logging average_loss on each iteration
+				file_writer.add_scalar('since_last', since_last, iteration)   # logging since_last on each iteration
+			#     file_writer.add_scalar('label' , label, iteration)
+			#     file_writer.add_scalar('prediction', prediction, iteration)
+			#     file_writer.add_histogram('label-prediction', [label, prediction], iteration)
+			#     file_writer.add_histogram('label', label, iteration)
```
lokitoth, line 180: [Minor] When you are ready to get this merged, remember to remove commented out code.

python/vowpalwabbit/DFtoVWtoTensorboard.py
```diff
@@ -0,0 +1,182 @@
...
+			example = self.vw.parse(vw_format)       # parse the string format, it returns an example object
+			self.vw.learn(example)                  # learn on example
+
+			label = pyvw.get_label(example, self.vw.get_label_type())
+			prediction = pyvw.get_prediction(example, self.vw.get_prediction_type())
+			num_features = example.get_feature_number()    
+
+			self.vw.finish_example(example)  # Any use of vw object should be done after this and use of example before this 
+
+			sum_loss_since_last = self.vw.get_sum_loss() - sum_loss  # vw.get_sum_loss() return current sum loss, sum_loss variable right now holds sum loss of previous iteration
```
lokitoth, line 163: [Minor][Perf] You should consider making metrics calculation/printing here be optional (similar to how you made the tensorboard output optional.

python/vowpalwabbit/DFtoVWtoTensorboard.py
```diff
@@ -0,0 +1,182 @@
...
+		"""
+		print( 'average_loss:{:.6f}'.format(average_loss) , end='\t')
+		print('since_last:{:.6f}'.format(since_last), end='\t')
+		print('label:', label, end='\t')    
+		print('prediction:', prediction, end='\t')    
+		print('num_features:', num_features) 
+
+
+	#------------------------------------------------------------------------------------------------		
+	def fit(self, vw_to_tensorboard=None):
```
lokitoth, line 131: Consider merging this class with the pyvw class, in particular, pulling code in the learning loop here into `pyvw.learn`: This will allow everyone who uses pyVW to use your VWtoTensorboard integration, not only those using DFtoVW.

python/vowpalwabbit/DFtoVWtoTensorboard.py
lokitoth: I agree with @etiennekintzler: The extra code in `.fit` is much smaller than the code + header of the helper functions, and what is going on is fairly straightforward. The only thing I would suggest is to do the zero check directly, rather than relying on the exception:

```
   return (sum_loss_since_last / weighted_examples_since_last) if weighted_examples_since_last  != 0 else 0
```

python/vowpalwabbit/DFtoVWtoTensorboard.py
lokitoth: The way you can do this is similar to how the `pyvw.learn` method deals with different types of input: Check the type explicitly.

lokitoth (approved):
Looks good.

python/vowpalwabbit/DFtoVWtoTensorboard.py
```diff
@@ -0,0 +1,90 @@
...
+		vw		 : A vowpalwabbit object
+ 				This would be used for learning using the algorithm specified in the object, and also to compute a few metrics
+
+		Returns
+		-------
+
+		self : DFtoVWtoTensorboard
+		"""
+		self.vw = vw  # As this needs to be remembered
+		self.vw_formatted_data = df_to_vw.convert_df()
```
etiennekintzler, line 68: This part introduce a tight coupling with `DFtoVW` class that is not necessary IMO. 

The examples could be the result of `DFtoVW.convert_df()` **but could well have been directly built by the user**. It seems better to just pass a list of string `vw_examples` as argument instead of a `DFtoVW` object.

python/examples/iris-df-to-vw-to-tensorboard/iris-pandas-df-to-vw-and-tensorboard-logs.ipynb
```diff
@@ -0,0 +1,396 @@
...
+    "# converting \"Species\" categorical column to integer\n",
+    "def categorical_to_int(specie):\n",
+    "    if specie == 'Iris-setosa':\n",
+    "        return 1\n",
+    "    elif specie == 'Iris-versicolor':\n",
+    "        return 2\n",
+    "    elif specie == 'Iris-virginica':\n",
+    "        return 3\n",
+    "    \n",
+    "df['Species'] = df['Species'].apply(categorical_to_int)\n",
```
etiennekintzler, line 165: If you want you can also used the [pandas.Series.map method](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.map.html) which is more straightforward for what you want to do:

```python
df['Species'] = df['Species'].map({'Iris-setosa': 1, 'Iris-versicolor': 2, 'Iris-virginica': 3})
```

python/examples/iris-df-to-vw-to-tensorboard/iris-pandas-df-to-vw-and-tensorboard-logs.ipynb
```diff
@@ -0,0 +1,396 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "target_col = \"Species\"\n",
+    "\n",
+    "# features = [Feature(col) for col in df.columns if col != target_col]  # target column not part of Features\n",
+    "features = [Feature(col) for col in df.columns if col != target_col and col != 'Id']  # 'Id' column also not part of Features along with target col\n",
```
etiennekintzler, line 178: More straightforward ;-)
```python
features = [Feature(col) for col in df.columns if col not in [target_col, 'Id']]
```

python/examples/iris-df-to-vw-to-tensorboard/iris-pandas-df-to-vw-and-tensorboard-logs.ipynb
speedhunter001: It does look more straightforward, gonna change that

python/vowpalwabbit/DFtoVWtoTensorboard.py
speedhunter001: I will look into that

jackgerrits:
It seems like we could reasonably implement this externally to the core library? Adding a Tensorboard dependency seems pretty heavy if it can be avoided

lokitoth:
Hey @jackgerrits - could you expand a bit on how you think it would be layered? Are you thinking of creating a proxy object that drives the underlying vw object? Or were you thinking more along the lines of expecting a specific callback interface into the normal learning code, like we do now, just without specifically typing it, then implementing the TW/TB projections externally?

jackgerrits:
I am primarily referring to a sort of proxy object. It seems like the information required for this module to operate is available "externally" to the individual predict and learn calls. Therefore, providing a separate, and importantly optional, module which wraps the necessary predict and learn calls seems to be a more flexible design.

The biggest factor here is that adding `tensorboardX` and `tensorwatch` dependencies to the core `vowpalwabbit` library is not a viable path forward.

jackgerrits:
Going to go ahead and close this. If you would like to revisit in future please feel free to reopen.

PR #3028, opened 2021-05-28
chore: Add deprecation template

kumpera:
This is an initial proposal of how to track all items under our deprecation and removal schedule.

A few things we should consider is how we'd track the lifecycle of those issue.

We could use tags to signal their status or a project board. What do you think?


jackgerrits (approved):

.github/ISSUE_TEMPLATE/deprecation.md
```diff
@@ -0,0 +1,31 @@
...
+assignees: ''
+
+---
+
+#### Describe the feature to be deprecated
+When describing the feature, focus on the user visible aspects of it.
+For example:
+
+- Are we removing a C++ API method?
+- Are we removing a command line switch?
```
jackgerrits, line 15:
```suggestion
- Are we removing a command line option?
```

PR #3030, opened 2021-06-01
refactor: Avoid v_array copies

bassmang:


vowpalwabbit/cb_explore_adf_common.h
```diff
@@ -123,16 +123,16 @@ inline void cb_explore_adf_base<ExploreType>::predict(
...
+    label_example->l.cb = std::move(data._empty_label);
   }
 
   data.explore.predict(base, examples);
 
   if (label_example != nullptr)
   {
     // predict path, restore label
-    label_example->l.cb = data._action_label;
+    label_example->l.cb = std::move(data._action_label);
```
jackgerrits, line 135: Shouldn't `data._empty_label` be moved back into down here?

bassmang: Just changed this in the last commit

olgavrou (approved):

vowpalwabbit/cb_explore_adf_common.h
olgavrou: @jackgerrits should we also reset the label here? i.e. clear the array and reset the cost?

jackgerrits: Yeah we probably should

vowpalwabbit/cb_explore_adf_common.h
```diff
@@ -123,16 +123,18 @@ inline void cb_explore_adf_base<ExploreType>::predict(
...
+    data._empty_label.costs.delete_v();
```
peterychang, line 136: Do we need to delete the v_array? Can we just clear it instead?

vowpalwabbit/explore_eval.cc
```diff
@@ -127,13 +127,15 @@ void do_actual_learning(explore_eval& data, multi_learner& base, multi_ex& ec_se
...
-    label_example->l.cb = data.empty_label;
+    data.action_label = std::move(label_example->l.cb);
+    label_example->l.cb = std::move(data.empty_label);
   }
   multiline_learn_or_predict<false>(base, ec_seq, data.offset);
 
   if (label_example != nullptr)  // restore label
-    label_example->l.cb = data.action_label;
+    label_example->l.cb = std::move(data.action_label);
+    data.empty_label.costs.delete_v();
```
jackgerrits, line 138: `delete_v` should no longer be used. Do you mean `clear` instead?

vowpalwabbit/cb_explore_adf_common.h
bassmang: Just recommitted with clear

vowpalwabbit/explore_eval.cc
bassmang: Just recommitted with clear

olgavrou (approved):

PR #3031, opened 2021-06-01
feat: [metrics] add ccb metrics

lalo:


jackgerrits (approved):

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -556,6 +566,21 @@ void output_example(vw& all, ccb& c, multi_ex& ec_seq)
         float l = CB_ALGS::get_cost_estimate(
             outcome->probabilities[TOP_ACTION_INDEX], outcome->cost, preds[i][TOP_ACTION_INDEX].action);
         loss += l * preds[i][TOP_ACTION_INDEX].score;
+
+        if (c._metrics)
+        {
+          c._metrics->metric_sum_cost += outcome->cost;
+
+          if (outcome->probabilities[TOP_ACTION_INDEX].action == 0)
```
lalo, line 574: ?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -556,6 +566,21 @@ void output_example(vw& all, ccb& c, multi_ex& ec_seq)
         float l = CB_ALGS::get_cost_estimate(
```
lalo, line 566: why is i==0 special

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -556,6 +566,21 @@ void output_example(vw& all, ccb& c, multi_ex& ec_seq)
...
             outcome->probabilities[TOP_ACTION_INDEX], outcome->cost, preds[i][TOP_ACTION_INDEX].action);
         loss += l * preds[i][TOP_ACTION_INDEX].score;
+
+        if (c._metrics)
```
lokitoth, line 570: How does this get initialized/cleared? Is there any way for the user to control the computation of metrics?

lalo: _metrics is a unique_ptr that ends up being nullptr if the user did not supply --extra_metrics arg. For now, most metrics get computed only for users that specify this flag. There are some exceptions of metrics we already compute for other reasons and we just reuse for this feature.

To initialize we would need to add support for it (fairly trivial, adding a bool to the persist_metrics fn similar to save_load read bool), clearing the values might be the same thing as initializing. It might make sense to add initialize from the previous metrics file specified.

PR #3034, opened 2021-06-03
[wip] slim cats with learners

olgavrou:
this is a very hacky **very very very non-ready** version of cats in slims re-using the reduction learners

most of the changes consist in removing any paths that contain `throw`'s and pulling in predict paths into header files so that they can be included into `vw_predict_slim.h`, a better solution would be to split predict and learn paths or something else entirely. I also want to later experiment with the idea of re-using setup functions

the interesting part is `vw_slim_predict.h` where the learners are manually activated and then called from slim predict

just opening up for clarity or in case anyone sees something extremely wrong with this approach


rajan-chari:
Some questions around size:
1) The delta size of the library?  
2) Any feel for how big an exe might be that uses cats?  For example using the release mode unit test executable.  Although that includes test data itself.

vowpalwabbit/slim/include/vw_slim_predict.h
```diff
@@ -372,7 +484,118 @@ class vw_predict
...
+    RETURN_ON_FAIL(mp.read_weights<W>(_predict_info->_weights, _num_bits, _stride_shift));
+
+    // TODO return error if activated but these are not initialized
+    if (pmf_to_pdf != nullptr && cats_tree != nullptr)
+    {
+      // TODO plug in interactions reduction and maybe metrics (or maybe figure out how to use setup functions)
+
+      // cats_tree -> binary -> gd
+      // create gd
+      auto* ret = VW::LEARNER::make_base_learner(
```
lalo, line 496: you might be able to leverage setup_base_i, maybe with custom setup functions specialized for slim? Just a thought, not sure if its actually useful here or not

olgavrou: Yes that might be the way to go

lalo: yes, you mentioned this a while back. I didn't realize it was about this PR

PR #3039, opened 2021-06-04
refactor: Refactor cb_sample.cc to use builder to set features of learner

bassmang:
Refactor cb_sample.cc to use builder to set features of learner


bassmang:
> [Nit] If possible, please add a short description of what the patch is intended to do.

Yep just added it

jackgerrits:
Can you please update the title too?

bassmang:
> Can you please update the title too?

Yes just changed that too

PR #3040, opened 2021-06-04
feat: add several functions to namespaced_features

jackgerrits:
Adds:
- `clear`
- Makes `get_indices` a constant time operation
- `get_index_for_hash`

Rename:
- `index` to `ns_index` in some places to clarify it is the 1 byte namespace index and not a `size_t` collection index

### DIFF
--- a/vowpalwabbit/namespaced_features.cc
+++ b/vowpalwabbit/namespaced_features.cc
@@ -23,13 +23,15 @@ const features* namespaced_features::get_feature_group(uint64_t hash) const
   return &_feature_groups[it->second];
 }
 
-std::vector<namespace_index> namespaced_features::get_indices() const
+const std::set<namespace_index>& namespaced_features::get_indices() const { return _contained_indices; }
+
+namespace_index namespaced_features::get_index_for_hash(uint64_t hash) const
 {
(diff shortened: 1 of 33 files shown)


vowpalwabbit/namespaced_features.cc
```diff
@@ -23,13 +23,15 @@ const features* namespaced_features::get_feature_group(uint64_t hash) const
...
+
+namespace_index namespaced_features::get_index_for_hash(uint64_t hash) const
 {
-  auto indices_copy = _namespace_indices;
-  std::sort(indices_copy.begin(), indices_copy.end());
-  auto last = std::unique(indices_copy.begin(), indices_copy.end());
-  indices_copy.erase(last, indices_copy.end());
-  return indices_copy;
+  auto it = _hash_to_index_mapping.find(hash);
+#ifndef VW_NOEXCEPT
```
olgavrou, line 31: if this is used with noexcept it will fail miserably, maybe add:

```
#else 
 if (it == end())
    return {};
```

and return an empty set

jackgerrits: Addressed

Issue #2474, opened 2020-06-18
Easier access to learned feature weight in python wrapper


etiennekintzler:
### Problem
In the python wrapper, **getting learned (features) weight should be easier**.

Currently the method to get learned weight of a given feature is :
1. find the hash value of the feature name using `pyvw.vw.hash_feature` using the feature name as 1st arg and the _hash_ of the namespace (being 0 for the naked namespace) as 2nd arg.
2. feed the feature name hash value to `pyvw.vw.get_weight`

**These two steps are not straightforward to figure out for newcomers**, and assume one is familiar with the hashing trick (to grasp why the learned feature are not at location 0, 1, 2, 3). 
 
### Solution

The solution is easy, just define the following function/method:
```python
def get_weight_from_name(model, name, namespace=""):
    return model.get_weight(model.hash_feature(name, model.hash_space(namespace)))
```

### Working example

Below is a quick example. I use _monkey patching_ on the second line (`pyvw.vw.get_weight_from_name = get_weight_from_name`) to add the method to the class at runtime but of course this method should be written directly in the class.

```python
from vowpalwabbit import pyvw
pyvw.vw.get_weight_from_name = get_weight_from_name
model = pyvw.vw(quiet=True)

train_examples = [
  "0 | price:.23 sqft:.25 age:.05 2006",
  "1 | price:.18 sqft:.15 age:.35 1976",
  "0 | price:.53 sqft:.32 age:.87 1924",
]

for example in train_examples:
    model.learn(example)
```
Then a quick listcomp to get the tuples _(feature name, learned weight)_: 
```python
[(name, model.get_weight_from_name(name)) for name in ["price", "sqft", "age"]]
```
The output is:
```python
[('price', 0.0018481612205505371),
 ('sqft', 0.021296292543411255),
 ('age', 0.0042297691106796265)]
```

If necessary I can do a PR.

Yash621:
Can I work on this issue?

priyanshuone6:
The `get_weight_from_name` is already implemented, is this still an ongoing issue?

jackgerrits:
You're right this is resolved. Closing

PR #3042, opened 2021-06-05
feat: [pyvw] Get weight from name

etiennekintzler:
Fix issue #2474: Add method to access the weight based on the feature name and the namespace name (if any).

The method `vw.get_weight_from_name` just takes the feature name and the namespace name as arguments and return the weight (a float). It is based on the method `vw.get_weight`.

I also add a test. Note that it tests equality between same feature weights and non nullity but doesn't directly test weight value as I don't know whether this value will always be the same between releases.

### Example

```python
from vowpalwabbit.pyvw import vw

# Learning
model = vw(quiet=True)

examples = [
    "1 |type color=blue square |dimensions width:2 height:4",
    "2 |type color=red square |dimensions width:3 height:4",
    "0 |type color=red circle |dimensions radius:1",
]

for ex in examples:
    model.learn(ex)

# Printing weights
ns_with_feat_names = {
    "type": ["color=blue", "color=red", "square", "circle"],
    "dimensions": ["width", "height"]
}

for ns_name, feat_names in ns_with_feat_names.items():
    print("namespace:", ns_name)
    for feat_name in feat_names:
        weight = model.get_weight_from_name(feat_name, ns_name)
        print(f"{feat_name.rjust(15)}: {round(weight, 4)}")
```

returns:

```python
Namespace: type
     color=blue: 0.1346
      color=red: 0.1243
         square: 0.2655
         circle: -0.1258
Namespace: dimensions
          width: 0.0933
         height: 0.0664

```


python/vowpalwabbit/pyvw.py
```diff
@@ -501,6 +501,25 @@ def get_weight(self, index, offset=0):
         """
         return pylibvw.vw.get_weight(self, index, offset)
 
+    def get_weight_from_name(self, feature_name, namespace_name=""):
```
jackgerrits, line 504: Maybe it makes sense for the default `namespace_name` to be `" "` since in text format it would correspond to `| feature`. " " is the default namespace

etiennekintzler: Yep sure !

Oddly it works whether the namespace name is a empty string `""`, one space `" "` or multiple spaces.

etiennekintzler:
The coverage test keeps failing. The reason seems not linked to the PR's code, would you have any idea about how to solve it @jackgerrits  ?

```sh
->  Pinging Codecov
https://codecov.io/upload/v4?package=bash-1.0.3&token=secret&branch=get_weight_from_name&commit=807dc93ffee69454ed75098ab6ed584d33dd2abb&build=909653602&build_url=http%3A%2F%2Fgithub.com%2FVowpalWabbit%2Fvowpal_wabbit%2Factions%2Fruns%2F909653602&name=&tag=&slug=VowpalWabbit%2Fvowpal_wabbit&service=github-actions&flags=&pr=3042&job=Coverage&cmd_args=Z
{'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
404
==> Uploading to Codecov
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 1066k    0     0  100 1066k      0  5169k --:--:-- --:--:-- --:--:-- 5150k
100 1066k  100   171  100 1066k    610  3807k --:--:-- --:--:-- --:--:-- 3807k
    {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
Error: Process completed with exit code 1.
```

jackgerrits:
It seems there may be some sort of spurious failure with either Codecov or GitHub actions. I have seen this before and I think it resolved itself after some amount of time. I'll keep a watch on it look deeper if it remains an issue

python/tests/test_pyvw.py
```diff
@@ -376,6 +376,19 @@ def test_example_features():
     ex.push_namespace(ns2)
     assert ex.pop_namespace()
 
+
+def test_get_weight_name():
+    model = vw(quiet=True)
+    model.learn("1 | a a b c |ns x")
+    assert model.get_weight_from_name("a") != 0.
```
lalo, line 383: what does get_weight_from_name behave when you ask for a feature that you haven't used before?

lalo: might be interesting to test out with --random_weights too

python/vowpalwabbit/pyvw.py
```diff
@@ -501,6 +501,25 @@ def get_weight(self, index, offset=0):
         """
         return pylibvw.vw.get_weight(self, index, offset)
 
+    def get_weight_from_name(self, feature_name, namespace_name=" "):
```
lalo, line 504: another fn that could be useful would be get_weight_from_interaction and ask for two features/namespaces, we just need to figure out the exact hash formula for interactions

lalo: not needed for this PR, it's just a thought

python/vowpalwabbit/pyvw.py
```diff
@@ -501,6 +501,25 @@ def get_weight(self, index, offset=0):
...
+            The name of the feature
+        namespace_name : str, by default is ""
+            The name of the namespace where the feature lives
+
+        Returns
+        -------
+        weight : float
+            Weight for the given feature and namespace name
+        """
+        space_hash = self.hash_space(namespace_name)
```
lokitoth, line 519: This method should take an optional "string value" input for JSON-based inputs, and will likely need special-casing for [chain_hash](https://github.com/VowpalWabbit/vowpal_wabbit/blob/21030f5c97b71f91224bfb44c4381316cf82efeb/vowpalwabbit/parse_args.cc#L456-L460) mode.

The key bit is that in the case of string-values features in JSON, there are two behaviours: One which concatenates the value to the string, and one which uses the hash-chaining trick. This currently defaults to former and lets you switch to the latter, but this is intended to change in a future version.

lalo: The wiki also mentions this: 

https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Input-format

> Features is a sequence of whitespace separated strings, each of which is optionally followed by a float (e.g., NumberOfLegs:4.0 HasStripes) or a string (e.g., city:paris). Each string is a feature and the value is the feature value for that example. Omitting a feature means that its value is zero. Including a feature but omitting its value means that its value is 1. When a string is supplied as the feature value, the value is 1 and the feature index is calculated as hash(feature_value, hash(feature_name, namespace_hash)) where hash's signature is hash(input, seed). This chained hashing is denoted by a ^ in the audit output.

python/tests/test_pyvw.py
etiennekintzler: > how does get_weight_from_name behave when you ask for a feature that you haven't used before?

It returns `0.`  which is default initial value for feature name that have not appeared (yet) in the learning phase.

etiennekintzler: > might be interesting to test out with --random_weights too

with `--random_weight` I am afraid we **won't be able to distinguish the weight initialization process from the weight update process**. To put differently, in the case of random weight initialization:
- every weight will be non 0 (except in the extremely rare case where the random weight is 0), which makes the test on nullity useless
- the equality between two weights  (for 2 features that appear in the same context such as  "b" and "c") will fail  since these 2 weights will still be different after 1 update. They might converge asymptotically (i.e after repeated updates) but that's another story.

python/vowpalwabbit/pyvw.py
etiennekintzler: > This method should take an optional "string value" input for JSON-based inputs, and will likely need special-casing for chain_hash mode.
>
> The key bit is that in the case of string-values features in JSON, there are two behaviours: One which concatenates the value to the string, and one which uses the hash-chaining trick. This currently defaults to former and lets you switch to the latter, but this is intended to change in a future version.

I am not sure how to integrate it since I have never used JSON-based inputs. Could you provide me with an example using the method `vw.get_weight`. Thanks !

python/vowpalwabbit/pyvw.py
etiennekintzler: > another fn that could be useful would be get_weight_from_interaction and ask for two features/namespaces, we just need to figure out the exact hash formula for interactions


Yes that would be quite useful !

python/vowpalwabbit/pyvw.py
jackgerrits: Chain hash actually works in text format too. It is for the use case of text feature values when you still want to use `:`. Activated with `--chain_hash` it would look like:
```
|test name:value
```

Would result in a feature index of:
```c++
hash("value", hash("name", hash("test", hash_seed)))

// Where hash("test", hash_seed) is the same as self.hash_space("test")
```

I think it could be reasonably modeled as the function taking an optional `feature_value_string` parameter and doing the chain hash in that case and if that parameter is `None` then using your current impl

lokitoth: The JSON analogue of your VW text example `1 | a a b c |ns x` is:


```json
{
   "_label": 1,
   "a": 2,
   "b": 1,
   "c": 1,
   "ns": 
   {
      "x": 1
   }
}
```

Not sure if is legal to have two of the same key in a single json object, so I combined them - thus they are not 100% equivalent.

You would also need to pass the "--json" flag to the vw object on init to switch the parser into JSON mode.

etiennekintzler: I try to parse this example using `--json` flag: 

```python
ex_json = {"_label": 1, "a": 2, "b": 1, "c": 1, "ns": {"x": 1}}
ex_json_str = str(ex_json)

model = vw("--json")
model.parse(ex_json_str)
```
but I get the following error : `RuntimeError: JSON parser error at 1: Missing a name for object member.. Handler: State: Default`

lokitoth: Let me take a look at what is going on here with the parsing, but it seems like you should be able to test chain-hash mode in text mode as well, as per @jackgerrits in the meanwhile

olgavrou: @etiennekintzler if you try `model.parse(json.dumps(ex_json))` should work

Also we can merge this now and open a ticket for chain hash or do you prefer to have it go in with this PR?

olgavrou (approved):

python/vowpalwabbit/pyvw.py
etiennekintzler: You are right @olgavrou `json.dumps(ex_json)` does work. The difference with `ex_json_str` is the presence of double quote for keys (which is JSON valid) vs single quote for the cast `str(ex_json)`.

Regarding chain hash, I don't feel competent tackling it since I haven't used the functionnality.  Based on what @jackgerrits said I tried the following code but didn't work out (weight is 0.0):

```python
from vowpalwabbit.pyvw import vw

model = vw("--chain_hash")
for ex in ["1 |test name:value"]:
    model.learn(ex)

ns_hash = model.hash_space("test")
key_hash = model.hash_feature("name", ns_hash)
value_hash = model.hash_feature("value", key_hash)
model.get_weight(value_hash) # 0.0
```

I am using the last vw version (8.11.0). Do you also have the same result ?

Otherwise, I think you can merge this PR if it looks good to you !

jackgerrits: Looks right to me at first glance so I'm not sure what's up there. I think we should merge as this is super useful! Then someone can tackle chain hash separately with an optional argument

olgavrou: created issue to track this #3081

olgavrou: also yes @etiennekintzler tried it on master and I get the same as you

olgavrou: chain hash applied the parse mask which is probably why the above isn't working see [here](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/vw.h#L203)

lokitoth: Looks like the plan is to deal with chain-hash after getting this PR in, right @olgavrou, @etiennekintzler? If so, looks good.

lokitoth (approved):

python/vowpalwabbit/pyvw.py
olgavrou: @lokitoth yup see #3081

PR #3046, opened 2021-06-07
feat: Migrate example_predict::feature_space to namespaced_features

jackgerrits:


olgavrou (approved):
nit: the PR name is a bit misleading (since it's the feature group info and the iterator that is being migrated)

jackgerrits:
namespaced_features is the structure that holds all of the features. The old static 256 length array of features is being replaced with this new type. I updated the title to clarify it is a member that is changing not the owning structure

PR #3048, opened 2021-06-07
ci: Reduce CI burden

jackgerrits:
- Only run slow CIs when relevant paths change and 
- Only build master and release push based triggers

### DIFF
--- a/.github/workflows/build_python_wheels.yml
+++ b/.github/workflows/build_python_wheels.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - master
-      - releases/**
+      - 'releases/**'
   pull_request:
     branches:
       - '*'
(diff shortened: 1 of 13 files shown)


lalo (approved):

olgavrou (approved):

.github/workflows/build_python_wheels.yml
```diff
@@ -7,8 +7,7 @@ on:
       - 'releases/**'
   pull_request:
     branches:
-      - master
-
+      - '*'
```
lokitoth, line 10: Is it intended that you will build on PR to any branch? Or am I misunderstanding what is going on here?

jackgerrits: Yes that is intended. This is useful for example for a feature branch

PR #3052, opened 2021-06-08
refactor: move reduction stack building out of parse_args.cc

lalo:
- most of the code from parse_args.cc moved to reduction_stack.cc
- build state is not part of all object anymore, see changes in global_data.h


vowpalwabbit/cached_learner.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "global_data.h"  // to get vw struct
+#include "options.h"      // to get options_i
+
+namespace VW
+{
+struct cached_learner : public setup_base_fn
+{
+  VW::LEARNER::base_learner* _cached = nullptr;
```
olgavrou, line 10: should _cached be private here?

vowpalwabbit/reduction_stack.h
```diff
@@ -0,0 +1,26 @@
...
+struct status_quo : public setup_base_fn
+{
+  std::vector<std::string> enabled_reductions;
+
+  void print_enabled_reductions(vw& all);
+
+  status_quo(vw& all);
+
+  // this function consumes all the reduction_stack until it's able to construct a base_learner
+  // same signature as the old setup_base(...) from parse_args.cc
```
olgavrou, line 20: nit: `same signature as ...` do we want this as a comment in the code? maybe a comment in the PR for tracking?

vowpalwabbit/reduction_stack.h
```diff
@@ -0,0 +1,26 @@
...
+
+#include "reductions_fwd.h"
+
+struct vw;
+
+typedef VW::LEARNER::base_learner* (*reduction_setup_fn)(VW::setup_base_fn&, VW::config::options_i&, vw&);
+
+namespace VW
+{
+struct status_quo : public setup_base_fn
```
olgavrou, line 11: `status_quo` -> `default_reduction_stack_setup` or something more descriptive, `status_quo` will loose it's meaning in a couple of months

vowpalwabbit/reduction_stack.h
```diff
@@ -0,0 +1,26 @@
...
+{
+  std::vector<std::string> enabled_reductions;
+
+  void print_enabled_reductions(vw& all);
+
+  status_quo(vw& all);
+
+  // this function consumes all the reduction_stack until it's able to construct a base_learner
+  // same signature as the old setup_base(...) from parse_args.cc
+  VW::LEARNER::base_learner* operator()(VW::config::options_i& options, vw& all);
```
olgavrou, line 21: I think this needs an `override` since it's implementing a pure virtual function from base

vowpalwabbit/cached_learner.h
lalo: done

vowpalwabbit/reduction_stack.h
lalo: done

vowpalwabbit/reduction_stack.h
lalo: done

vowpalwabbit/reduction_stack.cc
```diff
@@ -0,0 +1,270 @@
...
+}
+
+namespace VW
+{
+default_reduction_stack_setup::default_reduction_stack_setup(vw& all)
+{
+  // push all reduction functions into the stack
+  prepare_reductions(reduction_stack);
+  // populate setup_fn -> name map to be used to lookup names in setup_base
+  all.build_setupfn_name_dict(reduction_stack);
```
lalo, line 220: this also could be moved from all struct to default_reduction_stack_setup

olgavrou (approved):

lalo:
This should go before https://github.com/VowpalWabbit/vowpal_wabbit/pull/2971

vowpalwabbit/reduction_stack.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include "reductions_fwd.h"
+
+struct vw;
+
+typedef VW::LEARNER::base_learner* (*reduction_setup_fn)(VW::setup_base_fn&, VW::config::options_i&, vw&);
```
rajan-chari, line 7: setup_fn here is a pure virtual class with instantiations that create their own reduction stack.   Can the same thing can not be achieved if setup_fn decides what stack to instantiate?  This achieves the same thing without adding another parameter to the setup function of each reduction.

lalo: - The concern of adding an extra parameter is interesting, since in another PR after this one you could easily collapse options and all parameters inside setup_base_fn interface. If you care about parameter count, the count would get reduced to 1 and this 'object' starts to represent exactly what it is: a collection of configs and a magic box that gives you the base learner. You get the side benefit of having specific APIs that hide away VW all global state.
- Yes, you could achieve the same thing with a separate and alternate impls of current setup functions. I'm proposing that you can get for free with a single impl of setup function by breaking that coupling between setup functions all over the place and parse_args.h:setup_base fn. You can override this dependency on runtime by following the proposed interface in this PR.

lalo: https://github.com/VowpalWabbit/vowpal_wabbit/blob/d5bb16e2227148082aadf48daa6be050f139321d/vowpalwabbit/parse_args.cc#L1415

You can imagine alternate builders with other behaviour, like cached_learner that always returns the learner which it was constructed with like so: https://github.com/VowpalWabbit/vowpal_wabbit/blob/d5bb16e2227148082aadf48daa6be050f139321d/vowpalwabbit/cached_learner.h

rajan-chari: Makes sense.  setup_fn evolving to incorporate a factory for base reductions and other facilities that a reduction needs seems like a good general direction.  This makes it easier for a reduction to live in a different host outside vw.exe as long as the host provides all the same facilities.... like a factory for the next reduction, a random number generator, log facility, state save/resume facility etc.

rajan-chari (approved):

vowpalwabbit/binary.h
```diff
@@ -9,6 +9,6 @@ namespace VW
 {
 namespace binary
 {
-VW::LEARNER::base_learner* binary_setup(VW::config::options_i& options, vw& all);
+VW::LEARNER::base_learner* binary_setup(setup_base_fn& setup_base_fn, VW::config::options_i& options, vw& all);
```
kumpera, line 12:
```suggestion
VW::LEARNER::base_learner* binary_setup(setup_base_fn& setup_base, VW::config::options_i& options, vw& all);
```
nit: arg name consistency.

vowpalwabbit/global_data.h
```diff
@@ -276,8 +278,7 @@ struct vw
 
   size_t length() { return (static_cast<size_t>(1)) << num_bits; };
 
-  std::vector<std::tuple<std::string, reduction_setup_fn>> reduction_stack;
-  std::vector<std::string> enabled_reductions;
+  std::unique_ptr<VW::default_reduction_stack_setup> learner_builder;
```
kumpera, line 281: The learner_builder should be a pointer to an abstract object, otherwise we won't be able to effectively override its behavior.

vowpalwabbit/reduction_stack.h
```diff
@@ -0,0 +1,26 @@
...
+
+namespace VW
+{
+struct default_reduction_stack_setup : public setup_base_fn
+{
+  std::vector<std::string> enabled_reductions;
```
kumpera, line 13: Public fields is usually bad.

kumpera (approved):
One thing about this PR that doesn't address is the decoupling between the transient state of build the reduction stack and the final state of the object.

It's an usual sign of problematic design when long lived objects retain vestigial setup data. In this case default_reduction_stack_setup keeps reduction_stack around.

I don't think this issue should block this PR as it's progress in the right direction.

vowpalwabbit/reduction_stack.h
lalo: I'll address this with https://github.com/VowpalWabbit/vowpal_wabbit/pull/3176

vowpalwabbit/parse_args.cc
```diff
@@ -1807,11 +1573,20 @@ vw* initialize(std::unique_ptr<options_i, options_deleter_type> options, io_buf*
...
+        cout << all.options->help({});
+      }
       exit(0);
     }
 
-    print_enabled_reductions(all);
+    // output list of enabled reductions
+    if (all.learner_builder)
+    {
+      if (!all.logger.quiet && !all.options->was_supplied("audit_regressor"))
```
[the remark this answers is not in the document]
lalo: This is a 'thing' I've seen across the code, audit_regressor being checked everytime we print something. Not sure why is that but it's big enough to get its own PR or refactor.

vowpalwabbit/global_data.h
lalo: This will be removed after merging with https://github.com/VowpalWabbit/vowpal_wabbit/pull/3176

vowpalwabbit/reduction_stack.h
lalo: done

python/pylibvw.cc
```diff
@@ -17,6 +17,7 @@
 #include "slates_label.h"
 #include "simple_label_parser.h"
 #include "shared_data.h"
+#include "reduction_stack.h"
```
lalo, line 20:
```suggestion
```

vowpalwabbit/parse_args.cc
```diff
@@ -1631,7 +1409,15 @@ void parse_modules(
...
 
-  parse_reductions(options, all);
+  // create reduction stack builder instance
+  auto learner_builder = VW::make_unique<VW::default_reduction_stack_setup>(all);
+  // kick-off reduction setup functions
+  all.l = learner_builder->operator()(options, all);
+
+  // explicit destroy of learner_builder state
+  // avoids misuse of this interface:
+  learner_builder.reset();
```
lalo, line 1419: added this as a 'canary in the coalmine' for future PRs that expect learner_builder instance that is being passed thru setup reduction functions to be a long-lived one, when it should not be the case.

kumpera: Good idea.
We can revisit this later if we need some sort of clone & slice functionality for things like automl in case we want, for example, to have coin as a challenger to SGD. I currently have no idea on how that would it look like.

kumpera (approved):
👍

vowpalwabbit/global_data.h
```diff
@@ -55,7 +55,8 @@
 typedef float weight;
 
 typedef std::unordered_map<std::string, std::unique_ptr<features>> feature_dict;
-typedef VW::LEARNER::base_learner* (*reduction_setup_fn)(VW::config::options_i&, vw&);
+// typedef VW::LEARNER::base_learner* (*setup_base_i)(VW::config::options_i&, vw&);
```
lalo, line 58:
```suggestion
```

PR #3053, opened 2021-06-08
refactor: Refactor cats_tree.cc to use builder to set features of learner

bassmang:


vowpalwabbit/cats_tree.cc
```diff
@@ -360,16 +361,18 @@ base_learner* setup(options_i& options, vw& all)
     options.replace("link", "glf1");
   }
 
-  auto tree = scoped_calloc_or_throw<cats_tree>();
+  auto tree = VW::make_unique<cats_tree>();
```
jackgerrits, line 364: It looks like `_cost_star, _a, _b` are unitialized after this call.

bassmang: Just recommitted. Let me know if manually initializing _a and _b is preferred to creating a constructor in node_cost.

vowpalwabbit/cats_tree.h
```diff
@@ -72,7 +73,7 @@ struct cats_tree
   uint64_t app_seed = uniform_hash("vw", 2, 0);
   std::string tree_stats_to_string();
   min_depth_binary_tree _binary_tree;
-  float _cost_star;
+  float _cost_star = 0;
```
jackgerrits, line 76: Should be a float literal `0.f`

vowpalwabbit/cats_tree.h
```diff
@@ -54,6 +54,7 @@ struct node_cost
 {
   uint32_t node_id;
   float cost;
+  node_cost(uint32_t node_id = 0, float cost = 0) : node_id(node_id), cost(cost) {}
```
jackgerrits, line 57: This is nit, so feel free to wontfix. I'd say updating lines 55 and 56 is preferable as default params in constructors is a but confusing and the default constructor compared to this constructor with default parameters is ambiguous.  (well default won't be generated in this case, and I think default default constructor is usually best)
```c++
uint32_t node_id = 0;
float cost = 0.f;
```

vowpalwabbit/cats_tree.cc
jackgerrits: Having the node_cost initialization done internally is good as it affects any other usage too. So this approach is better

vowpalwabbit/cats_tree.h
bassmang: I actually tried this initially, however it causes a compilation error with commands such as "_a = {node_id, _cost_star};"

jackgerrits: I see. In that case I think we should have two constructors. Default and two parameter rather than a since one with default parameters.

bassmang: Ok just recommitted.

jackgerrits (approved):

PR #3055, opened 2021-06-09
fix: Fix of learning on data with pdrop

### LINKED ISSUE #3054: Learning on data with pdrop is wrong
#### Describe the bug
cb.json
{"_label_cost":-1.0,"_label_probability":0.5,"_label_Action":2,"_labelIndex":1,"a":[2,1],"c":{"shared":{"f":"1"},"_multi":[{"action":{"f":"1"}},{"action":{"f":"2"}}]},"p":[0.5,0.5]}

cb_pdrop.json
`{"_label_cost":-1.0,"_label_probability":0.5,"_label_Action":2,"_labelIndex":1,"a":[2,1],"c":{"shared":{"f":"1"},"_multi":[{"action":{"f":"1"}},{"action":{"f":"2"}}]},"p":[0.5,0.5], pdrop:0.9}
`

#### To Reproduce
1. `vw --cb_explore_adf --dsjson -d cb.json --invert_hash model.txt`
2. `vw --cb_explore_adf --dsjson -d cb_pdrop.json --invert_hash model_pdrop.txt` 
3. Compare model.txt and model_pdrop.txt and see that weights in model.txt has higher abs value than for model_pdrop.txt

#### Expected behavior
Example with pdrop should have higher impact on learning

ataymano:
This one is fixing only learning.
Separate bug for issue with reported loss is created (it is less urgent since probably affecting metrics only, but fixing it breaking many tests)
Fixes #3054


ataymano:
#3054

vowpalwabbit/parse_example_json.h
```diff
@@ -1670,11 +1670,11 @@ inline void apply_pdrop(vw& all, float pdrop, v_array<example*>& examples)
 {
   if (all.example_parser->lbl_parser.label_type == label_type_t::cb)
   {
-    for (auto& e : examples) { e->l.cb.weight = 1 - pdrop; }
+    for (auto& e : examples) { e->l.cb.weight /= 1 - pdrop; }
```
lokitoth, line 1673: [Minor] Consider adding a comment explaining what is going on here, mathematically.

jackgerrits (approved):

vowpalwabbit/parse_example_json.h
```diff
@@ -1670,11 +1670,11 @@ inline void apply_pdrop(vw& all, float pdrop, v_array<example*>& examples)
 {
   if (all.example_parser->lbl_parser.label_type == label_type_t::cb)
```
lalo, line 1671: is this bug just relevant for json parsing?

ataymano: in the way that "pdrop" as concept exists only in json format.

vowpalwabbit/parse_example_json.h
ataymano: done

vowpalwabbit/parse_example_json.h
```diff
@@ -1668,13 +1668,15 @@ void read_line_json(vw& all, v_array<example*>& examples, char* line, example_fa
 
 inline void apply_pdrop(vw& all, float pdrop, v_array<example*>& examples)
 {
+  // Event with certain pdrop had (1-pdrop) as probability to survive, 
+  // so it is one of (1 / (1-pdrop)) events that we should learn on, and weight should be updated accordingly.
```
olgavrou, line 1672: doesn't this imply that there are `1-pdrop` events so each one has a probability of `1/(1-pdrop)`?

should we multiplying with `1-pdrop` instead of `1/(1-pdrop)`?

olgavrou: ok I understand we want higher impact (weight) for higher pdrop which is why we are dividing here (maybe the comment got me confused)

different concern: if pdrop is 1 we are dividing by zero here

ataymano: yeah that is true. updated.

olgavrou (approved):

vowpalwabbit/parse_example_json.h
lalo: why not set the weight to 0, wouldnt that be equivalent to dropping all events, pdrop of 1

ataymano: it may work, but could also be some corner case (if there is division by weight somewhere), which has to be extra tested.

jackgerrits (approved):

PR #3057, opened 2021-06-09
feat: add cache helper for reading and writing types directly

jackgerrits:


vowpalwabbit/io_buf.h
```diff
@@ -212,6 +212,36 @@ class io_buf
...
+
+  template <typename T>
+  T read_value(const char* debug_name = nullptr)
+  {
+    char* c;
+    T value;
+    if (buf_read(c, sizeof(T)) < sizeof(T))
+    {
+      if (debug_name != nullptr)
+      { THROW("Failed to read cache value: " << debug_name << ", with size: " << sizeof(T)); }
```
lokitoth, line 233: This is probably out of scope for this change, but does it make sense to have a mode where an error in reading from the cache would cause the learning to switch back to reading from non-cache? Or is the coordination hard enough to make it not worth the time?

jackgerrits: In the current landscape the coordination to achieve that is hard enough to not make it worth it. I can see it being a nice thing to have in future though.

olgavrou (approved):

PR #3058, opened 2021-06-09
refactor: Organize cache usage behind two functions for testing

jackgerrits:
- Make cache usage const correct
- Add test for feature reading and writing
- Add v_array `operator==`

### DIFF
--- a/vowpalwabbit/v_array.h
+++ b/vowpalwabbit/v_array.h
@@ -413,6 +413,12 @@ public:
 
     return false;
   }
+
+  friend bool operator==(const v_array<T>& lhs, const v_array<T>& rhs)
+  {
+    if (lhs.size() != rhs.size()) { return false; }
+    return std::equal(lhs.begin(), lhs.end(), rhs.begin());
+  }
 };
 
 template <class T>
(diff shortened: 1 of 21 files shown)


vowpalwabbit/v_array.h
```diff
@@ -412,6 +412,12 @@ struct v_array<T, typename std::enable_if<std::is_trivially_copyable<T>::value>:
 
     return false;
   }
+
+  friend bool operator==(const v_array<T>& lhs, const v_array<T>& rhs)
```
olgavrou, line 416: is this being used in this PR?

olgavrou: the benchmarks for text parsing seem to be complaining a bit for this PR (cache looks good) so was wondering what it could be

jackgerrits: It's being used in the tests to compare v_arrays, but not in the library code anywhere (explicitly). I am not sure what could really be having that effect on the benchmarks as this PR is just a rearrange, adding consts and adding a test

jackgerrits: Just verified that this is not called in the benchmarks

olgavrou (approved):

jackgerrits:
I removed the const additions and the move to numeric casts and this seems to have resolved the performance issues

PR #3064, opened 2021-06-11
refactor: isolate processing of interaction args

lalo:


vowpalwabbit/parse_args.cc
```diff
@@ -614,6 +614,105 @@ std::string spoof_hex_encoded_namespaces(const std::string& arg)
   return res;
 }
 
+void process_quadratics(vw& all, std::vector<std::vector<namespace_index>>& decoded_interactions,
```
lokitoth, line 617: [Minor] Can these two (/three?) methods be unified behind a template?

PR #3069, opened 2021-06-14
fix: Baseline should signal its enablement through an index and not a feature

jackgerrits:
This unblocks namespaced features

### DIFF
--- a/vowpalwabbit/baseline.cc
+++ b/vowpalwabbit/baseline.cc
@@ -2,6 +2,8 @@
 // individual contributors. All rights reserved. Released under a BSD (revised)
 // license as described in the file LICENSE.
 
+#include "baseline.h"
+
 #include <cfloat>
 #include <cerrno>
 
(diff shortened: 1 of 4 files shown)


vowpalwabbit/baseline.cc
```diff
@@ -15,47 +17,25 @@ using namespace VW::config;
...
 bool baseline_enabled(example* ec)
 {
-  auto& fs = ec->feature_space[message_namespace];
-  for (auto& f : fs)
-  {
-    if (f.index() == baseline_enabled_idx) return f.value() == 1;
-  }
-  return false;
+  auto it = std::find(ec->indices.begin(), ec->indices.end(), baseline_enabled_message_namespace);
+  return it != ec->indices.end();
```
lalo, line 38: line 190 of the right side (in case you are on diff mode), implies that this reduction only works with the global interactions and not the per example ones, is this an issue?

vowpalwabbit/constant.h
```diff
@@ -31,7 +31,7 @@ constexpr unsigned char spelling_namespace = 133;      // this is \x85
 constexpr unsigned char conditioning_namespace = 134;  // this is \x86
 constexpr unsigned char dictionary_namespace = 135;    // this is \x87
 constexpr unsigned char node_id_namespace = 136;       // this is \x88
-constexpr unsigned char message_namespace = 137;       // this is \x89
+constexpr unsigned char baseline_enabled_message_namespace = 137;  // this is \x89
```
lalo, line 34: was message_namespace used for baseline.cc before? or is this being reused?

vowpalwabbit/baseline.cc
jackgerrits: I'm not sure how this reduction is used. It may be an issue.

vowpalwabbit/constant.h
jackgerrits: It was only used by baseline.cc. I am making it clearer what it does

lalo (approved):

lalo (changes requested):
it would've been better to add the test before you change the code, but I guess its the same impl. your call

lalo (approved):

jackgerrits:
Test added separately in #3070

PR #3071, opened 2021-06-15
fix: cmake build

AdamStepan:
This MR fixes:
- build with external RapidJSON library
- cmake install target


jackgerrits (approved):

jackgerrits:
Thanks for the fixes @AdamStepan !

PR #3076, opened 2021-06-16
feat: add chained_proxy_iterator to make it easier to iterate entire namespace index groups

jackgerrits:


vowpalwabbit/namespaced_features.cc
```diff
@@ -158,6 +158,124 @@ generic_range<namespaced_features::const_indexed_iterator> namespaced_features::
   return {namespace_index_begin(ns_index), namespace_index_end(ns_index)};
 }
 
+VW::chained_proxy_iterator<namespaced_features::indexed_iterator, features::audit_iterator>
```
olgavrou, line 161: would a comment header explaining a bit what this code is doing here might be helpful

jackgerrits: Added some comments

vowpalwabbit/namespaced_features.cc
```diff
@@ -158,6 +158,128 @@ generic_range<namespaced_features::const_indexed_iterator> namespaced_features::
...
+  {
+    --end_it;
+    inner_it = features::audit_iterator{};
+  }
+  else
+  {
+    --end_it;
+    inner_it = (*begin_it).audit_begin();
+  }
+  // end_it always points to the last valid outer iterator instead of the actual end iterator of the outer collection.
```
olgavrou, line 179: do you mean this?

```suggestion
  // end_it always points to the last valid inner iterator instead of the actual end iterator of the outer collection.
```

jackgerrits: Outer is correct since end_it is the iterator to the list and we point to the last list and not the end iterator of the lists

test/unit_test/namespaced_features_test.cc
```diff
@@ -36,3 +36,124 @@ BOOST_AUTO_TEST_CASE(namespaced_features_test)
...
+  fs3.push_back(1.0, 7);
+  fs3.push_back(1.0, 8);
+
+  auto it = feature_groups.namespace_index_begin_proxy('a');
+  BOOST_REQUIRE_EQUAL((*it).index(), 1);
+  it += 4;
+  BOOST_REQUIRE_EQUAL((*it).index(), 5);
+  it += 3;
+  BOOST_REQUIRE_EQUAL((*it).index(), 8);
+}
```
olgavrou, line 159: could we possibly add a while loop test? I am having trouble understanding the end

jackgerrits: What would I assert in the body of the while loop?

olgavrou: we could check that all the values are 1?

jackgerrits: That makes sense! Okay will add

olgavrou (approved):

PR #3078, opened 2021-06-17
build: Enable building Python library using VS2019

lokitoth:
* Enables user-supplied CMake generator --global-option
* Updates setup script to use Python setuptools to discover headers

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,7 @@ import platform
 import sys
 from codecs import open
 from distutils.command.clean import clean as _clean
+from distutils.sysconfig import get_python_inc
 from setuptools import setup, Extension, find_packages, Distribution as _distribution
 from setuptools.command.build_ext import build_ext as _build_ext
 from setuptools.command.sdist import sdist as _sdist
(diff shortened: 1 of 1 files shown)


jackgerrits (approved):

olgavrou (commented):
Does this get exercised in a CI?

lokitoth:
@olgavrou Yes, in multiple different OSes and target versions - so it spotted my typo around which cmake_generator variable I was using, and failed multiple times.

olgavrou:
> @olgavrou Yes, in multiple different OSes and target versions - so it spotted my typo around which cmake_generator variable I was using, and failed multiple times.
> 
> Or do you mean whether we have a VS2019 CI? The answer to that is no, not at the moment.

Yes I meant in our existing CI collection, @jackgerrits added [this](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/.github/workflows/build_windows_cmake.yml#L53) relatively recently would it be worth building this there?

lokitoth:
@olgavrou Honestly, probably not really. We should check whether it is reasonable to move to VS 2019 for building Python by default and simply migrate to it. If it does not reduce Windows version compatibility to a place we do not want it. The goal of this PR was simply to make it possible to inject a custom toolchain.

olgavrou:
> @olgavrou Honestly, probably not really. We should check whether it is reasonable to move to VS 2019 for building Python by default and simply migrate to it. If it does not reduce Windows version compatibility to a place we do not want it. The goal of this PR was simply to make it possible to inject a custom toolchain.

Fair enough

olgavrou (approved):

PR #3090, opened 2021-06-23
ci: Build docs as part of CI

jackgerrits:


.github/workflows/build_docs.yml
```diff
@@ -59,14 +63,14 @@ jobs:
   upload:
     needs: [cpp, python]
     runs-on: ubuntu-latest
-    # The upload step should only be run on the main repository.
-    if: github.repository == 'VowpalWabbit/vowpal_wabbit'
+    # The upload step should only be run on the main repository and only for pushes or releases (not pull requests).
```
lalo, line 66: so we get both latest master and releases docs at the same time?

jackgerrits: Not at the same time, but this CI deploys docs for both master and for tagged releases

lalo (approved):

PR #3094, opened 2021-06-23
refactor: migrate namespaced_features to ezexample.h

bassmang:


vowpalwabbit/ezexample.h
```diff
@@ -131,43 +129,44 @@ class ezexample
...
-    if (ec->indices.empty())
+    if (ec->feature_space.empty())
     {
       current_seed = 0;
       current_ns = 0;
     }
     else
     {
-      if (ns_exists[static_cast<int>(current_ns)])
+      if (ns_exists[current_seed])
```
jackgerrits, line 158: ns_exists is still the bool array right? If so, then current_seed will quickly over index it

bassmang: Yes just switched that back

jackgerrits (approved):

PR #3096, opened 2021-06-24
test: add tests for namespaces with same first letter

bassmang:


vowpalwabbit/slim/test/ut_vw.cc
```diff
@@ -127,10 +128,10 @@ void run_predict_in_memory(
     // 1 |0 0:1 |a 0:2
     example_predict_builder b00(&ex[0], (namespace_index)0);
     b00.push_feature(0, 1.f);
-    example_predict_builder b0a(&ex[0], "a");
+    example_predict_builder b0a(&ex[0], (char*)"a");
```
jackgerrits, line 131: I do not think you need these casts?

PR #3105, opened 2021-06-25
ci: Add Python type checking

jackgerrits:


jackgerrits:
This is having troubles with the native dependency. I'll need to look further into how to resolve that

lalo (approved):

PR #3110, opened 2021-06-28
refactor: python bindings

bassmang:


python/pylibvw.cc
```diff
@@ -542,32 +542,85 @@ std::string varray_char_to_string(v_array<char>& a)
...
+  auto it = ec->begin();
+  for (int i = 0; i < distance; ++i, ++it);
+  return it.index();
+}
 
-uint32_t ex_num_features(example_ptr ec, unsigned char ns) { return (uint32_t)ec->feature_space[ns].size(); }
+uint32_t ex_num_features(example_ptr ec, char* ns) {
+  auto fs = ec->feature_space;
+  uint64_t hash = hashstring(ns, strlen(ns), 0);
+  return (strlen(ns) == 1) ? fs.namespace_index_end_proxy(ns[0]) - fs.namespace_index_begin_proxy(ns[0]) : fs[hash].size();
```
jackgerrits, line 557: It is still valid to have a namespace of length 1 and access it using the hash so we need to make sure it works with that. So it may best to use an explicit bool parameter for switching behavior

python/pylibvw.cc
```diff
@@ -542,32 +542,85 @@ std::string varray_char_to_string(v_array<char>& a)
...
-float ex_sum_feat_sq(example_ptr ec, unsigned char ns) { return ec->feature_space[ns].sum_feat_sq; }
+float ex_sum_feat_sq(example_ptr ec, char* ns)
+{
+  auto fs = ec->feature_space;
+  if (strlen(ns) == 1) 
+  {
+    auto beg = fs.namespace_index_begin(ns[0]);
+    auto end = fs.namespace_index_end(ns[0]);
+    float sum_feat_sq = 0.f;
+    for (auto it = beg; it != end; ++it) sum_feat_sq += (*it).sum_feat_sq;
```
jackgerrits, line 602: You can use `namespace_index_range` for cases like this

python/pylibvw.cc
```diff
@@ -542,32 +542,85 @@ std::string varray_char_to_string(v_array<char>& a)
...
 
-uint32_t ex_num_features(example_ptr ec, unsigned char ns) { return (uint32_t)ec->feature_space[ns].size(); }
+uint32_t ex_num_features(example_ptr ec, char* ns) {
+  auto fs = ec->feature_space;
+  uint64_t hash = hashstring(ns, strlen(ns), 0);
+  return (strlen(ns) == 1) ? fs.namespace_index_end_proxy(ns[0]) - fs.namespace_index_begin_proxy(ns[0]) : fs[hash].size();
+}
 
-uint32_t ex_feature(example_ptr ec, unsigned char ns, uint32_t i)
+uint32_t ex_feature(example_ptr ec, char* ns, uint32_t i)
```
jackgerrits, line 560: All of these `char*` should be `const char*`

python/pylibvw.cc
```diff
@@ -655,39 +711,67 @@ void ex_push_dictionary(example_ptr ec, vw_ptr vw, py::dict& dict)
     if (ns_e().length() < 1) continue;
     py::extract<py::list> list_e(objectVal);
     py::list list = list_e();
-    char ns = ns_e()[0];
+    std::string ns_str = ns_e();
```
jackgerrits, line 714: ns_e() is already a string copy, you don't need to copy it into an additional buffer you can pass .c_str() directly to ex_ensure_namespace_exists

python/pylibvw.cc
```diff
@@ -655,39 +711,67 @@ void ex_push_dictionary(example_ptr ec, vw_ptr vw, py::dict& dict)
...
+    char* ns = new char [ns_str.length()+1];
```
jackgerrits, line 715: This is leaking memory

python/pylibvw.cc
```diff
@@ -655,39 +711,67 @@ void ex_push_dictionary(example_ptr ec, vw_ptr vw, py::dict& dict)
...
 {
-  if (ec->feature_space[ns].size() == 0) return false;
-  float val = ec->feature_space[ns].values.back();
-  ec->feature_space[ns].values.pop_back();
-  if (ec->feature_space[ns].indicies.size() > 0) ec->feature_space[ns].indicies.pop_back();
-  if (ec->feature_space[ns].space_names.size() > 0) ec->feature_space[ns].space_names.pop_back();
+  uint64_t hash = hashstring(ns, strlen(ns), 0);
+  features& feats = ec->feature_space[hash];
+  if (feats.size() == 0) return false;
+  float val = feats.values.back();
```
jackgerrits, line 727: All if statements require braces

python/pylibvw.cc
```diff
@@ -655,39 +711,67 @@ void ex_push_dictionary(example_ptr ec, vw_ptr vw, py::dict& dict)
...
+  if (strlen(ns) == 1)
+  {
+    auto beg = ec->feature_space.namespace_index_begin(ns[0]);
+    auto end = ec->feature_space.namespace_index_end(ns[0]);
+    std::vector<uint64_t> rem_hashes;
+    for (auto it = beg; it != end; ++it) {
+      features& feats = *it;
+      ec->num_features -= feats.size();
+      feats.sum_feat_sq = 0.f;
+      feats.clear();
```
jackgerrits, line 748: There's no need to do the sum_feat_sq and clear operations if you're about to remove it

python/pylibvw.cc
```diff
@@ -655,39 +711,67 @@ void ex_push_dictionary(example_ptr ec, vw_ptr vw, py::dict& dict)
...
 {
   if (ec->indices.size() == 0) return false;
-  unsigned char ns = ec->indices.back();
-  ec->indices.pop_back();
-  ex_erase_namespace(ec, ns);
+  uint64_t hash = ec->feature_space.get_last_hash();
+  features& feats = ec->feature_space[hash];
+  ec->num_features -= feats.size();
+  feats.sum_feat_sq = 0.f;
+  feats.clear();
```
jackgerrits, line 772: same here

python/pylibvw.cc
```diff
@@ -700,38 +784,38 @@ void unsetup_example(vw_ptr vwP, example_ptr ae)
...
   if (all.ignore_some) { THROW("error: cannot unsetup example when some namespaces are ignored!"); }
 
   if (all.skip_gram_transformer != nullptr && !all.skip_gram_transformer->get_initial_ngram_definitions().empty())
   { THROW("error: cannot unsetup example when ngrams are in use!"); }
 
   if (all.add_constant)
   {
-    ae->feature_space[constant_namespace].clear();
-    int hit_constant = -1;
+    (*ae->feature_space.namespace_index_begin(constant_namespace)).clear();
```
jackgerrits, line 794: The constant namespace should be removed not just cleared

jackgerrits (changes requested):

python/pylibvw.cc
```diff
@@ -544,22 +544,46 @@ std::string my_get_tag(example_ptr ec) { return varray_char_to_string(ec->tag);
 
 uint32_t ex_num_namespaces(example_ptr ec) { return (uint32_t)ec->indices.size(); }
 
-unsigned char ex_namespace(example_ptr ec, uint32_t ns) { return ec->indices[ns]; }
+unsigned char ex_namespace(example_ptr ec, uint32_t distance) 
+{
+  auto it = ec->begin();
+  for (int i = 0; i < distance; ++i, ++it);
```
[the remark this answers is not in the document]
bassmang: It did not so I added one

python/pylibvw.cc
bassmang: I changed the logic in these cases to check if the length is 1, and if the hash does not exist

python/pylibvw.cc
bassmang: This works for most cases, however if you need to access the hash I dont think this is possible

python/pylibvw.cc
bassmang: Changed

python/pylibvw.cc
bassmang: Changed

python/pylibvw.cc
bassmang: The constant namespace is removed at the end of the function, along with any namespaces which are after it

python/vowpalwabbit/pyvw.py
```diff
@@ -883,8 +883,8 @@ def __init__(self, ex, id):
             if len(id) == 0:
                 id = " "
             self.id = None  # we don't know and we don't want to do the linear search required to find it
-            self.ns = id[0]
-            self.ord_ns = ord(self.ns)
+            self.ns = id
```
jackgerrits, line 886: Storing the hashed namespace here will allow it to be reused rather than hashed in every function call.

PR #3111, opened 2021-06-28
refactor: Java migration for namespaced features migration

jackgerrits:


java/src/main/c++/jni_spark_vw.cc
```diff
@@ -384,24 +384,14 @@ JNIEXPORT void JNICALL Java_org_vowpalwabbit_spark_VowpalWabbitExample_clear(JNI
...
     JNIEnv* env, jobject exampleObj, jchar ns, jint weight_index_base, jdoubleArray values)
 {
   INIT_VARS
 
   try
   {
-    addNamespaceIfNotExists(all, ex, ns);
-
-    auto features = ex->feature_space.data() + ns;
+    auto& features = ex->feature_space.get_or_create_feature_group(weight_index_base, ns);
```
rajan-chari, line 394: How does this work?  Just gets the first feature group that matches ns?

java/src/main/java/org/vowpalwabbit/spark/VowpalWabbitExample.java
```diff
@@ -33,19 +33,34 @@
...
+
+        /**
+     * Adds the sparse features values to the supplied namespace.
+     *
+     * @param ns      the first character of the namespace.
+     * @param baseIndex the namespace hash (e.g. hash(namespace))
+     * @param indices the indices of each corresponding feature value.
+     * @param values  the feature values.
+     */
+    public native void addToNamespaceSparse(char ns, int baseIndex, int[] indices, double[] values);
```
rajan-chari, line 63: Wondering if there should be another addToNamespace overload that takes the whole namespace string.

jackgerrits: This is simply enabling the spark java interface to work with this. The spark java interface requires manual hashing of your own namespace and was an intentional decision when it was created for performance reasons.

java/src/main/c++/jni_spark_vw.cc
jackgerrits: To create a namespace you must know both the hash and index for it, so the creation function need the complete info for creation. Duplicates are not allowed, so if the hash already exists you get back that existing feature_group

PR #3120, opened 2021-07-02
fix: Check label type for shared features, Resolves #3088

### LINKED ISSUE #3088: Shared features not working with --csoaa_ldf flag
#### Describe the bug
Using the `--csoaa_ldf m` flag on a dataset with shared features produces buggy output.

#### To Reproduce
1. Used this specific command line: `./vw --csoaa_ldf m ex.txt`
2. Used `ex.txt`:
```
shared | g
1:0 | Real_f Auto_f RealAuto_f
2:1 | Toy_f  Auto_f ToyAuto_f
3:1 | Real_f Air_f  RealAir_f
4:1 | Toy_f  Air_f  ToyAir_f
```

#### Expected behavior
Num weight bits = 18                                                                                                                                                                                                                                                                                                                                      learning rate = 0.5                                                                                                                                                                                                                                                                                                                                       initial_t = 0                                                                                                                                                                                                                                                                                                                                             power_t = 0.5                                                                                                                                                                                                                                                                                                                                             using no cache                                                                                                                                                                                                                                                                                                                                            Reading datafile = ex                                                                                                                                                                                                                                                                                                                                     num sources = 1                                                                                                                                                                                                                                                                                                                                           average  since         example        example  current  current  current                                                                                                                                                                                                                                                                                  loss     last          counter         weight    label  predict features                                                                                                                                                                                                                                                                                  0.000000 0.000000            1            1.0    known        1       16                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            finished run                                                                                                                                                                                                                                                                                                                                              number of examples = 1                                                                                                                                                                                                                                                                                                                                    weighted example sum = 1.000000                                                                                                                                                                                                                                                                                                                           weighted label sum = 0.000000                                                                                                                                                                                                                                                                                                                             average loss = 0.000000                                                                                                                                                                                                                                                                                                                                   total feature number = 16   


#### Observed Behavior
Num weight bits = 18                                                                                                                                                                                                                                                                                                                                      learning rate = 0.5                                                                                                                                                                                                                                                                                                                                       initial_t = 0                                                                                                                                                                                                                                                                                                                                             power_t = 0.5                                                                                                                                                                                                                                                                                                                                             using no cache                                                                                                                                                                                                                                                                                                                                            Reading datafile = ex                                                                                                                                                                                                                                                                                                                                     num sources = 1                                                                                                                                                                                                                                                                                                                                           Enabled reductions: gd, scorer, csoaa_ldf, shared_feature_merger                                                                                                                                                                                                                                                                                          average  since         example        example  current  current  current                                                                                                                                                                                                                                                                                  loss     last          counter         weight    label  predict features                                                                                                                                                                                                                                                                                  [warning] NAN prediction in example 1, forcing 0                                                                                                                                                                                                                                                                                                          [warning] NAN prediction in example 1, forcing 0                                                                                                                                                                                                                                                                                                          [warning] NAN prediction in example 1, forcing 0                                                                                                                                                                                                                                                                                                          [warning] NAN prediction in example 1, forcing 0                                                                                                                                                                                                                                                                                                          -340282346638528859811704183484516925440.000000 -340282346638528859811704183484516925440.000000            1            1.0    known        0       18                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              finished run                                                                                                                                                                                                                                                                                                                                              number of examples = 1                                                                                                                                                                                                                                                                                                                                    weighted example sum = 1.000000                                                                                                                                                                                                                                                                                                                           weighted label sum = 0.000000                                                                                                                                                                                                                                                                                                                             average loss = -340282346638528859811704183484516925440.000000                                                                                                                                                                                                                                                                                            total feature number = 18     

#### Environment
Buggy behavior was in master branch, correct behavior was in old version.
Linux command line environment

bassmang:


vowpalwabbit/shared_feature_merger.cc
```diff
@@ -47,7 +47,18 @@ void predict_or_learn(sfm_data& data, VW::LEARNER::multi_learner& base, multi_ex
 
   multi_ex::value_type shared_example = nullptr;
 
-  const bool has_example_header = CB::ec_is_example_header(*ec_seq[0]);
+  bool has_example_header = false;
+  // TODO: Use label type from base when available -- currently hacky method to find label type
+  if (ec_seq[0]->l.cb.costs.size() > 0) { has_example_header = CB::ec_is_example_header(*ec_seq[0]); }
```
jackgerrits, line 52: A reduction should never read more than 1 label is a certain invocation. The label type should be checked in setup and then only that one checked

jackgerrits (changes requested):

vowpalwabbit/shared_feature_merger.cc
bassmang: Ok I added a static label_type to the shared_feature_merger namespace which is set in setup

jackgerrits: It should not be static, that means we can only ever use a single instance of shared_feature_merger

bassmang: Got it, removed static

vowpalwabbit/learner.cc
```diff
@@ -73,6 +73,10 @@ inline bool example_is_newline_not_header(example& ec, vw& all)
   // If we are using CCB, test against CCB implementation otherwise fallback to previous behavior.
   bool is_header = false;
   if (all.example_parser->lbl_parser.label_type == label_type_t::ccb) { is_header = CCB::ec_is_example_header(ec); }
+  else if (all.example_parser->lbl_parser.label_type == label_type_t::cs)
+  {
+    is_header = COST_SENSITIVE::ec_is_example_header(ec);
+  }
   else
   {
     is_header = CB::ec_is_example_header(ec);
```
lokitoth, line 82: Does it make sense to have this be a function, especially since you duplicate the logic in shared_feature_merger.cc?

There are a few other place we check for headers, too, and I am unsure whether they are doing it correctly in all situations:

* https://github.com/VowpalWabbit/vowpal_wabbit/blob/6335dd190d682bd98a01999a4b5b091ea69e6dae/vowpalwabbit/explore_eval.cc#L66
* https://github.com/VowpalWabbit/vowpal_wabbit/blob/21030f5c97b71f91224bfb44c4381316cf82efeb/vowpalwabbit/search.cc#L1277

(There are other places in search). If it is not clear what the right behaviour is for search, do not let it block this fix, though if you could create an issue tracking checking whether the is_example_header is properly checked in search assigned to me that would be appreciated.

jackgerrits: You would either need to overload it per label_type or pass the label_type as a parameter, but yes a central place to check this would be nice

jackgerrits (approved):

vowpalwabbit/learner.cc
bassmang: I moved the logic to a function in learner.h which takes label_type as an input. Let me know if there's a better place to move this function.

PR #3128, opened 2021-07-06
refactor: deprecate ezexample

jackgerrits:


lalo (approved):

library/ezexample_predict.cc
```diff
@@ -1,8 +1,12 @@
 #include <stdio.h>
 #include "../vowpalwabbit/parser.h"
 #include "../vowpalwabbit/vw.h"
+
+// ezexample is deprecated and will be removed in VW 9.0.
```
lalo, line 5: maybe add a comment that references to the correct example struct cpp/h file

jackgerrits: I can add a link to the issue which details migration

PR #3129, opened 2021-07-06
fix: [Python] Fix default parameter `convert_labels` in `sklearn_vw.VWRegressor`

### LINKED ISSUE #3125: `VWRegressor` should have default parameter `convert_labels` set to False
The class `VWRegressor`  has default parameter `convert_labels=False` since it directly inherits from class `VW`.  It should not be the case in a regression setting, as it will convert all continuous data to `{-1, 1}` (from my understanding of the function `tovw`  [in module sklearn_vw](https://github.com/VowpalWabbit/vowpal_wabbit/blob/b06f7678846a4f6ab44bdf6379286b48d0ab7f2d/python/vowpalwabbit/sklearn_vw.py#L1009)).

Here are reproducible python code that illustrates the problem:

```python
from vowpalwabbit.sklearn_vw import VWRegressor
from sklearn.metrics import mean_squared_error
import numpy as np
import pandas as pd

# Data generation process
np.random.seed(0)
n = 200
intercept = 6
coef = np.array([2.4, 1.2])
x = np.random.normal(size=(n, len(coef)))
noise = np.random.normal(size=n, scale=.25)
y = intercept + (x @ coef) + noise

# Models definition
current_default = VWRegressor()
current_default.fit(y=y, X=x)

better_default = VWRegressor(convert_labels=False)
better_default.fit(y=y, X=x)

# Models predictions
y_pred_current_default = current_default.predict(x)
y_pred_better_default = better_default.predict(x)

# Scores
mse_current = mean_squared_error(y_true=y, y_pred=y_pred_current_default)
mse_better = mean_squared_error(y_true=y, y_pred=y_pred_better_default)

# Plot
ax = pd.DataFrame([y, y_pred_current_default, y_pred_better_default]).T.plot(alpha=.6)
ax.legend(["truth", f"current default (mse: {mse_current:.1f})", f"better default (mse: {mse_better:.1f})"])
```

with the plot:

image

I suspect this is the issue that @adr-arroyo faces in  https://github.com/VowpalWabbit/vowpal_wabbit/issues/3107 (since he tried regression)

etiennekintzler:
Fix #3125 

Parameter `convert_labels` is now set to `True`.

I also had to add a specific test since `TestVWRegressor`  inherits from class `BaseVWTest` that has the following test
```python
def test_repr(self):
    model = self.estimator()
    expected = self.estimator.__name__ + "(convert_labels: True, convert_to_vw: True, passes: 1, quiet: True)"
    assert expected == model.__repr__()
```
which would have failed since `convert_labels` is now `False`.


python/vowpalwabbit/sklearn_vw.py
```diff
@@ -866,6 +866,10 @@ def _more_tags(self):
 class VWRegressor(VW, RegressorMixin):
     """Vowpal Wabbit Regressor model """
 
+    def __init__(self, **kwargs):
+        kwargs["convert_labels"] = False
```
jackgerrits, line 870: Should this only be set if it wasn't explicitly provided by the caller?

jackgerrits (approved):

python/vowpalwabbit/sklearn_vw.py
etiennekintzler: Yes you're right.

PR #3135, opened 2021-07-07
refactor: remove all usages of v_init and deprecate

jackgerrits:


lalo (approved):

vowpalwabbit/v_array.h
```diff
@@ -415,6 +415,7 @@ struct v_array<T, typename std::enable_if<std::is_trivially_copyable<T>::value>:
 };
 
 template <class T>
+VW_DEPRECATED("Use v_array's constructor directly instead. This will be removed in VW 9.0.")
 inline v_array<T> v_init()
```
lalo, line 419: if we were to delete it today with this PR, are all the references in this repo to it gone?

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -188,7 +188,7 @@ void save_action_scores(ccb& data, decision_scores_t& decision_scores)
   data.exclude_list[original_index_of_chosen_action] = true;
 
   decision_scores.emplace_back(std::move(pred));
-  data.shared->pred.a_s = v_init<ACTION_SCORE::action_score>();
+  data.shared->pred.a_s = v_array<ACTION_SCORE::action_score>();
```
peterychang, line 191: would `clear()` work? I think it should

vowpalwabbit/v_array.h
jackgerrits: yes

peterychang (approved):
I would recomment using `clear()` to get moved v_arrays into a known state, but otherwise looks good!

PR #3136, opened 2021-07-07
refactor: deprecate some v_array functions

jackgerrits:
deprecate: `v_array::find_sorted`, `v_array::unique_add_sorted`, `v_array::contain_sorted`, `v_array_contains`

### DIFF
--- a/vowpalwabbit/log_multi.cc
+++ b/vowpalwabbit/log_multi.cc
@@ -175,7 +175,7 @@ bool children(log_multi& b, uint32_t& current, uint32_t& class_index, uint32_t l
   auto& preds = b.nodes[current].preds;
   node_pred val_to_insert(label);
   auto found_it = std::lower_bound(preds.begin(), preds.end(), val_to_insert);
-  if(found_it == preds.end() || !(*found_it == val_to_insert)) { found_it = preds.insert(found_it, val_to_insert); }
+  if (found_it == preds.end() || !(*found_it == val_to_insert)) { found_it = preds.insert(found_it, val_to_insert); }
   class_index = static_cast<uint32_t>(std::distance(preds.begin(), found_it));
   b.nodes[current].preds[class_index].label_count++;
 
--- a/vowpalwabbit/log_multi.cc
+++ b/vowpalwabbit/log_multi.cc
@@ -175,8 +175,8 @@ bool children(log_multi& b, uint32_t& current, uint32_t& class_index, uint32_t l
   auto& preds = b.nodes[current].preds;
   node_pred val_to_insert(label);
   auto found_it = std::lower_bound(preds.begin(), preds.end(), val_to_insert);
-  if(found_it == preds.end() || *found_it != val_to_insert) { preds.insert(found_it, val_to_insert); }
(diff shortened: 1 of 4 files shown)


vowpalwabbit/log_multi.cc
```diff
@@ -172,7 +172,10 @@ void display_tree_dfs(log_multi& b, const node& node, uint32_t depth)
 
 bool children(log_multi& b, uint32_t& current, uint32_t& class_index, uint32_t label)
 {
-  class_index = static_cast<uint32_t>(b.nodes[current].preds.unique_add_sorted(node_pred(label)));
+  auto& preds = b.nodes[current].preds;
```
peterychang, line 175: it looks like `unique_add_sorted` will only add the value if it doesn't already exist in the vector

vowpalwabbit/v_array.h
```diff
@@ -377,12 +378,15 @@ struct v_array<T, typename std::enable_if<std::is_trivially_copyable<T>::value>:
     else  // size = 1, ele = 1, _begin[0] = 0
       return b;
   }
+
+  VW_DEPRECATED("Use insert and std::upper_bound instead. This will be removed in VW 9.0.")
```
peterychang, line 382: to reproduce this behavior, you'd actually have to do a lower_bound and check if the value is equal.. we might need to keep this function around because there's no easy way to reproduce exactly what it does...

jackgerrits: Hmm, you can do an upper_bound then check prev if it is equal to the value to be inserted.

peterychang: lower_bound gives you the first index thats >= the target, which will either give you an iterator to the value you want or the upper_bound if it doesn't exist

vowpalwabbit/log_multi.cc
peterychang: lower_bound gives you the behavior you want without having to do the iteration subtraction
Just make sure you check if the returned iterator is end() before you try to dereference it

```
found_it = lower_bound();
if(found_it != vec.end() && *found_it != val_to_insert) { vec.insert(); }
```

vowpalwabbit/v_array.h
jackgerrits: If you use upper bound and go back one then you only need to traverse the list once compared to if you use lower_bound to determine membership then upper bound to find the iterator to insert at

vowpalwabbit/log_multi.cc
jackgerrits: This requires calling both lower and upper bound though

```
found_it = lower_bound();
if(found_it == vec.end() || *found_it != val_to_insert) { vec.insert(upper_bound); }
```

peterychang: upper_bound gives you strictly > while lower_bound is >=. So if the value doesn't exist, then it gives you exactly the same result as an upper_bound call would give

jackgerrits: Ah I see, that makes sense

peterychang (approved):

PR #3141, opened 2021-07-08
fix: Fix CB label caching

### LINKED ISSUE #3139: CB labels do not get cached correctly
CB labels do not get cached correctly: Their weight does not get cached.

data.txt:
```vw-text
shared | a:1 b:0.5
0:0.1:0.75 | a:0.5 b:1 c:2
| a:1 c:3

shared | s_1 s_2
0:1.0:0.5 | a:1 b:1 c:1
| a:0.5 b:2 c:1
```

When running VW without the cache (or the when generating the cache) the output is:
```
finished run
number of examples = 2
weighted label sum = 0.000000
average loss = 1.066667
total feature number = 21
```
Command line = `--cb_adf -c -k -d data.txt`

When running VW the second time (with cache):
```
finished run
number of examples = 2
weighted label sum = 0.000000
average loss = n.a.
total feature number = 21
```
Command line = `--cb_adf -c -d data.txt`

lokitoth:
The label caching code was not following the io_buf contract expectations: It was not updating the buffer pointer after writing using `.set()`, and it was not properly requesting space for the weight field. This prevented the weight field from being persisted.

The fix is to update the caching code to properly `.set()` when writing using `buf_write()`, and to use `write_value()`/`read_value()` when storing the weight.

Fixes #3139

### DIFF
--- a/test/runtests_parser.py
+++ b/test/runtests_parser.py
@@ -64,13 +64,13 @@ class Test:
             self.vw_command = " ".join(self.vw_command.split()[1:])
 
             # check what input files this command needs
-            input_file_flags = ["-d", "-i", "--dictionary", "--feature_mask"]
+            input_file_flags = ["-d", "-i", "--dictionary", "--feature_mask", "--cache_file"]
             files = []
             for flag in input_file_flags:
                 next_files = Parser.get_values_of_vwarg(self.vw_command, flag)
 
                 # some cleanup only when its dictionary
-                if flag is "--dictionary":
+                if flag == "--dictionary":
                     next_files = [f.split(":")[-1]  for f in next_files]
                     dpath = Parser.get_values_of_vwarg(self.vw_command, "--dictionary_path")
                     if dpath:
@@ -83,6 +83,7 @@ class Test:
 
             # get output files and register as creator of files
             files = Parser.get_values_of_vwarg(self.vw_command, "-f")
+            files += Parser.get_values_of_vwarg(self.vw_command, "--cache_file")
             for f in files:
                 Test.register_output(int(self.id), f)
 
@@ -93,6 +94,13 @@ class Test:
             for f in files:
                 if "model-sets" not in f:
                     depends_on.append(Test.filename_to_test_id(f))
+
+            files = Parser.get_values_of_vwarg(self.vw_command, "--cache_file")
+            for f in files:
+                parent_test = Test.filename_to_test_id(f)
+                if str(parent_test) != self.id:
+                    depends_on.append(parent_test)
+
             if depends_on:
                 self.depends_on = depends_on
         
--- a/test/runtests_parser.py
+++ b/test/runtests_parser.py
@@ -108,10 +108,6 @@ class Test:
 
             files = Parser.get_values_of_vwarg(self.vw_command, "--cache_file")
             for f in files:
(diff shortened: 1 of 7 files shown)


vowpalwabbit/cb_label_parser.h
```diff
@@ -68,15 +63,25 @@ char* bufcache_label(LabelT& ld, char* c)
...
+void cache_label_additional_fields(LabelT& ld, io_buf& cache)
+{
+  cache.write_value(ld.weight);
 }
 
 template <typename LabelT = CB::label, typename LabelElmT = cb_class>
 void cache_label(LabelT& ld, io_buf& cache)
 {
   char* c;
   cache.buf_write(c, sizeof(size_t) + sizeof(LabelElmT) * ld.costs.size());
```
jackgerrits, line 80: `cache.buf_write(c, sizeof(size_t) + sizeof(LabelElmT) * ld.costs.size());` should probably be done internally to the call of `bufcache_label_elements`, since it is overloaded based on type for continuous right?

lokitoth: I believe the point is that it computes the right thing based on that `sizeof()` (which is what the change in the Continuous Action commit did) - since `LabelElmT` is the POD containing just the element data, so it matches the read/write size. The only place where they differ is that the continuous label does not have additional fields (weight).

jackgerrits: That's not obvious at all but I see what you mean. I don't see any benefit of doing it that way over just delegating responsibility of requesting enough buffer space to the `bufcache_label_elements` call itself.

lokitoth: It seems like we should pull `bufcache_label_elements` back into `cache_label`, since it never gets overloaded anyways.

lokitoth:
The Linux failure on Azure DevOps is funny: It is bad interference between the breaking out of tests individually (which runs RunTests individually for each test) and RunTests trying to clean up `*/*.cache` when it starts.

lokitoth:
Just talked to @lalo - turns out there the runtests_parser.py file needs to be updated for this new type of inter-test dependency (--cache_file), which also reminds me that I need to make the same change for the C# Unit Tests.

test/runtests_parser.py
```diff
@@ -93,6 +105,19 @@ def clean(self):
             for f in files:
                 if "model-sets" not in f:
                     depends_on.append(Test.filename_to_test_id(f))
+
+            files = Parser.get_values_of_vwarg(self.vw_command, "--cache_file")
+            for f in files:
+                # If the file has not been registered as an output
+                if not Test.has_output(f):
+                    continue
```
lalo, line 113: this if should never be executed bc of line 99. /I think/

lokitoth: Good catch. Removed.

vowpalwabbit/cb_label_parser.h
lokitoth: I pulled in the element cache/read methods into the main label cache/read methods, and replaced most of the explicit io_buf buffer management with `[read/write]_value`, but there is an odd lack-symmetry around label caching forcing the read to be able to fail transparently when at EOF.

jackgerrits (approved):

PR #3142, opened 2021-07-09
test: get type information of VW options

lalo:


python/pylibvw.cc
```diff
@@ -108,14 +108,16 @@ class OptionManager
   template <typename T>
   py::object* value_to_pyobject(VW::config::typed_option<T>& opt)
   {
+    T not_supplied;
```
jackgerrits, line 111: This is unitialized
```suggestion
    T not_supplied{};
```

jackgerrits (approved):

PR #3143, opened 2021-07-09
fix: Multiline Example handling when in multi-pass mode

### LINKED ISSUE #3106: Can't do multiple passes with --cb_adf format
I am attempting to do multiple passes over a cb_adf dataset to hopefully improve the quality of my q function. I'm thinking of trying an offline bandit using the whole dataset and multiple passes instead of online with iterative updates. However, I get the following error after the first pass:
```
libc++abi.dylib: terminating with uncaught exception of type VW::vw_exception: cb_adf: badly formatted example, only one line can have a cost
[1]    90720 abort      vw --cb_adf --passes 2 -c -d train.dat
```
Here is my command and dataset for reproducibility:
`vw --cb_adf --passes 2 -c -d train.dat`

train.dat
```
shared | a:1 b:0.5
0:0.1:0.75 | a:0.5 b:1 c:2
| a:1 c:3

shared | s_1 s_2
0:1.0:0.5 | a:1 b:1 c:1
| a:0.5 b:2 c:1
```
I'm using version 8.10.1 on OSX. I found this [SO post](https://stackoverflow.com/questions/59466988/vowpal-wabbit-contextual-bandit-simulation-error) and commit https://github.com/VowpalWabbit/vowpal_wabbit/commit/431c270742e76c7524413c33e76e47851cba549e by @jackgerrits that maybe was supposed to fix this but also could be unrelated.

Are multiple passes not supported for --cb_adf? If so, maybe some better error messaging might be useful here?

lokitoth:
The logic for assembling and dispatching multiline examples was not handling end-of-pass examples correctly: When they are empty, they did not properly complete the `multi_ex` object and dispatch it to `learn_multi_ex`. Instead, it silently finished the end-of-pass example. Then when examples came in from a previous pass, they get appended to the inflight multi_ex, resulting in a malformed example in many cases (e.g. for cb_adf, this results in an example with multiple labeled actions)

Fixes #3106


vowpalwabbit/learner.cc
```diff
@@ -45,12 +45,13 @@ void learn_multi_ex(multi_ex& ec_seq, vw& all)
   as_multiline(all.l)->finish_example(all, ec_seq);
 }
 
+template <bool finish_example=true>
```
lokitoth, line 48: Can be removed - the example cannot be finished early because that erases `ec->end_pass` information.

jackgerrits (approved):

test/RunTests
```diff
@@ -2473,4 +2473,8 @@ train-sets/ref/slates_simple_w_interactions.stderr
 {VW} --cb_adf --cache_file models/cb_test_cache.ldf.cache -d train-sets/cb_test.ldf
     train-sets/ref/cb_adf_mtr_cache_2.stderr
 
+# Test 344: check cb_adf with multiple passes
+{VW} --cb_adf -c --passes 2 -d train-sets/cb_adf_multipass_crash_1.data --holdout_off
```
lalo, line 2477: can you add a --cb test with --passes 2 too?

lalo: --cb will reroute to cb_adf

lokitoth: I do not think this code change will be hit by it? Since it is a single-line top learner. It generates the multi-ex and drives the base learner internally:

* Create multi_ex via [cbify_adf_data::copy_example_to_adf()](https://github.com/VowpalWabbit/vowpal_wabbit/blob/f2bd8e5e7f015e9fb852afb38bb084aaa3f9ae5b/vowpalwabbit/cbify.cc#L143)
* Invoke base learner [directly](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/cb_to_cb_adf.cc#L80)

Happy to add the test anyways.

lalo: that changed with cb_to_cb_adf, check reduction stack of this test as an example: https://github.com/VowpalWabbit/vowpal_wabbit/blob/473f06263b874af1a2ff6103daf97765e29dec53/test/RunTests#L2461

lalo (approved):

PR #3146, opened 2021-07-12
fix: Adds random seed when saving VW model, Resolves #3116

### LINKED ISSUE #3116: CATS does not save the model
[003-Cats-Tutorial.ipynb.zip](https://github.com/VowpalWabbit/vowpal_wabbit/files/6748242/003-Cats-Tutorial.ipynb.zip)
I train a multi-armed bandit model for a continuous action space.
After training, I save the model, but after unloading the model, the progress is not saved or the model does not predict on old examples.

`
vw.predict(ex)
`
Result:
`
(13.885909080505371, 0.40625)
`

```
vw.save("mab.model")

model = pyvw.vw("mab.model")

model.predict(ex)
```
Result:
`
0.0
`

If I use loading with `--quiet -i`, then a bug comes when predicting

```
RuntimeError: Continuous actions labels require the first word to be ca

```

#### Environment
VW 8.10.2 on MAC OS with the Python bindings

bassmang:


test/RunTests
```diff
@@ -2485,4 +2485,14 @@ train-sets/ref/slates_simple_w_interactions.stderr
 {VW} -d train-sets/cb_multi_cost.dat --cb 4 -k -c --passes 2 --holdout_off --cb_force_legacy
     train-sets/ref/cb_multi_cost_multipass_legacy.stderr
```
olgavrou, line 2487: Windows CI doesn't run json/dsjson tests, you can make the windows ci skip them [here](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/cs/unittest/RunTests.tt#L24)

vowpalwabbit/vw_versions.h
```diff
@@ -22,4 +22,5 @@
 #define VERSION_FILE_WITH_CB_TO_CBADF "8.11.0"  // first version that maps --cb to use cb_adf
 #define VERSION_FILE_WITH_REG_CB_SAVE_RESUME "8.11.0"     // version that ensures accuracy of loaded models in regcb
 #define VERSION_FILE_WITH_SQUARE_CB_SAVE_RESUME "8.11.0"  // version that ensures accuracy of loaded models in squarecb
-#define VERSION_FILE_WITH_FIRST_SAVE_RESUME "8.11.0"      // version that ensures accuracy of loaded models in first
\ No newline at end of file
+#define VERSION_FILE_WITH_FIRST_SAVE_RESUME "8.11.0"      // version that ensures accuracy of loaded models in first
+#define VERSION_FILE_WITH_RANDOM_SEED "8.11.1"            // version which saves random seed from vw in model file
```
olgavrou, line 26: we are about to release 8.11.0 so I think you can just re-use that version here, no need to use 8.11.1 as far as I can tell

vowpalwabbit/parse_regressor.cc
```diff
@@ -316,6 +316,12 @@ void save_load_header(
         }
       }
 
+      if (!read || all.model_file_ver >= VERSION_FILE_WITH_RANDOM_SEED)
```
olgavrou, line 319: I like that this is going into the general save_load header but I am concerned that this might break forward compatibility on models and so might be safer to put this into the cats reduction where the seed is being used

test/RunTests
bassmang: Thanks!

vowpalwabbit/parse_regressor.cc
bassmang: I had this fix in a lower reduction initially to solve the bug, however the random seed is used in other places like ccb, so the thinking here was that including it in the VW model would avoid other current / future bugs

vowpalwabbit/vw_versions.h
bassmang: I used 8.11.1 to ensure that model files before this could still be read properly, without the random seed. @rajan-chari do you have thoughts on this?

olgavrou: but we haven't released 8.11.0 yet we can't jump to 8.11.1

PR #3150, opened 2021-07-13
fix: Use random state in sample_pdf, Resolves #3116 

### LINKED ISSUE #3116: CATS does not save the model

bassmang:


test/unit_test/explore_test.cc
```diff
@@ -97,13 +97,13 @@ BOOST_AUTO_TEST_CASE(sample_continuous_action_statistical)
   float pdf_value;
   constexpr float range_min = .0f;
   constexpr float range_max = 100.0f;
-  uint64_t random_seed = 7791;
+  std::shared_ptr<rand_state> random_seed = std::make_shared<rand_state>(7791);
```
jackgerrits, line 100:
```suggestion
  auto random_seed = std::make_shared<rand_state>(7791);
```

explore/explore_internal.h
```diff
@@ -340,8 +341,8 @@ namespace exploration
   // 2) Does not normalize the scores (unlike sample_after_normalization)
   // 3) Scores need not add up to one.
   template <typename It>
-  int sample_scores(
-      uint64_t* p_seed, It scores_first, It scores_last, uint32_t& chosen_index, std::random_access_iterator_tag)
+  int sample_scores(std::shared_ptr<rand_state> p_seed, It scores_first, It scores_last, uint32_t& chosen_index,
```
jackgerrits, line 344: This function is part of the explore library which is independent, and therefore cannot depend on `rand_state`. The inner random state should be retrieved before this is called, then it should be updated with the updated state once `sample_scores` is complete.

vowpalwabbit/sample_pdf.cc
```diff
@@ -34,10 +34,10 @@ struct sample_pdf
   int learn(example& ec, experimental::api_status* status);
   int predict(example& ec, experimental::api_status* status);
 
-  void init(single_learner* p_base, uint64_t* p_random_seed);
+  void init(single_learner* p_base, std::shared_ptr<rand_state>&& random_state);
```
jackgerrits, line 37: Why make this an rvalue? This should be a value which is moved inside the call

vowpalwabbit/sample_pdf.cc
```diff
@@ -71,10 +71,10 @@ int sample_pdf::predict(example& ec, experimental::api_status*)
   return VW::experimental::error_code::success;
 }
 
-void sample_pdf::init(single_learner* p_base, uint64_t* p_random_seed)
+void sample_pdf::init(single_learner* p_base, std::shared_ptr<rand_state>&& random_state)
```
jackgerrits, line 74:
```suggestion
void sample_pdf::init(single_learner* p_base, std::shared_ptr<rand_state> random_state)
```

vowpalwabbit/sample_pdf.cc
```diff
@@ -71,10 +71,10 @@ int sample_pdf::predict(example& ec, experimental::api_status*)
...
 {
   _base = p_base;
-  _p_random_state = p_random_seed;
+  _p_random_state = random_state;
```
jackgerrits, line 77:
```suggestion
  _p_random_state = std::move(random_state);
```

jackgerrits (changes requested):

vowpalwabbit/sample_pdf.cc
olgavrou: Do we want to move the shared pointer here?

jackgerrits: If the argument is by value then moving it here is what we want. If the arg is by ref then letting it copy here is good.

olgavrou: yeah was wondering if it's safe but looks like all.`get_random_state()` returns a copy of the shared pointer, so passing it as an rvalue and moving it seems fine (and probably a tad faster)

jackgerrits: Either way works

test/unit_test/explore_test.cc
bassmang: I reverted explore_internal.h so seed can be a uint64_t here again

explore/explore_internal.h
bassmang: Changed to get seed in sample_pdf before calling this

jackgerrits (approved):

olgavrou (approved):

PR #3153, opened 2021-07-13
refactor: move to 2d list based grouping of namespaces

jackgerrits:


vowpalwabbit/chained_proxy_iterator.h
```diff
@@ -52,39 +52,53 @@ struct chained_proxy_iterator
...
+  // UB if diff < 0
   chained_proxy_iterator& operator+=(difference_type diff)
   {
-    for (size_t i = 0; i < diff; i++) { operator++(); }
-    return *this;
+    assert(diff >= 0);
+    while (true)
+    {
+      auto current_group_distance_to_end = std::distance(_current, (*_outer_current).feats.audit_end());
+      if (diff > current_group_distance_to_end)
```
peterychang, line 70: what happens if you try to increment beyond the end of the list? say, `iter += INT_MAX`? It looks like an infinite loop or undefined behavior. Did i miss something?

vowpalwabbit/namespaced_feature_store.cc
```diff
@@ -7,344 +7,89 @@
...
-  // If any groups are left empty, remove them.
-  for (auto it = _legacy_indices_to_index_mapping.begin(); it != _legacy_indices_to_index_mapping.end();)
-  {
-    if (it->second.empty())
+    if (!_saved_feature_group_nodes.empty())
     {
-      // There are no more feature groups which correspond to this index.
-      _contained_indices.erase(it->first);
-      it = _legacy_indices_to_index_mapping.erase(it);
+      _feature_groups[ns_index].splice(
```
peterychang, line 19: only the first element is being moved?

vowpalwabbit/chained_proxy_iterator.h
jackgerrits: If you += beyond the end of the list it will be undefined behavior, same as with any STL iterator.

> The behavior is undefined if the specified sequence of increments or decrements would require that a non-incrementable iterator (such as the past-the-end iterator) is incremented, or that a non-decrementable iterator (such as the front iterator or the singular iterator) is decremented. 

https://en.cppreference.com/w/cpp/iterator/advance

vowpalwabbit/namespaced_feature_store.cc
jackgerrits: Yep it is taking a single node and moving it over to be used

vowpalwabbit/namespaced_feature_store.h
```diff
@@ -4,253 +4,217 @@
...
-#include <set>
+#include <list>
+#include <array>
+#include <vector>
 
 #include "feature_group.h"
-#include "generic_range.h"
 #include "chained_proxy_iterator.h"
 
 typedef unsigned char namespace_index;
```
olgavrou, line 15: sort-of unrelated to PR but can this be a `using namespace_index` instead of `typedef`?

vowpalwabbit/namespaced_feature_store.h
```diff
@@ -4,253 +4,217 @@
...
+  namespaced_feature_store& operator=(const namespaced_feature_store& other)
+  {
+    _feature_groups = other._feature_groups;
+    _legacy_indices_existing = other._legacy_indices_existing;
+    return *this;
+  }
+  namespaced_feature_store(namespaced_feature_store&& other) = default;
+  namespaced_feature_store& operator=(namespaced_feature_store&& other) = default;
+
+  inline size_t size() const
```
olgavrou, line 103: depending on how often we call `size()` is it worth re-calculating this or keeping the size ready every time something is added/removed?

vowpalwabbit/namespaced_feature_store.h
```diff
@@ -4,253 +4,217 @@
...
+  {
+    size_t accumulator = 0;
+    for (auto ns_index : _legacy_indices_existing) { accumulator += _feature_groups[ns_index].size(); }
+    return accumulator;
+  }
+  inline bool empty() const { return _legacy_indices_existing.empty(); }
```
olgavrou, line 109: nit: spaces in between methods

vowpalwabbit/namespaced_feature_store.h
```diff
@@ -4,253 +4,217 @@
...
+    return &it->feats;
+  }
 
-  // The following are experimental and may be superseded with namespace_index_begin_proxy
-  // Returns empty range if not found
-  std::pair<indexed_iterator, indexed_iterator> get_namespace_index_groups(namespace_index ns_index);
-  // Returns empty range if not found
-  std::pair<const_indexed_iterator, const_indexed_iterator> get_namespace_index_groups(namespace_index ns_index) const;
+  inline features& get(namespace_index ns_index, uint64_t hash)
+  {
```
olgavrou, line 132: so we should only call `get` if we are 100% sure that the ns/hash has features, otherwise we can still call `get_or_null`

vowpalwabbit/namespaced_feature_store.h
```diff
@@ -4,253 +4,217 @@
...
-  features& get_or_create_feature_group(uint64_t hash, namespace_index ns_index);
+  features& get_or_create(namespace_index ns_index, uint64_t hash);
 
-#ifndef VW_NOEXCEPT
-  // These will throw if the hash does not exist
-  const features& operator[](uint64_t hash) const;
-  features& operator[](uint64_t hash);
-#endif
+  // UB if the list itself is modified.
+  inline std::list<namespaced_features>& get_list(namespace_index index) { return _feature_groups[index]; }
```
olgavrou, line 150: can the list ever be modified by a different thread while we are holding a reference to it from here?

vowpalwabbit/namespaced_feature_store.h
jackgerrits: Yeah this is inconsistent, all new places i use `using`. I agree with you

vowpalwabbit/namespaced_feature_store.h
jackgerrits: Yeah I wondered this, I don't thing its often and I wanted to avoid that state for now at least

vowpalwabbit/namespaced_feature_store.h
jackgerrits: Correct

vowpalwabbit/namespaced_feature_store.h
jackgerrits: No it shouldnt be because once parsing is done and it is passed to the learner that parse thread will never touch it again

olgavrou: ok so this is more of a note to the developer

jackgerrits: Yep

lalo (approved):

vowpalwabbit/namespaced_feature_store.h
```diff
@@ -4,253 +4,218 @@
...
-  HashT hash() { return _namespace_hashes[_index]; }
-
-  bool operator==(const iterator_t& rhs) { return _index == rhs._index; }
-  bool operator!=(const iterator_t& rhs) { return _index != rhs._index; }
-};
-
-/// Insertion or removal will result in this value in invalidated.
-template <typename IndicesT, typename FeaturesT, typename IndexT, typename HashT>
+template <typename ArrayT, typename IndexItT, typename ListT>
 class indexed_iterator_t
```
olgavrou, line 20: can we add a comment on top of `indexed_iterator_t` explaining what it does? I understand that it returns whatever the first argument is pointing to each time based on an index that is incremented

vowpalwabbit/chained_proxy_iterator.h
```diff
@@ -52,39 +52,53 @@ struct chained_proxy_iterator
...
   }
 
-  // TODO jump full feature groups.
+  // UB if diff < 0
   chained_proxy_iterator& operator+=(difference_type diff)
   {
-    for (size_t i = 0; i < diff; i++) { operator++(); }
-    return *this;
+    assert(diff >= 0);
+    while (true)
```
olgavrou, line 67: I am afraid of `while (true)` loops since some one at some point might forget to add an exit somewhere. Is there maybe a way of limiting this to something specific?

jackgerrits: While true is a bit scary, I'm not sure what I should change this to though and still feel confident in the condition

olgavrou: what about ! the first if statement?

jackgerrits: Resolved

olgavrou (approved):

PR #3157, opened 2021-07-14
refactor: add label and use builder to set learner features in sample_pdf.cc

bassmang:


vowpalwabbit/rand_state.h
```diff
@@ -10,7 +10,7 @@
 struct rand_state
 {
 private:
-  uint64_t random_state;
+  uint64_t random_state = 0;
```
jackgerrits, line 13: Not needed as line 16 initializes this

vowpalwabbit/sample_pdf.cc
```diff
@@ -112,13 +112,16 @@ LEARNER::base_learner* sample_pdf_setup(options_i& options, vw& all)
...
-  auto p_reduction = scoped_calloc_or_throw<sample_pdf>();
+  auto p_reduction = VW::make_unique<sample_pdf>();
   p_reduction->init(as_singleline(p_base), all.get_random_state());
 
-  LEARNER::learner<sample_pdf, example>& l = init_learner(p_reduction, as_singleline(p_base), predict_or_learn<true>,
-      predict_or_learn<false>, 1, prediction_type_t::action_pdf_value, all.get_setupfn_name(sample_pdf_setup));
+  auto* l = make_reduction_learner(std::move(p_reduction), as_singleline(p_base), predict_or_learn<true>,
+      predict_or_learn<false>, all.get_setupfn_name(sample_pdf_setup))
+                .set_prediction_type(prediction_type_t::action_pdf_value)
+                .set_label_type(label_type_t::simple)
```
jackgerrits, line 121: This learner does not care what the label type is, and should therefore assume whatever bases label type is. It would be better if we could mark it as explicitly pass through but not sure the best way to do that.

For now the best way to communicate that base's label type should be used is actually to omit `set_label_type`. Can you add a comment in place of this line saying it assumes base's label type?

bassmang: Just added in a comment

jackgerrits (approved):

PR #3160, opened 2021-07-15
docs: Move examples and tutorials into docs dir

jackgerrits:


python/docs/source/examples/contextual_bandit.ipynb
```diff
@@ -284,9 +250,12 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.4"
+   "version": "3.8.5"
```
olgavrou, line 253: Should the interpreter hash (and the display_name above) be removed/made more generic?

jackgerrits: I'm pretty sure this is just the version of the kernel which executed the notebook? It isn't a version requirement AFAIK

PR #3162, opened 2021-07-15
refactor: CCB warnings and lint fixes

jackgerrits:


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -36,11 +36,11 @@ using namespace VW::config;
 namespace logger = VW::io::logger;
 
 template <typename T>
-void return_v_array(v_array<T>&& array, VW::v_array_pool<T>& pool)
+void return_v_array(v_array<T>& array, VW::v_array_pool<T>& pool)
 {
   array.clear();
   pool.reclaim_object(std::move(array));
-  array.clear();
+  array = v_array<T>{};
```
jackgerrits, line 43: This fixes the use after move which although may be valid in this case, most linters are going to warn loudly for it

lalo (approved):

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -130,7 +130,7 @@ bool split_multi_example_and_stash_labels(const multi_ex& examples, ccb& data)
       case example_type::slot:
         data.slots.push_back(ex);
         break;
-      default:
+      case unset:
```
lalo, line 133: what is unset

jackgerrits: It is the other enum value

lalo: we dont need a default case anymore?

jackgerrits: Yeah now that I look at this again it seems perhaps the `default` was better and less likely to be misused in future

lalo (approved):

PR #3166, opened 2021-07-15
refactor!: Remove deprecated things in v_array

jackgerrits:


vowpalwabbit/v_array.h
```diff
@@ -352,136 +323,8 @@ struct v_array<T, typename std::enable_if<std::is_trivially_copyable<T>::value>:
     if (_end == _end_array) reserve_nocheck(2 * capacity() + 3);
     new (_end++) T(std::forward<Args>(args)...);
   }
-
-  VW_DEPRECATED("Use std::lower_bound instead. This will be removed in VW 9.0.")
```
peterychang, line 356: didn't we just deprecate this stuff a few weeks ago?

vowpalwabbit/v_array.h
```diff
@@ -172,40 +172,14 @@ struct v_array<T, typename std::enable_if<std::is_trivially_copyable<T>::value>:
...
   inline size_t capacity() const { return _end_array - _begin; }
 
-  // maintain the original (deprecated) interface for compatibility. To be removed in VW 10
-  VW_DEPRECATED(
-      "v_array::resize() is deprecated. Use reserve() instead. For standard resize behavior, use "
-      "resize_but_with_stl_behavior(). This will be removed in VW 9.0.")
-  void resize(size_t length) { reserve_nocheck(length); }
-
   // change the number of elements in the vector
-  // to be renamed to resize() in VW 10
```
peterychang, line 208: we're just changing the version we intended to do this on?

jackgerrits: As can be seen in the above removal the deprecation attribute and comment were out of sync. The deprecation attribute take precedence. Given that, this is updated to be in like with the rest of the changes in this file.

vowpalwabbit/v_array.h
jackgerrits: Yes, prior to the 8.11 release. That is the last release this function will be in.

peterychang (approved):

PR #3175, opened 2021-07-19
refactor: test/benchmark vector instead of list

jackgerrits:


olgavrou:
could a map also work instead of a linked list? so you could have an array of maps

jackgerrits:
Given the low number of elements stored in that collection I think the overhead of a map is going to make it slower. At least thats what I noticed in past tests of a slightly different setup

PR #3178, opened 2021-07-20
refactor: Namespace (feature group) lookup during parsing should happen once per feature group

rajan-chari:
This is in preparation for vw namespace change.

1) Feature group lookups are about to get more expensive.  In existing parser, this happens once per feature instead of once per feature group.  Adjusting this to once per feature group.
2) Also moved feature group related functionality to a helper class.


vowpalwabbit/parse_example.cc
```diff
@@ -50,29 +50,113 @@ int read_features_string(vw* all, v_array<example*>& examples)
...
+      , _current_example(current_example)
+      , _redefine_char_flag(redefine_char_flag)
+      , _redefine_char_map(redefine_char_map)
+      , _namespace_dictionaries_map(namespace_dictionaries_map)
+      , _affix_features(affix_features)
+      , _spelling_features(spelling_features)
+      , _hash_function(hash_fn)
+  {
+  }
+  inline unsigned char feature_group_char() const { return _feature_group_char; }
```
olgavrou, line 81: aren't member function inherently inline in C++?

vowpalwabbit/parse_example.cc
```diff
@@ -50,29 +50,113 @@ int read_features_string(vw* all, v_array<example*>& examples)
...
+    _hash = _hash_function(_feature_group_name.begin(), _feature_group_name.length(), _hash_seed);
+    _p_current_feature_group = nullptr;  // reset cached feature group
+    _affix = (*_affix_features)[_feature_group_char];
+    _spelling = (*_spelling_features)[_feature_group_char];
+    _namespace_dictionaries = &(*_namespace_dictionaries_map)[feature_group_char()];
+  }
+  inline void set_current_fg()
+  {
+    _feature_group_name = " ";
+    _feature_group_char = ' ';
```
olgavrou, line 102: can we use `default_namespace` here instead of `" "` and `' '`?

vowpalwabbit/parse_example.cc
```diff
@@ -50,29 +50,113 @@ int read_features_string(vw* all, v_array<example*>& examples)
...
+    _hash = this->_hash_seed == 0 ? 0 : uniform_hash("", 0, this->_hash_seed);
+    _p_current_feature_group = nullptr;  // reset cached feature group
+    _affix = (*_affix_features)[_feature_group_char];
+    _spelling = (*_spelling_features)[_feature_group_char];
+    _namespace_dictionaries = &(*_namespace_dictionaries_map)[feature_group_char()];
+  }
+  inline const VW::string_view& get_fg_name() const { return _feature_group_name; }
```
jackgerrits, line 109: Return the string view by value

vowpalwabbit/parse_example.cc
```diff
@@ -168,165 +252,165 @@ class TC_parser
...
-      // Case where string:float
-      else if (!feature_name.empty())
-      {
-        word_hash = (_p->hasher(feature_name.begin(), feature_name.length(), _channel_hash) & _parse_mask);
-      }
-      // Case where :float
       else
       {
-        word_hash = _channel_hash + _anon++;
+        fs.space_names.emplace_back(audit_strings(_feature_group_helper.get_fg_name(), feature_name.to_string()));
```
jackgerrits, line 312:
```suggestion
        fs.space_names.emplace_back(_feature_group_helper.get_fg_name(), feature_name.to_string());
```

vowpalwabbit/parse_example.cc
```diff
@@ -374,12 +458,8 @@ class TC_parser
...
     {
       // NameSpaceInfo --> 'String' NameSpaceInfoValue
-      _index = (unsigned char)(_line[_read_idx]);
-      if (_redefine_some) _index = (*_redefine)[_index];  // redefine _index
-      if (_ae->feature_space[_index].size() == 0) _new_index = true;
-      VW::string_view name = read_name();
-      if (audit) { _base = name; }
-      _channel_hash = _p->hasher(name.begin(), name.length(), this->_hash_seed);
+      const VW::string_view name = read_name();
+      _feature_group_helper.set_current_fg(name);
```
jackgerrits, line 462:
```suggestion
      _feature_group_helper.set_current_fg(read_name());
```

vowpalwabbit/parse_example.cc
```diff
@@ -168,165 +252,165 @@ class TC_parser
...
-          affix >>= 4;
+          v_array<char> affix_v;
+          if (_feature_group_helper.feature_group_char() != ' ')
+            affix_v.push_back(_feature_group_helper.feature_group_char());
+          affix_v.push_back(is_prefix ? '+' : '-');
+          affix_v.push_back('0' + static_cast<char>(len));
+          affix_v.push_back('=');
+          affix_v.insert(affix_v.end(), affix_name.begin(), affix_name.end());
+          affix_v.push_back('\0');
+          affix_fs.space_names.emplace_back(audit_strings("affix", affix_v.begin()));
```
jackgerrits, line 347:
```suggestion
          affix_fs.space_names.emplace_back("affix", affix_v.begin());
```

vowpalwabbit/parse_example.cc
```diff
@@ -168,165 +252,165 @@ class TC_parser
...
+    {
       if (!string_feature_value.empty())
       {
-        // chain hash is hash(feature_value, hash(feature_name, namespace_hash)) & parse_mask
-        word_hash = (_p->hasher(string_feature_value.begin(), string_feature_value.length(),
-                         _p->hasher(feature_name.begin(), feature_name.length(), _channel_hash)) &
-            _parse_mask);
+        std::stringstream ss;
+        ss << feature_name << "^" << string_feature_value;
+        fs.space_names.emplace_back(audit_strings(_feature_group_helper.get_fg_name(), ss.str()));
```
jackgerrits, line 308:
```suggestion
        fs.space_names.emplace_back(_feature_group_helper.get_fg_name(), ss.str());
```

vowpalwabbit/parse_example.cc
```diff
@@ -50,29 +50,113 @@ int read_features_string(vw* all, v_array<example*>& examples)
...
+private:
+  inline features& get_feature_group(features*& p_features_ref, unsigned char namespace_char)
+  {
+    if (p_features_ref == nullptr)
+    {
+      p_features_ref = &_current_example->feature_space[namespace_char];
+
+      if (p_features_ref->empty()) _current_example->indices.emplace_back(namespace_char);
+    }
+    return {*p_features_ref};
```
jackgerrits, line 124: The brackets here are going to cause a temp copy of features to be returned, should just be the ref itself no?
```suggestion
    return *p_features_ref;
```

vowpalwabbit/parse_example.cc
rajan-chari: makes sense

vowpalwabbit/parse_example.cc
rajan-chari: Yep.

rajan-chari: Probably will get optimized out.  Still no reason to have it.

jackgerrits: As it is right now it is returning a reference to a temporary object which is undefined behavior.

vowpalwabbit/parse_example.cc
rajan-chari: I see member functions that have function definitions within class definitions are implicitly inlined.  Ok.  Will change this.

rajan-chari: done.

vowpalwabbit/parse_example.cc
rajan-chari: Is there a string version of default_namespace?

vowpalwabbit/parse_example.cc
```diff
@@ -50,29 +50,113 @@ int read_features_string(vw* all, v_array<example*>& examples)
...
+      , _hash_function(hash_fn)
+  {
+  }
+  inline unsigned char feature_group_char() const { return _feature_group_char; }
+  inline uint64_t feature_group_hash() const { return _hash; }
+  inline features& get_current_fg() { return get_feature_group(_p_current_feature_group, feature_group_char()); }
+  inline features& get_affix_fg() { return get_feature_group(_p_affix_feature_group, affix_namespace); }
+  inline features& get_spelling_fg() { return get_feature_group(_p_spelling_feature_group, spelling_namespace); }
+  inline features& get_dictionary_fg() { return get_feature_group(_p_dictionary_feature_group, dictionary_namespace); }
+  inline void set_current_fg(const VW::string_view& fg_name)
```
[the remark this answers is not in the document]
rajan-chari: done.

vowpalwabbit/parse_example.cc
rajan-chari: done

vowpalwabbit/parse_example.cc
```diff
@@ -50,29 +50,114 @@ int read_features_string(vw* all, v_array<example*>& examples)
...
+  VW::string_view _feature_group_name;
+  uint64_t _affix;
+  bool _spelling;
+  namespace_dictionaries* _namespace_dictionaries;
+  char_map* _redefine_char_map;
+  dictionaries_map* _namespace_dictionaries_map;
+  affix_map* _affix_features;
+  bool_map* _spelling_features;
+  hash_func_t _hash_function;
+  const char* _default_namespace_str = " ";
```
jackgerrits, line 145: Why make this is a member? Seems reasonable to make it a const global?

rajan-chari: yeah makes sense.

test/benchmarks/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ set(all_sources
   benchmark_main.cc
   standalone/benchmark_text_input.cc
   standalone/rcv1_benchmarks.cc
+  standalone/string_view_benchmarks.cc
```
jackgerrits, line 5: Though this is an interesting thing to see, this does not seem relevant to VW's performance. Adding this is going to make the benchmarks CI slower, which is already one of the slower runs.

jackgerrits (changes requested):

test/benchmarks/CMakeLists.txt
rajan-chari: I ran the benchmarks to see how long they take.  It looks like this adds less than a second to the time to run the process.  However, I think the more important question you ask is what kind of benchmarks should we have?  Perhaps we can discuss this more during a VW session?

jackgerrits: We use the benchmarks CI to make a determination about the performance implications of a PR. From my perspective benchmarking `std::string_view` by ref or value is irrelevant to the goal of that CI and just introduces noise.

rajan-chari: Let's find a good place for it then.  We should be micro benchmarking small code blocks to help guide dev choices.

jackgerrits: The `std::string_view` benchmark was helpful in answering the question of value vs ref. Now that that is answered I do not see the value of persisting this benchmark. Ephemeral code to measure things is important, but persisting all of this code we write is not necessarily the right thing to do.

Benchmarks for components we develop and maintain are certainly helpful as they allow us to track performance over time. Benchmarks for components we do not develop are helpful to answer "this or that" questions but are not useful to track over time as they are beyond our control.

I'm all for micro-benchmarking of small code blocks we develop.

rajan-chari: There are a few very different points raised.  Let me summarize and address them here.

1) This microbenchmark answers the question of ref vs not.  Once there is an answer, we don't need to keep the benchmark around.
a) On the contrary, we should preserve these benchmarks that help make decisions so we don't have to revisit this question.  It makes perf knowledge institutional vs tribal.  

2) Benchmarks CI is used to determine perf implications of a PR.  This benchmark will not help with that goal.
a) VW::string_view is an alias for boost::string_view or boost::string_ref or could be something else in the future.  That might be an argument for leaving this where it is.

3) Benchmark CI takes a long time.  We should not be slowing it down further.
a) I measured the impact of this benchmark relative to the other ones.  It takes approx. 1 second. 

4) Persisting all of this code we write is not necessarily the right thing to do.
a) Yes.  I agree.

olgavrou: If we want to document this decision the [coding guidelines](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Coding-guidelines) is the way to go here

I love that we benchmark these things but this is likely to be a useful benchmark when we (a) remove boost (b) update c++ version or (c) update cmake 

The time to run this might only be one second but developers will have the extra cognitive load of either learning to ignore this benchmark (never good to get into the mental state of ignoring things) in the CI run and/or opening the file to see what this benchmark does and why when all other benchmarks have gone up/down this one has stayed the same

If we want microbenchmarks of small code blocks then (in a separate PR from this one) we can create a different CI run or benchmark category that clearly states what these benchmarks are doing

vowpalwabbit/parse_example.cc
```diff
@@ -50,29 +50,113 @@ int read_features_string(vw* all, v_array<example*>& examples)
...
+  {
+    _feature_group_name = fg_name;
+    auto namespace_char = static_cast<unsigned char>(_feature_group_name[0]);
+    if (_redefine_char_flag) namespace_char = (*_redefine_char_map)[namespace_char];  // redefine feature_group_char
+    _feature_group_char = namespace_char;
+    _hash = _hash_function(_feature_group_name.begin(), _feature_group_name.length(), _hash_seed);
+    _p_current_feature_group = nullptr;  // reset cached feature group
+    _affix = (*_affix_features)[_feature_group_char];
+    _spelling = (*_spelling_features)[_feature_group_char];
+    _namespace_dictionaries = &(*_namespace_dictionaries_map)[feature_group_char()];
```
olgavrou, line 97: nit: does this need to be dereferenced and then referenced? Can we just do:

`_namespace_dictionaries = _namespace_dictionaries_map`?

vowpalwabbit/parse_example.cc
```diff
@@ -50,29 +50,113 @@ int read_features_string(vw* all, v_array<example*>& examples)
...
+      , _hash_function(hash_fn)
+  {
+  }
+  unsigned char feature_group_char() const { return _feature_group_char; }
+  uint64_t feature_group_hash() const { return _hash; }
+  features& get_current_fg() { return get_feature_group(_p_current_feature_group, feature_group_char()); }
+  features& get_affix_fg() { return get_feature_group(_p_affix_feature_group, affix_namespace); }
+  features& get_spelling_fg() { return get_feature_group(_p_spelling_feature_group, spelling_namespace); }
+  features& get_dictionary_fg() { return get_feature_group(_p_dictionary_feature_group, dictionary_namespace); }
+  void set_current_fg(const VW::string_view fg_name)
```
olgavrou, line 87: nit:

since `fg_name` is passed by value I think `const` is redundant

olgavrou (commented):
I like this PR, tidies up `parse_example`:)

test/benchmarks/CMakeLists.txt
rajan-chari: I am fine with another location in the repo for microbenchmarks used to justify coding decisions around performance.  I think this is essential if performance is important.  I can move this to a different PR.

vowpalwabbit/parse_example.cc
rajan-chari: I don't see how.  The [] operator returns a reference of the value stored in the container.  & is used to get the address to that reference.  Maybe parentheses can make things more clear.

vowpalwabbit/parse_example.cc
rajan-chari: const tells the compiler to enforce that fg_name won't change inside the function and perhaps allows opportunities for optimization?

jackgerrits (approved):

vowpalwabbit/parse_example.cc
olgavrou: but since it's passed by value it does it matter if it changes inside the function, the guarantee stays the same from the caller point of view afaik

vowpalwabbit/parse_example.cc
olgavrou: oh, nevemind, I didn't see the [] operator, github diff put it on a different line

vowpalwabbit/parse_example.cc
olgavrou: I guess it doesn't matter either way (hence the nit)

jackgerrits: When the argument is a value anyway there is little reason to make it const. The compiler isn't really going to do anything extra because of the const (and it rarely does because const in C++ isnt very binding...).

olgavrou (approved):

jackgerrits:
Different approach to extents makes this no longer relevant for its original purpose.

PR #3180, opened 2021-07-21
refactor: replace c-style with c++ casts

jackgerrits:


olgavrou (approved):

vowpalwabbit/vw_allreduce.h
```diff
@@ -15,11 +15,11 @@ void all_reduce(vw& all, T* buffer, const size_t n)
   switch (all.all_reduce_type)
   {
     case AllReduceType::Socket:
-      ((AllReduceSockets*)all.all_reduce)->all_reduce<T, f>(buffer, n);
+      dynamic_cast<AllReduceSockets*>(all.all_reduce)->all_reduce<T, f>(buffer, n);
```
olgavrou, line 18: `dynamic_cast` might return a `nullptr` afaik if it can't do the cast, do we want to check for that?

PR #3181, opened 2021-07-21
feat: [Tutorial] DFtoVW tutorial v2

etiennekintzler:
Version 2 of https://github.com/VowpalWabbit/vowpal_wabbit/pull/3068

Typo corrected + some simplifications.

As @jackgerrits suggested it would have been better to have the underlying dataset as zip but I didn't find it out (it is available as zip on kaggle but one has to log in).


jackgerrits:
Is the content of the two copies the same? If so then the `python/examples/` file should not be added. All examples should go in the other location and we should probably add a readme pointing people there since it is the expected location.

etiennekintzler:
> Is the content of the two copies the same? If so then the `python/examples/` file should not be added. All examples should go in the other location and we should probably add a readme pointing people there since it is the expected location.

That's a mistake on my side, I forgot to remove the previous files

> The logo in the title of the notebook is causing the website rendering to go wonky. Can it be removed?

Sure, done !

jackgerrits (approved):

PR #3187, opened 2021-07-26
fix: lbl_parser should be set after calling setup base

lalo:


jackgerrits (approved):

vowpalwabbit/cats_tree.cc
```diff
@@ -372,6 +372,7 @@ base_learner* setup(options_i& options, vw& all)
                 .set_prediction_type(prediction_type_t::multiclass)
                 .set_label_type(label_type_t::cb)
                 .build();
+  all.example_parser->lbl_parser = CB::cb_label;
```
lalo, line 375: set_label_type on line 373: looks like its not using that label to set example_parser

PR #3189, opened 2021-07-27
refactor: rename setup_base to stack_builder

lalo:
- removes parse_args.h where possible (mostly setup reductions)
- renames setup_base to stack_builder
- collapses all and options into stack_builder to reduce args to setup function to 1
- adds interface to access options and all via stack_builder
- renames operator() of setup_base to setup_base_learner()
- fwds (todo) get_setupfn_name calls via stack_builder (impl should migrate from all to s_b)

binary.cc setup function is completely all free, and a couple of others

Good to review first: reductions_fwd.h, parse_args.cc and binary.cc

### DIFF
--- a/vowpalwabbit/audit_regressor.cc
+++ b/vowpalwabbit/audit_regressor.cc
@@ -4,7 +4,6 @@
 
 #include "reductions.h"
 #include "interactions.h"
-
 #include "vw.h"
 #include "shared_data.h"
 #include "gd.h"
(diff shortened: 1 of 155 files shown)


vowpalwabbit/audit_regressor.cc
```diff
@@ -4,7 +4,7 @@
 
 #include "reductions.h"
 #include "interactions.h"
-#include "parse_args.h"
+
```
lalo, line 7:
```suggestion
```

jackgerrits (approved):

PR #3190, opened 2021-07-27
feat: ability to customize learner builder in vw.h

lalo:
This PR adds the following experimental init fn to vw.h:

```
// Experimental (VW::setup_base_i):
vw* initialize_with_builder(std::string s, io_buf* model = nullptr, bool skipModelLoad = false,
    trace_message_t trace_listener = nullptr, void* trace_context = nullptr,
    std::unique_ptr<VW::setup_base_i> = nullptr);
```

This allows any user of VW to alter the reduction stack or augment it as they find fit. See unit tests for examples of flexibility. They are able to inherit from our concrete impls of setup_base_i or provide their own.

This could also be leveraged by https://github.com/VowpalWabbit/vowpal_wabbit/pull/2971

### DIFF
--- a/vowpalwabbit/parse_args.cc
+++ b/vowpalwabbit/parse_args.cc
@@ -1410,7 +1410,10 @@ void parse_modules(
   parse_output_preds(options, all);
 }
 
-void parse_reductions(vw& all, std::unique_ptr<VW::setup_base_i> learner_builder)
+// note: Although we have the option to override setup_base_i,
+// the most common scenario is to use the default_reduction_stack_setup.
+// Expect learner_builder to be nullptr most/all of the cases.
+void instantiate_learner(vw& all, std::unique_ptr<VW::setup_base_i> learner_builder)
 {
   if (!learner_builder)
   { learner_builder = VW::make_unique<VW::default_reduction_stack_setup>(all, *all.options.get()); }
@@ -1590,7 +1593,7 @@ vw* initialize_with_builder(std::unique_ptr<options_i, options_deleter_type> opt
 
     std::vector<std::string> dictionary_nses;
     parse_modules(*all.options.get(), all, interactions_settings_duplicated, dictionary_nses);
-    parse_reductions(all, std::move(learner_builder));
+    instantiate_learner(all, std::move(learner_builder));
     parse_sources(*all.options.get(), all, *model, skipModelLoad);
 
     // we must delay so parse_mask is fully defined.
@@ -1668,8 +1671,8 @@ vw* initialize_with_builder(int argc, char* argv[], io_buf* model, bool skipMode
       std::move(options), model, skipModelLoad, trace_listener, trace_context, std::move(learner_builder));
 }
 
-vw* initialize(int argc, char* argv[], io_buf* model, bool skipModelLoad, trace_message_t trace_listener,
-    void* trace_context)
+vw* initialize(
+    int argc, char* argv[], io_buf* model, bool skipModelLoad, trace_message_t trace_listener, void* trace_context)
 {
   return initialize_with_builder(argc, argv, model, skipModelLoad, trace_listener, trace_context, nullptr);
 }
@@ -1683,7 +1686,8 @@ vw* initialize_with_builder(std::string s, io_buf* model, bool skipModelLoad, tr
 
   try
   {
-    ret = initialize_with_builder(argc, argv, model, skipModelLoad, trace_listener, trace_context, std::move(learner_builder));
+    ret = initialize_with_builder(
+        argc, argv, model, skipModelLoad, trace_listener, trace_context, std::move(learner_builder));
   }
   catch (...)
   {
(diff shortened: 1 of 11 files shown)


vowpalwabbit/vw.h
```diff
@@ -44,13 +44,16 @@ namespace VW
     (2) The code is not yet reentrant.
    */
 vw* initialize(std::unique_ptr<config::options_i, options_deleter_type> options, io_buf* model = nullptr,
-    bool skipModelLoad = false, trace_message_t trace_listener = nullptr, void* trace_context = nullptr);
+    bool skipModelLoad = false, trace_message_t trace_listener = nullptr, void* trace_context = nullptr,
```
jackgerrits, line 47: I do not think this should be added to the standard `initialize` calls if it is for testing. Can this be an internal only `initialize` implementation?

lalo: I was planning on refactoring python reduction work to re-use this work, and it would use this same vw.h API. Thoughts?

lalo: it could also be a test and binding only API (controlled via cmake)

jackgerrits: I see, I still think keeping it separate to these entrypoints may make things a little easier. Can it be called `initialize_custom` or something?

vowpalwabbit/parse_args.cc
```diff
@@ -1643,16 +1662,32 @@ vw* initialize(std::string s, io_buf* model, bool skipModelLoad, trace_message_t
   return ret;
 }
 
-vw* initialize_escaped(
-    std::string const& s, io_buf* model, bool skipModelLoad, trace_message_t trace_listener, void* trace_context)
+vw* initialize_with_builder(int argc, char* argv[], io_buf* model, bool skipModelLoad, trace_message_t trace_listener,
+    void* trace_context, std::unique_ptr<VW::setup_base_i> learner_builder)
+{
+  std::unique_ptr<options_i, options_deleter_type> options(
+      new config::options_boost_po(argc, argv), [](VW::config::options_i* ptr) { delete ptr; });
```
olgavrou, line 1669: can we use `VW::make_unique` here instead of `new`?

vowpalwabbit/parse_args.cc
```diff
@@ -1408,9 +1408,20 @@ void parse_modules(
...
+// note: Although we have the option to override setup_base_i,
+// the most common scenario is to use the default_reduction_stack_setup.
+// Expect learner_builder to be nullptr most/all of the cases.
+void instantiate_learner(vw& all, std::unique_ptr<VW::setup_base_i> learner_builder)
+{
+  if (!learner_builder)
+  { learner_builder = VW::make_unique<VW::default_reduction_stack_setup>(all, *all.options.get()); }
+  else
+  {
+    learner_builder->delayed_attach(all, *all.options.get());
```
olgavrou, line 1422: nit: rename suggestion `delayed_state_attach`

vowpalwabbit/vw.h
lalo: done

vowpalwabbit/parse_args.cc
```diff
@@ -1624,10 +1637,17 @@ vw* initialize(std::unique_ptr<options_i, options_deleter_type> options, io_buf*
   }
 }
 
-vw* initialize(std::string s, io_buf* model, bool skip_model_load, trace_message_t trace_listener, void* trace_context)
+vw* initialize(std::unique_ptr<options_i, options_deleter_type> options, io_buf* model, bool skip_model_load,
```
lalo, line 1640: most of these fns have been shuffled around, see vw.h for a diff on what actually changed from the public side.

jackgerrits (approved):

vowpalwabbit/parse_args.cc
lalo: This is moved code, so we can address in separate PR. Although Jack confirmed that VW::make_unique right now doesn't have support for custom delete.

vowpalwabbit/parse_args.cc
lalo: done

vowpalwabbit/reduction_stack.h
```diff
@@ -23,8 +28,10 @@ struct default_reduction_stack_setup : public setup_base_i
   std::string get_setupfn_name(reduction_setup_fn setup) override;
 
 private:
-  std::vector<std::tuple<std::string, reduction_setup_fn>> reduction_stack;
   VW::config::options_i* options_impl = nullptr;
   vw* all_ptr = nullptr;
+
+protected:
```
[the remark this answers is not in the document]
lalo: no, this allows anybody to inherit this class and alter the stack without having to figure out where the setup functions of the reductions live. You get access to it on runtime, and before actually running setup. The tests are just demonstrating what's now possible with this new api.

olgavrou (approved):

PR #3193, opened 2021-07-27
feat: Add benchmarks to track full namespace performance

bassmang:


test/benchmarks/standalone/benchmark_text_input.cc
```diff
@@ -85,6 +85,54 @@ static void benchmark_ccb_adf_learn(benchmark::State& state, std::string feature
   vw->finish_example(examples);
 }
 
+static void benchmark_cb_adf_large(
+    benchmark::State& state, int num_feature_groups, bool same_first_char, bool interactions)
+{
+  std::string cmd = "--cb_explore_adf --quiet";
+  if (interactions) { cmd += " -q ::"; }
+  auto vw = VW::initialize(cmd, nullptr, false, nullptr, nullptr);
+  int batch_size = 100;       // 20000;
```
olgavrou, line 94: what are these comments next to each variable?

test/benchmarks/standalone/benchmark_text_input.cc
```diff
@@ -85,6 +85,54 @@ static void benchmark_ccb_adf_learn(benchmark::State& state, std::string feature
...
+      std::ostringstream action_ss;
+      if (ac == action_ind) { action_ss << action_ind << ":1.0:0.5 "; }
+      action_ss << "|";
+      if (same_first_char) { action_ss << "f"; }
+      action_ss << (rand() % num_feature_groups);
+      for (int action_feat = 0; action_feat < action_feats_count; ++action_feat)
+      { action_ss << " " << (rand() % action_feats_size); }
+      examples.push_back(VW::read_example(*vw, action_ss.str()));
+    }
+    examples_vec.push_back(examples);
```
olgavrou, line 122: would the benchmark be more accurate if the examples (or example strings) were built once and then the batch loop is done inside the (for (auto _ :state)) loop?

otherwise I am not sure how this will be counted since google benchmarks benchmark the code inside the `for state` loop and we are doing that `batch_size` amount of times instead of one `for state` loop with `batch_size` executions

olgavrou (approved):

jackgerrits (approved):

PR #3194, opened 2021-07-28
fix: move some cb metrics so that they are available during ccb

olgavrou:
moved some metrics from `output_example` which doesn't get called during ccb reduction, into `learn` which will get called


vowpalwabbit/cb_explore_adf_common.h
```diff
@@ -152,6 +152,18 @@ inline void cb_explore_adf_base<ExploreType>::learn(
     data.explore.learn(base, examples);
     if (data._metrics)
     {
+      size_t num_features = 0;
+      size_t num_namespaces = 0;
+
+      for (const auto& example : examples)
+      {
+        num_features += example->get_num_features();
```
lalo, line 160: @jackgerrits get_num_features can be called only after learn right?

jackgerrits: Yes, the interacted count requires learn be called

olgavrou: oh so should be moved after learn

test/test-sets/ref/metrics.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{"total_predict_calls":3,"total_learn_calls":3,"sfm_count_learn_example_with_shared":3,"cbea_labeled_ex":3,"cbea_predict_in_learn":0,"cbea_label_first_action":0,"cbea_label_not_first":3,"cbea_non_zero_cost":2,"cbea_min_actions":12,"cbea_max_actions":12,"total_log_calls":0,"cbea_sum_cost":-2.0,"cbea_sum_cost_baseline":0.0,"cbea_avg_feat_per_event":1221.0,"cbea_avg_actions_per_event":12.0,"cbea_avg_ns_per_event":105.0,"cbea_avg_feat_per_action":101.0,"cbea_avg_ns_per_action":8.0,"number_skipped_events":0,"number_events_zero_actions":0,"line_parse_error":0,"first_event_id":"0074434d3a3a46529f65de8a59631939","first_event_time":"","last_event_id":"9077f996581148978a0ebe2484260dab","last_event_time":""}
```
lalo, line 1: This file having a diff is interesting, its a cb test. It should have been exactly the same, right?

olgavrou: yeah I am wondering about that actually and trying to see why... maybe by the time we call `output_example` any extra interactions would have been cleaned up and now they are not? not sure

lalo: if anything, it was buggy before since here you are checking right on the spot. In the other case, it was under finish call

lalo: the cause probably is that before finish was being called with the top example being sent which has shared, now we are reporting post shared merger. This is a bug of shared feature merger.

lalo (approved):

PR #3201, opened 2021-08-02
refactor: Add label and use builder to set learner features in ftrl.cc

bassmang:


vowpalwabbit/ftrl.cc
```diff
@@ -25,26 +26,26 @@ namespace logger = VW::io::logger;
...
 
 struct ftrl
 {
-  vw* all;  // features, finalize, l1, l2,
-  float ftrl_alpha;
-  float ftrl_beta;
+  vw* all = nullptr;  // features, finalize, l1, l2,
+  float ftrl_alpha = 0.f;
+  float ftrl_beta = 0.f;
   struct ftrl_update_data data;
```
jackgerrits, line 44: While you're here can you update this?

```suggestion
  ftrl_update_data data;
```

vowpalwabbit/ftrl.cc
```diff
@@ -434,17 +436,25 @@ base_learner* ftrl_setup(VW::setup_base_i& stack_builder)
...
+        stack_builder.get_setupfn_name(ftrl_setup) + "-" + algorithm_name, prediction_type_t::scalar,
+        label_type_t::simple)
+            .set_learn_returns_prediction(learn_returns_prediction)
+            .set_params_per_weight(UINT64_ONE << all.weights.stride_shift())
+            .set_sensitivity(sensitivity)
+            .set_multipredict(multipredict<false>)
+            .set_save_load(save_load)
+            .set_end_pass(end_pass)
+            .build();
   return make_base(*l);
```
jackgerrits, line 459: There's a lot of duplication here. Can you do the shared things outside of the `if`?

vowpalwabbit/learner.h
```diff
@@ -1063,9 +1066,9 @@ reduction_no_data_learner_builder<ExampleT, BaseLearnerT> make_no_data_reduction
   return builder;
 }
 
-template <class DataT, class ExampleT>
+template <class DataT, class BaseLearnerT, class ExampleT>
```
jackgerrits, line 1069: Why is this needed? The point of a base learner is that it doesn't take in a base itself.

jackgerrits (changes requested):

vowpalwabbit/ftrl.cc
jackgerrits: Hm yeah I see, thats pretty cumbersome isnt it

vowpalwabbit/learner.h
jackgerrits: A base learner should never actually use the base_learner parameter (pretty sure it is a nullptr here), which is why it is specified as a base_learner explicitly. I'd say this is a bug in the ftrl code and the the signature of those funcs need to change.

vowpalwabbit/ftrl.cc
jackgerrits: Suggestion:
```c++
... predict_ptr = (all.audit || all.hash_inv) ? predict<true> : predict<false>;
... multipredict_ptr = (all.audit || all.hash_inv) ? multipredict<true> : multipredict<false>;

l = VW::LEARNER::make_base_learner(std::move(b), learn_ptr, predict_ptr,
        stack_builder.get_setupfn_name(ftrl_setup) + "-" + algorithm_name + "-audit", prediction_type_t::scalar,
        label_type_t::simple)
            .set_learn_returns_prediction(learn_returns_prediction)
            .set_params_per_weight(UINT64_ONE << all.weights.stride_shift())
            .set_sensitivity(sensitivity)
            .set_multipredict(multipredict_ptr)
            .set_save_load(save_load)
            .set_end_pass(end_pass)
            .build();
```

vowpalwabbit/ftrl.cc
```diff
@@ -434,17 +436,25 @@ base_learner* ftrl_setup(VW::setup_base_i& stack_builder)
...
+            .set_end_pass(end_pass)
+            .build();
   else
-    l->set_multipredict(multipredict<false>);
-  l->set_save_load(save_load);
-  l->set_end_pass(end_pass);
+    l = VW::LEARNER::make_base_learner(std::move(b), learn_ptr, predict<false>,
+        stack_builder.get_setupfn_name(ftrl_setup) + "-" + algorithm_name, prediction_type_t::scalar,
+        label_type_t::simple)
+            .set_learn_returns_prediction(learn_returns_prediction)
```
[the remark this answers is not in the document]
bassmang: based on the original implementation it's only set for 1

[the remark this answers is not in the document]
jackgerrits: I see, that seems like a bug since audit should not change if it returns a prediction

vowpalwabbit/ftrl.cc
bassmang: Also added conditional for name

PR #3203, opened 2021-08-03
refactor: Add label and use builder to set learner features in oaa.cc

bassmang:


vowpalwabbit/oaa.cc
```diff
@@ -264,51 +264,71 @@ VW::LEARNER::base_learner* oaa_setup(VW::setup_base_i& stack_builder)
   }
 
   oaa* data_ptr = data.get();
-  VW::LEARNER::learner<oaa, example>* l;
+  uint64_t kVal = data->k;
```
jackgerrits, line 267: Change to match naming scheme
```suggestion
  uint64_t k_value = data->k;
```

vowpalwabbit/oaa.cc
```diff
@@ -264,51 +264,71 @@ VW::LEARNER::base_learner* oaa_setup(VW::setup_base_i& stack_builder)
...
+    else
+    {
+      learn_ptr = learn<false, false, false>;
+      pred_ptr = predict<false, false, false>;
+      name_addition = "";
+    }
   }
 
+  all.example_parser->lbl_parser = MULTICLASS::mc_label;
+  all.example_parser->lbl_parser.label_type = label_type_t::multiclass;
```
jackgerrits, line 317: This field should not be changed in a reduction, it is intended to only be set by the label parser itself.

jackgerrits (changes requested):

vowpalwabbit/oaa.cc
bassmang: Changed

vowpalwabbit/oaa.cc
bassmang: Got it, removed `all.example_parser->lbl_parser.label_type = label_type_t::multiclass;`

vowpalwabbit/oaa.cc
```diff
@@ -264,51 +264,70 @@ VW::LEARNER::base_learner* oaa_setup(VW::setup_base_i& stack_builder)
...
+    }
   }
 
+  all.example_parser->lbl_parser = MULTICLASS::mc_label;
+
   if (data_ptr->num_subsample > 0)
   {
-    l->set_learn(learn_randomized);
-    l->set_finish_example(MULTICLASS::finish_example_without_loss<oaa>);
-    l->learn_returns_prediction = false;
```
jackgerrits, line 310: `learn_returns_prediction` which makes this line a noop. Was this intentional? @rajan-chari

test/benchmarks/standalone/benchmark_text_input.cc
```diff
@@ -91,7 +91,7 @@ static void benchmark_cb_adf_large(
   std::string cmd = "--cb_explore_adf --quiet";
   if (interactions) { cmd += " -q ::"; }
   auto vw = VW::initialize(cmd, nullptr, false, nullptr, nullptr);
-  int example_size = 100;
+  int example_size = 10;
```
jackgerrits, line 94: Whats the reason behind this change?

bassmang: I've been seeing the benchmarking CI fail a lot recently, and thought it may have to do with these new benchmarks I added being too large

jackgerrits: Ah yeah those are tricky. Cool lets see if this helps

vowpalwabbit/oaa.cc
rajan-chari: The semantics and name of that variable was reversed so this is likely vestigial and can be removed.
There are reductions with conditional behavior for learn_returns_predicttion where being explicit reduces cognitive burden.

vowpalwabbit/oaa.cc
```diff
@@ -264,51 +264,70 @@ VW::LEARNER::base_learner* oaa_setup(VW::setup_base_i& stack_builder)
...
     {
       auto loss_function_type = all.loss->getType();
       if (loss_function_type != "logistic")
         *(all.trace_message) << "WARNING: --probabilities should be used only with --loss_function=logistic"
                              << std::endl;
       // the three boolean template parameters are: is_learn, print_all and scores
-      l = &LEARNER::init_multiclass_learner(data, base, learn<false, true, true>, predict<false, true, true>,
-          all.example_parser, data->k, stack_builder.get_setupfn_name(oaa_setup) + "-prob", prediction_type_t::scalars);
-      all.example_parser->lbl_parser.label_type = label_type_t::multiclass;
+      learn_ptr = learn<false, true, true>;
```
rajan-chari, line 284: I am wondering if along with this change, we could start to make parameterization of template functions more explicit.

for example in this case it would look something like
learn_ptr = learn<!Print_All, Scores, Probabilities>

This would really help with functions in gd.

bassmang: I added some directives defining some constants on top as true, so now we can define the function pointers like this:

learn_ptr = learn<!PRINT_ALL, SCORES, PROBABILITIES>;

This should make things more clear

PR #3207, opened 2021-08-04
feat: print expected input label and output pred type

lalo:


jackgerrits (approved):
Cool!

vowpalwabbit/learner.h
```diff
@@ -32,13 +32,14 @@ enum class prediction_type_t
   multiclass,
   multilabels,
   prob,
-  multiclassprobs,
+  multiclassprobs,  // not in use?
```
lalo, line 35: this seems to not be referenced anywhere

vowpalwabbit/parse_args.cc
```diff
@@ -1616,6 +1616,12 @@ vw* initialize_with_builder(std::unique_ptr<options_i, options_deleter_type> opt
 
     print_enabled_reductions(all, enabled_reductions);
 
+    if (!all.logger.quiet)
+    {
+      *(all.trace_message) << "Input label = " << to_string(all.example_parser->lbl_parser.label_type) << std::endl;
```
lalo, line 1621: we could also print something fancier rather than the explicit type, but at least this gives a clue on what was the computed expected input and what is going to be the output type pred (in theory)

vowpalwabbit/learner.h
jackgerrits: Maybe thats the corresponding type for when multilabels is used as a prediction. Althought I cannot see it used anywhere

olgavrou (approved):

vowpalwabbit/learner.h
lalo:
```suggestion
  multiclassprobs,  // not in use (technically oaa.cc)
```

vowpalwabbit/learner.h
```diff
@@ -32,13 +32,14 @@ enum class prediction_type_t
...
   multilabels,
   prob,
-  multiclassprobs,
+  multiclassprobs,  // not in use (technically oaa.cc)
   decision_probs,
   action_pdf_value,
   active_multiclass
 };
 
 const char* to_string(prediction_type_t prediction_type);
```
jackgerrits, line 41: These functions already exist now in label_type.h and prediction_type.h

lalo: thanks

jackgerrits (approved):

PR #3208, opened 2021-08-04
feat: extent based namespaces

jackgerrits:


vowpalwabbit/feature_group.cc
```diff
@@ -84,6 +94,12 @@ void features::concat(const features& other)
 
   if (!other.space_names.empty())
   { space_names.insert(space_names.end(), other.space_names.begin(), other.space_names.end()); }
+
+  for (const auto& ns_extent : other.namespace_extents)
+  {
+    namespace_extents.emplace_back(
+        ns_extent.begin_index + extent_offset, ns_extent.end_index + extent_offset, ns_extent.hash);
+  }
 }
```
bassmang, line 103: Could add logic here to combine last extent of first group with first extent of last group if they share the same hash

vowpalwabbit/feature_group.h
```diff
@@ -289,10 +365,14 @@ struct features
...
   using const_audit_iterator = audit_features_iterator<const feature_value, const feature_index, const audit_strings>;
+  using extent_iterator = ns_extent_iterator<features, audit_iterator, std::vector<VW::namespace_extent>::iterator>;
+  using const_extent_iterator =
+      ns_extent_iterator<const features, const_audit_iterator, std::vector<VW::namespace_extent>::const_iterator>;
 
-  v_array<feature_value> values;               // Always needed.
-  v_array<feature_index> indicies;             // Optional for sparse data.
-  std::vector<audit_strings> space_names;      // Optional for audit mode.
+  v_array<feature_value> values;           // Always needed.
+  v_array<feature_index> indicies;         // Optional for sparse data.
```
bassmang, line 373: nit: Is this v_array still optional as per the comment?

vowpalwabbit/feature_group.h
```diff
@@ -352,6 +437,15 @@ struct features
   void push_back(feature_value v, feature_index i);
   bool sort(uint64_t parse_mask);
 
+  void start_ns_extent(uint64_t hash);
+  void end_ns_extent();
+
+  bool all_extents_complete()
+  {
+    return std::all_of(namespace_extents.begin(), namespace_extents.end(),
+        [](const VW::namespace_extent& obj) { return obj.begin_index != obj.end_index && obj.end_index != 0; });
```
bassmang, line 446: Could be safer to use `obj.begin_index < obj.end_index`

vowpalwabbit/feature_group.h
jackgerrits: Yes, but in practice it is almost always used

vowpalwabbit/feature_group.cc
jackgerrits: Yeah youre right they should merge here if needed

vowpalwabbit/feature_group.cc
```diff
@@ -84,6 +94,24 @@ void features::concat(const features& other)
 
   if (!other.space_names.empty())
   { space_names.insert(space_names.end(), other.space_names.begin(), other.space_names.end()); }
+
+  // If the back of the current list and the front of the other list have the same hash then merge the extent.
+  size_t offset = 0;
+  if (!namespace_extents.empty() && !other.namespace_extents.empty())
+  {
+    if (namespace_extents.back().hash == other.namespace_extents.front().hash)
```
bassmang, line 102: These if statements can be combined to `if (!namespace_extents.empty() && !other.namespace_extents.empty() && !namespace_extents.empty() && !other.namespace_extents.empty()`

jackgerrits: Good point!

bassmang (approved):

vowpalwabbit/feature_group.h
```diff
@@ -41,6 +41,38 @@ struct feature
...
+
+  namespace_extent(size_t begin_index, uint64_t hash) : begin_index(begin_index), hash(hash) {}
+
+  size_t begin_index = 0;
+  size_t end_index = 0;
+  uint64_t hash = 0;
+
+  friend bool operator==(const namespace_extent& lhs, const namespace_extent& rhs)
+  {
+    return lhs.begin_index == rhs.begin_index && lhs.end_index == rhs.end_index && lhs.hash == rhs.hash;
```
rajan-chari, line 63: Does it make sense to start with the hash first since begin_index is likely to be the same in many cases?

jackgerrits: yeah that sounds reasonable

vowpalwabbit/feature_group.cc
```diff
@@ -48,6 +49,14 @@ void features::truncate_to(size_t i, float sum_feat_sq_of_removed_section)
   if (indicies.end() != indicies.begin()) { indicies.resize_but_with_stl_behavior(i); }
 
   if (space_names.size() > i) { space_names.erase(space_names.begin() + i, space_names.end()); }
+
+  while (!namespace_extents.empty() && namespace_extents.back().begin_index >= i) { namespace_extents.pop_back(); }
```
rajan-chari, line 53: as an alternative std::find_if and std::erase can be used here

vowpalwabbit/feature_group.cc
```diff
@@ -76,6 +85,7 @@ void features::concat(const features& other)
   assert(!(!empty() && (space_names.empty() != other.space_names.empty())));
```
rajan-chari, line 85: This comment makes this condition much easier to read.

vowpalwabbit/feature_group.cc
jackgerrits: pop_back is used instead of a combo of remove_if and erase because pop_back is guaranteed to be constant time. If erase is used then it can be used in such a way to be linear over the size of the container. 

The idea is that removals *always* occur from the back and so if the situation cannot be implemented with pop_back then there is some change of semantics that goes against the design.

test/unit_test/feature_group_test.cc
```diff
@@ -39,34 +43,85 @@ BOOST_AUTO_TEST_CASE(unique_feature_group_test)
...
+  BOOST_REQUIRE(unflattened_list == extents);
+
+  extents = {{1, 2, 123}};
+
+  flat_list = VW::details::flatten_namespace_extents(extents, 2);
+  unflattened_list = VW::details::unflatten_namespace_extents(flat_list);
+  BOOST_REQUIRE(unflattened_list == extents);
+
+  extents = {{1, 2, 1}, {5, 8, 2}, {8, 12, 3}, {13, 14, 4}};
+  flat_list = VW::details::flatten_namespace_extents(extents, 18);
```
lalo, line 100: how are the 2's and 18 calculated from the extents?

jackgerrits: 18 is the length of the feature_group. This is separate because there can be a range with no hash (I intend to make it required, but it isn't in this PR). In this example positions 14, 15, 16 and 17 do not have a hash since they are after the last extent but before the end at position 18.

An extent is a tuple of `(start_index, end_index, hash)` so this maps to `{1, 2, 1}`. The range is `[start_index, end_index)`

vowpalwabbit/feature_group.cc
jackgerrits: I agree it's the same complexity, it's more about clearly stating we always remove from the back. This would make it a little less likely to be changed to some inner removal in future. The constructions of extents (should) ensure there are no empty extents in the middle and for an empty namespace there would be nothing to remove.

vowpalwabbit/feature_group.cc
```diff
@@ -140,13 +227,50 @@ bool features::sort(uint64_t parse_mask)
   apply_permutation_in_place(values, dest_index_vec);
```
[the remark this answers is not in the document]
jackgerrits: I don't understand the comment? The sort order is determined through the corresponding elements of the two vectors then the determined sort order is applied to both values and indices

vowpalwabbit/parser/flatbuffer/parse_example_flatbuffer.cc
```diff
@@ -177,6 +177,11 @@ void parser::parse_namespaces(vw* all, example* ae, const Namespace* ns)
   ae->indices.push_back(temp_index);
 
   auto& fs = ae->feature_space[temp_index];
+  if (flatbuffers::IsFieldPresent(ns, Namespace::VT_EXTENTS))
```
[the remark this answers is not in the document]
jackgerrits: Yep, you make an excellent point. The changes to the schema don't seem right - I'll revisit this.

jackgerrits:
@rajan-chari, the flatbuffer changes should be better now

vowpalwabbit/feature_group.cc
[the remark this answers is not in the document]
rajan-chari: base on function definition:
template <typename IndexVec, typename ValVec, typename Compare>
std::vector<std::size_t> sort_permutation(const IndexVec& index_vec, const ValVec& value_vec, const Compare& compare)

values and indicies seem to be reversed for the call:
auto dest_index_vec = sort_permutation(values, indicies, comparator);

vowpalwabbit/feature_group.cc
```diff
@@ -140,13 +248,50 @@ bool features::sort(uint64_t parse_mask)
...
```
rajan-chari, line 248: Worried about allocations that can happen in steady state:
1) sort operation (seems avoidable)
2) flatten_namespace

vowpalwabbit/feature_group.cc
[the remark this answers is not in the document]
jackgerrits: Ah I see, I didn't realize you were referring to the args of the function and then here.

vowpalwabbit/parser/flatbuffer/schema/example.fbs
```diff
@@ -10,6 +10,7 @@ table Namespace {
   name:string;
   hash:uint8;
   features:[Feature];
+  full_hash:uint64;
```
rajan-chari, line 13: a /// comment for this element would be useful.

vowpalwabbit/parser/flatbuffer/schema/example.fbs
```diff
@@ -10,6 +10,7 @@ table Namespace {
   name:string;
   hash:uint8;
```
rajan-chari, line 11: a /// comment for this element would be useful.

vowpalwabbit/parser/flatbuffer/schema/example.fbs
jackgerrits: Will do. I'd rather call this hash but that is already taken by the other element which is really the index.

rajan-chari:
> @rajan-chari, the flatbuffer changes should be better now

Yep.  Looks good.  Thanks.

vowpalwabbit/parser/flatbuffer/schema/example.fbs
rajan-chari: Yeah.  I agree with your preference.

vowpalwabbit/feature_group.cc
jackgerrits: Sort is very rarely used, it is only used with `limit` and `sort_features`. It is also done only once per example, so in practice this will not matter

vowpalwabbit/parser/flatbuffer/schema/example.fbs
jackgerrits: Should I switch them? I am worried that will break things. What do you think?

jackgerrits: Going to leave as is with a comment

vowpalwabbit/feature_group.cc
jackgerrits: As discussed the apply function was rolled into one to avoid reallocating the done vec

PR #3212, opened 2021-08-11
feat: VW fuzzer

peterychang:
Adding test binary, python helper script, and initial CB models/commands


test/tools/fuzzing/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+add_executable(test_initialize main.cc)
```
jackgerrits, line 1: Can this name be more descriptive?

test/tools/fuzzing/README.txt
```diff
@@ -0,0 +1,16 @@
+To use this tool you will need to download and build Google's AFL fuzzer https://github.com/google/AFL
```
jackgerrits, line 1: Make this markdown and use code blocks for readability

test/tools/fuzzing/fuzz.py
```diff
@@ -0,0 +1,129 @@
...
+import os
+from pathlib import Path
+import threading
+import time
+import shlex
+import subprocess
+import sys
+import signal
+
+def find_in_path(paths, file_matcher, debug_file_name):
```
jackgerrits, line 14: Maybe leave a comment to say this is the same impl as in run_tests.py so someone can update both if it ever needs to change

test/tools/fuzzing/fuzz.py
```diff
@@ -0,0 +1,129 @@
...
+        '--test_bin_path', help="Specify test binary to use. Otherwise, binary will be searched for in build directory")
+    parser.add_argument(
+        '-t', '--timeout', type=int, default=60,  help="Max runtime for each fuzzing test in minutes. 0 will disable the timeout")
+    parser.add_argument(
+        '-f', '--full_tests',  action='store_true', help="Run full tests. This disables the '-d' option in AFL")
+    #parser.add_argument(
+    #    '-p', '--parallel', type=int, default=1,  help="Number of CPUs to use per test.")
+    args = parser.parse_args()
+
+    # Unfortunately, there doesn't appare to be a sane way to run different tests concurrently. Therefore this script will
```
jackgerrits, line 69:
```suggestion
    # Unfortunately, there doesn't appear to be a sane way to run different tests concurrently. Therefore this script will
```

test/tools/fuzzing/fuzz.py
```diff
@@ -0,0 +1,129 @@
...
+            # VW test binary and options
+            cmd.extend([test_bin, row['command'], '--quiet', '--no_stdin', '-i', '@@'])
+
+            timeout = None
+            if args.timeout != 0:
+                timeout = 60*args.timeout
+
+            # shlex.join is only available in python 3.8+
+            cmd = ' '.join(cmd)
+            print("running ", cmd)
```
jackgerrits, line 115: Nit: throughout
```suggestion
            print(f"running {cmd}")
```

test/tools/fuzzing/fuzz.py
```diff
@@ -0,0 +1,129 @@
+import argparse
```
jackgerrits, line 1: Please run `black` over this file for consistent formatting

test/tools/fuzzing/main.cc
```diff
@@ -0,0 +1,18 @@
...
+
+#include "vw.h"
+#include "vw_exception.h"
+#include <memory>
+
+int main(int argc, char** argv)
+{
+  std::unique_ptr<VW::config::options_i> ptr(new VW::config::options_boost_po(argc, argv));
+  try {
+    vw* all = VW::initialize(*ptr);
```
jackgerrits, line 11: You need to call finish otherwise this is a leak

test/tools/fuzzing/fuzz.py
```diff
@@ -0,0 +1,129 @@
...
+    print("Using VW binary: {}".format((test_bin)))
+
+    try:
+        os.mkdir('output')
+    except FileExistsError:
+        pass
+
+    # Create a temporary working directory for any output files VW produces
+    try:
+        os.mkdir('tmpwd')
```
jackgerrits, line 87: You can a temp dir to avoid having to clear it between runs. https://stackoverflow.com/questions/3223604/how-to-create-a-temporary-directory-and-get-its-path-file-name

peterychang: This doesn't clear anything out, it just creates a temp working directory and runs the tests in there. I suppose I could use tempfile, but it might be nice to keep it around to see what cache files get generated.

test/tools/fuzzing/main.cc
peterychang: Finish is slow, and I don't particularly care if there's a leak. The process will get scrapped and all the memory will get cleaned up regardless of what happens. This test is literally just testing the initialization step

jackgerrits: You can call delete on all, which should be faster than finish?

jackgerrits (approved):

test/tools/fuzzing/main.cc
peterychang: There are a lot of crashes we won't catch with this test, notably anything dealing with the learners. I say we set up another test that does a full end-to-end workflow. It'll be MUCH slower, but there's not much we can do about that

PR #3213, opened 2021-08-11
wip:  Fixes from issues found during fuzz testing 

rajan-chari:


vowpalwabbit/learner.h
```diff
@@ -415,7 +415,16 @@ struct learner
   // called anytime saving or loading needs to happen. Autorecursive.
   inline void save_load(io_buf& io, const bool read, const bool text)
   {
-    save_load_fd.save_load_f(save_load_fd.data, io, read, text);
+    try
+    {
+      save_load_fd.save_load_f(save_load_fd.data, io, read, text);
```
lalo, line 420: do we have more info on the impl of save_load that is throwing?

vowpalwabbit/parse_args.cc
```diff
@@ -1625,6 +1625,16 @@ vw* initialize_with_builder(std::unique_ptr<options_i, options_deleter_type> opt
...
     return &all;
   }
+  catch (VW::save_load_model_exception& e)
+  {
+    const bool DETETE_ALL = true, FINALIZE_REDUCTIONS=true;
+    std::stringstream better_msg;
+    better_msg << e.what() << " , model files = ";
+    for(const auto& regressor_name : all.initial_regressors)
+      better_msg << regressor_name << ", ";
+    finish(all, DETETE_ALL, !FINALIZE_REDUCTIONS);
```
jackgerrits, line 1635: Since we are tearing down VW, all we actually need to do here is `delete &all`, all cleanup is encapsulated in the destructor. That means you don't need to conditionally turn off the finalize regressor logic.

vowpalwabbit/learner.h
rajan-chari: There are a couple of ways of looking at this question.
1) Is there a specific answer to this question in the fuzzed models?
a) All the models I have looked at so far ~10 out of 50 all throw in cb_adf_explore_cover::save_load

2) Can we augment the exception with type information of the reduction implementing save_load?
a) Not at the moment.  For example the "name" of the learner instance is shared_feature_merger.  But the actual implementation is in cb_adf_explore_cover and the exception is thrown deeper in the callstack in bin_text_read_write_fixed_validated()

rajan-chari: If you are asking whether learner name is available at this point so better_msg from line 425 can contain the name, the answer is no.  learner here is going to say "shared_feature_merger::save_load" where as the error was in "cb_adf_explore_cover::save_load"

vowpalwabbit/vw.h
```diff
@@ -73,7 +73,7 @@ const char* are_features_compatible(vw& vw1, vw& vw2);
 /*
   Call finish() after you are done with the vw instance.  This cleans up memory usage.
  */
-void finish(vw& all, bool delete_all = true);
+void finish(vw& all, bool delete_all = true, bool finalize_reductions = true);
```
jackgerrits, line 76:
```suggestion
void finish(vw& all, bool delete_all = true);
```

vowpalwabbit/parse_args.cc
```diff
@@ -1567,6 +1567,18 @@ vw* initialize_with_builder(std::unique_ptr<options_i, options_deleter_type> opt
 
     return &all;
   }
+  catch (VW::save_load_model_exception& e)
+  {
+    const bool DETETE_ALL = true;
```
jackgerrits, line 1572:
```suggestion

```

vowpalwabbit/parse_args.cc
```diff
@@ -1567,6 +1567,18 @@ vw* initialize_with_builder(std::unique_ptr<options_i, options_deleter_type> opt
...
+    std::stringstream better_msg;
+    better_msg << e.what() << " , model files = ";
+    for(const auto& regressor_name : all.initial_regressors)
+      better_msg << regressor_name << ", ";
```
jackgerrits, line 1575:
```suggestion
    auto msg = fmt::format("{}, model files = {}", e.what(), fmt::join(all.initial_regressors, ", ");
```

PR #3218, opened 2021-08-13
refactor: Make interactions use iterators

jackgerrits:


vowpalwabbit/interactions_predict.h
```diff
@@ -205,46 +202,38 @@ inline void generate_interactions(const std::vector<std::vector<namespace_index>
...
-        fgd->loop_end = ft_cnt - 1;  // saving number of features for each namespace
-        fgd->ft_arr = &ft;
-        ++fgd;
+        state_data.push_back(feature_gen_data{ft.audit_begin(), ft.audit_end()});
       }
 
       // if any of interacting namespace has 0 features - whole interaction is skipped
       if (must_skip_interaction) continue;  // no_data_to_interact
 
+      auto fgd = state_data.data();                             // always equal to first ns
```
peterychang, line 223: nit: might be a little easier to read with
```
auto fgd = &state_data.front();
auto fgd2 = &state_data.back();
```

Doesn't really make a difference though, I just hate doing pointer math when it can be avoided

peterychang (approved):

vowpalwabbit/interactions_predict.h
jackgerrits: Something feels a little off about taking an address of a vector item even though it is totally legit

PR #3224, opened 2021-08-13
fix: fix undefined behavior in compile_gram

peterychang:


peterychang:
This is actually defined behavior, just maybe not desirable behavior. Closing until I can figure out exactly what we want to have happen here

PR #3227, opened 2021-08-17
fix: fix fallthrough warning on GCC C++11/C++14

jackgerrits:


explore/future_compat.h
```diff
@@ -17,7 +17,11 @@
 #else
 #define VW_STD17_CONSTEXPR
 #define VW_ATTR(name)
-#    define VW_FALLTHROUGH  // fall through
+#    if defined(__GNUC__) && __GNUC__ >= 7
```
lalo, line 20: do we need to update something here https://github.com/VowpalWabbit/vowpal_wabbit/blob/a9da12d32e7b4d027896762d98b3dcfdef34506e/explore/hash.h#L100

peterychang (approved):

explore/future_compat.h
jackgerrits: Nope, it is used the same.

PR #3228, opened 2021-08-17
refactor: prepare interactions for ranges

jackgerrits:


vowpalwabbit/interactions_predict.h
```diff
@@ -60,6 +60,13 @@ struct feature_gen_data
   }
 };
 
+inline bool has_empty_interaction(
+    const std::array<features, NUM_NAMESPACES>& feature_groups, const std::vector<namespace_index>& namespace_indexes)
+{
+  return !std::all_of(namespace_indexes.begin(), namespace_indexes.end(),
```
peterychang, line 66: use `std::any_of`, it should be faster

vowpalwabbit/interactions_predict.h
```diff
@@ -71,9 +78,42 @@ inline float INTERACTION_VALUE(float value1, float value2) { return value1 * val
 
 // #define GEN_INTER_LOOP
 
+std::tuple<std::pair<features::const_audit_iterator, features::const_audit_iterator>,
```
peterychang, line 81: I know its kind of bad practice, but should we typedef this nasty return type for readability's sake?

vowpalwabbit/interactions_predict.h
```diff
@@ -71,9 +78,42 @@ inline float INTERACTION_VALUE(float value1, float value2) { return value1 * val
...
+    std::pair<features::const_audit_iterator,
+        features::const_audit_iterator>> inline generate_quadratic_char_combination(const std::array<features,
```
peterychang, line 83: nit: the spacing here is hard to read. Can we go with something like
```
inline generate_quadratic_char_combination(
    const std::array<features, NUM_NAMESPACES>& feature_groups,
    namespace_index ns_idx1,
    namespace_index ns_idx2
)
```
?

vowpalwabbit/interactions_predict.h
```diff
@@ -94,6 +134,184 @@ inline void inner_kernel(DataT& dat, features::const_audit_iterator& begin, feat
...
+    {
+      // f3 x k*(f2 x k*f1)
+      if (Audit)
+      { audit_func(inner_second_begin.audit() != nullptr ? inner_second_begin.audit() : &EMPTY_AUDIT_STRINGS); }
+      feature_index halfhash = FNV_prime * (halfhash1 ^ inner_second_begin.index());
+      feature_value ft_value = INTERACTION_VALUE(first_ft_value, inner_second_begin.value());
+
+      auto begin = third_begin;
+      // next index differs for permutations and simple combinations
+      if (same_namespace2) { begin += j; }
```
peterychang, line 206: is it possible for `first_begin == second_begin == third_begin`? In this case, would it be possible to iterate begin past `third_end`?

vowpalwabbit/interactions_predict.h
```diff
@@ -94,6 +134,184 @@ inline void inner_kernel(DataT& dat, features::const_audit_iterator& begin, feat
...
+  // preparing state data
+  for (const auto& r : range) { state_data.emplace_back(r.first, r.second); }
+
+  if (!permutations)  // adjust state_data for simple combinations
+  {                   // if permutations mode is disabled then namespaces in ns are already sorted and thus grouped
+    // (in fact, currently they are sorted even for enabled permutations mode)
+    // let's go throw the list and calculate number of features to skip in namespaces which
+    // repeated more than once to generate only simple combinations of features
+
+    for (auto fgd = state_data.data() + (state_data.size() - 1); fgd > state_data.data(); --fgd)
```
peterychang, line 235: can you use `rbegin()` and `rend()` instead of the pointer math? Not sure if its better though since you want to end on the second element instead of the first..

vowpalwabbit/interactions_predict.h
jackgerrits: I'm not changing the logic but they would be equal on the first iteration, and then never again after that

vowpalwabbit/interactions_predict.h
jackgerrits: I'd rather leave this logic as is for now and circle back in a separate PR, but I do really like the idea of using reverse iterators

vowpalwabbit/interactions_predict.h
jackgerrits: I'd prefer to just use what the formatter says. I think this is better after the typedef

peterychang (approved):

PR #3231, opened 2021-08-17
feat: support format strings for error messages

jackgerrits:


olgavrou (approved):

jackgerrits:
Trying to make this work well when slim can't take dependencies is enough for me

PR #3233, opened 2021-08-18
fix: Catch segfaults dense parameters

bassmang:
This catches the two crashes from the 24h_fuzz_2 directory with ids 000048 and 000051


olgavrou (commented):
can we add a test that exercises this crash and fix?

vowpalwabbit/array_parameters_dense.h
```diff
@@ -85,7 +85,11 @@ class dense_parameters
   const_iterator cend() { return const_iterator(_begin + _weight_mask + 1, _begin, stride()); }
 
   inline const weight& operator[](size_t i) const { return _begin[i & _weight_mask]; }
-  inline weight& operator[](size_t i) { return _begin[i & _weight_mask]; }
+  inline weight& operator[](size_t i)
+  {
+    if (_begin == nullptr) { THROW("Error: Dense parameters not initialized."); }
```
jackgerrits, line 90: Putting a throw in this operator could slow things down quite a bit. Consider making it an assert? Can this condition be handled somewhere else in a sort of validation step?

bassmang: Yes there are a few places I can add this check. This is being invoked by this function:
`inline weight& strided_index(size_t index) { return operator[](index << _stride_shift); }`
I think it make sense to assert `not_null()` within the strided_index function. Alternatively I could check `not_null()` at an earlier stage, but that might leave future errors.

jackgerrits: Generally for a function like `operator[]` you define under what conditions it is undefined behavior. We can assert that state in a debug build. I feel like the initialized check should be done once after the model is loaded?

bassmang: Ok yea I changed the check to when the model is loaded

vowpalwabbit/gd.cc
```diff
@@ -1066,7 +1069,11 @@ void save_load(gd& g, io_buf& model_file, bool read, bool text)
       save_load_online_state(all, model_file, read, text, g.total_weight, &g);
     }
     else
+    {
+      if (!all.weights.sparse && !all.weights.dense_weights.not_null())
```
jackgerrits, line 1073: We could call `!all.weights.not_null()` and it will cover both checking for sparse and dense.

jackgerrits (approved):

PR #3235, opened 2021-08-18
refactor: remove not needed comparator

jackgerrits:
The default vector < operator will result in the same behavior as this comparator implemented.

### DIFF
--- a/vowpalwabbit/interactions.cc
+++ b/vowpalwabbit/interactions.cc
@@ -154,21 +154,6 @@ void eval_count_of_generated_ft(bool permutations, const std::vector<std::vector
   }
 }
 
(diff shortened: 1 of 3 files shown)


vowpalwabbit/interactions.cc
```diff
@@ -154,21 +154,6 @@ void eval_count_of_generated_ft(bool permutations, const std::vector<std::vector
   }
 }
 
-bool sort_interactions_comparator(const std::vector<namespace_index>& a, const std::vector<namespace_index>& b)
-{
-  if (a.size() != b.size()) { return a.size() > b.size(); }
```
peterychang, line 159: Is this desired behavior? If so, then we'll need a comparator that does this bit

jackgerrits:
This comparator implements size sorting then contents sorting. So lexicographically cannot be used.

PR #3237, opened 2021-08-19
Revert "fix: fix comparator for interactions sorting"

olgavrou:
Reverts VowpalWabbit/vowpal_wabbit#3236

This was correct as it was, double checked, we always sorted cubics before quadratics (checked out the code before any of the `-q :: speedup` changes and verified (i.e. commit `b29f803f1902553092d777481f703dc4c619662f`)) I do remember testing the comparator extensively to be sure that interaction functionality was not changed


jackgerrits:
Synced offline. This is fine as the model diffs are very small.

PR #3239, opened 2021-08-19
refactor: change interactions to be a list of interaction_terms

jackgerrits:
Next in the series: https://github.com/jackgerrits/vowpal_wabbit/tree/jagerrit/new_interactions_parsing


test/unit_test/interactions_test.cc
```diff
@@ -27,6 +27,28 @@ struct eval_gen_data
   }
 };
```
olgavrou, line 29: based on a comment from `interaction_term.cc` if it is legit to compare char's with full namespace's then can we add a test here for that too?

vowpalwabbit/interactions.h
```diff
@@ -74,8 +77,16 @@ std::vector<std::vector<namespace_index>> compile_interaction(
   }
 
   // Quadratic fast path or generic generation function.
-  auto result = num_wildcards == 2 ? expand_quadratics_wildcard_interactions(leave_duplicate_interactions, indices)
-                                   : generate_func(indices, num_wildcards);
+  auto result_indices = num_wildcards == 2
+      ? expand_quadratics_wildcard_interactions(leave_duplicate_interactions, indices)
+      : generate_func(indices, num_wildcards);
+  std::vector<std::vector<INTERACTIONS::interaction_term>> result;
```
olgavrou, line 83: this is on the hot path right? it might be worth seeing if we can have the `result` vector be a member var who's size gets stabilized over runs and see if that affects the benchmarks

jackgerrits: I wouldn't say this is on the hot path. It's only run when the set of seen namespaces change

vowpalwabbit/interaction_term.h
```diff
@@ -0,0 +1,73 @@
...
+    case interaction_term_type::ns_char:
+      return "ns_char";
+    case interaction_term_type::full_name:
+      return "full_name";
+    default:
+      return "unknown";
+  }
+}
+
+struct interaction_term
```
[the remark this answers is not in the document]
jackgerrits: The keyword of struct/class makes no functional difference

vowpalwabbit/interaction_term.cc
```diff
@@ -0,0 +1,72 @@
...
+  // It doesn't make sense to get the hash for a wildcard.
+  assert(!_wildcard);
+  return _ns_hash;
+}
+
+bool interaction_term::wildcard() const { return _wildcard; }
+
+interaction_term_type interaction_term::type() const { return _type; }
+
+bool operator<(const interaction_term& lhs, const interaction_term& rhs)
```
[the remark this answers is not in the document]
jackgerrits: Yep it is valid and it is the first term in the comparison so they will be grouped separately in the same container.

test/unit_test/interactions_test.cc
jackgerrits: I haven't implemented hash based expansion (next PR). I will add a test which sorts interactions with both character and hash based interactions in that PR.

vowpalwabbit/interaction_term.h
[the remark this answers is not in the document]
olgavrou: it doesn't but it is a more of a social construct that `struct`s are used for data transfer objects without any logic while `class`s are used for encapsulation and business logic

vowpalwabbit/interactions.cc
```diff
@@ -39,12 +40,15 @@ void eval_count_of_generated_ft(bool permutations, const std::vector<std::vector
       size_t num_features_in_inter = 1;
       float sum_feat_sq_in_inter = 1.;
 
-      for (namespace_index ns : inter)
+      for (auto ns : inter)
```
bassmang, line 43:
```suggestion
      for (const auto& ns : inter)
```

vowpalwabbit/interactions_predict.h
```diff
@@ -97,11 +120,12 @@ std::tuple<features_range_t, features_range_t, features_range_t> inline generate
...
 
 std::vector<features_range_t> inline generate_generic_char_combination(
-    const std::array<features, NUM_NAMESPACES>& feature_groups, const std::vector<namespace_index>& namespace_indexes)
+    const std::array<features, NUM_NAMESPACES>& feature_groups,
+    const std::vector<INTERACTIONS::interaction_term>& terms)
 {
   std::vector<features_range_t> inter;
-  for (const auto namespace_index : namespace_indexes)
-  { inter.emplace_back(feature_groups[namespace_index].audit_begin(), feature_groups[namespace_index].audit_end()); }
+  for (const auto term : terms)
```
bassmang, line 127:
```suggestion
  for (const auto& term : terms)
```

bassmang (approved):

vowpalwabbit/interactions.h
olgavrou: should we be concerned that some of the benchmarks are a bit red?

jackgerrits: Hmm, some of them are fine and some are bit red yeah. The ones that were added recently seem to have more variance I've noticed, so I am less concerned about those.

In general, I did expect this PR to have a slight negative effect on performance. Given this is adding a branch into the interactions stuff it seems reasonable. One possible way is to keep both "types" of interactions completely separate and then invoke the interactions loop twice - one for each type. This introduces more complexity elsewhere so I don't know if it would pay off. I haven't implemented it since it will require a fair few changes.

vowpalwabbit/interaction_term.h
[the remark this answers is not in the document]
jackgerrits: I wasn't aware of that distinction. I've always associated `class`/`struct` in c++ as a style choice. The codebase seems to use struct most of the time, so I used that.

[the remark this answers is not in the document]
olgavrou: The main difference is that by default everything in a `struct` is public vs everything in a `class` being private, resulting in natural encapsulation in the `class` case. But up to you

vowpalwabbit/interactions.h
olgavrou: As long as we are aware of where the regressions come from, so that in the end we can focus on those first to optimize should the final solution be too slow, then that's fine by me

jackgerrits: Yeah I 100% agree

olgavrou (approved):

vowpalwabbit/interaction_term.h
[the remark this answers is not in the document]
jackgerrits: I certainly don't feel strongly about it, so I just pushed a change changing it to class :)

jackgerrits:
This has been superseded by #3304 as it is more efficient to manage two lists.

PR #3248, opened 2021-08-20
feat: Adding metric for sum of original_label_cost

peterychang:


vowpalwabbit/vw_exception.h
```diff
@@ -96,6 +96,21 @@ class vw_unrecognised_option_exception : public vw_exception
   ~vw_unrecognised_option_exception() noexcept override = default;
 };
 
+class save_load_model_exception : public vw_exception
```
jackgerrits, line 99: Is this meant to be here?

vowpalwabbit/parse_slates_example_json.h
```diff
@@ -291,6 +292,8 @@ void parse_slates_example_dsjson(vw& all, v_array<example*>& examples, char* lin
       {
         assert(false);
       }
+
+      if (current_obj.HasMember("_original_label_cost")) { data->originalLabelCost = current_obj["_original_label_cost"].GetFloat(); }
```
jackgerrits, line 296: You should check whether it is a float

vowpalwabbit/metrics.cc
```diff
@@ -70,6 +70,8 @@ void list_to_json_file(dsjson_metrics* ds_metrics, std::string filename, metric_
       writer.String(ds_metrics->LastEventId.c_str());
       writer.Key("last_event_time");
       writer.String(ds_metrics->LastEventTime.c_str());
+      writer.Key("cbea_sum_cost_original");
```
lalo, line 73: cbea is the prefix used by cb_explore_adf, in this case this value comes from dsjson, I would rename to dsjon_sum_cost_original?

vowpalwabbit/vw_exception.h
peterychang: no, it is not haha

jackgerrits (approved):

PR #3250, opened 2021-08-23
refactor: use new builder for marginal and other fixes

jackgerrits:


vowpalwabbit/marginal.cc
```diff
@@ -15,82 +17,110 @@ namespace MARGINAL
...
+      , unweighted_marginals(unweighted_marginals)
+      , compete(compete)
+      , m_weights(m_weights)
+      , m_loss_function(m_loss_function)
+      , m_shared_data(m_shared_data)
+  {
+    id_features.fill(false);
+  }
+
+  data(float initial_numerator, float initial_denominator, float decay, bool update_before_learn,
```
olgavrou, line 49: i am guessing `all` will never be `nullptr` right?

olgavrou (approved):

vowpalwabbit/marginal.cc
jackgerrits: It should never be in the way it is currently used. Especially since it is not stored I can change it to a reference

PR #3252, opened 2021-08-23
feat: test

lalo:
- no interactions vs hard-coded runtime interactions
    - save load working
    - basic tests
    - todo: fix parser side


test/unit_test/automl_test.cc
```diff
@@ -10,16 +10,19 @@
 #include <boost/test/test_tools.hpp>
 
 #include "test_common.h"
-#include "metric_sink.h"
 #include "simulator.h"
 #include "reductions_fwd.h"
+#include "test_red.cc"
```
kumpera, line 15: that's.... nasty.

kumpera: A better approach would be to have a test_red_private.h that is used here and don't force us to include the implementation

lalo: this is a draft btw, we're still focused on features and we're cutting corners in plenty of places so beware (see learner.h for more). There's a couple of items to revisit once we're feature/functional complete in terms of the mvp

lalo: that is a good idea, I've seen dirs titled internal for those sort of internal headers that shouldn't be consumed for the public

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,664 @@
...
+      const std::map<namespace_index, size_t>&, const std::set<namespace_index>&) = 0;
+};
+
+struct quadratic_exclusion_oracle
+{
+  // This code is primarily borrowed from expand_quadratics_wildcard_interactions in
+  // interactions.cc. It will generate interactions with -q :: and exclude namespaces
+  // from the corresponding stride. This function can be swapped out depending on
+  // preference of how to generate interactions from a given set of exclusions.
+  interaction_vec gen_interactions(
```
lalo, line 195: transforms exclusions -> interactions expected by VW

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,675 @@
...
+        if (index_queue.empty() && !repopulate_index_queue()) { continue; }
+        config_index = index_queue.top().second;
+        index_queue.pop();
+        scores[stride].config_index = config_index;
+        // Regenerate interactions each time an exclusion is swapped in
+        live_interactions[stride].clear();
+        live_interactions[stride] = oc.gen_interactions(ns_counter, configs[config_index].exclusions);
+        // We may also want to 0 out weights here? Currently keep all same in stride position
+      }
+      configs[config_index].budget--;
```
lalo, line 396: We might not need to do this, if budget is compared with update_count. They might be tracking the same state right now.

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,675 @@
...
+
+  // This function defines the logic update the live configs' budgets, and to swap out
+  // new configs when the budget runs out.
+  void update_live_configs()
+  {
+    for (size_t stride = 0; stride < max_live_configs; ++stride)
+    {
+      int config_index = scores[stride].config_index;
+      // Special case to initialize first champion to -q ::
+      if (config_index == -2)
```
lalo, line 372: This special case can be removed if this logic is moved to the constructor of config_manager. This will allow us to remove the need for this -2 case. It could also be moved to the switch case of idle, when preparing the first experiment.

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,675 @@
...
+        // Regenerate interactions each time an exclusion is swapped in
+        live_interactions[stride].clear();
+        live_interactions[stride] = oc.gen_interactions(ns_counter, configs[config_index].exclusions);
+        // We may also want to 0 out weights here? Currently keep all same in stride position
+      }
+      configs[config_index].budget--;
+    }
+  }
+
+  void update_champ()
```
lalo, line 400: this function might belong on scored_config, or some sort of static compare between scored_configs

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,675 @@
...
+        float priority = get_priority(redo_index);
+        index_queue.push(std::make_pair(priority, redo_index));
+      }
+    }
+    return !index_queue.empty();
+  }
+
+  // This function defines the logic update the live configs' budgets, and to swap out
+  // new configs when the budget runs out.
+  void update_live_configs()
```
lalo, line 366: we should be able to swap or specialize the impl of this function. i.e. figure out a way to extend the budget to the best challenger with zero budget

vowpalwabbit/automl.cc
lalo: We can have a method on scored_config report back how much budget it has left by comparing budget against update_count.

vowpalwabbit/automl.cc
bassmang: I think the function will need to remain in config manager since it updates the champ, but I can move most of the comparing logic to scored_config

lalo: yea maybe lets think about it some more

vowpalwabbit/automl.cc
bassmang: Updated. We should note that this changes the functionality from "swap out after X # of examples as non-champ" -> "swap out if not champ at example #X"

vowpalwabbit/automl.cc
bassmang: This case is fixed by regenerating interactions whenever a new namespace is seen

vowpalwabbit/automl.cc
bassmang: I can move the functionality of handling a config out of budget to a different function, which will make it clearer how to update this logic

test/unit_test/automl_test.cc
lalo: removed .cc include, resolving this convo.

vowpalwabbit/automl.h
```diff
@@ -0,0 +1,148 @@
...
+  virtual void persist(metric_sink&) = 0;
+};
+
+struct config_manager : config_manager_base
+{
+  config_state current_state = Idle;
+  size_t county = 0;
+  size_t current_champ = 0;
+  const size_t budget;
+  const size_t max_live_configs;
```
jackgerrits, line 103: I think const member variables should generally be avoided due to the pitfalls associated (https://lesleylai.info/en/const-and-reference-member-variables/)

vowpalwabbit/automl.h
```diff
@@ -0,0 +1,148 @@
...
+  void gen_exclusion_configs(const multi_ex&);
+  bool repopulate_index_queue();
+  void handle_empty_budget(scored_config&);
+  void update_live_configs();
+};
+
+struct automl
+{
+  config_manager cm;
+  vw* all;                              //  TBD might not be needed
```
jackgerrits, line 137: Should be initialized to nullptr

vowpalwabbit/automl.h
```diff
@@ -0,0 +1,148 @@
...
+  LEARNER::multi_learner* adf_learner;  //  re-use print from cb_explore_adf
```
jackgerrits, line 138: Should be initialized to nullptr

vowpalwabbit/learner.h
```diff
@@ -115,6 +115,17 @@ struct save_metric_data
   fn save_metric_f = nullptr;
 };
 
+// used for automl
+struct multiconfig_data
```
jackgerrits, line 119: What is a multiconfig? Can you please document this function type here and/or [here](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/What-is-a-learner%3F)

vowpalwabbit/learner.h
```diff
@@ -173,6 +186,7 @@ inline void increment_offset(example& ex, const size_t increment, const size_t i
 
 inline void increment_offset(multi_ex& ec_seq, const size_t increment, const size_t i)
 {
+  // std::cerr<<increment*i<<std::endl;
```
jackgerrits, line 189: Remove?

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,743 @@
...
+  all.interactions.clear();
+  assert(all.interactions.empty() == true);
+
+  // make sure we setup the rest of the stack with cleared interactions
+  // to make sure there are not subtle bugs
+  auto* base_learner = stack_builder.setup_base_learner();
+
+  assert(all.interactions.empty() == true);
+
+  assert(all.weights.sparse == false);
```
jackgerrits, line 693: If this invariant is required you'll need to convert it to a failure in release builds

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,743 @@
...
+  if (!options.add_parse_and_check_necessary(new_options)) return nullptr;
+
+  auto data = VW::make_unique<automl>(budget, all.model_file_ver, max_live_configs);
+  // all is not needed but good to have for testing purposes
+  data->all = &all;
+  assert(max_live_configs <= MAX_CONFIGS);
+
+  // override and clear all the global interactions
+  // see parser.cc line 740
+  all.interactions.clear();
```
jackgerrits, line 684: You should raise an error if this is non-empty as that would be surprising behavior

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,743 @@
...
+      size_t other_ns_size = ns_pair.second.size();
+      bin_text_read_write_fixed(model_file, (char*)&ns, sizeof(ns), "", read, msg, text);
+      bin_text_read_write_fixed(
+          model_file, reinterpret_cast<char*>(&other_ns_size), sizeof(other_ns_size), "", read, msg, text);
+      for (const auto& other_ns : ns_pair.second)
+      { bin_text_read_write_fixed(model_file, (char*)&other_ns, sizeof(other_ns), "", read, msg, text); }
+    }
+  }
+
+  if (!read || true)
```
jackgerrits, line 184: This will always evaluate to true

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,743 @@
...
+    {
+      auto idx2 = (*jt).first;
+      // Only handles ':' for now, extend to handle specific namespaces later as needed
+      if (exclusions.find(idx2) != exclusions.end() && exclusions.at(idx2).find(':') != exclusions.at(idx2).end())
+      { continue; }
+      interactions.insert({idx1, idx2});
+      interactions.insert({idx2, idx2});
+    }
+  }
+  interaction_vec v = interaction_vec(interactions.begin(), interactions.end());
```
jackgerrits, line 219: To avoid some copies here you could move the results:
```c++
return interaction_vec(std::make_move_iterator(interactions.begin()), std::make_move_iterator(interactions.end()));
```

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,743 @@
...
+  metrics.int_metrics_list.emplace_back("current_champ", current_champ);
+  for (size_t stride = 0; stride < scores.size(); ++stride)
+  { scores[stride].persist(metrics, "_" + std::to_string(stride)); }
+}
+
+// This sets up example with correct ineractions vector
+void config_manager::apply_config(example* ec, size_t stride)
+{
+  if (ec == nullptr) { return; }
+  if (stride < max_live_configs) { ec->interactions = &(scores[stride].live_interactions); }
```
jackgerrits, line 531: Are wildcard interactions ever used? I imagine not?

How does this reduction interact with CCB? Since it modifies the interactions state?

jackgerrits (changes requested):

vowpalwabbit/automl.cc
lalo: I was not considering CCB compat for MVP. Right now I'm failing on setup if ccb is enabled.

lalo: We do have a pending task to audit all those reductions flagged as possibly modifying interaction state.

vowpalwabbit/array_parameters_dense.h
```diff
@@ -114,6 +114,64 @@ class dense_parameters
...
 
+  void swap_offsets(size_t offset_1, size_t offset_2, size_t params_per_problem)
+  {
+    float temp = 0.f;
+    assert(offset_1 < offset_2);
+    for (iterator iter = begin(); iter != end(); ++iter)
+    {
+      // assert(*iter == _begin[(count << stride_shift()) & _weight_mask]);
+      size_t actual_index = iter.index() >> stride_shift();
+      size_t actual_type = actual_index & (params_per_problem - 1);
```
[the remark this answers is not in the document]
lalo: you are right, misnomer. It represents the offset embedded in the last bits of that index, so maybe a better name would be current_offset and rename actual_index to prestride_index

vowpalwabbit/array_parameters_dense.h
```diff
@@ -114,6 +114,64 @@ class dense_parameters
     for (iterator iter = begin(); iter != end(); ++iter) (&(*iter))[offset] = 0;
   }
 
+  void swap_offsets(size_t offset_1, size_t offset_2, size_t params_per_problem)
```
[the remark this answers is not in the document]
lalo: I believe its not equivalent, but I need to double check that weights currently has a copy of ppp. I'll add a comment. Although this method has been deleted, but I'll add it to the others.

vowpalwabbit/array_parameters_dense.h
```diff
@@ -114,6 +114,64 @@ class dense_parameters
...
+      size_t actual_type = actual_index & (params_per_problem - 1);
+      if (actual_type == offset_1)
+      {
+        float* other = &_begin[(actual_index + (offset_2 - offset_1)) << stride_shift()];
+
+        if (*other != 0.f || *iter != 0.f)
+        {
+          temp = *other;
+          (&(*other))[0] = (&(*iter))[0];
+          (&(*iter))[0] = temp;
```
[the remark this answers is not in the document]
lalo: not relevant anymore

vowpalwabbit/array_parameters_dense.h
```diff
@@ -114,6 +114,64 @@ class dense_parameters
...
   }
 
+  void swap_offsets(size_t offset_1, size_t offset_2, size_t params_per_problem)
+  {
+    float temp = 0.f;
+    assert(offset_1 < offset_2);
+    for (iterator iter = begin(); iter != end(); ++iter)
+    {
+      // assert(*iter == _begin[(count << stride_shift()) & _weight_mask]);
+      size_t actual_index = iter.index() >> stride_shift();
```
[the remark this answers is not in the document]
lalo: I documented with a test.

vowpalwabbit/array_parameters_dense.h
```diff
@@ -114,6 +114,64 @@ class dense_parameters
...
+          (&(*iter))[0] = temp;
+        }
+      }
+    }
+  }
+
+  void copy_offsets(size_t from, size_t to, size_t params_per_problem)
+  {
+    std::int64_t diff = to - from;
+    // assert(from < to);
```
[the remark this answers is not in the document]
lalo: they are stale, deleted actually.

vowpalwabbit/array_parameters_dense.h
```diff
@@ -114,6 +114,64 @@ class dense_parameters
...
+    for (iterator iter = begin(); iter != end(); ++iter)
+    {
+      // assert(*iter == _begin[(count << stride_shift()) & _weight_mask]);
+      size_t actual_index = iter.index() >> stride_shift();
+      size_t actual_type = actual_index & (params_per_problem - 1);
+      if (actual_type == from)
+      {
+        float* other = &_begin[(actual_index + diff) << stride_shift()];
```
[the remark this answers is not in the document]
lalo: cop-out answer but correct usage of the api, in theory both numbers have to be less than all.wpp, I guess we can add an assert?

vowpalwabbit/array_parameters_dense.h
```diff
@@ -114,6 +114,64 @@ class dense_parameters
...
+
+        if (*other != 0.f || *iter != 0.f) { (&(*other))[0] = (&(*iter))[0]; }
```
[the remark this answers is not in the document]
lalo: should I update the whole file to keep it 'themed'?

vowpalwabbit/array_parameters_dense.h
```diff
@@ -114,6 +114,64 @@ class dense_parameters
...
+
+  // void copy_offsets(size_t from, size_t to, size_t width = 1, size_t params_per_problem = 1)
+  // {
+  //   for (iterator iter = begin(); iter != end(); ++iter)
+  //   {
+  //     for (size_t j = 0; j < width; j++) { (&(*iter))[to + j] = (&(*iter))[from + j]; }
+  //   }
+  // }
+
+  void clear_offset(size_t offset, size_t params_per_problem)
```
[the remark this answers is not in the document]
lalo: yes, that's what this type of low level api is trying to enable. The complicated part is also handling from the learner stack and all the reductions, for that see learner.h diff. The operation is needed, we need to apply a mask to figure out which offset that index belongs to + the other 3 floats gd.cc expects (offset 1-3)

vowpalwabbit/array_parameters_dense.h
```diff
@@ -114,6 +114,64 @@ class dense_parameters
...
+  {
+    for (iterator iter = begin(); iter != end(); ++iter)
+    {
+      size_t actual = (iter.index() >> stride_shift()) & (params_per_problem - 1);
+      if (actual == offset && *iter != 0.f) { (&(*iter))[0] = 0; }
```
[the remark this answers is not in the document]
lalo: it's going to have to do some extra work with offsets +1,2,3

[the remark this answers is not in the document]
lalo: or with 1 << stride_shift() amount of offsets

vowpalwabbit/automl.h
```diff
@@ -0,0 +1,148 @@
...
+{
+using namespace_index = unsigned char;
+using interaction_vec = std::vector<std::vector<namespace_index>>;
+
+VW::LEARNER::base_learner* automl_setup(VW::setup_base_i&);
+
+namespace helper
+{
+void fail_if_enabled(vw&, std::string);
+bool cmpf(float, float, float);
```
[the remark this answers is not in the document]
lalo: thanks

vowpalwabbit/learner.h
lalo:
```suggestion
```

vowpalwabbit/automl.cc
lalo: I think at this point it already has the parsed interactions of the cmd line populated here, but now you are making me doubt.

vowpalwabbit/automl.cc
lalo: Right now we only have a dense impl, so most likely yes.

vowpalwabbit/automl.h
bassmang: Cool didnt know that

vowpalwabbit/array_parameters_dense.h
```diff
@@ -114,6 +114,64 @@ class dense_parameters
     for (iterator iter = begin(); iter != end(); ++iter) (&(*iter))[offset] = 0;
   }
 
+  void swap_offsets(size_t offset_1, size_t offset_2, size_t params_per_problem)
+  {
+    float temp = 0.f;
+    assert(offset_1 < offset_2);
+    for (iterator iter = begin(); iter != end(); ++iter)
```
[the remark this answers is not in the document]
lalo: resolving bc outdated

vowpalwabbit/automl.h
```diff
@@ -0,0 +1,148 @@
...
+  std::map<namespace_index, std::set<namespace_index>> exclusions;
+  size_t budget;
+
+  exclusion_config(size_t budget = 10) : budget(budget) {}
+
+  void save_load_exclusion_config(io_buf&, bool, bool);
+};
+
+// all possible states of config_manager
+enum config_state
```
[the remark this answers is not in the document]
bassmang: This may overcomplicated things with save_load

vowpalwabbit/automl.cc
bassmang: That was a placeholder for the versions check. Thats now made in the main automl save_load so this `if` can be removed

vowpalwabbit/automl.cc
bassmang: I think it might be even to here to just pass in the vector

vowpalwabbit/learner.h
lalo: it's a proposal that goes hand in hand with automl reduction but there isn't a concrete implementation of this functionality right now. The goal is to have a way to signal the stack to copy 'state' specific to offsets that the reduction has, permitting for a single stack to keep track of multiple 'states'. In this case an upper reduction can signal the lower stack to clear offset Foo or copy the values from offset Bar. These offsets should match the offsets the reduction is receiving when being called on learn/predict, which would be the way for this theoretical reduction to keep some state per different offset received (if needed).

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,834 @@
...
+            break;
+          }
+        }
+        champ_ips = configs[config_index].ips;
+        champ_lower_bound = configs[config_index].lower_bound;
+      }
+      else
+      {
+        configs[scores[current_champ].config_index].budget *= 2;
+        configs[scores[current_champ].config_index].state = Inactive;
```
lalo, line 470: I suspect this is wrong, if the champ gets over-passed by a single config one that shouldn't kick it that champ out, that should just promote the other one to be active. The hard thing here is that now you have to kick out the worst one.

lalo: althought the champ might get pushed back in on the next go bc it'll have a high ips probably, but haven't checked that logic

bassmang: Yes this whole `else` branch is the case when an inactive config takes over the champ. I could kick out the worst performing live config instead. That (and handling the index queue more efficiently) are both possible but seem unnecessary since this case really shouldn't happen.

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,834 @@
...
+    if (removed) { index_queue.pop(); }
+  }
+  // TODO: Add logic to erase index from configs map if wanted
+  if (index_queue.empty() && !repopulate_index_queue()) { return; }
+  score.reset_stats();
+  configs[config_index].state = Inactive;
+  config_index = index_queue.top().second;
+  index_queue.pop();
+  score.config_index = config_index;
+  configs[config_index].state = Live;
```
lalo, line 391: somewhere around here we copy the champs weights or clear/random

vowpalwabbit/automl.cc
lalo: synced offline: lower priority since this case might be impossible to hit or uncommon

bassmang: solved the issue in latest commit

test/unit_test/automl_weights_test.cc
```diff
@@ -0,0 +1,168 @@
...
+}  // namespace vw_hash_helpers
+
+using namespace vw_hash_helpers;
+
+// 0) runs after 1331 learned examples from simulator.h
+// 1) clears (set to zero) offset 1
+// 2) copies offset 2 -> offset 1
+// n) asserts weights before/after every operation
+// NOTE: interactions are currently 0 for offset 0 since
+// config 0 is hard-coded to be empty interactions for now.
```
lalo, line 62: not true anymore, reminder to re-read comments and update

vowpalwabbit/automl.cc
```diff
@@ -494,16 +503,12 @@ void config_manager::update_champ()
     }
 
     // If challenger is worse ('worse function from Chacha')
-    if (configs[config_index].ips < champ_lower_bound)
-    {
-      // TODO: Update worse function, this seems way too strict (removes configs instantly)
-      // configs[config_index].state = Removed;
-    }
+    if (worse(configs[config_index], champ_config)) { configs[config_index].state = Removed; }
```
lalo, line 506: should this be an else if?

bassmang: Both cases should never be true, but yes else if looks better

vowpalwabbit/array_parameters_dense.h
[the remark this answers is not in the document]
jackgerrits: It would be better if the documentation could reside in the code too. A test is better than nothing but wont always be found when looking at the code

vowpalwabbit/array_parameters_dense.h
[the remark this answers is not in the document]
jackgerrits: If its assertable we should totally add it

vowpalwabbit/array_parameters_dense.h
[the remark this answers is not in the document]
jackgerrits: Yeah I would vote for updating it. But it is a bit of personal preference

vowpalwabbit/learner.h
jackgerrits: Is the registering of multiple states done at setup time or is it dynamic?

vowpalwabbit/array_parameters_dense.h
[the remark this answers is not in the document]
jackgerrits: Yeah I've always thought the fact that *iter returns a reference to a single float but we always just turn it back into a pointer and index into it just mega dangerous and icky. Ideally this could be span so the caller of this can't misuse it.

You're using it now thought - so if you're using the rest of the stride then no prob. Feel free to resolve.

vowpalwabbit/automl.h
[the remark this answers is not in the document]
jackgerrits: How so? The enum class is backed by whatever type you want. Make it some known type and cast to and from it.

python/tests/test_tutorial.py
```diff
@@ -14,11 +16,14 @@ class Simulator:
 
     users = ['Tom', 'Anna']
     times_of_day = ['morning', 'afternoon']
-    actions = ["politics", "sports", "music", "food", "finance", "health", "camping"]
+    # actions = ["politics", "sports", "music", "food", "finance", "health", "camping"]
```
olgavrou, line 19: nit: remove commented out code

vowpalwabbit/array_parameters_dense.h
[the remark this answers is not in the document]
lalo: renamed, resolving for now. feel free to open another one on the latest diff.

python/tests/test_tutorial.py
```diff
@@ -14,11 +16,14 @@ class Simulator:
...
+    actions = ["politics", "sports", "music"]
 
-    def __init__(self, debug_logfile=None, seed=10):
+    def __init__(self, debug_logfile=None, seed=10, has_automl=False):
         random.seed(seed)
 
+        self.has_aml = has_automl
```
olgavrou, line 25: nit:
```suggestion
        self.has_automl = has_automl
```

python/tests/test_tutorial.py
```diff
@@ -107,27 +150,36 @@ def run_simulation(self, vw, num_iterations, users, times_of_day, actions, cost_
             # We negate this so that on the plot instead of minimizing cost, we are maximizing reward
             self.ctr.append(-1*self.cost_sum/i)
 
+            if self.has_aml and i % 1 == 0:
```
olgavrou, line 153: isn't ` i % 1` always zero?

lalo: yes it is. I have to remove it but it was useful for debugging at some point. This file is less relevant though maybe spend more time on the cpp unit tests.

test/train-sets/ref/help.stdout
```diff
@@ -200,6 +200,12 @@ Audit Regressor:
                         this mode.
 Autolink:
   --autolink arg        create link function with polynomial d
+Debug: automl reduction:
+  --automl arg (=3, )                set number of live configs
+  --global_lease arg (=10, )         set initial lease for automl interactions
+  --cm_type arg (=interaction, )     set type of config manager
+  --priority_type arg (=none, )      set type of config manager in {none, le}
```
olgavrou, line 207: whats `le`?

olgavrou: also both arguments `cm_type` and `priority_type` have a description of set type of config manager, is that by mistake?

vowpalwabbit/array_parameters_dense.h
```diff
@@ -114,6 +114,33 @@ class dense_parameters
     for (iterator iter = begin(); iter != end(); ++iter) (&(*iter))[offset] = 0;
   }
 
+  void copy_offsets(const size_t from, const size_t to, const size_t params_per_problem)
+  {
+    std::int64_t diff = to - from;
```
olgavrou, line 119:
```suggestion
    int64_t diff = to - from;
```

test/RunTests
```diff
@@ -2503,4 +2503,10 @@ train-sets/ref/slates_simple_w_interactions.stderr
 ./negative-test.sh {VW} --no_stdin --cb_explore_adf --cover 3 --noconstant --quiet -i model-sets/corrupted_weights_gd_mf.model
     train-sets/ref/corrupt_weights_gd_mf.stderr
 
+# Test 351: should be the same output as 158
+# except the line of enabled_reductions
+{VW} --save_resume -f automl.model --automl 3 --priority_type le -d train-sets/decisionservice.json --dsjson --cb_explore_adf --epsilon 0.2 -P 1 -p cbe_adf_dsjson.predict --extra_metrics metrics_.json
```
rajan-chari, line 2508: missing decisionservice.json

test/train-sets/ref/help.stdout
bassmang: `le` (least-exclusion) is a different priority algorithm to decide which config is swapped in next when one runs out of lease.

bassmang: And yes that's a type with the descriptions

test/RunTests
bassmang: I believe that file exists

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,917 @@
...
+{
+  switch (current_state)
+  {
+    case Idle:
+      process_namespaces(ec);
+      exclusion_configs_oracle();
+      current_state = Experimenting;
+      break;
+
+    case Collecting:
```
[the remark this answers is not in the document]
bassmang: This is still a placeholder for now. In the current implementation, we begin in `Idle` mode for the first example, and this example is used to collect information about the namespaces used, which is used by the `oracle` to generate new configs. In the future we may want to go through the first X examples in `Collecting` mode in order to give the oracle a better idea of what namespaces are being used before generating new configs. Another idea here is to use `--passes` to pass through all the data once (in batch mode), in order to collect all information about the namespaces before generating configs and learning.

test/train-sets/ref/help.stdout
olgavrou: can we add the long name for `le` in the help?

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,917 @@
...
+  configs[0] = conf;
+  scored_config sc;
+  scores.push_back(sc);
+}
+
+void interaction_config_manager::one_step(const multi_ex& ec)
+{
+  switch (current_state)
+  {
+    case Idle:
```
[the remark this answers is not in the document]
bassmang: This mode is intended for use if we want to turn off any collecting or learning. I think it makes more sense to begin in `Collecting` mode and transition `Experimenting` after the first example for now, and keep `Idle` as a placeholder.

test/train-sets/ref/help.stdout
bassmang: Sure

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,917 @@
...
+Predict/Incur Loss -> offset_learn(): Updates weights and learns on live configs
+Better -> better(): Changes champ if challenger is better
+Worse -> worse(): Removes challenger is much worse than champ
+*/
+
+namespace VW
+{
+namespace automl
+{
+namespace details
```
olgavrou, line 53: `details` as in implementation details? Can this namesepace be more specific? e.g. `internals` or something

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,917 @@
...
+// fail if incompatible reductions got setup
+// todo: audit if they reference global all interactions
+void fail_if_enabled(vw& all, const std::set<std::string>& not_compat)
+{
+  std::vector<std::string> enabled_reductions;
+  if (all.l != nullptr) all.l->get_enabled_reductions(enabled_reductions);
+
+  for (const auto& reduction : enabled_reductions)
+  {
+    if (not_compat.count(reduction) > 0) THROW("automl.cc: plz no bad stack" + reduction);
```
olgavrou, line 64: can this error message be more descriptive?

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,917 @@
...
+  metrics.int_metrics_list.emplace_back("upcnt" + suffix, update_count);
+  metrics.float_metrics_list.emplace_back("ips" + suffix, current_ips());
+  distributionally_robust::ScoredDual sd = chisq.recompute_duals();
+  metrics.float_metrics_list.emplace_back("bound" + suffix, static_cast<float>(sd.first));
+  metrics.float_metrics_list.emplace_back("w" + suffix, last_w);
+  metrics.float_metrics_list.emplace_back("r" + suffix, last_r);
+  metrics.int_metrics_list.emplace_back("conf_idx" + suffix, config_index);
+}
+
+float scored_config::current_ips() const { return ips / update_count; }
```
olgavrou, line 159: check update_count not zero?

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,917 @@
...
+interaction_config_manager::interaction_config_manager(size_t global_lease, size_t max_live_configs, uint64_t seed,
+    size_t priority_challengers,
+    float (*calc_priority)(const exclusion_config&, const std::map<namespace_index, size_t>&))
+    : global_lease(global_lease)
+    , max_live_configs(max_live_configs)
+    , seed(seed)
+    , priority_challengers(priority_challengers)
+    , calc_priority(calc_priority)
+{
+  random_state.set_random_state(seed);
```
olgavrou, line 240: does the random_state also need to be save_resumed?

vowpalwabbit/automl.cc
bassmang: This was mentioned in a previous comment and we changed it from `helper` to `details` for consistency with other reductions. The scope of this namespace is only the `fail_if_enabled` function, which should be removed eventually, so this point may be moot.

vowpalwabbit/automl.cc
bassmang: updated to `THROW("Error: automl does not yet support this reduction: " + reduction);`

vowpalwabbit/automl.cc
bassmang: Good catch

vowpalwabbit/automl.cc
bassmang: Yes it does. The random seed is used to generate new configs. We want to ensure that the next generated configs will be the same regardless of whether we saved/loaded in the middle

vowpalwabbit/automl.cc
olgavrou: ok

vowpalwabbit/automl.h
```diff
@@ -0,0 +1,154 @@
...
+
+struct scored_config
+{
+  VW::distributionally_robust::ChiSquared chisq;
+  float ips = 0.0;
+  float last_w = 0.0;
+  float last_r = 0.0;
+  size_t update_count = 0;
+  size_t config_index = 0;
+  bool regular = false;
```
olgavrou, line 42: regular as in non-champ?

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,917 @@
...
+        scores[worst_stride].reset_stats();
+        scores[worst_stride].config_index = config_index;
+        configs[scores[worst_stride].config_index].state = Live;
+        gen_quadratic_interactions(worst_stride);
+        if (scores[worst_stride].regular)
+        {
+          scores[worst_stride].regular = false;
+          scores[current_champ].regular = true;
+        }
+        current_champ = worst_stride;
```
olgavrou, line 521: why does the current champ get set to the worst stride? seems counter intuitive

vowpalwabbit/automl.h
bassmang: This is a newer idea which Paul suggested as an implementation of the `median` function in Algorithm 2 of Chacha. Basically we set half (or some number with the `--priority_challengers` flag) of the live configs as `regular`, and the other half as `priority`. The champ is a priority config, but is used for predictions while the other configs are not. When a priority config runs out of lease, its lease it doubled and it's able to keep running. When a regular config runs out of lease, we first compare it's `lower_bound` to each of the priority config's `ips`. If the lower bound is higher, we'll swap the priorities (priority becomes regular, regular becomes priority), then double its lease and keep running. If the `lower_bound` of the regular config can't beat the `ips` of any priority configs, then it is swapped out for some new config on the `index_queue`. The idea is that some live slots should be used for priority challengers to have more time to beat the champ, while others are used to swap out and try new configs more consistently.

vowpalwabbit/automl.cc
bassmang: This is a very awkward case which Chacha specifies but might not make sense in a VW context. In `update_champ()`, we compare the champ against all configs (live and non-live). This is the case where a non-live config overtakes the champ (with some lower_bound from a past run). Instead of replacing the champ with the non-live config that overtook it, we find the worst live config, and replace that instead. So `worst_stride` specifies the worst live config, we swap that out with the previously non-live config which is now champ, then we set `current_champ = worst_stride` since that's now the index with the champ.

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,917 @@
...
+
+  CB::cb_class logged{};
+  size_t labelled_action = 0;
+  if (is_learn)
+  {
+    const auto it = std::find_if(ec.begin(), ec.end(), [](example* item) { return !item->l.cb.costs.empty(); });
+
+    if (it != ec.end())
+    {
+      logged = (*it)->l.cb.costs[0];
```
olgavrou, line 737: is this correct? if `cb.costs.empty()` is true that means that there maybe is a label but it could just be the shared label here, no?

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,917 @@
...
+// be the config with the least exclusion. Note that all configs will run to lease
+// before priorities and lease are reset.
+float calc_priority_least_exclusion(const exclusion_config& config, const std::map<namespace_index, size_t>& ns_counter)
+{
+  float priority = 0.f;
+  for (const auto& ns_pair : config.exclusions) { priority -= ns_counter.at(ns_pair.first); }
+  return priority;
+}
+
+// Same as above, returns 0 (includes rest to remove unused variable warning)
```
olgavrou, line 805: you can do `(void)variable;` to remove the unused var warning

olgavrou (approved):
I have reservations about using `stride` as var name in the `automl.cc` file because it is overloading the weights stride, but other than that looks good to me :)

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,917 @@
...
+{
+  random_state.set_random_state(seed);
+  exclusion_config conf(global_lease);
+  conf.state = Live;
+  configs[0] = conf;
+  scored_config sc;
+  scores.push_back(sc);
+}
+
+void interaction_config_manager::one_step(const multi_ex& ec)
```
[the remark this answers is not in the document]
olgavrou: looks like `config_manager` is super generic and so it can be open to many different automl implementations, no?

vowpalwabbit/array_parameters_dense.h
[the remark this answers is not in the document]
lalo: this will be addressed in documentation

vowpalwabbit/array_parameters_dense.h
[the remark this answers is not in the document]
lalo: documentation will include this

vowpalwabbit/array_parameters_dense.h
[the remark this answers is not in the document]
lalo: we can think about it once we have spans, or let me know if you come up with something that fits.

vowpalwabbit/automl.cc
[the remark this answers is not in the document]
bassmang: @rajan-chari I moved one_step() to the automl struct and put schedule(), config_oracle() etc in the config_manager base class. There are a few ways we could go about swapping out these functions. Since their inputs/outputs may vary based on config_manager type, algorithm used, etc... Im not certain it makes sense to swap in new function pointers (like what is done with calc_priority_empty/calc_priority_least_exclusion). Instead we could keep these functions generic (void with no inputs), and their behavior could change within a given implementation (with some if/else), based on some command-line argument.

vowpalwabbit/learner.h
lalo: it should be done at setup, yet its something that right now the lower reductions don't have any knowledge of. We could figure out a way to communicate this, but we don't have a lower reduction right now that is aware of this 'offset' specific state. This would be the next steps of this PR, I assume.

vowpalwabbit/automl.cc
lalo: added throw when sparse

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,930 @@
...
+  float priority = 0.f;
+  for (const auto& ns_pair : config.exclusions) { priority -= ns_counter.at(ns_pair.first); }
+  return priority;
+}
+
+// Same as above, returns 0 (includes rest to remove unused variable warning)
+float calc_priority_empty(const exclusion_config& config, const std::map<namespace_index, size_t>& ns_counter)
+{
+  (void)config;
+  (void)ns_counter;
```
jackgerrits, line 824:
```suggestion
  _UNUSED(config);
  _UNUSED(ns_counter);
```

vowpalwabbit/automl.h
```diff
@@ -0,0 +1,161 @@
...
+
+  void update(float w, float r);
+  void save_load_scored_config(io_buf&, bool, bool);
+  void persist(metric_sink&, const std::string&);
+  float current_ips() const;
+  void reset_stats();
+};
+
+// all possible states of exclusion config
+enum config_state
```
jackgerrits, line 55: This should be an enum class

vowpalwabbit/automl.h
```diff
@@ -0,0 +1,161 @@
...
+  float lower_bound = std::numeric_limits<float>::infinity();
+  config_state state = New;
+
+  exclusion_config(size_t lease = 10) : lease(lease) {}
+
+  void save_load_exclusion_config(io_buf&, bool, bool);
+};
+
+// all possible states of config_manager
+enum config_manager_state
```
jackgerrits, line 77: `enum class`

vowpalwabbit/learner.h
```diff
@@ -895,6 +925,22 @@ struct common_learner_builder
     _learner->persist_metrics_fd.base = _learner->learn_fd.base;
     return *static_cast<FluentBuilderT*>(this);
   }
+
+  FluentBuilderT& set_copy(void (*fn_ptr)(DataT&, size_t& from, size_t& to))
```
jackgerrits, line 929: this name is super vague. Can you please update this https://github.com/VowpalWabbit/vowpal_wabbit/wiki/What-is-a-learner%3F ?

vowpalwabbit/learner.h
```diff
@@ -895,6 +925,22 @@ struct common_learner_builder
...
+  {
+    _learner->copy_clear_fd.copy_notify_f = (multiconfig_data::fn_copy)fn_ptr;
+    _learner->copy_clear_fd.data = _learner->learn_fd.data;
+    _learner->copy_clear_fd.base = _learner->learn_fd.base;
+    return *static_cast<FluentBuilderT*>(this);
+  }
+
+  FluentBuilderT& set_reset_zero(void (*fn_ptr)(DataT&, size_t&))
```
jackgerrits, line 937: Is reset_zero the same as clear offset?

vowpalwabbit/learner.h
```diff
@@ -252,15 +265,16 @@ struct learner
...
   func_data finisher_fd;
   std::string name;  // Name of the reduction.  Used in VW_DBG to trace nested learn() and predict() calls
 
-  std::shared_ptr<void> learner_data;
 
   learner(){};  // Should only be able to construct a learner through
                 // init_learner function
 
 public:
+  std::shared_ptr<void> learner_data;  // oops for now
```
jackgerrits, line 277: Excuse me, why is this now public with a comment like that?

vowpalwabbit/automl.h
```diff
@@ -0,0 +1,161 @@
...
+  void config_oracle();
+  void process_namespaces(const multi_ex&);
+  void schedule();
+  void update_champ();
+
+private:
+  void gen_quadratic_interactions(size_t);
+  bool better(const exclusion_config&, const exclusion_config&) const;
+  bool worse(const exclusion_config&, const exclusion_config&) const;
+  float (*calc_priority)(const exclusion_config&, const std::map<namespace_index, size_t>&);
```
jackgerrits, line 143: Use a `using` alias so this is more readable

vowpalwabbit/learner.h
```diff
@@ -444,6 +458,20 @@ struct learner
     persist_metrics_fd.base = learn_fd.base;
   }
 
+  // non-recursive
+  void copy_offset_based(size_t from, size_t to)
```
jackgerrits, line 462: What does based mean?

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,930 @@
...
+
+    auto* l = make_reduction_learner(std::move(data), as_multiline(base_learner),
+        learn_automl<interaction_config_manager, true>, predict_automl<interaction_config_manager, true>,
+        stack_builder.get_setupfn_name(automl_setup))
+                  .set_params_per_weight(ppw)  // refactor pm
+                  .set_finish_example(finish_example<interaction_config_manager>)
+                  .set_save_load(save_load_aml<interaction_config_manager>)
+                  .set_persist_metrics(persist<interaction_config_manager>)
+                  .set_prediction_type(base_learner->pred_type)
+                  .set_label_type(label_type_t::cb)
```
jackgerrits, line 916: Is this a guarantee?

vowpalwabbit/learner.h
```diff
@@ -444,6 +458,20 @@ struct learner
     persist_metrics_fd.base = learn_fd.base;
   }
 
+  // non-recursive
```
jackgerrits, line 461: Please add a doc comment

vowpalwabbit/learner.h
```diff
@@ -444,6 +458,20 @@ struct learner
...
+  void copy_offset_based(size_t from, size_t to)
+  {
+    // apply increment to match learn(ex, i) behaviour
+    copy_clear_fd.copy_notify_f(copy_clear_fd.data, *copy_clear_fd.base, increment * from, increment * to);
+  }
+
+  // non-recursive
```
jackgerrits, line 468: please add a doc comment

jackgerrits (changes requested):
The changes made to `learner.h` require some naming improvement and further documentation for me to be satisfied merging that in there

vowpalwabbit/array_parameters_dense.h
```diff
@@ -114,6 +114,49 @@ class dense_parameters
     for (iterator iter = begin(); iter != end(); ++iter) (&(*iter))[offset] = 0;
   }
 
+  void copy_offsets(const size_t from, const size_t to, const size_t params_per_problem)
```
jackgerrits, line 117: Generally speaking we don't use const for value args across the codebase so I'd prefer without, but if you feel strongly thats okay

lalo: I wanted to communicate that those vars don't get used as buffers or mutate around in this function, since the fn has already a lot going on.

vowpalwabbit/learner.h
lalo: a reduction can choose to call to clear offset on the weights, the point of this function is a way for an upper reduction like automl to signal a reduction below to reset anything that has to do with a given offset.

vowpalwabbit/learner.h
lalo: The point of this function is a way for an upper reduction like automl to signal a reduction below to copy the state from one offset to another. I'll document it.

vowpalwabbit/learner.h
jackgerrits: I was mainly asking why the naming is different between the setter and how it is called:
`set_reset_zero` vs `clear_offset_based`

vowpalwabbit/learner.h
jackgerrits: Are `from` and `to` in_out parameters? Can you also rename then to something like `from_offset`

vowpalwabbit/automl.cc
bassmang: This should probably be taken out. As of now automl only support `cb`, but should have a greater scope in the future.

jackgerrits (approved):

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,876 @@
...
+}
+
+// this is the registered learn function for this reduction
+// mostly uses config_manager and actual_learn(..)
+template <typename CMType, bool is_explore>
+void learn_automl(automl<CMType>& data, multi_learner& base, multi_ex& ec)
+{
+  assert(data.all->weights.sparse == false);
+
+  bool is_learn = true;
```
rajan-chari, line 700: This can be removed since is_learn is always true.

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,876 @@
...
+{
+  for (size_t live_slot = 0; live_slot < cm->scores.size(); ++live_slot)
+  {
+    for (example* ex : ec) { cm->apply_config(ex, live_slot); }
+
+    auto restore_guard = VW::scope_exit([this, &ec, &live_slot] {
+      for (example* ex : ec) { this->cm->revert_config(ex); }
+    });
+
+    if (!base.learn_returns_prediction) { base.predict(ec, live_slot); }
```
rajan-chari, line 668: Nice!

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,876 @@
...
+
+// inner loop of learn driven by # MAX_CONFIGS
+template <typename CMType>
+void automl<CMType>::offset_learn(multi_learner& base, multi_ex& ec, CB::cb_class& logged, size_t labelled_action)
+{
+  for (size_t live_slot = 0; live_slot < cm->scores.size(); ++live_slot)
+  {
+    for (example* ex : ec) { cm->apply_config(ex, live_slot); }
+
+    auto restore_guard = VW::scope_exit([this, &ec, &live_slot] {
```
rajan-chari, line 664: is live_slot needed for this lambda?

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,876 @@
...
+
+void interaction_config_manager::revert_config(example* ec) { ec->interactions = nullptr; }
+
+template <typename CMType, bool is_explore>
+void predict_automl(automl<CMType>& data, multi_learner& base, multi_ex& ec)
+{
+  size_t champ_live_slot = data.cm->current_champ;
+  for (example* ex : ec) { data.cm->apply_config(ex, champ_live_slot); }
+
+  auto restore_guard = VW::scope_exit([&data, &ec, &champ_live_slot] {
```
rajan-chari, line 649: champ_live_slot needed for this lambda?

vowpalwabbit/automl.h
```diff
@@ -0,0 +1,173 @@
...
+  bool repopulate_index_queue();
+  bool swap_eligible_to_inactivate(size_t);
+};
+
+template <typename CMType>
+struct automl
+{
+  automl_state current_state = automl_state::Collecting;
+  std::unique_ptr<CMType> cm;
+  vw* all = nullptr;                              //  TBD might not be needed
```
rajan-chari, line 163: Remove?

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,876 @@
...
+
+  details::fail_if_enabled(all,
+      {"ccb_explore_adf", "audit_regressor", "baseline", "cb_explore_adf_rnd", "cb_to_cb_adf", "cbify", "replay_c",
+          "replay_b", "replay_m", "memory_tree", "new_mf", "nn", "stage_poly"});
+
+  // only this has been tested
+  if (base_learner->is_multiline)
+  {
+    // fetch cb_explore_adf to call directly into the print routine twice
+    data->adf_learner = as_multiline(base_learner->get_learner_by_name_prefix("cb_explore_adf_"));
```
rajan-chari, line 852: does base->print_example not work?  Seems like it would be good to avoid making assumption about the stack if possible.

rajan-chari: or if the functionality of cb_explore_adf print() is needed then perhaps refer to that function directly?

vowpalwabbit/automl.cc
```diff
@@ -0,0 +1,876 @@
...
+  }
+
+  VW::finish_example(all, ec);
+}
+
+template <typename CMType>
+void save_load_aml(automl<CMType>& d, io_buf& model_file, bool read, bool text)
+{
+  if (model_file.num_files() == 0) { return; }
+  VW::model_utils::process_model_field(model_file, d.current_state, read, "_aml_state", text);
```
rajan-chari, line 752: Some encapsulation here might make sense:
auto_ml.save_load()
  if(model...
  process_model_file(...)
  cm->save_load(..)

vowpalwabbit/automl.h
```diff
@@ -0,0 +1,173 @@
...
+
+  // Public Chacha functions
+  void config_oracle();
+  void pre_process(const multi_ex&);
+  void schedule();
+  void update_champ();
+
+private:
+  void gen_quadratic_interactions(size_t);
+  bool better(const exclusion_config&, const exclusion_config&) const;
```
rajan-chari, line 150: Toying with the idea of making better, worse independent of config_manager implentation.... 

If better & worse accepted a class representing model performance "model_stats", then better or worse can be independent of config_manager.
If that were the case, current better() fn could be implmented outside the config_manager and be called better_lower_bound(stat1, stat2) for example.

vowpalwabbit/automl.h
```diff
@@ -0,0 +1,173 @@
...
+};
+
+template <typename CMType>
+struct automl
+{
+  automl_state current_state = automl_state::Collecting;
+  std::unique_ptr<CMType> cm;
+  vw* all = nullptr;                              //  TBD might not be needed
+  LEARNER::multi_learner* adf_learner = nullptr;  //  re-use print from cb_explore_adf
+  ACTION_SCORE::action_scores champ_a_s;          // a sequence of classes with scores.  Also used for probabilities.
```
rajan-chari, line 165: champ_a_s can be private to automl

lalo: We need them public currently for the tests, we would have to put getters and that sounds to formal for just testing.

vowpalwabbit/automl.cc
lalo: this has always been like this, or what is the meaning of savings? Plus it doesn't matter since automl only has been tested with cb_explore_adf and I believe that one is set to false. So we still call predict, let me know if I'm out of date about this Griffin.

bassmang: No this looks good to me

vowpalwabbit/automl.cc
bassmang: Im planning on changing all the save_load stuff once this is merged. Check out this PR to see the direction:
https://github.com/VowpalWabbit/vowpal_wabbit/pull/3358
It should all get a lot cleaner

vowpalwabbit/automl.h
bassmang: champ_a_s actually isnt being used in any of our unit tests (only champ from config_manager is) so I think we could change this

vowpalwabbit/automl.h
bassmang: I think that makes sense but I'm concerned its preemptive. This would require us to template the automl, config_manager, and exclusion_config structs, as well as many functions within them, and up to 3 new flags for stats_type, better_func, worse_func (which may or may not be relevant in the future). This may overcomplicate things depending on how automl gets configured in the future

vowpalwabbit/automl.cc
rajan-chari: Adding dependency on a very specific reduction to print out progress seems like excessive coupling to me.  It would be less troubling if this was not getting merged to master.  My worry is that this will never get addressed.  Having said that, adding a tech-debt section to the project stories and putting this a story under it seems like a way to move forward.  Let me know if that makes sense and also questions about where/how to add it.

PR #3255, opened 2021-08-23
fix: fix warning is search from unsafely using realloc

jackgerrits:


lalo:
is there a test exercising this code path?

jackgerrits:
Apparently not based on codecov, do you know how I can exercise it?

PR #3258, opened 2021-08-23
ci: Run clang tidy checks for each PR

jackgerrits:
- Since checks will be required to pass they should be scoped back. I suggest we just start with the decided coding guideline of braces required for every block
- **Will not pass** until the checks are fixed in master


vowpalwabbit/cbify.cc
```diff
@@ -343,7 +343,7 @@ void predict_or_learn(cbify& data, single_learner& base, example& ec)
   if (use_cs)
     csl = std::move(ec.l.cs);
   else
-    ld = std::move(ec.l.multi);
+    ld = ec.l.multi;
```
peterychang, line 346: i guess we don't foresee any changes where this would make a difference?

jackgerrits: This resolves the lint of moving a trivially copyable type having no effect. If this object becomes non-trivial where the move would be necessary then this CI will tell us because of the `performance-unnecessary-copy-initialization` check.

peterychang (approved):
one nit: If you're gonna remove the `std::move` calls, maybe add a comment somewhere noting that `ec.l.cs` contains a vector so `std::move` is necessary, whereas `ec.l.multi` is a POD type so its unnecessary

jackgerrits:
That's kind of the point in making this an automated check though - we can enforce that it is done right in case we forget to check the type before copying/moving

jackgerrits:
Updated to allow for moving trivially copyable types to avoid this confusion.

PR #3265, opened 2021-08-26
refactor: Encapsulate resetting and current file in io_buf

jackgerrits:


olgavrou (commented):
is this reset functionality being tested anywhere?

jackgerrits:
It is used during multiple passes - so it is being exercised

olgavrou (approved):

PR #3270, opened 2021-08-30
feat: support skip reasons in run_tests.py

jackgerrits:


test/run_tests.py
```diff
@@ -600,6 +603,8 @@ def convert_tests_for_flatbuffers(tests, to_flatbuff, working_dir, color_enum):
         # pdrop is not supported in fb, so 327-331 are excluded
         # 336, 337, 338 - the FB converter script seems to be affecting the invert_hash
         if str(test_id) in ('300', '189', '312', '316', '318', '319', '324', '325', '326', '327', '328', '329', '330', '331', '336', '337', '338'):
+            test['skip'] = True
+            test['skip_reason'] = "test skipped for automatic converted flatbuffer tests for unknown reason"
```
lalo, line 607: some of these we do know, so in theory we could split them now that we have skip reason

jackgerrits: That sounds good!

lalo (approved):

PR #3272, opened 2021-08-31
fix: properly release memory in the event of an error

jackgerrits:


vowpalwabbit/main.cc
```diff
@@ -128,7 +128,7 @@ int main(int argc, char* argv[])
   {
     // TODO: If loggers are instantiated within struct vw, this line lives outside of that. Log as critical for now
     std::cerr << "[critical] vw (" << e.Filename() << ":" << e.LineNumber() << "): " << e.what() << std::endl;
-    exit(1);
+    return 1;
```
jackgerrits, line 131: return instead of exit allows destructors to run

olgavrou (approved):

PR #3281, opened 2021-09-02
test: unit tests improvements

lalo:
1) remove print from explore_test.cc
2) shift things around from tutorial_test.cc to simulator.h
3) add callback hook to simulator

### DIFF
--- a/test/unit_test/explore_test.cc
+++ b/test/unit_test/explore_test.cc
@@ -49,8 +49,11 @@ BOOST_AUTO_TEST_CASE(new_sample_pdf)
   the_pdf.push_back({2.f, 3.5f, 0.1f / 1.5f});
   the_pdf.push_back({3.5f, 4.5f, 0.8f / 1.0f});
   the_pdf.push_back({4.5f, 6.2f, 0.1f / 1.7f});
-  const auto str_pdf = to_string(the_pdf);
-  std::cout << str_pdf << std::endl;
+  const auto str_pdf = to_string(the_pdf);  // 2-3.5:0.0666667,3.5-4.5:0.8,4.5-6.2:0.0588235
+  // avoid float precision compare
+  BOOST_CHECK_EQUAL(str_pdf.find("2-3.5:"), 0);
+  BOOST_CHECK(str_pdf.find(",3.5-4.5:0.8,4.5-6.2:") > 0);
+
   uint64_t seed = 7791;
   float chosen_action = 0.f;
   float pdf_value = 0.f;
--- a/test/unit_test/simulator.h
+++ b/test/unit_test/simulator.h
@@ -17,11 +17,11 @@ using callback_map = typename std::map<int, std::function<bool(vw*)>>;
 
 class cb_sim
 {
-  float USER_LIKED_ARTICLE = -1.f;
-  float USER_DISLIKED_ARTICLE = 0.f;
-  const std::vector<std::string> users{"Tom", "Anna"};
-  const std::vector<std::string> times_of_day{"morning", "afternoon"};
-  const std::vector<std::string> actions{"politics", "sports", "music", "food", "finance", "health", "camping"};
+  const float USER_LIKED_ARTICLE = -1.f;
+  const float USER_DISLIKED_ARTICLE = 0.f;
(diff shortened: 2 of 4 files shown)


test/unit_test/explore_test.cc
```diff
@@ -49,8 +49,11 @@ BOOST_AUTO_TEST_CASE(new_sample_pdf)
   the_pdf.push_back({2.f, 3.5f, 0.1f / 1.5f});
   the_pdf.push_back({3.5f, 4.5f, 0.8f / 1.0f});
   the_pdf.push_back({4.5f, 6.2f, 0.1f / 1.7f});
-  const auto str_pdf = to_string(the_pdf);
-  std::cout << str_pdf << std::endl;
+  const auto str_pdf = to_string(the_pdf);  // 2-3.5:0.0666667,3.5-4.5:0.8,4.5-6.2:0.0588235
+  // avoid float precision compare
+  BOOST_CHECK_EQUAL(str_pdf.find("2-3.5:"), 0);
+  BOOST_CHECK(str_pdf.find(",3.5-4.5:0.8,4.5-6.2:") > 0);
```
jackgerrits, line 55: This wont be resilient to float differences. Is that okay? Also if the to_string changes then this test will break

jackgerrits (approved):

test/unit_test/explore_test.cc
lalo: well its all generated from right above, any suggestions on how to replace that print? or just delete or what

test/unit_test/simulator.h
```diff
@@ -5,6 +5,10 @@
 
 namespace simulator
 {
+// maps an int: # learned examples
+// with a function to 'test' at that point in time in the simulator
+using callback_map = typename std::map<int, std::function<bool(vw*)>>;
```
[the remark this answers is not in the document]
lalo: it will be later i think, or what makes you say that?

test/unit_test/explore_test.cc
jackgerrits: You could pick out the bits of it and assert that segment exists. It's way more work and not really worth it. I just wanted to point it out. Let's not worry about it

test/unit_test/simulator.h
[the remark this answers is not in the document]
jackgerrits: `std::map` is not ambiguous so you don't need to tell the compiler is a type? I'm not 100% so leaving it here is fine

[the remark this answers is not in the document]
lalo: yea the benefit is low, i was thinking for default arg to some function. It looks a bit cleaner with a short typedef

test/unit_test/explore_test.cc
lalo: if it passes it stays, if not we can later revisit. It is already wrong printing random stuff.

lalo:
related: https://github.com/VowpalWabbit/vowpal_wabbit/pull/3252

bassmang (approved):

test/unit_test/simulator.h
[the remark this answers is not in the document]
olgavrou: `typename` shouldn't be needed since there isn't a need to specify that this is a type, I think it is mostly used with templates to distinguish any ambiguity

test/unit_test/simulator.h
```diff
@@ -26,5 +30,10 @@ class cb_sim
   const std::string& choose_user();
   const std::string& choose_time_of_day();
   std::vector<float> run_simulation(vw*, int, bool = true, int = 1);
+  std::vector<float> run_simulation_hook(vw*, int, callback_map&, bool = true, int = 1);
 };
+
+std::vector<float> _test_helper(const std::string&, int = 3000, int = 10);
+std::vector<float> _test_helper_save_load(const std::string&, int = 3000, int = 10);
+std::vector<float> _test_helper_hook(const std::string&, callback_map&, int = 3000, int = 10);
```
olgavrou, line 38: where is this being called from?

lalo: automl PR for more examples

PR #3287, opened 2021-09-02
refactor: migrate pmf_to_pdf to builder

jackgerrits:


vowpalwabbit/pmf_to_pdf.cc
```diff
@@ -295,13 +295,17 @@ base_learner* setup(VW::setup_base_i& stack_builder)
...
 
-  learner<pmf_to_pdf::reduction, example>& l =
-      init_learner(data, p_base, learn, predict, 1, prediction_type_t::pdf, stack_builder.get_setupfn_name(setup));
-
-  return make_base(l);
+  auto* l = VW::LEARNER::make_reduction_learner(
+      std::move(data), p_base, learn, predict, stack_builder.get_setupfn_name(setup))
+                .set_prediction_type(prediction_type_t::pdf)
+                .set_label_type(label_type_t::continuous)
+                // .set_output_label_type(label_type_t::cb)
```
lalo, line 305: is this a todo?

lalo (approved):
check question, but lgtm

vowpalwabbit/pmf_to_pdf.cc
jackgerrits: This is ready for when we extend the learner to understand both sides of the input/output. So far it only supports input labels and output predictions. To fully typecheck eventually we will need both.

PR #3282, opened 2021-09-02
feat: Interaction Grounded Learning

robotsorcerer:
This PR adds the implementation for the [Interaction Grounded Learning](https://arxiv.org/pdf/2106.04887.pdf) algorithm. It includes the following 

+ `interaction_ground.h`
+ `interaction_ground.cc`
+ `reductions.h`
+ `reductions.cc`
+ `cbify.h`
+ `cbify.cc`

PR #3289, opened 2021-09-03
feat: Interaction ground

lakehanne:
Removed `vw& all` offending line.


lalo:
closing since it should be equivalent to #3282

lalo:
closing since it should be equivalent to #3282

PR #3294, opened 2021-09-03
feat: update active_interactor.py for Python3

jackgerrits:


lalo (approved):
how hard is to add a test on CI? could be a straight python command right after we run py.test

jackgerrits:
Discussed offline - tricky since the script relies on an interactive session

PR #3296, opened 2021-09-03
ci: add workflow to produce artifacts and make it possible to just package executables

jackgerrits:


.github/workflows/package_linux.yml
```diff
@@ -0,0 +1,47 @@
...
+        shell: bash
+        run: |
+          cd build
+          cpack -G TGZ
+      - name: Package Deb
+        shell: bash
+        run: |
+          cd build
+          cpack -G DEB
+      - name: Upload tar.gz
```
lalo, line 38: this could also be used to package the header files and vw library right? or at least in theory

lalo: mention in the PR desc the musl callout

jackgerrits: Yes, the problem is that libvw is usually used statically - so you've got lots of compatibility issues there and so doing this in a portable way needs something like nix (which is already done). Shipping the DLL this way would be good but I just omitted here for simplicity. Musl would still not be the whole story I think since I am not sure about binary compat across compilers, so you'd need to match those too (nix to the rescue). Additionally, since the static lib depends on spdlog and fmt those need to be shipped too if you decide to package a static lib (and it just feels odd to do this) - that's the primary reason for the VW_INSTALL_EXE change in this PR.

PR #3300, opened 2021-09-07
refactor: Move empty interaction check into inner scope where size is known

jackgerrits:


vowpalwabbit/interactions_predict.h
```diff
@@ -62,11 +62,30 @@ struct feature_gen_data
   }
 };
 
+inline bool term_is_empty(namespace_index term, const std::array<features, NUM_NAMESPACES>& feature_groups)
+{
+  return feature_groups[term].empty();
+}
+
+inline bool has_empty_interaction_quadratic(
```
olgavrou, line 70: should we assert for the `[0], [1], [2]` indexes here?

olgavrou (approved):

PR #3301, opened 2021-09-07
fix: remove v_array copy from --onethread and std::function from parse_dispatch

jackgerrits:


vowpalwabbit/learner.cc
```diff
@@ -237,12 +237,21 @@ class ready_examples_queue
...
-  example* pop() { return _index < _examples.size() ? _examples[_index++] : nullptr; }
+  example* pop()
+  {
+    assert(_examples != nullptr);
+    return _index < _examples->size() ? (*_examples)[_index++] : nullptr;
+  }
 
 private:
-  v_array<example*> _examples;
+  const v_array<example*>* _examples;
```
olgavrou, line 254: why was this made a pointer? who owns the `_examples` memory?

jackgerrits: A v_array is passed in it needs to be converted to this interface to be processed. In the past this relied on the shallow copy of the v_array to function efficiently. `_examples` is owned by whoever calls parse_dispatch and we are done with it by the end of the call

olgavrou (approved):

PR #3302, opened 2021-09-07
Add new safety reduction that uses baseline if it is better than the current policy.

kumpera:
The reduction builds a CI around the baseline action  and use it instead of the policy as the greedy action
if its lower bound is higher than the policy expected reward.

This PR does change how ChiSquared is serialized a few ways:

- Don't persist derived data (the duals stuff)
- Don't serialize constructor args.
This is better left to caller as some of those could come from CLI args in which case they shouldn't be serialized.
XXX Maybe pass a bool to control when to do it?
- Pass an extra char* name prefix. This allow us to serialize multiple instances in a readable model and be able to distinguish them.


vowpalwabbit/baseline_challenger_cb.cc
```diff
@@ -0,0 +1,196 @@
...
+
+    //must not be const since we mutate it at the end
+    auto& action_scores = examples[0]->pred.a_s;
+    const uint32_t chosen_action = get_chosen_action(action_scores);
+
+    if (is_learn)
+    {
+      const auto lbl_example = std::find_if(examples.begin(), examples.end(), [](example* item) { return !item->l.cb.costs.empty(); });
+      const auto labelled_action = static_cast<uint32_t>(std::distance(examples.begin(), lbl_example));
+      if (lbl_example != examples.end()) {
```
jackgerrits, line 92: This check should be done before you cast to a uint32_t so it is still an iterator

vowpalwabbit/baseline_challenger_cb.cc
```diff
@@ -0,0 +1,196 @@
...
+        const CB::cb_class logged = (*lbl_example)->l.cb.costs[0];
```
jackgerrits, line 93:
```suggestion
        const CB::cb_class& logged = (*lbl_example)->l.cb.costs[0];
```

vowpalwabbit/baseline_challenger_cb.cc
```diff
@@ -0,0 +1,196 @@
...
+        policy_expectation.update(w, r);
+        baseline.update(w2, r);
+      }
+
+      multiline_learn_or_predict<true>(base, examples, examples[0]->ft_offset);
+    }
+
+    //We play baseline if policy expectation is worse than the baseline lower bound
+    double ci = baseline.lower_bound();
+    double exp = policy_expectation.current();
```
jackgerrits, line 108: rename as exp is messing up syntax highlighting

vowpalwabbit/baseline_challenger_cb.cc
```diff
@@ -0,0 +1,196 @@
...
+  if (!options.add_parse_and_check_necessary(new_options)) {
+    return nullptr;
+  }
+
+  if (!options.was_supplied("cb_adf"))
+    { THROW("cb_challenger requires cb_explore_adf or cb_adf"); }
+
+  if (!all.logger.quiet)
+  {
+    *(all.trace_message) << "Using CB baseline challenger" << std::endl;
```
jackgerrits, line 177: This will already be conveyed in the `enabled reductions` component of the output so probably shouldnt be printed

test/unit_test/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 add_executable(vw-unit-test.out
   automl_test.cc
+  baseline_cb_test.cc
```
lalo, line 3: this file is missing or it changed name

test/unit_test/baseline_cb_test.cc
```diff
@@ -0,0 +1,143 @@
...
+    if(s <= probs[i]) {
+      return i;
+    }
+    s -= probs[i];
+  }
+  return 0; //error
+}
+
+
+int get_int_metric(const VW::metric_sink& metrics, const std::string& metric_name)
```
lalo, line 42: Should we try to avoid having almost the same util code duplicated across tests? https://github.com/VowpalWabbit/vowpal_wabbit/blob/6670fc8db7cc44e65f84e5538c9bf285b204dd1c/test/unit_test/automl_test.cc#L25

lalo: Although its just a find_if, but makes me think that these sort of utility function might even be useful as part of metric_sink. Right now its pretty barebones, just a collection of lists so not exactly sure if it worth it

test/unit_test/baseline_cb_test.cc
```diff
@@ -0,0 +1,143 @@
...
+#include <boost/test/unit_test.hpp>
+#include <boost/test/test_tools.hpp>
+
+#include "test_common.h"
+
+#include <vector>
+
+#include "vw.h"
+
+void make_example(multi_ex& examples, vw& vw, int arm, float* costs, float* probs)
```
lalo, line 17: nit: namespace for helper functions

test/unit_test/baseline_cb_test.cc
kumpera: metric_sink is not a great API as far as partial usage.
My preference would be something like dotnet's perf counters.
IE, pseudo-code 
```
vw::perf_counter<int> is_baseline = all->get_counter<int>("baseline_cb_baseline_in_use");
///the test code.
BOOST_CHECK_EQUAL(is_baseline.sample(), 0);
```

vowpalwabbit/baseline_challenger_cb.cc
kumpera: yeah... at least it didn't change the point to exp.

vowpalwabbit/baseline_challenger_cb.cc
```diff
@@ -0,0 +1,177 @@
...
+    std::stringstream msg;
+    if (!read) { msg << name << "_expectation_sum " << std::fixed << sum << "\n"; }
+    bin_text_read_write_fixed_validated(io, reinterpret_cast<char*>(&sum), sizeof(sum), "", read, msg, text);
+
+    if (!read) { msg << name << "_expectation_n " << std::fixed << n << "\n"; }
+    bin_text_read_write_fixed_validated(io, reinterpret_cast<char*>(&n), sizeof(sum), "", read, msg, text);
+  }
+
+private:
+  double tau, sum, n;
```
jackgerrits, line 51: nit: one member per line

jackgerrits (approved):

vowpalwabbit/distributionally_robust.cc
```diff
@@ -162,6 +162,34 @@ ScoredDual ChiSquared::recompute_duals()
   return duals;
 }
 
+void ChiSquared::save_load(io_buf& model_file, bool read, bool text, const char* name)
+{
+  if (model_file.num_files() == 0) { return; }
+  std::stringstream msg;
+
+#define save_load_field(field)                                                            \
```
jackgerrits, line 170: I've added equivalent functionality with this PR: #3326. It's a tiny bit more verbose but does not rely on implicit variable names.

```c++
save_load_field(n);
```
becomes
```c++
VW::process_model_field(io, n, read, "_chisq_n", text);
```

kumpera: It won't work OOTB because I need to dynamically create the read string, but it's worth simplifying the macro.

jackgerrits: What do you mean by that? I believe it is supported. Please see the doc for the func: https://vowpalwabbit.org/docs/vowpal_wabbit/cpp/latest/namespaceVW_1_1model__utils.html#a8369057d5016e6cfdf02f0ea80570572

PR #3303, opened 2021-09-07
test: add kill cache to test 344 so it can be rerun without failure

jackgerrits:


olgavrou (approved):

test/RunTests
```diff
@@ -2474,7 +2474,7 @@ train-sets/ref/slates_simple_w_interactions.stderr
     train-sets/ref/cb_adf_mtr_cache_2.stderr
 
 # Test 344: check cb_adf with multiple passes
-{VW} --cb_adf -c --passes 2 -d train-sets/cb_adf_multipass_crash_1.data --holdout_off
+{VW} --cb_adf -k -c --passes 2 -d train-sets/cb_adf_multipass_crash_1.data --holdout_off
```
lalo, line 2477: can you explain why we need this?

jackgerrits: If you run the benchmark once it will produce a cache file. Then when you run it again it will use the cache file instead of the text input. This breaks the stderr diff test. -k means it always uses the text file and is a consistent run

lalo: but then this implies that the second pass on this test will also re-read and not use the cache?

jackgerrits: -k only affects the first pass, the second pass will still use the cache as intended

PR #3309, opened 2021-09-09
refactor: Add label and use builder to set learner features in cb_to_cb_adf and cbify

bassmang:


vowpalwabbit/cbify.cc
```diff
@@ -753,59 +753,95 @@ base_learner* cbify_setup(VW::setup_base_i& stack_builder)
...
+      predict_ptr = predict_adf<false>;
+      finish_ptr = MULTICLASS::finish_example<cbify&>;
+      name_addition = "-adf";
+      label_type = label_type_t::multiclass;
+      all.example_parser->lbl_parser = MULTICLASS::mc_label;
     }
+    l = make_reduction_learner(
+        std::move(data), base, learn_ptr, predict_ptr, stack_builder.get_setupfn_name(cbify_setup) + name_addition)
+            .set_finish_example(finish_ptr)
+            .set_label_type(label_type)
```
jackgerrits, line 789: What's the prediction type in this case?

bassmang: Looks like multiclass, I'll put that in

jackgerrits (approved):

PR #3310, opened 2021-09-09
refactor: Add label and use builder to set learner features in cbzoo and confidence

bassmang:


vowpalwabbit/cbzo.cc
```diff
@@ -362,13 +363,14 @@ base_learner* setup(VW::setup_base_i& stack_builder)
   data->min_prediction_supplied = options.was_supplied("min_prediction");
   data->max_prediction_supplied = options.was_supplied("max_prediction");
 
-  learner<cbzo, example>& l = init_learner(data, get_learn(all, policy, feature_mask_off), get_predict(all, policy), 0,
-      prediction_type_t::pdf, stack_builder.get_setupfn_name(setup));
+  auto* l = make_base_learner(std::move(data), get_learn(all, policy, feature_mask_off), get_predict(all, policy),
+      stack_builder.get_setupfn_name(setup), prediction_type_t::pdf, label_type_t::continuous)
+                .set_params_per_weight(0)
```
jackgerrits, line 368: This seems odd - but it's how it was before. At least it is clear that's what the zero is doing now

jackgerrits (approved):

vowpalwabbit/cbzo.cc
bassmang: Yea it does seem odd

PR #3311, opened 2021-09-10
test: add cb_sim ref to test callback

lalo:


lalo:
see #3252 for actual usage

bassmang (approved):

PR #3314, opened 2021-09-13
refactor: explicitly pass buffer into parser read call

jackgerrits:


olgavrou (approved):
General note: when we update rlclientlib `binary_parser` will also have to be modified to accept this

PR #3317, opened 2021-09-13
fix: don't use heap for io_buf

jackgerrits:
Will conflict with #3314 so will need to be fixed up after that is merged


vowpalwabbit/parser.h
```diff
@@ -57,7 +55,7 @@ struct parser
   VW::object_pool<example> example_pool;
   VW::ptr_queue<example> ready_parsed_examples;
 
-  std::unique_ptr<io_buf> input;  // Input source(s)
+  io_buf input;  // Input source(s)
```
olgavrou, line 58: I wonder if this was heap allocated on purpose to avoid potential stack overflow from IO

jackgerrits: The buffer itself in io_buf must be heap allocated as it grows. This just prevents the io_buf object being on the heap, and it instead inline on the object

olgavrou (approved):

PR #3319, opened 2021-09-13
fix: perf issues found by clang-tidy

jackgerrits:
Found with checks:
- performance-unnecessary-copy-initialization
- performance-for-range-copy
- performance-move-const-arg


jackgerrits:
Updated due to a clash with a global gnu function `major`

lalo (approved):

PR #3320, opened 2021-09-13
ci: use sys deps to improve build time

jackgerrits:
The 18.04 image includes sys deps for spdlog and fmt, use them to reduce build times


jackgerrits:
Hm either does not improve or hampers, based on a single sample.

PR #3321, opened 2021-09-14
fix: active was using inf values for sensitivity

jackgerrits:


vowpalwabbit/active.cc
```diff
@@ -87,7 +90,12 @@ void predict_or_learn_active(active& a, single_learner& base, example& ec)
   if (ec.l.simple.label == FLT_MAX)
   {
     const float threshold = (a._shared_data->max_label + a._shared_data->min_label) * 0.5f;
+    // We want to understand the change in prediction if the label were to be
+    // the opposite of what was predicted. 0 and 1 are used for the expected min
+    // and max labels to be coming in from the active interactor.
+    ec.l.simple.label = (ec.pred.scalar >= threshold) ? ACTIVE_MIN_LABEL : ACTIVE_MAX_LABEL;
```
JohnLangford, line 96: I believe that threshold should be defined based on ACTIVE_MIN_LABEL and ACTIVE_MAX_LABEL, and that these values should be data dependent.

jackgerrits: If `ACTIVE_MIN_LABEL` and `ACTIVE_MAX_LABEL` are defined based on **labels** seen so far, how do we then use this threshold to compare to `ec.pred.scalar` which is in a completely different scale `[-50,50]`?

JohnLangford: Logistic regression, which is probably being used here, uses labels of {-1,+1}, so the threshold is 0, just as it is for [-50,50].

jackgerrits: Is `ec.confidence = fabsf(ec.pred.scalar - threshold) / base.sensitivity(ec);` still correct in this case?

JohnLangford: It should be.  Essentially, we're asking "how much update is required to cause the prediction to go the other way?"

jackgerrits: Okay got it. I've updated the PR to track min and max seen labels.

PR #3323, opened 2021-09-14
refactor: Add label and use builder to set learner features in cs_active and explore_eval

bassmang:


vowpalwabbit/cs_active.cc
```diff
@@ -354,18 +354,36 @@ base_learner* cs_active_setup(VW::setup_base_i& stack_builder)
...
-  learner<cs_active, example>& l = simulation
-      ? init_learner(data, as_singleline(stack_builder.setup_base_learner()), predict_or_learn<true, true>,
-            predict_or_learn<false, true>, data->num_classes, prediction_type_t::active_multiclass,
-            stack_builder.get_setupfn_name(cs_active_setup) + "-sim", true)
-      : init_learner(data, as_singleline(stack_builder.setup_base_learner()), predict_or_learn<true, false>,
-            predict_or_learn<false, false>, data->num_classes, prediction_type_t::active_multiclass,
-            stack_builder.get_setupfn_name(cs_active_setup), true);
+  void (*learn_ptr)(cs_active & cs_a, single_learner & base, example & ec);
+  void (*predict_ptr)(cs_active & cs_a, single_learner & base, example & ec);
+  size_t ws = data->num_classes;
```
jackgerrits, line 359: This doesn't need to be changed in the `if` statement, so you could pass it directly on line 377?

jackgerrits (approved):

vowpalwabbit/cs_active.cc
bassmang: I can move it after the `if/else` but it does need to be set before `make_reduction_learner` because of `std::move(data)`

jackgerrits: Oh I see, good point! Feel free to leave as is

PR #3326, opened 2021-09-15
refactor: use helper method for model variables

jackgerrits:


lalo (approved):

vowpalwabbit/io_buf.h
```diff
@@ -361,6 +363,30 @@ inline size_t bin_text_read_write_fixed_validated(
...
 
+namespace VW
+{
+template <typename T>
+size_t process_model_field(io_buf& io, T& var, bool read, VW::string_view name, bool text)
+{
+  // It is not valid to read a text based field.
+  assert(!(read && text));
+
+  auto* data = reinterpret_cast<char*>(&var);
```
olgavrou, line 374: we are assuming anything `T` can be cast into a char representation right?

jackgerrits: Yes, we are. We can add some template checks to make sure it is a type where this makes sense. POD is probably what we want to check for. This being a template allows us to overload for types that can't be trivially bit cast to bytes (like std::string)

test/pred-sets/ref/ccb_implicit_and_explicit_interactions.inv
```diff
@@ -8,7 +8,7 @@ lda:0
 0 skip:
 options: --cb_adf --cb_explore_adf --cb_sample --cb_type mtr --ccb_explore_adf --csoaa_ldf multiline --csoaa_rank --interactions AAA --interactions ::: --quadratic :: --quadratic AB
 Checksum: 2813477556
-CCB: has_seen_multi_slot_example = 1
+CCB: has_seen_multi_slot_example = true
```
lalo, line 11: can this sort of change bite us back when we read them? this mismatch across versions, i.e. reading and old readeable model with this new impl?

test/pred-sets/ref/ccb_implicit_and_explicit_interactions.inv
```diff
@@ -1,5 +1,5 @@
 Version 8.11.0
-Id 
+Id
```
lalo, line 2: also applies to this space, unless its trimmed when parsing it could shift everything as we read.

test/pred-sets/ref/ccb_implicit_and_explicit_interactions.inv
jackgerrits: Readable models cannot be parsed by VW, the are an output only thing.

test/pred-sets/ref/ccb_implicit_and_explicit_interactions.inv
jackgerrits: Same answer as before, these files cannot be read

olgavrou (approved):

PR #3327, opened 2021-09-16
feat: Privacy Preserving Learning

manavsinghal157:
Part of the _Empirical Analysis of Privacy Preserving Learning_ Project.

This PR introduces a command line argument that implements aggregated learning by saving only those features that have seen a minimum threshold of users thus upholding the privacy of the user.

### **Methodology:**

- For each feature, a 32-bit vector is defined. (`vowpalwabbit/array_parameters.h` and `vowpalwabbit/array_parameters_dense.h`)
- We calculate a 5-bit hash of the tag of the example. (`vowpalwabbit/parser.cc`)
- For each feature weight updated by a non-zero value, we use the 5-bit hash to look up a bit in the 32-bit vector and set it to 1.(`vowpalwabbit/gd_predict.h` -> (`vowpalwabbit/array_parameters.h` and `vowpalwabbit/array_parameters_dense.h`))
- When saving the weights into a file, we calculate the number of bits set to 1 for a feature. If it is greater than the threshold, the weights for that feature are saved. (`vowpalwabbit/gd.cc`->(`vowpalwabbit/array_parameters.h` and `vowpalwabbit/array_parameters_dense.h`))

_(The default value of the threshold is 10)_

### **This PR includes:**

- Command line argument to activate privacy preservation and set the threshold. (`vowpalwabbit/parse_args.cc`)
- Runtests to test the desired output on a small dataset. (`test/RunTests`)
- Unit-tests for checking output when threshold is reached for a feature and when it is not. (`test/unit_test/weights_test.cc`)
- Benchmarks to test time taken for learning in privacy preserving method. (`test/benchmarks/standalone/benchmark_text_input.cc `)

### **Implementation details:**

`--privacy_activation` : To activate the feature 
`--privacy_activation_threshold arg (=10)` : To set the threshold

### **Future Work:**

- Implement the feature for save_resume.
- Work on aggregations in the online setting.

Wiki page for the same : https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Privacy-Preserving-Learning


olgavrou:
if benchmarks are unacceptable we can put this change behind a compile time flag and compile the changes out with `#ifdef`'s

olgavrou:
Some runs look relatively ok: https://github.com/VowpalWabbit/vowpal_wabbit/runs/3623944328?check_suite_focus=true

vowpalwabbit/array_parameters_dense.h
```diff
@@ -47,10 +49,19 @@ class dense_iterator
...
+  {
+    uint64_t tag_hash;
+    bool is_set = false;
+  };
   weight* _begin;
   uint64_t _weight_mask;  // (stride*(1 << num_bits) -1)
   uint32_t _stride_shift;
   bool _seeded;  // whether the instance is sharing model state with others
+  size_t _privacy_activation_threshold;
+  std::unordered_map<uint64_t, std::bitset<32>> _feature_bitset;  // define the bitset for each feature
```
lalo, line 63: this might allocate on learn/predict, in case something pops up on the benchmarks

olgavrou: reserved size during construction

olgavrou:
> Some runs look relatively ok: https://github.com/VowpalWabbit/vowpal_wabbit/runs/3623944328?check_suite_focus=true
> 
> this one seems to suffer a bit: `bench_text_io_buf/120_string_fts`

latest benchmarks look fine after pre-allocating the bitset maps

vowpalwabbit/parse_args.cc
```diff
@@ -529,6 +529,12 @@ void parse_feature_tweaks(
 
   option_group_definition feature_options("Feature options");
   feature_options
+      .add(make_option("privacy_activation", all.privacy_activation)
```
jackgerrits, line 532: It looks like this isn't supported in every base learner. However, for a feature such as this I think it's very important to fail if a user requested `privacy_activation` but the stack's base learner wont do it

olgavrou: We can check again but last time we checked all base learners would eventually use the gd methods that have been altered here and the tag is setup by setup base so it looks like all base learners can support the feature

jackgerrits: `gd_mf` is one which will be unsupported. Also, how does this interact with `print` and `sender`?

olgavrou: `gd_mf` and `print` use `GD::foreach_feature` as far as I can see so this change should be supported there. I am not sure what `sender` does but it doesn't look like it is using weights at all.

We are looking into an alternative implementation with @manavsinghal157 of putting this change into the weight's `[]` operator so then any reduction using the `weights` will get this feature for sure (need to check the perf on that change)

Do you think it will still be a problem in that case? I am worried that if we have to add a warning in all the base learners then we might add a new base learner and forget to add the warning there so it's best to make sure this feature is used in all base learners

lalo: that warning system could be generic and done once for all args, it would just be maybe out of scope for this PR?

We would have to keep track of args that were parsed but their 'root' necessary arg was not fulfilled. That would imply it's an error, and right now i think we quietly fail to report that the arg didnt have a side-effect on the instantiation. It gets tricky since some args are repeated across reductions, but shouldn't be that complicated.

jackgerrits: You're right, I missed that in `gd_mf`! VW has a bit of a problem with knowing if an argument you supplied is actually used or not, hence my concern with an option centered around ensuring privacy. But really, this issue is not created by or limited to this new option. Your explanation is good, it seems we are sufficiently covered here. In fact this is probably one of the more universal options given the nature of it being in the base of the stack.

Hmm, `sender` is used with daemon mode so it isn't really a base learner in the traditional sense.

olgavrou: This is the alternative approach [here](https://github.com/VowpalWabbit/vowpal_wabbit/pull/3334/files#diff-f75d24249a7ddfb0203db0f25db4adec31f15f48abcf32bed9fef843f9496bfcR170)

PR #3332, opened 2021-09-17
fix: `self.classes_[indices]` in sklearn VW

sergeyf:
On line 833 there is the following: `return self.classes_[indices]`, which will error out unless `self.classes_` is a numpy array.

### DIFF
--- a/python/vowpalwabbit/sklearn_vw.py
+++ b/python/vowpalwabbit/sklearn_vw.py
@@ -931,7 +931,7 @@ class VWMultiClassifier(VWClassifier):
                 self.estimator_ = {estimator: n_classes}
                 if n_classes is not None:
                     if self.classes_ is None:
-                        self.classes_ = range(1, n_classes + 1)
+                        self.classes_ = np.array(list(range(1, n_classes + 1)))
                     break
             else:
                 # use oaa by default and determine classes from y
--- a/python/vowpalwabbit/sklearn_vw.py
+++ b/python/vowpalwabbit/sklearn_vw.py
@@ -931,7 +931,7 @@ class VWMultiClassifier(VWClassifier):
                 self.estimator_ = {estimator: n_classes}
                 if n_classes is not None:
                     if self.classes_ is None:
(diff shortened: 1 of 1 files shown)


jackgerrits:
Thanks for submitting a fix! Just to confirm, this should cover both normal lists and numpy arrays now? I do wonder why it was originally written as it was?

sergeyf:
Can you please clarify what you mean by `this should cover both normal lists and numpy arrays now`?

Also, any idea why checks are failing?

jackgerrits:
> Can you please clarify what you mean by this should cover both normal lists and numpy arrays now?
Does this code now work for both `np.array` and Python's builtin `list`?

The Python build failure is weird - failing in setting up cmake. The upload coverage is a flaky test unfortunately. I've restarted both.

sergeyf:
Ah you mean does it work if `self.classes_` is both `list` and `np.array`? I think it's only ever `np.array` as it is defined as follows in a single place only: `self.classes_, y = np.unique(y, return_inverse=True)`

Thanks for restarting the tests!

jackgerrits:
Ah I see, I am not sure I quite follow the reason for this change then as you originally stated then:
> `return self.classes_[indices]`, which will error out unless `self.classes_` is a numpy array.

If it is only ever a numpy array then shouldn't the line being changed be fine?

sergeyf:
Sorry I wrote the wrong thing previously.  Here is the full context of where the bug was

```
        scores = self.decision_function(X)
        if len(scores.shape) == 1:
            indices = (scores > 0).astype(np.int)
        else:
            indices = scores.argmax(axis=1)
        return self.classes_[indices]
```

So `indices` is *always* a `np.array`, and `self.classes_` used to be a `range` so `self.classes_[indices]` gave an error.

jackgerrits:
Ah I see now, thanks very much for the clarification. Appreciate the fix @sergeyf and prompt responses! :)

sergeyf:
Happy to help!

PR #3338, opened 2021-09-24
refactor: Apply clang-tidy performance fixes

jackgerrits:


vowpalwabbit/parse_args.cc
```diff
@@ -88,7 +89,18 @@ std::string find_in_path(const std::vector<std::string>& paths, const std::strin
 #endif
   for (const auto& path : paths)
   {
-    std::string full = VW::ends_with(path, delimiter) ? (path + fname) : (path + delimiter + fname);
+    std::string full;
```
peterychang, line 92: nit: slightly more verbose than it needs to be, but the readability probably makes up for that. Alternative would be
```
std::string full = path;
if(!vw::ends_with(path, delimiter)) {
  full += delimiter;
}
full += fname;
```

vowpalwabbit/vw.h
```diff
@@ -47,21 +47,21 @@ vw* initialize(std::unique_ptr<config::options_i, options_deleter_type> options,
     bool skip_model_load = false, trace_message_t trace_listener = nullptr, void* trace_context = nullptr);
 vw* initialize(config::options_i& options, io_buf* model = nullptr, bool skip_model_load = false,
     trace_message_t trace_listener = nullptr, void* trace_context = nullptr);
-vw* initialize(std::string s, io_buf* model = nullptr, bool skip_model_load = false,
+vw* initialize(const std::string& s, io_buf* model = nullptr, bool skip_model_load = false,
```
peterychang, line 50: These changes break the interface (definitely the ABI, possibly the API in some cases). Does that matter?

jackgerrits: Any `std::string` passed to this function will work as before (just without copying!) and we certainly do not guarantee any sort of ABI here. So I think its fine

peterychang (approved):

PR #3342, opened 2021-09-28
refactor: change counters to better reflect usage of pool

jackgerrits:


vowpalwabbit/parse_example.cc
```diff
@@ -147,7 +147,8 @@ class TC_parser
     {
       _v = float_feature_value = 0.f;
       // syntax error
-      parserWarning("malformed example! '|', ':', space, or EOL expected after : \"", _line.substr(0, _read_idx), "\"");
+      parserWarning("malformed example! '|', ':', space, or EOL expected after : \"", _line.substr(0, _read_idx), "\"",
+          _ae->example_counter);
```
lalo, line 151: Does example_counter corresponds to the line number of the source text file? or once we have multiple threads parsing will it diverge from that?

jackgerrits: I think it does match up at the moment. There is no concrete implementation/spec for multiple threads parsing and so we don't really know if it will diverge from that. I think it will require specific care we we want it to still correlate.

lalo (approved):

PR #3343, opened 2021-09-29
fix: fix bug where read_lines would ignore given length

jackgerrits:


jackgerrits:
In the WASM bindings I am making use of the text_reader entrypoint. I trim the line to remove extraneous whitespace, which surfaced that the function ignored the given length which it must critically use. It cannot assume the input is a null terminated char sequence.

lalo (approved):

PR #3346, opened 2021-09-30
refactor: rename vw -> VW::workspace

### LINKED ISSUE #2391: Proposal: rename ::vw to VW::workspace
In order to make progress in the area of namespacing and code organization I would like to make the following proposal:

- Rename `::vw` object to `VW::context`
    - https://github.com/VowpalWabbit/vowpal_wabbit/blob/17dad786022c6f56b8ee567d4e7b654a07780b89/vowpalwabbit/global_data.h#L361
- Support a compat.h header that exposes `::vw` as a type inherited from `VW::context`. This should assist existing users to migrate

I do not think there is a way to move the object out of the global namespace without some breakage but we will do a best effort at minimizing this cost.

Motivations:
- The `vw` object, which we often call all, is in the global namespace. Namespacing is an important part of organizing a library and is something we should resolve sooner rather than later.
- Renaming the object is also desired as, while possible, it is not a great idea to name a type the same as a namespace.

There has been existing discussion of the larger issue here: https://github.com/VowpalWabbit/vowpal_wabbit/issues/1769, but this issue is a targeted proposal.

jackgerrits:
Resolves #2391

Remains compatible with existing usages of `vw` due to alias. The next step would be deprecating the alias and ensuring we use it nowhere.


lalo (approved):
looks good, also good sign that pyvw is not updated yet not broken by this change

olgavrou (approved):

PR #3354, opened 2021-10-01
test: Use proper substitution names to support run_tests.py

jackgerrits:


test/RunTests
```diff
@@ -330,6 +330,11 @@ sub next_test() {
          $cmd =~ s/{VW}/$VW$opt_O/g;
     }
 
+    $spanning_tree_bin = '../build/cluster/spanning_tree';
```
jackgerrits, line 333: It's still hard coded in this file since it is much more work to find it. In `run_tests.py` it is found dynamically

lalo (approved):

PR #3358, opened 2021-10-04
refactor: Migrate chisq save_load to process_model_field

bassmang:


vowpalwabbit/distributionally_robust.cc
```diff
@@ -7,6 +7,44 @@
...
+size_t process_model_field(
+    io_buf& io, VW::distributionally_robust::Duals& duals, bool read, const std::string&, bool text)
+{
+  if (io.num_files() == 0) { return 0; }
+  process_model_field(io, duals.unbounded, read, "_duals_unbounded", text);
+  process_model_field(io, duals.kappa, read, "_duals_kappa", text);
+  process_model_field(io, duals.gamma, read, "_duals_gamma", text);
+  process_model_field(io, duals.beta, read, "_duals_beta", text);
+  process_model_field(io, duals.n, read, "_duals_n", text);
+  return 0;
```
jackgerrits, line 21: This should return the total bytes read or written. Adding up the other calls should suffice.

vowpalwabbit/distributionally_robust.cc
```diff
@@ -7,6 +7,44 @@
...
+  process_model_field(io, chisq.rmin, read, "_chisq_chi_rmin", text);
+  process_model_field(io, chisq.rmax, read, "_chisq_chi_rmax", text);
+  process_model_field(io, chisq.n, read, "_chisq_chi_n", text);
+  process_model_field(io, chisq.sumw, read, "_chisq_chi_sumw", text);
+  process_model_field(io, chisq.sumwsq, read, "_chisq_chi_sumwsq", text);
+  process_model_field(io, chisq.sumwr, read, "_chisq_chi_sumwr", text);
+  process_model_field(io, chisq.sumwsqr, read, "_chisq_chi_sumwsqr", text);
+  process_model_field(io, chisq.sumwsqrsq, read, "_chisq_chi_sumwsqrsq", text);
+  process_model_field(io, chisq.delta, read, "_chisq_chi_delta", text);
+  return 0;
```
jackgerrits, line 44: Must return number of bytes read/written

vowpalwabbit/distributionally_robust.h
```diff
@@ -139,63 +134,11 @@ namespace distributionally_robust
       return duals.second.qfunc(w, r);
     }
 
-    void save_load(io_buf& model_file, bool read, bool text)
```
jackgerrits, line 142: Where is this function being called? You are going to need to ensure the original callsite is using `VW::model_utils::process_model_field`

vowpalwabbit/distributionally_robust.h
```diff
@@ -30,26 +43,8 @@ namespace distributionally_robust
...
-      if (!read) { msg << "_duals_gamma " << gamma << "\n"; }
-      bin_text_read_write_fixed(model_file, reinterpret_cast<char*>(&gamma), sizeof(gamma), read, msg, text);
-
-      if (!read) { msg << "_duals_beta " << beta << "\n"; }
-      bin_text_read_write_fixed(model_file, reinterpret_cast<char*>(&beta), sizeof(beta), read, msg, text);
-
-      if (!read) { msg << "_duals_n " << n << "\n"; }
-      bin_text_read_write_fixed(model_file, reinterpret_cast<char*>(&n), sizeof(n), read, msg, text);
-    }
+    friend size_t VW::model_utils::process_model_field<>(
```
jackgerrits, line 46: If using `<>` then you should have `template <>`. In fact you should be able to remove `<>` if you add `template <>` as the template type can then be deduced from the args.

Additionally, why is this declaration different from that on line 26?

vowpalwabbit/distributionally_robust.h
```diff
@@ -139,63 +134,11 @@ namespace distributionally_robust
...
-      bin_text_read_write_fixed(model_file, reinterpret_cast<char*>(&sumwsqrsq), sizeof(sumwsqrsq), read, msg, text);
-
-      if (!read) { msg << "_chisq_chi_delta " << delta << "\n"; }
-      bin_text_read_write_fixed(model_file, reinterpret_cast<char*>(&delta), sizeof(delta), read, msg, text);
-    }
-
     ScoredDual recompute_duals();
     static double chisq_onedof_isf(double alpha);
     const double& effn() { return n; }
+    friend size_t VW::model_utils::process_model_field(
```
jackgerrits, line 140: Same as other friend func for consistency

jackgerrits (changes requested):

vowpalwabbit/distributionally_robust.h
bassmang: This should be deleted. I was using a specialized template at first, but it works the same without any template.

vowpalwabbit/distributionally_robust.h
bassmang: Same as above

vowpalwabbit/distributionally_robust.h
bassmang: This isnt being called anywhere in master, but it's being used (and tested) in the automl branch.

jackgerrits (approved):

PR #3359, opened 2021-10-04
Adds new metric for aggregating original label cost for first slot 

slahabar:


vowpalwabbit/metrics.cc
```diff
@@ -72,6 +72,8 @@ void list_to_json_file(dsjson_metrics* ds_metrics, const std::string& filename,
       writer.String(ds_metrics->LastEventTime.c_str());
       writer.Key("dsjson_sum_cost_original");
       writer.Double(ds_metrics->DsjsonSumCostOriginal);
+      writer.Key("dsjson_sum_cost_original_first_slot");
```
lalo, line 75: what exactly is the difference between dsjson_sum_cost_original and dsjson_sum_cost_original_first_slot? Do we ever need both at the same time?

lalo: 1) If we don't need both at the same time, an if else statement depending on cb vs ccb might be good. 2) Also I would add to the name dsjson_sum_cost_original_first_slot the keyword ccb.

lalo: synced offline

slahabar: Addressed your comments. 
Keeping the metric name as it is (more generic) since we could possibly use this for slates. Slates is not a requirement right now by APS. Emitting the metric only when it is ccb required by APS.

lalo: this is the stack for slates: gd, generate_interactions, scorer-identity, csoaa_ldf-rank, cb_adf,
cb_explore_adf_greedy, cb_sample, shared_feature_merger,
ccb_explore_adf, slates

slahabar: makes sense

vowpalwabbit/metrics.cc
```diff
@@ -31,7 +31,7 @@ struct metrics_data
   size_t predict_count = 0;
 };
 
-void list_to_json_file(dsjson_metrics* ds_metrics, const std::string& filename, metric_sink& metrics)
+void list_to_json_file(dsjson_metrics* ds_metrics, const std::string& filename, metric_sink& metrics, std::vector<std::string> enabled_reductions)
```
olgavrou, line 34:
```suggestion
void list_to_json_file(dsjson_metrics* ds_metrics, const std::string& filename, metric_sink& metrics, const std::vector<std::string>& enabled_reductions)
```

olgavrou (approved):

vowpalwabbit/metrics.cc
slahabar: Addressed

vowpalwabbit/metrics.cc
```diff
@@ -70,8 +70,13 @@ void list_to_json_file(dsjson_metrics* ds_metrics, const std::string& filename,
       writer.String(ds_metrics->LastEventId.c_str());
       writer.Key("last_event_time");
       writer.String(ds_metrics->LastEventTime.c_str());
-      writer.Key("dsjson_sum_cost_original");
-      writer.Double(ds_metrics->DsjsonSumCostOriginal);
+      if (std::find(enabled_reductions.begin(), enabled_reductions.end(), "ccb_explore_adf") != enabled_reductions.end())
+      {        
+        writer.Key("dsjson_sum_cost_original");
```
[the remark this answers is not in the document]
slahabar: Addressed

jackgerrits (approved):

lokitoth (approved):

PR #3360, opened 2021-10-05
feat: ability to capture the audit output using the vwdll api

aaiyer:
ability to capture the audit output using the vwdll api instead of only printing it to stdout


jackgerrits:
Hi @aaiyer, thanks for opening this, apologies for not getting back to you sooner. I agree, the audit information is not well exposed to lib consumers. The current implementation of audit is very command line focused and has usability issues. I think when we expose it to library consumers we need to have a more usable structured output and I'm not sure what that is going to look like yet

jackgerrits:
This is ready to go apart from a test that is consistently failing in the Windows CI which utilizes the checked in solution. I am not sure what the issue is here since I cannot repro it. The test works fine in the CMake based Windows CI, CMake and checked in solution locally for me and under Ubuntu.

vowpalwabbit/global_data.h
```diff
@@ -264,6 +264,8 @@ struct workspace
 
   vw_logger logger;
   bool audit;     // should I print lots of debugging information?
+  std::shared_ptr<std::vector<char>> _audit_buffer;
```
lokitoth, line 267: [Minor] Consider keeping the naming convention for members consistent within objects.

lokitoth (approved):
Looks good overall. Had a minor comment, but do not block on it if time is of the essence.

vowpalwabbit/global_data.h
jackgerrits: Happy to fix

PR #3362, opened 2021-10-05
fix: Add --probabilities flag to multilabel_oaa, Resolves #3331

### LINKED ISSUE #3331: CSOAA in Python not returning probabilities
Hello,

Thanks for the great work on VW over the many years!

I'm using the Python package with the `csoaa` option to deal with a multilabel problem. It seems to train fine, but I can't get it to return probabilities. This is how I trained without any errors:

```
from vowpalwabbit import pyvw

vw = pyvw.vw(quiet=True, csoaa=y_train.shape[1], b=26, ngram="d2", probabilities=True)
passes = 1
for n in range(passes):
    print(n)
    if n >= 1:
        examples = shuffle(Xy_train)
    else:
        examples = Xy_train
    for idx, example in enumerate(examples):
        vw.learn(example)
```

Where `Xy_train` has things like 

```
Xy_train[0]
Out: '1:1.0 2:1.0 3:1.0 4:0.0 5:1.0 6:1.0 7:1.0 8:1.0 9:1.0 10:1.0 11:1.0 12:1.0 13:1.0 14:1.0 15:1.0 16:1.0 17:1.0 18:1.0 19:1.0 20:1.0 21:1.0 22:1.0 23:1.0 24:1.0 |text a transient based real time scheduling algorithm in fms 
```

At test time, I have data like:

```
Xy_val[0]
Out: '1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 |text covid associated ards treated with dexamethasone codex study design and rationale for a randomized trial
```

And doing `vw.predict(Xy_val[0])` probabilities aren't returned, but just a single integer. How can one get probabilities out of this? I've tried `vw.predict(Xy_val[0], i)` for various `i` and no luck.

Thank you.

bassmang:
1. Adds --probabilities flag to multilabel_oaa
2. Throws error if a user tries to use --probabilities with --csoaa
3. Automatically inserts "--link logistic" if using --probabilities with --oaa or --multilable_oaa
4. Breaks previous output of "--oaa --probabilities". Previously the logic of "--link logistic" was being applied after the link (scorer.cc) reduction. This lead to some values being stored incorrectly. The previous implementation of this also explicitly required the user to not use "--link logistic," and would apply the logic at a later stage.


vowpalwabbit/multilabel_oaa.cc
```diff
@@ -64,18 +91,46 @@ VW::LEARNER::base_learner* multilabel_oaa_setup(VW::setup_base_i& stack_builder)
...
+  size_t ws = data->k;
+
+  if (data->probabilities)
+  {
+    options.insert("link", "logistic");
+    options.add_and_parse(new_options);
+    pred_type = prediction_type_t::scalars;
+    auto loss_function_type = all.loss->getType();
+    if (loss_function_type != "logistic")
+      *(all.trace_message) << "WARNING: --probabilities should be used only with --loss_function=logistic" << std::endl;
```
jackgerrits, line 113: Since this is new code I think you should use the logger to emit a warning

vowpalwabbit/multilabel_oaa.cc
```diff
@@ -50,12 +54,35 @@ void predict_or_learn(multi_oaa& o, VW::LEARNER::single_learner& base, example&
                         multilabels.label_v[multilabel_index], o.k - 1);
     }
   }
-  ec.pred.multilabels = preds;
-  ec.l.multilabels = multilabels;
+  if (o.probabilities)
+  {
+    float sum = std::accumulate(ec.pred.scalars.begin(), ec.pred.scalars.end(), 0.f);
+    for (size_t i = 0; i < o.k; ++i) { ec.pred.scalars[i] /= sum; }
```
jackgerrits, line 60: Could you add an assert that `ec.pred.scalars.size() == o.k`?

vowpalwabbit/multilabel_oaa.cc
```diff
@@ -5,15 +5,18 @@
 #include <cfloat>
 #include "reductions.h"
 #include "vw.h"
+#include "shared_data.h"
 
 #include "io/logger.h"
 
 using namespace VW::config;
+using namespace VW::LEARNER;
```
jackgerrits, line 13: We should avoid usings like this whenever possible. Of course this comes under personal preference but it seems like there are only 3 functions from this namespace used here so in my opinion we should not use this `using`

vowpalwabbit/oaa.cc
```diff
@@ -233,11 +233,17 @@ VW::LEARNER::base_learner* oaa_setup(VW::setup_base_i& stack_builder)
...
-      .add(make_option("probabilities", probabilities).help("predict probabilites of all classes"))
+      .add(make_option("probabilities", probabilities).help("predict probabilities of all classes"))
       .add(make_option("scores", scores).help("output raw scores per class"));
 
   if (!options.add_parse_and_check_necessary(new_options)) return nullptr;
 
+  if (probabilities)
+  {
+    options.insert("link", "logistic");
+    options.add_and_parse(new_options);
```
jackgerrits, line 244: I do not think this line is necessary. For reference we add options elsewhere and do not use it (https://github.com/VowpalWabbit/vowpal_wabbit/blob/5f13251cbca146bacc7bb6f4e3c390dd6857117a/vowpalwabbit/cb_explore.cc#L365)

vowpalwabbit/multilabel_oaa.cc
bassmang: Yea took it out, it was only being used in two places here

vowpalwabbit/multilabel_oaa.cc
```diff
@@ -50,12 +53,36 @@ void predict_or_learn(multi_oaa& o, VW::LEARNER::single_learner& base, example&
                         multilabels.label_v[multilabel_index], o.k - 1);
     }
   }
-  ec.pred.multilabels = preds;
-  ec.l.multilabels = multilabels;
+  if (o.probabilities)
+  {
+    assert(ec.pred.scalars.size() == o.k);
+    float sum = std::accumulate(ec.pred.scalars.begin(), ec.pred.scalars.end(), 0.f);
```
JohnLangford, line 59: On the next line (it won't let me comment there?) I don't think we should be normalize the probabilities.  In a multilabel context, the appropriate normalizer is the expected number of labels, not 1.

vowpalwabbit/oaa.cc
```diff
@@ -148,7 +148,7 @@ void predict(oaa& o, LEARNER::single_learner& base, example& ec)
       float sum_prob = 0;
       for (uint32_t i = 0; i < o.k; i++)
       {
-        ec.pred.scalars[i] = 1.f / (1.f + correctedExp(-o.pred[i].scalar));
+        ec.pred.scalars[i] = o.pred[i].scalar;
```
JohnLangford, line 151: Just to confirm: the combination of this and line 241 results in no change in test cases, yes?

PR #3378, opened 2021-10-08
build: add absl source dependency and make available to vw build

jackgerrits:
Because the old Windows solution still exists this PR includes a VERY hacky way to get it to work. It pulls in the required absl sources into the build. However, the required sources is determined by manually finding what is needed by the targets which are consumed.

Currently supported targets are `span` and `inlinedvector`. 

This should be temporary as the checked in solution is on its way out.


jackgerrits:
Going to close this until we converge on the single build system making this easier.

PR #3434, opened 2021-11-01
refactor: Use enum class instead of bare enum

### LINKED ISSUE #1855: Proposal: Change int-based enums to enum classes
Rather than using size_t with \#define'd values or const numbers, we should use enum classes, which are more readable and relatively type safe

jackgerrits:
resolves #1855


jackgerrits:
the attached issue is a long standing tech debt item. Tackling it now prior to 9 release.

PR #3435, opened 2021-11-01
feat: Output specific message when stdin is used to try avoid confusion

jackgerrits:


test/test-sets/ref/backwards.stderr
```diff
@@ -3,7 +3,7 @@ learning rate = 0.5
 initial_t = 0
 power_t = 0.5
 using no cache
-Reading datafile = 
+Reading from stdin since no datafile given
```
lalo, line 6: why not keep same format and say something like 
Reading datafile = None (stdin) or something like that? I guess it could get confusing if there is a file name called none

lalo (approved):

test/test-sets/ref/backwards.stderr
jackgerrits: Yeah keeping the same format might be good. I tried to make it clear why VW is sitting waiting as it is a point of confusion that has come up a number of time.s

PR #3438, opened 2021-11-01
feat: [automl] Add feature to clear configs on champ change

bassmang:


vowpalwabbit/automl.cc
```diff
@@ -403,7 +415,28 @@ void interaction_config_manager::update_champ()
       configs[config_index].state = VW::automl::config_state::Removed;
     }
   }
-  if (champ_change) { config_oracle(); }
+  if (champ_change)
+  {
+    if (keep_configs) { config_oracle(); }
```
lalo, line 420: nit: since both of those branches end with config_oracle() call, you could also have just one if(!keep_configs) {//content of the else } and then the config_oracle call.

bassmang: Changed

lalo (approved):

PR #3439, opened 2021-11-02
feat: make json parser usable without `all`

jackgerrits:


vowpalwabbit/vw.h
```diff
@@ -190,6 +191,13 @@ inline uint64_t chain_hash(vw& all, const std::string& name, const std::string&
       all.parse_mask;
 }
 
+inline uint64_t chain_hash_static(
```
lalo, line 194: could the automl_weights_test.cc over in the unit test dir be of any assistance here? Might be easy to cross-reference, but not sure

vowpalwabbit/parse_example_json.h
```diff
@@ -1628,14 +1637,16 @@ struct VWReaderHandler : public rapidjson::BaseReaderHandler<rapidjson::UTF8<>,
 {
   Context<audit> ctx;
 
-  void init(vw* all, v_array<example*>* examples, rapidjson::InsituStringStream* stream, const char* stream_end,
+  void init(const label_parser& lbl_parser, hash_func_t hash_func, uint64_t hash_seed, uint64_t parse_mask,
```
olgavrou, line 1640: this is getting big, wondering if we should define a specific passthrough struct for all these arguments

vowpalwabbit/vw.h
jackgerrits: Useful in what sense? This is just the function above but using explicit args and not the things nested in `all`

vowpalwabbit/parse_example_json.h
jackgerrits: Agreed, if you are okay with it it will come along in another PR

vowpalwabbit/vw.h
lalo: true, just to wire it to a test but seems pretty straightforward.

lalo (approved):

PR #3440, opened 2021-11-02
feat: expose holdout sum loss in python bindings

jackgerrits:
Added in response to a question on Gitter

### DIFF
--- a/python/pylibvw.cc
+++ b/python/pylibvw.cc
@@ -870,6 +870,7 @@ float get_loss(example_ptr ec) { return ec->loss; }
 float get_total_sum_feat_sq(example_ptr ec) { return ec->get_total_sum_feat_sq(); }
 
 double get_sum_loss(vw_ptr vw) { return vw->sd->sum_loss; }
+double get_holdout_sum_loss(vw_ptr vw) { return vw->sd->holdout_sum_loss; }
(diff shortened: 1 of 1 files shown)


python/pylibvw.cc
```diff
@@ -870,6 +870,7 @@ float get_loss(example_ptr ec) { return ec->loss; }
 float get_total_sum_feat_sq(example_ptr ec) { return ec->get_total_sum_feat_sq(); }
 
 double get_sum_loss(vw_ptr vw) { return vw->sd->sum_loss; }
+double get_holdout_sum_loss(vw_ptr vw) { return vw->sd->holdout_sum_loss; }
```
lalo, line 873: should we add ac heck for holdout_set_off just like https://github.com/VowpalWabbit/vowpal_wabbit/blob/700efac9eb94045d57407a475e8a4190d43f430a/vowpalwabbit/parse_args.cc#L1782

maybe even throw if theres no holdout?

olgavrou (approved):

python/pylibvw.cc
jackgerrits: I think if there is no holdout the sum of cost will just be zero based on where it is set in shared_data

lalo: yea not much to do since we can only return double, I was going to say we could return a python None in that case but hopefully whoever calls knows what they are doing.

PR #3441, opened 2021-11-02
feat: [automl] add oracle for all one-diff configs

bassmang:


jackgerrits (approved):

vowpalwabbit/automl.cc
```diff
@@ -596,6 +637,7 @@ VW::LEARNER::base_learner* automl_setup(VW::setup_base_i& stack_builder)
   std::string priority_type;
   int priority_challengers;
   bool keep_configs = false;
+  std::string oracle_type;
```
olgavrou, line 640: does this need to be persisted to the model during save_load?

olgavrou: oh never mind I can see it is kept via the args

vowpalwabbit/automl.cc
```diff
@@ -210,35 +239,47 @@ void interaction_config_manager::pre_process(const multi_ex& ecs)
...
-    namespace_index ns1 = champ_interactions[rand_ind][0];
-    namespace_index ns2 = champ_interactions[rand_ind][1];
-    std::map<namespace_index, std::set<namespace_index>> new_exclusions(
-        configs[scores[current_champ].config_index].exclusions);
-    new_exclusions[ns1].insert(ns2);
-    // Note that configs are never actually cleared, but valid_config_size is set to 0 instead to denote that
-    // configs have become stale. Here we try to write over stale configs with new configs, and if no stale
-    // configs exist we'll generate a new one.
-    if (valid_config_size < configs.size())
+    for (uint64_t i = 0; i < CONGIGS_PER_CHAMP_CHANGE; ++i)
```
olgavrou, line 244: nit: 
```suggestion
    for (uint64_t i = 0; i < CONFIGS_PER_CHAMP_CHANGE; ++i)
```

vowpalwabbit/automl.cc
```diff
@@ -202,6 +203,34 @@ void interaction_config_manager::pre_process(const multi_ex& ecs)
   }
 }
 
+// Helper function to insert new configs from oracle into map of configs as well as index_queue.
+// Handles creating new config with exclusions or overwriting stale configs to avoid reallocation.
+void interaction_config_manager::insert_config(
```
olgavrou, line 208: can this be `std::move`'d into here and then moved into the `.exclusions` field?

vowpalwabbit/automl.cc
```diff
@@ -202,6 +203,34 @@ void interaction_config_manager::pre_process(const multi_ex& ecs)
...
+    const std::map<namespace_index, std::set<namespace_index>>& new_exclusions)
+{
+  // Note that configs are never actually cleared, but valid_config_size is set to 0 instead to denote that
+  // configs have become stale. Here we try to write over stale configs with new configs, and if no stale
+  // configs exist we'll generate a new one.
+  if (valid_config_size < configs.size())
+  {
+    configs[valid_config_size].exclusions.clear();
```
olgavrou, line 216: `clear()` seems redundant since we are replacing the previous exclusion map

vowpalwabbit/automl.cc
bassmang: Good catch!

vowpalwabbit/automl.cc
bassmang: Yes, changed

vowpalwabbit/automl.cc
bassmang: Removed

vowpalwabbit/automl.cc
olgavrou: which formatter? The reason is that it's a needless copy since the `new_exclusions` object is being created outside of this function with the sole purpose of being stored in the configs right?

olgavrou: oh ok I see what you mean from your previous commit, in order to actually move the signature of the function needs to change too to:

`
void interaction_config_manager::insert_config(std::map<namespace_index, std::set<namespace_index>>&& new_exclusions)
`

bassmang: Right forgot the rvalue in the signature. Just added it back with some other moves

vowpalwabbit/automl.cc
```diff
@@ -202,6 +203,32 @@ void interaction_config_manager::pre_process(const multi_ex& ecs)
...
+    configs[valid_config_size].exclusions = std::move(new_exclusions);
+    configs[valid_config_size].lease = global_lease;
+    configs[valid_config_size].ips = 0;
+    configs[valid_config_size].lower_bound = std::numeric_limits<float>::infinity();
+    configs[valid_config_size].state = VW::automl::config_state::New;
+  }
+  else
+  {
+    exclusion_config conf(global_lease);
+    conf.exclusions = new_exclusions;
```
olgavrou, line 224:
```suggestion
    conf.exclusions = std::move(new_exclusions);
```

lalo (approved):

PR #3442, opened 2021-11-03
refactor: convert RunTests.tt to consume json test description

jackgerrits:


cs/unittest/RunTests.tt
```diff
@@ -6,121 +6,102 @@
 <#@ import namespace="System.Text" #>
 <#@ import namespace="System.Text.RegularExpressions" #>
 <#@ import namespace="System.Collections.Generic" #>
+<#@ assembly name="Newtonsoft.Json" #>
```
lalo, line 9: maybe do the expanded hintpath https://github.com/VowpalWabbit/vowpal_wabbit/blob/5954860bfdb57e722d13c068cb75478c372abedd/cs/unittest/cs_unittest.csproj#L113 ugly but might work

jackgerrits: its worth trying

lalo: now cmake fails and the other one passes :|

jackgerrits: That's not great...

cs/unittest/RunTests-cmake.tt
```diff
@@ -0,0 +1,200 @@
...
+
+var testRoot = Path.Combine(Path.GetDirectoryName(Host.TemplateFile), @"..\..\test");
+
+List<Test> tests;
+using (StreamReader sr = File.OpenText(Path.Combine(testRoot, "runtests.AUTOGEN.json")))
+{
+    tests = JsonConvert.DeserializeObject<List<Test>>(sr.ReadToEnd());
+}
+
+var skipList = new[] { 13, 32, 39, 258, 40, 259, 41, 260, 59, 60, 61, 66, 68, 90,
```
lalo, line 23: having two files with skip list will be interesting in the long term, and a bit confusing

jackgerrits: Yeah agreed. Chatted with jacob and have an idea how to merge the files

lalo: the newtonsoft dependency could be refactored out of the tt file, this might imply rewriting how this test file gets called

lokitoth (approved):
Well done finding a way to get it to work in both using one file

PR #3443, opened 2021-11-03
feat: add one_of method, standardize help output, Resolves #3318

### LINKED ISSUE #3318: feature request: options: add one_of(...) method
There’s a class of options, including `cb_type`, that should be defined as follows:
 
```
      .add(make_option("cb_type", type_string)
               .keep()
               .one_of(“ips”, “dr”, “mtr”)
               .help("contextual bandit method to use in {ips,dr,mtr}. Default: mtr"));
```
 
This way we can bake in a validation step on the parser from one of the pre-defined options. There’s a couple of other flags that are set between different pre-defined values. The help message can be auto-generated based on the one_of arguments.
 
The error would still need to be handled instead of doing the printing warning and picking a default.

current impl: we manually have to check if the user selected one of the correct options
https://github.com/VowpalWabbit/vowpal_wabbit/blob/40285770d918fb181ddff5d0a89c86b940b69d60/vowpalwabbit/cb_adf.cc#L503

update options_i to have one_of method:
https://github.com/VowpalWabbit/vowpal_wabbit/blob/40285770d918fb181ddff5d0a89c86b940b69d60/vowpalwabbit/options.h#L62

add logic to check that the user's input matches the known options ("ips", "dr", "mtr") fail if they dont match.
https://github.com/VowpalWabbit/vowpal_wabbit/blob/40285770d918fb181ddff5d0a89c86b940b69d60/vowpalwabbit/options_boost_po.cc#L145

bassmang:
Additions:
1) Include one_of method to specify possible options. This can handle sets of `string` or `int` type. It will throw a well-formatted error if a non-included option is given. It will also automatically add these choices to the help message.
2) Remove all `.`'s add end of current help messages for consistency
3) Create general format in help page capitalizing all first letters for each reduction, and `Options:` at the end.
4) Capitalize first letter of all help messages for consistency

Fixes:
1) Previous help message seemed wrong for `search_rollout` and `search_rollin`. It did not include all options checked for in search reduction, updated to include these options.
2) `--kernel` help message did not include `poly`

cb_type Handling:
There are reductions which use(d) cb_type. The following are their update choices:
1) cb_adf.cc: choices: `{"ips", "dm", "dr", "mtr", "sm"}`, default: `mtr`. This is consistent with old behavior.
2) cb_algs.cc: choices: `{"ips", "dm", "dr", "mtr", "sm"}`, default: `dr`. This is tricky. The reduction specifies that `{'ips','dm','dr'}` are the only valid choices, but there are valid stacks which pass down `mtr` from `cb_adf.cc` and don't use the flag. I think the best route here is to allow all choices.
3) cb_explore_adf_cover.cc: choices: `{"ips", "dr", "mtr"}`, default: `mtr`. This is consistent with old behavior.
4) cb_explore_adf_regcb.cc: choices: `{mtr}`, default: `mtr`. This is consistent with old behavior.
5) cb_explore_adf_squarecb.cc: choices: `{mtr}`, default: `mtr`. This is consistent with old behavior.
6) cb_to_cb_adf.cc: Removed cb_type flag since it wasn't being used.


test/train-sets/ref/cb_bad_one_of.stderr
```diff
@@ -0,0 +1,9 @@
+Error: Error: 'bad_type' is not a valid choice for option --cb_type
```
jackgerrits, line 1: Can the error say what are valid choices?

bassmang: Yea I can add that in

jackgerrits:
Different places which use `cb_type` support only a subset of the types at each location. We should be consistent with old behavior or double check that the change to allowing all is valid.

vowpalwabbit/cb_explore_adf_squarecb.cc
```diff
@@ -351,7 +351,9 @@ base_learner* setup(VW::setup_base_i& stack_builder)
                .help("Upper bound on cost. Only used with --elim"))
       .add(make_option("cb_type", type_string)
                .keep()
-               .help("contextual bandit method to use in {ips,dr,mtr}. Default: mtr"));
+               .default_value("mtr")
+               .one_of({"ips", "dm", "dr", "mtr", "sm"})
```
lalo, line 355: Similar to Jack's comment, in this case this reduction only works with mtr and if you scroll down in this file you can see it's handled. Not sure if the right approach is to only mention "mtr" here in one_of. John might have a stronger opinion. For the rest of the reductions, just make sure that after parsing you take a look at how this arg gets handled, to see if it was a case of wrong help message or correct.

vowpalwabbit/mwt.cc
```diff
@@ -210,7 +210,11 @@ void save_load(mwt& c, io_buf& model_file, bool read, bool text)
     policy_data& pd = c.evals[policy];
     if (read) msg << "evals: " << policy << ":" << pd.action << ":" << pd.cost << " ";
     bin_text_read_write_fixed_validated(
-        model_file, reinterpret_cast<char*>(&c.evals[policy]), sizeof(policy_data), read, msg, text);
+        model_file, reinterpret_cast<char*>(&c.evals[policy].cost), sizeof(double), read, msg, text);
```
lalo, line 213: interesting that this got surfaced in this PR

cs/unittest/RunTests.tt
```diff
@@ -16,7 +16,7 @@ var lines = File.ReadAllLines(Path.Combine(testRoot, "RunTests"))
     .ToList();
 
 var skipList = new[] { 13, 32, 39, 258, 40, 259, 41, 260, 59, 60, 61, 66, 68, 90,
-	25, 26, 349, 350, 356, // crash
+	25, 26, 349, 350, 356, 357, 358, // crash
```
lalo, line 19: this has interesting implications for consumers, since before we would not crash, and default to X

vowpalwabbit/mwt.cc
bassmang: This issue was somewhat mysterious. I saw there were valgrind errors in the current version on master which werent showing up in the python script, not sure where that started. Splitting it up into the three members of the struct gets rid of the issue.

olgavrou: does any CI pipeline catch the valgrind errors you are talking about?

bassmang: Yea I actually just tried running the python test locally on master:
`python3 run_tests.py -f -j 2 -t 123 --include_flatbuffers --valgrind --clean_dirty`
and it is throwing an error. Not sure why the CI isn't seeing this

olgavrou: ok we should really really fix that then, can you open a tracking issue with some details?

bassmang: https://github.com/VowpalWabbit/vowpal_wabbit/issues/3445

vowpalwabbit/cb_adf.cc
```diff
@@ -491,16 +491,18 @@ base_learner* cb_adf_setup(VW::setup_base_i& stack_builder)
...
                .keep()
                .default_value(0.f)
-               .help("Clipping probability in importance weight. Default: 0.f (no clipping)."))
+               .help("Clipping probability in importance weight. Default: 0.f (no clipping)"))
       .add(make_option("cb_type", type_string)
                .keep()
-               .help("contextual bandit method to use in {ips, dm, dr, mtr, sm}. Default: mtr"));
+               .default_value("mtr")
+               .one_of({"ips", "dm", "dr", "mtr", "sm"})
+               .help("Contextual bandit method to use"));
```
lalo, line 505: default value got chopped from help message

vowpalwabbit/options.h
```diff
@@ -53,9 +56,33 @@ struct option_builder
     return *this;
   }
 
+  template <typename U>
+  void add_help(std::ostringstream&, const U&)
+  { /* cannot handle non-string or arithmetic types */
```
lalo, line 61: maybe worth throwing?

vowpalwabbit/options.h
```diff
@@ -124,12 +158,38 @@ struct typed_option : base_option
...
 
+  template <typename U>
+  std::string invalid_choice_error(const U&)
+  {
+    return "";
+  }
+  std::string invalid_choice_error(const std::string& value)
+  {
+    std::ostringstream err_msg;
+    err_msg << fmt::format("Error: '{}' is not a valid choice for option --{}", value, m_name);
```
lalo, line 169: I think when throwing you already print this information, check stderr file for duplication of same message

vowpalwabbit/automl.cc
```diff
@@ -597,27 +597,32 @@ VW::LEARNER::base_learner* automl_setup(VW::setup_base_i& stack_builder)
...
+      .add(make_option("cm_type", cm_type)
+               .keep()
+               .default_value("interaction")
+               .one_of({"interaction"})
+               .help("Set type of config manager"))
       .add(make_option("priority_type", priority_type)
                .keep()
                .default_value("none")
-               .help("set function to determine next config {none, least_exclusion}"))
+               .one_of({"none", "least_exclusion"})
```
lalo, line 619: we should be able to delete the if at 631 in this file

vowpalwabbit/automl.cc
```diff
@@ -597,27 +597,32 @@ VW::LEARNER::base_learner* automl_setup(VW::setup_base_i& stack_builder)
...
                .default_value(3)
-               .help("set number of live configs"))
+               .help("Set number of live configs"))
       .add(make_option("global_lease", global_lease)
                .keep()
                .default_value(10)
-               .help("set initial lease for automl interactions"))
-      .add(make_option("cm_type", cm_type).keep().default_value("interaction").help("set type of config manager"))
+               .help("Set initial lease for automl interactions"))
+      .add(make_option("cm_type", cm_type)
```
lalo, line 611: this is not being used right?

vowpalwabbit/cb_adf.cc
bassmang: Yea I'm not certain if it's necessary to keep these explicit default messages since theyre auto-generated in help.stdout

vowpalwabbit/automl.cc
bassmang: Not really, `"interaction"` is the only choice here

vowpalwabbit/automl.cc
bassmang: We still need the `if` at 631 to set the function pointer based on the input, but we can delete the `THROW` after the `else if`

vowpalwabbit/options.h
bassmang: This part generates the message, there's only one `THROW` for this message on 362 in options.h, not sure why its showing up twice

vowpalwabbit/mwt.cc
jackgerrits: I'm pretty sure what is going on here is that the struct itself has padding in some environments and is tightly packed elsewhere. The compiler version in CI probably just doesn't do it.

The valgrind issue repros for me locally, and so it is complaining that the packing bytes are understandably uninitialized. So, by splitting it up and writing each member it will always work.

You can force tight packing with `__attribute__ ((packed))` and doing so fixes the valgrind issue. The size of the struct when tightly packed is 13 as you would expect and without it is 16, so theres 3 padding bytes in there.

For structures we serialize we need to be careful about this memory alignment because if one system uses padding and the other doesn't then the model just wont load. It's a good thing this was found! I think the right fix in this case is to serialize the members separately as was done, but I'm not sure how we can check for this in future other than through valgrind

bassmang: I updated the issue above noting that its an issue with struct padding. I'll make a separate PR trying to find other places where this issue's occurring

vowpalwabbit/cats_tree.cc
```diff
@@ -340,16 +340,16 @@ base_learner* setup(setup_base_i& stack_builder)
...
   uint32_t num_actions;  // = K = 2^D
   uint32_t bandwidth;    // = 2^h#
   std::string link;
   new_options.add(make_option("cats_tree", num_actions).keep().necessary().help("CATS Tree with <k> labels"))
       .add(make_option("tree_bandwidth", bandwidth)
                .default_value(0)
                .keep()
-               .help("tree bandwidth for continuous actions in terms of #actions"))
-      .add(make_option("link", link).keep().help("Specify the link function: identity, logistic, glf1 or poisson"));
+               .help("Tree bandwidth for continuous actions in terms of #actions"))
```
jackgerrits, line 351: The check for glfw below is redundant now and cant be reached. Perhaps adding the explanation to the help string here would be useful and then removing the check below?

vowpalwabbit/cb_explore_adf_regcb.cc
```diff
@@ -270,12 +270,14 @@ base_learner* setup(VW::setup_base_i& stack_builder)
...
-      .add(make_option("cb_max_cost", max_cb_cost).keep().default_value(1.f).help("upper bound on cost"))
+      .add(make_option("cb_min_cost", min_cb_cost).keep().default_value(0.f).help("Lower bound on cost"))
+      .add(make_option("cb_max_cost", max_cb_cost).keep().default_value(1.f).help("Upper bound on cost"))
       .add(make_option("first_only", first_only).keep().help("Only explore the first action in a tie-breaking event"))
       .add(make_option("cb_type", type_string)
                .keep()
-               .help("contextual bandit method to use in {ips,dr,mtr}. Default: mtr"));
+               .default_value("mtr")
+               .one_of({"mtr"})
+               .help("Contextual bandit method to use"));
```
jackgerrits, line 280: Same here. Explaining why there is only one option seems important

vowpalwabbit/cb_explore_adf_rnd.cc
```diff
@@ -270,18 +270,18 @@ base_learner* setup(VW::setup_base_i& stack_builder)
...
-      .add(make_option("epsilon", epsilon).keep().allow_override().help("minimum exploration probability"))
-      .add(make_option("rnd", numrnd).keep().necessary().help("rnd based exploration"))
+      .add(make_option("epsilon", epsilon).keep().allow_override().help("Minimum exploration probability"))
+      .add(make_option("rnd", numrnd).keep().necessary().help("Rnd based exploration"))
       .add(make_option("rnd_alpha", alpha)
                .keep()
                .allow_override()
                .default_value(0.1f)
-               .help("ci width for rnd (bigger => more exploration on repeating features)"))
+               .help("Ci width for rnd (bigger => more exploration on repeating features)"))
```
jackgerrits, line 279:
```suggestion
               .help("CI width for rnd (bigger => more exploration on repeating features)"))
```

vowpalwabbit/cb_explore_adf_squarecb.cc
```diff
@@ -351,7 +351,9 @@ base_learner* setup(VW::setup_base_i& stack_builder)
                .help("Upper bound on cost. Only used with --elim"))
       .add(make_option("cb_type", type_string)
                .keep()
-               .help("contextual bandit method to use in {ips,dr,mtr}. Default: mtr"));
+               .default_value("mtr")
+               .one_of({"mtr"})
+               .help("Contextual bandit method to use"));
```
jackgerrits, line 356: Here too

vowpalwabbit/options.h
```diff
@@ -53,9 +56,34 @@ struct option_builder
...
+        if (first) { first = false; }
+        else
+        {
+          help_w_additions << ", ";
+        }
+        add_help(help_w_additions, v);
+      }
+      help_w_additions << "}";
+    }
+    m_option_obj.m_help = help_w_additions.str();
```
jackgerrits, line 86: This could be replaced with something like:

```c++
m_option_obj.m_help = fmt::format("{}. Choices {{{}}}", fmt::join(m_option_obj.one_of(), ", ");
```

vowpalwabbit/options.h
```diff
@@ -124,12 +159,38 @@ struct typed_option : base_option
...
+    {
+      if (first) { first = false; }
+      else
+      {
+        err_msg << ", ";
+      }
+      err_msg << v;
+    }
+    err_msg << "}";
+    return err_msg.str();
```
jackgerrits, line 183: Should also be able to be replaced by fmt::join

vowpalwabbit/options.h
```diff
@@ -139,6 +200,10 @@ struct typed_option : base_option
     THROW("typed_option does not contain value. use value_supplied to check if value exists.")
   }
 
+  void set_one_of(const std::set<value_type>& one_of_set) { m_one_of = one_of_set; }
+
+  std::set<value_type> one_of() { return m_one_of; }
```
jackgerrits, line 205:
```suggestion
  const std::set<value_type>& one_of() const { return m_one_of; }
```

vowpalwabbit/options.h
```diff
@@ -288,14 +354,24 @@ struct option_group_definition
     return check_if_all_necessary_enabled;
   }
 
+  // will check if one_of condition is met for all options
+  bool check_one_of() const
+  {
+    for (const auto& option : m_options)
+    {
+      if (option->m_one_of_err.length() > 0) { THROW(option->m_one_of_err); }
```
jackgerrits, line 362:
```suggestion
      if (!option->m_one_of_err.empty()) { THROW(option->m_one_of_err); }

```

vowpalwabbit/mwt.cc
olgavrou: @jackgerrits should we be determining the packing with the attribute you suggested? If we just initialize what we need to won't that be enough?

vowpalwabbit/cb_explore_adf_regcb.cc
bassmang: Also added for squarecb

jackgerrits (approved):

PR #3446, opened 2021-11-05
feat: [py] add git commit hash to library version

lalo:


jackgerrits:
The package version is available in the version module https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/vowpalwabbit/version.py

lalo:
@jackgerrits does that one show the git commit? that's what I'm after, for unofficial builds that share the same vw version of latest CI

jackgerrits:
No it doesn't, just the triplet. Probably makes sense to try and put the git commit here to keep things consistent

olgavrou (approved):

jackgerrits (approved):

PR #3453, opened 2021-11-09
test: make --fuzzy_compare the default

jackgerrits:
In the majority of use cases of the test script fuzzy comparing floats is the intended behavior. To make it a little simpler to invoke the script make fuzzy compare the default and support disabling it through `--fuzzy_compare false`


jackgerrits:
Going to

PR #3454, opened 2021-11-09
test: migrate `make test` to use run_tests.py

jackgerrits:
Requires #3450


test/CMakeLists.txt
```diff
@@ -3,24 +3,20 @@ add_subdirectory(tools)
...
+    COMMAND python3 run_tests.py --fuzzy_compare --exit_first_fail --epsilon 0.001 --clean_dirty --extra_options=--onethread ${FB_FLAG}
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
   )
 
   add_test(
     NAME RunTests_pass_2
-    COMMAND
-      ${CMAKE_COMMAND} -E env PATH=${CMAKE_CURRENT_BINARY_DIR}/../vowpalwabbit/:$ENV{PATH}
-      ./RunTests -d -fe -E 0.001 ${FB_FLAG}
+    COMMAND python3 run_tests.py --fuzzy_compare --exit_first_fail --epsilon 0.001 --clean_dirty ${FB_FLAG}
```
olgavrou, line 19: should we force a clean dirty here?

jackgerrits: No we shouldn't but we have to for now.

As soon as we are solely relying on the json file we can update the `run_test.py` and the test spec to properly sandbox the bash based tests. Then once that is done the test directory should remain clean after a run and we can remove `--clean_dirty`. Then a dirty directory will become a bug we can investigate.

jackgerrits: Because this runs two tests one after another in this `make test` setup we must either do clean or ignore here. Once the tests do not affect the test dir anymore then we can remove it.

I opted to clean instead of ignore as if we ignore the result of the test could differ based on files that are left around.

olgavrou: Yeah that's the biggest thing I am concerned about, deleting some ones files without any warning can be pretty catastrophic depending on what they have in that dir which we can't really know. What we can do is not run `ctest` in the CI but run `ctest -VV -R <name of test>`, and add a git clean dirty in the CI script vs in the actual ctest?

olgavrou (approved):

test/CMakeLists.txt
jackgerrits: yep agreed lets avoid even risking it. Changed to `--ignore_dirty`

PR #3457, opened 2021-11-11
fix: fix flat_example save_load by handling all components

bassmang:
Ongoing effort to address this issue:
https://github.com/VowpalWabbit/vowpal_wabbit/issues/3445


vowpalwabbit/kernel_svm.cc
```diff
@@ -881,3 +882,12 @@ VW::LEARNER::base_learner* kernel_svm_setup(VW::setup_base_i& stack_builder)
 
   return make_base(*l);
 }
+
+/*namespace VW
```
jackgerrits, line 886: Looks like it is not needed

vowpalwabbit/model_utils.h
```diff
@@ -97,6 +99,64 @@ size_t write_model_field(io_buf& io, const T& var, const std::string& name_or_re
...
+{
+  size_t bytes = 0;
+  uint64_t str_size;
+  bytes += read_model_field(io, str_size);
+  std::ostringstream os;
+  for (uint64_t i = 0; i < str_size; ++i)
+  {
+    char c;
+    bytes += read_model_field(io, c);
+    os << c;
```
jackgerrits, line 139: Rather than reading one by one you could use the c-style string constructor

vowpalwabbit/example.cc
```diff
@@ -306,6 +307,73 @@ void return_multiple_example(VW::workspace& all, v_array<example*>& examples)
   examples.clear();
 }
 
+namespace model_utils
+{
+size_t read_model_field(io_buf& io, polylabel& poly)
```
jackgerrits, line 312: This sort of introduces a parallel and incompatible way of serializing labels (`label_parser.cache-label`), It might be good to consider if we can reuse that instead

jackgerrits (commented):
Are you sure this change is binary compatible with the previous code (minus potential padding bits)?

vowpalwabbit/kernel_svm.cc
bassmang: Removed

vowpalwabbit/model_utils.h
bassmang: Do you mean making another overload to read_model_field for c-style strings?

jackgerrits: It's pretty inefficient to serialize/deserialize a string one character at a time. `bin_read_fixed` or something like that seems appropriate here. Since you know the size already you can just memcpy the contents

bassmang: Makes sense, updated

bassmang:
Resolved by separate PRs

PR #3464, opened 2021-11-12
fix: fix crash in daemon mode children, deadlock in endpass and add logging for any future crashes

jackgerrits:


vowpalwabbit/parse_dispatch_loop.h
```diff
@@ -40,6 +40,8 @@ void parse_dispatch(VW::workspace& all, DispatchFuncT& dispatch)
         all.example_parser->lbl_parser.default_label(examples[0]->l);
         examples[0]->end_pass = true;
         all.example_parser->in_pass_counter = 0;
+        // Since this example gets finished, we need to keep the counter correct.
+        all.example_parser->num_setup_examples++;
```
olgavrou, line 44: is this correct? my understanding is that if we come into the else statement then the example is one we don't learn on so we don't count it (i.e. it's end_pass or newline)

jackgerrits: The other side of this is a counter for the number of examples which were finished. And since the end pass example is finished we must increment this counter or we sit waiting on a condition variable whose condition will never be satisfied.

olgavrou: is this only for deamon mode?

jackgerrits: Yep, daemon mode is the only thing which consumes this counter

olgavrou (approved):

PR #3465, opened 2021-11-12
test: remove RunTests, update docs, refer to json file directly

jackgerrits:


test/core.vwtest.json
```diff
@@ -4221,5 +4221,70 @@
...
+    ]
+  },
+  {
+    "id": 362,
+    "desc": "--probabilities --oaa mixed-indexing",
+    "vw_command": "-d train-sets/probabilities_mixed_ind.dat --probabilities --oaa=4 --loss_function=logistic -p oaa_mixed_probabilities.predict",
+    "diff_files": {
+      "stderr": "train-sets/ref/oaa_mixed_probabilities.stderr",
+      "stdout": "train-sets/ref/oaa_mixed_probabilities.stdout",
+      "oaa_mixed_probabilities.predict": "pred-sets/ref/oaa_mixed_probabilities.predict"
```
bassmang, line 4284: nit: It may be easier to rename this field to "predict" then extrapolate the file name from the value

jackgerrits: The key of this mapping is the filename in the test dir, that is the file it searches for to compare to the ref file which is the value. stderr and stdout are special values. Everything else is just a file name

bassmang: Yes I meant we could use the ref file to extrapolate the file we search for, assuming they both have the same name

jackgerrits: I feel like it is easier to be explicit here, otherwise we have some implicit mapping between the filename produced by the test and the file it is being compared to

bassmang (approved):

test/runtests_parser.py
```diff
@@ -1,293 +0,0 @@
...
-            )
-        self.is_bash_command = True
-        self.add_vw_command(line)
-
-    def clean(self):
-        if not self.is_bash_command:
-            self.vw_command = " ".join(self.vw_command.split()[1:])
-
-            # check what input files this command needs
-            input_file_flags = [
```
lalo, line 77: is this processing done elsewhere now? this and checking for missing and repeat id numbers?

jackgerrits: Input files are up to the test author. I think checking for missing and duplicate ids is missing, I'll add that in

lalo: input files and test dependency could still be aided by a script, or at least as a verification step

jackgerrits: You can't depend on it for bash_command tests. I guess you could make it automatic for vw_command tests and explicit for bash_command tests. I initially didn't do that as it felt inconsistent. It would reduce effort in adding new tests though.

Issue #439, opened 2014-11-06
reporting "current features" with shared features


martinpopel:
[shared](http://www.umiacs.umd.edu/~hal/tmp/multiclassVW.html) features are very helpful with `csoaa_ldf`, but the number of "current features" reported by VW is broken.

```
echo '1:0 |T a |S x1 x2 x3
2:1 |T b |S x1 x2 x3

' | vw --csoaa_ldf=m -qST
```

Repeating the shared features on each line manually: reports "current features" = 7 and "total feature number" = 14.

```
echo 'shared |S x1 x2 x3
1:0 |T a
2:1 |T b

' | vw --csoaa_ldf=m -qST
```

Using the "shared" keyword works the same, but the reported number of  "current features" = 1 and "total feature number" = 2.

I would think the two examples above should be equivalent.

Moreover, the "current features" should be equal to the "total feature number" when there is just one (though multiline) example in the dataset ("number of examples per pass" = 1).
- [x] "current features" now include all lines of an ldf example (not just the first line)
- [x] "current features" now include also "shared" features in ldf examples
- [ ] "current features" don't include --quadratic and --cubic features where one namespace is "shared", because VW::setup_example is called before the namespaces from "shared" are added via add_example_namespaces_from_example.

JohnLangford:
Martin, do you see a fix for this?  Seems difficult.

JohnLangford:
Maybe the right thing to do is to move the calculation of the number of features to bottom of the stack rather than the top (as it is at present)?

martinpopel:
Yes, I didn't see any easy solution, so I left this. And yes, moving the calculation in the stack may help, but I was not sure if this does not break other code or make something slower. This is a low priority issue. One solution is just detect shared ldf features and --quadratic or --cubic and report "?" as the number of current features. (I am quite busy next few weeks/months.)

JohnLangford:
The number of features is mostly just used for reporting purposes so moving
it to the bottom of the stack is probably fine w.r.t. the final printout.

However, the intermediate printouts are less clear.  These are typically
handled at the top of the stack, and they report on the example at the top
of the stack for debugging purposes.  In some sense, _all_ the "examples"
which make up an LDF example should be lumped together and reported on
simultaneously.

-John

JohnLangford:
How does this issue interact with Alex's interactionis patch?

martinpopel:
The problem is still here, just with different numbers (because I fixed two of the three issues).

[...]

PR #3470, opened 2021-11-17
fix: cb feature count off for shared

bassmang:


olgavrou (approved):

jackgerrits:
It looks like its possible this might fix this issue: https://github.com/VowpalWabbit/vowpal_wabbit/issues/439

bassmang:
> It looks like its possible this might fix this issue: #439

@jack it does seem to fix it for cb, but the feature count still still off for ccb (and likely other reductions) so not sure if we can close it

PR #3478, opened 2021-11-17
feat: Allow options_boost_po to be created from a command line string directly

jackgerrits:


jackgerrits:
Requiring usage of an escaped string seems to not really be a viable solution based on the many ways this is used.

PR #3479, opened 2021-11-18
fix: fix ccb feature count, remove unnecessary slot copy, fix message

bassmang:


vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -489,6 +489,16 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
         save_action_scores(data, decision_scores);
         VW_DBG(examples) << "ccb "
                          << "slot:" << slot_id << " " << ccb_decision_to_string(data) << std::endl;
+        for (const auto& ex : data.cb_ex)
```
olgavrou, line 492: since this is done during learn_or_predict, won't the shared features be merged into the actions and slots already by the shared_feature_merger?

bassmang: Yes they will be. This is basically a new way to count features for ccb. The feature count of each slot will include all single features and interactions for each action that slot processes

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -489,6 +489,16 @@ void learn_or_predict(ccb& data, multi_learner& base, multi_ex& examples)
...
+        {
+          if (CB::ec_is_example_header(*ex)) { slot->num_features = (data.cb_ex.size() - 1) * ex->num_features; }
```
lalo, line 494: I'm not sure any example will return true here if this gets run after shared_feature_merger, the previous PR made sense bc it was osn the finish path and that one gets run after shared_feature_merger reverts the merge but that shouldn't apply here in predict/learn

olgavrou (approved):

PR #3487, opened 2021-11-29
ci: Add explicit pandas install for benchmark CI

jackgerrits:
Benchmark CI is currently failing with a missing dependency


olgavrou (approved):

olgavrou:
Do we have any idea why the Linux CI is failing out of the blue?

jackgerrits:
gbench had a code change which added pandas. The root requirements.txt was updated but not the one in the tools directory. It may be worth just consuming the root requirements file to avoid this.

jackgerrits:
It's a flaky test

PR #3493, opened 2021-12-01
refactor: disable prediction with ccb and cb_type mtr

bassmang:


vowpalwabbit/cb_adf.cc
```diff
@@ -261,6 +263,11 @@ void cb_adf::learn_MTR(multi_learner& base, multi_ex& examples)
   examples[_gen_cs.mtr_example]->num_features = nf;
   examples[_gen_cs.mtr_example]->weight = old_weight;
   std::swap(_gen_cs.mtr_ec_seq[0]->pred.a_s, _a_s_mtr_cs);
+  if (_ccb)
```
jackgerrits, line 266: It is illegal for this reduction to set a label other than its specified output label type.

vowpalwabbit/cb_adf.cc
```diff
@@ -485,6 +492,7 @@ base_learner* cb_adf_setup(VW::setup_base_i& stack_builder)
   bool rank_all;
   float clip_p;
   bool no_predict;
+  bool ccb = false;
```
jackgerrits, line 495: This reduction knowing about ccb is a big break of encapsulation.

olgavrou: Would it make sense to use the same condition as `conditional_contextual_bandit` setup, i.e. `if (options.was_supplied("quiet") && !options.was_supplied("p")) { data->no_pred = true; }` and have a no-pred boolean here too?

vowpalwabbit/cb_adf.cc
```diff
@@ -261,6 +263,11 @@ void cb_adf::learn_MTR(multi_learner& base, multi_ex& examples)
   examples[_gen_cs.mtr_example]->num_features = nf;
   examples[_gen_cs.mtr_example]->weight = old_weight;
   std::swap(_gen_cs.mtr_ec_seq[0]->pred.a_s, _a_s_mtr_cs);
+  if (_ccb)
+  {
+    examples[0]->l.conditional_contextual_bandit.explicit_included_actions.clear();
```
olgavrou, line 268: Could re-use `reduction_features` to communicate extra information between reductions that don't fit in the label or prediction and explicitly extend `ccb_reduction_features.h` to hold a field specifically for this

olgavrou (commented):
Can we add a test to exercise this code?

vowpalwabbit/cb_adf.cc
jackgerrits: I would say that modifying the reduction features of a problem type other than yourself is also a break of label/prediction layering rules

olgavrou: hm so then the only way is via the `pred.a_s` but somehow making it clear that it's not a prediction?

jackgerrits: Or by returned a different prediction type when this mode is enabled

vowpalwabbit/cb_adf.cc
bassmang: Yes I just had this here to see the perf impact, I removed all changes to cb_adf in this PR

vowpalwabbit/cb_adf.cc
bassmang: It can all be handled in ccb so I removed this as well

vowpalwabbit/cb_adf.cc
bassmang: This step isn't necessary and actually messed with some other CCB functionality. The cb labels and predictions are already in ccb so we dont need to do anything here

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -211,6 +212,26 @@ void save_action_scores(ccb& data, decision_scores_t& decision_scores)
   data.shared->pred.a_s.clear();
 }
 
+void remove_labeled_action(ccb& data, multi_ex& examples)
+{
+  int32_t action_index = -1;
+  for (size_t i = 0; i < examples.size(); i++)
+  {
+    CB::label& ld = examples[i]->l.cb;
```
jackgerrits, line 220:
```suggestion
    const CB::label& ld = examples[i]->l.cb;
```

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -211,6 +212,26 @@ void save_action_scores(ccb& data, decision_scores_t& decision_scores)
   data.shared->pred.a_s.clear();
 }
 
+void remove_labeled_action(ccb& data, multi_ex& examples)
```
jackgerrits, line 215:
```suggestion
void remove_labeled_action(ccb& data, const multi_ex& examples)
```

jackgerrits (approved):

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -211,6 +212,26 @@ void save_action_scores(ccb& data, decision_scores_t& decision_scores)
   data.shared->pred.a_s.clear();
 }
 
+void remove_labeled_action(ccb& data, const multi_ex& examples)
```
olgavrou, line 215:
```suggestion
void add_chosen_action_to_exclude_list(ccb& data, const multi_ex& examples)
```

or something similar, remove implies that something is actually removed, i.e. that this method has side effects

olgavrou (approved):
approve with a suggestion

PR #3495, opened 2021-12-01
build: Remove tox

jackgerrits:
- Remove tox as it is deprecated and doesn't work


lalo (approved):

setup.py
```diff
@@ -62,213 +70,154 @@ def get_ext_filename_without_platform_suffix(filename):
...
-    distclass=Distribution,
-    cmdclass={
-        'build_ext': BuildPyLibVWBindingsModule,
-        'clean': Clean,
-        'sdist': Sdist,
-        'test': Tox,
-        'install_lib': InstallLib
-    },
-    # tox.ini handles additional test dependencies
-    tests_require=['tox'],
```
lalo, line 273: what's the deal with the checked in file tox.ini?

setup.py
```diff
@@ -62,213 +70,154 @@ def get_ext_filename_without_platform_suffix(filename):
...
+    author_email="scott.d.graham@gmail.com",
+    license="BSD 3-Clause License",
     classifiers=[
-        'Development Status :: 4 - Beta',
-        'Intended Audience :: Science/Research',
-        'Topic :: Scientific/Engineering',
-        'Topic :: Scientific/Engineering :: Information Analysis',
-        'License :: OSI Approved :: BSD License',
-        'Programming Language :: Python :: 3',
+        "Development Status :: 4 - Beta",
```
lalo, line 208: is is time to remove this one?

setup.py
jackgerrits: Tox from what I can tell is a testing framework for testing different python versions. We get this testing coverage through the CI matrix and tox hasn't been used for a long time. I think it is safe to remove

PR #3497, opened 2021-12-02
build: update instructions for building macos wheels

### LINKED ISSUE #3491: Provide Python binary wheels for MacOS arm64
We don't currently upload or produce binary wheels for MacOS M1 arm resulting in the installer falling back to source builds.

jackgerrits:
Wheels for 3.8 and 3.9 are uploaded. 3.10 is not yet doable. Closes #3491


python/PACKAGE.rst
```diff
@@ -10,6 +10,7 @@ PyPI Deployment Process
     * `Linux`_
     * `Windows`_
     * `MacOS`_
+        * CI doesnt produce `arm64` wheels. Use build_python_wheels.sh_ to build them locally on an ARM based MacOS device.
```
lalo, line 13: did you test this on the m1?

lalo (approved):

python/PACKAGE.rst
jackgerrits: Yep

PR #3498, opened 2021-12-02
ci: Build Python 3.10 wheels for Windows

jackgerrits:
There are two things bundled up in this PR.
1. Migration from windows-2016 to windows-2019 for build images
2. Add Python 3.10 CI

For `1.`, this is required since Github will be retiring the `windows-2016` environment in March 2022. This change has a rather drastic impact on our CI. `windows-2016` had VS2017, but `windows-2019` removes this and forces usage of VS2019. 

Older `vcpkg` does not build on newer Visual studio. So this required an upgrade of `vcpkg` and a rethinking of how to pin the required python versions.

Additionally, the `windows-2019` image uses the Win11 SDK and Python 3.6, 3.7 and 3.8 require patching to be able to work with that SDK. This is why custom overlay ports had to be added for those versions.


peterychang (approved):
why vcpkg... just why

Issue #3163, opened 2021-07-15
`--save_resume` to become default for model saving and loading


jackgerrits:
When a model is saved currently the default is just to store just the model weights. A flag `--save_resume` is available which additionally saves the trainer state which allows you to continue training additional training state. This is required when resuming for online learning.

We have noticed over time that most users expect to be able to resume training from a saved model and would therefore need to supply the `--save_resume` flag. This is unintuitive for most users. We will be changing default behavior to be as if `--save_resume` were passed, unless `--save_bare` is supplied in which case the behavior will be the old default logic.

The extra state for save resume does increase model size, approximately 2 or 3 orders of magnitude. However, VW models are already very small often in the order of single digit megabytes. It is unlikely to matter practically.

We believe this to be a positive change without any major drawbacks, but it is important to consider this fact if you are currently using VW to initialize weights for a batch training scenario.

## When would I want to use `--save_bare`?
The extra trainer state is not needed for predictions. Therefore, if the model is for predictions only then using `--save_bare` may make sense. Do note thought that the extra trainer state will have no negative impact on predictions so this is only necessary if you want to reduce model size.

If you use VW in a batch setting and only load a model with the intention of initializing weights then the extra trainer state may not be desired as it affects the rate of learning per parameter. In practice, this seems to be insignificant to training performance though. 

## When does this take effect?
The default will change in the next release of VW, 9.0.0. But as always, we opened this issue for discussion in case there are issues with the plan.

rajan-chari:
Instead of 2 or 3 orders of magnitude -- assuming you meant 2x - 3x?

jackgerrits:
I was just repeated what John said, but I do agree. The extra space should be in the ~3 or 4x and not orders of magnitude

lokitoth:
I suspect this will depend a little on the base learner and algorithm.

Most differences are all contained to the GD::save_load_regressor/GD::save_load_online_state. There the number of additional values stored depends on the true base learner / algorithm:

| Base Learner / Variant | Extra Floats Per Weight | Notes |
| -----------:------------ | ------------------------- | :------- |
| GD - adaptive+normalized | 2 | |
| GD - adaptive | normalized | 1 | |
| GD - plain                            | 0 | |
| OjaNewton                          | 0 | |
| SVRG                                   | 0 | Note that SVRG requests extra space for parameter always, but does not seem to do anything extra special in save-resume |
| FTRL - Proximal-FTRL          | 2 | |
[...]

PR #3502, opened 2021-12-02
feat!: Remove --save_resume and add --predict_only_model. Resolves: #3163

bassmang:
feat: Remove save_resume and add save_bare. Resolves: VowpalWabbit#3163


jackgerrits:
What is the expected behavior when a user loads a 8.11.0 model which has save resume? And if it doesnt?

bassmang:
@jackgerrits both of those models should be loaded in and have the same results as before. A new model with no flag would be the same as an old model with --save_resume, and a new model with --save_bare would be the same as an old model with no flag.

jackgerrits:
Excellent, that sounds reasonable

olgavrou:
> @jackgerrits both of those models should be loaded in and have the same results as before. A new model with no flag would be the same as an old model with --save_resume, and a new model with --save_bare would be the same as an old model with no flag.

what about a new model with no `--save_bar` (i.e. it has `save_resume` on) would it load fine and continue learning on an old vw instance?

bassmang:
> > @jackgerrits both of those models should be loaded in and have the same results as before. A new model with no flag would be the same as an old model with --save_resume, and a new model with --save_bare would be the same as an old model with no flag.
> 
> what about a new model with no `--save_bar` (i.e. it has `save_resume` on) would it load fine and continue learning on an old vw instance?

Yes this works fine too and can continue learning -- just tested locally

olgavrou (approved):

vowpalwabbit/parse_args.cc
```diff
@@ -1143,7 +1148,12 @@ void parse_output_model(options_i& options, VW::workspace& all)
   if (!all.final_regressor_name.empty() && !all.logger.quiet)
     *(all.trace_message) << "final_regressor = " << all.final_regressor_name << endl;
 
-  if (options.was_supplied("invert_hash")) all.hash_inv = true;
+  if (options.was_supplied("invert_hash")) { all.hash_inv = true; }
+  if (save_resume)
+  {
+    logger::errlog_warn("--save_resume flag is deprecated -- learning can now continue on saved models by default!");
```
jackgerrits, line 1154: Small thing, I don't think the ! is necessary.
```suggestion
    logger::errlog_warn("--save_resume flag is deprecated -- learning can now continue on saved models by default.");
```

jackgerrits (approved):

PR #3504, opened 2021-12-03
feat: Warn if options are passed which cannot be reached due to unsupplied necessary args

### LINKED ISSUE #3351: Setup should fail if an option is passed which is not used by the activated reductions
There are many command line options and a major point of confusion is knowing if an option is being used. Is it hard to know which reduction an option belongs to without looking at code.

We have enough information in the setup process to determine if an option is passed by the user but the `necessary` option of the option group in which it resides was not activated. If this is the case we should exit with an error message stating the option was passed but the required reduction which consumes it was not enabled.

jackgerrits:
If a user passes an option that will not be used then a warning will be produced. For example:
```
[warning] Option 'epsilon' depends on another option (or combination of options) which was not supplied. Possible combinations of options which would enable this option are:
	cb_explore_pdf
	warm_cb
	cb_explore_adf, bag
	cb_explore_adf, cover
	cb_explore_adf, first
	cb_explore_adf, synthcover
	cb_explore_adf, rnd
	cb_explore_adf, softmax
	cb_explore_adf
	cb_explore
```

This change will suggest a user to use `--cb_explore_adf --regcb --regcbopt` if  just `--cb_explore_adf --regcbopt` was passed

Resolves #3351


vowpalwabbit/cb_explore_adf_regcb.cc
```diff
@@ -279,7 +280,7 @@ base_learner* setup(VW::setup_base_i& stack_builder)
                .one_of({"mtr"})
                .help("Contextual bandit method to use. RegCB only supports supervised regression (mtr)"));
 
-  options.add_and_parse(new_options);
+  options.add_parse_and_check_necessary(new_options);
```
lalo, line 283: this might be incorrect bc of line 285

vowpalwabbit/cb_explore_adf_greedy.cc
```diff
@@ -90,11 +90,12 @@ VW::LEARNER::base_learner* setup(VW::setup_base_i& stack_builder)
   new_options
       .add(make_option("cb_explore_adf", cb_explore_adf_option)
                .keep()
+               .necessary()
```
lalo, line 93: this works because of line 102 right?

vowpalwabbit/cb_explore_adf_regcb.cc
jackgerrits: Hmm good point. So it is `cb_explore_adf` and `regcb` or `regcbopt` right? How should we handle that?

vowpalwabbit/cb_explore_adf_greedy.cc
jackgerrits: I'll need to look closer here.

lalo: i think it should work but yea i would double check

vowpalwabbit/ftrl.cc
```diff
@@ -401,33 +401,57 @@ base_learner* ftrl_setup(VW::setup_base_i& stack_builder)
...
+  else if (pistol_enabled)
   {
     b->ftrl_alpha = options.was_supplied("ftrl_alpha") ? b->ftrl_alpha : 1.0f;
     b->ftrl_beta = options.was_supplied("ftrl_beta") ? b->ftrl_beta : 0.5f;
   }
-  else if (coin)
+  else if (coin_enabled)
   {
     b->ftrl_alpha = options.was_supplied("ftrl_alpha") ? b->ftrl_alpha : 4.0f;
     b->ftrl_beta = options.was_supplied("ftrl_beta") ? b->ftrl_beta : 1.0f;
```
bassmang, line 457: nit: Can we combine the `if...else/ifs` in lines 444-458 with those on lines 469-500?

vowpalwabbit/cbify.cc
```diff
@@ -745,7 +745,7 @@ base_learner* cbify_setup(VW::setup_base_i& stack_builder)
     }
   }
 
-  if (data->use_adf)
+  if (data->use_adf && (options.was_supplied("regcb") || options.was_supplied("squarecb")))
```
JohnLangford, line 748: Why gateway on regcb or squarecb?  It seems like the ability to encode losses should be independent of the chosen algorithm.

vowpalwabbit/options_boost_po.cc
```diff
@@ -144,8 +158,24 @@ void options_boost_po::add_and_parse(const option_group_definition& group)
 
 bool options_boost_po::add_parse_and_check_necessary(const option_group_definition& group)
 {
-  this->add_and_parse(group);
-  return group.check_necessary_enabled(*this) && group.check_one_of();
+  this->internal_add_and_parse(group);
+
```
JohnLangford, line 162: I'm not quite following this bit of logic.

vowpalwabbit/cbify.cc
jackgerrits: The reason for this gateway is that supplying `cb_min_cost` when no reduction actually consumes the option will produce an error. By checking that either one of those is in use we avoid showing a warning to a user that they did not cause

vowpalwabbit/options_boost_po.cc
jackgerrits: There are two ways in which an option group is added:
- `add_and_parse`, or
- `add_parse_and_check_necessary`

The semantic difference here is options added by `add_and_parse` are assumed to be always reachable. Whereas options added with `add_parse_and_check_necessary` are only reachable if the corresponding necessary options of the given group have been supplied.

For example, for CCB `--all_slots_loss` only used if CCB is enabled with `--ccb_explore_adf`.

bassmang (approved):

PR #3506, opened 2021-12-03
fix: shared_feature_merge should enable if base is of expected type

jackgerrits:
Previously shared feature merger was enabled if the expected options were passed. However, a better check if if the received `base` learner is one of the expected reductions.


lalo (approved):

vowpalwabbit/shared_feature_merger.cc
```diff
@@ -13,19 +13,30 @@
...
 {
-static const std::vector<std::string> option_strings = {
+static const std::vector<std::string> reduction_names_to_enable_for = {
     "csoaa_ldf", "wap_ldf", "cb_adf", "explore_eval", "cbify_ldf", "cb_explore_adf", "warm_cb"};
 
-bool use_reduction(config::options_i& options)
+bool use_reduction(const VW::LEARNER::base_learner& base_learner)
 {
-  for (const auto& opt : option_strings)
+  // TODO - workout a better check to determine if we should merge a shared header.
```
olgavrou, line 28: why is this `reduction_names_to_enable_for` not a good enough way to determine if we should merge a shared header?

jackgerrits: It's not that bad. It would just be nice if we were enabling for label types which have shared header, not reductions which we know happen to use those label types and therefore may have shared headers.

PR #3508, opened 2021-12-06
fix: cb/ccb loss fix

### LINKED ISSUE #3056: Fix reported loss for data with pdrop
#### Describe the bug
#2076 was created by mistake and expected behavior should be exactly opposite.
pdrop should be taken into account for both numerator and denumerator for reported loss, so average loss for training on:
- some cb example
- the same cb example with different pdrop
should be the same

ataymano:
Incorporating weight into average loss numerator

Fixes #3056

### DIFF
--- a/vowpalwabbit/conditional_contextual_bandit.cc
+++ b/vowpalwabbit/conditional_contextual_bandit.cc
@@ -606,7 +606,7 @@ void output_example(VW::workspace& all, ccb& c, multi_ex& ec_seq)
       {
         const float l = CB_ALGS::get_cost_estimate(
             outcome->probabilities[TOP_ACTION_INDEX], outcome->cost, preds[i][TOP_ACTION_INDEX].action);
-        loss += l * preds[i][TOP_ACTION_INDEX].score;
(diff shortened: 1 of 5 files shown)


ataymano:
Fixes #3056

jackgerrits (approved):

olgavrou (approved):

vowpalwabbit/conditional_contextual_bandit.cc
```diff
@@ -606,7 +606,7 @@ void output_example(VW::workspace& all, ccb& c, multi_ex& ec_seq)
       {
         const float l = CB_ALGS::get_cost_estimate(
             outcome->probabilities[TOP_ACTION_INDEX], outcome->cost, preds[i][TOP_ACTION_INDEX].action);
-        loss += l * preds[i][TOP_ACTION_INDEX].score;
+        loss += l * preds[i][TOP_ACTION_INDEX].score * ec_seq[SHARED_EX_INDEX]->weight;
```
ataymano, line 609: using weight from shared example in order to be consistent with line 623

PR #3514, opened 2021-12-07
feat: warn if --quantile_tau is not used

jackgerrits:


vowpalwabbit/parse_args.cc
```diff
@@ -1048,6 +1048,12 @@ void parse_example_tweaks(options_i& options, VW::workspace& all)
   }
 
   all.loss = getLossFunction(all, loss_function, loss_parameter);
+  if (options.was_supplied("quantile_tau") && all.loss->getType() != "quantile")
```
lalo, line 1051: why would this happen though?

lalo: and why only test for quantile_tau, I'm assuming this can happen with any other loss related arg that is later overridden somehow

jackgerrits: quantile is the only loss function which accepts an argument.

lalo (approved):

PR #3520, opened 2021-12-07
 - initial commit

singlis:


jackgerrits:
Why are the fmt and spdlog submodules changed?

python/demos/Non_Stationary_Exploration.ipynb
```diff
@@ -0,0 +1,273 @@
+{
```
jackgerrits: Does it make sense to move this notebook to `python/docs/source/examples` so that it can be rendered on the website?

python/demos/Non_Stationary_Exploration.ipynb
```diff
@@ -0,0 +1,273 @@
...
...4f2r+g3OhwHfq/QzWY7zocLPgGmV2jYBrrLWHgm0awn8zRjTwVq7FbgAKLbW3lHpnLdO8P4iIiJ1Rt3jRURE6kdRecIesB7wAR2/xbWWlienla4F8O/yHdbavTgV/W9z/RP5HrAQKDHGeIwxHqAMWAb0qyHWygk7xnG7MeYLY8zhSucCpH+LeC4CFlT+mVhrt+D0BhhUpe3S8oQ9oHx8fPnP6VOglTHmeWPMd40xzb9FPCIiIrVGSbuIiEj9KK68Ya31A16cSvSp2ltlu7Sm9wjsr4uZ39sCt+Ik25W/rgVSqrTdWcP5t+N0S38bGI5T3R4eOPZt4m0N7Khh/w6qd2uv6WcUfF9r7WLgeuCcQHy7jTGvGWOq3peIiEi9UPd4ERGRhuEoTtftyhrqOOrdOFX9x2o4VlJlu6a1Za8B3rLW3lm+wxhzyWnEswdoV8P+doFjp8Ra+xLwUqDr/JU4wwueAy759iGKiIh8O0raRUREGoYC4CpjjLvSkmiXhzMgnCp0lDHGFegZUO4dnMnZVllrvd/iuk2pXhm/8RjvfzKV9/eBHxljflFpTHtHYADw/75FfEBwiMGLxpjzgdHf9joiIiKnQ0m7iIhIw/AKMB5nUrRXgCxgbHhD4kucoXR3GGP+C+y31n4N3At8ArxrjJkJbMOZHf87wBZr7aMnuO6/gTuNMb8APgd+SM1V7C+By4wx38OpmOdaa3fX0O5+nFnm/22MeZCK2eP3UHNvgGMyxkwP3MtinA8WzsFJ2P9zKtcRERGpLRrTLiIi0gBYa/+DM6P6d4F/4iyzFu7q7uvAkzjLtX0EzAKw1ubhTDhXADyMU3l/AGeJuI9P4rq/x1lWbSrwKpCIM468ql8DeYE2n+Ak99VYaz8Dyte8fxH4P+Ar4MIqy72djI9xEvXHcBL1e3Fm9590itcRERGpFcbamoaaiYiIiIiIiEi4qdIuIiIiIiIi0kBpTLuIiEgjZ4wxgPs4TWylye9ERESkAVGlXUREpPG7mOprqlf+2hS+0EREROR4VGkXERFp/D4Fzj/O8aprq4uIiEgDoYnoRERERERERBoodY8XERERERERaaCUtIuIiIiIiIg0UEraRURERERERBooJe0iIiIiIiIiDZSSdhEREREREZEGSkm7iIiIiIiISAP1/wGftwmo1YU0fwAAAABJRU5ErkJggg==\n",
```
jackgerrits: I think its best not to commit the graphics, they will be auto rendered when built for the docs site

test/unit_test/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 add_executable(vw-unit-test.out
+  epsilon_decay_test.cc
```
jackgerrits, line 2: please add in correct place alphabetically

test/unit_test/epsilon_decay_test.cc
```diff
@@ -0,0 +1,100 @@
+
+#include <boost/test/tools/old/interface.hpp>
```
jackgerrits, line 2: nit: remove extra newline

test/unit_test/simulator.h
```diff
@@ -44,22 +44,23 @@ class cb_sim
   std::string action_ns;
 
   cb_sim(uint64_t = 0);
-  float get_reaction(const std::map<std::string, std::string>&, const std::string&, bool = false);
+  float get_reaction(const std::map<std::string, std::string>&, const std::string&, bool = false, bool = false);
```
jackgerrits, line 47: why are there args with no name?

vowpalwabbit/Makefile
```diff
@@ -1,8 +0,0 @@
-default:
```
jackgerrits, line 1: This shouldn't be removed

vowpalwabbit/reductions/epsilon_decay.cc
```diff
@@ -0,0 +1,240 @@
...
+  uint64_t ppw = 1;
+  while (ppw < model_count) { ppw *= 2; }
+  for (; first != end; ++first)
+  {
+    first->reset_stats(epsilon_decay_alpha, epsilon_decay_tau);
+    weights.dense_weights.clear_offset(first->get_model_idx(), ppw);
+  }
+}
+
+float decayed_epsilon(uint64_t update_count) { return static_cast<float>(pow(update_count + 1, -1.f / 3.f)); }
```
jackgerrits, line 56:
```suggestion
float decayed_epsilon(uint64_t update_count) { return static_cast<float>(std::pow(update_count + 1, -1.f / 3.f)); }
```

vowpalwabbit/reductions/epsilon_decay.cc
```diff
@@ -0,0 +1,240 @@
...
+{
+  for (; first != end; ++first, ++n_first) { std::iter_swap(first, n_first); }
+  return n_first;
+}
+
+template <class ForwardIt>
+void reset_models(ForwardIt first, ForwardIt end, parameters& weights, double epsilon_decay_alpha,
+    double epsilon_decay_tau, uint64_t model_count)
+{
+  uint64_t ppw = 1;
```
jackgerrits, line 47: What is ppw? Can you update name or add a comment

vowpalwabbit/reductions/epsilon_decay.h
```diff
@@ -0,0 +1,64 @@
...
+
+namespace VW
+{
+namespace epsilon_decay
+{
+VW::LEARNER::base_learner* epsilon_decay_setup(VW::setup_base_i&);
+
+struct epsilon_decay_score : scored_config
+{
+  epsilon_decay_score() : VW::scored_config() {}
```
jackgerrits, line 19: I dont think this is necessary, can be replaced with default

```suggestion
  epsilon_decay_score() = default;
```

vowpalwabbit/reductions/epsilon_decay.h
```diff
@@ -0,0 +1,64 @@
...
+    {
+      epsilon_decay_score s(epsilon_decay_alpha, epsilon_decay_tau, i);
+      scored_configs.push_back(s);
+    }
+  }
+  std::vector<epsilon_decay_score> scored_configs;
+  uint64_t min_scope;
+  double epsilon_decay_alpha;
+  double epsilon_decay_tau;
+  parameters& weights;
```
jackgerrits, line 52: Should these be prefixed with _ or m_?

vowpalwabbit/reductions/epsilon_decay.cc
```diff
@@ -0,0 +1,240 @@
...
+
+  // make sure we setup the rest of the stack with cleared interactions
+  // to make sure there are not subtle bugs
+  auto* base_learner = stack_builder.setup_base_learner();
+  if (base_learner->is_multiline())
+  {
+    auto* learner = VW::LEARNER::make_reduction_learner(std::move(data), VW::LEARNER::as_multiline(base_learner), learn,
+        predict, stack_builder.get_setupfn_name(epsilon_decay_setup))
+                        .set_params_per_weight(ppw)
+                        .set_output_prediction_type(base_learner->get_output_prediction_type())
```
jackgerrits, line 188: Please set all of:
- `set_output_prediction_type`
- `set_input_prediction_type`
- `set_output_label_type`
- `set_input_label_type`

vowpalwabbit/reductions/epsilon_decay.cc
```diff
@@ -0,0 +1,240 @@
...
+                        .set_save_load(save_load_epsilon_decay)
+                        .build();
+
+    return VW::LEARNER::make_base(*learner);
+  }
+  else
+  {
+    // not implemented yet
+    THROW("Epsilon decay does not supported for single line learners yet");
```
jackgerrits, line 197: I would avoid saying "yet"

```suggestion
    THROW("Epsilon decay  (--epsilon_decay) does not supported for single line learners");
```

bassmang:
> Why are the fmt and spdlog submodules changed?

Not sure, reverted it.

python/demos/Non_Stationary_Exploration.ipynb
bassmang: This notebook is pretty experimental, but I can clean it up then move it

python/demos/Non_Stationary_Exploration.ipynb
bassmang: Removed

test/unit_test/simulator.h
bassmang: Updated the doc with arguments

vowpalwabbit/Makefile
bassmang: restored

vowpalwabbit/reductions/epsilon_decay.cc
bassmang: params_per_weight -- changed

vowpalwabbit/reductions/epsilon_decay.h
bassmang: I don't see a strong reason to, I think accessing with `data._weights` etc.. doesnt look great

vowpalwabbit/reductions/epsilon_decay.cc
bassmang: added

test/unit_test/simulator.h
```diff
@@ -43,23 +43,26 @@ class cb_sim
...
+  void call_if_exists(VW::workspace& vw, multi_ex& ex, const callback_map& callbacks, const size_t event);
 };
 
-std::vector<float> _test_helper(const std::string&, size_t = 3000, int = 10);
-std::vector<float> _test_helper_save_load(const std::string&, size_t = 3000, int = 10);
-std::vector<float> _test_helper_hook(const std::string&, callback_map&, size_t = 3000, int = 10);
+std::vector<float> _test_helper(const std::string& vw_arg, size_t num_iterations = 3000, int seed = 10);
+std::vector<float> _test_helper_save_load(const std::string& vw_arg, size_t num_iterations = 3000, int seed = 10);
+std::vector<float> _test_helper_hook(const std::string& vw_arg, callback_map& hooks, size_t num_iterations = 3000,
+    int seed = 10, const std::vector<uint64_t>& swap_after = std::vector<uint64_t>());
```
lalo, line 67: swap_after could have a more interesting implementation if applied via a normal test hook, since every test hook already receives an instance of the simulator to apply any state change. But since it's testing, this can later be improved and for now just a nit.

vowpalwabbit/reductions/epsilon_decay.h
```diff
@@ -0,0 +1,64 @@
...
+  epsilon_decay_score(double alpha, double tau, uint64_t model_idx)
+      : VW::scored_config(alpha, tau), _model_idx(model_idx)
+  {
+  }
+  float get_upper_bound() const { return this->current_ips(); }
+  float get_lower_bound() const { return _lower_bound; }
+  uint64_t get_model_idx() const { return _model_idx; }
+  void update_bounds(float w, float r);
+
+  float _lower_bound = 0.f;
```
rajan-chari, line 29: private variables?

vowpalwabbit/reductions/epsilon_decay.h
```diff
@@ -0,0 +1,64 @@
...
+      , epsilon_decay_tau(epsilon_decay_tau)
+      , weights(weights)
+  {
+    for (uint64_t i = 0; i < num_configs; ++i)
+    {
+      epsilon_decay_score s(epsilon_decay_alpha, epsilon_decay_tau, i);
+      scored_configs.push_back(s);
+    }
+  }
+  std::vector<epsilon_decay_score> scored_configs;
```
jackgerrits, line 48: member fields should be prefixed with a `_`

vowpalwabbit/reductions/epsilon_decay.h
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) by respective owners including Yahoo!, Microsoft, and
+// individual contributors. All rights reserved. Released under a BSD (revised)
+// license as described in the file LICENSE.
+#pragma once
+
+#include "array_parameters.h"
+#include "distributionally_robust.h"
+#include "reductions_fwd.h"
```
jackgerrits, line 8:
```suggestion
#include "vw_fwd.h"
```

vowpalwabbit/reductions/epsilon_decay.h
```diff
@@ -0,0 +1,64 @@
...
+#include "scored_config.h"
+
+namespace VW
+{
+namespace epsilon_decay
+{
+VW::LEARNER::base_learner* epsilon_decay_setup(VW::setup_base_i&);
```
jackgerrits, line 15: Please move the setup function into `VW::reductions` namespace

vowpalwabbit/reductions/epsilon_decay.cc
```diff
@@ -0,0 +1,247 @@
...
+                        .set_output_prediction_type(base_learner->get_output_prediction_type())
+                        .set_save_load(save_load_epsilon_decay)
+                        .build();
+
+    return VW::LEARNER::make_base(*learner);
+  }
+  else
+  {
+    // not implemented
+    THROW("--epsilon_decay does not supported for single line learners");
```
jackgerrits, line 204:
```suggestion
    THROW("--epsilon_decay is not supported for single line learners");
```

vowpalwabbit/reductions/epsilon_decay.cc
```diff
@@ -0,0 +1,247 @@
...
+{
+  update(w, r);
+
+  // update the lower bound
+  distributionally_robust::ScoredDual sd = this->chisq.recompute_duals();
+  _lower_bound = static_cast<float>(sd.first);
+}
+
+template <class ForwardIt>
+ForwardIt swap_models(ForwardIt first, ForwardIt n_first, ForwardIt end)
```
jackgerrits, line 39: Put things what are not exported from the .h in an anonymous namespace

vowpalwabbit/reductions/epsilon_decay.h
```diff
@@ -0,0 +1,64 @@
...
+  double epsilon_decay_alpha;
+  double epsilon_decay_tau;
+  parameters& weights;
+};
+
+}  // namespace epsilon_decay
+
+namespace model_utils
+{
+size_t read_model_field(io_buf&, VW::epsilon_decay::epsilon_decay_score&);
```
jackgerrits, line 59: What are these needed for? If they are only required for the save_load of this reduction then they should be able to be only in the cc file and we should avoid exporting them here

vowpalwabbit/reductions/epsilon_decay.h
```diff
@@ -0,0 +1,64 @@
...
+#include "reductions_fwd.h"
+#include "scored_config.h"
+
+namespace VW
+{
+namespace epsilon_decay
+{
+VW::LEARNER::base_learner* epsilon_decay_setup(VW::setup_base_i&);
+
+struct epsilon_decay_score : scored_config
```
jackgerrits, line 17: Please move other things like this into `VW::reductions::epsilon_decay` namespace

jackgerrits (changes requested):

vowpalwabbit/reductions/epsilon_decay.h
jackgerrits: The convention for members is to use a prefix so that they can be distinguished

python/demos/Non_Stationary_Exploration.ipynb
jackgerrits: I think as long as the notebook specifies that it's fine

vowpalwabbit/reductions/epsilon_decay.h
bassmang: these are accessed directly so cannot be private

vowpalwabbit/reductions/epsilon_decay.h
```diff
@@ -0,0 +1,64 @@
...
+  {
+    for (uint64_t i = 0; i < num_configs; ++i)
+    {
+      epsilon_decay_score s(epsilon_decay_alpha, epsilon_decay_tau, i);
+      scored_configs.push_back(s);
+    }
+  }
+  std::vector<epsilon_decay_score> scored_configs;
+  uint64_t min_scope;
+  double epsilon_decay_alpha;
```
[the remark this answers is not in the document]
bassmang: added

vowpalwabbit/reductions/epsilon_decay.cc
bassmang: I added these functions to these functions to the .h instead

vowpalwabbit/reductions/epsilon_decay.h
bassmang: fair enough -- changed

vowpalwabbit/reductions/epsilon_decay.h
```diff
@@ -0,0 +1,64 @@
...
+  double epsilon_decay_tau;
+  parameters& weights;
+};
```
rajan-chari, line 53: Wondering what happens during vw shutdown.  Is it possible to try to access this during shutdown when it's cleaned up?  (Probably not, but wondering)

vowpalwabbit/reductions/epsilon_decay.cc
```diff
@@ -0,0 +1,247 @@
...
+  uint64_t params_per_weight = 1;
+  while (params_per_weight < model_count) { params_per_weight *= 2; }
+  for (; first != end; ++first)
+  {
+    first->reset_stats(epsilon_decay_alpha, epsilon_decay_tau);
+    weights.dense_weights.clear_offset(first->get_model_idx(), params_per_weight);
+  }
+}
+
+float decayed_epsilon(uint64_t update_count) { return static_cast<float>(std::pow(update_count + 1, -1.f / 3.f)); }
```
rajan-chari, line 58: Is there a paper with reference to why 1/cube_root(n) makes sense?  A reference to that here would be useful.

vowpalwabbit/reductions/epsilon_decay.h
bassmang: This actually doesn't work. If I move these to only the cc file, the save_load for scored_config will be used instead of epsilon_decay_score (which inherits the former), since a vector of these are being saved, and it would not be visible to model_utils.h

jackgerrits: Ooof that's not good... I wonder if there is a way to ensure the parent is not considered for overload resolution and only the concrete type. I can't think of a way though

vowpalwabbit/reductions/epsilon_decay.h
bassmang: Not sure this is possible. Is there a reason you want to check this?

vowpalwabbit/reductions/epsilon_decay.cc
bassmang: I'll ask John in the next sync

vowpalwabbit/reductions/epsilon_decay.cc
```diff
@@ -0,0 +1,247 @@
...
+            (std::pow(champion_iter->update_count, static_cast<float>(model_idx) / model_count)))
+    {
+      auto n_iter = swap_models(candidate_iter + 1, candidate_iter, end_iter);
+      reset_models(n_iter, end_iter, data.weights, data.epsilon_decay_alpha, data.epsilon_decay_tau, model_count);
+      break;
+    }
+  }
+}
+
+void save_load_epsilon_decay(epsilon_decay_data& epsilon_decay, io_buf& io, bool read, bool text)
```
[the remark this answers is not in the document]
bassmang: I dont think this is necessary. If all.save_resume is false (--predict_only_model is turned on with some existing model) then the weights should not be updated, and a new model should not be saved, since the current model being used is only for prediction.

vowpalwabbit/reductions/epsilon_decay.h
```diff
@@ -0,0 +1,73 @@
...
+      , _weights(weights)
+  {
+    for (uint64_t i = 0; i < num_configs; ++i)
+    {
+      epsilon_decay_score s(epsilon_decay_alpha, epsilon_decay_tau, i);
+      _scored_configs.push_back(s);
+    }
+  }
+  std::vector<epsilon_decay_score> _scored_configs;
+  uint64_t _min_scope;
```
rajan-chari, line 57: 1) Prepend with '_' for class member variables so it's easier to read?
2) Can some of these be private or protected?

vowpalwabbit/reductions/epsilon_decay.cc
```diff
@@ -0,0 +1,256 @@
...
+
+  // Update model count to be 2^n
+  uint64_t params_per_weight = 1;
+  while (params_per_weight < model_count) { params_per_weight *= 2; }
+
+  // Scale confidence interval by number of examples
+  float scaled_alpha = _epsilon_decay_alpha / model_count;
+
+  auto data = VW::make_unique<VW::reductions::epsilon_decay::epsilon_decay_data>(
+      model_count, _min_scope, scaled_alpha, _epsilon_decay_tau, all.weights);
```
rajan-chari, line 231: Checks for sane model_count?  perhaps > 0?  @peter pointed out that iterators later in the code make this assumption.

vowpalwabbit/reductions/epsilon_decay.h
rajan-chari: Line number of the comment seems off.  I was wondering about parameters& weights member variable.

vowpalwabbit/reductions/epsilon_decay.cc
```diff
@@ -0,0 +1,256 @@
...
+    {
+      auto n_iter = swap_models(candidate_iter + 1, candidate_iter, end_iter);
+      reset_models(n_iter, end_iter, data._weights, data._epsilon_decay_alpha, data._epsilon_decay_tau, model_count);
+      break;
+    }
+  }
+}
+
+void save_load_epsilon_decay(
+    VW::reductions::epsilon_decay::epsilon_decay_data& epsilon_decay, io_buf& io, bool read, bool text)
```
rajan-chari, line 179: My point here was that for prediction you only need the champion model.  So it's wasteful to save the weights for all the models all the time.  I don't know of a good way to only save the weights for one model though.  This is controlled by make_learner_reduction params_per_weight.  As far as I know it's not possible to change this once the reduction is created.

rajan-chari (approved):
I had a few additional minor comments.  Please take a look.  Thanks.

vowpalwabbit/reductions/epsilon_decay.cc
bassmang: added check

vowpalwabbit/reductions/epsilon_decay.cc
bassmang: Yes I believe this would require us to spin off a new model with only one set of weights. Eduardo had previously brought this up as an extension to automl. I think this would require a separate PR.

PR #3522, opened 2021-12-08
feat: mean field implementation for logistic regression

akshaykr:
Allows logistic regression to take non-binary costs via mean field implementation. E.g., for loss p in [0,1] update is p*update(1) + (1-p)*update(-1)


vowpalwabbit/loss_functions.cc
```diff
@@ -172,14 +172,19 @@ class logloss : public loss_function
 
   float getLoss(shared_data*, float prediction, float label) override
   {
+    if (label >= 0.f && label < 1.f)
+      return label * getLoss(nullptr, prediction, 1.f) + (1 - label) * getLoss(nullptr, prediction, -1.f);
     // TODO: warning or error?
     if (label != -1.f && label != 1.f)
-      logger::log_warn("You are using label {} not -1 or 1 as loss function expects!", label);
+      logger::log_warn("You are using label {} not -1 or 1 or in [0,1] as loss function expects!", label);
```
olgavrou, line 179: nit: I know it was not introduced by this change, but can we make the warning less aggressive by removing the `!`, and also removing the `You`?
```suggestion
      logger::log_warn("The label {} is not -1 or 1 or in [0,1] as loss function expects", label);
```

olgavrou (approved):

olgavrou (approved):

PR #3527, opened 2021-12-09
refactor: Use model_utils for label cache

bassmang:


vowpalwabbit/cb.cc
```diff
@@ -235,3 +217,57 @@ label_parser cb_eval = {
     VW::label_type_t::cb_eval};
 
 }  // namespace CB_EVAL
+
```
olgavrou, line 220: nit: add empty line in between functions

vowpalwabbit/no_label.cc
```diff
@@ -42,7 +42,8 @@ label_parser no_label_parser = {
     [](polylabel& /* label */, reduction_features& /* red_features */, VW::label_parser_reuse_mem& /* reuse_mem */,
         const VW::named_labels* /* ldict */, const std::vector<VW::string_view>& words) { parse_no_label(words); },
     // cache_label
-    [](const polylabel& /* label */, const reduction_features& /* red_features */, io_buf& /* cache */) {},
+    [](const polylabel& /* label */, const reduction_features& /* red_features */, io_buf& /* cache */,
+        const std::string&, bool) -> size_t { return 1; },
```
olgavrou, line 46: return 1 or 0?

bassmang: I used 1 for consistency since read_cached_label returns 1, they should probably both return 0 though

olgavrou (approved):
LGTM, we checked for backwards compatibility right?

Approving with suggestion to make `return 1` -> `return 0` when we dont' read anything

Also we should think if making `read_model_field` return `uint32_t` instead of a `size_t` is needed (different PR if so)

PR #3529, opened 2021-12-09
feat: More consistent and controllable output

jackgerrits:
- Cleans up warning and error messages to be less aggressive
- Send all warnings and errors to structured logging instead of driver
- Uses a file based logging API
- Redirects logs to C API if trace message function given (logs messages didn't make it there in the past)
- Allow log level to be set
- Allow driver and logs to be turned off independently
- Allow logs and driver logs to be sent to specifically stdout or stderr
- Add compat mode to preserve current logging locations by default


vowpalwabbit/parse_args.cc
```diff
@@ -280,14 +278,14 @@ void parse_diagnostics(options_i& options, VW::workspace& all)
...
+  size_t upper_limit = 0;
   option_group_definition diagnostic_group("Diagnostic");
   diagnostic_group.add(make_option("version", version_arg).help("Version information"))
       .add(make_option("audit", all.audit).short_name("a").help("Print weights of features"))
       .add(make_option("progress", progress_arg)
                .short_name("P")
                .help("Progress update frequency. int: additive, float: multiplicative"))
-      .add(make_option("quiet", all.logger.quiet).help("Don't output diagnostics and progress updates"))
-      .add(make_option("limit_output", all.logger.upper_limit).help("Avoid chatty output. Limit total printed lines"))
+      .add(make_option("limit_output", upper_limit).help("Avoid chatty output. Limit total printed lines"))
```
lalo, line 288: not sure if this one belongs here or with the others below (where you add the new one)

jackgerrits: Good catch, it definitely should go there

vowpalwabbit/binary.cc
```diff
@@ -15,14 +16,18 @@
...
-namespace logger = VW::io::logger;
-
 namespace VW
 {
 namespace binary
 {
+struct binary_data
+{
+  VW::io::logger logger;
+  explicit binary_data(VW::io::logger logger) : logger(std::move(logger)) {}
```
olgavrou, line 26:
```suggestion
  explicit binary_data(VW::io::logger&& logger) : logger(std::move(logger)) {}
```
?

olgavrou: is this a copy of the `all.logger`?

vowpalwabbit/cb_explore.cc
```diff
@@ -54,6 +54,7 @@ struct cb_explore
   bool nounif = false;
   bool epsilon_decay = false;
   VW::version_struct model_file_version;
+  VW::io::logger* logger;
```
olgavrou, line 57: other reductions get a copy to the `all.logger` object and others hold a pointer to it?

vowpalwabbit/cb_adf.cc
```diff
@@ -85,19 +83,22 @@ struct cb_adf
   const bool _rank_all;
   const float _clip_p;
 
+  VW::io::logger _logger;
```
olgavrou, line 86: nit: inconsistent naming in reductions? `_logger` vs `logger`?

vowpalwabbit/cs_active.cc
```diff
@@ -336,18 +335,17 @@ base_learner* cs_active_setup(VW::setup_base_i& stack_builder)
   data->t = 1;
 
   auto loss_function_type = all.loss->getType();
-  if (loss_function_type != "squared") THROW("error: you can't use non-squared loss with cs_active");
+  if (loss_function_type != "squared") THROW("You can't use non-squared loss with cs_active");
```
olgavrou, line 338: In the non-aggressive effort should we avoid `You` altogether? i.e. `Non-squared loss can not be used with cs_active`?

vowpalwabbit/gd.cc
```diff
@@ -543,7 +542,8 @@ inline void pred_per_update_feature(norm_data& nd, float x, float& fw)
       if (x2 > x2_max)
       {
         norm_x2 = 1;
-        logger::errlog_error("your features have too much magnitude");
+        assert(nd.logger != nullptr);
+        nd.logger->err_error("your features have too much magnitude");
```
olgavrou, line 546: same here: `the features have too much magnitude` ?

vowpalwabbit/gd_mf.cc
```diff
@@ -264,7 +264,7 @@ void save_load(gdmf& d, io_buf& model_file, bool read, bool text)
 
   if (model_file.num_files() > 0)
   {
-    if (!all.weights.not_null()) { THROW("Error: Model weights not initialized."); }
+    if (!all.weights.not_null()) { THROW("Model weights not initialized when trying to load into them."); }
```
olgavrou, line 267: I'm not sure I understand the `when trying to load into them`, is it `when trying to load them`?

vowpalwabbit/gen_cs_example.cc
```diff
@@ -11,28 +11,26 @@
...
 using namespace CB;
 
-
-float safe_probability(float prob)
+float safe_probability(float prob, VW::io::logger& logger)
 {
   if (prob <= 0.)
   {
-    logger::log_warn("Probability {} is not possible, replacing with 1e-3.  Fix your dataset. ", prob);
+    logger.out_warn("Probability {} is not possible, replacing with 1e-3.  Fix your dataset. ", prob);
```
olgavrou, line 24: oooof the `Fix your dataset` :D Should we replace with: `There seems to be something wrong with the dataset`?

vowpalwabbit/global_data.cc
```diff
@@ -212,24 +210,25 @@ void workspace::finish_example(multi_ex& ec)
...
     if (isdigit(limit[0]))
     {
       int n = atoi(limit.c_str());
-      logger::errlog_warn("limiting to {} features for each namespace.", n);
+      logger.err_warn("limiting to {} features for each namespace.", n);
       for (size_t j = 0; j < 256; j++) dest[j] = n;
     }
     else if (limit.size() == 1)
-      logger::log_error("You must specify the namespace index before the n");
+      logger.out_error("You must specify the namespace index before the n");
```
olgavrou, line 226:
```suggestion
      logger.out_error("The namespace index must be specified before the n");
```

vowpalwabbit/lda_core.cc
```diff
@@ -1327,7 +1330,7 @@ base_learner* lda_setup(VW::setup_base_i& stack_builder)
 
   if (all.eta > 1.)
   {
-    logger::errlog_warn("your learning rate is too high, setting it to 1");
+    all.logger.err_warn("your learning rate is too high, setting it to 1");
```
olgavrou, line 1333:
```suggestion
    all.logger.err_warn("The learning rate is too high, setting it to 1");
```

vowpalwabbit/lda_core.cc
```diff
@@ -534,7 +532,8 @@ float lda::digamma(float x)
       return ldamath::digamma<float, lda_math_mode::USE_SIMD>(x);
     default:
       // Should not happen.
-      logger::errlog_critical("lda::digamma: Trampled or invalid math mode, aborting");
+      std::cerr << "lda::digamma: Trampled or invalid math mode, aborting" << std::endl;
```
olgavrou, line 535: there are a couple of extra `;` in this file. Why where these log lines replaced with `std::err`?

vowpalwabbit/parser.cc
```diff
@@ -279,7 +286,7 @@ void make_write_cache(VW::workspace& all, std::string& newname, bool quiet)
   io_buf& output = all.example_parser->output;
   if (output.num_files() != 0)
   {
-    *(all.trace_message) << "Warning: you tried to make two write caches.  Only the first one will be made." << endl;
+    all.logger.err_warn("You tried to make two write caches. Only the first one will be made.");
```
olgavrou, line 289:
```suggestion
    all.logger.err_warn("There was an attempt to make two write caches. Only the first one will be made.");
```

vowpalwabbit/binary.cc
jackgerrits: Yeah copying that object is intended to be the standard usage pattern since it is just a shared pointer

vowpalwabbit/cb_explore.cc
jackgerrits: I need to check this specific place but the logger object cannot be initialized, so either the constructor initializes it or it must be some optional value. Clearly optional would be better

vowpalwabbit/cb_adf.cc
jackgerrits: I opted to be consistent within this reduction, but it does mean it is globally inconsistent now

vowpalwabbit/cs_active.cc
jackgerrits: Definitely, thanks for catching this

vowpalwabbit/gd_mf.cc
jackgerrits: This essentially means the weights object was not initialized/created/allocated before we try to write to it

vowpalwabbit/gen_cs_example.cc
jackgerrits: Hahah, yeah that's much nicer :)

vowpalwabbit/lda_core.cc
jackgerrits: This is a deep math function and and hitting this `default` indicates a deep issue (we abort here...). Rather than funneling a logger here and potentially having perf impact I figured the message + abort is the closer analogue. Messy situation in general, but I  agree with the abort here

vowpalwabbit/boosting.cc
```diff
@@ -391,12 +393,13 @@ VW::LEARNER::base_learner* boosting_setup(VW::setup_base_i& stack_builder)
...
 
-  logger::errlog_info("Number of weak learners = {}", data->N);
-  logger::errlog_info("Gamma = {}", data->gamma);
+  all.logger.err_info("Number of weak learners = {}", data->N);
+  all.logger.err_info("Gamma = {}", data->gamma);
 
   data->C = std::vector<std::vector<int64_t> >(data->N, std::vector<int64_t>(data->N, -1));
   data->t = 0;
   data->all = &all;
+  data->logger = all.logger;
```
olgavrou, line 402: didn't we already copy the logger into `data` at the constructor line 374?

vowpalwabbit/gd.cc
```diff
@@ -1284,9 +1284,11 @@ base_learner* setup(VW::setup_base_i& stack_builder)
   if (g->adax && !all.weights.adaptive) THROW("Cannot use adax without adaptive");
 
   if (pow(static_cast<double>(all.eta_decay_rate), static_cast<double>(all.numpasses)) < 0.0001)
-    *(all.trace_message) << "Warning: the learning rate for the last pass is multiplied by: "
-                         << pow(static_cast<double>(all.eta_decay_rate), static_cast<double>(all.numpasses))
-                         << " adjust --decay_learning_rate larger to avoid this." << std::endl;
+  {
```
olgavrou, line 1287: some trace messages are replaced with logged messages (like here) but some are not (e.g. cs_active) what's the logic behind replacing some and not others? is it that if it's an error or warning it now goes to logger but otherwise stays on driver which can be set independently?

vowpalwabbit/parse_example.cc
```diff
@@ -76,13 +74,14 @@ class TC_parser
   uint32_t _hash_seed;
   uint64_t _parse_mask;
   std::array<std::vector<std::shared_ptr<feature_dict>>, NUM_NAMESPACES>* _namespace_dictionaries;
+  VW::io::logger* logger;
```
olgavrou, line 77: I am guessing are we making this a pointer because we parse many examples and don't want to copy the logger every time?

olgavrou (approved):

vowpalwabbit/gd.cc
jackgerrits: I replaced trace messages that were obviously a warning or error, if I missed any please point them out. I left things that sort of just look like driver sort of updates. I'm not sure whether to move those to an info level log or not. Its still not exactly clear what exactly constitutes the driver output vs info

vowpalwabbit/parse_example.cc
jackgerrits: Yep, that is what I was thinking. And the lifetime of the object is well known so we don't need to worry it being invalidated

vowpalwabbit/boosting.cc
jackgerrits: You're right. Thanks!

vowpalwabbit/parser.cc
```diff
@@ -590,8 +595,7 @@ void enable_sources(VW::workspace& all, bool quiet, size_t passes, input_options
       catch (std::exception const&)
       {
         // when trying to fix this exception, consider that an empty filename_to_read is valid if all.stdin_off is false
-        if (!filename_to_read.empty())
-        { *(all.trace_message) << "can't open '" << filename_to_read << "', sailing on!" << endl; }
+        if (!filename_to_read.empty()) { all.logger.err_error("can't open '{}', sailing on", filename_to_read); }
```
bassmang, line 598: Would it be too radical to change this?

bassmang: Also something to think about is if people are currently looking for `sailing on!` with the exclamation point in their code like here:
https://github.com/VowpalWabbit/vowpal_wabbit/issues/1895

test/train-sets/ref/help.stdout
```diff
@@ -524,274 +228,8 @@ Gradient Descent Options:
...
-                                       lambda=1, 3: center lambda=epsilon/(1+ep
-                                       silon), 4: center lambda=epsilon/(1+epsi
-                                       lon), min lambda=0, max lambda=1); the 
-                                       rest of candidate lambda values are 
-                                       generated using a doubling scheme. 
-                                       Choices: {1, 2, 3, 4}
-  --overwrite_label arg (=1, )         The label used by type 3 corruptions 
-                                       (overwriting)
-  --sim_bandit                         Simulate contextual bandit updates on 
-                                       warm start examples
```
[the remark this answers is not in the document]
jackgerrits: There's a check in the help string generation function which does enabled reduction filtering. The check means that you will filter if there are **any** other options passed. Seems like it could cause confusion but I will work around it for this PR

vowpalwabbit/allreduce_sockets.cc
```diff
@@ -300,7 +292,7 @@ void AllReduceSockets::broadcast(char* buffer, const size_t n)
     if (socks.parent != -1)
     {
       // there is data to be read from the parent
-      if (parent_read_pos == n) THROW("I think parent has no data to send but he thinks he has");
+      if (parent_read_pos == n) THROW("I think parent has no data to send but they think they do.");
```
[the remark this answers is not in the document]
jackgerrits: The comment above sort of contradicts the error message though

vowpalwabbit/parser.cc
jackgerrits: Alright I'll leave the `!` in there

vowpalwabbit/allreduce_sockets.cc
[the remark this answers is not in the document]
bassmang: could we do `There is no data to be read from the parent` then? The current message could be confusing

cs/cli/vw_base.cpp
```diff
@@ -237,9 +237,12 @@ void VowpalWabbitBase::SaveModel(String^ filename)
     throw gcnew ArgumentException("Filename must not be null or empty");
 
   String^ directoryName = System::IO::Path::GetDirectoryName(filename);
-
   if (!String::IsNullOrEmpty(directoryName))
-  { System::IO::Directory::CreateDirectory(directoryName);
```
lalo, line 242: whats wrong with this api? is it different than the one you mention CreateDirectoryA?

jackgerrits: Some header change I think caused it to not be able to find this function any more. I figured since the .net core change is incoming I would not spend too much time on it

vowpalwabbit/parse_args.cc
```diff
@@ -1202,16 +1193,94 @@ ssize_t trace_message_wrapper_adapter(void* context, const char* buffer, size_t
...
+                     "deprecated so it is recommended that stdout or stderr is chosen."))
+      .add(make_option("limit_output", upper_limit)
+               .default_value(0)
+               .help("Avoid chatty output. Limit total printed lines. 0 means unbounded."));
+
+  options->add_and_parse(logging_options);
+
+  if (quiet)
+  {
+    // If a user supplies an explicit log level, then we'll respect that.
```
lalo, line 1233: but specifying --quiet still overrides that log level right?

bassmang (approved):

vowpalwabbit/parse_args.cc
```diff
@@ -1202,16 +1193,94 @@ ssize_t trace_message_wrapper_adapter(void* context, const char* buffer, size_t
...
+               .default_value("stderr")
+               .one_of({"stdout", "stderr"})
+               .help("Specify the stream to output driver output to."))
+      .add(make_option("log_level", log_level)
+               .default_value("info")
+               .one_of({"info", "warn", "error", "critical", "off"})
+               .help("Log level for logging messages. Specifying this wil override --quiet for log output."))
+      .add(make_option("log_output_stream", log_output_stream)
+               .default_value("compat")
+               .one_of({"stdout", "stderr", "compat"})
```
lalo, line 1221: does it ignore caps when checking for one of these?

vowpalwabbit/io/logger.h
```diff
@@ -8,47 +8,224 @@
...
+  out,
+  err,
+  compat
+};
 
-template<typename FormatString, typename... Args>
-  void log_error(const FormatString &fmt, Args&&...args)
+output_location get_output_location(const std::string& name);
+
+namespace details
```
lalo, line 45: interesting refactor

lalo (approved):

vowpalwabbit/parse_args.cc
jackgerrits: As it is right now in the PR. Explicit log_level overrides quiet. Do you think quiet should always silence everything?

vowpalwabbit/parse_args.cc
jackgerrits: No, it is case sensitive. This is how existing places in VW that compare to "stdout" work

vowpalwabbit/parse_args.cc
jackgerrits: Yeah i think quiet should have highest priority

Issue #3445, opened 2021-11-05
Ensure bin_text_read_write_fixed_validated is only called on basic types to avoid struct padding issues


bassmang:
#### Describe the bug
The number of bytes in a struct is compiler-dependent due to padding. To ensure that models are saved and loaded properly ensure that bin_text_read_write_fixed_validated or other functions which read and write models are only called on basic types.

#### To Reproduce
This issue is compiler dependent so catching it could be tricky, but may show up when running tests with valgrind.

PR #3530, opened 2021-12-10
refactor: clarify examples and features cache reading and properly cache flat_example

bassmang:
Resolves: #3445


vowpalwabbit/cache.h
```diff
@@ -19,17 +19,16 @@ struct cache_temp_buffer;
...
+std::pair<size_t, bool> read_cached_index_and_features(io_buf& input, unsigned char& index, features& ours);
 
 namespace VW
 {
 uint32_t convert(size_t number);
 // What is written by write_example_to_cache can be read by read_example_from_cache
 void write_example_to_cache(io_buf& output, example* ae, label_parser& lbl_parser, uint64_t parse_mask,
     VW::details::cache_temp_buffer& temp_buffer);
-int read_example_from_cache(io_buf& input, example* ae, label_parser& lbl_parser, bool sorted_cache);
+int read_example_from_cache(VW::workspace* all, io_buf& buf, v_array<example*>& examples);
```
olgavrou, line 33: why do we need to pass in `all` here? In general it's best to avoid passing the all object when it's not needed, and passing in more specific args to functions is desirable, so I am wondering why we are going from `io_buf` to `all`?

vowpalwabbit/cache.cc
```diff
@@ -100,68 +157,20 @@ int VW::read_example_from_cache(io_buf& input, example* ae, label_parser& lbl_pa
...
-      VW::io::logger::errlog_error("truncated example! {} {} ", temp, char_size + sizeof(size_t));
-      return 0;
-    }
-
-    index = *reinterpret_cast<unsigned char*>(c);
-    c += sizeof(index);
-
+    unsigned char index;
+    features ours;
+    std::pair<size_t, bool> feat_size_and_sorted = read_cached_index_and_features(input, index, ours);
```
olgavrou, line 164: this is creating a pair object every time we read an example from cache, we should really check the benchmarks to see if there are any regressions

vowpalwabbit/cache.h
bassmang: This function definition is required because it's passed in as a function pointer to the parser. This is replacing the old `read_cached_features`

vowpalwabbit/example.cc
```diff
@@ -309,6 +310,45 @@ void return_multiple_example(VW::workspace& all, v_array<example*>& examples)
...
+  lbl_parser.cache_label(fe.l, fe._reduction_features, io);
+  bytes += write_model_field(io, fe.tag_len, upstream_name + "_tag_len", text);
+  bytes += write_model_field(io, fe.tag == nullptr, upstream_name + "_tag_is_null", text);
+  if (!(fe.tag == nullptr)) { bytes += write_model_field(io, *fe.tag, upstream_name + "_tag", text); }
+  bytes += write_model_field(io, fe.example_counter, upstream_name + "_example_counter", text);
+  bytes += write_model_field(io, fe.ft_offset, upstream_name + "_ft_offset", text);
+  bytes += write_model_field(io, fe.global_weight, upstream_name + "_global_weight", text);
+  bytes += write_model_field(io, fe.num_features, upstream_name + "_num_features", text);
+  bytes += write_model_field(io, fe.total_sum_feat_sq, upstream_name + "_total_sum_feat_sq", text);
+  cache_index_and_features(io, 0, fe.fs, parse_mask);
```
olgavrou, line 348: It would be good for the methods of reading/writing to mirror each other, i.e. have `[read_cached|cache]_index` and `[read_cached|cache+features]` or `[read_cached|cache]_index_and_features`

bassmang: Yes I agree

vowpalwabbit/cache.cc
```diff
@@ -70,98 +70,102 @@ void VW::write_example_to_cache(io_buf& output, example* ae, label_parser& lbl_p
...
+  }
+  input.set(c);
+  return total;
+}
 
-    char* end = c + storage;
+int VW::read_example_from_cache(VW::workspace* all, io_buf& input, v_array<example*>& examples)
+{
+  char* read_ptr;
+  if (input.buf_read(read_ptr, sizeof(uint64_t)) < sizeof(uint64_t)) { return 0; }
```
jackgerrits, line 146: Even though `size` had an "unused for now" comment, can you please leave a comment to describe what the value is. This was added to be able to skip cached examples on a read failure and is intended to be used.

vowpalwabbit/cache.cc
```diff
@@ -70,98 +70,102 @@ void VW::write_example_to_cache(io_buf& output, example* ae, label_parser& lbl_p
...
 {
   temp_buffer._backing_buffer->clear();
-  lbl_parser.cache_label(ae->l, ae->_reduction_features, temp_buffer._temporary_cache_buffer, "", false);
-  cache_features(temp_buffer._temporary_cache_buffer, ae, parse_mask);
-  temp_buffer._temporary_cache_buffer.flush();
+  io_buf& temp_cache = temp_buffer._temporary_cache_buffer;
+  lbl_parser.cache_label(ae->l, ae->_reduction_features, temp_cache, "_label", false);
+  cache_tag(temp_cache, ae->tag);
+  temp_cache.write_value<unsigned char>(ae->is_newline ? newline_example : non_newline_example);
+  temp_cache.write_value<unsigned char>(static_cast<unsigned char>(ae->indices.size()));
```
jackgerrits, line 77: Maybe add an assert here that it somehow this value is not larger

vowpalwabbit/example.h
```diff
@@ -194,6 +195,11 @@ void return_multiple_example(VW::workspace& all, v_array<example*>& examples);
 
 using example_factory_t = example& (*)(void*);
 
+namespace model_utils
+{
+size_t read_model_field(io_buf&, flat_example&, label_parser&);
+size_t write_model_field(io_buf&, const flat_example&, const std::string&, bool, label_parser&, uint64_t);
```
jackgerrits, line 201: Put the argument names in the header

vowpalwabbit/kernel_svm.cc
```diff
@@ -322,22 +257,22 @@ void save_load_svm_model(svm_params& params, io_buf& model_file, bool read, bool
...
+      auto* tmp = &calloc_or_throw<svm_example>();
+      VW::model_utils::read_model_field(model_file, *fec, params.all->example_parser->lbl_parser);
+      tmp->ex = *fec;
       model->support_vec.push_back(tmp);
     }
     else
     {
-      fec = &(model->support_vec[i]->ex);
-      save_load_flat_example(model_file, read, fec);
+      VW::model_utils::write_model_field(model_file, model->support_vec[i]->ex, "_flat_example", false,
```
jackgerrits, line 274: Is the model binary output identical with the previous method?

bassmang: The previous method called `bin_write_fixed` / ``bin_read_fixed` on a flat example which resulted in a segfault. I made the new method analogous to how example is cached.

vowpalwabbit/cache.cc
bassmang: Readded a comment with `size`

vowpalwabbit/cache.cc
bassmang: Larger than what?

jackgerrits: Than an unsigned char

jackgerrits (approved):

vowpalwabbit/cache.cc
bassmang: Added

Issue #3501, opened 2021-12-02
Overflowing handling in coin


ataymano:
#### Describe the bug
It is possible to turn coin weights into inf with single example

#### To Reproduce
data.txt
```
shared |User device:10000000000
0:-1:0.5 |Action ad:1
|Action ad:1
```
command line:
`vw --cb_explore_adf --coin -d data.txt -q ::  --invert_hash weights.txt`

weights.txt:
```
Version 8.11.0
Id 
Min label:-1
Max label:0
bits:18
lda:0
0 ngram:
0 skip:
options: --cb_adf --cb_explore_adf --cb_type mtr --coin --csoaa_ldf multiline --csoaa_rank --quadratic ::
Checksum: 970181571
event_sum 1
action_sum 2
:0
User^device:10281:-nan(ind)
Constant:116060:-nan(ind)
Action^ad:119437:-nan(ind)
Action^ad*User^device:154590:-nan(ind)
User^device*User^device:200866:-nan(ind)
Action^ad*Action^ad:237946:-nan(ind)
```

#### Expected behavior
Weights should not be nan since it is unrecoverable

#### Observed Behavior
Weights are nan => Any further learning on good data is impossible

#### Environment
Windows 11, master 12/02/2021

ataymano:
@bremen79, could you please suggest the right fix here?
Root cause is that all weights depend on average feature vector norm, so once we overflow it, nan(ind) is propagated to all weights.
I have prepared naive fix candidate - validating norm_sum before committing it and skipping example in case of overflowing:
https://github.com/VowpalWabbit/vowpal_wabbit/pull/3532 
but happy to hear other options.

bremen79:
Sorry, I am bit rusty on this code and I don't see immediately why it overflows.
Moreover, if I remember correctly, this part of the code is taken with minor changes from the gd code, see gd.cc lines 576-587.
So, does it overflow there as well?

ataymano:
oh you are right - it can be fixed with just operation order change. Let me send new PR

bremen79:
Nice, it seems good to me.

PR #3532, opened 2021-12-14
No inf(nan) weights assignment during coin betting

ataymano:


ataymano:
Attempt to fix #3501

vowpalwabbit/ftrl.cc
```diff
@@ -357,10 +365,16 @@ template <bool audit>
...
-  coin_betting_update_after_prediction(a, ec);
+  if (coin_betting_predict(a, base, ec))
+  {
+    if (audit) GD::print_audit_features(*(a.all), ec);
+    // update state based on the prediction
+    coin_betting_update_after_prediction(a, ec);
+  }
+  else
+  {
+    logger::errlog_error("Your features have too much magnitude. Example is ignored");
```
olgavrou, line 376:
```suggestion
    logger::errlog_error("The features have too much magnitude. Example is ignored");
```

I think we should start avoiding using `You` in error logs where possible (see https://github.com/VowpalWabbit/vowpal_wabbit/pull/3529 trying to get error messages to be less agressive)

ataymano: Yeah makes sense.

test/train-sets/ref/regression_simple_1var_overflow.stderr
```diff
@@ -0,0 +1,38 @@
...
+power_t = 0.5
+using no cache
+Reading datafile = train-sets/regression_simple_1var_overflow.txt
+num sources = 1
+Enabled reductions: ftrl-Coin Betting, generate_interactions, scorer-identity, count_label
+Input label = simple
+Output pred = scalar
+average  since         example        example  current  current  current
+loss     last          counter         weight    label  predict features
+[error] Your features have too much magnitude. Example is ignored
```
olgavrou, line 20: error msg changed so this also needs to change (Tests failing)

vowpalwabbit/ftrl.cc
```diff
@@ -221,22 +221,30 @@ void inner_coin_betting_update_after_prediction(ftrl_update_data& d, float x, fl
...
 }
 
-void coin_betting_predict(ftrl& b, base_learner&, example& ec)
+bool coin_betting_predict(ftrl& b, base_learner&, example& ec)
 {
+  constexpr float x2_max = FLT_MAX;
   b.data.predict = 0;
   b.data.normalized_squared_norm_x = 0;
 
   size_t num_features_from_interactions = 0;
```
olgavrou, line 230: do we want to count the num_features_from_iteractions if we are not going to use this example to update? Should lines 231 and 232 also go in the `if` statement?

ataymano: not sure.
right now, this variable is responsible for features from interactions only.
in current test there are 11 examples with 1 feature (+ 1 constant + 1 interaction).
Without change that you are proposing final example counter is 33 (although 1 example is ignored).
If we apply it, then counter drops to 32 which seems inconsistent to me - it would make sense to return 30 if we do not want to count ignored examples.
And not sure about any existing conventions what we should do in this case since there is no precedent of ignoring examples during learning as far as i know.

olgavrou: ok fair enough, seems like we should leave count as is then

olgavrou (approved):

PR #3533, opened 2021-12-14
feat: enable 0-indexed labels for csoaa

bassmang:


vowpalwabbit/csoaa.cc
```diff
@@ -34,22 +34,34 @@ namespace CSOAA
...
+  int indexing = -1;
+  bool search = false;
   polyprediction* pred = nullptr;
   ~csoaa() { free(pred); }
 };
 
 template <bool is_learn>
 inline void inner_loop(single_learner& base, example& ec, uint32_t i, float cost, uint32_t& prediction, float& score,
-    float& partial_prediction)
+    float& partial_prediction, int& indexing)
```
olgavrou, line 45:
```suggestion
    float& partial_prediction, int indexing)
```

if `indexing` isn't modified by the function it should be passed in by value since it's a built in type

vowpalwabbit/csoaa.cc
```diff
@@ -137,13 +189,15 @@ base_learner* csoaa_setup(VW::setup_base_i& stack_builder)
   VW::workspace& all = *stack_builder.get_all_pointer();
   auto c = VW::make_unique<csoaa>();
   option_group_definition new_options("Cost Sensitive One Against All");
-  new_options.add(
-      make_option("csoaa", c->num_classes).keep().necessary().help("One-against-all multiclass with <k> costs"));
+  new_options
+      .add(make_option("csoaa", c->num_classes).keep().necessary().help("One-against-all multiclass with <k> costs"))
+      .add(make_option("indexing", c->indexing).one_of({0, 1}).keep().help("Choose between 0 or 1-indexing"));
```
olgavrou, line 194: should this be `necessary`? what happens if it isn't supplied?

bassmang: It will auto-detect if the flag isnt supplied. Eg if k=4 and 4 is seen it will become 1-indexed, if 0 is seen it will become 0-indexed, which effects how warnings are given for future mislabels

olgavrou: should we output an `[info]` msg with the auto indexing chosen?

bassmang: Yea good idea -- added it for oaa too

olgavrou (approved):

Issue #3526, opened 2021-12-09
Update dump_options to understand one_of


jackgerrits:
The dump options program needs to be updated so it can understand and produce output for the new `one_of` type. 

https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/utl/dump_options/main.cc

PR #3540, opened 2021-12-17
fix: add one_of to dump_options, Resolves: #3526

bassmang:
Resolves: VowpalWabbit#3526


utl/dump_options/main.cc
```diff
@@ -284,6 +284,9 @@ struct options_exporter : options_i
       option_obj.AddMember("short_name", short_name_value, allocator);
       option_obj.AddMember("keep", option->m_keep, allocator);
       option_obj.AddMember("necessary", option->m_necessary, allocator);
+      rapidjson::Value one_of_value;
+      one_of_value.SetString(option->m_one_of_str, allocator);
+      option_obj.AddMember("one_of", one_of_value, allocator);
```
jackgerrits, line 289: I feel like this should only be added if the option is in fact a one_of option? Additionally, this should be a list of possible values and not just a formatted string

bassmang: Would it be sufficient to use a list of possible values, each as a string? (as opposed to string or int). Since untyped base_option's are passed in here, it wouldn't be possible to maintain the types for one_of unless we typed base_option, or used some caching scheme to decipher the type here. Also short_name is included regardless of being set so I'll also change that to only include if it's set.

jackgerrits: The base option has a type id you can use to dynamic cast the option. That's what we're doing for the other things.

Sounds like a good idea to fix short option. Maybe we should use `null` to indicate a missing optional value?

bassmang: I just used the empty string to detect if a short name is not present, since this is the default value if nothing is indicated.

vowpalwabbit/options.h
```diff
@@ -119,6 +120,7 @@ struct base_option
   size_t m_type_hash;
   std::string m_help = "";
   std::string m_short_name = "";
+  std::string m_one_of_str = "";
```
jackgerrits, line 123: Why not store the set instead of serializing and deserializing later?

bassmang: The set can either be `int` or `string` which would require templating.

jackgerrits: Typed option is templated based on the value of the option, that seems like the logical place for this to live

bassmang: Changed to dynamic cast base option to typed option

jackgerrits (approved):

PR #3541, opened 2021-12-21
test: fix test false positive, add types, use pathlib, change from format to fstrings

jackgerrits:
- Fix false positive test passing
  - Before this change `difflib` was used to determine if there were changes. If there was a diff, then we compared each different line to see if there were float differences that were close enough. However, if the diff differed because of ordering of lines this method would falsely accept the diff as close enough. This is resolved by not using `difflib` for the actual comparison and only using it to generate a diff should they be deemed to be different with a line by line comparison
- Add types
- Use `pathlib.Path` instead of string based paths whenever possible
- Use f-strings now that we can safely assume at least Python 3.6 is being used


test/run_tests.py
```diff
@@ -42,20 +42,87 @@ class Result(Enum):
     SKIPPED = 3
 
 
-def try_decode(binary_object):
+class Completion:
```
ataymano, line 45: Looks like type hinting is forgotten here

jackgerrits: I explicitly typed `completed` in `__init__`, and left `lock` and `condition` to be implicitly typed based on their initialization

ataymano: not sure if I understand. I was all methods signatures - id/success are unhinted.

test/run_tests.py
```diff
@@ -42,20 +42,87 @@ class Result(Enum):
...
+        checks: Dict[str, Union[StatusCheck, DiffCheck]],
+        skip_reason: Optional[str] = None,
+    ):
+        self.id = id
+        self.result = result
+        self.checks = checks
+        self.skip_reason = skip_reason
+
+
+def try_decode(binary_object) -> str:
```
ataymano, line 112: and the same here - type hint for binary_object?

test/run_tests.py
jackgerrits: Oh youre so right i missed those, thanks! I misunderstood originally

test/run_tests.py
jackgerrits: I'm not actually sure what the type of binary object is, I'll look into that one

test/run_tests.py
```diff
@@ -130,21 +200,25 @@ def is_line_different(output_line, ref_line, epsilon):
...
             return (
                 True,
-                "Floats don't match {} {}".format((output_token), (ref_token)),
+                f"Floats don't match {output_token} {ref_token}",
                 found_close_floats,
             )
         else:
             if output_token != ref_token:
+                print(output_tokens)
+                print(ref_tokens)
```
bassmang, line 209: Are these prints meant to be here?

test/run_tests.py
```diff
@@ -370,131 +373,89 @@ def run_command_line_test(
...
             if output_file == "stdout":
                 output_content = stdout
             elif output_file == "stderr":
                 output_content = stderr
             else:
-                output_file_working_dir = os.path.join(working_dir, output_file)
-                if os.path.isfile(output_file_working_dir):
-                    output_content = open(
-                        output_file_working_dir, "r", encoding="utf-8"
+                output_file_working_dir = current_test_working_dir / output_file
```
bassmang, line 394: Cool didnt know this was possible

test/run_tests.py
jackgerrits: No they arent, thanks for catching!

test/run_tests.py
jackgerrits: Yeah pathlib is super nice

bassmang (approved):

Issue #3043, opened 2021-06-07
Label accessors for CCB, Slates, and Continuous Actions missing in pyvw


lokitoth:
The pylibvw example object has accessors for most labels ~and predictions~, except for the newest ones: CCB, Slates and Continuous Actions. We should ensure these are implemented ~, and backfill [`pyvw.get_prediction()`]~(https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/vowpalwabbit/pyvw.py#L211) with them.

jackgerrits:
Predictions are implemented: 
- CCB, Slates - [decision_scores](https://github.com/VowpalWabbit/vowpal_wabbit/blob/8786469ee0b062c6c443c10b86c6c9cedd6e47b7/python/vowpalwabbit/pyvw.py#L256)
- Continuous actions - [PDF and PDF_Value](https://github.com/VowpalWabbit/vowpal_wabbit/blob/8786469ee0b062c6c443c10b86c6c9cedd6e47b7/python/vowpalwabbit/pyvw.py#L257)

Labels are indeed missing

jonpsy:
Would like to take a stab at this, since I don't see much progress here.

jackgerrits:
#2500 is a bit different, it is for the Pandas to VW converter.

An example for this PR would be: https://github.com/VowpalWabbit/vowpal_wabbit/blob/a95be782de07d7a1619eef3b704c062aea12100b/python/vowpalwabbit/pyvw.py#L1031

jonpsy:
Okay, but it looks like I need to define the accessors in the `pyvwlib.cc` file as well? Then I have to register them to be exportable as python functions.

jackgerrits:
Yes, there will need to be some new functions available in the binding layer

jonpsy:
Let's continue the discussion over the PR. Thanks.

PR #3546, opened 2021-12-29
feat: Label accessor for CCB

jonpsy:
Aims to add label accessors to the aforementioned classes. @jackgerrits @lokitoth  #3043 

**TODO**:

- [x] CCB .cc label accessors
- [x] Python interface implementation
 ~Slates .cc label accessors~
 ~Continuous Action .cc label accessors~


python/pylibvw.cc
```diff
@@ -857,6 +857,37 @@ uint32_t ex_get_cbandits_class(example_ptr ec, uint32_t i) { return ec->l.cb.cos
...
+}
+float ex_get_ccbandits_cost(example_ptr ec)
+{ 
+  auto* outcome_ptr = ec->l.conditional_contextual_bandit->conditional_contextual_bandit_outcome;
+  return outcome_ptr == nullptr ? 0.f : outcome_ptr->cost;
+}
+uint32_t ex_get_ccbandits_class(example_ptr ec, uint32_t i)
+{
+  auto* outcome_ptr = ec->l.conditional_contextual_bandit->conditional_contextual_bandit_outcome;
+  return outcome_ptr == nullptr ? -1 : outcome_ptr->probabilities[i]->action;
```
bassmang, line 883: unsigned int cannot be negative, 0 should be fine instead

bassmang:
Hi @jonpsy,
This looks like a good start. I added a couple of comments on your code. I would suggest that you use `ccb` instead of `ccbandit` for clarity. Additionally you should include functions for other members of the ccb label such as `weight` and `explicit_included_actions`. If you do the same for slates and continuous_actions, and put in the python interface it should be good to go.

jonpsy:
Lovely, thanks. I'll get onto it!

python/pylibvw.cc
```diff
@@ -869,24 +868,35 @@ const char* ex_get_ccbandits_type(example_ptr ec)
...
+  auto* outcome_ptr = ec->l.conditional_contextual_bandit.conditional_contextual_bandit_outcome;
   return outcome_ptr == nullptr ? 0.f : outcome_ptr->probabilities[i]->score;
 }
+float ex_get_ccb_weight(example_ptr ec) { return ec->l.conditional_contextual_bandit.weight; }
+uint32_t ex_get_ccb_num_included_actions(example_ptr ec)
+{
+  auto label = ec->l.conditional_contextual_bandit;
+  return label.type != CCB::slot ? 0 : label.explicit_included_actions.size();
+}
+int ex_get_included_action(example_ptr ec, uint32_t i)
```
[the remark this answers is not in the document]
jonpsy: Okay, should I return a `boost::py::list` by defining a custom method to convert `varraytopylist` OR I can use `boost_indexing_suite` to define `uvec` like this:

```c++
BOOST_PYTHON_MODULE(pylibvw)
{  
    ...
  // define unsigned v_array class
  py::class_<v_array<uint32_t> >("uvec")
      .def(vector_indexing_suite<v_array<uint32_t> >());
}
```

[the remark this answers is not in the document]
bassmang: it's probably best to return `py::list` as with similar functions such as `ex_get_action_scores`

python/pylibvw.cc
```diff
@@ -857,6 +857,47 @@ uint32_t ex_get_cbandits_class(example_ptr ec, uint32_t i) { return ec->l.cb.cos
 float ex_get_cbandits_probability(example_ptr ec, uint32_t i) { return ec->l.cb.costs[i].probability; }
 float ex_get_cbandits_partial_prediction(example_ptr ec, uint32_t i) { return ec->l.cb.costs[i].partial_prediction; }
 
+const char* ex_get_ccb_type(example_ptr ec)
+{
+  switch(ec->l.conditional_contextual_bandit.type)
+  {
+    case CCB::example_type::shared:
```
jackgerrits, line 864: We should something enum like rather than a string. For example:
```c++
const size_t pSCALAR = 0;
```

python/pylibvw.cc
```diff
@@ -857,6 +857,47 @@ uint32_t ex_get_cbandits_class(example_ptr ec, uint32_t i) { return ec->l.cb.cos
...
+    case CCB::example_type::slot:
+      return "slot";
+    default:
+      return "unset";
+  }
+}
+float ex_get_ccb_cost(example_ptr ec)
+{
+  auto* outcome_ptr = ec->l.conditional_contextual_bandit.conditional_contextual_bandit_outcome;
+  return outcome_ptr == nullptr ? 0.f : outcome_ptr->cost;
```
jackgerrits, line 877: If the outcome doesn't exist, Python should probably receive None instead of some default possible valid value

jackgerrits (commented):
What is the intended Python interface? I think adding that first so we can review and agree on it before implementing the glue functions would be better

jonpsy:
> What is the intended Python interface? I think adding that first so we can review and agree on it before implementing the glue functions would be better

From the [wiki](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Conditional-Contextual-Bandit#vw-text-format) page, let's say we use the same example as was used there. Then the label should could be as:

```python
shared
shared
action
slot top_action:cost:prob, action:prob, action:prob <explicitly_included_actions>
```
@jackgerrits  Let me know if this sounds good to you?

python/vowpalwabbit/pyvw.py
```diff
@@ -1157,6 +1157,47 @@ def __str__(self):
             ["{}:{}:{}".format(c.action, c.cost, c.probability) for c in self.costs]
         )
 
+class ccb_label(abstract_label):
+    """Class for conditional contextual bandits VW label"""
+
+    def __init__(self, type = "unset", cost=0, top_class=0, top_probability=0, explicitly_included_actions=[]):
+        abstract_label.__init__(self)
+        if isinstance(cost, example):
```
jackgerrits, line 1165: Should be `type` as that is the first arg no?
```suggestion
        if isinstance(type, example):
```

python/vowpalwabbit/pyvw.py
```diff
@@ -1157,6 +1157,47 @@ def __str__(self):
...
+                self.weight = weight
+                self.top_action, self.top_probability = top_prediction
+                self.label = self.top_action
+                self.cost = cost
+                self.explicitly_included_actions = explicitly_included_actions
+        self.out_label = wclass(ex.get_ccb_type(), get_prediction(ex, pylibvw.vw.pDECISION_SCORES),
+                ex.get_ccb_cost(), ex.get_ccb_weight(), ex.get_ccb_explicitly_included_actions())
+
+    def __str__(self):
+        return f"{self.type}:{str(self.top_action or ' ')}:{str(self.top_probability or ' ')}:{self.explicitly_included_actions}"
```
jackgerrits, line 1200: This format is wrong. The label format can be found [here](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Conditional-Contextual-Bandit#vw-text-format)

python/vowpalwabbit/pyvw.py
```diff
@@ -1157,6 +1157,47 @@ def __str__(self):
...
 
+class ccb_label(abstract_label):
+    """Class for conditional contextual bandits VW label"""
+
+    def __init__(self, type = "unset", cost=0, top_class=0, top_probability=0, explicitly_included_actions=[]):
+        abstract_label.__init__(self)
+        if isinstance(cost, example):
+            self.from_example(cost)
+        else:
+            self.type = type
```
jackgerrits, line 1168: This object does not map to the contextual bandit label object. See [here](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Conditional-Contextual-Bandit#label-type) for what it contains.

I was thinking something along the lines of:
```Python
class SlotOutcome:
  cost: float
  action_probs: List[Tuple[int, float]]

class CCBLabel:
  type: CCBLabelType
  explicit_included_actions: List[int]
  weight: float
  outcome: Optional[SlotOutcome]
```

jonpsy: @jackgerrits Updated it. Also, I think we should use snake case for naming classes?

python/vowpalwabbit/pyvw.py
```diff
@@ -1157,6 +1159,59 @@ def __str__(self):
...
+class ccb_label(abstract_label):
+  type: ccb_label_type = ccb_label_type.UNSET
+  explicit_included_actions: List[int] = []
+  weight: float = 0
+  outcome: Optional[slot_outcome] = None
+"""Class for conditional contextual bandits VW label"""
+
+    def __init__(self, type: ccb_label_type, explicit_included_actions: List[int], weight: float, outcome: Optional[slot_outcome], ex=None):
+        abstract_label.__init__(self)
+        if isinstance(ex, example):
```
jonpsy, line 1201: @jackgerrits @bassmang  I'm a little confused here, how do I keep an optional `example` as an argument here. This way user can extract information directly from example. I don't think python supports multiple constructors. Let me know your thoughts!

python/vowpalwabbit/pyvw.py
jackgerrits: This file is far from consistent, but for new code we should try and follow the [PEP 8 guideline](https://www.python.org/dev/peps/pep-0008/#class-names) of PascalCase for classes.

python/vowpalwabbit/pyvw.py
jackgerrits: In the same way other label types do it. For example here: https://github.com/VowpalWabbit/vowpal_wabbit/blob/6318d17f36599cab70902edf7de1dc142538c326/python/vowpalwabbit/pyvw.py#L1038

`label` is the first arg. It is either a `float` or an `example`. This just makes use of the fact it is a positional argument.

jonpsy: I was thinking of  this approach:

```python
def __init__(self, type, ...., **kwargs):
  ex = kwargs.get('ex')
  if isinstance(ex, example):
      ....
```

python/vowpalwabbit/pyvw.py
```diff
@@ -1158,6 +1160,72 @@ def __str__(self):
         )
 
 
+class CCBLabelType(enum.Enum):
```
jonpsy, line 1163: This is used by both `Slates` & `CCB`. Perhaps we can name it as just `LabelType`?

python/vowpalwabbit/pyvw.py
jonpsy: Correct me if I'm wrong, you're asking whether we should allow single positional arg input to labels (and expecting that to be `example` class`). 

I think nope, this method is much more clear in its intent.

python/pylibvw.cc
```diff
@@ -857,6 +871,49 @@ uint32_t ex_get_cbandits_class(example_ptr ec, uint32_t i) { return ec->l.cb.cos
 float ex_get_cbandits_probability(example_ptr ec, uint32_t i) { return ec->l.cb.costs[i].probability; }
 float ex_get_cbandits_partial_prediction(example_ptr ec, uint32_t i) { return ec->l.cb.costs[i].partial_prediction; }
 
+const size_t ex_get_ccb_type(example_ptr ec)
```
jackgerrits, line 874: redundant const 
```suggestion
size_t ex_get_ccb_type(example_ptr ec)
```

python/pylibvw.cc
```diff
@@ -560,6 +565,15 @@ std::string varray_char_to_string(v_array<char>& a)
   return ret;
 }
 
+template <class T>
+py::list varray_to_pylist(const v_array<T>& a)
+{
+	py::list list;
+  for (const auto& elem : a):
+    list.append(a)
+	return list;
```
jackgerrits, line 574: This wont compile. 
```suggestion
 	py::list list;
  for (const auto& elem : a) 
  {
    list.append(a);
  }
 	return list;
```

python/pylibvw.cc
```diff
@@ -857,6 +871,49 @@ uint32_t ex_get_cbandits_class(example_ptr ec, uint32_t i) { return ec->l.cb.cos
...
+      return tSHARED;
+    case CCB::example_type::action:
+      return tACTION;
+    case CCB::example_type::slot:
+      return tSLOT;
+    default:
+      return tUNSET;
+  }
+}
+float ex_get_ccb_cost(example_ptr ec)
```
jackgerrits, line 888: This return type does not seem right for something that can return `None`?

python/pylibvw.cc
```diff
@@ -857,6 +871,49 @@ uint32_t ex_get_cbandits_class(example_ptr ec, uint32_t i) { return ec->l.cb.cos
...
+{
+  auto* outcome_ptr = ec->l.conditional_contextual_bandit.conditional_contextual_bandit_outcome;
+  return outcome_ptr == nullptr ? Py_None : outcome_ptr->cost;
+}
+size_t ex_get_ccb_class(example_ptr ec, uint32_t i)
+{
+  auto* outcome_ptr = ec->l.conditional_contextual_bandit.conditional_contextual_bandit_outcome;
+  return outcome_ptr == nullptr ? Py_None : outcome_ptr->probabilities[i]->action;
```
jackgerrits, line 896: Please add a bounds check for `i`

python/pylibvw.cc
```diff
@@ -857,6 +871,49 @@ uint32_t ex_get_cbandits_class(example_ptr ec, uint32_t i) { return ec->l.cb.cos
...
+}
+float ex_get_ccb_probability(example_ptr ec, uint32_t i)
+{
+  auto* outcome_ptr = ec->l.conditional_contextual_bandit.conditional_contextual_bandit_outcome;
+  return outcome_ptr == nullptr ? Py_None : outcome_ptr->probabilities[i]->score;
+}
+float ex_get_ccb_weight(example_ptr ec) { return ec->l.conditional_contextual_bandit.weight; }
+uint32_t ex_get_ccb_num_included_actions(example_ptr ec)
+{
+  auto label = ec->l.conditional_contextual_bandit;
```
jackgerrits, line 906: This is currently making a copy
```suggestion
  const auto& label = ec->l.conditional_contextual_bandit;
```

python/pylibvw.cc
```diff
@@ -857,6 +871,49 @@ uint32_t ex_get_cbandits_class(example_ptr ec, uint32_t i) { return ec->l.cb.cos
...
+  return label.explicit_included_actions.size();
+}
+py::list ex_get_ccb_explicitly_included_actions(example_ptr ec)
+{
+  auto label = ec->l.conditional_contextual_bandit;
+  if (label.type != CCB::slot)
```
jackgerrits, line 912: Please add braces

python/pylibvw.cc
```diff
@@ -857,6 +871,49 @@ uint32_t ex_get_cbandits_class(example_ptr ec, uint32_t i) { return ec->l.cb.cos
...
+}
+float ex_get_ccb_weight(example_ptr ec) { return ec->l.conditional_contextual_bandit.weight; }
+uint32_t ex_get_ccb_num_included_actions(example_ptr ec)
+{
+  auto label = ec->l.conditional_contextual_bandit;
+  return label.explicit_included_actions.size();
+}
+py::list ex_get_ccb_explicitly_included_actions(example_ptr ec)
+{
+  auto label = ec->l.conditional_contextual_bandit;
```
jackgerrits, line 911:
```suggestion
  const auto& label = ec->l.conditional_contextual_bandit;
```

python/pylibvw.cc
```diff
@@ -857,6 +871,49 @@ uint32_t ex_get_cbandits_class(example_ptr ec, uint32_t i) { return ec->l.cb.cos
...
+{
+  auto* outcome_ptr = ec->l.conditional_contextual_bandit.conditional_contextual_bandit_outcome;
+  return outcome_ptr == nullptr ? Py_None : outcome_ptr->cost;
+}
+size_t ex_get_ccb_class(example_ptr ec, uint32_t i)
+{
+  auto* outcome_ptr = ec->l.conditional_contextual_bandit.conditional_contextual_bandit_outcome;
+  return outcome_ptr == nullptr ? Py_None : outcome_ptr->probabilities[i]->action;
+}
+float ex_get_ccb_probability(example_ptr ec, uint32_t i)
```
jackgerrits, line 898: This return type does not seem right for something that can return None?

python/pylibvw.cc
```diff
@@ -857,6 +871,49 @@ uint32_t ex_get_cbandits_class(example_ptr ec, uint32_t i) { return ec->l.cb.cos
...
+    default:
+      return tUNSET;
+  }
+}
+float ex_get_ccb_cost(example_ptr ec)
+{
+  auto* outcome_ptr = ec->l.conditional_contextual_bandit.conditional_contextual_bandit_outcome;
+  return outcome_ptr == nullptr ? Py_None : outcome_ptr->cost;
+}
+size_t ex_get_ccb_class(example_ptr ec, uint32_t i)
```
jackgerrits, line 893: This return type does not seem right for something that can return None?

python/vowpalwabbit/pyvw.py
```diff
@@ -1158,6 +1160,74 @@ def __str__(self):
...
+    SHARED = pylibvw.vw.tSHARED
+    ACTION = pylibvw.vw.tACTION
+    SLOT = pylibvw.vw.tSLOT
+
+
+class CCBSlotOutcome:
+    cost: float
+    action_probs: List[Tuple[int, float]]
+
+    def __init__(self, cost=0, action_probs=[], **kwargs):
```
jackgerrits, line 1174: Add types?

python/vowpalwabbit/pyvw.py
```diff
@@ -1158,6 +1160,74 @@ def __str__(self):
...
+        ex = kwargs.get("ex")
```
jackgerrits, line 1175: add type?

python/vowpalwabbit/pyvw.py
```diff
@@ -1158,6 +1160,74 @@ def __str__(self):
...
+        if isinstance(ex, example):
+            self.from_example(ex)
+        else:
+            self.cost = cost
+            self.action_probs = action_probs
+
+    def from_example(self, ex):
```
jackgerrits, line 1182: add type

python/vowpalwabbit/pyvw.py
```diff
@@ -1158,6 +1160,74 @@ def __str__(self):
...
+    explicit_included_actions: List[int]
+    weight: float
+    outcome: Optional[CCBSlotOutcome]
+
+    def __init__(
+        self,
+        type=CCBLabelType.UNSET,
+        explicit_included_actions=[],
+        weight=1,
+        outcome=None,
```
jackgerrits, line 1209: Arg types

python/vowpalwabbit/pyvw.py
```diff
@@ -1158,6 +1160,74 @@ def __str__(self):
...
+                CCBSlotOutcome(ex=ex) if self.type == CCBLabelType.SLOT else None
+            )
+        else:
+            self.type = type
+            self.explicit_included_actions = explicit_included_actions
+            self.weight = weight
+            self.outcome = outcome
+
+    def __str__(self):
+        return f"ccb {self.type} {str(self.outcome or '')} {str(self.explicit_included_actions or '')}"
```
jackgerrits, line 1228: `self.type` needs some sort of string conversion since it is an enum here

jonpsy: `self.type.name.lower()` used this trick.

python/pylibvw.cc
```diff
@@ -857,6 +873,50 @@ uint32_t ex_get_cbandits_class(example_ptr ec, uint32_t i) { return ec->l.cb.cos
...
+uint32_t ex_get_ccb_num_included_actions(example_ptr ec)
+{
+  const auto& label = ec->l.conditional_contextual_bandit;
+  return label.explicit_included_actions.size();
+}
+py::object* to_pyobject ex_get_ccb_class(example_ptr ec, uint32_t i)
+{
+  if (i >= ex_get_ccb_num_included_actions(ec)) { THROW("Class index out of bounds"); }
+  auto* outcome_ptr = ec->l.conditional_contextual_bandit.conditional_contextual_bandit_outcome;
+  return outcome_ptr == nullptr ? Py_None : value_to_pyobject<uint32_t>(outcome_ptr->probabilities[i]->action);
```
jonpsy, line 904: @jackgerrits I suppose this is the correct way?

jackgerrits: Looks like object could be used for this?
https://www.boost.org/doc/libs/1_64_0/libs/python/doc/html/tutorial/tutorial/object.html

Not sure if this will work but a look at that file makes me think it should work this way:
```c++
py::object to_pyobject ex_get_ccb_class(example_ptr ec, uint32_t i)
{
  if (i >= ex_get_ccb_num_included_actions(ec)) { THROW("Class index out of bounds"); }
  auto* outcome_ptr = ec->l.conditional_contextual_bandit.conditional_contextual_bandit_outcome;
  return outcome_ptr == nullptr ? py::object() : py::object(outcome_ptr->probabilities[i]->action);
}
```

jackgerrits:
Looks like the CI is not passing for this PR. Could you also please add some tests [here](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/tests/test_ccb.py). In particular exercising the `None` return since we aren't confident on that one

python/pylibvw.cc
```diff
@@ -1335,7 +1395,21 @@ BOOST_PYTHON_MODULE(pylibvw)
...
       .def("get_cbandits_partial_prediction", &ex_get_cbandits_partial_prediction,
           "Assuming a contextual_bandits label type, get the partial prediction for a given pair (i=0.. "
-          "get_cbandits_num_costs)");
+          "get_cbandits_num_costs)")
+      .def("get_ccb_type", &ex_get_ccb_type,
+          "Assuming a conditional_contextual_bandits label type, get the type of example")
+      .def("get_ccb_cost", &ex_get_ccb_cost,
+          "Assuming a conditional_contextual_bandits label type, get the cost of the given label")
+      ex_.def("get_ccb_num_included_actions", &ex_get_ccb_num_included_actions,
+          "Assuming a conditional_contextual_bandits label type, get the set of included actions.")
```
bassmang, line 1404:
```suggestion
          "Assuming a conditional_contextual_bandits label type, get the number of included actions.")
```

bassmang (commented):
@jonpsy there seem to be some compilation errors which you can see here:
https://github.com/VowpalWabbit/vowpal_wabbit/runs/4790458435?check_suite_focus=true
Try to compile locally with `python3 setup.py install` to catch the errors

python/vowpalwabbit/pyvw.py
```diff
@@ -1360,6 +1423,9 @@ def __init__(
         self.labelType: Optional[LabelType] = None
         if isinstance(labelType, LabelType):
             label_int = labelType.value
+            # Keep self.labelType as None if provided label is DEFAULT.
+            if labelType != LabelType.DEFAULT:
+                self.labelType = labelType
```
jonpsy, line 1428: @bassmang  @jackgerrits  For some reason we don't assign `self.labelType` in this block? Caught this while testing.

python/vowpalwabbit/pyvw.py
```diff
@@ -1224,6 +1232,61 @@ def __init__(self, action: int, score: float):
...
+        explicit_included_actions: List[int] = [],
+        weight: float = 1,
+        outcome: Optional[CCBSlotOutcome] = None,
+        **kwargs,
+    ):
+        AbstractLabel.__init__(self)
+        ex: example = kwargs.get("ex")
+        if isinstance(ex, example):
+            if not ex.labelType == LabelType.CONDITIONAL_CONTEXTUAL_BANDIT:
+                raise ValueError("Example should have CCB label type.")
```
jonpsy, line 1273: I think other classes should follow this, or am I missing something?

python/vowpalwabbit/pyvw.py
bassmang: Good catch!

python/vowpalwabbit/pyvw.py
bassmang: Yes this would be safer -- adding in a separate PR

bassmang (commented):
Hi @jonpsy 
The changes look good. The last thing I would add is a function `test_ccb_label()` similar to `test_slates_label()` here:
https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/tests/test_pyvw.py#L207

To make sure everything is being set right for a `shared` `action` and `slot` CCB example.

python/vowpalwabbit/pyvw.py
jonpsy: Okay, removing it (and the corresponding test) for the sake of maintaining scope.

python/tests/test_ccb.py
```diff
@@ -111,3 +111,14 @@ def test_ccb_single_slot_and_cb_non_equivalence_with_slot_features():
...
     # does not apply and so we expect there to be more weights in the CCB model.
     assert ccb_num_weights > cb_num_weights
+
+
+def test_ccb_non_slot_none_outcome():
+    model = pyvw.vw(quiet=True)
+    example = pyvw.example(
+        vw=model, labelType=pyvw.LabelType.CONDITIONAL_CONTEXTUAL_BANDIT
+    )
+    label = pyvw.CCBLabel(example)
```
jackgerrits, line 121: The label should be retrieved as:
```
label = example.get_label(pyvw.CCBLabel)
```

python/tests/test_ccb.py
```diff
@@ -111,3 +111,14 @@ def test_ccb_single_slot_and_cb_non_equivalence_with_slot_features():
     # Since there was at least one slot feature supplied, the equivalent mode
     # does not apply and so we expect there to be more weights in the CCB model.
     assert ccb_num_weights > cb_num_weights
+
+
+def test_ccb_non_slot_none_outcome():
+    model = pyvw.vw(quiet=True)
```
jackgerrits, line 117:
```suggestion
    model = pyvw.Workspace(quiet=True)
```

python/tests/test_ccb.py
```diff
@@ -111,3 +111,14 @@ def test_ccb_single_slot_and_cb_non_equivalence_with_slot_features():
...
+    example = pyvw.example(
```
jackgerrits, line 118:
```suggestion
    example = pyvw.Example(
```

python/vowpalwabbit/pyvw.py
```diff
@@ -1270,6 +1277,73 @@ def __init__(self, action: int, score: float):
...
+        out = "{}:{}:{}".format(top_action, round(self.cost, 2), round(top_score, 2))
+        for action_score in self.action_probs[1:]:
+            out += f",{action_score.action}:{action_score.score}"
+        return out
+
+
+class CCBLabel(AbstractLabel):
+    """Class for conditional contextual bandits VW label"""
+
+    def __init__(
```
jackgerrits, line 1309: This will need to be updated the match the new way from this PR: https://github.com/VowpalWabbit/vowpal_wabbit/pull/3615

jonpsy:
Okay. It'd be best to merge this sooner. Any further re-factor would include `CCB` rather than repeatedly changing this file w.r.t master branch.

python/vowpalwabbit/pyvw.py
```diff
@@ -1278,22 +1278,19 @@ def __init__(self, action: int, score: float):
...
+    def __init__(self, cost: float = 0, action_probs: List[ActionScore] = []):
+        self.cost = cost
+        self.action_probs = action_probs
 
-    def from_example(self, ex: "example"):
-        self.cost = ex.get_ccb_cost()
-        self.action_probs = []
+    @staticmethod
+    def from_example(ex: "Example"):
+        cost = ex.get_ccb_cost()
```
jackgerrits, line 1287: cost is optional based on the c++ code, so this won't match the types defined on line 1281

python/vowpalwabbit/pyvw.py
```diff
@@ -1308,25 +1305,24 @@ class CCBLabel(AbstractLabel):
...
+        self.explicit_included_actions = explicit_included_actions
+        self.weight = weight
+        self.outcome = outcome
+
+    @staticmethod
+    def from_example(ex: "Example"):
+        type = ex.get_ccb_type()
+        explicit_included_actions = ex.get_ccb_explicitly_included_actions()
+        weight = ex.get_ccb_weight()
+        outcome = CCBSlotOutcome.from_example(ex) if type == CCBLabelType.SLOT else None
```
jackgerrits, line 1324: You only want to set an outcome if one exists, not if it is a slot example. The c++ should project a `has_outcome` sort of function so that can be checked

python/vowpalwabbit/pyvw.py
```diff
@@ -1278,22 +1278,19 @@ def __init__(self, action: int, score: float):
...
-            self.action_probs = action_probs
+    def __init__(self, cost: float = 0, action_probs: List[ActionScore] = []):
+        self.cost = cost
+        self.action_probs = action_probs
 
-    def from_example(self, ex: "example"):
-        self.cost = ex.get_ccb_cost()
-        self.action_probs = []
+    @staticmethod
+    def from_example(ex: "Example"):
```
jackgerrits, line 1286: I think a `from_example` function is too much given that `CCBSlotOutcome` is not a label itself. `CCBLabel.from_example` should handle extracting outcome information if there is one

python/vowpalwabbit/pyvw.py
jonpsy: Question: Only slot type has outcome. For the slots which don't it can be an empty list inside `CCBSlotOutcome`.

python/pylibvw.cc
```diff
@@ -923,6 +931,50 @@ float ex_get_slates_probability(example_ptr ec, uint32_t i)
...
+    default:
+      return tUNSET;
+  }
+}
+py::object ex_get_ccb_cost(example_ptr ec)
+{
+  auto* outcome_ptr = ec->l.conditional_contextual_bandit.outcome;
+  return outcome_ptr == nullptr ? py::object() : py::object(outcome_ptr->cost);
+}
+uint32_t ex_get_ccb_num_included_actions(example_ptr ec)
```
jackgerrits, line 953:
```suggestion
uint32_t ex_get_ccb_num_explicitly_included_actions(example_ptr ec)
```

python/vowpalwabbit/pyvw.py
```diff
@@ -1270,6 +1277,69 @@ def __init__(self, action: int, score: float):
...
+class CCBSlotOutcome:
+    def __init__(self, cost: float = 0, action_probs: List[ActionScore] = []):
+        self.cost = cost
+        self.action_probs = action_probs
+
+    @staticmethod
+    def from_example(ex: "Example"):
+        cost = ex.get_ccb_cost()
+        action_probs = []
+        for i in range(ex.get_ccb_num_included_actions()):
```
jackgerrits, line 1289: `explicit_included_actions` does not tell you how many fields there are in the `probabilities` list. You need a function to get length of `probabilities`

python/vowpalwabbit/pyvw.py
jackgerrits: Yes but a slot is not guaranteed to have an outcome. There being an outcome or  not is actually an important distinction for two given labels, even if they are both slots

bassmang:
> Okay. It'd be best to merge this sooner. Any further re-factor would include `CCB` rather than repeatedly changing this file w.r.t master branch.

@jonpsy Agreed we should merge this sooner. It should be good to merge once the last few comments are resolved. One last thing, to get the C++ formatting tests to pass you'll want to run:
`git fetch upstream`
`git diff upstream/master...HEAD -U0 --no-color | clang-format-diff -r '^.*\.(cc|h)$' -p1 -i`

and for python:
`python3 -m black python/vowpalwabbit`

Both in the base directory, then add and commit the files which get altered.

python/vowpalwabbit/pyvw.py
jonpsy: Okay, also added a `slot pred`  to test which confirms the following behaviour

`ccb slot |` type of example should have no outcome even though its a slot type.

bassmang (approved):

jackgerrits (approved):
Looks great! Thank you! There are a couple of small things I'd like to cleanup but I can take that on in a separate PR

Issue #3556, opened 2022-01-04
Various warnings when building VW in Windows


bassmang:
#### To Reproduce
Build VW in Windows with Visual Studio 2019 x64.

#### Expected behavior
Compile cleanly without warnings

#### Observed Behavior
Multiple warnings such as:
`conditional expression is constant`
`declaration of 'num_actions' hides class member`
etc...

#### Environment
What version of VW did you use?
Master branch
What OS or language did you use?
Windows Visual Studio 2019 x64 Debug mode

PR #3560, opened 2022-01-04
fix: Suppress build warnings on windows

zwd-ms:
Addresses #3556


vowpalwabbit/cbify.cc
```diff
@@ -109,6 +109,8 @@ void finish_cbify_reg(cbify_reg& data, std::ostream* trace_stream)
   if (trace_stream != nullptr) (*trace_stream) << "Max Cost=" << data.max_cost << std::endl;
 }
 
+VW_WARNING_STATE_PUSH
```
olgavrou, line 112: could this warning be fixed with a rename instead ?

zwd-ms: yes that could be a fix, and I actually renamed the other case.
But since this is an init method, i thought maybe it makes sense to use the same name?
Instead of suppressing this warning, another way is perhaps to refactor using setter methods. WDYT?

explore/future_compat.h
```diff
@@ -62,12 +62,16 @@
     #define VW_WARNING_DISABLE_DEPRECATED_USAGE    VW_WARNING_DISABLE(4996)
 #  define VW_WARNING_DISABLE_BADLY_FORMED_XML VW_WARNING_DISABLE(4635)
 #  define VW_WARNING_DISABLE_CPP_17_LANG_EXT VW_WARNING_DISABLE(4984)
+#  define VW_WARNING_DISABLE_COND_CONST_EXPR VW_WARNING_DISABLE(4127)
+#  define VW_WARNING_DISABLE_HIDE_CLASS_MEMBER VW_WARNING_DISABLE(4458)
```
jackgerrits, line 66: I dont think this is a warning we want to suppress

explore/future_compat.h
```diff
@@ -62,12 +62,16 @@
     #define VW_WARNING_DISABLE_DEPRECATED_USAGE    VW_WARNING_DISABLE(4996)
 #  define VW_WARNING_DISABLE_BADLY_FORMED_XML VW_WARNING_DISABLE(4635)
 #  define VW_WARNING_DISABLE_CPP_17_LANG_EXT VW_WARNING_DISABLE(4984)
```
jackgerrits, line 64: It looks at least from first glance that we don't need this suppression if we are using the language level detecting macro https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-c4984?view=msvc-170

bassmang (commented):
Looks good. I agree that renaming is better than disabling the hidden member warning

vowpalwabbit/cbify.cc
olgavrou: isn't this complaining about the function parameters i.e. `const std::size_t num_actions` and `std::size_t increment`? if so can't we just rename those to `num_actions_` and `increment_`?

zwd-ms: done, thanks

explore/future_compat.h
zwd-ms: Removed.

bassmang (approved):

PR #3561, opened 2022-01-05
fix: resolve segfaults in test_search

bassmang:


vowpalwabbit/search.cc
```diff
@@ -2555,6 +2557,8 @@ base_learner* setup(VW::setup_base_i& stack_builder)
   new_options.add(make_option("search_task", task_string)
                       .keep()
                       .necessary()
+                      .one_of({"sequence", "sequencespan", "sequence_ctg", "argmax", "sequence_demoldf",
```
jackgerrits, line 2560: Based on the help string shouldnt "list" be one of these?

bassmang: yes good catch!

jackgerrits (approved):

PR #3563, opened 2022-01-05
feat: Add epsilon_reduction_features for decay of exploration

bassmang:
Gives ability to update epsilon value of specific examples from upstream reductions. Can be used in decay of exploration algorithm to dynamically set epsilon.


vowpalwabbit/epsilon_reduction_features.h
```diff
@@ -0,0 +1,23 @@
...
+{
+namespace cb_explore_adf
+{
+namespace greedy
+{
+struct reduction_features
+{
+  float epsilon;
+
+  reduction_features() { epsilon = -1.f; }
```
rajan-chari, line 17: Wondering if -1.f should be a constant (INVALID_EPSILON ?) and then reference that? I realize this is not standard practice in VW :).

rajan-chari: Another idea might be to add a is_valid_epslion() method to reduction features.  This way no one needs to know about it.

bassmang: Yea good point, I added in is_valid_epsilon()

vowpalwabbit/cb_explore_adf_greedy.cc
```diff
@@ -50,19 +50,21 @@ cb_explore_adf_greedy::cb_explore_adf_greedy(float epsilon, bool first_only)
 void cb_explore_adf_greedy::update_example_prediction(multi_ex& examples)
 {
   ACTION_SCORE::action_scores& preds = examples[0]->pred.a_s;
-
   uint32_t num_actions = static_cast<uint32_t>(preds.size());
 
+  auto& ep_fts = examples[0]->_reduction_features.template get<reduction_features>();
```
rajan-chari, line 55: Hi,  A couple of questions...

1) Wondering how persistence would work for epsilon.  What happens during prediction for a stack that has non-stationary exploration reduction?

2) Wondering if it makes sense to grab epsilon from example, set _epsilon and let everything work as is?

Thoughts?

bassmang: 1. That should depend on how the non-stationary reduction manages / updates epsilon. If the epsilon is set in reduction_features for an example it will always use that epsilon. 
2. Yes that would work too. I chose to not do that to keep the internal state of `cb_explore_adf_greedy` separate but it should be the same

bassmang: 1. That should depend on how the non-stationary reduction manages / updates epsilon. If the epsilon is set in reduction_features for an example it will always use that epsilon. 
2. Yes that would work too. I chose to not do that to keep the internal state of `cb_explore_adf_greedy` separate but it should be the same

rajan-chari (approved):

PR #3565, opened 2022-01-06
fix: Stop search_generate from segfaulting

### LINKED ISSUE #3133: Search examples under library fail to run
Both test_search and search_generate segfault when running

bassmang:
Resolves #3133


library/search_generate.cc
```diff
@@ -476,11 +476,12 @@ void train()
...
-    run_istream(gen, "phrase-table.tr", true);
-    run_istream(gen, "phrase-table.tr", false, 300000);
-    run_istream(gen, "phrase-table.te", false, 100000);
+    //run_istream(gen, "phrase-table.tr", true);
+    //run_istream(gen, "phrase-table.tr", false, 300000);
+    //run_istream(gen, "phrase-table.te", false, 100000);
+    run_easy();
   }
-  VW::finish(vw_obj);
+  VW::finish(vw_obj, false);
```
jackgerrits, line 484: What calls delete now?

bassmang: It wasn't being deleted. Just updated it to delete and commented out this line `Generator gen(vw_obj, nullptr); // &dict);` which was causing a memory freeing error

jackgerrits: But wont calling finish with false mean it wont be deleted?
```
void finish(VW::workspace& all, bool delete_all)
```

bassmang: Yes I removed `false`

jackgerrits (approved):

PR #3569, opened 2022-01-06
docs: add default values to some options

jackgerrits:


vowpalwabbit/parse_args.cc
```diff
@@ -1299,8 +1306,9 @@ VW::workspace& parse_args(
 
     option_group_definition update_args("Update");
     update_args.add(make_option("learning_rate", all.eta).help("Set learning rate").short_name("l"))
-        .add(make_option("power_t", all.power_t).help("T power value"))
+        .add(make_option("power_t", all.power_t).default_value(0.5f).help("T power value"))
```
olgavrou, line 1309: these defaults were being set at some later point right?

jackgerrits: They were defaults in global_data.cc

olgavrou (approved):

PR #3572, opened 2022-01-06
docs: help string updates

jackgerrits:
- Add [reduction] for reduction option groups
- Expand column width 
- Consistent width for each group
- Add help message for necessary args


bassmang (approved):

vowpalwabbit/options_boost_po.cc
```diff
@@ -71,10 +71,12 @@ void options_boost_po::internal_add_and_parse(const option_group_definition& gro
 {
   m_option_group_dic[m_current_reduction_tint].push_back(group);
 
-  po::options_description new_options(group.m_name);
+  po::options_description new_options(group.m_name, 100);
```
olgavrou, line 74: what are these numbers? can we name them?

jackgerrits: They are widths in characters, I chose them arbitrarily based on what looks better

PR #3576, opened 2022-01-10
Add Linux AArch64 wheel build support

odidev:
Added linux aarch64 wheel build support.
Related to https://github.com/VowpalWabbit/vowpal_wabbit/issues/3547, @jackgerrits Could you please review this PR?


jackgerrits:
Is there are a way to make the test step work for ARM wheels?

jackgerrits:
Is there are a way to make the test step work for ARM wheels? Just installing and running pytest would be enough for this I'd sat

jackgerrits:
It's quite slow to build. If there's no way to speed it up, running for every PR is not really practical

odidev:
> It's quite slow to build. If there's no way to speed it up, running for every PR is not really practical

The slow builds for aarch64 are due to qemu emulation. We can do either of the following: 

1. Add a condition on aarch64 jobs to run only during release. 
2. Use another CI, such as CircleCI as it provides AArch64 build agents and native build would be faster.

jackgerrits:
Would there be a way to make the test step work via qemu or not really? Running this on a nightly basis seems like a reasonable step to ensure coverage but now bog down PRs

odidev:
I tried running tests for aarch64 wheels but I’m getting some assertion errors which might be due to different architecture causing floating-point error. Logs for the same are attached below: 
https://github.com/odidev/vowpal_wabbit/runs/4786107263?check_suite_focus=true 

Could you please share your feedback on the same?

jackgerrits:
Yep, looking at the output they all seems like floating point errors. They seem fine. We can just relax the tolerance a bit and it should be good.

I think we should go with a daily scheduled build at first (https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#schedule). Then at least there is coverage but it doesn't bog down PRs. It will take much longer to look into CircleCI

odidev:
> Yep, looking at the output they all seems like floating point errors. They seem fine. We can just relax the tolerance a bit and it should be good.
> 
> I think we should go with a daily scheduled build at first (https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#schedule). Then at least there is coverage but it doesn't bog down PRs. It will take much longer to look into CircleCI

Done

jackgerrits:
The commit needs to have a "type:" (https://www.conventionalcommits.org/en/v1.0.0/) `build` or `ci` seems appropriate

.github/workflows/build_python_wheels.yml
```diff
@@ -8,6 +8,8 @@ on:
   pull_request:
     branches:
       - '*'
```
jackgerrits, line 10: This will still run per PR. I think the ARM specific stuff should be moved into its own workflow that only runs as cron

odidev: Done

jackgerrits (approved):

PR #3581, opened 2022-01-10
feat: Use an Enum instead of integers for label and prediction types in Python

jackgerrits:


python/pylibvw.cc
```diff
@@ -370,6 +373,10 @@ label_parser* get_label_parser(VW::workspace* all, size_t labelType)
       return &VW::slates::slates_label_parser;
     case lCONTINUOUS:
       return &VW::cb_continuous::the_label_parser;
+    case lCONTEXTUAL_BANDIT_EVAL:
+      return &CB_EVAL::cb_eval;
+    case lMULTILABEL:
+      return &MULTILABEL::multilabel;
```
bassmang, line 379: Is no_label handled?

jackgerrits: It's not. The label parser for that isn't really useful because it doesnt actually have a label in my opinion. It's still not exactly clear to me the best way to handle no_label so I decided to not try and tackle that in this PR

bassmang: Yea I agree no_label_parser doesnt make much sense (especially cache_label / read_cached_label), something to think about in the future.

bassmang (approved):

PR #3583, opened 2022-01-10
revert: "feat: [py] add git commit hash to library version (#3446)"

jackgerrits:
This reverts commit 3994a3faccfce778acbee2f19b71e65499de5651.

Seeing behavior where pip installs many packages of the same version with different +metadata. Reverting out of caution before release.


lalo (approved):

setup.py
```diff
@@ -197,15 +196,6 @@ def build(self):
 with open(config_path, encoding='utf-8') as f:
     version = f.readline().strip()
 
-try:
```
lalo, line 200: if you could enable this expanded version when self.debug is true that would be helpful for my use case

jackgerrits:
This is concerning. https://stackoverflow.com/questions/45745884/pip-install-semantic-version-with-metadata-not-upgrading Maybe it is best to test on PyPi test?

jackgerrits:
Spoke with @lalo, and we are going to revert this for now

PR #3584, opened 2022-01-10
chore: trim commit ref from version in docs generation

jackgerrits:


bassmang (approved):

jackgerrits:
No longer needed since the commit hash was removed from the version

PR #3585, opened 2022-01-11
fix: vwdll unit test memory leak

### LINKED ISSUE #3559: vw-unit-test.out vw_dll_parse_escaped memory errors in Windows
#### Describe the bug
Memory errors are thrown when running test `vw_dll_parse_escaped` with vw-unit-test.out on Windows. The memory errors occur from line 109 of vwdll_test.cc
`  BOOST_CHECK_THROW(VW_InitializeA("-d test\\ --nonexistent_option --quiet"), VW::vw_unrecognised_option_exception);
`

#### Observed Behavior
Detected memory leaks!
Dumping objects ->
{47717730} normal block at 0x000002224D038150, 48 bytes long.
 Data: <-d test\ --nonex> 2D 64 20 74 65 73 74 5C 20 2D 2D 6E 6F 6E 65 78
{47717729} normal block at 0x000002224E45DF50, 16 bytes long.
 Data: <X  l8           > 58 CD B9 6C 38 00 00 00 00 00 00 00 00 00 00 00
Object dump complete.

#### Environment
Master branch
Visual Studio 2019, x64 Debug mode, Boost 1.75.0

bassmang:
Resolves #3559


vowpalwabbit/vwdll.cpp
```diff
@@ -59,23 +59,21 @@ VW_DLL_PUBLIC VW_HANDLE VW_CALLING_CONV VW_InitializeEscaped(const char16_t* pst
 
 
 VW_DLL_PUBLIC VW_HANDLE VW_CALLING_CONV VW_InitializeA(const char * pstrArgs)
-{ string s(pstrArgs);
-  VW::workspace* all = VW::initialize(s);
+{
+  VW::workspace* all = VW::initialize(static_cast<std::string>(pstrArgs));
```
jackgerrits, line 63: I really don't think that cast is safe. What about constructing a local string was an issue before?

bassmang: The local string wasn't being freed an unrecognized option exception was thrown. Wrapping the logic here in a `try...catch` also works to solve the issue

jackgerrits: It doesn't really make sense why the string would not get cleaned up. The exception would just cause the stack to unwind right? What else could be going on here?

jackgerrits (approved):

PR #3586, opened 2022-01-11
feat: Separate scored_config from automl

bassmang:
Useful as separate component for decay of exploration


vowpalwabbit/automl.h
```diff
@@ -37,24 +38,13 @@ void fail_if_enabled(VW::workspace&, const std::set<std::string>&);
 constexpr uint64_t MAX_CONFIGS = 10;
 constexpr uint64_t CONFIGS_PER_CHAMP_CHANGE = 5;
 
-struct scored_config
+struct aml_score : VW::scored_config
```
jackgerrits, line 41: The objects in this header should be in the `VW` namespace

jackgerrits (approved):

PR #3599, opened 2022-01-12
feat: Update label class names, deprecate old names, add strong typing

bassmang:


python/vowpalwabbit/pyvw.py
```diff
@@ -1015,24 +1015,30 @@ def push_features(self, ns, featureList):
...
 
-class abstract_label:
+class AbstractLabel:
     """An abstract class for a VW label."""
 
     def __init__(self):
         pass
 
-    def from_example(self, ex):
+    def from_example(self, ex: pylibvw.example):
```
jackgerrits, line 1024: Dont we want this to be the Python type? `Example`?

python/vowpalwabbit/pyvw.py
```diff
@@ -1080,18 +1091,22 @@ def __str__(self):
...
+        AbstractLabel.__init__(self)
         if isinstance(label, example):
             self.from_example(label)
         else:
             self.prediction = prediction
 
-    def from_example(self, ex):
-        self.prediction = get_prediction(ex, PredictionType.MULTICLASSPROBS)
+    def from_example(self, ex: pylibvw.example):
+        self.prediction = get_prediction(ex, pylibvw.vw.pMULTICLASSPROBS)
```
jackgerrits, line 1109: Why remove the enum?

```suggestion
        self.prediction = get_prediction(ex, PredictionType.MULTICLASSPROBS)
```

bassmang: whoops typo from merging

python/vowpalwabbit/pyvw.py
bassmang: The issue here is that `example` is defined after `SimpleLabel`, and `example` uses `SimpleLabel` as a default value so it has to be defined after

jackgerrits: This forward declaration problem has been solved in Python 3.7 (https://www.python.org/dev/peps/pep-0563/#enabling-the-future-behavior-in-python-3-7) using `from __future__ import annotations` to delay evaluation of annotations. We are still supporting Python3.6 though. Python 3.6 reached end of support as of December 2021, so perhaps we can use this soon.

I think until we can use that another option is to write the annotation as a string literal (see here: https://www.python.org/dev/peps/pep-0484/#id28). Pytype still evaluates this correctly as far as I can tell and it allows for use of a later defined type

```suggestion
    def from_example(self, ex: "example"):
```

bassmang: Cool added in the string literal

jackgerrits (approved):

PR #3603, opened 2022-01-13
feat: pyvw label accessor for slates

bassmang:
Addresses #3043


python/pylibvw.cc
```diff
@@ -897,6 +902,28 @@ float ex_get_cb_continuous_pdf_value(example_ptr ec, uint32_t i)
...
+    default:
+      return tUNSET;
+  }
+}
+float ex_get_slates_weight(example_ptr ec) { return ec->l.slates.weight; }
+bool ex_get_slates_labeled(example_ptr ec) { return ec->l.slates.labeled; }
+float ex_get_slates_cost(example_ptr ec) { return ec->l.slates.cost; }
+uint32_t ex_get_slates_slot_id(example_ptr ec) { return ec->l.slates.slot_id; }
+size_t ex_get_slates_num_probabilities(example_ptr ec) { return ec->l.slates.probabilities.size(); }
+uint32_t ex_get_slates_action(example_ptr ec, uint32_t i) { return ec->l.slates.probabilities[i].action; }
```
jackgerrits, line 924: Bounds check

jackgerrits (approved):

python/vowpalwabbit/pyvw.py
```diff
@@ -1215,6 +1210,70 @@ def __str__(self):
         )
 
 
+class SlatesLabelType(IntEnum):
+    UNSET = pylibvw.vw.tUNSET
```
jackgerrits, line 1214: I like this. We should do that for the other enums too

bassmang: Yea agreed

PR #3608, opened 2022-01-14
refactor: Use cpp label enums in pyvw

bassmang:


jackgerrits (approved):

python/pylibvw.cc
```diff
@@ -1279,15 +1280,20 @@ BOOST_PYTHON_MODULE(pylibvw)
       .def_readonly("lDefault", lDEFAULT,
           "Default label type (whatever vw was initialized with) -- used as input to the example() initializer")
       .def_readonly("lBinary", lBINARY, "Binary label type -- used as input to the example() initializer")
+      .def_readonly("lSimple", lSIMPLE, "Simple label type -- used as input to the example() initializer")
       .def_readonly("lMulticlass", lMULTICLASS, "Multiclass label type -- used as input to the example() initializer")
       .def_readonly("lCostSensitive", lCOST_SENSITIVE,
           "Cost sensitive label type (for LDF!) -- used as input to the example() initializer")
       .def_readonly("lContextualBandit", lCONTEXTUAL_BANDIT,
           "Contextual bandit label type -- used as input to the example() initializer")
+      .def_readonly("lMax", lMAX, "Max label type -- used as input to the example() initializer")
```
jackgerrits, line 1289: I don't know what max is. Do you? Is there a way we can deprecate it to avoid confusion?

bassmang: Don't know what this is either -- I can put in a deprecation message

Issue #3179, opened 2021-07-20
Training cannot be continued with --regcb or --regcbopt after model reload (--save_resume turned on).


alxbar75:
#### Describe the bug
A vowpal wabbit model, initialized with --regcb or --regcbopt and --save_resume, is not improving during training after it has been reloaded.

#### To Reproduce
Example Code, 6 different contexts with 3 different actions, each combination having a probability of giving a reward (a click, cost=-1). Without saving/loading, the model converges to a good and sometimes optimal solution. With save/loading, the model is behaving differently.

To see convergence set `store_and_load = False`
```python
import operator
import random
import numpy as np
from matplotlib import pyplot as plt
from vowpalwabbit import pyvw

plt.ion()

def running_mean(x, N):
    cumsum = np.cumsum(np.insert(x, 0, 0))
    return (cumsum[N:] - cumsum[:-N]) / float(N)


def to_vw_example_format(context, cb_label=None):
    if cb_label is not None:
        chosen_action, cost, prob = cb_label
    example_string = ""
    example_string += "shared |User device={} \n".format(context)
    for action in action_space:
        if cb_label is not None and action == chosen_action:
            example_string += "1:{}:{} ".format(cost, prob)
        example_string += "|Action ad={} \n".format(action)
    # Strip the last newline
    return example_string[:-1]


# definition of problem to solve. Goal: playing out the article with highest ctr given a device
context_to_action_ctr = {
    "device-1": {"article-1": 0.05, "article-2": 0.06, "article-3": 0.04},  # 2
    "device-2": {"article-1": 0.08, "article-2": 0.07, "article-3": 0.05},  # 1
    "device-3": {"article-1": 0.01, "article-2": 0.04, "article-3": 0.09},  # 3
    "device-4": {"article-1": 0.04, "article-2": 0.04, "article-3": 0.045},  # 3
    "device-5": {"article-1": 0.09, "article-2": 0.01, "article-3": 0.07},  # 1
    "device-6": {"article-1": 0.03, "article-2": 0.09, "article-3": 0.04}  # 2
}
action_space = list(context_to_action_ctr["device-1"].keys())

vw = f"--cb_explore_adf -q UA -q UU --save_resume --regcbopt --mellowness 0.0003 --power_t 0 -l 0.5"

actor = pyvw.vw(vw)

# vw performs better with this variable set to False
store_and_load = False

random_rewards = []
actor_rewards = []
optimal_rewards = []

[...]

PR #3609, opened 2022-01-14
feat!: Persist learning_rate, power_t, and initial_t for loaded models

bassmang:
Addresses: #3179


vowpalwabbit/parse_args.cc
```diff
@@ -1381,7 +1392,6 @@ VW::workspace& parse_args(
 
     parse_diagnostics(*all.options, all);
 
-    all.initial_t = static_cast<float>(all.sd->t);
```
JohnLangford, line 1384: Why this change?

bassmang: This line has been moved to `1093` in the same file. The entire option_group_definition had to be moved to a later section in the parser, namely after
`load_header_merge_options(*all.options, all, *model, interactions_settings_duplicated);`
is called so that we have access to the loaded arguments. This line needs to be moved because it is set of the parsed arguments.

JohnLangford: Sounds good.

vowpalwabbit/parse_args.cc
```diff
@@ -1070,6 +1070,29 @@ void parse_example_tweaks(options_i& options, VW::workspace& all)
...
+               .default_value(0.5f)
+               .keep()
+               .allow_override()
+               .help("Set learning rate")
+               .short_name("l"))
+      .add(make_option("power_t", all.power_t).default_value(0.5f).keep().allow_override().help("T power value"))
+      .add(make_option("decay_learning_rate", all.eta_decay_rate)
+               .default_value(1.f)
+               .help("Set Decay factor for learning_rate between passes"))
+      .add(make_option("initial_t", all.sd->t).keep().allow_override().help("Initial t value"))
```
jackgerrits, line 1087: Why is this needed as keep. Both t and initial_t are serialized as part of a save_resume model

bassmang: I dont think being part of save_resume is sufficient here. If you save and load a model without `keep()`, the output will print `initial_t` with the default value (although internally initial_t and sd->t should restore properly)

bassmang: save_resume isn't sufficient here because it is called after the initial output printing, so if you save and load a model with some `initial_t` it will print the default value (although internally it should restore initial_t at a later point)

jackgerrits: But now dont we have duplicated/potentially conflicting initial_t information between the command line and the model?

bassmang: I removed `keep()` and `allow_override()` from `initial_t`, and moved the printing to after the model is loaded. This is actually a bug fix since `initial_t` is often altered before being save_loaded, and was previously printing out the default for loaded models.

jackgerrits (approved):

Issue #3197, opened 2021-07-30
object_pool destructor assertion error (Race condition)


bassmang:
[cb_adf_large.txt](https://github.com/VowpalWabbit/vowpal_wabbit/files/6909341/cb_adf_large.txt)

#### Describe the bug
In the destructor ~no_lock_object_pool() (object_pool.h, line 51), there is an assertion error where m_pool.size() != size(). In this case, we will get m_pool_size() == 265 and size() == 264 conditionally. When running in debug mode, this assertion will throw an error sometimes, and run well other times, suggesting a race condition.

#### To Reproduce
1) Build vw in debug mode
2) vw --cb_explore_adf -d cb_adf_large.txt -q ::
Use the attached .txt file

#### Environment
Master branch with Linux

PR #3617, opened 2022-01-18
fix: double free of end_pass example causing tests to fail randomly

olgavrou:
Resolves: #3197 

Reason:

this issue surfaces itself when the `object_pool` happens to be full during the reading of the very last example that is going to have `end_pass` set to `true`

```
          parser -> reached end of file and want to get another example to set it to end_pass=true (but object pool is full, so just wait for an example to be finished)
learner -> consume a bunch of learnable examples from ready_example queue
learner -> is newline? yes so finish newline example (still holding the ptr to the newline example that was just finished and returned to object pool)
learner -> do_learn (still holding the ptr to the newline example that was just finished and returned to object pool)
          parser -> hey an example is on the object pool, you can stop being blocked and read it (same ptr as was just finished and returned to object pool)
          parser -> set that example's end_pass = true
learner -> do_learn returns, the example who's ptr you still have, is end_pass? yes!
learner -> finish example again (this is the problematic finish example, because learner called finish on it twice!)
          parser -> dispatch end_pass=true example to learner
learner -> consume example from ready queue
learner -> is end_pass? yes! finish same example again
```


vowpalwabbit/learner.cc
```diff
@@ -186,10 +186,10 @@ class multi_example_handler
       _context.template process<multi_ex, learn_multi_ex>(ec_seq);
       ec_seq.clear();
     }
-
-    // Send out the end-of-pass notification after doing learning
-    if (ec->end_pass)
+    else if (ec->end_pass)
```
olgavrou, line 189: `else if` branch will be true when we call `on_example` again with the new `end_pass` example object after it is `pop`ed from the `ready_example` queue

vowpalwabbit/learner.cc
```diff
@@ -186,15 +175,21 @@ class multi_example_handler
...
       // Because the end_pass example is used to complete the in-flight multi_ex prior
       // to this call we should have no more in-flight multi_ex here.
       assert(ec_seq.empty());
       _context.template process<example, end_pass>(*ec);
     }
+    else if (ec->is_newline)
+    {
+      // Because the is_newline example is used to complete the in-flight multi_ex prior
+      // to this call we should have no more in-flight multi_ex here.
+      assert(ec_seq.empty());
```
rajan-chari, line 190: If there are two new lines in a row, will this get triggered?

olgavrou: yes, if it's a newline that comes right after the multi-line examples then learn will be called and then finish on the first newline. Then the second newline will not trigger a learn and will just be cleaned up

jackgerrits (approved):

PR #3623, opened 2022-01-20
refactor: change python module names to align with PEP8

### LINKED ISSUE #3127: [python] Improve modules names and import
**Module names**
The modules names could be improved: 
- `DFtoVW` is not [pep8 compliant](https://www.python.org/dev/peps/pep-0008/#package-and-module-names):
> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability.

It could be renamed `dftovw` (baring pep8 it's really rare to see capitals in module name).
- `sklearn_vw` could just be renamed `sklearn`: the "_vw" doesn't bring much information as one already knows he's using vowpalwabbit package.

A simple warning could be issued so it doesn't break the backward compatibility.

**Import**
This might be more a question of taste by why not import directly `vw` from the _root_ of the package (i.e from `vowpalwabbit`), since it's the main class that people are expected to use in the package ? To be more clear, in `__init__.py` you'd import vw from .pyvw so you could then do `from vowpalwabbit import vw`.

jackgerrits:
Resolves #3127


python/vowpalwabbit/__init__.py
```diff
@@ -1,16 +1,74 @@
 # -*- coding: utf-8 -*-
 """Python interfaces for VW"""
 
-import importlib
+import importlib as _importlib
+import warnings as _warnings
```
bassmang, line 5: Why change the name?

python/vowpalwabbit/DFtoVW.py
```diff
@@ -2,6 +2,8 @@
 import numpy as np
 import pandas as pd
 
+print(__name__)
```
bassmang, line 5: Assuming this was for testing

python/vowpalwabbit/__init__.py
jackgerrits: Since they aren't exports of the package itself I think _ means it is hidden?

python/vowpalwabbit/DFtoVW.py
jackgerrits: yep thanks for catching!

olgavrou (approved):

bassmang (approved):

etiennekintzler:
great @jackgerrits  ! much cleaner and coherent

PR #3629, opened 2022-01-20
feat: get_label and get_prediction work the same way, more consistent usage of enums

jackgerrits:


python/pylibvw.cc
```diff
@@ -416,6 +416,10 @@ size_t my_get_label_type(VW::workspace* all)
   {
     return lCONTINUOUS;
   }
+  else if (lp->parse_label == MULTILABEL::multilabel.parse_label)
+  {
+    return lMULTILABEL;
+  }
```
bassmang, line 422: we also need cb_eval here. It may also be better to use lSIMPLE here to phase out lBINARY

python/vowpalwabbit/pyvw.py
```diff
@@ -338,25 +328,40 @@ def get_prediction(ec, prediction_type: Union[int, PredictionType]):
...
+    label_type: LabelType,
+) -> Union[
+    "SimpleLabel",
+    "MulticlassLabel",
+    "CostSensitiveLabel",
+    "CBLabel",
+    "CCBLabel",
+    "SlatesLabel",
+    "CBContinuousLabel",
+]:
```
bassmang, line 348: Can we use `AbstracLabel` here instead?

python/vowpalwabbit/pyvw.py
```diff
@@ -709,9 +707,9 @@ def predict(self, ec, prediction_type: Optional[Union[int, PredictionType]] = No
             prediction_type = self.get_prediction_type()
 
         if isinstance(ec, Example):
-            prediction = get_prediction(ec, prediction_type)
+            prediction = ec.get_prediction(prediction_type)
         else:
-            prediction = get_prediction(ec[0], prediction_type)
+            prediction = ec[0].get_prediction(prediction_type)
```
bassmang, line 712: nit: `elif isinstance(ec, List)` to make it clear what this is looking for

python/vowpalwabbit/pyvw.py
```diff
@@ -1847,18 +1851,121 @@ def iter_features(self):
...
+        self, label_class: Optional[Union[int, LabelType, AbstractLabel]] = None
+    ) -> Union[
+        "SimpleLabel",
+        "MulticlassLabel",
+        "CostSensitiveLabel",
+        "CBLabel",
+        "CCBLabel",
+        "SlatesLabel",
+        "CBContinuousLabel",
+    ]:
```
bassmang, line 1864: Also would be nice if `AbstracLabel` can be used here

python/vowpalwabbit/pyvw.py
```diff
@@ -1847,18 +1851,121 @@ def iter_features(self):
...
+    def get_prediction(
+        self, prediction_type: Optional[Union[int, PredictionType]] = None
+    ) -> Any:
+        """Get prediction object from this example.
+
+        Parameters
+        ----------
+
+        prediction_type :
+            - None should be used unless you know what you are doing.
```
bassmang, line 1916: Same as above

python/vowpalwabbit/pyvw.py
```diff
@@ -1847,18 +1851,121 @@ def iter_features(self):
...
+        """Get the label object of this example.
 
         Parameters
         ----------
 
-        label_class : label classes
-            Get the label of the example of label_class type, by default is
-            Simplelabel
+        label_class :
+            - None should be used unless you know what you are doing.
```
bassmang, line 1871: Message seems a bit intimidating. Maybe combine with last comment.

python/vowpalwabbit/pyvw.py
jackgerrits: Ah I just realized this is wrong anyway. It should be Type[AbstractLabel] but I don't think we can do that until 3.9. I'll leave this one unannotated

python/pylibvw.cc
jackgerrits: We could also potentially move to use he output label type of the top of the reduction stack

python/vowpalwabbit/pyvw.py
jackgerrits: If an AbstractLabel is returned then the specific functionality of each type is not accessible

python/pylibvw.cc
bassmang: might be risky but that would be cool

python/vowpalwabbit/pyvw.py
jackgerrits: Actually it is available

python/vowpalwabbit/pyvw.py
jackgerrits: On line 692 the type has already been checked to be example or list. Adding another type check here seems redundant

python/pylibvw.cc
jackgerrits: Yeah lets not for now -  but we can work towards it later

bassmang (approved):

PR #3632, opened 2022-01-20
feat: add cb_eval label accessor and fix CBLabel

bassmang:


python/vowpalwabbit/__init__.py
```diff
@@ -8,6 +8,7 @@
     "ActionScore",
     "CBContinuousLabel",
     "CBContinuousLabelElement",
+    "CBEvalLabel",
```
jackgerrits, line 11: This also needs to be added to the import below

python/tests/test_pyvw.py
```diff
@@ -148,6 +149,19 @@ def test_CBLabel():
     del model
 
 
+def test_CBEvalLabel():
+    model = Workspace(cb=4, eval=True, quiet=True)
+    cbel = pyvw.CBEvalLabel.from_example(model.example("3 1:10:0.5 |"))
```
jackgerrits, line 154: nit: prefer the `vowpalwabbit.CBEvalLabel` import

python/vowpalwabbit/pyvw.py
```diff
@@ -1249,14 +1249,55 @@ def from_example(ex: "Example"):
...
         )
 
 
+class CBEvalLabel(AbstractLabel):
+    """Class for contextual bandits eval VW label"""
+
+    def __init__(
+        self,
+        action: int = 0,
+        cb_label: CBLabel = CBLabel(),
```
jackgerrits, line 1266: I kind of don't think all of the labels should have these defaults. When would a user actually need to create a label is some default state?

python/vowpalwabbit/pyvw.py
```diff
@@ -1249,14 +1249,55 @@ def from_example(ex: "Example"):
...
+    ):
+        AbstractLabel.__init__(self)
+        self.action = action
+        self.cb_label = cb_label
+
+    @staticmethod
+    def from_example(ex: "example"):
```
jackgerrits, line 1273:
```suggestion
    def from_example(ex: "Example"):
```

python/vowpalwabbit/pyvw.py
```diff
@@ -1249,14 +1249,55 @@ def from_example(ex: "Example"):
...
+        return (
+            str(self.action)
+            + " "
+            + " ".join(
+                [
+                    "{}:{}:{}".format(c.action, c.cost, c.probability)
+                    for c in self.cb_label.costs
+                ]
+            )
+        )
```
jackgerrits, line 1298: We can use fstrings now so its possible make this a little cleaner

python/vowpalwabbit/pyvw.py
```diff
@@ -1457,7 +1498,7 @@ def __init__(
             it would be from a VW data file into an example (and
             "setup_example" is run). if it is a dict, then we add all
             features in that dictionary. finally, if it's a function,
-            we (repeatedly) execute it fn() until it's not a function
+        we (repeatedly) execute it fn() until it's not a function
```
jackgerrits, line 1501: looks to be a mistake

python/vowpalwabbit/pyvw.py
bassmang: You think it would be better to make it Optional?

jackgerrits: Why optional? I think the values should be required without defaults

bassmang: I set action to 0 to be consistent with C++ but I agree that isnt a good default

jackgerrits: Since the primary way of creation in python is from an example object we don't need the defaults for anything in particular. I was just thinking as a way to make it harder to create a potentially nonsensical default object. This is really not a big deal though

python/vowpalwabbit/pyvw.py
```diff
@@ -1262,8 +1262,8 @@ class CBEvalLabel(AbstractLabel):
 
     def __init__(
         self,
-        action: int = 0,
-        cb_label: CBLabel = CBLabel(),
+        action: Optional[int] = None,
+        cb_label: Optional[CBLabel] = None,
     ):
```
jackgerrits, line 1267: I'd really rather not have optional fields. If youre using this type they need to be set

python/vowpalwabbit/pyvw.py
bassmang: Makes sense I got rid of the default

python/vowpalwabbit/pyvw.py
bassmang: removed

jackgerrits (approved):

PR #3633, opened 2022-01-21
feat: pyvw label accessors for multilabel

bassmang:


python/pylibvw.cc
```diff
@@ -1008,6 +1008,13 @@ py::object ex_get_ccb_explicitly_included_actions(example_ptr ec)
   return varray_to_pylist(label.explicit_included_actions);
 }
 
+size_t ex_get_multilabel_num_labels(example_ptr ec) { return ec->l.multilabels.label_v.size(); }
```
jackgerrits, line 1011: It might be simpler to just return a py::list for this one?

bassmang: Sure that makes it a bit simpler

python/vowpalwabbit/pyvw.py
```diff
@@ -1468,6 +1468,31 @@ def __str__(self):
         )
 
 
+class CBContinuousLabelElement:
```
jackgerrits, line 1471: Looks like this was added by mistake?

bassmang: good catch!

jackgerrits (approved):

PR #3639, opened 2022-01-24
feat: add types to vowpalwabbit.dftovw

jackgerrits:
- Add types for all functions
- Convert docstrings to Google format as it plays nicer with automatic docstring type insertion (coming in separate PR)
    - This means the types will be in the docs automatically from the type hints
- `AttributeDescriptor` was is complex. I opted to represent it with `Any` for now as it is mostly internal and not user facing
- `Hashable` is used a fair bit for column names. Is this right?

CC: @etiennekintzler

### DIFF
--- a/python/vowpalwabbit/dftovw.py
+++ b/python/vowpalwabbit/dftovw.py
@@ -1,6 +1,7 @@
 import warnings
 import numpy as np
 import pandas as pd
+from typing import Any, Hashable, List, Optional, Tuple, Type, Union
 
 
 class _Col:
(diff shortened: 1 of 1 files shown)


bassmang (approved):

python/vowpalwabbit/dftovw.py
```diff
@@ -193,16 +176,13 @@ def __init__(self, attribute_name, expected_type, min_value=None, max_value=None
         self.min_value = min_value
         self.max_value = max_value
 
-    def __set__(self, instance, arg):
+    def __set__(self, instance: "AttributeDescriptor", arg: str) -> None:
```
etiennekintzler, line 179: Shouldn't it be 
```python
instance: Any
```

python/vowpalwabbit/dftovw.py
```diff
@@ -634,6 +556,9 @@ def __init__(self, label_list):
 
         # Unpack columns
         columns = set()
+        # Override the type to Any as labels have complex machinery that is not yet communicated via its type.
+        label_list: List[Any] = label_list
```
etiennekintzler, line 560: not sure why it is necessary ?

python/vowpalwabbit/dftovw.py
```diff
@@ -699,64 +624,62 @@ class DFtoVW:
...
+            Union[
+                SimpleLabel,
+                MulticlassLabel,
+                MultiLabel,
+                ContextualbanditLabel,
+                List[MultiLabel],
+                List[ContextualbanditLabel],
+            ]
+        ] = None,
+        tag: Optional[str] = None,
```
etiennekintzler, line 640: I think tag could be any `Hashable`

etiennekintzler:
I am late to the party but well done @jackgerrits for this meticulous work.

Also you were right on replacing `str` to `Hashable` on a lot of labels (the docstring was indicating `str` which was somewhat incorrect)

> AttributeDescriptor was is complex. I opted to represent it with Any for now as it is mostly internal and not user facing

Yes, also I would make this class private (add `_`) as it's not user facing.

python/vowpalwabbit/dftovw.py
jackgerrits: This is needed because functionality exposed through AttributeDescriptors extensions are used just below and the ContextualbanditLabel and MultiLabel types do not expose it. So its a type error with those types as is.

PR #3644, opened 2022-01-25
fix: fix type and deprecate legacy example construction method

jackgerrits:


python/vowpalwabbit/pyvw.py
```diff
@@ -1403,16 +1403,20 @@ def __init__(
                 If initString is a string, then this string is parsed as
                 it would be from a VW data file into an example (and
                 "setup_example" is run). if it is a dict, then we add all
-                features in that dictionary. finally, if it's a function,
+                features in that dictionary. Finally, if it's a function,
                 we (repeatedly) execute it fn() until it's not a function
-                any more(for lazy feature computation). By default is None
+                any more(for lazy feature computation). Passing a function is deprecated. By default is None
```
lalo, line 1408: why are we removing this?

jackgerrits: This being a potentially infinitely called function is (as far as I can tell) impossible to type and it is not clear if and when it is being used. If its sole existence is to support lazy evaluation I think it should be limited to a single call at the very least. Besides that, I don't really think it is the responsibility of the library to support lazy evaluation of parameter (slippery slope: why not all params). Since it is Python if user required a lazy evaluated parameter for this call it should be quite simple to create a lazily evaluated wrapper

lalo (approved):

PR #3646, opened 2022-01-25
fix: support sklearn_vw name

jackgerrits:


jackgerrits:
The global import is necessary for compatibility

bassmang (approved):

PR #3654, opened 2022-01-27
refactor: store l1 and l2 as model state and not command line arguments

jackgerrits:
This change is important as these options being serialized as a float or double matters as enough precision is lost to affect save_resume model in the float case. Moving this to the model resolves the todo, makes it consistent with all other state and removes the need for options to understand double precision.

This change automatically supports "upgrading" models. If an old keep style model is loaded it will update the state accordingly and on subsequent save/load will use the save load data section

When loading value is
- If command line value is non-default, use that value
- Else if model is non-default, use that value
- Else use default

Must merge after #3656


vowpalwabbit/vw_versions.h
```diff
@@ -59,5 +59,9 @@ constexpr VW::version_struct VERSION_PASS_UINT64{8, 3, 3};
 
 /// Added serialized seen min and max labels in the --active reduction
 constexpr VW::version_struct VERSION_FILE_WITH_ACTIVE_SEEN_LABELS{9, 0, 0};
+
+/// Moved option values from command line to model data
+constexpr VW::version_struct VERSION_FILE_WITH_L1_AND_L2_STATE_IN_MODEL_DATA{8, 11, 0};
```
jackgerrits, line 64: Decide if okay for 9 or later and update accordingly

test/train-sets/ref/automl_readable.txt
```diff
@@ -1,5 +1,5 @@
 Version 8.11.0
```
rajan-chari, line 1: Will this have the right version once VW 9 is released?

jackgerrits: When version.txt changes then these files are updated

vowpalwabbit/vw_versions.h
lokitoth: We should make sure we decide this before merging the PR, since it could be a little awkward to have to change this once it is in the repo, and it could cause odd situations where a source-build does the wrong thing.

jackgerrits: The source build should do the right thing according to version.txt. It is just whether or not this functionality is "turned on"

vowpalwabbit/gd.cc
```diff
@@ -1211,12 +1230,10 @@ base_learner* setup(VW::setup_base_i& stack_builder)
                .default_value(0.f)
                .help("Degree of l2 regularization applied to activated sparse parameters"))
       .add(make_option("l1_state", all.sd->gravity)
-               .keep(all.save_resume)
-               .default_value(0.)
+               .default_value(L1_STATE_DEFAULT)
```
rajan-chari, line 1233: If the model file does not have l1,l2 (i.e. older model), and we don't pick up l1, l2 values from command line will this be breaking change?

jackgerrits: No, if neither the model file or command line contains l1/l2 information then the default value will be used as before. There is no change in behavior

jackgerrits:
Test 1 in slow.vwtest.json handles saving and loading models with --l1 and --l2 which causes l1_state and l2_state to be modified. (https://github.com/VowpalWabbit/vowpal_wabbit/blob/2f189ca300ddef39196d17bc9358a74fc5d539ef/test/save_resume_test.py#L190) This is how I found the issue in the first place.

I can add more tests though. I'll test the override behavior.

vowpalwabbit/vw_versions.h
jackgerrits: Will update this to 9.0.0 prior to merging

bassmang (approved):

PR #3655, opened 2022-01-27
fix: report regcb predictions as probabilities

bassmang:
Resolves: #3179


jackgerrits (approved):

test/train-sets/ref/regcb_save.stderr
```diff
@@ -1,5 +1,6 @@
 creating quadratic features for pairs: UA
 final_regressor = models/regcb_ld.model
+predictions = regcb_pred.predict
```
ataymano, line 3: is this file missing?

ataymano (approved):

PR #3658, opened 2022-01-28
fix: only persist learning_rate and power_t for save_resume -- also dont double parse update_options

bassmang:


jackgerrits (approved):

vowpalwabbit/parse_args.cc
```diff
@@ -1532,14 +1536,14 @@ void parse_modules(options_i& options, VW::workspace& all, bool interactions_set
...
 
-  parse_update_options(options, all);
-
   parse_feature_tweaks(options, all, interactions_settings_duplicated, dictionary_namespaces);  // feature tweaks
 
   parse_example_tweaks(options, all);  // example manipulation
 
   parse_output_model(options, all);
 
+  parse_update_options(options, all);
```
jackgerrits, line 1545: Is this reorder still needed?

bassmang: Yes because `parse_output_model(options, all);` determings `all.save_resume` which is used in `parse_update_options(options, all);`

jackgerrits: Got it thanks!

PR #3661, opened 2022-01-28
Update PACKAGE.rst

bassmang:


bassmang:
Adding twine check to CI instead

PR #3663, opened 2022-01-30
feat: Assert correct labelType in from_example

jonpsy:
Related #3546 

@bassmang  We went over this in the PR and over IRC as well. I've only tested the correct error message in `test_ccb.py`. I'll assume that should prove its working? Thanks!


jackgerrits:
The labelType in the wrapped Python example is sort of like a hint right now. I am not confident it always correctly represents the help label type. Here is a counter example:
```Python
import vowpalwabbit
workspace = vowpalwabbit.Workspace(quiet=True) # Using default simple labels
print(workspace.get_label_type())
ex = workspace.parse("456 | a b", vowpalwabbit.LabelType.CONTEXTUAL_BANDIT)
print(ex.get_label(vowpalwabbit.LabelType.SIMPLE))
print(ex.get_label(vowpalwabbit.LabelType.CONTEXTUAL_BANDIT))
```

The above seems to be a bug but I think it is also indicative of the root problem that an example doesn't know for sure what label it has. Until that is true I am not sure we should check for such a constraint.

bassmang:
@jonpsy closing this PR as per the concerns above. We may want to revisit this in the future when we can make stronger guarantees on the correctness of labelTypes. Thanks for you contribution!

PR #3667, opened 2022-01-31
build: update fmt and spdlog submodules

### LINKED ISSUE #3666: update fmt and spdlog
Test and support newer fmt and spdlog.

From testing downstream in the conda-forge feedstock (also see failures in #3667), we see that the build fails when using newer fmt and spdlog, for now highest working versions are fmt = 7.x and spdlog 1.8.x.

ngam:
Fixes #3666 

fmt: 7.1.3 --> 8.1.1
spdlog: v1.8.2 --> v1.9.2

### DIFF
--- a/ext_libs/ext_libs.cmake
+++ b/ext_libs/ext_libs.cmake
@@ -2,8 +2,8 @@
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 if(FMT_SYS_DEP)
-  # fmt was initially built against 7.1.3. Its possible earlier versions will also work, but that needs to be tested
-  find_package(fmt 7.1.3 REQUIRED)
+  # fmt is now built against 8.1.1. Its possible earlier versions will also work, but that needs to be tested
+  find_package(fmt REQUIRED)
 else()
   if(VW_INSTALL)
     set(FMT_INSTALL ON CACHE BOOL "install fmt library" FORCE)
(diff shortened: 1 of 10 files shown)


ngam:
Hi @jackgerrits, I committed your PR #3669 here, and it seems it is working fine. I will take this as a patch downstream soon (not sure when you are planning to update the modules for vw). 

This may resolve some issues with the cross-compilation, but not guaranteed. We likely have to figure out a separate solution for that.

jackgerrits:
I'm going to merge the other first then we can update here as I want proof in the history of it working with the older submodule versions.

ngam:
> I'm going to merge the other first then we can update here as I want proof in the history of it working with the older submodule versions.

Yes, of course. You may want to do more thorough testing and get input from users/maintainers in case some other pieces strictly need the older versions. We can also do some more testing downstream for these sorts of things. In conda-forge, we have global pinnings where all packages are instructed to migrate for certain dependencies and so there is a high chance we see issues to do with these sorts of things earlier than here. That's why working together on this can be quite beneficial for both of all.

Once we have everything sorted, I will add some documentation about using vw from conda-forge (since this is a much safer approach than pypi, imo)

jackgerrits:
I think if the changes still work on the previous pinned version as well as latest versions then we're good here. These dependencies are quite stable.

Docs around consumption through conda are certainly welcome! :)

ngam:
We got an error with windows --- I personally have no clue whatsoever about anything to do windows, haven't used a windows devices for nearly 15 years...

jackgerrits:
Its a flaky test

jackgerrits (approved):

PR #3690, opened 2022-02-03
refactor: decouple help message generation from option parsing implementation

jackgerrits:


vowpalwabbit/main.cc
```diff
@@ -39,10 +39,12 @@ int main(int argc, char* argv[])
   driver_config.add(make_option("onethread", should_use_onethread).help("Disable parse thread"));
   driver_config.add(make_option("log_level", log_level)
                         .default_value("info")
+                        .hidden()
                         .one_of({"info", "warn", "error", "critical", "off"})
                         .help("Log level for logging messages. Specifying this wil override --quiet for log output"));
   driver_config.add(make_option("log_output", log_output_stream)
                         .default_value("stdout")
+                        .hidden()
```
bassmang, line 47: Why are we hiding these?

vowpalwabbit/cli_help_formatter.cc
```diff
@@ -0,0 +1,187 @@
...
+    };
+    void visit(typed_option<int64_t>& option) override
+    {
+      if (!option.one_of().empty()) { m_ret_val = fmt::format("{}", fmt::join(option.one_of(), ", ")); }
+    };
+    void visit(typed_option<int32_t>& option) override
+    {
+      if (!option.one_of().empty()) { m_ret_val = fmt::format("{}", fmt::join(option.one_of(), ", ")); }
+    };
+    void visit(typed_option<bool>& /*option*/) override{};
```
bassmang, line 38: is bool relevant for one_of?

vowpalwabbit/cli_help_formatter.cc
```diff
@@ -0,0 +1,187 @@
...
+    void visit(typed_option<float>& /*option*/) override{};
```
bassmang, line 39: is this meant to be different than int, uint, etc..

vowpalwabbit/cli_help_formatter.cc
```diff
@@ -0,0 +1,187 @@
...
+      }
+      std::string option_name_str = ss_option_name.str();
+
+      std::stringstream ss_description;
+      ss_description << option->m_help;
+
+      ss_description << " (type: " << type_string(*option);
+
+      if (!default_value_string(*option).empty()) { ss_description << ", default: " << default_value_string(*option); }
+      if (option->m_keep) { ss_description << ", keep"; }
```
bassmang, line 164: May want to put one_of after default. Wondering if users will know what `keep` and `necessary` mean.

vowpalwabbit/main.cc
jackgerrits: These are only here to get the value. The standard registration place of these is parse args

vowpalwabbit/cli_help_formatter.cc
jackgerrits: It is not, that's why it unhandled

vowpalwabbit/cli_help_formatter.cc
jackgerrits: One of is after default. I think keep and necessary are important pieces of info we need to a better job at explaining. They allow you to understand how the command line is interpreted and stored

vowpalwabbit/main.cc
bassmang: Ahh nice

vowpalwabbit/cli_help_formatter.cc
jackgerrits: Not sure I understand?

vowpalwabbit/cli_help_formatter.cc
bassmang: I was suggesting putting one_of after default and before `keep` and `necessary` since `type`, `default` and `one_of` are all related to what a user can input. And yea I agree keep and necessary are important, just wasn't sure it would be clear what they meant. If you release a doc later when you're done refactoring something like `keep: persist this value when saving and loading a model` would be useful

jackgerrits: Oh I see what you're saying yeah that makes sense

vowpalwabbit/cli_help_formatter.cc
bassmang: I meant this `if (!option.one_of().empty()) { m_ret_val = fmt::format("{}", fmt::join(option.one_of(), ", "));`

But realized it wouldnt make much sense to require specific float values

bassmang (approved):

PR #3694, opened 2022-02-04
feat: enable standalone consumption of boost-math by default and remove remaining boost pieces

jackgerrits:
- Flesh out standalone CI
- Add boost-math vendor dep
- Remove all references to boost program options
- Rewrite the few tools using boost program options to use VW options CLI
- Remove usages of BOOST_STATIC_ASSERT


library/recommend.cc
```diff
@@ -106,37 +106,55 @@ int main(int argc, char* argv[])
...
-    po::notify(vm);
-  }
-  catch (std::exception &e)
-  {
-    cout << endl << argv[0] << ": " << e.what() << endl << endl << desc << endl;
-    exit(2);
-  }
-
-  if (vm.count("help"))
+  // po::variables_map vm;
```
lalo, line 109: is this pending work?

jackgerrits: Should be removed

lalo (approved):

PR #3695, opened 2022-02-04
feat: support vendored builds of zlib

jackgerrits:


bassmang (approved):

ext_libs/zlib.cmake
```diff
@@ -0,0 +1,151 @@
+# This is a patched version of zlib's CMakeLists.txt.
```
olgavrou, line 1: are we intentionally checking in this file? How will it evolve with zlib versions?

jackgerrits: ZLib hasnt released in 5 years and it doesnt seem like thats going to change. This PR may be dead in the water because ZLIB edits their source code in the build process which is just a no go

jackgerrits: @lokitoth had the great idea if running the build from the binary directory!

PR #3697, opened 2022-02-04
refactor: Move options into `config/`, split files and move impls to cc

jackgerrits:


olgavrou (approved):

test/train-sets/ref/cb_bad_one_of_string.stdout
```diff
@@ -1 +1 @@
-[critical] vw (options.h:345): Error: 'bad_type' is not a valid choice for option --cb_type. Please select from {dm, dr, ips, mtr, sm}
```
bassmang, line 1: Hopefully this change won't happen so much anymore :)

vowpalwabbit/config/options.cc
```diff
@@ -77,7 +78,10 @@ bool options_i::add_parse_and_check_necessary(const option_group_definition& gro
     m_options[option->m_name] = option;
   }
 
-  return group.check_necessary_enabled(*this) && group.check_one_of();
+  auto necessary_enabled = group.check_necessary_enabled(*this);
+  if (necessary_enabled) { group.check_one_of(); }
```
bassmang, line 82: Are there not groups without any `necessary` options that we still want to check one_of, Eg most of the groups in parse_args

test/train-sets/ref/cb_bad_one_of_string.stdout
jackgerrits: Haha yes I hope so!

vowpalwabbit/config/options.cc
jackgerrits: Yeah they are checked in the other `add_and_parse` function. In the case of necessary we only check the `one_of` if the group was enabled. This is due to `cb_type` having different sets of allowed values.

bassmang (approved):

PR #3699, opened 2022-02-07
feat: Implement standalone command line options parsing

jackgerrits:
This PR adds a new implementation to the `options_i` interface. It can replace boost program options as the command line parser.

VW progressively adds options and expects correct parsing of the options it knows about so far. This parser is generally split into two phases:
1. Associate tokens with options
2. Convert option associated tokens with the required type

Phase 1 will treat any option it does not know about as positional. When it eventually does know about an option it will consume subsequent tokens based on the type of the option.

- bool: always zero tokens
- scalar: single token 
- vector: all tokens until the next option like thing is seen
  - Added complexity here is that the first token is always consumed whether it looks like an option or not

Short form options can have their value embedded in the same token as the option itself. For a given flag f the value 'value' can be supplied as `-fvalue`

Long form options can have their value embedded in the same token as the option itself. For a given option 'option' the value 'value' can be supplied as `--option=value`

These values follow the same rules as values that follow a given option as a separate token. In other words they are treated as if they were a separate option.

Positional values can only be retrieved after all option descriptions have been supplied.

Phase 1 is run again after every `add_and_parse`. In this way the understanding of the command line is improved until all information is known.

Caveat (which exists today too): if a value is a valid option and we learn about the value option before learning about the preceeding option then the value option will be extracted and used. This is a pretty weird case and the parser has always been like that so it's just something good to understand. This seems to just be a limitation of operating with ambiguity,

For example if `scalar_string` is scalar and `bool` is a bool option. If we learn about `scalar_string` before `bool` then bool is not extracted but if we learn about `scalar_string` after `bool` then `bool` will be extracted and used as well.

```
--scalar_option --bool
```


vowpalwabbit/config/options_cli.cc
```diff
@@ -0,0 +1,365 @@
...
+}
+
+bool options_cli::was_supplied(const std::string& key) const { return m_tokens.find(key) != m_tokens.end(); }
+
+void options_cli::check_unregistered(VW::io::logger& logger)
+{
+  for (auto const& supplied : m_supplied_options)
+  {
+    if (m_defined_options.count(supplied) == 0)
+    { THROW_EX(VW::vw_unrecognised_option_exception, "unrecognised option '--" << supplied << "'") }
```
bassmang, line 330: what if an unrecognized short name is provided?

jackgerrits: Good point we can detect based on length

vowpalwabbit/config/options_cli.cc
```diff
@@ -0,0 +1,365 @@
...
+    return std::equal(b.begin(), b.end(), a.begin(),
+        [](unsigned char left, unsigned char right) { return std::tolower(left) == std::tolower(right); });
+  }
+  return false;
+}
+
+bool to_bool(VW::string_view token)
+{
+  if (case_insensitive_equal(token, "true")) { return true; }
+  if (case_insensitive_equal(token, "false")) { return false; }
```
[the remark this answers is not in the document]
jackgerrits: Will be removed since bool values arent being used

vowpalwabbit/config/options_cli.cc
```diff
@@ -0,0 +1,365 @@
...
+
+namespace VW
+{
+namespace config
+{
+bool is_number(VW::string_view s)
+{
+  size_t end_idx = 0;
+  const auto f = parseFloat(s.begin(), end_idx, s.end());
+  if ((end_idx == 0 && !s.empty()) || std::isnan(f)) { return false; }
```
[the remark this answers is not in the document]
jackgerrits: this impl is changed in an uncommitted revision

jackgerrits:
TODO: there is another behavior with command line parsing.

- If `--` (terminator) is encountered while parsing all values which follow it should be interpreted as positional arguments
- If the `--` arg is consumed as a value ie `--q --` then it isn't interpreted as a terminator

vowpalwabbit/config/options_cli.cc
```diff
@@ -0,0 +1,450 @@
...
+bool is_number(const std::string& str)
+{
+  if (str.empty()) { return false; }
+
+  char* ptr;
+  std::strtof(str.c_str(), &ptr);
+  return (*ptr) == '\0';
+}
+
+bool is_number(VW::string_view str) { return is_number(std::string{str}); }
```
bassmang, line 36: Can we put the implementation here and remove the `const std::string& str` overload since it's only used as an intermediary

vowpalwabbit/config/options_cli.cc
```diff
@@ -0,0 +1,486 @@
...
+  {
+    const auto result_name = "__POSITIONAL__";
+    result[result_name].push_back(current_token);
+    command_line.pop();
+    return;
+  }
+
+  auto current_opt = current_token[1];
+
+  std::vector<VW::string_view> current_tokens;
```
bassmang, line 192: Is this for the case of `-o=XXX`? If so it would be better to check for the equal sign

vowpalwabbit/config/options_cli.cc
jackgerrits: This is actually necessary. Tokens are stored as string_views which are *not* null terminated, `strtof` requires a null terminated string. So the overload is actually creating a temporary just to facilitate this call. It's a bit gross I agree

vowpalwabbit/config/options_cli.cc
jackgerrits: You cannot use an equals sign in a short option. It is simply -oXXX

bassmang (approved):

vowpalwabbit/config/options_cli.cc
```diff
@@ -0,0 +1,486 @@
...
+  scalar,
+  boolean,
+  vector
+};
+
+option_type get_option_type(const base_option& option)
+{
+  if (option.m_type_hash == typeid(bool).hash_code()) { return option_type::boolean; }
+  if (option.m_type_hash == typeid(std::vector<std::string>).hash_code()) { return option_type::vector; }
+  return option_type::scalar;
```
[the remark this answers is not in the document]
jackgerrits: The problem with checking for scalar is it can be one of many types. typed_option has a static_assert which ensures scalar is the only other possible thing it can be at this point

vowpalwabbit/config/options_cli.cc
```diff
@@ -0,0 +1,486 @@
...
+    // If there is no known option we are going to just treat it as an unknown option.
+    const auto result_name = "__POSITIONAL__";
+    result[result_name].push_back(current_token);
+    return;
+  }
+
+  const auto& option_spec = *known_opt->second;
+  consume_tokens(option_spec, command_line, current_tokens);
+
+  // Will create vector if not exists - this is intentional.
```
olgavrou, line 170: nit: what is this comment reffering to?

vowpalwabbit/config/options_cli.cc
```diff
@@ -0,0 +1,486 @@
...
+  template <typename T>
+  void handle_typed_option(typed_option<T>& option)
+  {
+    auto tokens_it = m_tokens.find(option.m_name);
+    bool option_was_supplied = tokens_it != m_tokens.end();
+
+    // Default case
+    if (!option_was_supplied)
+    {
+      if (option.default_value_supplied()) { option.value(option.default_value(), true); }
```
olgavrou, line 254: what happens to this option if it was not supplied and there is no default?

vowpalwabbit/config/options_cli.cc
```diff
@@ -0,0 +1,486 @@
...
+  // Will create vector if not exists - this is intentional.
+  auto& result_tokens = result[option_spec.m_name];
+  for (auto& token : current_tokens) { result_tokens.push_back(token); }
+}
+
+namespace VW
+{
+namespace config
+{
+namespace details
```
olgavrou, line 223: `VW::config::details` is a bit generic? maybe `token_parser` or `token_handler` or something else for a namespace?

vowpalwabbit/config/options_cli.cc
jackgerrits: No value is written, it is essentially untouched

vowpalwabbit/config/options_cli.cc
jackgerrits: `details` is a standard namespace for implementation details that the end user should not use

olgavrou: that's good to know :)

PR #3700, opened 2022-02-07
feat: implement command line split function

jackgerrits:


vowpalwabbit/parse_primitives.cc
```diff
@@ -53,6 +53,82 @@ std::vector<std::string> escaped_tokenize(char delim, VW::string_view s, bool al
...
+bool is_escape_char(char c) { return c == '\\'; }
+
+template <typename It>
+char unescape_char(It char_to_unescape_it, It end)
+{
+  if (char_to_unescape_it == end) { THROW("unescape_char: unexpected end of string while unescaping"); }
+
+  char c = *char_to_unescape_it;
+
+  if (c == 'n') { return '\n'; }
```
bassmang, line 69: Should there be more special cases here like `\t`

vowpalwabbit/parse_primitives.cc
```diff
@@ -53,6 +53,82 @@ std::vector<std::string> escaped_tokenize(char delim, VW::string_view s, bool al
...
+      {
+        current.append(1, *begin);
+      }
+    }
+    else if (is_quote(*begin))
+    {
+      if (inside_quote && quote_char == *begin)
+      {
+        inside_quote = !inside_quote;
+        quote_char = '\0';
```
bassmang, line 108: Not sure this line is needed

vowpalwabbit/parse_primitives.cc
```diff
@@ -53,6 +53,82 @@ std::vector<std::string> escaped_tokenize(char delim, VW::string_view s, bool al
...
+      else
+      {
+        current.append(1, *begin);
+      }
+    }
+    else if (is_quote(*begin))
+    {
+      if (inside_quote && quote_char == *begin)
+      {
+        inside_quote = !inside_quote;
```
bassmang, line 107: nit: `inside_quote = false`

vowpalwabbit/parse_primitives.cc
```diff
@@ -53,6 +53,82 @@ std::vector<std::string> escaped_tokenize(char delim, VW::string_view s, bool al
...
+        quote_char = '\0';
+      }
+      else if (!inside_quote)
+      {
+        inside_quote = !inside_quote;
```
bassmang, line 112: nit: `insdie_quote = true`

vowpalwabbit/parse_primitives.cc
jackgerrits: It is. `quote_char` is used to match the corresponding quote `"` or `'`. \0 is not in that range and used as a special value to make the above conditional work. This is what allows sequences like this to work `"'"`

bassmang: Right so quote_char is only relevant when inside_quote == true. Since inside_quote is set to false here, quote_char won't be relevant again until it's reset in the next conditional.

jackgerrits: Oh you're so right. Good catch!

vowpalwabbit/parse_primitives.cc
jackgerrits: Doesn't seem super necessary but I guess it doesnt hurt to add

PR #3707, opened 2022-02-10
fix: attempt to fix flaky daemon test

jackgerrits:
- its possible startup time has changed enough that it is possible for vw to not be ready by the time the netcat command is run
- trying to connect to a non-open port results in the same error for me locally


jackgerrits:
Looks to be consistent so far

olgavrou (approved):

PR #3709, opened 2022-02-10
fix: Move automl between cb_explore_adf and cb_adf on reduction stack

bassmang:


jackgerrits:
What's the rationale behind this change?

pmineiro (approved):

jackgerrits:
You should ensure AutoML expects action_scores instead of action_probs as its input prediction type

PR #3710, opened 2022-02-10
fix: parse command lines according to quotes and escape sequences

jackgerrits:
In the past VW's command line parsing did not respect quotes and escapes. This brings us in line with what a user would expect of a command line.

CC @ankane This will be a more robust fix to the #2124


bassmang (approved):

jackgerrits:
I'm going to close this for now. I think something needs to be done here but I am not confident what the right way to do it is given windows command lines too. While it feels like a bug fix this may break some people so I want to be more careful too.

ankane:
Hey @jackgerrits, fwiw, two options that come to mind are:

1. Have a method to set options that's independent of command line args (this is typically what I've seen other ML libraries do, like [XGBoost](https://github.com/dmlc/xgboost/blob/93eebe86643aadac55a86e13c11cee697a6706f5/include/xgboost/c_api.h#L788-L790))
2. Have a method that accepts an array of strings (args), so no parsing is needed

jackgerrits:
My goal is to move the API towards an array of strings so the user can decide how they wish to parse a command line. It seems unnecessary to take this complexity on within the library. So maybe the solution here is to better expose that array of strings capability out through the binding layers instead

PR #3720, opened 2022-02-14
ci: consolidate lint CIs

jackgerrits:


bassmang (commented):
Wondering if we should apply python and C++ linting to the entire project

bassmang (approved):
Expanding scope of linting in separate PR

PR #3721, opened 2022-02-14
ci: consolidate python wheel building into 1 workflow

jackgerrits:


.github/workflows/python_wheels.yml
```diff
@@ -0,0 +1,330 @@
+name: Python
```
olgavrou, line 1: do we need a small description at the beginning of this file or do you think if someone just reads it is clear enough?

olgavrou (approved):

PR #3729, opened 2022-02-17
feat: Made it possible to pass no labels in DFtoVW.from_colnames

Ark-kun:
`DFtoVW.from_colnames` is useful, but cannot be used when creating datasets for prediction.

### DIFF
--- a/python/vowpalwabbit/dftovw.py
+++ b/python/vowpalwabbit/dftovw.py
@@ -702,10 +702,10 @@ class DFtoVW:
     @classmethod
     def from_colnames(
         cls,
-        y: Union[Hashable, List[Hashable]],
-        x: Union[Hashable, List[Hashable]],
         df: pd.DataFrame,
-        label_type: str = "simple_label",
+        x: Union[Hashable, List[Hashable]],
+        y: Optional[Union[Hashable, List[Hashable]]] = None,
+        label_type: Optional[str] = "simple_label",
     ) -> "DFtoVW":
         """Build DFtoVW instance using column names only.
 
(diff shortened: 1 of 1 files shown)


python/vowpalwabbit/dftovw.py
```diff
@@ -702,10 +702,10 @@ def __init__(
     @classmethod
     def from_colnames(
         cls,
-        y: Union[Hashable, List[Hashable]],
-        x: Union[Hashable, List[Hashable]],
         df: pd.DataFrame,
-        label_type: str = "simple_label",
+        x: Union[Hashable, List[Hashable]],
+        y: Optional[Union[Hashable, List[Hashable]]] = None,
+        label_type: Optional[str] = "simple_label",
```
jackgerrits, line 708: This change breaks existing code which uses this. While less ergonomic if you just make `y` optional and dont reorder the arguments then existing usage should be unaffected.

python/vowpalwabbit/dftovw.py
```diff
@@ -702,10 +702,10 @@ def __init__(
...
     ) -> "DFtoVW":
         """Build DFtoVW instance using column names only.
```
jackgerrits, line 711: The doc string needs to be updated to reflect this change

jackgerrits (changes requested):
Change makes sense to me. Just need to ensure that the API is not broken.

Can you please add a test to https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/python/tests/test_DFtoVW.py exercising this?

jackgerrits:
CC @etiennekintzler

python/vowpalwabbit/dftovw.py
```diff
@@ -743,25 +743,28 @@ def from_colnames(
             "multi_label": MultiLabel,
         }
 
-        if label_type not in dict_label_type:
-            raise ValueError(
-                "'label_type' should be either of the following string: {label_types}".format(
-                    label_types=repr(list(dict_label_type.keys()))[1:-1]
+        if label_type:
```
etiennekintzler, line 746: it should be:

```python
if y:
```

since we test whether y is `None`

python/vowpalwabbit/dftovw.py
etiennekintzler: I agree

etiennekintzler:
I think that this improvement is very welcomed as well :)

I would slightly reorganize the code as follow:
```python
        x = x if isinstance(x, list) else [x]
        namespaces = Namespace(features=[Feature(value=colname) for colname in x])
        if not y:
            return cls(namespaces=namespaces, label=None, df=df)
        else:
            dict_label_type = {
                 "simple_label": SimpleLabel,
                 "multiclass_label": MulticlassLabel,
                 "multi_label": MultiLabel,
            }
            # NOT SHOWN HERE: chunk of code with the label type etc.
            label = dict_label_type[label_type](y)
            return cls(namespaces=namespaces, label=None, df=df)
```

python/vowpalwabbit/dftovw.py
Ark-kun: >This change breaks existing code which uses this.

This is technically true. However it would only break the code that uses positional arguments instead of keyword arguments. I've looked at all GitHub code that uses `from_colnames` and it looks like all calls use keyword arguments.
https://github.com/search?q=from_colnames&type=code

>if you just make y optional and dont reorder the arguments

Unfortunately, Python does not allow optional parameters before required parameters.
So, I cannot make `y` optional without making `x` and `df` optional. // In the future, we should try to put "more required" parameters first.

The only other solution I can think of is pretty ugly: leave `y` required, but make it accept `None`.
What do you think?

jackgerrits: > This is technically true. However it would only break the code that uses positional arguments instead of keyword arguments. I've looked at all GitHub code that uses from_colnames and it looks like all calls use keyword arguments.

I know this seems minor and probably is low impact but we're making a conscious effort to have a more stable API so I think we should adhere to that here. 

> The only other solution I can think of is pretty ugly: leave y required, but make it accept None.
What do you think?

Sorry, when I said make it optional what I really meant was what you described. I agree it's ugly, but I can't think of another way to enable this in a non-breaking way.

jackgerrits:
Addressed in #3875

PR #3735, opened 2022-02-18
fix: several bug fixes for options_cli

jackgerrits:
- Don't check `__POSITIONAL__` for unregistered
- Fix false positive for `was_supplied`
- Compare values instead of strings for scalar non string types


vowpalwabbit/config/options_cli.cc
```diff
@@ -393,9 +413,25 @@ bool options_cli::was_supplied(const std::string& key) const
...
+  auto short_option_found = std::any_of(m_command_line.begin(), m_command_line.end(),
+      [&short_key](const std::string& arg) { return VW::starts_with(arg, short_key); });
+  if (short_option_found) { return true; }
+
+  const auto long_key = "--" + key;
+  auto long_option_found =
+      std::any_of(m_command_line.begin(), m_command_line.end(), [&long_key](const std::string& arg) {
+        // We need to check that the option starts with --key_name, but we also need to ensure that either the whole
+        // token matches or we hit an equals sign denoting the end of the option name. If we don't do this --csoaa and
+        // --csoaa_ldf would incorrectly match.
```
lalo, line 430: was this a bug fixed?

lalo (approved):

vowpalwabbit/config/options_cli.cc
jackgerrits: It's a bug fixed in the new options implementation yep

PR #3739, opened 2022-02-22
refactor: dont parse options header using boost

jackgerrits:


test/train-sets/ref/truncated_model_crash.stdout
```diff
@@ -1 +1 @@
-[critical] vw (io_buf.h:366): model load failed. Error Details: Unexpected end of file encountered., model files = model-sets/truncated_model
+[critical] vw (options_boost_po.cc:107): the argument ('-cb_adf') for option '--bit_precision' is invalid
```
jackgerrits, line 1: This file contains some junk in the command line which results in `-cb_adf` being a token. We know that command lines do not contain short options so we not interpret this as a value, for which it gets assigned to `--bit_precision` and fails. I believe in the past boost was interpreting this as a short option and moving forward based on that. 

The error message is bad but I believe it is more correct now. One thing I don't understand though is that for this particular test the checksum is calculated based on the invalid input, whereas if a scenario like this were to happen then most likely the checksum would not match the input and we would have a better error message.

vowpalwabbit/parse_args.cc
```diff
@@ -1438,93 +1438,57 @@ bool check_interaction_settings_collision(options_i& options, const std::string&
   return file_options_has_interaction;
 }
 
-void merge_options_from_header_strings(const std::vector<std::string>& strings, bool skip_interactions,
-    VW::config::options_i& options, bool& is_ccb_input_model)
+bool is_opt_long_option_like(VW::string_view token) { return token.find("--") == 0 && token.size() > 2; }
+
+// The model file contains a command line but it has much greater constraints than the user supplied command line. These
+// constraints greatly help us unambiguously process it. The command line will ONLY consist of bool switches or options
+// with a single value. However, the tricky thing here is that there is no way to disambiguate something that looks like
```
olgavrou, line 1445: this comment leaves me wondering how we handle it though

jackgerrits: I believe all things that looks like options are processed as such. So if you have a value which looks like an option it will be erroneously processed as an option.

lalo (approved):

olgavrou (approved):

PR #3740, opened 2022-02-22
refactor: Remove getRevertingWeight from loss function implementations.

zwd-ms:


vowpalwabbit/loss_functions.cc
```diff
@@ -105,19 +102,20 @@ class classic_squaredloss : public loss_function
     return 2.f * (label - prediction) * update_scale;
   }
 
-  float getRevertingWeight(shared_data* sd, float prediction, float eta_t) override
+  float getSquareGrad(float prediction, float label) override
```
jackgerrits, line 105: Why reorder these lines?

zwd-ms: This is to remove (not reorder) getRevertingWeight, which is not used anywhere in code.

jackgerrits: Ohh it just causes a cascading weird diff? Got it

jackgerrits (approved):

PR #3741, opened 2022-02-22
fix: more robust table formatting

### LINKED ISSUE #3737: Training output is misaligned
#### Describe the bug

```
0.000000 0.000000            1            1.0   0.0000   0.0000        7
0.500000 1.000000            2            2.0   1.0000   0.0000        7
0.480043 0.460086            4            4.0   0.0000   0.0805        5
23.608095 46.736146            8            8.0  16.3000   2.9737        7
12.132763 0.657431           16           16.0   0.0000   0.4977        7
11.581626 11.030488           32           32.0   2.0000   1.1618        7
8.434629 5.287633           64           64.0   1.5000   1.3844        8
5.997804 3.560979          128          128.0   8.1000   2.5878        8
5.201175 4.404545          256          256.0   0.0000   1.0806        8
5.017200 4.833226          512          512.0   0.0000   1.3598        8
```

#### Expected behavior
I expect the output to be properly aligned:

```
average   since         example        example  current  current  current
loss      last          counter         weight    label  predict features
 0.000000  0.000000            1            1.0   0.0000   0.0000        7
 0.500000  1.000000            2            2.0   1.0000   0.0000        7
 0.480043  0.460086            4            4.0   0.0000   0.0805        5
23.608095 46.736146            8            8.0  16.3000   2.9737        7
12.132763  0.657431           16           16.0   0.0000   0.4977        7
11.581626 11.030488           32           32.0   2.0000   1.1618        7
 8.434629  5.287633           64           64.0   1.5000   1.3844        8
 5.997804  3.560979          128          128.0   8.1000   2.5878        8
 5.201175  4.404545          256          256.0   0.0000   1.0806        8
 5.017200  4.833226          512          512.0   0.0000   1.3598        8
```

#### Observed Behavior

#### Environment
What version of VW did you use? `9.0.1` on `alpine`

jackgerrits:
Resolves #3737

My suggestion for reviewing is to turn off stderr files from the diff (if you just want to see code changes):
image


test/train-sets/ref/cats-pdf-train.stderr
```diff
@@ -13,14 +13,14 @@ power_t = 0.5
 Enabled reductions: ftrl-Coin Betting, scorer-identity, binary, cats_tree, get_pmf, pmf_to_pdf, cb_explore_pdf, cats_pdf
 Input label = continuous
 Output pred = pdf
-average  since         example        example  current  current  current
-loss     last          counter         weight    label  predict features
-0.657567 0.657567            1            1.0 {185.121,0.657567,6.20426e-05}                10
```
ataymano, line 18: Looks like not just reformatting, but some information loss here. Is it intentional?

jackgerrits: Yes, this is intentional. If the label or prediction overflow their column they are cut off. In this particular example the predictions were empty but that was hidden by the overflow of the old example. 

We can instead allow label/prediction to overflow to the next line(s) so no information is lost. What do you think?

ataymano: not sure actually. Can we maybe just expand columns limit a bit? with current approach seems like output for cats is always useless.
Allowing overflowing to the next line may complicate parsing of this output (although it is not a great idea anyway)

jackgerrits: In this PR I expanded label/pred columns from 8 to 14 chars. I can do more, but they are still not going to fit I think. The other issue here is that when we serialize these predictions and labels we don't reduce the decimal precision so they often have 7 or so digits after the decimal place and take up a lot of width. There is also no notion of the canonical string representation of a prediction which I feel would be handy.

ataymano: yeah that would be great. seems to be something that could be done as followup work though

ataymano (approved):

PR #3743, opened 2022-02-22
refactor: replace used options impl with options_cli

jackgerrits:
- Since opts are provided by the core lib now spanning tree was removed from the solution


cluster/spanning_tree_main.cc
```diff
@@ -20,51 +24,48 @@ int getpid() { return (int)::GetCurrentProcessId(); }
...
+
+  VW::config::options_cli opts(std::vector<std::string>(argv + 1, argv + argc));
+
+  VW::config::option_group_definition desc("Spanning Tree");
+  desc.add(VW::config::make_option("nondaemon", nondaemon).help("Run spanning tree in foreground"));
+  desc.add(VW::config::make_option("help", help).short_name('h').help("Print help message"));
+  desc.add(VW::config::make_option("port", port)
+               .short_name('p')
+               .default_value(26543)
+               .help("Port number for spanning tree to listen on"));
```
bassmang, line 48: nit: can we keep make this `.add()...add()...add()`

test/unit_test/options_boost_po_test.cc
```diff
@@ -238,20 +237,6 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(incorrect_option_type_str_for_int, T, option_types
   BOOST_CHECK_THROW(options->add_and_parse(arg_group), VW::vw_argument_invalid_value_exception);
 }
 
-BOOST_AUTO_TEST_CASE(multiple_locations_one_option)
```
bassmang, line 241: why remove this test?

jackgerrits: The new implementations doesn't really care about this situation, whereas boost in the past threw in this case im pretty sure. It's purely a developer side thing

bassmang (approved):

test/unit_test/options_boost_po_test.cc
olgavrou: is there no value in still keeping the test and not checking for throw, but just checking that it works fine?

vowpalwabbit/parse_args.cc
```diff
@@ -1802,7 +1803,8 @@ VW::workspace* initialize_with_builder(int argc, char* argv[], io_buf* model, bo
     trace_message_t trace_listener, void* trace_context, std::unique_ptr<VW::setup_base_i> learner_builder)
 {
   std::unique_ptr<options_i, options_deleter_type> options(
-      new config::options_boost_po(argc, argv), [](VW::config::options_i* ptr) { delete ptr; });
+      new config::options_cli(std::vector<std::string>(argv + 1, argv + argc)),
```
olgavrou, line 1806: can we use vw's make unique here?

jackgerrits: You can't use a deleter with `std::make_unique` or `VW::make_unique` unfortunately

PR #3746, opened 2022-02-22
fix: split apart newlines in log lines forwarded to python

jackgerrits:
If a batch of output came in one go the log received did not properly split apart each line before.

### DIFF
--- a/python/vowpalwabbit/pyvw.py
+++ b/python/vowpalwabbit/pyvw.py
@@ -353,10 +353,15 @@ class _log_forward:
         self.messages = []
 
     def log(self, msg):
-        self.current_message += msg
-        if "\n" in self.current_message:
+        if "\n" in msg:
+            components = msg.split("\n")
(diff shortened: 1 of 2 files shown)


python/vowpalwabbit/pyvw.py
```diff
@@ -353,10 +353,15 @@ def __init__(self):
...
-        self.current_message += msg
-        if "\n" in self.current_message:
+        if "\n" in msg:
+            components = msg.split("\n")
+            self.current_message += components[0]
             self.messages.append(self.current_message)
-            self.current_message = ""
+            for component in components[1:-1]:
+                self.messages.append(component)
+            self.current_message = components[-1]
```
bassmang, line 362: Seems like lines 358-362 can be compressed to
```
for component in components[:-1]:
     self.messages.append(component)
self.current_message = components[-1]
```

bassmang (approved):

PR #3748, opened 2022-02-23
feat: add pred/lbl types to cb algos

bassmang:


vowpalwabbit/reductions/cb/cb_to_cb_adf.cc
```diff
@@ -198,26 +194,25 @@ VW::LEARNER::base_learner* cb_to_cb_adf_setup(VW::setup_base_i& stack_builder)
...
     data->adf_learner = as_multiline(base->get_learner_by_name_prefix("cb_adf"));
-    pred_type = VW::prediction_type_t::multiclass;
   }
 
   auto* l = make_reduction_learner(
       std::move(data), base, predict_or_learn<true>, predict_or_learn<false>, all.get_setupfn_name(cb_to_cb_adf_setup))
-                .set_output_prediction_type(pred_type)
                 .set_input_label_type(VW::label_type_t::cb)
+                .set_output_label_type(VW::label_type_t::cb)
+                .set_input_prediction_type(base->get_output_prediction_type())   // action_scores or action_probs
```
jackgerrits, line 211: We know which one it will be based on data->explore_mode. So we should set the specific type

vowpalwabbit/reductions/cb/cbify.cc
```diff
@@ -890,10 +890,12 @@ base_learner* cbifyldf_setup(VW::setup_base_i& stack_builder)
   multi_learner* base = as_multiline(stack_builder.setup_base_learner());
   auto* l = make_reduction_learner(std::move(data), base, do_actual_learning_ldf, do_actual_predict_ldf,
       stack_builder.get_setupfn_name(cbifyldf_setup))
-                .set_output_prediction_type(VW::prediction_type_t::multiclass)
                 .set_input_label_type(VW::label_type_t::cs)
+                .set_output_label_type(VW::label_type_t::cb)
+                .set_input_prediction_type(base->get_output_prediction_type())  // action_scores or action_probs
```
jackgerrits, line 895: We know based on `do_actual_learning_ldf`, so we should set the specific type

vowpalwabbit/reductions/cb/cb_to_cb_adf.cc
```diff
@@ -194,22 +209,25 @@ VW::LEARNER::base_learner* cb_to_cb_adf_setup(VW::setup_base_i& stack_builder)
 
   // see csoaa.cc ~ line 894 / setup for csldf_setup
   all.example_parser->emptylines_separate_examples = false;
+  VW::prediction_type_t out_pred_type;
 
   if (data->explore_mode)
   {
     data->adf_learner = as_multiline(base->get_learner_by_name_prefix("cb_explore_adf_"));
+    out_pred_type = base->get_output_prediction_type();
```
jackgerrits, line 217: We're still not able to determine action_scores vs action_probs here. The `finish_example` does something and it probably depends on being either scores or probs right?

PR #3753, opened 2022-02-25
refactor: cleanup prediction formatting

jackgerrits:
- Add `VW::fmt_float` which supports specifying decimal precision
- Add `VW::to_string` for most prediction types for canonical representation
- Migrate usages to `VW::to_string`
- Move prediction debug prints into own file


test/pred-sets/ref/cats.predict
```diff
@@ -1,57 +1,57 @@
...
-15160.8,0.000160436
-17345.2,0.000160436
-14788.8,0.000160436
-14527.2,0.000160436
-16322,0.000160436
-12985.2,0.000160436
-13590.6,0.000160436
-14753.8,0.000160436
-17058.4,0.000160436
-13428,0.000160436
```
jackgerrits, line 22: This is an interesting one. It is using the fmtlib mode defined as:
> The default precision is as high as needed to represent the particular value.

test/train-sets/ref/cats_save.stderr
```diff
@@ -15,17 +15,17 @@ power_t = 0.5
...
 Output pred = action_pdf_value
-average  since         example        example        current        current  current 
-loss     last          counter         weight          label        predict features 
-0.000000 0.000000            1            1.0 {0.00120175... 0.00120175,...        6 
-0.000000 0.000000            2            2.0 {0.582045,0... 0.611352,0....        6 
-0.000000 0.000000            4            4.0 {10.3501,0,... 2.23432,0.4...        6 
-0.000000 0.000000            8            8.0 {2.40591,0,... 22.5925,0.0...        6 
+average  since         example        example        current        current  current
+loss     last          counter         weight          label        predict features
+0.000000 0.000000            1            1.0     {0,0,0.01}         0,0.01        6
```
ataymano, line 20: just an option to consider:
make formatting related to number of significant bits to avoid having pure 0 - it may be misleading in situation like this since pure 0 float somehow gives impression of variable that have never been changed

jackgerrits: Which column are you referring to? The left two I didn't change, they used fixed formatting of 7 (think) precision. The approach I went for for the other numbers is 2 decimal places of precision and any trailing zeroes removed.

ataymano: I mean situation when 0.00120175 turns into 0.
It may be useful to output it either 1.2e-3 or atleast something like "0..."

ataymano: but it is nit for sure

jackgerrits: Yeah that thing is tricky.  For 0.00120175 like you said with only two decimal places of precision it just becomes 0. However if you use significant digits then for large numbers you lose too much information such as 1234 becomes 1200 which seems like a pretty significant loss of into. The medium I landed on was to support two decimal places of precision and all digits left of the decimal.

ataymano: Yeah I see. I think it is ok to do it this way.

test/train-sets/ref/squarecb_load.stderr
```diff
@@ -10,15 +10,15 @@ power_t = 0.5
...
-0.000000 0.000000            1            1.0       4:0:0.14     0:0.911854       42 
-0.000000 0.000000            2            2.0       0:0:0.14     0:0.818539       42 
-0.000000 0.000000            4            4.0       1:0:0.14     2:0.893115       42 
-0.000000 0.000000            8            8.0       3:0:0.14     2:0.693382       42 
-0.000000 0.000000           16           16.0       3:0:0.14     2:0.693382       42 
-0.000000 0.000000           32           32.0       3:0:0.14     2:0.893115       42 
--0.56272 -1.12544           64           64.0       0:0:0.77     0:0.818539       42 
+average  since         example        example        current        current  current
+loss     last          counter         weight          label        predict features
+0.000000 0.000000            1            1.0       4:0:0.14         0:0.91       42
```
lalo, line 15: what file change fixed the missing spaces on current predict column?

jackgerrits: It was actually fixed in the original PR [here](https://github.com/VowpalWabbit/vowpal_wabbit/pull/3741/commits/9bdeb79806cc4d0dbc904e8ec74947c91829f9e5). But because we trim the diff it wasnt triggering failures.

lalo (approved):

PR #3762, opened 2022-03-02
ci: generate v141, v142 native nugets

jackgerrits:
# Native Nuget Packages

There are native nuget packages available for the core VW libraries and the main command line executable. They can be found as artifacts produced by this [CI job](https://github.com/VowpalWabbit/vowpal_wabbit/actions/workflows/build_nugets.yml).

There is a package available per toolset version and architecture. Currently supported:
- v141
- v142

Details of Nuget packages:
- 64 bit only
- Static library only (VW does not currently support dynamic linkage)
- Targeting the multi-threaded dynamically linked runtime
- Contains artifacts for debug and release mode builds
- Contains the necessary dependent static libs and headers (fmt, spdlog, zlib, rapidjson)

How to use:
1. Install the nuget to your package directory
2. Include the `.targets` file into your project

For example:

`myproj.vcxproj`:
```xml
<ImportGroup Label="ExtensionTargets">
    <Import Project="$(ProjectDir)packages\VowpalWabbitNative-v142-x64.9.0.0\build\vowpalwabbit.targets" Condition="Exists('$(ProjectDir)packages\VowpalWabbitNative-v142-x64.9.0.0\build\vowpalwabbit.targets')" />
</ImportGroup>
```

Or if you need to support both x86 and x64 in the same project both nugets can be included next to each other:
```xml
<ImportGroup Label="ExtensionTargets">
    <Import Project="$(ProjectDir)packages\VowpalWabbitNative-v142-x64.9.0.0\build\vowpalwabbit.targets" Condition="Exists('$(ProjectDir)packages\VowpalWabbitNative-v142-x64.9.0.0\build\vowpalwabbit.targets')" />
    <Import Project="$(ProjectDir)packages\VowpalWabbitNative-v142-x86.9.0.0\build\vowpalwabbit.targets" Condition="Exists('$(ProjectDir)packages\VowpalWabbitNative-v142-x86.9.0.0\build\vowpalwabbit.targets')" />
</ImportGroup>
```
## How to build locally

See the build process in the [build job](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/.github/workflows/build_nugets_.yml)


.github/workflows/build_nugets.yml
```diff
@@ -0,0 +1,98 @@
...
+          -DVW_BOOST_MATH_SYS_DEP=Off
+          -DDO_NOT_BUILD_VW_C_WRAPPER=On
+          -DVW_INSTALL=On
+      - name: Build and install
+        run: |
+          cmake --build build --config Debug -t vw_io vw vw-bin allreduce spanning_tree
+          cmake --build build --config Release -t vw_io vw vw-bin allreduce spanning_tree
+
+          # Install debug first so that release overwrites the exe and nothing else
+          cmake --install build --prefix ./nuget_staging --config Debug
```
olgavrou, line 48: so the executable we get is only release right?

olgavrou: why do we need the debug build?

olgavrou (approved):

.github/workflows/build_nugets.yml
jackgerrits: > so the executable we get is only release right?
 Correct

> why do we need the debug build?
Since VS is multi-config you can build both in the same solution. It also seems to sorta be a convention among nuget packages to supply both release and debug bits

PR #3773, opened 2022-03-04
build: remove Windows solution

jackgerrits:
Building on Windows is still supported. The recommended way of doing so going forward is to use CMake to generate a solution file and then build using that.

This checked in solution file is being removed so that there are not duplicated build systems that need to be kept in sync. This checked in solution also had some bugs around concurrency causing build failures.


lalo (approved):

lalo:
can you add in this description and/or make sure that the wiki calls out that you can use cmake to generate a solution file.

PR #3785, opened 2022-03-08
fix: issues in json comparison in run_tests.py

jackgerrits:


test/run_tests.py
```diff
@@ -475,8 +477,9 @@ def run_command_line_test(
                 continue
 
             if ref_file.endswith(".json"):
-                output_json = json.loads(output_content)
-                ref_json = json.loads(ref_content)
+                # Empty strings are falsy. Fall back to an empty dict if the file is empty
+                output_json = json.loads(output_content) if output_content else dict()
```
lalo, line 481: shouldn't we complain here about missing file?

jackgerrits: If we get here the file exists but is empty. If you do json.loads on an empty string it fails

jackgerrits:
No longer needed. #3786 brought in these fixes.

PR #3787, opened 2022-03-09
fix: check if base learner is GD when using invert_hash

jackgerrits:
Requires changes from #3786


jackgerrits:
I misunderstood, ftrl actually uses the audit code of GD to implement audit and invert hash.

PR #3796, opened 2022-03-14
feat: support online state and excluding feature names in dump json model

jackgerrits:


vowpalwabbit/global_data.cc
```diff
@@ -257,17 +260,47 @@ std::string dump_weights_to_json_weight_typed(
...
+          adaptive_value = rapidjson::kNullType;
+          normalized_value = current_weight_state[1];
+        }
+        if (parameter_holder.adaptive && parameter_holder.normalized)
+        {
+          adaptive_value = current_weight_state[1];
+          normalized_value = current_weight_state[2];
+        }
+
+        extra_state_value.AddMember("adaptive", adaptive_value, allocator);
```
ataymano, line 300: is it possible to make gd.cc responsible for adding this since this is gd-specific?

jackgerrits: Yeah I agree it would be good. I am not sure of a good way to make that happen at the moment to be honest.

ataymano: I see - looks like the fact that we have adaptive/normalized field as array_parameters members is hack as well.
Does it mean that if we try to output state with non-gd learner, we just not going to see extra output for now?

jackgerrits: The dump_json_weights feature explicitly restricts usage to GD for now. I figured it was better to restrict usage until we work out a way to better support other use cases.

ataymano (approved):

PR #3804, opened 2022-03-17
build: use centralized forward declaration header to reduce dependencies

jackgerrits:


cs/cli/vw_arguments.h
```diff
@@ -5,10 +5,13 @@
 #pragma once
 
 #include <msclr\marshal_cppstd.h>
-#include "vw.h"
-#include "config/cli_options_serializer.h"
+
 #include <algorithm>
 
+#include "config/cli_options_serializer.h"
```
peterychang, line 11: I've typically tried to include system headers after the local ones since it helps avoid order-dependency in include files. That way you'll get an error if a header is missing something

java/src/main/c++/jni_spark_vw.cc
```diff
@@ -1,13 +1,17 @@
 #include "jni_spark_vw.h"
-#include "vw_exception.h"
+
+#include <algorithm>
+#include <exception>
+
 #include "best_constant.h"
-#include "util.h"
 #include "config/cli_options_serializer.h"
+#include "config/options.h"
```
peterychang, line 8: nit: group includes in different directories separately from local ones?

test/unit_test/loss_functions_test.cc
```diff
@@ -2,10 +2,12 @@
 // individual contributors. All rights reserved. Released under a BSD (revised)
 // license as described in the file LICENSE.
 
-#include <boost/test/unit_test.hpp>
+#include "loss_functions.h"
```
peterychang, line 5: include ordering

vowpalwabbit/ccb_label.h
```diff
@@ -5,17 +5,18 @@
...
+#include "ccb_reduction_features.h"
 #include "label_parser.h"
 #include "v_array.h"
-// TODO: This header can be removed once type and explicit_included_actions are removed from the label
-#include "ccb_reduction_features.h"
-#include "fmt/format.h"
-#include "io_buf.h"
 #include "vw_string_view.h"
 
+// TODO: This header can be removed once type and explicit_included_actions are removed from the label
```
peterychang, line 17: Does this comment go with `ccb_reduction_features.h`?

cs/cli/vw_arguments.h
jackgerrits: Discussed directly. Will make this change in a separate PR.

java/src/main/c++/jni_spark_vw.cc
jackgerrits: Its still technically a local search path include though. Clang-format also treats them all as the same unit too due to being a local include so I think this would be hard to change

test/unit_test/loss_functions_test.cc
jackgerrits: Will be fixed in style change update

vowpalwabbit/ccb_label.h
jackgerrits: Looks like it. Seems that clang ignored the association. Ill update the comment

peterychang (approved):

PR #3805, opened 2022-03-17
refactor: add cb_type to output message

bassmang:


vowpalwabbit/parse_args.cc
```diff
@@ -1728,6 +1728,10 @@ std::unique_ptr<VW::workspace> initialize_internal(std::unique_ptr<options_i, op
     *(all->trace_message) << "initial_t = " << all->sd->t << endl;
     *(all->trace_message) << "power_t = " << all->power_t << endl;
     if (all->numpasses > 1) *(all->trace_message) << "decay_learning_rate = " << all->eta_decay_rate << endl;
+    if (all->options->was_supplied("cb_type"))
```
jackgerrits, line 1731: If this is not explicitly supplied in the command line or added then it wont show up. Some but not all of the reduction which use "cb_type" will add in the default into the command line. It might be worth ensuring all of them insert it into the command line if it wasnt supplied to make sure this is consistent

bassmang: Yes this adds it in for default if not specified in the command line as well

jackgerrits: I don't follow. `was_supplied` doesn't change the underlying command line

bassmang: Yes from what I can see all reductions that default cb_type if not supplied will actually insert it into the command lint

jackgerrits: CCB, regcb and squarecb dont

lalo: when it does insert it, it does the pattern of re-parsing the args to make sure to keep track of the change. Which is a weird pattern that could be fixed.

jackgerrits (approved):

PR #3814, opened 2022-03-18
refactor: Begin reduction ns cleanup and migrate to vw_fwd

jackgerrits:
- The goal here is that all code under the reductions directory must be in the `VW::reductions` namespace
- Nested namespaces are allowed


vowpalwabbit/reductions/active_cover.cc
```diff
@@ -2,10 +2,13 @@
 // individual contributors. All rights reserved. Released under a BSD (revised)
 // license as described in the file LICENSE.
 
+#include "reductions/active_cover.h"
```
lalo, line 5: reductions/ is not needed right?

lalo (approved):

PR #3815, opened 2022-03-19
refactor: upgrade to python3 for tools under utl

HollowMan6:
Python 2 was deprecated on January 1, 2020. Many up-to-date Linux distributions now use Python 3 as the default. It's time for the tools under `utl` to get upgraded to Python 3.

* Add a shebang for `active_interactor.py`
* Refactor `csv2vw`, `vw-csv2bin`, `vw-lda` for python3
* Explicitly specify `vw-hyperopt.py` to use python3 (So that we can also directly run it even on old Linux distributions who still use Python 2 as the default)

Signed-off-by: Hollow Man <hollowman@opensuse.org>

### DIFF
--- a/utl/active_interactor.py
+++ b/utl/active_interactor.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 import socket
 import argparse
 from typing import Callable, ContextManager, Optional, TextIO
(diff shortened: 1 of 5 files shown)


jackgerrits (approved):
Thanks for the PR! Looks good to me

lalo:
Related: I noticed we have a copy of vw-varinfo but a python3 version is available here: https://github.com/arielf/weight-loss/blob/master/vw-varinfo2

Issue #3782, opened 2022-03-08
ci: test failure intermittent


lalo:
`    Failed TestSharedModel [1 s]

CUSTOMBUILD : error Message:  [D:\a\vowpal_wabbit\vowpal_wabbit\vw\build\test_with_output.vcxproj]
     Test method cs_unittest.TestCbAdfClass.TestSharedModel threw exception: 
  System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
    Stack Trace:
        at VW.finish(workspace* , Boolean )
     at VW.VowpalWabbitBase.InternalDispose() in D:\a\vowpal_wabbit\vowpal_wabbit\vw\cs\cli\vw_base.cpp:line 157
     at VW.VowpalWabbitBase.~VowpalWabbitBase() in D:\a\vowpal_wabbit\vowpal_wabbit\vw\cs\cli\vw_base.cpp:line 94
     at VW.VowpalWabbitBase.Dispose(Boolean A_0) in C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\include\sstream:line 0
     at VW.VowpalWabbitBase.Dispose()
     at VW.VowpalWabbit`2.Dispose(Boolean disposing) in D:\a\vowpal_wabbit\vowpal_wabbit\vw\cs\cs\VowpalWabbit.cs:line 506
     at VW.VowpalWabbit`2.Dispose() in D:\a\vowpal_wabbit\vowpal_wabbit\vw\cs\cs\VowpalWabbit.cs:line 496
     at cs_unittest.TestCbAdfClass.TestSharedModel() in D:\a\vowpal_wabbit\vowpal_wabbit\vw\cs\unittest\TestCbAdf.cs:line 166
  `

see https://github.com/VowpalWabbit/vowpal_wabbit/runs/5469418230?check_suite_focus=true

I've only seen this on "Windows / C++ - CMake only" CI

JohnLangford:
Fixed in #4838

PR #3816, opened 2022-03-20
ci: adapt Windows CMake workflow to windows 2022

### LINKED ISSUE #3750: ci: add windows-latest to Windows CI matrix
The current [Windows CI](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/.github/workflows/build_windows_cmake.yml) fails for Windows-2022 I think because the Vcpkg version is too old.

```
Run D:\a\vowpal_wabbit\vowpal_wabbit/vcpkg/bootstrap-vcpkg.bat
Could not find MSBuild version with C++ support. VS2015, VS2017, or VS2019 (with C++) needs to be installed.
At D:\a\vowpal_wabbit\vowpal_wabbit\vcpkg\scripts\bootstrap.ps1:195 char:5
+     throw "Could not find MSBuild version with C++ support. VS2015, V ...
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (Could not find ...o be installed.:String) [], RuntimeException
    + FullyQualifiedErrorId : Could not find MSBuild version with C++ support. VS2015, VS2017, or VS2019 (with C++) ne 
   eds to be installed.
```

We should fix this and add windows-latest back to the build matrix.

HollowMan6:
Resolves #3750: https://github.com/HollowMan6/vowpal_wabbit/runs/5615976754?check_suite_focus=true

* Upgrade vcpkg to 2022.03.10
* Remove the cmake -G option for adapting to the Visual Studio version.
As Visual Studio 16 2019 is not available in windows-2022 but Visual Studio 17 2022.
Cmake will be clever enough to find out the correct version as there's only [Visual Studio Enterprise 2022 in windows-2022](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md#visual-studio-enterprise-2022) and  [Visual Studio Enterprise 2019 in windows-2019 and windows-latest(currently)](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#visual-studio-enterprise-2019).

Test with both windows 2019 and 2022 can be found here: https://github.com/HollowMan6/vowpal_wabbit/actions/runs/2011159605

I've also checked the #3782, it seems like that the error arrives stochastically. Maybe it's some kind of bug from Visual Studio's component, but can't be sure what kind of bug it is because of its randomness, moreover maybe it's just some kind of memory shortage.

Signed-off-by: Hollow Man <hollowman@opensuse.org>

### DIFF
--- a/.github/workflows/build_windows_cmake.yml
+++ b/.github/workflows/build_windows_cmake.yml
@@ -11,16 +11,18 @@ on:
 
 jobs:
   check:
-    runs-on: windows-2019
     strategy:
+      fail-fast: false
       matrix:
         build_config: ["Debug", "Release"]
+        os: ["windows-2019", "windows-2022"]
+    runs-on: ${{ matrix.os }}
     env:
       CMAKE_BUILD_DIR: ${{ github.workspace }}/vw/build
       SOURCE_DIR: ${{ github.workspace }}/vw
       VCPKG_ROOT: ${{ github.workspace }}/vcpkg
-      # 2020.11-1
-      VCPKG_REF: ec6fe06e8da05a8157dc8581fa96b36b571c1bd5
+      # 2022.03.10
+      VCPKG_REF: af2287382b1991dbdcb7e5112d236f3323b9dd7a
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


jackgerrits:
Rather than replacing 2019 with 2022 I think it would be preferable to add to the matrix so we have coverage on both for now. Bumping vcpkg and removing -G looks fine to me

HollowMan6:
> I think we may need to add the OS to the vcpkg cache key as they are going to default to two different toolchains

Done image

I now also set fail-fast to false because of the disturbing random error like #3782 . Now you can get every job in one workflow to run completely even one matrix just failed.

PR #3817, opened 2022-03-22
fix: Json parser does not assign weight to example

cheng-tan:


test/unit_test/test_common.cc
```diff
@@ -14,6 +14,7 @@ VW::multi_ex parse_json(VW::workspace& all, const std::string& line)
   VW::multi_ex result;
   for (size_t i = 0; i < examples.size(); ++i)
   {
+    examples[i]->weight = all.example_parser->lbl_parser.get_weight(examples[i]->l, examples[i]->_reduction_features);
```
jackgerrits, line 17: You should use `setup_example` here. `get_weight` is an implementation detail and may change.

PR #3827, opened 2022-03-28
refactor: namespacing of uniform_hash

jackgerrits:


vowpalwabbit/vw_fwd.h
```diff
@@ -15,7 +15,6 @@ class io_buf;
 class loss_function;
 class parameters;
 struct features;
-struct random_state;
```
jackgerrits, line 17: This doesn't actually exist and is covered by rand_state

PR #3841, opened 2022-04-01
refactor: complete migration of setup functions to reductions namespace

jackgerrits:


test/unit_test/pmf_to_pdf_test.cc
```diff
@@ -115,20 +120,20 @@ BOOST_AUTO_TEST_CASE(pmf_to_pdf_basic)
...
-  const VW::pmf_to_pdf::predictions_t prediction_scores{{action, 1.f}};
+  const ::predictions_t prediction_scores{{action, 1.f}};
 
-  const auto test_harness = VW::pmf_to_pdf::get_test_harness_reduction(prediction_scores);
+  const auto test_harness = ::get_test_harness_reduction(prediction_scores);
 
   VW::example ec;
 
-  auto data = scoped_calloc_or_throw<VW::pmf_to_pdf::reduction>();
+  auto data = scoped_calloc_or_throw<VW::reductions::pmf_to_pdf_reduction>();
```
olgavrou, line 129: is the second `_reduction` redundant here?

vowpalwabbit/reductions/cb/cb_adf.cc
```diff
@@ -274,42 +307,12 @@ void cb_adf::learn_MTR(multi_learner& base, VW::multi_ex& examples)
...
-    }
-  }
-
-  return ret;
-}
-
 void cb_adf::learn(multi_learner& base, VW::multi_ex& ec_seq)
 {
-  if (test_adf_sequence(ec_seq) != nullptr)
+  if (CB_ADF::test_adf_sequence(ec_seq) != nullptr)
```
olgavrou, line 312: not related to this PR but it is unfortunate that we have `CB_ADF` and `cb_adf` namespaces

test/unit_test/pmf_to_pdf_test.cc
jackgerrits: I think that just `pmf_to_pdf` is ambiguous when in that namespace there will be a `pmf_to_pdf_reduction` and `pmf_to_pdf_setup`

vowpalwabbit/reductions/cb/cb_adf.h
```diff
@@ -3,30 +3,22 @@
...
+namespace reductions
 {
-struct writer;
+VW::LEARNER::base_learner* cb_adf_setup(VW::setup_base_i& stack_builder);
 }
 }  // namespace VW
 
-VW::LEARNER::base_learner* cb_adf_setup(VW::setup_base_i& stack_builder);
-
+// TODO: Move these functions into VW lib and not reductions
```
olgavrou, line 19: is this a follow up todo?

vowpalwabbit/reductions/cb/cb_adf.cc
jackgerrits: It looks like `cb_adf` is just the class name

vowpalwabbit/reductions/cb/cb_algs.cc
```diff
@@ -23,31 +23,43 @@ using namespace GEN_CS;
...
+void generic_output_example(
+    VW::workspace& all, float loss, const VW::example& ec, const CB::label& ld, const CB::cb_class* known_cost)
 {
-  cb_to_cs cbcs;
-  VW::io::logger logger;
+  all.sd->update(ec.test_only, !CB::is_test_label(ld), loss, 1.f, ec.get_num_features());
 
-  cb(VW::io::logger logger) : logger(std::move(logger)) {}
-};
+  for (auto& sink : all.final_prediction_sink)
```
olgavrou, line 31: nit:
```suggestion
  for (const auto& sink : all.final_prediction_sink)
```

?

vowpalwabbit/reductions/cb/cb_adf.h
jackgerrits: Yeah, not sure if its what we want or not but we need to think about where cb related library functionality lives. Under reductions or somewhere more general

vowpalwabbit/reductions/cb/cb_algs.cc
jackgerrits: I always feel funny with a change like this because the const is not transitive to the contents of the unique ptr. Reading this look I'd probably make an assumption that whatever goes on inside the loop does not change final_prediction_sink which isn't really true

olgavrou (approved):

PR #3844, opened 2022-04-04
fix: revert accidental change to output format

jackgerrits:


vowpalwabbit/action_score.cc
```diff
@@ -43,10 +43,10 @@ std::string to_string(const ACTION_SCORE::action_scores& action_scores_or_probs,
   std::string delim;
   for (const auto& item : action_scores_or_probs)
   {
-    ss << delim << fmt::format("{},{}", item.action, VW::fmt_float(item.score, decimal_precision));
+    ss << delim << fmt::format("{}:{}", item.action, VW::fmt_float(item.score, decimal_precision));
```
lalo, line 46: if all tests pass then we have missed a test

lalo (approved):

vowpalwabbit/action_score.cc
jackgerrits: I agree, I don't think tests will pass though

jackgerrits: Actually, yep I think we're missing some tests here

ataymano (approved):

vowpalwabbit/action_score.cc
jackgerrits: So it is covered by tests, but it seems like the harness is failing to see the diff here

PR #3856, opened 2022-04-08
refactor: move more components into libs

jackgerrits:
Components libified:
- active_interactor
- allreduce
- config
- io

Some common functionality was moved into common to avoid a circular dependency between config and core


utl/dump_options/CMakeLists.txt
```diff
@@ -1,2 +1,2 @@
 add_executable(vw-dump-options main.cc)
-target_link_libraries(vw-dump-options vw RapidJSON)
+target_link_libraries(vw-dump-options PUBLIC VowpalWabbit::vw RapidJSON)
```
bassmang, line 2: Is there a reason this is PUBLIC?

vowpalwabbit/config/tests/options_cli_test.cc
```diff
@@ -2,27 +2,22 @@
...
 using namespace VW::config;
 
-using option_types = boost::mpl::vector<options_cli>;
-
-BOOST_AUTO_TEST_CASE_TEMPLATE(typed_options_parsing, T, option_types)
+TEST(options_cli_test, typed_options_parsing)
 {
   std::vector<std::string> args = {"--str_opt", "test_str", "-i", "5", "--bool_opt", "--float_opt", "4.3"};
-  auto options = VW::make_unique<T>(args);
+  auto options = std::unique_ptr<options_cli>(new options_cli(args));
```
bassmang, line 20: Why this change?

utl/dump_options/CMakeLists.txt
jackgerrits: You should always specify PUBLIC/PRIVATE/INTERFACE. This is an exe so it doesn't really matter much so I defaulted to PUBLIC

vowpalwabbit/config/tests/options_cli_test.cc
jackgerrits: The boost implementation is gone and so the test file no longer needs to be templated over two types. Its a lot simpler to just use 1 type so I moved back to that

bassmang (approved):

PR #3862, opened 2022-04-12
fix: fix build issue in std latest for consteval fmt constructors

### LINKED ISSUE #3861: [MSVC][std:c++latest] Vowpal_wabbit failed to build with error C7595 on MSVC
### Describe the bug

Hi all,

The MSVC team added Vowpal_wabbit as part of RWC testing. Recently we found error C7595 when building Vowpal_wabbit with '/std:c++latest'. I opened a internal bug and the developer considered this to be a source issue. Could you please take a quick look at the issue? Thanks in advance.

I attached the repro code and it includes our workaround. I was going to apply this patch at first, but many lines are triggering this issue, could you fix it?

**Refer to the developer's explanation:**
A C string argument is passed to the call to logger_impl::err_critical(), and the body of the function attempts to convert the string to a fmt::basic_format_string using the consteval constructor. The folding of the consteval constructor call rightfully fails, since the constructor argument is fmt, a function parameter that has unknown value outside of the invocation of its function. To avoid this error, the conversion to fmt::basic_format_string needs to occur at the call site for logger_impl::err_critical(), which can be forced by changing the function signature for logger_impl::err_critical() so that it takes a fmt::basic_format_string parameter. The other functions in logger_impl may need to be similarly changed to avoid similar issues.

### How to reproduce

**Repro:**
1. open VS2019 v16.11.11  x64 tools command
2. cl.exe /std:c++latest bug_1505946.cpp

Repro code: [bug_1505946.cpp.txt](https://github.com/VowpalWabbit/vowpal_wabbit/files/8471307/bug_1505946.cpp.txt)
Build log on MSVC: [build.log](https://github.com/VowpalWabbit/vowpal_wabbit/files/8471426/build.log)

**Error**:
Correspond https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/io/logger.h#L135

```bug_1505946.cpp
bug_1505946.cpp(132): error C7595: 'fmt::basic_format_string<char,size_t &>::basic_format_string': call to immediate function is not a constant expression
bug_1505946.cpp(132): note: failure was caused by a read of a variable outside its lifetime
bug_1505946.cpp(132): note: see usage of 'fmt'
bug_1505946.cpp(160): note: see reference to function template instantiation 'void VW::io::details::logger_impl::err_critical<char[90],size_t&>(const FormatString (&),size_t &)' being compiled
        with
        [
            FormatString=char [90]
        ]
bug_1505946.cpp(135): error C7595: 'fmt::basic_format_string<char,size_t &>::basic_format_string': call to immediate function is not a constant expression
bug_1505946.cpp(135): note: failure was caused by a read of a variable outside its lifetime
bug_1505946.cpp(135): note: see usage of 'fmt'
bug_1505946.cpp(139): error C7595: 'fmt::basic_format_string<char,size_t &>::basic_format_string': call to immediate function is not a constant expression
bug_1505946.cpp(139): note: failure was caused by a read of a variable outside its lifetime
bug_1505946.cpp(139): note: see usage of 'fmt'```

### Version

master branch 7916bd4

### OS

Windows

### Language

C++

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by jackgerrits**
Resolves #3861

### CODE REVIEW

File: vowpalwabbit/parse_args.cc [hunk 1]
```diff
@@ -717,8 +717,8 @@ void parse_feature_tweaks(options_i& options, VW::workspace& all, bool interacti
        interactions_settings_duplicated /*settings were restored from model file to file_options and overriden by params from command line*/)
   {
     all.logger.err_warn(
-        "model file has set of {-q, --cubic, --interactions} settings stored, but they'll be "
-        "OVERRIDDEN by set of {-q, --cubic, --interactions} settings from command line.");
+        "model file has set of {{-q, --cubic, --interactions}} settings stored, but they'll be "
+        "OVERRIDDEN by set of {{-q, --cubic, --interactions}} settings from command line.");
```
bassmang, line 721: is this tested anywhere?

jackgerrits: Yes it is hit in one of the integration tests

bassmang (approved):

PR #3874, opened 2022-04-14
ci: update coverage ignore paths

jackgerrits:


codecov.yml
```diff
@@ -13,9 +13,12 @@ coverage:
         # Coverage can drop by 0.5% before it is a failure.
         threshold: 0.5%
         paths:
-          - vowpalwabbit
-          - explore
-          - "!test"
+          - "vowpalwabbit/"
```
olgavrou, line 16: what about explore?

jackgerrits: Explore is no longer a top level directory, its a lib under vowpalwabbit

olgavrou (approved):

PR #3880, opened 2022-04-15
build: allow package_ROOT for search paths and dont guess zlib dir

jackgerrits:
We want to avoid adding search paths into our CMakeLists.txt file for where packages can be found. I think the more correct cmake "way" here to is to let the user specify the search path if the default find package cant find it. 

So in the case of zlib on mac the user would invoke cmake like so:
```
cmake -S . -B build -DSTATIC_LINK_VW=ON -DZLIB_ROOT=`brew --prefix zlib`
```


bassmang:
> We want to avoid adding search paths into our CMakeLists.txt file for where packages can be found. I think the more correct cmake "way" here to is to let the user specify the search path if the default find package cant find it.
> 
> So in the case of zlib on mac the user would invoke cmake like so:
> 
> ```
> cmake -S . -B build -DSTATIC_LINK_VW=ON -DZLIB_ROOT=`brew --prefix zlib`
> ```

Is the default find_path different than the one hard-coded?

jackgerrits:
Depends on if consumed via the system or brew. Based on [this issue](https://github.com/VowpalWabbit/vowpal_wabbit/issues/3879) that was just opened the brew install location changed at some point

bassmang (approved):

Issue #1895, opened 2019-05-28
Allow multiple data files as input


sharathmalladi:
#### Describe the bug
Call vw with a bad argument and notice that vw does not return a non-zero error code. To detect whether vw rejected the arguments would require us to read the output and look for a line that says "sailing on!" .. which is not really a robust mechanism to return an error response.

#### To Reproduce
For example (notice the vw parameters are "bad vw arguments" which are invalid parameters):
VW COMMAND:

E:\sharathm\github\sharathmalladi-mwt-ds\DataScience>vw bad vw arguments -d D:/tmp/124bb2ca-a99f-489e-b29c-bc142baa6f51\6359742a010048a58c1892eabd731d4c\6359742a010048a58c1892eabd731d4c_merged_data_2019-01-03_2019-01-03.json.gz -p D:/tmp/124bb2ca-a99f-489e-b29c-bc142baa6f51\6359742a010048a58c1892eabd731d4c\6359742a010048a58c1892eabd731d4c_merged_data_2019-01-03_2019-01-03.json.gz.Custom Policy 1.pred predictions = D:/tmp/124bb2ca-a99f-489e-b29c-bc142baa6f51\6359742a010048a58c1892eabd731d4c\6359742a010048a58c1892eabd731d4c_merged_data_2019-01-03_2019-01-03.json.gz.Custom

Num weight bits = 18

learning rate = 0.5

initial_t = 0

power_t = 0.5

using no cache

Reading datafile = bad

can't open 'bad', sailing on!

num sources = 0

average  since         example        example  current  current  current

loss     last          counter         weight    label  predict features

 

finished run

number of examples = 0

weighted example sum = 0.000000

weighted label sum = 0.000000

average loss = n.a.

total feature number = 0

 

E:\sharathm\github\sharathmalladi-mwt-ds\DataScience>echo %ERRORLEVEL%

0

#### Expected behavior
The error code after invoking vw should be non-zero since vw did not successfully output the predictions.

#### Observed Behavior
We instead get back an output that has a line that reads:
can't open 'bad', sailing on!

#### Environment
What version of VW did you use?
8.6.1

What OS or language did you use?
Windows command line

#### Additional context
None

jackgerrits:
[...]

PR #3891, opened 2022-04-21
feat: Allow multiple data files as input

lalo:
implements #1895

- [x] happy path
- [ ] cache debug


olgavrou (approved):

test/core.vwtest.json
```diff
@@ -5123,5 +5123,35 @@
...
+    },
+    "input_files": [
+      "train-sets/dsjson_cb_part1.json",
+      "train-sets/dsjson_cb_part2.json"
+    ]
+  },
+  {
+    "id": 400,
+    "desc": "test multi d positional -- based on test 316 to share same metrics of the complete file dsjson_cb.json (which is split into two here)",
+    "vw_command": "--dsjson --cb_explore_adf --epsilon 0.2 --quadratic GT --extra_metrics metrics_time.json train-sets/dsjson_cb_part1.json train-sets/dsjson_cb_part2.json",
```
ataymano, line 5144: do we want to allow multiple positional arguments for input file? even single one is constant source of problems when someone accidentally adding wrong options. Multiple positional arguments could be source of problems that could be hard to notice (like every wrong commandline option will be treated as input file?). Maybe we should explicitly deprecate that?

vowpalwabbit/core/src/parse_args.cc
```diff
@@ -419,13 +419,7 @@ input_options parse_source(VW::workspace& all, options_i& options)
   // Check if the options provider has any positional args. Only really makes sense for command line, others just return
   // an empty list.
   const auto positional_tokens = options.get_positional_tokens();
-  if (positional_tokens.size() == 1) { all.data_filename = positional_tokens[0]; }
-  else if (positional_tokens.size() > 1)
-  {
-    all.logger.err_warn(
```
lalo, line 425: I'm reverting this warning too, so maybe we should keep it?

test/train-sets/ref/rcv1_raw_cb_explore_pass2_split.stderr
```diff
@@ -0,0 +1,35 @@
+predictions = rcv1_raw_cb_explore.preds
+creating cache_file = train-sets/rcv1_raw_cb_small_p2.vw.merged.cache
```
lalo, line 2: gotcha: one cache per run

test/core.vwtest.json
jackgerrits: 100% agree. We should not be expanding the positional input.

ataymano: And probably the same about mixed mode (like one input is positional, another - explicit)?

lalo: yes, but the log should provide help to decipher it. See test 404 for a weird mixed mode side effect in ordering.

python/vowpalwabbit/pyvw.py
```diff
@@ -176,9 +176,11 @@ def __str__(self):
             if self.is_flag():
                 return "--{}".format(self.name)
             else:
-                # missing list case
                 if isinstance(self.value, list):
-                    return "**NOT_IMPL**"
+                    res = ""
+                    for ele in self.value:
+                        res += f"--{self.name} {ele} "
```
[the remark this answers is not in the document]
lalo: outdated diff but might still apply

cs/cli/vw_arguments.h
```diff
@@ -36,11 +36,16 @@ public ref class VowpalWabbitArguments
   float m_power_t;
 
   internal : VowpalWabbitArguments(VW::workspace* vw)
```
jackgerrits, line 38: Does this mean C# cannot support multiple data files?

cs/cli/vw_arguments.h
```diff
@@ -36,11 +36,16 @@ public ref class VowpalWabbitArguments
...
-      : m_data(gcnew String(vw->data_filename.c_str()))
-      , m_finalRegressor(gcnew String(vw->final_regressor_name.c_str()))
+      : m_finalRegressor(gcnew String(vw->final_regressor_name.c_str()))
       , m_testonly(!vw->training)
       , m_passes((int)vw->numpasses)
   {
+    if (vw->data_filenames.size() >= 1) { m_data = gcnew String(vw->data_filenames.front().c_str()); }
+    else
+    {
+      m_data = gcnew String("");
```
jackgerrits, line 46: Please throw rather than silently ignore multiple test files

cs/cli/vw_arguments.h
lalo: yes, CLI for now and python thanks to the duck typing and support for lists (-q and other args)

cs/cli/vw_arguments.h
lalo: added throw just in case, but not entirely sure it can reach that state.

lalo:
https://github.com/VowpalWabbit/vowpal_wabbit/blob/c9cb804a74c94df4a00756c78f6e204526fcb7a9/vowpalwabbit/core/src/parser.cc#L311

test/core.vwtest.json
ataymano: hm, it still seems too scary to be honest - basically we are introducing a lot of ambiguity in parsing - for example if you have file with the same name as valid command line argument: let's say you have files with names "-q" and "sa" - then what command line "--cb_explore_adf -q sa -d file.txt" is supposed to do?
I understand that it is extreme corner case, but that change is increasing their amount substantially.

vowpalwabbit/core/src/parse_args.cc
```diff
@@ -419,27 +421,58 @@ input_options parse_source(VW::workspace& all, options_i& options)
...
-        "Multiple data files passed as positional parameters, only the first one will be "
-        "read and the rest will be ignored.");
-  }
 
-  if (parsed_options.daemon || options.was_supplied("pid_file") || (options.was_supplied("port") && !all.active))
+  if (!positional_tokens.empty() && !all.data_filenames.empty())
+  { THROW("You cannot mix --data/-d argument with positional arguments. Pick one and use consistently."); }
```
lalo, line 426: added this throw to allow users to pick either all --data or positional, would this be helpful as a previous step to deprecating positional?

PR #3892, opened 2022-04-21
feat: large action spaces initial commit

olgavrou:
Initial PR for large action space

- includes Eigen as a submodule dependency (pull in Eigen dir to VW which is header only)
- add a `reductions/cb/details/` dir to place large action space internal reduction api so that it can be re-usable for testing
- lazily multiply each action's feature weights with a matrix of gaussian entries and create the matrix Q which is dense and has reduced the feature space to `d` where d is an input parameter


test/unit_test/CMakeLists.txt
```diff
@@ -56,6 +57,7 @@ add_executable(vw-unit-test.out
 
 # Add the include directories from vw target for testing
 target_link_libraries(vw-unit-test.out PRIVATE vw_core RapidJSON Boost::unit_test_framework)
+target_include_directories(vw-unit-test.out PRIVATE ${VW_CORE_SOURCE_DIR})
```
jackgerrits, line 60: Rather than this, add the src directory as part of the include for the core lib and then:
```suggestion
target_include_directories(vw-unit-test.out PRIVATE $<TARGET_PROPERTY:vw_core,INCLUDE_DIRECTORIES>)
```

vowpalwabbit/core/CMakeLists.txt
```diff
@@ -344,7 +346,7 @@ vw_add_library(
     SOURCES ${vw_core_headers} ${vw_core_sources}
     PUBLIC_DEPS vw_common vw_explore vw_allreduce vw_config ${spdlog_target} fmt::fmt
     PRIVATE_DEPS
-        ${CMAKE_DL_LIBS} ${LINK_THREADS} vw_io $<BUILD_INTERFACE:${boost_math_target}>
+        ${CMAKE_DL_LIBS} ${LINK_THREADS} vw_io $<BUILD_INTERFACE:${boost_math_target}> eigen
```
jackgerrits, line 349:
```suggestion
        ${CMAKE_DL_LIBS} ${LINK_THREADS} vw_io $<BUILD_INTERFACE:${boost_math_target}> $<BUILD_INTERFACE:eigen>
```

test/unit_test/CMakeLists.txt
olgavrou: done, let me know if you think it's something we should put into the vw_add_library section

vowpalwabbit/core/CMakeLists.txt
```diff
@@ -354,6 +356,7 @@ vw_add_library(
     ${ENABLE_INSTALL_OPTION}
 )
 
+target_include_directories(vw_core PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src)
```
jackgerrits, line 359: `CMAKE_CURRENT_LIST_DIR` is usually what you expect when you use a variable like this. `CMAKE_CURRENT_SOURCE_DIR` works but if this file was included then I think it would behave differently
```suggestion
target_include_directories(vw_core PRIVATE ${CMAKE_CURRENT_LIST_DIR}/src)
```

vowpalwabbit/core/include/vw/core/reductions/cb/cb_explore_adf_large_action_space.h
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) by respective owners including Yahoo!, Microsoft, and
+// individual contributors. All rights reserved. Released under a BSD (revised)
+// license as described in the file LICENSE.
+#pragma once
+
+#include "cb_explore_adf_common.h"
```
jackgerrits, line 6: Header not needed

vowpalwabbit/core/include/vw/core/reductions/cb/cb_explore_adf_large_action_space.h
```diff
@@ -0,0 +1,17 @@
...
+#include "vw/core/vw_fwd.h"
+
+#include <vector>
```
jackgerrits, line 9: header not needed

vowpalwabbit/core/src/reductions/cb/details/large_action_space.h
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) by respective owners including Yahoo!, Microsoft, and
+// individual contributors. All rights reserved. Released under a BSD (revised)
+// license as described in the file LICENSE.
+#pragma once
+
+#include "Eigen/Dense"
```
jackgerrits, line 6:
```suggestion
#include <Eigen/Dense>
```

vowpalwabbit/core/src/reductions/cb/details/large_action_space.h
```diff
@@ -0,0 +1,69 @@
...
+#include <iostream>
+#include <vector>
+
+namespace VW
+{
+namespace cb_explore_adf
+{
+struct cb_explore_adf_large_action_space
+{
+  uint64_t d = 0;
```
jackgerrits, line 21: Members should be prefixed with `_`

vowpalwabbit/core/src/reductions/cb/details/large_action_space.h
```diff
@@ -0,0 +1,69 @@
...
+  WeightsT& weights;
+  uint64_t column_index;
+  uint64_t seed;
+
+public:
+  LazyGaussianDotProduct(WeightsT& weights_, uint64_t column_index_, uint64_t seed_)
+      : weights(weights_), column_index(column_index_), seed(seed_)
+  {
+  }
+  inline float operator[](uint64_t index) const
```
jackgerrits, line 62: `inline` is redundant here

test/unit_test/cb_large_actions_test.cc
```diff
@@ -0,0 +1,104 @@
...
+// license as described in the file LICENSE.
+
+#include "reductions/cb/details/large_action_space.h"
+#include "test_common.h"
+#include "vw/core/rand48.h"
+#include "vw/core/rand_state.h"
+#include "vw/core/reductions/cb/cb_explore_adf_large_action_space.h"
+#include "vw/core/vw.h"
+
+#include <boost/test/unit_test.hpp>
```
jackgerrits, line 12: Maybe consider adding new tests in the new location and using gtest

vowpalwabbit/core/src/reductions/cb/details/large_action_space.h
```diff
@@ -0,0 +1,69 @@
...
+  template <bool is_learn>
+  void predict_or_learn_impl(VW::LEARNER::multi_learner& base, multi_ex& examples);
+};
+
+// this is meant to be called with foreach_feature
+// for each feature it will multiply the weight that corresponds to the index with the corresponding gaussian element
+// If the returned values are summed the end resutlt is the dot product between the weight vector of the features of an
+// example with a vector of gaussian elements
+template <typename WeightsT>
+struct LazyGaussianDotProduct
```
jackgerrits, line 50: Looks like this type can be just in the cc file

vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
```diff
@@ -0,0 +1,171 @@
...
+{
+  // create Q matrix with dimenstions Kxd where K = examples.size()
+  uint64_t num_actions = examples[0]->pred.a_s.size();
+  Q.resize(num_actions, d);
+
+  // TODO extend wildspace interactions before calling foreach
+  uint64_t row_index = 0;
+  for (auto* ex : examples)
+  {
+    if (!CB::ec_is_example_header(*ex))
```
jackgerrits, line 60: Given where this is in the stack this should be guaranteed? Maybe just make it an `assert`?

vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
```diff
@@ -0,0 +1,171 @@
...
+void cb_explore_adf_large_action_space::calculate_shrink_factor(const ACTION_SCORE::action_scores& preds, float min_ck)
+{
+  shrink_factors.clear();
+  for (size_t i = 0; i < preds.size(); i++)
+  { shrink_factors.push_back(std::sqrt(1 + d + (gamma / 4.0f * d) * (preds[i].score - min_ck))); }
+}
+
+inline void just_add_weights(float& p, float, float fw) { p += fw; }
+
+void cb_explore_adf_large_action_space::generate_Q(const multi_ex& examples)
```
jackgerrits, line 50: Does this produce the same result if called twice? If `Q` is not in the default state?

olgavrou: Yes it does, what is the default state of Q? This function resizes Q and re populates its cells

jackgerrits: Whatever state the default constructor places it in when the object is created

olgavrou: Yeah the function shouldn't care what state Q is in, it will take care of it

olgavrou: can add it as a test though

jackgerrits: Only if you want to, resize and fill makes sense to me

vowpalwabbit/core/src/reductions/cb/details/large_action_space.h
olgavrou: members or private members?

vowpalwabbit/core/src/reductions/cb/details/large_action_space.h
olgavrou: true

vowpalwabbit/core/src/reductions/cb/details/large_action_space.h
olgavrou: changed the private ones

vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
olgavrou: yup good point

vowpalwabbit/core/src/reductions/cb/details/large_action_space.h
jackgerrits: We weren't prescriptive but seems to imply all members.

> Class member variables should be prefixed with _. Example: uint32_t _number_of_actions;

https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Coding-guidelines#style

test/train-sets/ref/help.stdout
```diff
@@ -539,6 +539,12 @@ Weight Options:
                                             uint, keep, necessary)
     --replay_m_count arg                    How many times (in expectation) should each example be played
                                             (default: 1 = permuting) (type: uint, default: 1)
+[Reduction] Experimental: Contextual Bandit Exploration with ADF with large action space Options:
+    --cb_explore_adf                        Online explore-exploit for a contextual bandit problem with multiline
+                                            action dependent features (type: bool, keep, necessary)
+    --large_action_space                    Large action space exploration (type: bool, keep, necessary)
+    --max_actions arg                       Max number of actions to explore (type: uint, default: 50, keep)
+    --gamma arg                             Gamma hyperparameter (type: float, keep)
```
pmineiro, line 547: the `--gamma` hyperparameter is from squarecb.  we should reuse the arguments from there (there's another one iirc).

vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
```diff
@@ -0,0 +1,195 @@
...
+
+template <bool is_learn>
+void cb_explore_adf_large_action_space::predict_or_learn_impl(VW::LEARNER::multi_learner& base, multi_ex& examples)
+{
+  // Explore uniform random an epsilon fraction of the time.
+  if (is_learn) { base.learn(examples); }
+  else
+  {
+    base.predict(examples);
+
```
pmineiro, line 119: there should be a short circuit if there are less actions than the exploration budget.

vowpalwabbit/core/src/reductions/cb/details/large_action_space.h
olgavrou: Yeah we aren't. I don't like accessing public data members via `data._Q` though that seems wrong. This struct is going to evolve a bit so I'll modify it as we go along

test/train-sets/ref/help.stdout
olgavrou: synced offline, will get this reduction to call squarecb reduction

vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
olgavrou: synced offline, when d > num_actions, the spanner is simply [1:num_actions] (and no need to call SVD)

jackgerrits (approved):

PR #3903, opened 2022-05-02
Champ change audit

bassmang:


vowpalwabbit/core/include/vw/core/reductions/epsilon_decay.h
```diff
@@ -62,6 +64,8 @@ struct epsilon_decay_data
   double _epsilon_decay_alpha;  // Confidence interval
   double _epsilon_decay_tau;    // Count decay time constant
   parameters& _weights;
+  VW::io::logger& _logger;
```
jackgerrits, line 67: Make this a value rather than a reference so it has shared ownership over the logger

bassmang: updated

jackgerrits (approved):

PR #3905, opened 2022-05-03
fix: Add reduction features for simple label

bassmang:


vowpalwabbit/core/src/simple_label_parser.cc
```diff
@@ -106,11 +106,20 @@ VW::label_parser simple_label_parser = {
         const VW::named_labels* /*ldict*/, const std::vector<VW::string_view>& words,
         VW::io::logger& logger) { parse_simple_label(label.simple, red_features, words, logger); },
     // cache_label
-    [](const VW::polylabel& label, const VW::reduction_features&, io_buf& cache, const std::string& upstream_name,
-        bool text) { return VW::model_utils::write_model_field(cache, label.simple, upstream_name, text); },
+    [](const VW::polylabel& label, const VW::reduction_features& red_feats, io_buf& cache,
```
jackgerrits, line 109: There are some redundant functions in this file that are no longer needed. bufcache_simple_label etc. Can you remove them here?

bassmang: removed

jackgerrits (approved):

PR #3911, opened 2022-05-04
feat: enable using lambdas in foreach_feature

jackgerrits:


jackgerrits:
The perf difference due to lambda capture is unacceptable

PR #3920, opened 2022-05-05
fix: [Epsilon Decay] Keep weights consistent across models

bassmang:
Also adds `--constant_epsilon` flag


vowpalwabbit/core/src/reductions/epsilon_decay.cc
```diff
@@ -82,11 +82,14 @@ namespace
...
 {
-  auto& ep_fts = examples[0]->_reduction_features.template get<VW::cb_explore_adf::greedy::reduction_features>();
   uint64_t K = static_cast<uint64_t>(data._scored_configs.size());
-  auto& active_score = data._scored_configs[K - 1][K - 1];
-  ep_fts.epsilon = VW::reductions::epsilon_decay::decayed_epsilon(active_score.update_count);
-  base.predict(examples, active_score.get_model_idx());
+  if (!data._constant_epsilon)
+  {
+    auto& ep_fts = examples[0]->_reduction_features.template get<VW::cb_explore_adf::greedy::reduction_features>();
+    auto& active_score = data._scored_configs[K - 1][K - 1];
```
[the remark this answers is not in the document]
bassmang: changed

vowpalwabbit/core/include/vw/core/reductions/epsilon_decay.h
```diff
@@ -22,51 +22,62 @@ namespace epsilon_decay
...
-    uint64_t model_idx = 0;
+    uint64_t weight_idx = 0;
     for (uint64_t i = 0; i < num_configs; ++i)
     {
       std::vector<epsilon_decay_score> score_vec;
       for (uint64_t j = 0; j < i + 1; ++j)
       {
-        epsilon_decay_score s(epsilon_decay_alpha, epsilon_decay_tau, model_idx);
+        epsilon_decay_score s(epsilon_decay_alpha, epsilon_decay_tau);
         score_vec.push_back(s);
```
peterychang, line 56:
```suggestion
        score_vec.emplace_back(epsilon_decay_alpha, epsilon_decay_tau);
```

peterychang: there's no reason to create, copy, then destroy this object in every loop iteration

vowpalwabbit/core/include/vw/core/reductions/epsilon_decay.h
```diff
@@ -22,51 +22,62 @@ namespace epsilon_decay
...
-        ++model_idx;
       }
       _scored_configs.push_back(score_vec);
```
peterychang, line 58:
```suggestion
      _scored_configs.push_back(std::move(score_vec));
```

vowpalwabbit/core/include/vw/core/reductions/epsilon_decay.h
bassmang: updated

vowpalwabbit/core/src/reductions/epsilon_decay.cc
```diff
@@ -33,8 +35,129 @@ void epsilon_decay_score::update_bounds(float w, float r)
...
+}
+
+// Rebalance greater models to match lower shifted models
+void epsilon_decay_data::rebalance_greater_models(int64_t model_ind, int64_t swap_dist, int64_t num_models)
+{
+  int64_t greater_model = model_ind + swap_dist + 1;
+  for (int64_t curr_mod = greater_model; curr_mod < num_models; ++curr_mod)
+  {
+    for (int64_t score_ind = greater_model - swap_dist; score_ind > 0; --score_ind)
+    { _scored_configs[curr_mod][score_ind] = std::move(_scored_configs[curr_mod][score_ind - swap_dist]); }
```
peterychang, line 100: `std::move(_scored_configs[curr_mod][score_ind - swap_dist])`

The second index can become negative, which is undefined behavior

vowpalwabbit/core/src/reductions/epsilon_decay.cc
```diff
@@ -33,8 +35,129 @@ void epsilon_decay_score::update_bounds(float w, float r)
...
+  promote_model(model_ind, swap_dist);
+  rebalance_greater_models(model_ind, swap_dist, num_models);
+  clear_weights_and_scores(swap_dist, num_models);
+}
+
+void epsilon_decay_data::check_score_bounds()
+{
+  // If the lower bound of a model exceeds the upperbound of the champion, migrate the new model as
+  // the new champion.
+  auto num_models = static_cast<int64_t>(_scored_configs.size());
```
peterychang, line 129: you should make a variable.. something like
`size_t final_model_idx = _scored_configs.size() - 1;`
which would let you get rid of all this `num_models -1` business

vowpalwabbit/core/src/reductions/epsilon_decay.cc
```diff
@@ -33,8 +35,129 @@ void epsilon_decay_score::update_bounds(float w, float r)
...
+      break;
+    }
+  }
+}
+
+void epsilon_decay_data::check_horizon_bounds()
+{
+  // Check if any model counts are higher than the champion. If so, shift the model
+  // back to the beginning of the list and reset its counts
+  auto num_models = static_cast<int64_t>(_scored_configs.size());
```
peterychang, line 149: same as above

vowpalwabbit/core/src/reductions/epsilon_decay.cc
```diff
@@ -33,8 +35,129 @@ void epsilon_decay_score::update_bounds(float w, float r)
...
+  for (int64_t i = 0; i < num_models - 1; ++i)
+  {
+    if (_scored_configs[i][i].update_count > _min_scope &&
+        _scored_configs[i][i].update_count > std::pow(_scored_configs[num_models - 1][num_models - 1].update_count,
+                                                 static_cast<float>(i + 1) / num_models))
+    {
+      shift_model(i - 1, 1, num_models);
```
peterychang, line 156: `i` can become -1 in this case

vowpalwabbit/core/src/reductions/epsilon_decay.cc
```diff
@@ -82,131 +203,22 @@ namespace
 void predict(
     VW::reductions::epsilon_decay::epsilon_decay_data& data, VW::LEARNER::multi_learner& base, VW::multi_ex& examples)
 {
-  auto& ep_fts = examples[0]->_reduction_features.template get<VW::cb_explore_adf::greedy::reduction_features>();
-  uint64_t K = static_cast<uint64_t>(data._scored_configs.size());
-  auto& active_score = data._scored_configs[K - 1][K - 1];
-  ep_fts.epsilon = VW::reductions::epsilon_decay::decayed_epsilon(active_score.update_count);
-  base.predict(examples, active_score.get_model_idx());
+  uint64_t num_models = static_cast<uint64_t>(data._scored_configs.size());
```
peterychang, line 206: same as above

vowpalwabbit/core/src/reductions/epsilon_decay.cc
```diff
@@ -33,8 +35,129 @@ void epsilon_decay_score::update_bounds(float w, float r)
...
+// Promote model and all those lower with distance swap_dist
+void epsilon_decay_data::promote_model(int64_t model_ind, int64_t swap_dist)
+{
+  for (; model_ind >= 0; --model_ind)
+  {
+    for (int64_t score_ind = 0; score_ind < model_ind + 1; ++score_ind)
+    {
+      _scored_configs[model_ind + swap_dist][score_ind + swap_dist] = std::move(_scored_configs[model_ind][score_ind]);
+    }
+    std::swap(_weight_indices[model_ind + swap_dist], _weight_indices[model_ind]);
```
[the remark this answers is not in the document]
bassmang: Yes the indices do need to be swapped since we'll still need to use them after the actual weights are cleared

vowpalwabbit/core/src/reductions/epsilon_decay.cc
bassmang: Good catch, changed to `score_ind >= swap_dist`

vowpalwabbit/core/src/reductions/epsilon_decay.cc
bassmang: This is the expected behavior

vowpalwabbit/core/include/vw/core/reductions/epsilon_decay.h
```diff
@@ -22,47 +23,50 @@ namespace epsilon_decay
...
       , _epsilon_decay_tau(epsilon_decay_tau)
       , _weights(weights)
       , _logger(std::move(logger))
       , _log_champ_changes(log_champ_changes)
+      , _constant_epsilon(constant_epsilon)
   {
-    uint64_t model_idx = 0;
-    for (uint64_t i = 0; i < num_configs; ++i)
+    _scored_configs.reserve(model_count);
+    _weight_indices.resize(model_count);
```
peterychang, line 43: add a comment here mentioning that resize (rather than reserve) is required for iota

vowpalwabbit/core/include/vw/core/reductions/epsilon_decay.h
```diff
@@ -22,47 +23,50 @@ namespace epsilon_decay
...
+    std::iota(_weight_indices.begin(), _weight_indices.end(), 0);
+    for (uint64_t i = 0; i < model_count; ++i)
     {
-      std::vector<epsilon_decay_score> score_vec;
+      _scored_configs.emplace_back(i + 1);
```
peterychang, line 47: add a comment about what the `i+1` here is doing, since its not immediately obvious

vowpalwabbit/core/src/reductions/epsilon_decay.cc
```diff
@@ -26,6 +28,126 @@ namespace epsilon_decay
 {
 float decayed_epsilon(uint64_t update_count) { return static_cast<float>(std::pow(update_count + 1, -1.f / 3.f)); }
 
+void epsilon_decay_data::update_weights(VW::LEARNER::multi_learner& base, VW::multi_ex& examples)
+{
+  auto model_count = static_cast<int64_t>(_scored_configs.size());
+  CB::cb_class logged{};
+  uint64_t labelled_action = 0;
```
peterychang, line 35: I assume label 0 and default `cb_class` object is desired behavior when the `find_if()` call fails?

vowpalwabbit/core/src/reductions/epsilon_decay.cc
```diff
@@ -26,6 +28,126 @@ namespace epsilon_decay
...
+        const float w = (logged.probability > 0) ? a_s.score / logged.probability : 0;
+        for (int64_t j = 0; j <= i; ++j) { _scored_configs[i][j].update(w, r); }
+        break;
+      }
+    }
+  }
+}
+
+// Promote model and all those lower with distance swap_dist
+void epsilon_decay_data::promote_model(int64_t model_ind, int64_t swap_dist)
```
peterychang, line 66: nit: `model_ind` and `swap_dist` should probably be unsigned ints. I dont think negative values make sense or necessarily even work in this function?

vowpalwabbit/core/src/reductions/epsilon_decay.cc
```diff
@@ -26,6 +28,126 @@ namespace epsilon_decay
...
+
+// Clear values in removed weights and scores
+void epsilon_decay_data::clear_weights_and_scores(int64_t swap_dist, int64_t model_count)
+{
+  uint64_t params_per_weight = 1;
+  while (params_per_weight < static_cast<uint64_t>(model_count)) { params_per_weight *= 2; }
+  for (int64_t model_ind = 0; model_ind < model_count; ++model_ind)
+  {
+    for (int64_t score_ind = 0;
+         score_ind < std::min(static_cast<int64_t>(_scored_configs[model_ind].size()), swap_dist); ++score_ind)
```
peterychang, line 97: can you make `swap_dist` a size_t instead of doing this cast? Or does it need the ability to be negative?

vowpalwabbit/core/src/reductions/epsilon_decay.cc
bassmang: Updated so no updating happens if a label isnt found

vowpalwabbit/core/src/reductions/epsilon_decay.cc
bassmang: model_ind can become negative in one of the loops, I kept it all int so all the types match

PR #3930, opened 2022-05-11
feat: add experimental field for custom parser

jackgerrits:
This isn't attempting to replace the current method for setting the active parser, or solidify an interface. This is an experimental interface to be used for the external binary parser in rlclientlib. Once this is proven to work, any mention to BUILD_EXTERNAL_PARSER can be removed from this repo.


vowpalwabbit/core/include/vw/core/input_parser.h
```diff
@@ -0,0 +1,29 @@
...
+{
+namespace details
+{
+// Experimental.
+struct input_parser
+{
+  input_parser(std::string name) : _name(std::move(name)) {}
+  virtual ~input_parser() = default;
+
+  virtual bool next(VW::workspace& workspace_instance, io_buf& buffer, VW::v_array<VW::example*>& output_examples) = 0;
```
peterychang, line 21: We should probably switch to using vectors in the parsers before we solidify this interface (so we don't need to have a breaking change later)

jackgerrits: This interface is experimental, so by definition any changes are not breaking.

PR #3931, opened 2022-05-12
refactor: Convert parser usage of v_array<example*> to multi_ex

peterychang:


.vscode/settings.json
```diff
@@ -6,5 +6,8 @@
             ],
             "url": "./test/vwtest.schema.json"
         }
-    ]
+    ],
+    "files.associations": {
+        "cstdint": "cpp"
+    }
```
jackgerrits, line 12: Looks like an accidental add

jackgerrits (approved):

.vscode/settings.json
peterychang: why is this directory even in the repo...?

PR #3932, opened 2022-05-13
Fix: Add an expectile_q parameter.

zwd-ms:
Add a parameter for the expectile loss. Previously quantile_tau is reused, but they actually have different semantics.

### DIFF
--- a/vowpalwabbit/core/src/parse_args.cc
+++ b/vowpalwabbit/core/src/parse_args.cc
@@ -1073,6 +1073,7 @@ void parse_example_tweaks(options_i& options, VW::workspace& all)
   std::string named_labels;
   std::string loss_function;
   float quantile_loss_parameter = 0.0;
+  float expectile_loss_parameter = 0.0;
   float logistic_loss_min = 0.0;
   float logistic_loss_max = 0.0;
   uint64_t early_terminate_passes;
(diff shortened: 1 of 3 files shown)


vowpalwabbit/core/src/parse_args.cc
```diff
@@ -1110,6 +1111,8 @@ void parse_example_tweaks(options_i& options, VW::workspace& all)
       .add(make_option("quantile_tau", quantile_loss_parameter)
                .default_value(0.5f)
                .help("Parameter \\tau associated with Quantile loss. Defaults to 0.5"))
+      .add(make_option("expectile_q", expectile_loss_parameter)
```
jackgerrits, line 1114: Is  the default value `0`? If so, adding `.default_value(0.0f)` will be helpful to the end user

zwd-ms: It meant to be a "required" parameter when the expectile loss is used. We do not provide a default.
`0.0f` is actually invalid, and we check the range of the parameter value to make sure it is specified.

zwd-ms: Added "required" in help message.

jackgerrits: Do we check to make sure it is passed when quantile is being used?

jackgerrits: Sorry, for the required parameter for expectile loss do we error out if it isnt passed?

zwd-ms: Yes, we'll throw with an error message.

jackgerrits (approved):

PR #3950, opened 2022-05-23
fix: [automl] improve debug/introspection

lalo:
- add logger to automl
- use [NAME] for unprintable/blank char namespaces
- respect incoming example interactions, avoid setting it to nullptr


vowpalwabbit/core/src/reductions/automl.cc
```diff
@@ -545,6 +573,15 @@ void interaction_config_manager::revert_config(example* ec) { ec->interactions =
 template <typename CMType>
 void automl<CMType>::offset_learn(multi_learner& base, multi_ex& ec, CB::cb_class& logged, uint64_t labelled_action)
 {
+  VW::reductions::automl::interaction_vec_t* incoming_interactions = ec[0]->interactions;
+  for (VW::example* ex : ec)
+  {
+    _UNUSED(ex);
```
bassmang, line 579: removable?

lalo: without that line, the compiler complaints on Release mode because the assert gets compiled out and we have warnings as errors, so we get an error of unused var.

test/test-sets/ref/metrics_.json
```diff
@@ -1 +1 @@
...
...-q OO -q OR -q OS -q OT -q Oi -q Oj -q O[constant] -q RR -q RS -q RT -q Ri -q Rj -q R[constant] -q SS -q ST -q Si -q Sj -q S[constant] -q TT -q Ti -q Tj -q T[constant] -q ii -q ij -q i[constant] -q jj -q j[constant] -q [constant][constant] ","interactions_1":"-q EE -q EG -q EO -q ER -q ES -q ET -q Ei -q Ej -q E[constant] -q GG -q GM -q GO -q GR -q GS -q GT -q Gi -q Gj -q G[constant] -q MM -q MO -q MR -q MS -q MT -q Mi -q Mj -q M[constant] -q OO -q OR -q OS -q OT -q Oi -q Oj -q O[constant] -q RR -q RS -q RT -q Ri -q Rj -q R[constant] -q SS -q ST -q Si -q Sj -q S[constant] -q TT -q Ti -q Tj -q T[constant] -q ii -q ij -q i[constant] -q jj -q j[constant] -q [constant][constant] ","interactions_2":"-q EE -q EG -q EO -q ER -q ES -q ET -q Ei -q Ej -q E[constant] -q GG -q GM -q GO -q GR -q GS -q GT -q Gi -q Gj -q G[constant] -q MM -q MO -q MR -q MS -q MT -q Mi -q Mj -q M[constant] -q OO -q OR -q OS -q OT -q Oi -q Oj -q O[constant] -q RR -q RS -q RT -q Ri -q Rj -q R[constant] -q SS -q ST -q Si -q Sj -q S[constant] -q TT -q Ti -q Tj -q T[constant] -q ii -q ij -q i[constant] -q jj -q j[constant] -q [constant][constant] ","last_event_id":"9077f996581148978a0ebe2484260dab","last_event_time":""}
```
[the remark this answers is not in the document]
lalo: yeap, it is a json string for now. I guess we could do a native json but wanted to avoid that since I'm only using it to debug for now.

bassmang (approved):

PR #3953, opened 2022-05-24
fix: valgrind to fail when there are leaks

olgavrou:


lalo (approved):

test/unit_test/cb_large_actions_test.cc
```diff
@@ -19,9 +20,9 @@ using internal_action_space =
 BOOST_AUTO_TEST_CASE(creation_of_the_og_A_matrix)
 {
   auto d = 2;
-  auto& vw = *VW::initialize(
-      "--cb_explore_adf --large_action_space --max_actions " + std::to_string(d) + " --quiet --random_seed 5", nullptr,
-      false, nullptr, nullptr);
+  auto opts = VW::make_unique<VW::config::options_cli>(std::vector<std::string>{
```
jackgerrits, line 23: `VW::config::options_cli` constructor requires the tokens instead of the raw string. [`VW::split_command_line`](https://vowpalwabbit.org/docs/vowpal_wabbit/cpp/latest/namespaceVW.html#a634c99ddf9b3cc7c7f650def077a8d91) can be used if the string needs to be split, but in this case you can just write it out as the vector of args.

olgavrou: I'm just going to stick with non-experimental initializer for now

zwd-ms (approved):

PR #3955, opened 2022-05-26
feat: Apply shrink factor

olgavrou:
and unit tests for both epsilon greedy (no shrink factor applied) and square cb (where shrink factor should be applied)

unit tests are essentially the same but just looping over vw with square cb applied, where diag_M (diagonal matrix containing the shrink factors) is taken into account in the calculations or with epsilon greedy the diagonal scaling matrix is just the identity matrix

### DIFF
--- a/test/unit_test/cb_large_actions_test.cc
+++ b/test/unit_test/cb_large_actions_test.cc
@@ -844,7 +844,7 @@ BOOST_AUTO_TEST_CASE(check_finding_max_volume)
   auto& vw = *VW::initialize(
       "--cb_explore_adf --large_action_space --max_actions " + std::to_string(d) + " --quiet --random_seed 5", nullptr,
       false, nullptr, nullptr);
-  VW::cb_explore_adf::cb_explore_adf_large_action_space largecb(0, 1, false, &vw);
+  VW::cb_explore_adf::cb_explore_adf_large_action_space largecb(0, 1.f, 0.f, false, &vw);
   largecb.U = Eigen::MatrixXf{{1, 2, 3}, {4, 5, 6}, {7, 8, 9}, {0, 0, 0}, {7, 5, 3}, {6, 4, 8}};
   Eigen::MatrixXf X{{1, 2, 3}, {3, 2, 1}, {2, 1, 3}};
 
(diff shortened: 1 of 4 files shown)


test/unit_test/cb_large_actions_test.cc
```diff
@@ -79,192 +79,369 @@ BOOST_AUTO_TEST_CASE(creation_of_the_og_A_matrix)
...
+    std::vector<std::string> e_r;
+    vw.l->get_enabled_reductions(e_r);
+    if (std::find(e_r.begin(), e_r.end(), "cb_explore_adf_large_action_space") == e_r.end())
+    { BOOST_FAIL("cb_explore_adf_large_action_space not found in enabled reductions"); }
+
+    VW::LEARNER::multi_learner* learner =
+        as_multiline(vw.l->get_learner_by_name_prefix("cb_explore_adf_large_action_space"));
 
-    uint64_t num_actions = examples[0]->pred.a_s.size();
+    auto action_space = (internal_action_space*)learner->get_internal_type_erased_data_pointer_test_use_only();
```
lalo, line 213:
```suggestion
    internal_action_space* action_space = learner->get_internal_type_erased_data_pointer_test_use_only();
```
I think this should work too

test/unit_test/cb_large_actions_test.cc
```diff
@@ -79,192 +79,369 @@ BOOST_AUTO_TEST_CASE(creation_of_the_og_A_matrix)
...
+      examples.push_back(VW::read_example(vw, "0:0.4:0.5 | a_16 a_17 a_18"));
+
+      vw.learn(examples);
+      vw.finish_example(examples);
+    }
+
+    std::vector<std::string> e_r;
+    vw.l->get_enabled_reductions(e_r);
+    if (std::find(e_r.begin(), e_r.end(), "cb_explore_adf_large_action_space") == e_r.end())
+    { BOOST_FAIL("cb_explore_adf_large_action_space not found in enabled reductions"); }
```
lalo, line 208: get_learner_by_name throws if its not found, so maybe it can be minimized to BOOST_CHECK_NO_THROW ( ) but not sure if you can have an assignment inside that boost check function. also a nit, feel free to ignore

test/unit_test/cb_large_actions_test.cc
olgavrou: looks like the void pointer that is returned needs to be converted

vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
```diff
@@ -493,6 +508,12 @@ VW::LEARNER::base_learner* VW::reductions::cb_explore_adf_large_action_space_set
     options.insert("no_predict", "");
   }
 
+  if (options.was_supplied("squarecb"))
+  {
+    apply_shrink_factor = true;
+    gamma = options.get_typed_option<float>("gamma_scale").value();
```
zwd-ms, line 514: I think gamma needs to be computed like this (see squarecb wiki)?
gamma = gamma_scale * t^{gamma_exponent}

olgavrou: do you want to link the wiki for reference?

yes looking at the square cb code it looks like that, will need to also save load the model for that

zwd-ms: See "Options" section [here](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Contextual-Bandit-Exploration-with-SquareCB#options)

zwd-ms (approved):

PR #3959, opened 2022-05-27
feat: add support for dict features in pylibvw

mrucker:
Hi,

I've been using VW as a constituent learner in some ensemble learner research. During this research I've noticed that VW is a performance bottleneck on sparse data sets with many features. I decided to see how much of this bottleneck I could alleviate via optimizing the python interface. In my testing these code changes gave me approximately an 8x speed up in my use case.

Changes:
- Added `ex_push_feature_dict` to pylibvw.cc. This method mirrors `ex_push_feature_list` but expects a dict instead of a list. 
- Added a `test_example_features_dict` to test_pyvw.py

Regression Testing:
- These changes use a new data structure which we check for up-front. This means that all previous functionality is completely untouched. You can see this in `ex_push_dictionary`.

Naming Conventions:
- These changes do cause some naming confusion but I didn't feel comfortable changing too many names. Notably what is currently called `ex_push_dictionary` in c is `push_feature_dict` on the Example class. I'm happy to change the names for this pull request given direction.

Speed Ups:
- The speed improvements are due to more efficient translation of python objects to c objects 
  - Fewer translations -- using a dict I was able to avoid translating tuples for each key-value pair. 
  - Faster translations -- Python's C API translated more quickly than boost extraction (I don't know if this is true on all platforms)

Cross Platform:
- I did end up using a few C API methods that are not part of the stable ABI. We can replace these if necessary but I don't expect them to be a problem.

Benchmarking:
- Testing just the translation of py objects to c objects the new code has about an 8x speed up.
image

- Testing end to end in my learner that uses pyvw I observe average speed up around 1.3 times across 3 data sets
image


lalo (approved):

lalo:
Thanks for the contribution Mark 👍

mrucker:
Of course, thanks for the review and acceptance. I'm excited for these to get deployed to pypi.

lalo:
@mrucker if you are in a hurry you should be able to manually grab the wheel from our CI, it is outlined here:
https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Python#bleeding-edge-latest-commit-on-master

mrucker:
Oh nice, I didn't know those were available online. 

That's a good work around instead of doing a fresh build from source. Thanks!

PR #3960, opened 2022-05-29
feat: [automl] store exclusions as flat set

skzhang1:
first comment


test/unit_test/automl_test.cc
```diff
@@ -274,7 +276,7 @@ BOOST_AUTO_TEST_CASE(namespace_switch)
     auto champ_exclusions = aml->cm->configs[aml->cm->scores[aml->cm->current_champ].config_index].exclusions;
     BOOST_CHECK_EQUAL(champ_exclusions.size(), 1);
 
-    BOOST_CHECK(champ_exclusions.find('U') != champ_exclusions.end());
+    // BOOST_CHECK(champ_exclusions.find('U') != champ_exclusions.end()); TODO: how to handle this?
```
bassmang, line 279: You can check that 'U' doesn't exist in any of the champ's exclusions

lalo:
@bassmang should this be opt-in with a flag? also to easily compare between having this behavior vs not.

bassmang (approved):

Issue #3935, opened 2022-05-17
Indexing 0 wrong output order


FabianKaiser:
### Describe the bug

While trying to use 0 indexing when migrating to VW 9.1, I noticed the performance was much worse than when using 1 indexing.
This seems to not be an error of the model, but of the way the 0 indexed model outputs its probabilities.

The predictions of the 0 indexed model looked like this (data and code below)
[0.399, 0.038, 0.125, 0.111, 0.327]
[0.249, 0.095, 0.185, 0.147, 0.324]
[0.244, 0.102, 0.209, 0.159, 0.286]
[0.321, 0.136, 0.16, 0.184, 0.199]

While the predictions of the 1 indexed model looked like this.
[0.326, 0.4, 0.038, 0.126, 0.111]
[0.323, 0.249, 0.096, 0.185, 0.147]
[0.286, 0.245, 0.102, 0.21, 0.159]
[0.199, 0.322, 0.136, 0.16, 0.184]

It seems the 0 index classifier outputs its results at the last index of the output list, not the first (which would be expected). Therefore a conversion of index to classifier (and such class) does not work correctly with 0 indexed models.

### How to reproduce

```python
from vowpalwabbit import Workspace

texts = [
    'a b c d',
    'a b c e',
    'a b d e',
    'a b c',
    'a c d',
    'a b c d e f',
    'a b d',
    'a d e',
    'a b d',
    'c d e',
    'a c',
]
labels = [
    0,
    1,
    1,
    0,
    3,
    4,
    3,
    1,
    0,
    1,
    0,
]

tata = [
    'a b c d e',
    'a b',
    'c d',
    'e'
]

for ind in [0, 1]:

    print('------------')
    print(ind)
    print('------------')

    test = f'{ind}_data.vw'
    path = f'{ind}_model.vw'

    with open(test, "wb") as f:
        for label, text in zip(labels, texts):
            example = f'{label + ind} |text {text}\n'
            f.write(example.encode())

    params = {
        'loss_function': 'logistic',
        'data': test,
        'probabilities': True,
        'c': True,
        'k': True,
        'f': path,
        'compressed': True,
        'oaa': 5
    }
    model = Workspace(**params)

    for t in tata:
        print([round(p, 3) for p in model.predict(f'|text {t}')])
    print('------------')

    model.finish()

    del model

    params = {
        'loss_function': 'logistic',
        'probabilities': True,
        'predict_only_model': True,
        'i': path
    }

[...]

PR #3965, opened 2022-06-02
fix: maintain order of indexed probabilities in oaa and csoaa

bassmang:
resolves: #3935


vowpalwabbit/core/src/reductions/oaa.cc
```diff
@@ -34,10 +34,10 @@ struct oaa
   uint64_t num_subsample = 0;           // for randomized subsampling, how many negatives to draw?
   uint32_t* subsample_order = nullptr;  // for randomized subsampling, in what order should we touch classes
   size_t subsample_id = 0;              // for randomized subsampling, where do we live in the list
-  int indexing = -1;                    // for 0 or 1 indexing
   VW::io::logger logger;
+  uint32_t& indexing;  // for 0 or 1 indexing
```
lalo, line 38: can you add a comment for indexing == 2 or put it as a constant somewhere with a clear name?

vowpalwabbit/core/include/vw/core/global_data.h
```diff
@@ -329,6 +329,8 @@ struct workspace
   // hack to support cb model loading into ccb reduction
   bool is_ccb_input_model = false;
 
+  uint32_t indexing = 2;
```
lalo, line 332: maybe add the comment here too

vowpalwabbit/core/src/parse_args.cc
```diff
@@ -682,7 +682,8 @@ void parse_feature_tweaks(options_i& options, VW::workspace& all, bool interacti
                .help("Don't remove interactions with duplicate combinations of namespaces. For ex. this is a "
                      "duplicate: '-q ab -q ba' and a lot more in '-q ::'."))
       .add(make_option("quadratic", quadratics).short_name("q").keep().help("Create and use quadratic features"))
-      .add(make_option("cubic", cubics).keep().help("Create and use cubic features"));
+      .add(make_option("cubic", cubics).keep().help("Create and use cubic features"))
+      .add(make_option("indexing", all.indexing).one_of({0, 1}).keep().help("Choose between 0 or 1-indexing"));
```
lalo, line 686: verify that the logic of one_of doesn't clash with the default value of 2 of indexing

bassmang: Yea it works fine if the `indexing` option isn't supplied. 2 is meant as a placeholder if nothing is specified and shouldn't be an option for the user.

lalo (approved):

PR #3974, opened 2022-06-14
refactor: [automl] Don't allow inactive configs as champ, clean up lower/upper bound

bassmang:


vowpalwabbit/core/src/scored_config.cc
```diff
@@ -38,6 +38,10 @@ void scored_config::reset_stats(double alpha, double tau)
   update_count = 0;
 }
 
+float scored_config::lower_bound() { return chisq.lower_bound(); }
+
+float scored_config::upper_bound() { return current_ips(); }
```
lalo, line 43: I'm guessing this is temporary

vowpalwabbit/core/src/reductions/automl.cc
```diff
@@ -411,15 +409,17 @@ void interaction_config_manager::schedule()
...
 
-bool interaction_config_manager::better(const exclusion_config& challenger, const exclusion_config& champ) const
+bool interaction_config_manager::better(uint64_t challenger, uint64_t champ)
 {
-  return challenger.lower_bound > champ.ips;
+  return scores[challenger].lower_bound() > scores[champ].upper_bound();
 }
 
-bool interaction_config_manager::worse(const exclusion_config& challenger, const exclusion_config& champ) const
+bool interaction_config_manager::worse(uint64_t challenger, uint64_t champ)
```
lalo, line 417: you might be able to skip the unused if you remove the names from the function signature

vowpalwabbit/core/src/scored_config.cc
bassmang: Yes this should be temporary until we find something better for upper_bound

PR #3977, opened 2022-06-15
feat: [automl] apples to apples score comparison

bassmang:


test/unit_test/automl_test.cc
```diff
@@ -76,7 +76,7 @@ BOOST_AUTO_TEST_CASE(automl_first_champ_switch)
 {
   const size_t num_iterations = 1331;
   const size_t seed = 10;
-  const size_t deterministic_champ_switch = 161;
+  const size_t deterministic_champ_switch = 465;
```
lalo, line 79: is it worthwhile to assert that the scores are reset after the champ switch occurs?

bassmang: Yea good point

lalo (approved):

PR #3985, opened 2022-06-17
fix: match cpp cressieread with python

bassmang:


vowpalwabbit/core/src/distributionally_robust.cc
```diff
@@ -108,11 +108,11 @@ ChiSquared& ChiSquared::update(double w, double r)
   return *this;
 }
 
-double ChiSquared::qlb(double w, double r)
+double ChiSquared::qlb(double w, double r, double sign)
```
lalo, line 111: add comment here about the sign convention, just to be extra clear.

lalo (approved):

vowpalwabbit/core/src/distributionally_robust.cc
bassmang: updated

PR #3998, opened 2022-06-27
fix: [automl] patch state (gd/cb_adf) being shared by multiple configs

lalo:
- fix test to make sure that weights are equal even if the learning order changes
- add test that compares "-q ::" vs "automl", in a setting where the champ does not switch, therefore the weights have to be exactly the same (same floats, no precision error)
 

- [x] add extra state to model
- [ ] reset state after config switch?
- [ ] check what other feature state is being mutated on model (reversed vs non-reversed)


lalo:
This should be merged after https://github.com/VowpalWabbit/vowpal_wabbit/pull/4000

bassmang (approved):

PR #4001, opened 2022-06-28
feat: [automl] algo optimizations

bassmang:
- Removes `--keep_configs` flag
- Maintains old champ model and statistics after champ_switch
- Disallows exclusion repeats
- Handles weight swapping on champ_switch


test/unit_test/automl_test.cc
```diff
@@ -19,22 +19,6 @@ using simulator::callback_map;
 using simulator::cb_sim;
 using namespace VW::reductions::automl;
 
-#define TEST_SCORE_CLEARED(aml)                    \
```
lalo, line 22: is this temp

bassmang: this test isn't valid anymore since the scores aren't cleared on a champ change. The old and incoming champ both stay live

vowpalwabbit/core/include/vw/core/array_parameters_dense.h
```diff
@@ -205,7 +205,13 @@ class dense_parameters
         if (*other != 0.f || *iter != 0.f)
```
lalo, line 205: theres a bug here, a tiny one fixing in another PR

lalo (approved):

PR #4003, opened 2022-06-28
feat: [parameters_dense] add more operators to reduce repetition

lalo:


vowpalwabbit/core/include/vw/core/array_parameters_dense.h
```diff
@@ -41,7 +42,10 @@ class dense_iterator
   using pointer = T*;
   using reference = T&;
 
-  dense_iterator(T* current, T* begin, uint32_t stride) : _current(current), _begin(begin), _stride(stride) {}
+  dense_iterator(T* current, T* begin, uint32_t stride, uint32_t stride_shift)
```
jackgerrits, line 45: Stride shift can be calculated form stride, why set it specifically?

jackgerrits (approved):

PR #4014, opened 2022-06-30
feat: [cb_explore_adf] persist cb_adf scores optionally

lalo:


test/pred-sets/ref/cb_adf_scores_inside_explore.pred
```diff
@@ -0,0 +1,3 @@
+0:0,1:0,2:0
+1:0,0:0.239436
```
lalo, line 2: skipping 2?

lalo: that example has only 2 actions

bassmang (commented):
what is this being used for?

bassmang (approved):

PR #4017, opened 2022-06-30
feat: use std::make_unique when using C++14 or newer

jackgerrits:


vowpalwabbit/core/include/vw/core/memory.h
```diff
@@ -67,11 +67,16 @@ free_ptr<T> scoped_calloc_or_throw(Args&&... args)
 
 namespace VW
 {
+#if __cplusplus >= 201402L  // C++14 and beyond
```
lalo, line 70: do we have ci that tests with below c++14?

jackgerrits: All CI targets 11 unless USE_LATEST_STD is passed to cmake, so I think most are 11

lalo (approved):

PR #4019, opened 2022-07-01
feat: [Java] Support predict or learn directly from string

jackgerrits:
- Add support for calling learn or predict on single line examples
- Add support for passing a string to predict or learn
- Supports Text or Json/DSJson
- Expose `DecisionProbs` to `VowpalWabbitNative`


java/src/main/java/org/vowpalwabbit/spark/VowpalWabbitNative.java
```diff
@@ -52,6 +64,18 @@ public class VowpalWabbitNative implements Closeable {
...
 
+    private native Object predictFromString(String examplesString);
+
+    /**
+     * Invoke multi-line prediction. By default will interpret input as text format, pass `--dsjson` or `--json` to initialize to use DSJSON or JSON respectively.
+     *
+     * @param examplesString String representation of examples to learn from.
+     * @return the prediction.
+     */
+    public Object predict(String examplesString) {
```
eisber, line 75: remove it. no need to duplicate.

eisber (changes requested):
wow.

java/src/main/java/org/vowpalwabbit/spark/VowpalWabbitNative.java
jackgerrits: I added these so it just behaves as an overload rather than needing to call a different method. You don't think that's necessary?

eisber: I think the learnFromString is obvious enough. duplicate methods usually lead to confusion.

jackgerrits: Sounds good

eisber (approved):

PR #4026, opened 2022-07-06
fix: don't use no_predict by default in cb_adf_greedy

bassmang:


vowpalwabbit/core/src/reductions/cb/cb_explore_adf_greedy.cc
```diff
@@ -127,7 +127,6 @@ VW::LEARNER::base_learner* VW::reductions::cb_explore_adf_greedy_setup(VW::setup
   if (!options.was_supplied("cb_adf"))
   {
     options.insert("cb_adf", "");
-    options.insert("no_predict", "");
   }
```
lalo, line 130: this would have to be added outside the if always right?

lalo: The misuse here is that the first time it runs, it inserts no_predict, but nopredict is not marked as keep, the next time you load it, no predict will not be there. I don't remember how we ran into this issue though.

lalo:
closing, we can re-open if it becomes relevant again.

PR #4027, opened 2022-07-07
fix: keep --noconstant flag

bassmang:


lalo:
interesting that this bug was not found by this test https://github.com/VowpalWabbit/vowpal_wabbit/blob/01aad375f1855319cb18aff4c3e008c55ea66274/test/save_resume_test.py#L181

olgavrou:
> interesting that this bug was not found by this test, if it didn't fail it might mean that it was not affecting the model so the test might be wrong
> 
> https://github.com/VowpalWabbit/vowpal_wabbit/blob/01aad375f1855319cb18aff4c3e008c55ea66274/test/save_resume_test.py#L181

is this a bug? do we have the option of un-keeping the non constant if we decide we want to add it as a feature later?

bassmang:
> > interesting that this bug was not found by this test, if it didn't fail it might mean that it was not affecting the model so the test might be wrong
> > https://github.com/VowpalWabbit/vowpal_wabbit/blob/01aad375f1855319cb18aff4c3e008c55ea66274/test/save_resume_test.py#L181
> 
> is this a bug? do we have the option of un-keeping the non constant if we decide we want to add it as a feature later?

I don't think there would be an option to remove --noconstant once it has been kept

olgavrou:
> In general `keep()` shouldn't be used for runtime arguments and `keep()` should be used for arguments that are considered part of the model. The feature-related args like `interactions`, `cubic`, etc are kept and considered part of the model, so it would follow that `noconstant` is treated the same

fair enough

jackgerrits (approved):

PR #4030, opened 2022-07-07
fix: [largecb] Add more comments in code.

zwd-ms:


vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
```diff
@@ -478,10 +482,13 @@ void cb_explore_adf_large_action_space::update_example_prediction(VW::multi_ex&
   }
   else
   {
+    // When the number of actions is no larger than d, all actions are selected.
```
olgavrou, line 485:
```suggestion
    // When the number of actions is not larger than d, all actions are selected.
```

olgavrou (approved):

PR #4057, opened 2022-07-19
fix: dict processing in pyvw on linux operating systems

mrucker:
The c-api string encoding fails on linux operating systems when using the new dict features in vw examples. This pull request both fixes the bug and improves the unit test for the dict feature functionality so that it catches this bug.

### DIFF
--- a/python/pylibvw.cc
+++ b/python/pylibvw.cc
@@ -684,12 +684,12 @@ void ex_push_feature_dict(example_ptr ec, vw_ptr vw, unsigned char ns, PyObject*
   uint64_t ns_hash = VW::hash_space(*vw, ns_str);
   size_t count = 0;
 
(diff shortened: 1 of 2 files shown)


mrucker:
I think I got it this time. One of my inlined c statements in pylibvw.cc behaved differently on Linux. 

The unit tests catch the problem now. I moved the inlined statement to its own line. Hopefully the unit tests pass now.

mrucker:
@lalo How would you like to handle the pytype failing check? It's catching problems with code I didn't touch.

I'm happy to fix the problems as part of this pull request as well. I just wanted to make sure that's what you'd want since its a little outside of the scope of the initial pull request (that is, fixing the Linux bug with the new Python feature dict functionality).

lalo:
I'll take a look @mrucker

PR #4062, opened 2022-07-20
fix: [py] leave 1 CPU for OS when building bindings

lalo:


olgavrou (approved):

setup.py
```diff
@@ -146,9 +146,12 @@ def build_cmake(self, ext):
             cmake_args += [
                 "-DCMAKE_LIBRARY_OUTPUT_DIRECTORY=" + str(lib_output_dir),
             ]
+            cpu_count = multiprocessing.cpu_count()
+            if cpu_count > 2:
```
olgavrou, line 150: >=?
```suggestion
            if cpu_count >= 2:
```

lalo: intentional, but I might just add a new arg to override this behaviour

lalo:
a flag seems better

PR #4065, opened 2022-07-21
test: [learner] WIP: add simple counter for learn

lalo:


rajan-chari:
What's the cost of measuring performance per call reduction?  This could be behind a flag and something we pay for when needed.

rajan-chari:
Also seems like <chrono> uses QPC (at least for windows).  Is this a viable cross platform alternative?  
https://docs.microsoft.com/en-us/cpp/standard-library/chrono?view=msvc-170

lalo:
Since some reductions call into predict, the accounting of time of what is inside the reduction versus below becomes harder to split, and more if the user is also making calls into predict. This makes this change less valuable. ccb is one of the reductions that consistently calls into predict, causing some lower reductions to report 0 usage on learn time.

PR #4066, opened 2022-07-21
fix: Correct Feature Index Data Type

mrucker:
The feature index data type used by the Python layer was uint32_t while VW feature group actually used uint64_t. This pull request updates the Python layer to use VW's feature_index datatype. A unit test was added for this correction. This pull request also speeds up the Python feature_dict by 20% for integer feature values.

### DIFF
--- a/python/pylibvw.cc
+++ b/python/pylibvw.cc
@@ -684,12 +684,12 @@ void ex_push_feature_dict(example_ptr ec, vw_ptr vw, unsigned char ns, PyObject*
   uint64_t ns_hash = VW::hash_space(*vw, ns_str);
   size_t count = 0;
 
-  PyObject *name, *value;
+  PyObject *key, *value, *bytes;
   Py_ssize_t size = 0, pos = 0;
   float feat_value;
   uint64_t feat_index;
 
-  while (PyDict_Next(o, &pos, &name, &value))
+  while (PyDict_Next(o, &pos, &key, &value))
   {
     if (PyFloat_Check(value)) { feat_value = (float)PyFloat_AsDouble(value); }
     else if (PyLong_Check(value))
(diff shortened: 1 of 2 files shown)


python/pylibvw.cc
```diff
@@ -688,31 +691,32 @@ void ex_push_feature_dict(example_ptr ec, vw_ptr vw, unsigned char ns, PyObject*
...
-    {
-      feat_value = (float)PyLong_AsDouble(value);
-    }
+    if (PyLong_Check(value)) { feat_value = (float)PyLong_AsDouble(value); }
     else
     {
-      std::cerr << "warning: malformed feature in list" << std::endl;
-      continue;
+      feat_value = (float)PyFloat_AsDouble(value);
+      if (feat_value == -1 && PyErr_Occurred())
```
bassmang, line 702: Why this check?

mrucker: If `value` can't be turned into a float for some reason then this will get flagged (e.g., if `value` is a string). This is equivalent to the check on line 635 in ex_push_feature_list. In my benchmarks I found that doing the check this way was a tiny bit faster than trying to do error checking ourselves before calling `PyFloat_AsDouble`.

mrucker: Oh also, in case what you're asking is why the strange check for a value of -1 and then `PyErr_Occured`, [here's a link](https://docs.python.org/3/c-api/float.html#c.PyFloat_AsDouble) to the documentation. You can see that error handling is performed by checking for a value of -1 and then calling `PyErr_Occurred`. I'm assuming they do it this way for performance reasons, though I don't know.

bassmang: Cool thanks for the link

bassmang (approved):

python/pylibvw.cc
mrucker: Yeah, in my testing I've found that using the CPython functions directly is quite a bit faster than using the BoostPython methods, which is why I switched from the more readable boost methods to the CPython methods which are a little more cryptic.

PR #4073, opened 2022-07-27
feat: native CSV parsing

HollowMan6:
Previous reviews: https://github.com/HollowMan6/vowpal_wabbit/pull/1

A project of [Reinforcement Learning Open Source Fest 2022](https://vowpalwabbit.org/rlos/2022/projects#native-csv-parsing).

[Tutorial CSV](https://github.com/VowpalWabbit/vowpal_wabbit/files/9203605/iris.csv)

[Iris.csv](https://github.com/HollowMan6/vowpal_wabbit/files/8858418/Iris.csv)
[Iris.txt](https://github.com/HollowMan6/vowpal_wabbit/files/8646360/Iris.txt)
[iris2.csv](https://github.com/HollowMan6/vowpal_wabbit/files/8858420/iris2.csv)

Build CSV parser with cmake option `-DBUILD_CSV=On`

```bash
vw -d Iris.txt
vw --csv -d Iris.csv
vw --csv -d Iris2.csv  --named_labels Setosa,Versicolor,Virginica --oaa 3
```

Summary of features supported and tested:
1. Allows specifying the CSV field separator by --csv_separator, default is `,`, but `"` `|` or `:` are reserved and not allowed to use, since the double quote (`"`) is for escape, vertical bar(`|`) for separating the namespace and feature names, `:` can be used in labels.
2. For each separated field, auto remove the outer double-quotes of a cell when it pairs. `--csv_separator` symbols that appeared inside the double-quoted cells are not considered as a separator but a normal string character.
3. Double-quotes that appear at the start and end of the cell will be considered to enclose fields. Other quotes that appear elsewhere and out of the enclose fields will have no special meaning. (This is also how Microsoft Excel parses.)
4. If double-quotes are used to enclose fields, then a double-quote appearing inside a field must be escaped by preceding it with another double quote, and will remove that escape symbol during parsing.
5. Use header line for feature names (and possibly namespaces) / specify label and tag using `_label` and `_tag` by default. For each separated field in header except for tag and label, it may contain namespace and feature name separated by namespace separator, vertical bar(`|`).
6. `--csv_header` to override the CSV header by providing (namespace, `|` and)  feature name separated with `,`. By default, CSV files are assumed to  have a header with feature and/or namespaces names in the CSV first line.  You can override it by specifying `--csv_header`. Combined with `--csv_no_file_header`, we assume that there is no header in the CSV file and under such condition specifying `--csv_header` for the header is a must.
7. If the number of the separated fields for current parsing line is greater than the header, an error will be thrown.
8. Trim the field for ASCII "white space"(`\r\n\f\v`) as well as some UTF-8 BOM characters(`\xef\xbb\xbf`) before separation.
9.  If no namespace is separated, will use empty namespace.
10. Separator supports using `\t` to represent tabs. Otherwise, if assigning more than one character, an error will be thrown.
11. Directly read the label as string, interpret it using the VW text label parser.
12. Will try to judge if the feature values are float or string, if NaN, will consider it as a string. quoted numbers are always considered as strings.
13. If the feature value is empty, will skip that feature.
14. Reset the parser when EOF of a file is met (for possible multiple input file support).
15. Support using `--csv_ns_value` to scale the namespace values by specifying the float ratio.
e.g. --csv_ns_value=a:0.5,b:0.3,:8, which the namespace a has a ratio of 0.5, b of 0.3, empty namespace of 8, other namespaces of 1.
16. If all the cells in a line is empty, then consider it as an empty line. CSV is not a good fit for the multiline format, as evidenced by the large number of empty fields. Multi-line format often means different lines have different schemas. However, I still leave the empty line support to make sure that it’s flexible and extendable enough. In this case, users can still express multi-line examples in CSV files, although it is not listed as supported. We still throw an error if the number of fields separated by the line doesn’t match previous, even all the fields are empty, as this usually means typos that users may not intend.


rajan-chari (approved):

rajan-chari:
Looks very useful.  Thanks!

rajan-chari:
How does this interact with multiline examples?

rajan-chari:
Although not explicitly supported, I am wondering if having a multiline csv example in the test folder would be useful reference.

HollowMan6:
> Although not explicitly supported, I am wondering if having a multiline csv example in the test folder would be useful reference.

@rajan-chari Thanks for reviewing, I have added that in https://github.com/VowpalWabbit/vowpal_wabbit/pull/4073/commits/d7e5e914fac858082a7e67143fa4e8d341bfb83a

.github/workflows/build_nugets.yml
```diff
@@ -35,6 +35,7 @@ jobs:
           -T ${{ matrix.toolset }}
           -DNATIVE_NUGET_PLATFORM_TAG="${{ matrix.arch_config.arch }}"
           -DBUILD_FLATBUFFERS=Off
+          -DBUILD_CSV=Off
```
olgavrou, line 38: nit: `BUILD_CSV` is Off by default so can just omit in in these builds

HollowMan6: Fixed, thanks for reviewing!

CMakeLists.txt
```diff
@@ -117,6 +117,7 @@ option(VW_BOOST_MATH_SYS_DEP "Override using the submodule for boost math depend
 option(VW_ZLIB_SYS_DEP "Override using the submodule for zlib dependency. Instead will use find_package" ON)
 option(VW_GTEST_SYS_DEP "Override using fetch package for gtest dependency. Instead will use find_package" OFF)
 option(VW_BUILD_VW_C_WRAPPER "Enable building the c_wrapper project" ON)
+option(BUILD_CSV "Build csv parser" OFF)
```
byronxu99, line 120: I believe it's generally preferred to name your options like `VW_BUILD_CSV` to avoid duplicate names with other cmake projects

jackgerrits: +1 to this

olgavrou, line 120: maybe to match with the rest of the options:

```suggestion
option(VW_BUILD_CSV "Build csv parser" OFF)
```

vowpalwabbit/csv_parser/src/parse_example_csv.cc
```diff
@@ -0,0 +1,517 @@
...
+{
+namespace parsers
+{
+int parse_csv_examples(VW::workspace* all, io_buf& buf, VW::multi_ex& examples)
+{
+  bool keep_reading = all->custom_parser->next(*all, buf, examples);
+  return keep_reading ? 1 : 0;
+}
+
+void csv_parser::handling_csv_separator(std::string& str, const std::string& name)
```
olgavrou, line 24: since this method has side effects (i.e. might change the separator passed in as an argument) it would be nice to have the methods name declare its intentions e.g.
```suggestion
void csv_parser::set_csv_separator(std::string& str, const std::string& name)
```

CMakeLists.txt
HollowMan6: Fixed, thanks for reviewing!

HollowMan6: Fixed!

vowpalwabbit/csv_parser/src/parse_example_csv.cc
HollowMan6: Fixed!

olgavrou (approved):
LGTM

<3 the extensive test coverage

HollowMan6:
Just rebased the PR and it seems like that the CI error is not cause by this PR but recent commits in the master, as that passes for the previous run https://github.com/VowpalWabbit/vowpal_wabbit/actions/runs/2833443417

jackgerrits:
I agree, that break is not introduced by this PR. It snuck in in #4109.

jackgerrits:
Awesome stuff, merging now :)

PR #4077, opened 2022-07-29
build: Support vcpkg manifest and cmakepreset for easy building

jackgerrits:
This change makes it possible to build VowpalWabbit from a fresh clone with:

```
cmake --preset=vcpkg-debug
cmake --build --preset=vcpkg-debug
```

This also enables a user to open a fresh clone of VowpalWabbit in Visual Studio and build using CMake.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,6 @@
 [submodule "ext_libs/eigen"]
 	path = ext_libs/eigen
 	url = https://gitlab.com/libeigen/eigen.git
+[submodule "ext_libs/vcpkg"]
+	path = ext_libs/vcpkg
+	url = ../../microsoft/vcpkg.git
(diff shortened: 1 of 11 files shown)


.gitmodules
```diff
@@ -16,3 +16,6 @@
 [submodule "ext_libs/eigen"]
 	path = ext_libs/eigen
 	url = https://gitlab.com/libeigen/eigen.git
+[submodule "ext_libs/vcpkg"]
+	path = ext_libs/vcpkg
+	url = ../../microsoft/vcpkg.git
```
peterychang, line 21: we should use the whole url instead of a relative path for it

jackgerrits: I think relative path resolves https vs ssh based on your clone

lalo (approved):

PR #4078, opened 2022-07-29
feat: [las] One pass SVD implementation

olgavrou:
One Pass SVD:

A * Omega

where A is K x F
Omega is F x p 

AOmega is K x p

then we run Eigen svd on AOmega
U is truncated to d (`--max_actions` param)

`p = d + 5`

Omega is created by tossing a coin (i.e. rademacher randomness) achieved by checking the parity of the `combined_index`

This implementation will not result in `U, S, V` that can reconstruct `A` in an accurate manner but we only really care about the invariants that are tested in the unit tests:

- do identical actions have identical representations in U (for non-zero singular values)
- does an action that is a linear combination of two other actions have a representation in U that is a linear combination of the other two action's representations in U


vowpalwabbit/core/src/reductions/cb/details/large_action_space.h
```diff
@@ -129,19 +144,21 @@ struct spanner_state
...
-  uint64_t _d;
-  uint64_t _seed;
   spanner_state _spanner_state;
   shrink_factor_config _shrink_factor_config;
   aatop_impl _aatop_impl;
   vanilla_rand_svd_impl _vanilla_rand_svd_impl;
   model_weight_rand_svd_impl _model_weight_rand_svd_impl;
+  one_pass_svd_impl _one_pass_svd_impl;
   Eigen::MatrixXf U;
+  Eigen::MatrixXf AOmega;
```
lalo, line 161: duplicate? theres one in onepass too

vowpalwabbit/core/src/reductions/cb/details/large_action_space.h
```diff
@@ -156,6 +173,7 @@ struct cb_explore_adf_large_action_space
   void learn(VW::LEARNER::multi_learner& base, multi_ex& examples);
 
   void randomized_SVD(const multi_ex& examples);
+  void generate_AOmega(const multi_ex& examples);
```
lalo, line 176: is this being used?

vowpalwabbit/core/src/reductions/cb/details/large_action_space.h
olgavrou: yup removing

vowpalwabbit/core/src/reductions/cb/details/large_action_space.h
olgavrou: no should be removed thanks

vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
```diff
@@ -215,6 +223,7 @@ VW::LEARNER::base_learner* VW::reductions::cb_explore_adf_large_action_space_set
                .help("Online explore-exploit for a contextual bandit problem with multiline action dependent features"))
       .add(make_option("aatop", aatop))
       .add(make_option("model_weight", model_weight_impl))
+      .add(make_option("one_pass", one_pass_svd_impl))
```
lalo, line 226: nit: use one_of({"model_weight", "one_pass", etc})

pmineiro (approved):
It ended up so claan.

lalo (approved):

PR #4087, opened 2022-08-02
feat: (experimental) Initial support of merging VW models 

jackgerrits:
This is a first draft of implementing model merging. 

Currently only supports GD. Other reductions will run but not include pieces that should have been merged. It uses the logic that was used during AllReduce to implement the merge operation.

Exposed via:
- `vw-merge` tool
- Python API 
- Java (Spark) API
- C++ API

See [wiki page here](https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Model-merging)


vowpalwabbit/model_merger/src/main.cc
```diff
@@ -0,0 +1,184 @@
+// Copyright (c) by respective owners including Yahoo!, Microsoft, and
```
olgavrou, line 1: can the merge functions be called not via the merger tool but during normal vw execution by mistake? any reduction could potentially call merge should add a warning comment or something?

vowpalwabbit/core/src/merge.cc
```diff
@@ -0,0 +1,165 @@
...
+  VW::config::cli_options_serializer serializer;
+  for (auto const& option : workspace.options->get_all_options())
+  {
+    if (workspace.options->was_supplied(option->m_name) && option->m_keep) { serializer.add(*option); }
+  }
+
+  return serializer.str();
+}
+
+void validate_compatability(const std::vector<const VW::workspace*>& workspaces_to_merge, VW::io::logger* logger)
```
olgavrou, line 27:
```suggestion
void validate_compatibility(const std::vector<const VW::workspace*>& workspaces_to_merge, VW::io::logger* logger)
```

vowpalwabbit/core/src/reductions/gd.cc
```diff
@@ -41,6 +43,59 @@ using namespace VW::config;
...
+
+  // Weights have already been reweighted, so just accumulate.
+  for (const auto& model_num : source)
+  {
+    const auto& this_source = model_num.get();
+    // Intentionally add irrespective of stride.
+    const auto full_weights_size = length << weights.stride_shift();
+    for (uint64_t i = 0; i < full_weights_size; i++)
+    {
+      // Should per_model_weighting be used? adaptive has already caused reweighting?
```
olgavrou, line 83: is this question resolved?

vowpalwabbit/core/src/merge.cc
```diff
@@ -0,0 +1,165 @@
...
+  auto* target_learner = destination_model->l;
+  while (target_learner != nullptr)
+  {
+    if (target_learner->has_merge())
+    {
+      std::vector<void*> reduction_data_per_model;
+      for (const auto& model : workspaces_to_merge)
+      {
+        auto* source_data = model->l->get_learner_by_name_prefix(target_learner->get_name());
+        reduction_data_per_model.push_back(source_data->get_internal_type_erased_data_pointer_test_use_only());
```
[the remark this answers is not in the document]
olgavrou: like have the learner and reductions call merge data function so that data is not exposed

python/pylibvw.cc
```diff
@@ -283,6 +285,14 @@ vw_ptr my_initialize_with_log(py::list args, py_log_wrapper_ptr py_log)
 
 vw_ptr my_initialize(py::list args) { return my_initialize_with_log(args, nullptr); }
 
+boost::shared_ptr<VW::workspace> merge_workspaces(py::list workspaces)
+{
+  std::vector<const VW::workspace*> const_workspaces;
+  for (size_t i = 0; i < py::len(workspaces); i++)
+  { const_workspaces.push_back(py::extract<VW::workspace*>(workspaces[i])); }
+  return boost::shared_ptr<VW::workspace>(VW::merge_models(const_workspaces).release());
```
olgavrou, line 293: you should be able to initialize the shared pointer with the unique ptr, should not need to use release?

vowpalwabbit/core/tests/merge_test.cc
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) by respective owners including Yahoo!, Microsoft, and
```
olgavrou, line 1: love the test coverage

python/tests/test_pyvw.py
```diff
@@ -701,3 +701,25 @@ def test_random_weights_seed():
 
     dummy_ex_str = " | foo=bar"
     assert a.predict(dummy_ex_str) != b.predict(dummy_ex_str)
+
+
+def test_merge_models():
+    model1 = vowpalwabbit.Workspace(quiet=True, noconstant=True)
```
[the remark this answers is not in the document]
jackgerrits: I was going to count the number of non-zero weights and so wanted to ignore the constant weight but ended up not doing it. It has no effect on the test now

vowpalwabbit/model_merger/src/main.cc
jackgerrits: A reduction is unlikely to have the arguments available to be able to call it

vowpalwabbit/core/src/merge.cc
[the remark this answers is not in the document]
jackgerrits: You can't because you require the `data` object of every learner and not just a single one.

vowpalwabbit/core/src/reductions/gd.cc
jackgerrits: It is not.

jackgerrits: Yep it is

olgavrou (approved):

PR #4091, opened 2022-08-03
fix: clear labels and predictions when finishing examples

bassmang:


olgavrou (approved):

lalo:
The perf test that would've catch this might be interesting to merge before this PR and then see the improvement get registered on the ci to validate this PR.

bassmang:
Decided to call clear before cache_label as to not risk performance hit

PR #4112, opened 2022-08-11
fix: [automl] clang-tidy error

bassmang:


lalo:
fixed in this https://github.com/VowpalWabbit/vowpal_wabbit/pull/4111 to avoid merge conflict

PR #4116, opened 2022-08-12
[las] separate spanners and remove aatop SVD impl

olgavrou:
mostly just moving code around

- separate spanners into two file impls
- templetize to activate them
- remove aatop SVD implementation as it is and is going to remain incomplete


vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
```diff
@@ -26,68 +26,116 @@ namespace VW
 {
 namespace cb_explore_adf
 {
-template <typename impl_detail>
-void cb_explore_adf_large_action_space<impl_detail>::predict(VW::LEARNER::multi_learner& base, multi_ex& examples)
+struct A_triplet_constructor
```
olgavrou, line 29: moved from aatop_impl.cc

lalo (approved):

PR #4120, opened 2022-08-17
feat: confidence sequences estimator

bassmang:


vowpalwabbit/core/include/vw/core/confidence_sequence.h
```diff
@@ -0,0 +1,97 @@
...
+      if (lo != 0.0)
+      {
+        this->partials[i] = lo;
+        ++i;
+      }
+      x = hi;
+    }
+    this->partials[i] = x;
+    this->partials.resize(i + 1);
+    return *this;
```
lalo, line 41: make a test similar to this one distributionally_robust_test.cc

vowpalwabbit/core/include/vw/core/confidence_sequence.h
```diff
@@ -0,0 +1,113 @@
...
+      double hi = x + y;
+      double lo = y - (hi - x);
+      if (lo != 0.0)
+      {
+        this->partials[i] = lo;
+        ++i;
+      }
+      x = hi;
+    }
+    this->partials[i] = x;
```
ataymano, line 42: isn't i out of range here?

vowpalwabbit/core/include/vw/core/confidence_sequence.h
bassmang: Added test

vowpalwabbit/core/include/vw/core/confidence_sequence.h
bassmang: Good catch, added loop to ensure it is not out of range

lalo (approved):

vowpalwabbit/core/include/vw/core/confidence_sequence.h
ataymano: doesn't swapping this line with resizing work here?

bassmang: Yes that should work, updated

PR #4121, opened 2022-08-17
refactor: [automl] iterator based config generators

lalo:
exclusion_config renamed to ns_based_config to be able to express both exclusion and inclusion.


vowpalwabbit/core/src/reductions/details/automl_impl.h
```diff
@@ -96,32 +98,59 @@ struct config_oracle
...
+  {
+    return Iterator(champ_interactions.size() + champ_exclusions.size());
+  }
 };
 struct champdupe_impl
 {
-  void gen_exclusion_configs(config_oracle<champdupe_impl>* co, const interaction_vec_t& champ_interactions,
-      std::vector<exclusion_config>& configs, const std::map<namespace_index, uint64_t>& ns_counter);
+  Iterator begin() { return Iterator(); }
+  Iterator end() { return Iterator(3); }
```
bassmang, line 153: why 3?

lalo: champdupe is mostly for testing, hardcoded the 3 to make sure that if we run --automl 4 theres a config for each spot. I didn't want it to be infinite or bigger since we can also use it to test the lease mechanism, and the repopulate functionality.

vowpalwabbit/core/src/reductions/details/automl_impl.h
```diff
@@ -96,32 +98,59 @@ struct config_oracle
...
 struct one_diff_impl
 {
-  void gen_exclusion_configs(config_oracle<one_diff_impl>* co, const interaction_vec_t& champ_interactions,
-      std::vector<exclusion_config>& configs, const std::map<namespace_index, uint64_t>& ns_counter);
+  void gen_exclusion_config_at(const std::string& interaction_type, const interaction_vec_t& champ_interactions,
+      const size_t num, exclusion_set_t::iterator& exclusion, exclusion_set_t& new_exclusions);
+  Iterator begin() { return Iterator(); }
+  Iterator end(const interaction_vec_t& champ_interactions, const exclusion_set_t& champ_exclusions)
+  {
+    return Iterator(champ_interactions.size() + champ_exclusions.size());
```
bassmang, line 147: I think this would fail with CCB since we check this `(is_allowed_to_remove(ns1) && is_allowed_to_remove(ns2))` so there could be less configs

lalo: This would only express the max amount of configs, there could be less also bc of dups. The dup handling happens later in insert_config. I think with CCB you end up generating dups that don't get added everytime you hit a ccb related ns.

bassmang: Are there no scenarios where we iterate through the whole iterator which would make this unsafe?

lalo: we're already doing this, as in before this change. We're just creating dups in the worst case.

bassmang (approved):

PR #4123, opened 2022-08-17
fix: [automl] generate challengers on first example

lalo:


bassmang (commented):
A couple of comments
1) I think we should get rid of `automl_state` since `Collecting` is no longer used anywhere since everything is `Experimenting`
2) `insert_qcolcol` should be named something like `insert_no_exclusions` or `insert_starting_configuration` since it's not qcolcol with cubics / could be something else

bassmang (approved):

PR #4124, opened 2022-08-17
refactor: use std::isfinite instead of within bounds of min and max float

jackgerrits:
Ran my own benchmark which showed either equivalent or 20% faster than existing code. Curious to see results on perf benchmark.


lalo (approved):

jackgerrits:
No discernable improvement or loss in perf but would prefer to not change the hot path.

PR #4126, opened 2022-08-19
ci: check all runtests models for forwards model compatibility 

olgavrou:
generate step:

- build py vw for master
- get all cli args and inputs from runtests
- generate models using all cli args and datafiles using py vw

model load step:
- install the latest py vw released version
- get all cli args and input from runtests
- load all generated models run all the datafiles through them using py vw

so we keep learning after we load a model in all cases and we check that nothing crashes


.github/workflows/forward_model_load_check.yml
```diff
@@ -74,9 +74,10 @@ jobs:
       - uses: actions/download-artifact@v1
         with:
           name: vw_generated_models
-          path: vw_generated_models
+          path: .vw_runtests_model_gen_working_dir/test_models
```
olgavrou, line 77: default path here is the working directory and tilde doesn't work in placing it in path, hence the `mv` below

test/run_tests_model_gen_and_load.py
```diff
@@ -0,0 +1,236 @@
...
+Most, if not all, runtests have the -d option specified so using the cli directly in the python interface will train a model using the file specified.
+
+After the models are generated we can run the same runtests that will load each model (--load_models) by using the same python interface,
+and that will result in the datafile being processed again
+
+The idea is that we can generate the models using one vw version and load the models in an older (or newer) vw version and check for forwards (backwards) model compatibility
+"""
+
+default_working_dir_name = ".vw_runtests_model_gen_working_dir"
+default_test_file = "core.vwtest.json"
```
jackgerrits, line 26: To ensure you can run this script from any directory you could use pathlib:

```suggestion
default_test_file = Path(__file__).resolve().parent / "core.vwtest.json"
```

test/run_tests_model_gen_and_load.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+def create_test_dir(test_id, input_files, test_base_dir, test_ref_dir):
```
jackgerrits, line 29: Consider adding types to functions in this file to help whoever looks at this file next

test/run_tests_model_gen_and_load.py
olgavrou: this path is resolved a bit later

test/run_tests_model_gen_and_load.py
```diff
@@ -0,0 +1,236 @@
...
+    for f in input_files:
+        file_to_copy = None
+        search_paths = [test_ref_dir / f]
```
jackgerrits, line 32: nit: To me search path implies directories to look in, rather than full paths to check. Consider either renaming this or appending `f` inside the search loop below

test/run_tests_model_gen_and_load.py
olgavrou: done

test/run_tests_model_gen_and_load.py
```diff
@@ -0,0 +1,236 @@
...
+
+        shutil.copy(str(file_to_copy), str(test_dest_file))
+
+
+def generate_model(test_id, command, working_dir, color_enum=Color):
+    command = command + " --quiet "
+    print(f"{color_enum.LIGHT_CYAN}id: {test_id}, command: {command}{color_enum.ENDC}")
+    vw = vowpalwabbit.Workspace(command)
+
+    vw.save(f"{working_dir}/model_{test_id}.vw")
```
jackgerrits, line 80: construct the path using pathlib then convert it. Not sure if `vw.save` supports `Path` input or not
```suggestion
    vw.save(str(working_dir / f"model_{test_id}.vw"))
```

test/run_tests_model_gen_and_load.py
```diff
@@ -0,0 +1,236 @@
...
+    vw.finish()
+
+
+def load_model(test_id, command, working_dir, color_enum=Color):
+    command = command + " --quiet "
+    command = command + f" -i {working_dir}/model_{test_id}.vw "
```
jackgerrits, line 91: Construct the path with pathlib

test/run_tests_model_gen_and_load.py
```diff
@@ -0,0 +1,236 @@
...
+
+    # link is changed in some reductions so it will clash with saved model
+    if "--link" in command:
+        command = re.sub("--link [:a-zA-Z0-9_.\\-/]*", "", command)
```
jackgerrits, line 95: Commands of the form `--link=...` need to be handled too

test/run_tests_model_gen_and_load.py
olgavrou: it is being stringified before it is passed into vw.save already i.e. `f"{...}"`

olgavrou: but sure

test/run_tests_model_gen_and_load.py
olgavrou: there currently aren't any in runtests but can add for safety

jackgerrits (approved):

PR #4133, opened 2022-08-26
feat: Performance benchmarks for .NET C# bindings

byronxu99:
- Implemented C# versions of existing C++ standalone benchmarks
- Added CI job to run them after Nuget package build


jackgerrits:
I think a merge conflict got accidentally committed

.github/workflows/build_nugets.yml
```diff
@@ -364,13 +364,92 @@ jobs:
...
+
+          cd BenchmarkDotNet.Artifacts/results
+          for file in *-report-github.md; do
+            benchmark="${file%-report-github.md}"
+            echo "***** $benchmark *****"
+            # Remove lines not part of table (not between |...|)
+            # Remove markdown formatting **
+            # Remove first 3 columns of table
+            cat "$file" | sed '/^|.*|\s*$/!d' | sed 's/\*\*//g' | sed 's/^\(|[^|]*\)\{3\}//'
+            echo -e '\n'
```
jackgerrits, line 449: Could you output this to a job summary so it is easier to find? https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/

.vscode/settings.json
```diff
@@ -1,10 +0,0 @@
-{
```
jackgerrits, line 1: I think we should keep this file as it provides a schema for the test files.

cs/common/netstandard/CMakeLists.txt
```diff
@@ -16,11 +16,11 @@ if (vw_DOTNET_USE_MSPROJECT)
   include_external_msproject(VowpalWabbit.Common ${CMAKE_CURRENT_SOURCE_DIR}/vw.common.csproj)
 else()
   add_custom_target(VowpalWabbit.Common ALL
-    COMMAND ${DOTNET_COMMAND} build ${CMAKE_CURRENT_SOURCE_DIR} -o $<TARGET_FILE_DIR:vw.net.native> -v m --nologo --no-dependencies /clp:NoSummary
+    COMMAND ${DOTNET_COMMAND} build ${CMAKE_CURRENT_SOURCE_DIR} -o $<TARGET_FILE_DIR:vw.net.native> --configuration "$<IF:$<CONFIG:Debug>,Debug,Release>" -v m --nologo --no-dependencies /clp:NoSummary
```
jackgerrits, line 19: Do we need to handle RelWithDebInfo and MinSizeRel?

nuget/dotnet/benchmarks/BenchmarkCbAdfLearn.cs
```diff
@@ -0,0 +1,43 @@
...
+using BenchmarkDotNet.Attributes;
+
+[Config(typeof(VWBenchmarkConfig))]
+public class BenchmarkCbAdfLearn
+{
+    [Benchmark]
+    [ArgumentsSource(nameof(ArgsGenerator))]
+    public void Benchmark(Args args)
+    {
+        args.vw.Learn(args.examples);
```
jackgerrits, line 11: Does finish example need to be called? I am not sure how this is setup, so just wondering

.vscode/settings.json
byronxu99: This file is in the `.gitignore`, so I don't think it should have been commited in the first place.

If you think we should have the file, then it would make sense to also update gitignore.

nuget/dotnet/benchmarks/BenchmarkCbAdfLearn.cs
byronxu99: It's called in the native DLL wrapper here: https://github.com/byronxu99/vowpal_wabbit/blob/dev/DotNetBenchmarks/cs/vw.net.native/vw.net.workspace.cc#L289

which is in turn called by C# code here: https://github.com/byronxu99/vowpal_wabbit/blob/dev/DotNetBenchmarks/cs/vw.net/VowpalWabbit.cs#L630

.github/workflows/build_nugets.yml
byronxu99: Wow, that's a cool feature that I didn't know of!

.vscode/settings.json
jackgerrits: It's in the gitignore since vscode frequently edits the file. So we don't want to commit local user edits that vscode automatically does.

nuget/dotnet/benchmarks/BenchmarkCbAdfLearn.cs
jackgerrits: This means that the managed example object is technically invalidated after making this call. It would be quite easy for a user to think they can reuse the example object. Feels a bit scary

.vscode/settings.json
byronxu99: Makes sense. I've added it back.

cs/common/netstandard/CMakeLists.txt
byronxu99: I don't think dotnet supports CMake's idea of RelWithDebInfo and MinSizeRel. The way it is right now, both would be handled as Release if you were to use them. I think that's reasonable, as they are both variants of release builds.

We only build the Nuget package with the Release configuration, anyways.

nuget/dotnet/benchmarks/BenchmarkCbAdfLearn.cs
byronxu99: Looks like this is the same as how the old .NET Framework bindings did it: https://github.com/byronxu99/vowpal_wabbit/blob/0e503b39662ca14b64e3c887cb3799b6f50cf91e/cs/cli/vowpalwabbit.cpp#L212

I do agree that it seems a bit scary.

lokitoth: Yeah, this is an unfortunate legacy of the code as it was. What I would like to do going forward is to formalize the proposal for the c interface that you started @jackgerrits, and use a C#-ified layer as the new canonical interface of VW.Net.

.github/workflows/build_nugets.yml
rajan-chari: That's cool.

PR #4134, opened 2022-08-26
feat: Semantic Versioning

byronxu99:
Build nuget packages with SemVer version numbers generated from Git tags. Each new package will have a unique and increasing version number.


CMakeLists.txt
```diff
@@ -16,12 +16,12 @@ endif()
 # Right now this is done for the Python CIs.
 list(APPEND VCPKG_FEATURE_FLAGS "versions")
 
-# Read version into variable
-file(READ version.txt PACKAGE_VERSION)
+# Get version number
+execute_process(COMMAND "python3" "${CMAKE_CURRENT_LIST_DIR}/utl/version_number.py" OUTPUT_VARIABLE PACKAGE_VERSION)
```
jackgerrits, line 20: I thought this versioning scheme was just being used for nugets in https://github.com/VowpalWabbit/reinforcement_learning? I am quite hesitant to make a change to the version number structure in VW itself. 

Additionally, I do not think we should require `python3` to build VW. This is a potentially hefty dependency that complicates all of the packaging scenarios

byronxu99: Good point. We can't assume that we always have git metadata, like in cases such as building from tarball.

My motivation here was to have exactly one place where the version is stored. But it seems that we're stuck with both git tags for the Nuget and `version.txt` for other things.

jackgerrits (approved):

PR #4137, opened 2022-09-01
feat: simple thread pool

olgavrou:
introduced to be used in LAS

will probably evolve to include some sort of thread pool interface

when there is no work to be done threads wait on a conditional variable

if thread pool is initialized with zero size then the tasks are just executed sequentially on the same thread that called the thread pool


vowpalwabbit/core/include/vw/core/thread_pool.h
```diff
@@ -0,0 +1,187 @@
...
+#include <queue>
+#include <thread>
+
+// thread pool based off of the simple pool implementation from ccia:
+// https://github.com/anthonywilliams/ccia_code_samples
+
+namespace VW
+{
+template <typename T>
+class thread_safe_queue
```
jackgerrits, line 23: This is really similar to `queue.h` can we reuse it?

vowpalwabbit/core/include/vw/core/thread_pool.h
```diff
@@ -0,0 +1,187 @@
...
+    std::future<ResultType> result(task.get_future());
+
+    if (_threads.size() == 0)
+    {
+      task();
+      return result;
+    }
+
+    {
+      std::unique_lock<std::mutex> l(_mut);
```
zwd-ms, line 176: `push` has lock already?

vowpalwabbit/core/include/vw/core/thread_pool.h
```diff
@@ -0,0 +1,187 @@
...
+
+  // Enqueues a func task in the work queue for worker threads to execute.
+  template <typename F, typename... Args>
+  std::future<typename std::result_of<F(Args...)>::type> submit(F&& func, Args&&... args)
+  {
+    using ResultType = typename std::result_of<F(Args...)>::type;
+    std::packaged_task<ResultType()> task(std::bind(std::forward<F>(func), std::forward<Args>(args)...));
+    std::future<ResultType> result(task.get_future());
+
+    if (_threads.size() == 0)
```
zwd-ms, line 169: what's the use case for initializing a thread pool with zero threads?

vowpalwabbit/core/include/vw/core/thread_pool.h
```diff
@@ -0,0 +1,187 @@
...
+
+    function_wrapper(function_wrapper&& other) : _impl{std::move(other._impl)} {}
+
+    function_wrapper(const function_wrapper&) = delete;
+
+    function_wrapper(function_wrapper&) = delete;
+
+    function_wrapper& operator=(const function_wrapper&) = delete;
+
+    void operator()() { _impl->call(); }
```
jackgerrits, line 93: It looks like this means the thread pool cant be used for operations which return values?

vowpalwabbit/core/include/vw/core/thread_pool.h
```diff
@@ -0,0 +1,187 @@
...
+    {
+    }
+
+    function_wrapper() = default;
+
+    function_wrapper(function_wrapper&& other) : _impl{std::move(other._impl)} {}
+
+    function_wrapper(const function_wrapper&) = delete;
+
+    function_wrapper(function_wrapper&) = delete;
```
jackgerrits, line 89: Looks like copy constructor is done twice and move assignment was missed

vowpalwabbit/core/include/vw/core/thread_pool.h
```diff
@@ -0,0 +1,187 @@
...
+        while (!_task_queue.try_pop(task) && !_done) { _cv.wait(l); }
+      }
+      if (_done) { break; }
+      task();
+    }
+  }
+
+  std::atomic_bool _done;
+  std::condition_variable _cv;
+  mutable std::mutex _mut;
```
jackgerrits, line 134: I dont think this needs to be mutable

vowpalwabbit/core/include/vw/core/thread_pool.h
```diff
@@ -0,0 +1,187 @@
...
+  };
+
+  void worker()
+  {
+    while (!_done)
+    {
+      function_wrapper task;
+      {
+        std::unique_lock<std::mutex> l(_mut);
+        while (!_task_queue.try_pop(task) && !_done) { _cv.wait(l); }
```
zwd-ms, line 125: can we move the cv to `thread_safe_queue` and use a `wait_and_pop` instead?
I think that'll make the code logic cleaner

vowpalwabbit/core/include/vw/core/thread_pool.h
olgavrou: that one is a pointer queue though

vowpalwabbit/core/include/vw/core/thread_pool.h
olgavrou: maybe you don't want to branch your code based on zero threads vs more than one thread

plus this ensures that you don't just hang if you initialize with zero by mistake

jackgerrits: In my opinion initializing with zero threads should be an error. If the user doesn't want threading they shouldnt use a thread pool. It's not up to the thread pool to implement that

vowpalwabbit/core/include/vw/core/thread_pool.h
jackgerrits: Pointers are movable so usage of that type can probably be replaced to one common implementation that satisfies both

vowpalwabbit/core/include/vw/core/thread_pool.h
olgavrou: I like to think as scaling naturally with the available resources without having to have a different code path (similar to std async)

vowpalwabbit/core/include/vw/core/thread_pool.h
olgavrou: who owns the objects though?

olgavrou: might need to change its name from ptr_queue

vowpalwabbit/core/include/vw/core/thread_pool.h
```diff
@@ -0,0 +1,187 @@
...
+
+  ~thread_pool()
+  {
+    _done = true;
+    _cv.notify_all();
+  }
+
+  // Enqueues a func task in the work queue for worker threads to execute.
+  template <typename F, typename... Args>
+  std::future<typename std::result_of<F(Args...)>::type> submit(F&& func, Args&&... args)
```
jackgerrits, line 163: `std::result_of` was deprecated in 17 and removed in 20 so we should not use it.

We can instead use a trailing return type and decltype of funcs invocation. See here: https://github.com/mtrebi/thread-pool/blob/master/include/ThreadPool.h#L79

test/unit_test/thread_pool_test.cc
```diff
@@ -0,0 +1,115 @@
+// Copyright (c) by respective owners including Yahoo!, Microsoft, and
+// individual contributors. All rights reserved. Released under a BSD (revised)
+// license as described in the file LICENSE.
+
+#include "vw/core/thread_pool.h"
```
jackgerrits, line 5: Please add tests to core unit tests and not the boost project

vowpalwabbit/core/include/vw/core/thread_pool.h
olgavrou: that's true it isn't used in any read only functions

vowpalwabbit/core/include/vw/core/thread_pool.h
jackgerrits: I dont think pointers need special treatment. We should be able to reuse the same implementation

vowpalwabbit/core/include/vw/core/thread_pool.h
olgavrou: We can see what it looks like after I try out @jackgerrits suggestion to merge the existing queue's but I'll probably make it wait on `_task_queue.empty()` in that case. This way it is a good separation of concerns, the queue should not care when the thread pool is done

olgavrou: Hm that queue has set_done actually now that i think about it so might be doable

jackgerrits (approved):

zwd-ms (approved):

PR #4141, opened 2022-09-02
feat: [LAS] use thread pool in LAS one_pass implementation

olgavrou:


vowpalwabbit/core/src/reductions/cb/details/large_action/one_pass_svd_impl.cc
```diff
@@ -71,8 +69,18 @@ void one_pass_svd_impl::generate_AOmega(const multi_ex& examples, const std::vec
...
+
+  uint64_t row_index = 0;
+  for (auto* ex : examples)
+  {
+    _futures.emplace_back(_thread_pool.submit(
+        calculate_aomega_row, row_index, p, _all, _seed, ex, std::ref(AOmega), std ::ref(shrink_factors)));
     row_index++;
   }
+
+  for (auto& ft : _futures) { ft.get(); }
```
olgavrou, line 82: `calculate_aomega_row` could throw so we could wrap `ft.get()` in a try/catch and rethrow here (not sure how we would recover)

ZhangWangda (approved):

zwd-ms (approved):

vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
```diff
@@ -314,6 +315,10 @@ VW::LEARNER::base_learner* VW::reductions::cb_explore_adf_large_action_space_set
       .add(make_option("model_weight", model_weight_impl))
       .add(make_option("vanilla", use_vanilla_impl))
       .add(make_option("full_spanner", full_spanner))
+      .add(make_option("thread_pool_size", thread_pool_size)
+               .default_value(0)
```
ZhangWangda, line 319: if we want multithreading by default, perhaps use `std::thread::hardware_concurrency`

PR #4143, opened 2022-09-06
build: fix windows.h GetObject issue

jackgerrits:


byronxu99 (approved):
I remember once encountering this issue due to the order that header files were included. The error message made it really confusing to debug.

Good to have a proper fix for it.

PR #4144, opened 2022-09-08
refactor: [automl] allow_override for selected options

bassmang:


vowpalwabbit/core/src/reductions/automl.cc
```diff
@@ -185,6 +188,7 @@ VW::LEARNER::base_learner* VW::reductions::automl_setup(VW::setup_base_i& stack_
                .experimental())
       .add(make_option("oracle_type", oracle_type)
```
lalo, line 189: this case would be good to test, you would "depends_on" mechanism in https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/test/core.vwtest.json

bassmang: Added two tests which check all the overrides

lalo (approved):

PR #4146, opened 2022-09-12
test: [LAS ] add spanner tests and test file separation

olgavrou:
new tests in cb_las_spanner:

`check_spanner_rejects_same_actions`
`check_spanner_rejects_actions_that_are_linear_combinations_of_other_actions`


vowpalwabbit/core/src/reductions/cb/details/large_action/one_pass_svd_impl.cc
```diff
@@ -47,7 +47,8 @@ struct AO_triplet_constructor
 void one_pass_svd_impl::generate_AOmega(const multi_ex& examples, const std::vector<float>& shrink_factors)
 {
   auto num_actions = examples[0]->pred.a_s.size();
-  auto p = std::min(num_actions, _d + 5);
+  // one pass SVD is going to be less accurate than two pass SVD so we need to over-sample a bit more
+  auto p = std::min(num_actions, _d + 10);
```
zwd-ms, line 51: how do we know to use 5 or 10? shall this be a parameter?

zwd-ms (approved):

vowpalwabbit/core/src/reductions/cb/details/large_action/one_pass_svd_impl.cc
olgavrou: trial and error, we were between 5 and 10 initially, 5 made the unit test fail when switched from vanilla to one_pass but 10 was ok. You can essentially affect this number by tuning `d` so it seems like an overkill to tune this too

PR #4147, opened 2022-09-14
test: [LAS] calculate num of non-degenerate singular values

olgavrou:
and use it to make some tests more precise

### DIFF
--- a/test/unit_test/cb_las_spanner_test.cc
+++ b/test/unit_test/cb_las_spanner_test.cc
@@ -117,9 +117,7 @@ BOOST_AUTO_TEST_CASE(check_spanner_results_squarecb)
     std::vector<std::string> e_r;
     vw.l->get_enabled_reductions(e_r);
     if (std::find(e_r.begin(), e_r.end(), "cb_explore_adf_large_action_space") == e_r.end())
-    {
-      BOOST_FAIL("cb_explore_adf_large_action_space not found in enabled reductions");
-    }
+    { BOOST_FAIL("cb_explore_adf_large_action_space not found in enabled reductions"); }
 
     VW::LEARNER::multi_learner* learner =
         as_multiline(vw.l->get_learner_by_name_prefix("cb_explore_adf_large_action_space"));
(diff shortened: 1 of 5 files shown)


vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
```diff
@@ -133,6 +133,29 @@ void cb_explore_adf_large_action_space<randomized_svd_impl, spanner_impl>::rando
   _impl.run(examples, shrink_factors, U, _S, _V);
 }
 
+template <typename randomized_svd_impl, typename spanner_impl>
+size_t cb_explore_adf_large_action_space<randomized_svd_impl, spanner_impl>::number_of_non_degenerate_singular_values()
```
olgavrou, line 137: this could potentially be used to also return even less actions than specified by d

test/unit_test/cb_las_spanner_test.cc
```diff
@@ -601,26 +601,36 @@ BOOST_AUTO_TEST_CASE(check_spanner_rejects_same_actions)
 
 BOOST_AUTO_TEST_CASE(check_spanner_rejects_actions_that_are_linear_combinations_of_other_actions)
```
zwd-ms, line 602: Test name seems inconsistent with result. Is the expected behavior "reject" or "accept" the linear combination action?

test/unit_test/cb_las_spanner_test.cc
```diff
@@ -680,12 +687,10 @@ BOOST_AUTO_TEST_CASE(check_spanner_rejects_actions_that_are_linear_combinations_
       size_t encounters = 0;
       for (auto& a_s : preds)
       {
-        if (a_s.action == 1 && a_s.score != 0.f) { encounters++; }
-        if (a_s.action == 2 && a_s.score != 0.f) { encounters++; }
         if (a_s.action == 3 && a_s.score != 0.f) { encounters++; }
       }
 
-      BOOST_CHECK_EQUAL(encounters, 2);
+      BOOST_CHECK_EQUAL(encounters, 1);
```
zwd-ms, line 693: This is checking the fourth action (action_4 in the comment above) has non-zero score, right?
do we need to check other actions probabilities?

zwd-ms (approved):

test/unit_test/cb_las_spanner_test.cc
olgavrou: made it more generic

PR #4150, opened 2022-09-15
fix: remove problematic tag swap

tyclintw:
The first tags in the audit output is missing due to an incorrect swap.  Remove this swap as it is incorrect.


tyclintw:
Discussed offline with VW developers and we've decided this is not the correct approach. Abandoning.

PR #4162, opened 2022-09-22
fix: remove experimental and fix up model version test

bassmang:


test/run_tests_model_gen_and_load.py
```diff
@@ -134,15 +132,28 @@ def get_tests(
...
     for test in tests:
+        vw = vowpalwabbit.Workspace(test.command_line, quiet=True)
+        skip_cmd = False
+        for groups in vw.get_config().values():
+            for group in groups:
+                for opt in group[1]:
+                    if opt.value_supplied and (
+                        opt.experimental
+                        or opt.name == "bfgs"
+                        or (opt.name == "initial_regressor" and opt.value != [])
```
lalo, line 143: where is the --flatbuffer case covered? is flatbuffer tagged as experimental too?

bassmang: Yes flatbuffer is experimental, it actually needs to be checked separately because it segfaults when loading the model

olgavrou: we don't need to check flatbuffer files really, they repeat existing tests and the ones in the run tests are for input/output and don't add anything model-wise

bassmang: Yea I meant checked as in removed from the test set

.github/workflows/forward_model_load_check.yml
```diff
@@ -30,6 +30,12 @@ jobs:
       - name: Build wheel
         shell: bash
         run: |
+          git clone --recursive https://github.com/VowpalWabbit/vowpal_wabbit.git
```
lalo, line 33: you could merge this ci into this other one https://github.com/VowpalWabbit/vowpal_wabbit/blob/c7c14efc8e0e62a45ebc2917720823701a3ada44/.github/workflows/python_wheels.yml#L53

and re-use the built wheel

bassmang: This CI isn't building the most recent master's branch. I got it working here, think it makes sense to keep them separate

lalo (approved):

PR #4167, opened 2022-09-27
fix: test 67 windows failure

### LINKED ISSUE #3843: Test 67 fails on Windows
```
python .\test\run_tests.py --ignore_dirty --skip_spanning_tree_tests --vw_bin_path ./build/vowpalwabbit/Release/vw.exe -f -E 0.01 -t 67
```

The first difference of file paths can be resolved by simply removing the path special case for Windows. Then you are left with the following diff for test 67:

image

bassmang:
resolves #3843

### DIFF
--- a/vowpalwabbit/core/src/parse_args.cc
+++ b/vowpalwabbit/core/src/parse_args.cc
@@ -147,9 +147,7 @@ void parse_dictionary_argument(VW::workspace& all, const std::string& str)
   if (!all.quiet)
   {
     std::string out_file_name = file_name;
-#ifdef _WIN32
     std::replace(out_file_name.begin(), out_file_name.end(), '\\', '/');
-#endif
     *(all.trace_message) << "scanned dictionary '" << s << "' from '" << out_file_name << "', hash=" << std::hex
                          << fd_hash << std::dec << endl;
   }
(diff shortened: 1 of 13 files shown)


vowpalwabbit/core/src/parse_args.cc
```diff
@@ -67,6 +67,11 @@ uint64_t hash_file_contents(VW::io::reader* f)
   while (true)
   {
     ssize_t n = f->read(buf, 1024);
+#ifdef _WIN32
+    char* rem_buf;
+    if ((rem_buf = std::remove(std::begin(buf), std::end(buf), '\r')) == nullptr) { THROW("error: invalid buffer"); }
+    n -= 1024 - (rem_buf - buf);
```
jackgerrits, line 73: nit: extract 1024 into a constant that is used in both places

jackgerrits, line 73:
```suggestion
    n -= (1024 - std::distance(std::begin(buf), rem_buf));
```

jackgerrits (dismissed):

jackgerrits:
Now that this is fixed, can you unskip the test? https://github.com/VowpalWabbit/vowpal_wabbit/pull/3842

jackgerrits (approved):

PR #4172, opened 2022-09-28
ci: check model weights for gd-based tests for forward and backward compat

bassmang:


test/run_tests_model_gen_and_load.py
```diff
@@ -68,15 +68,23 @@ def create_test_dir(
...
+def generate_model_and_weights(
     test_id: int,
     command: str,
     working_dir: Path,
     color_enum: Type[Union[Color, NoColor]] = Color,
 ) -> None:
     print(f"{color_enum.LIGHT_CYAN}id: {test_id}, command: {command}{color_enum.ENDC}")
     vw = vowpalwabbit.Workspace(command, quiet=True)
-
+    weights_dir = working_dir / "../test_weights"
```
lalo, line 79: why is test_weights above working_dir?

bassmang: working_dir here the model directory, so the test_weights directory lives next to it

test/run_tests_model_gen_and_load.py
```diff
@@ -68,15 +68,23 @@ def create_test_dir(
...
+    weights_dir.mkdir(parents=True, exist_ok=True)
+    with open(weights_dir / f"weights_{test_id}.json", "w") as weights_file:
+        try:
+            weights_file.write(vw.json_weights())
+        except:
+            print(
+                f"{color_enum.LIGHT_PURPLE}Weights could not be loaded as base learner is not GD"
```
olgavrou, line 86: nit:
```suggestion
                f"{color_enum.LIGHT_PURPLE}Weights could not be generated as base learner is not GD"
```

olgavrou (approved):

test/run_tests_model_gen_and_load.py
peterychang: I think it would be better to make the parent directory the working_dir. That way you don't need to go out of your directory scope for any of the work here.

Another option would be to move the model_dir up a level, and make the entire structure look like
```
root_dir
  + outputs
    + model_dir
    + test_weights
```

Then outputs would be the working_dir in this case

bassmang: updated to this format

lokitoth (approved):
Looks good. One minor thing: If possible, please memorialize our logic for checking the weights in the PR description, so it does not get lost over time

PR #4183, opened 2022-10-04
ci: settle on consistent style and add warnings for violation

jackgerrits:
I would like to propose we change the preferred type name style and then eventually enforce this in CI.

I propose we change from the most common style of `snake_case` for type names to `PascalCase`, and enforce `snake_case` for namespaces. Once the changes are made we can promote this clang-tidy warning into an error.

**NOTE:** functions and variables remain as `snake_case`

The primary motivator of changing types from `snake_case` to `PascalCase` is it allows variables and types to have separate style. Often it makes sense to name a variable similar to the type. e.g. `vw::Example example;` vs `vw::example example;`

Additionally, requires enum values and macros be `UPPER_CASE`.

See here for a report detailing changes required so we adhere to these checks:
https://gist.github.com/jackgerrits/c8da9db95457c372ad72c4fd72aff4b7#file-report-md

Resolution:
After discussion we will keep using the existing default of `snake_case` because there was opposition to change. Enforcing consistency will go ahead as planned.


olgavrou (approved):
I like this as long as we don't introduce a breaking change in between minor releases (discussed offline, breaking change can be pushed back with aliasing)

jackgerrits:
100% agree on avoiding breaking changes here. Types which are considered part of the public API surface will have the necessary `using x_old = x_new;` put in place with a deprecated marker to avoid breaking changes and make it easier to migrate.

bassmang (approved):

PR #4203, opened 2022-10-07
refactor: [versioning] split options_config/setup

bassmang:


vowpalwabbit/core/src/reductions/cb/cb_algs.cc
```diff
@@ -134,96 +134,110 @@ void eval_finish_example(VW::workspace& all, cb& c, VW::example& ec)
   output_example(all, c, ec, ec.l.cb_eval.event);
   VW::finish_example(all, ec);
 }
-}  // namespace
```
rajan-chari, line 137: Nit:  // Struct to store options for v1 of cb_algs
Suggestion for convention for these structs?
for example options_<reduction>_v<version>?  so struct options_cb_algs_v1 ?

vowpalwabbit/core/src/reductions/cb/cb_algs.cc
```diff
@@ -134,96 +134,110 @@ void eval_finish_example(VW::workspace& all, cb& c, VW::example& ec)
...
+  if (cb_algs_opts == nullptr) { return nullptr; }
+  auto cb_algs_data = VW::make_unique<cb>(all.logger);
+  cb_algs_data->cbcs.num_actions = cb_algs_opts->num_actions;
+
+  cb_to_cs& c = cb_algs_data->cbcs;
 
   size_t problem_multiplier = 2;  // default for DR
-  c.cb_type = VW::cb_type_from_string(type_string);
+  c.cb_type = VW::cb_type_from_string(cb_algs_opts->type_string);
   switch (c.cb_type)
```
rajan-chari, line 200: Wondering if there should be a cb_type(not serialized?) and problem_multiplier(not serialized?) in cb_alg_options and the error checking below should live in get_cb_algs_options_instance ?

vowpalwabbit/core/src/reductions/cb/cb_explore.cc
```diff
@@ -362,61 +375,86 @@ base_learner* VW::reductions::cb_explore_setup(VW::setup_base_i& stack_builder)
...
+  cb_explore_data->epsilon = cb_explore_opts->epsilon;
+  cb_explore_data->bag_size = cb_explore_opts->bag_size;
+  cb_explore_data->cover_size = cb_explore_opts->cover_size;
+  cb_explore_data->nounif = cb_explore_opts->nounif;
+  cb_explore_data->psi = cb_explore_opts->psi;
+
+  cb_explore_data->_random_state = all.get_random_state();
+  uint32_t num_actions = cb_explore_data->cbcs.num_actions;
+
+  if (cb_explore_data->epsilon < 0.0 || cb_explore_data->epsilon > 1.0)
```
rajan-chari, line 414: Should this be in get_cb_explore_options_instance?

vowpalwabbit/core/src/reductions/cb/cb_explore.cc
```diff
@@ -362,61 +375,86 @@ base_learner* VW::reductions::cb_explore_setup(VW::setup_base_i& stack_builder)
...
-  data->cbcs.cb_type = VW::cb_type_t::dr;
-  data->model_file_version = all.model_file_ver;
+base_learner* VW::reductions::cb_explore_setup(VW::setup_base_i& stack_builder)
+{
+  options_i& options = *stack_builder.get_options();
+  VW::workspace& all = *stack_builder.get_all_pointer();
+  auto cb_explore_opts = CB_EXPLORE::get_cb_explore_options_instance(all, options);
+  if (cb_explore_opts == nullptr) { return nullptr; }
+  auto cb_explore_data = VW::make_unique<cb_explore>(all.logger);
+  cb_explore_data->cbcs.num_actions = cb_explore_opts->num_actions;
```
rajan-chari, line 403: Wondering if cb_explore_data should have cb_explore_opts eventually... although storing it as a pointer would mean an extra indirection everytime the variable is accessed...

vowpalwabbit/core/src/reductions/cb/cb_explore.cc
```diff
@@ -329,31 +329,44 @@ void generic_output_example(VW::workspace& all, float loss, VW::example& ec, CB:
   print_update_cb_explore(all, CB::is_test_label(ld), ec, sso);
 }
 
-}  // namespace CB_EXPLORE
+struct cb_explore_options_instance_v1
```
rajan-chari, line 332: Nit: Naming convention mentioned in cb_algs.cc ?

vowpalwabbit/core/src/reductions/cb/cb_to_cb_adf.cc
```diff
@@ -102,73 +102,71 @@ void finish_example(VW::workspace& all, cb_to_cb_adf& c, VW::example& ec)
...
-    Purpose: run before cb, cb_explore, cbify and cb_adf related reductions
-    This will 'translate' cb (non adf) input commands into their cb_adf counterparts
-    Except when:
-        - the model file loaded is from a version older than and not including 8.11.0
-        - user bypasses this translation step using '--cb_force_legacy'
 
-    Related files: cb_algs.cc, cb_explore.cc, cbify.cc
-*/
-VW::LEARNER::base_learner* VW::reductions::cb_to_cb_adf_setup(VW::setup_base_i& stack_builder)
+struct cb_to_cb_adf_options_instance_v1
```
rajan-chari, line 106: Nit: naming convention mentioned in cb_algs

vowpalwabbit/core/src/reductions/cb/cb_to_cb_adf.cc
```diff
@@ -178,36 +176,66 @@ VW::LEARNER::base_learner* VW::reductions::cb_to_cb_adf_setup(VW::setup_base_i&
...
+    Related files: cb_algs.cc, cb_explore.cc, cbify.cc
+*/
+VW::LEARNER::base_learner* VW::reductions::cb_to_cb_adf_setup(VW::setup_base_i& stack_builder)
+{
+  options_i& options = *stack_builder.get_options();
+  VW::workspace& all = *stack_builder.get_all_pointer();
+  auto cb_to_cb_adf_opts = get_cb_to_cb_adf_options_instance(all, options);
+  if (cb_to_cb_adf_opts == nullptr) { return nullptr; }
+
+  if (cb_to_cb_adf_opts->cb_to_cbadf_supplied)
```
rajan-chari, line 215: should this live in get___options_instance()?

vowpalwabbit/core/src/reductions/cb/cb_to_cb_adf.cc
```diff
@@ -178,36 +176,66 @@ VW::LEARNER::base_learner* VW::reductions::cb_to_cb_adf_setup(VW::setup_base_i&
...
 
   auto data = VW::make_unique<cb_to_cb_adf>();
-  data->explore_mode = override_cb_explore;
+  data->explore_mode = cb_to_cb_adf_opts->override_cb_explore;
   data->weights = &(all.weights);
 
   multi_learner* base = as_multiline(stack_builder.setup_base_learner());
 
-  if (num_actions <= 0) { THROW("cb num actions must be positive"); }
+  if (cb_to_cb_adf_opts->num_actions <= 0) { THROW("cb num actions must be positive"); }
```
rajan-chari, line 235: should this live in get____options_instance()

vowpalwabbit/core/src/reductions/cb/cb_algs.cc
bassmang: `cb_algs_options_instance_v1` this is the current convention I have been using. I agree it could be shorter

vowpalwabbit/core/src/reductions/cb/cb_algs.cc
bassmang: moved out most of this other than the logging

vowpalwabbit/core/src/reductions/cb/cb_explore.cc
bassmang: yea good call

vowpalwabbit/core/src/reductions/cb/cb_explore.cc
bassmang: thats doable, but goes against the general design of applying the options config to the data struct in setup

vowpalwabbit/core/src/reductions/cb/cb_to_cb_adf.cc
bassmang: makes sense, moved

vowpalwabbit/core/src/reductions/cb/cb_to_cb_adf.cc
bassmang: moved

vowpalwabbit/core/src/reductions/cb/cb_algs.cc
```diff
@@ -135,96 +135,113 @@ void eval_finish_example(VW::workspace& all, cb& c, VW::example& ec)
...
+struct options_cb_algs_v1
 {
-  options_i& options = *stack_builder.get_options();
-  VW::workspace& all = *stack_builder.get_all_pointer();
-  auto data = VW::make_unique<cb>(all.logger);
+  uint32_t num_actions;
   std::string type_string = "dr";
   bool eval = false;
   bool force_legacy = true;
+  size_t problem_multiplier = 2;  // default for DR
```
jackgerrits, line 145: Problem multiplier is a result of other options and not something actually configurable by a user. Based on that it seems like it shouldn't be in this struct?

vowpalwabbit/core/src/reductions/cb/cb_explore.cc
jackgerrits: We should not worry about a pointer indirection in code of this kind until profiling suggests it is an issue

vowpalwabbit/core/src/reductions/cb/cb_algs.cc
bassmang: fair point, moved the problem_multiplier logic to setup

vowpalwabbit/core/src/reductions/cb/cb_explore.cc
```diff
@@ -366,61 +379,84 @@ base_learner* VW::reductions::cb_explore_setup(VW::setup_base_i& stack_builder)
...
     // none of the relevant options are set, default to cb
     std::stringstream ss;
-    ss << data->cbcs.num_actions;
+    ss << cb_explore_opts->num_actions;
     options.insert("cb", ss.str());
   }
+  cb_explore_opts->cover_supplied = options.was_supplied("cover");
+  cb_explore_opts->epsilon_supplied = options.was_supplied("epsilon");
+  cb_explore_opts->bag_supplied = options.was_supplied("bag");
+  cb_explore_opts->first_supplied = options.was_supplied("first");
```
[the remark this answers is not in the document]
bassmang: that's true but they're still relevant to this reduction given the 4 unique predict_or_learn functions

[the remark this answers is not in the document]
jackgerrits: With the split apart options functions - is the idea that the function should be independent? Or does it still follow the current sequence of operations? It feels like if possible it should be defined such that it is independent and callable on its own. If thats the case, these fields probably should not exist in that configuration object. It feels a bit too complex for me to clearly work through in a comment though

vowpalwabbit/core/src/reductions/cb/cb_explore.cc
```diff
@@ -333,31 +333,44 @@ void generic_output_example(VW::workspace& all, float loss, VW::example& ec, CB:
...
+struct options_cb_explore_v1
+{
+  uint32_t num_actions;
+  uint64_t tau;
+  float epsilon;
+  uint64_t bag_size;
+  uint64_t cover_size;
+  bool nounif;
+  float psi;
+  bool cover_supplied;
```
peterychang, line 345: Since there's only one valid configuration for these options, I suggest an enum class

```
enum class cb_explore_types
{
  GREEDY,
  COVER,
  COVER_WITH_EPSILON,
  BAG,
  FIRST
}
```

That will programmatically restrict the reduction to its only valid options (default will be GREEDY, then overwrite with something else if necessary)

rajan-chari: When serialized, enums cause an unusual problem.  In most serialize/deserialize protocols, it's not possible to deserialize a valid enum when new items are added, even it is to check for error.  I started out defining enums for serializable fields and then had to switch to integers for this reason.  

Defining enums and setting it to a serializable int is ok though.

vowpalwabbit/core/include/vw/core/reductions/slates.h
```diff
@@ -16,7 +16,7 @@ namespace VW
 {
 namespace reductions
 {
-class slates_data
+class slates_data_obj
```
olgavrou, line 19: I'm not crazy about postfixing class names with `obj` I find it redundant and misleading (`slate_reduction` is the object `slates_data_obj` is the type)

bassmang: yea `slates_reduction` is fine, I have been renaming the data instance in setup to be `slate_data` which was causing the issue

jackgerrits: We should prefer naming for the type and if the object has to be named something weird then so be it. One of the unfortunate outcomes of using snake_case for type names...

bassmang:
NOTE: Revisit as versioning progresses

PR #4207, opened 2022-10-08
refactor: use stream insertion instead of fmtlib specialization

jackgerrits:
fmtlib automatically uses operator<< for custom types. We should provide those when necessary instead of adding formatter specialization because it does not add an unnecessary dependency and they are more generally usable.

Small fix in automl - anonymous namespaces should not be used in headers.
vw_string_view_fmt.h is no longer needed as operator<< is provided.


jackgerrits:
According to the docs this is opt in, but it seems to work by default for me locally https://fmt.dev/latest/api.html#std-ostream-support

PR #4211, opened 2022-10-11
style: fix some more style issues

jackgerrits:


library/libsearch.h
```diff
@@ -15,87 +15,88 @@ license as described in the file LICENSE.
 #  include <memory>
 
 template <class INPUT, class OUTPUT>
-class SearchTask
+class SearchTask  // NOLINT
```
bassmang, line 18: what does NOLINT mean?

jackgerrits: It tells clang-tidy to ignore this line

bassmang (approved):

PR #4219, opened 2022-10-12
refactor: No RapidJSON in header files

byronxu99:
Hopefully this is a permanent fix for the GetObject issue
- Move implementation of functions using RapidJSON to `.cc` source files
- Replace the header files with just a declaration of the function
- Explicitly instantiate templates for the two possible values of `<bool audit>`
- Simplify CMake files now that RapidJSON no longer needs to be exported


jackgerrits (approved):
This is so great! Thank you!

PR #4232, opened 2022-10-18
refactor: reduce build time

jackgerrits:
- Move spdlog to private dependency in vw_io by adding virtual interface for log sink
  - This has a massive impact on build time because the spdlog header is relatively heavy and caused MANY heavy template instantiations
- Make vw_fwd header lighter
- Split apart some other headers

On my machine my build time went from about **30s to 17s**.


vowpalwabbit/core/include/vw/core/gen_cs_example.h
```diff
@@ -5,7 +5,7 @@
 #include "vw/common/vw_exception.h"
 #include "vw/core/reductions/cb/cb_algs.h"
 #include "vw/core/scope_exit.h"
-#include "vw/core/vw.h"
+// #include "vw/core/vw.h"
```
byronxu99, line 8: delete this line instead of commenting it out?

olgavrou (approved):

lalo (approved):

cs/cli/vw_example.cpp
```diff
@@ -2,6 +2,8 @@
 // individual contributors. All rights reserved. Released under a BSD (revised)
 // license as described in the file LICENSE.
 
+#define NOMINMAX
```
lalo, line 5: this used to be inside a ifdef _WIN32, is there a reason to not follow that

jackgerrits: CLI will only build on windows as far as I know so I figured it wasnt necessary

PR #4242, opened 2022-10-21
style: another round of style updates

jackgerrits:


peterychang (approved):
still not a fan of the snake cased object names, but otherwise looks fine

PR #4243, opened 2022-10-21
feat: [LAS] sparse Rademacher

olgavrou:
Sparse Rademacher

also:

- fix seed setting
- add scaling in AOmega multiplication
- fix some tests

### DIFF
--- a/test/unit_test/cb_las_one_pass_svd_test.cc
+++ b/test/unit_test/cb_las_one_pass_svd_test.cc
@@ -25,13 +25,19 @@ BOOST_AUTO_TEST_CASE(check_AO_same_actions_same_representation)
   std::vector<VW::workspace*> vws;
 
   auto* vw_zero_threads = VW::initialize("--cb_explore_adf --large_action_space --full_predictions --max_actions " +
-          std::to_string(d) + " --quiet --random_seed 5 --thread_pool_size 0",
+          std::to_string(d) + " --quiet --random_seed 1 --thread_pool_size 0",
       nullptr, false, nullptr, nullptr);
 
   vws.push_back(vw_zero_threads);
 
+  auto* vw_zero_threads_zs = VW::initialize("--cb_explore_adf --large_action_space --full_predictions --max_actions " +
+          std::to_string(d) + " --quiet --random_seed 0 --thread_pool_size 0",
+      nullptr, false, nullptr, nullptr);
+
+  vws.push_back(vw_zero_threads_zs);
+
   auto* vw_threads = VW::initialize("--cb_explore_adf --large_action_space --full_predictions --max_actions " +
-          std::to_string(d) + " --quiet --random_seed 5",
+          std::to_string(d) + " --quiet --random_seed 2",
       nullptr, false, nullptr, nullptr);
 
   vws.push_back(vw_threads);
@@ -72,7 +78,7 @@ BOOST_AUTO_TEST_CASE(check_AO_same_actions_same_representation)
       vw.predict(examples);
 
       // representation of actions 2 and 3 (duplicates) should be the same in U
-      BOOST_CHECK_EQUAL(action_space->explore.U.row(1).isApprox(action_space->explore.U.row(2)), true);
+      BOOST_CHECK_EQUAL(action_space->explore.U.row(1).isApprox(action_space->explore.U.row(2), FLOAT_TOL), true);
 
       vw.finish_example(examples);
     }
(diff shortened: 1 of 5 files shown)


test/unit_test/cb_las_one_pass_svd_test.cc
```diff
@@ -58,21 +58,21 @@ BOOST_AUTO_TEST_CASE(check_AO_same_actions_same_representation)
       VW::multi_ex examples;
 
       examples.push_back(VW::read_example(vw, "shared |U b c"));
```
olgavrou, line 60: adding namespaces in these examples would make the tests fail without these changes

test/unit_test/cb_las_spanner_test.cc
```diff
@@ -811,4 +837,76 @@ BOOST_AUTO_TEST_CASE(check_spanner_with_actions_that_are_linear_combinations_of_
   }
 }
 
+BOOST_AUTO_TEST_CASE(check_singular_value_sum_diff_for_diff_ranks_is_small)
```
olgavrou, line 840: this checks the scaling factor that was added in this PR, without this the diff would be > 500

olgavrou: this effects the non_degenerate singular value count if the first singular values can be huge without the scaling

vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
```diff
@@ -182,7 +182,7 @@ void cb_explore_adf_large_action_space<randomized_svd_impl, spanner_impl>::updat
       return;
     }
 
-    auto non_degen = std::min(_d, static_cast<uint64_t>(number_of_non_degenerate_singular_values() + 1));
+    auto non_degen = std::min(_d, static_cast<uint64_t>(number_of_non_degenerate_singular_values()));
```
olgavrou, line 185: not needed with sparse rademacher

test/unit_test/cb_las_spanner_test.cc
```diff
@@ -741,11 +767,11 @@ BOOST_AUTO_TEST_CASE(check_spanner_with_actions_that_are_linear_combinations_of_
 
       examples.push_back(VW::read_example(vw, "| 1:0.1 2:0.12 3:0.13 b200:2 c500:9"));
 
-      examples.push_back(VW::read_example(vw, "| a_1:0.5 a_2:0.65 a_3:0.12 a100:4 a200:33"));
-      examples.push_back(VW::read_example(vw, "| a_1:0.8 a_2:0.32 a_3:0.15 a100:0.2 a200:0.2"));
+      examples.push_back(VW::read_example(vw, "| a_1:0.1 a_2:0.25 a_3:0.12 a100:1 a200:0.1"));
```
olgavrou, line 770: toned these valued down since the non degen singular values are around 7, and once the singular values start dwindling one of the two first actions gets selected if their volume is quite high

vowpalwabbit/core/src/reductions/cb/details/large_action/one_pass_svd_impl.cc
```diff
@@ -77,8 +96,15 @@ class AO_triplet_constructor
...
   uint64_t _seed;
   float& _final_dot_product;
+  // index mapped used to select sparsity or not from value map
+  constexpr static std::array<size_t, 2> index_map = {0, 2};
+  constexpr static std::array<float, 4> value_map = {0.f, 0.f, 1.f, -1.f};
 };
 
+// definition at namespace scope
+constexpr std::array<size_t, 2> AO_triplet_constructor::index_map;
+constexpr std::array<float, 4> AO_triplet_constructor::value_map;
```
jackgerrits, line 106: What are these lines for?

vowpalwabbit/core/src/reductions/cb/details/large_action/one_pass_svd_impl.cc
```diff
@@ -77,8 +96,15 @@ class AO_triplet_constructor
   uint64_t _column_index;
   uint64_t _seed;
   float& _final_dot_product;
+  // index mapped used to select sparsity or not from value map
+  constexpr static std::array<size_t, 2> index_map = {0, 2};
+  constexpr static std::array<float, 4> value_map = {0.f, 0.f, 1.f, -1.f};
```
jackgerrits, line 101:
```suggestion
  constexpr static const std::array<size_t, 2> INDEX_MAP = {0, 2};
  constexpr static const std::array<float, 4> VALUE_MAP = {0.f, 0.f, 1.f, -1.f};
```

olgavrou: aren't all constexpr variables already const?

jackgerrits: Yeah I think you're right, can remove the cosnt

olgavrou: capitalized

vowpalwabbit/core/src/reductions/cb/details/large_action/one_pass_svd_impl.cc
olgavrou: you need to declare the constexpr static members at namespace scope ([resolved automatically in c++17](https://en.cppreference.com/w/cpp/language/static#Constant_static_members))

olgavrou: and can not declare them inline because (and I quote) "inline variables are only available with c++17"

PR #4261, opened 2022-11-01
feat: Add explicit simd implementation for one pass svd in large action spaces.

zwd-ms:
Add an explicit SIMD implementation using AVX-512 for the one pass svd part in large actions. For now, it only works with quadratic interactions.

By default, VW builds with `VW_BUILD_LAS_WITH_SIMD=ON` so it includes the SIMD binaries. The compile flag can be turned off, for example when some CI runs on unsupported architectures (or when someone building on really old platforms), but all tests should build and pass on a recent CPU.

To enable vectorization at runtime, pass in `--explicit_simd` on vw command line. This flag is advisory: it only works if the program detects that the platform actually supports the needed instructions.

Next steps include adding alternative implementations using other extension sets (e.g. AVX2)~~, and adding automatic switching between scalar and simd code paths using CPUID info at runtime, instead of using the command-line flag~~.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -9,6 +9,9 @@
     flake-utils.lib.eachDefaultSystem (system:
       let pkgs = nixpkgs.legacyPackages.${system}; in
       let
+        # Note: this compile_commands.json does not work with clangd because it
+        # uses the nix clang-wrapper which abstracts enough away so that clangd
+        # can no longer find the right includes.
         generate-compile-commands = ''
           echo -n "Generating compile_commands.json... "
           rm -rf $TMPDIR/compile_commands_build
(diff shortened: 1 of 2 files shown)


vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
```diff
@@ -258,15 +258,15 @@ void generate_Z(const multi_ex& examples, Eigen::MatrixXf& Z, Eigen::MatrixXf& B
 template <typename T, typename S>
 cb_explore_adf_large_action_space<T, S>::cb_explore_adf_large_action_space(uint64_t d, float gamma_scale,
     float gamma_exponent, float c, bool apply_shrink_factor, VW::workspace* all, uint64_t seed, size_t total_size,
-    size_t thread_pool_size, size_t block_size, implementation_type impl_type)
+    size_t thread_pool_size, size_t block_size, bool use_explicit_simd, implementation_type impl_type)
```
olgavrou, line 261: is there a reason we are enabling this by both turning on a compile time flag and a cli argument?

olgavrou: just saw the remark in the description of the PR

olgavrou (commented):
A separate CI job should be added that turns on the compile time flag and runs the related tests (like we had for LAS while it was behind a compile time flag). It can run only for linux architectures

olgavrou (approved):

PR #4262, opened 2022-11-02
fix: explore_eval don't learn if logged action not in predicted actions

olgavrou:


vowpalwabbit/core/src/reductions/explore_eval.cc
```diff
@@ -164,6 +165,8 @@ void do_actual_learning(explore_eval& data, multi_learner& base, VW::multi_ex& e
       if (data.known_cost.action == a_s[i].action) { action_probability = a_s[i].score; }
     }
 
+    if (action_probability == 0) { return; }
+
     float threshold = action_probability / data.known_cost.probability;
 
     if (!data.fixed_multiplier) { data.multiplier = std::min(data.multiplier, 1 / threshold); }
```
olgavrou, line 172: now that probs can be zero, this ` 1 / threshold` could cause division by zero

vowpalwabbit/core/src/reductions/explore_eval.cc
```diff
@@ -164,6 +165,8 @@ void do_actual_learning(explore_eval& data, multi_learner& base, VW::multi_ex& e
       if (data.known_cost.action == a_s[i].action) { action_probability = a_s[i].score; }
     }
 
+    if (action_probability == 0) { return; }
```
jackgerrits, line 168:
```suggestion
    if (action_probability == 0.f) { return; }
```

Also, should this be a comparison with some tolerance instead of exact?

olgavrou: if it is LAS then the action will be missing (action_mask is after explore_eval) and the for loop won't assign anything and so the action_probability will be as it was set on line 162

otherwise we go to previous explore_eval and it might be small but not zero so don't think we should skip in that case

vowpalwabbit/core/src/reductions/explore_eval.cc
```diff
@@ -183,15 +191,18 @@ void do_actual_learning(explore_eval& data, multi_learner& base, VW::multi_ex& e
...
 
+      // restore logged example
       if (threshold > 1)
       {
         float inv_threshold = 1.f / threshold;
         for (auto& ec : ec_seq) { ec->weight *= inv_threshold; }
       }
+
       ec_found->l.cb.costs[0].probability = data.known_cost.probability;
       data.update_count++;
```
ataymano, line 207: are we ok to skip this?

ataymano (approved):

jackgerrits (approved):

PR #4264, opened 2022-11-03
feat: Adding EMT reduction.

mrucker:
This pull request contains an implementation of the EMT reduction (https://arxiv.org/abs/2210.14077) along with unit tests.


vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,928 @@
...
+namespace reductions
+{
+namespace eigen_memory_tree
+{
+///////////////////////Helper/////////////////////////////
+//////////////////////////////////////////////////////////
+float median(std::vector<float>& array)
+{
+  int size = array.size();
+  std::sort(array.begin(), array.end());
```
ataymano, line 46: is it beneficial to have linear complexity algorithm here?
i.e. stl ones: https://cplusplus.com/reference/algorithm/nth_element/

vowpalwabbit/core/include/vw/core/reductions/eigen_memory_tree.h
```diff
@@ -0,0 +1,123 @@
...
+
+  int max_size;
+
+  emt_LRU(unsigned long max_size);
+  K bound(K item);
+};
+
+struct emt_node
+{
+  emt_node* parent; // parent node
```
ataymano, line 79: are we using parent anywhere?

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
mrucker: Oh, nice, I didn't know about that method. That's an easy change to make to be safe. Jack had also mentioned maybe there was an existing algorithm in VW that is even better? For what it is worth, in practice I don't think it'll matter too much. We only calculate the median when we split a node in our binary tree, which means it's calculated with log(N) frequency. And when we do calculate it is only on the items in the leaf node being split.

vowpalwabbit/core/include/vw/core/reductions/eigen_memory_tree.h
mrucker: Good catch. No I don't think we are. I think I just added that out of habit.

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,935 @@
...
+///////////////////////Helper/////////////////////////////
+//////////////////////////////////////////////////////////
+float median(std::vector<float>& array)
+{
+  int size = array.size();
+  auto nth = array.begin() + size / 2; 
+
+  if (size % 2 == 0) {
+    std::nth_element(array.begin(), nth  , array.end());
+    std::nth_element(array.begin(), nth-1, array.end());
```
ataymano, line 50: >>The other elements are left without any specific order, except that none of the elements preceding nth are greater than it, and none of the elements following it are less.

Looks like this call is invalidating guaraantees about nth from previous line.
We probably should either:
1) get the *nth value between these calls
2) replace second nth_element call with something like (constant factor seems to be better here)
```
auto n_minus_1th = std::max_element(array.begin(), array.begin() + size / 2 - 1)
```

mrucker: I grabbed the nth between the calls. I wasn't entirely sure grabbing the max was guaranteed to work given nth_element. Feel free to change it.

vowpalwabbit/core/include/vw/core/reductions/eigen_memory_tree.h
mrucker: I've cleaned this up.

mrucker:
I finished a major cleaning pass. I think EMT is nearly done. With this pass I:

1. Removed all new statements and replaced then with VW::make_unique
2. Removed all class attributes which weren't actually used (e.g., parent)
3. Refactored to remove two different enums that weren't actually needed
4. Removed unneeded destructors now that we're using make_unique
5. Replaced hacky uses of VW internal classes with std data types
6. Added a lot of unit tests
7. Replaced all couts with trace_message
8. Validated the performance of this implementation against the 206 datasets used in the original publication

vowpalwabbit/core/include/vw/core/reductions/eigen_memory_tree.h
```diff
@@ -0,0 +1,111 @@
...
+
+float emt_inner(emt_feats, emt_feats);
+
+emt_feats emt_scale(emt_feats, float);
+
+float emt_norm(emt_feats);
+
+emt_feats emt_normalize(emt_feats);
+
+emt_feats emt_scale_add(float const, emt_feats const, float const, emt_feats const);
```
ataymano, line 47: Looks like we are doing copying of emt_feats objects in all functions where emt_feats is part of signature.
Can it be changed to references or pointers?

mrucker: Yeah, that's a really good idea. Almost certainly. I'll make a performance pass over the code and improve this.

mrucker: I just finished an optimization pass to tune all function signatures (along with a few other optimizations). 

On a large test data set EMT now runs in about 4 seconds where it used to run in about 6 seconds.

vowpalwabbit/core/include/vw/core/reductions/eigen_memory_tree.h
```diff
@@ -49,6 +49,24 @@ emt_feats emt_scale_add(float, const emt_feats&, float, const emt_feats&);
 
 emt_feats emt_router_eigen(std::vector<emt_feats>&, VW::rand_state&);
 
+template<typename I, typename R>
```
jackgerrits: Nit: `I` is too short. At least `It` or better, `RandomIt` is preferred since ideally we want a random access iterator here

vowpalwabbit/core/include/vw/core/reductions/eigen_memory_tree.h
```diff
@@ -49,6 +49,24 @@ emt_feats emt_scale_add(float, const emt_feats&, float, const emt_feats&);
...
+void emt_shuffle(I first, I last, R* rng)
```
jackgerrits, line 53: Based on the implementation this will only work with a `VW::rand_state`, so I think we should just remove the template for `R` and use that type.

mrucker: Agreed, that's what I had originally but the compiler was yelling at me.

> invalid use of incomplete type ‘class VW::rand_state’

Any suggestions on this?

vowpalwabbit/core/include/vw/core/reductions/eigen_memory_tree.h
mrucker: Good suggestion. Made the recommended change.

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+  // we will populate this examples features over and over.
+  ex = VW::make_unique<VW::example>();
+  ex->interactions = new std::vector<std::vector<VW::namespace_index>>();
+  ex->extent_interactions = new std::vector<std::vector<extent_term>>();
+  ex->indices.push_back(0);
+}
+
+emt_tree::~emt_tree()
+{
+  delete ex->interactions;
```
ataymano, line 127: can nullptr dereferencing happen here in case of 
````
{
  emt_tree a;
}
```

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+{
+  leaf_split = 100;
+
+  scorer_type = emt_scorer_type::self_consistent_rank;
+  router_type = emt_router_type::eigen;
+
+  begin = 0;
+
+  all = nullptr;
+  ex = nullptr;
```
ataymano, line 113: can be removed?

mrucker: Or you mean begin? It looks like it.

mrucker: If you see any small refactors like this feel free to make changes yourself.

ataymano: I am sorry for short comment. I was meaning "ex=nullptr" which is initialized 2 lines later.

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+                     "consistent rank)"))
+      .add(make_option("emt_router", router_type)
+               .keep()
+               .one_of({1, 2})
+               .default_value(2)
+               .help("Indicates the type of router to use (1=random;2=eigen)"));
+
+  if (!options.add_parse_and_check_necessary(new_options)) { return nullptr; }
+
+  t->scorer_type = static_cast<VW::reductions::eigen_memory_tree::emt_scorer_type>(scorer_type);
```
ataymano, line 756: it looks a bit dangerous (i e if someone adds new scorer_type in the middle). Probably helper conversion from string constant to enum (either with switch/case or using std::map) with explicit error for unknown names is a bit better here

vowpalwabbit/core/include/vw/core/reductions/eigen_memory_tree.h
```diff
@@ -0,0 +1,127 @@
...
+void emt_scale(emt_feats&, float);
+
+void emt_normalize(emt_feats&);
+
+emt_feats emt_scale_add(float, const emt_feats&, float, const emt_feats&);
+
+emt_feats emt_router_eigen(std::vector<emt_feats>&, VW::rand_state&);
+
+template <typename RandomIt>
+void emt_shuffle(RandomIt first, RandomIt last, VW::rand_state* rng)
```
jackgerrits, line 54: We expect this to be non-null and we don't check for null, so we should use a reference instead.

vowpalwabbit/core/include/vw/core/reductions/eigen_memory_tree.h
```diff
@@ -0,0 +1,127 @@
...
+
+namespace VW
+{
+namespace reductions
+{
+VW::LEARNER::base_learner* eigen_memory_tree_setup(VW::setup_base_i& stack_builder);
+
+namespace eigen_memory_tree
+{
+using emt_feats = std::vector<std::pair<int, float>>;
```
jackgerrits, line 21: Are these hashes? If so, then we should use the same type as elsewhere `uint64_t`

vowpalwabbit/core/include/vw/core/reductions/eigen_memory_tree.h
```diff
@@ -0,0 +1,127 @@
...
+
+struct emt_lru
+{
+  using K = emt_example*;
+  using V = std::list<K>::iterator;
+
+  std::list<K> list;
+  std::unordered_map<K, V> map;
+
+  unsigned long max_size;
```
jackgerrits, line 87: Explicitly sized types should be used instead. e.g. `uint64_t`

vowpalwabbit/core/include/vw/core/reductions/eigen_memory_tree.h
```diff
@@ -0,0 +1,127 @@
...
+  VW::workspace* all;
+  std::shared_ptr<VW::rand_state> _random_state;
+
+  uint32_t leaf_split;  // how many memories before splitting a leaf node
+  emt_scorer_type scorer_type;
+  emt_router_type router_type;
+
+  std::unique_ptr<VW::example> ex;  // we create one of these which we re-use so we don't have to reallocate examples
+
+  long begin;  // for timing performance
```
jackgerrits, line 116: Can this be an explicitly sized type?

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+
+namespace VW
+{
+namespace reductions
+{
+namespace eigen_memory_tree
+{
+////////////////////////////definitions/////////////////////
+////////////////////////////////////////////////////////////
+emt_example::emt_example() { label = 0; }
```
jackgerrits, line 42: IMO, just set label to 0 in the class definition and then don't define a constructor and let the compiler do it

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+    return last_value;
+  }
+  else { return nullptr; }
+}
+
+emt_node::emt_node()
+{
+  left = nullptr;
+  right = nullptr;
+  router_decision = 0;
```
jackgerrits, line 100: Same comment here about doing it in the class def

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+  delete ex->interactions;
+  delete ex->extent_interactions;
+}
+
+////////////////////////end of definitions/////////////////
+///////////////////////////////////////////////////////////
+
+///////////////////////Helper/////////////////////////////
+//////////////////////////////////////////////////////////
+float emt_median(std::vector<float>& array)
```
jackgerrits, line 136: This can be a const input

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+{
+  int size = array.size();
+  auto nth = array.begin() + size / 2;
```
jackgerrits, line 139:
```suggestion
  const auto nth = array.begin() + size / 2;
```

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+
+  if (size % 2 == 0)
+  {
+    std::nth_element(array.begin(), nth, array.end());
+    auto v1 = (*nth);
+    std::nth_element(array.begin(), nth - 1, array.end());
+    auto v2 = *(nth - 1);
```
jackgerrits, line 146:
```suggestion
    const auto v2 = *(nth - 1);
```

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+    std::nth_element(array.begin(), nth, array.end());
+    return *nth;
+  }
+}
+
+float emt_inner(const emt_feats& xs, const emt_feats& ys)
+{
+  float sum = 0;
+  unsigned long xi = 0;
+  unsigned long yi = 0;
```
jackgerrits, line 160: Prefer explicit sized types, unless it needs to be `size_t` which looks like it may be the case.

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+  b.begin = clock();
+}
+/////////////////end of eigen_memory_tree///////////////////
+////////////////////////////////////////////////////////////
+
+///////////////////Save & Load//////////////////////////////////////
+////////////////////////////////////////////////////////////////////
+void save_load_examples(emt_node& n, io_buf& model_file, bool& read, bool& text, std::stringstream& msg)
+{
+  WRITEITVAR(n.examples.size(), "n_examples", n_examples);
```
jackgerrits, line 639: `WRITEITVAR` should not be used. Please use utility functions in `VW::model_utils`

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+  if (read)
+  {
+    for (uint32_t i = 0; i < n_examples; i++) { n.examples.push_back(VW::make_unique<emt_example>()); }
+  }
+
+  for (auto& e : n.examples)
+  {
+    for (auto& p : e->base)
+    {
+      WRITEIT(p.first, "base_index");
```
jackgerrits, line 650: `WRITEIT` should not be used. Please use utility functions in `VW::model_utils`

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+      WRITEIT(p.second, "full_value");
+    }
+  }
+}
+
+void save_load_weights(emt_node& n, io_buf& model_file, bool& read, bool& text, std::stringstream& msg)
+{
+  for (auto& p : n.router_weights)
+  {
+    WRITEIT(p.first, "router_index");
```
jackgerrits, line 665: `WRITEIT` should not be used. Please use utility functions in `VW::model_utils`

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+    WRITEIT(p.second, "router_value");
+  }
+}
+
+std::unique_ptr<emt_node> save_load_node(
+    emt_tree& b, std::unique_ptr<emt_node> n, io_buf& model_file, bool& read, bool& text, std::stringstream& msg)
+{
+  WRITEITVAR(!read && !n, "is_null", is_null);
```
jackgerrits, line 673: Essentially, you need to implement `read_model_field` and `write_model_field` for your new types to allow for serialization/deserialization.

jackgerrits (changes requested):

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+    b.all->ignore_linear['x'] = true;
+    b.all->ignore_linear['z'] = true;
+
+    scorer_features(ex1.full, out.feature_space['x']);
+    scorer_features(ex2.full, out.feature_space['z']);
+
+    // when we receive ex1 and ex2 their features are indexed on top of eachother. In order
+    // to make sure VW recognizes the features from the two examples as separate features
+    // we apply a map of multiplying by 2 and then offseting by 1 on the second example.
+    for (auto& j : out.feature_space['x'].indices) { j = j * 2; }
```
ataymano, line 394: is it safe to get above (potentially) the maximum hash here?

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+}
+
+void scorer_example(emt_tree& b, const emt_example& ex1, const emt_example& ex2)
+{
+  VW::example& out = *b.ex;
+
+  if (b.scorer_type == emt_scorer_type::self_consistent_rank)
+  {
+    out.indices.clear();
+    out.indices.push_back('x');
```
ataymano, line 358: Can you please define constants for 'x'/'z'?

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+void scorer_features(const emt_feats& f1, features& out)
+{
+  out.clear();
+  for (auto p : f1)
+  {
+    if (p.second != 0) { out.push_back(p.second, p.first); }
+  }
+}
+
+void scorer_example(emt_tree& b, const emt_example& ex1, const emt_example& ex2)
```
ataymano, line 351: probably outside of the scope of this pr, but to consider in the future: as far as i understand randomness is the only reason for emt_tree being nonconst here, so maybe it is worth to consider having rng as mutable.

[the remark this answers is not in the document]
ataymano: just to confirm what kind of asymmetry i mean:
scorer_example({0: a, 1: b}, {0: c})  = some_f({0: a, 1: c, 2: b})
scorer_example({0:c}, {0: a, 1: b}) = some_f({0: c, 1: a, 3: b})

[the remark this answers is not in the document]
mrucker: Good catch. The asymmetry is not desired.  Here's what we really wanted...

scorer_example({0: a, 1: b}, {0: c}) = some_f({00:ac, 10:bc})

It that possible using vw's interactions. I think I did the offset because it seemed to me like VW wasn't calculating the interactions at all for features in different namespaces with identical indexes (which I know is a strange use case to begin with).

That is, users could say --interactions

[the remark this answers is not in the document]
mrucker: (also worth noting, before moving all this to VW we were manually calculating these features so this wasn't an issue)

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,787 @@
...
+  auto end_pass = VW::reductions::eigen_memory_tree::end_pass;
+  auto save_load = VW::reductions::eigen_memory_tree::save_load_tree;
+
+  auto l = make_reduction_learner(std::move(t), as_singleline(stack_builder.setup_base_learner()), learn, predict,
+      stack_builder.get_setupfn_name(eigen_memory_tree_setup))
+               .set_learn_returns_prediction(true)  // we set this to true otherwise bounding doesn't work as well
+               .set_end_pass(end_pass)
+               .set_save_load(save_load)
+               .set_output_prediction_type(pred_type)
+               .set_input_label_type(label_type);
```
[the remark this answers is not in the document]
mrucker: @jackgerrits What would the input prediction type be? Is that like an initial prediction that we would update?

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
mrucker: @ataymano That's why I explicitly set the values of the enum... Does it still seem dangerous to you?

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
mrucker: @jackgerrits I don't think I can do this because there is another constructor defined on this class.

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
mrucker: @jackgerrits Can it? My compiler complains. I think it is because we re-order the array?

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
mrucker: @ataymano Yes, I wouldn't have the time for that.

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
jackgerrits: In that case you can just mark it in the header like so:
```c++
struct emt_example
{
  emt_feats base;  // base example only includes the base features without interaction flags
  emt_feats full;  // full example includes the interactions that were passed in as flags
  uint32_t label = 0;

  emt_example() = default // <-----------
  emt_example(VW::workspace&, VW::example*);
};

```

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
jackgerrits: You're right, I didn't realize `std::nth_element` mutates the input.

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
jackgerrits: https://github.com/VowpalWabbit/vowpal_wabbit/blob/5ca2a7f54a9bf02794c1417b68439538d638555c/vowpalwabbit/core/src/reductions/epsilon_decay.cc#L274

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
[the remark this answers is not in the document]
jackgerrits: When you call `base.predict` what is the prediction type you are expecting from below this reduction? Similarly, when calling `base.learn` what are you setting the label to for that call?

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
mrucker: ooh fancy, I like it

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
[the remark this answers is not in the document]
mrucker: got it: when I call base.predict and base.learn I expect/provide a simple regression label.

vowpalwabbit/core/tests/eigen_memory_tree_test.cc
```diff
@@ -328,4 +328,34 @@ TEST(emt_tests, test_emt_shuffle)
...
 
+TEST(emt_tests, emt_save_load)
+{
+  auto vw_save = VW::initialize_experimental(vwtest::make_args("--quiet", "--emt", "--emt_leaf", "5"));
+
+  for (int i = 0; i < 10; i++)
+  {
+    auto* ex = VW::read_example(*vw_save, std::to_string(i) + " | " + std::to_string(i));
+    vw_save->learn(*ex);
+    VW::finish_example(*vw_save, *ex);
```
jackgerrits, line 339: You need to call `vw_save->finish_example(*ex)` instead of this.

ataymano:
was playing around to understand better and the following input data is crashing release build (but not crashing debug one):
```
n = 10
m = 10
with open('data.txt', 'w') as f:
    for i in range(n):
        for j in range(m):
            v = j / m
            l = int(v * 2)
            f.write(f'{l} | 1:{v} 5:{v}\n')
```
Looks like there is ub somewhere

ataymano:
playing with it to understand better.
On the following data:
```
n = 10
m = 10
with open('data.txt', 'w') as f:
    for i in range(n):
        for j in range(m):
            v = j / m
            l = int(v * 2)
            f.write(f'{l} | 2:{v} 100:{v}\n')
```
following command line
```
-d data.txt --readable_model r_model.txt --emt --noconstant
```
is crashing with 
image
looking.

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -0,0 +1,815 @@
...
+  bytes += read_model_field(io, node.right);
+  bytes += read_model_field(io, node.router_weights);
+  bytes += read_model_field(io, node.examples);
+  return bytes;
+}
+
+size_t write_model_field(
+    io_buf& io, const reductions::eigen_memory_tree::emt_node& node, const std::string& upstream_name, bool text)
+{
+  size_t bytes = 0;
```
ataymano, line 699: readable model looks like that now:
```
...
emtrootrightrightrightright.is_null() = false
emtrootrightrightrightrightrouter_decision = 0.3114859461784363
emtrootrightrightrightrightleft.is_null() = false
emtrootrightrightrightrightleftrouter_decision = 0
emtrootrightrightrightrightleftleft.is_null() = true
emtrootrightrightrightrightleftright.is_null() = true
emtrootrightrightrightrightleftrouter_weights.size() = 0
emtrootrightrightrightrightleftexamples.size() = 4
emtrootrightrightrightrightleftexamples[0].is_null() = false
emtrootrightrightrightrightleftexamples[0]base.size() = 3
emtrootrightrightrightrightleftexamples[0]base[0].first = 2
emtrootrightrightrightrightleftexamples[0]base[0].second = 0.07031249
emtrootrightrightrightrightleftexamples[0]base[1].first = 100
emtrootrightrightrightrightleftexamples[0]base[1].second = 0.07031249
emtrootrightrightrightrightleftexamples[0]base[2].first = 116060
emtrootrightrightrightrightleftexamples[0]base[2].second = 1
emtrootrightrightrightrightleftexamples[0]full.size() = 3
emtrootrightrightrightrightleftexamples[0]full[0].first = 2
emtrootrightrightrightrightleftexamples[0]full[0].second = 0.07031249
emtrootrightrightrightrightleftexamples[0]full[1].first = 100
emtrootrightrightrightrightleftexamples[0]full[1].second = 0.07031249
emtrootrightrightrightrightleftexamples[0]full[2].first = 116060
emtrootrightrightrightrightleftexamples[0]full[2].second = 1
emtrootrightrightrightrightleftexamples[0]label = 1
emtrootrightrightrightrightleftexamples[1].is_null() = false
emtrootrightrightrightrightleftexamples[1]base.size() = 3
emtrootrightrightrightrightleftexamples[1]base[0].first = 2
emtrootrightrightrightrightleftexamples[1]base[0].second = 0.062499993
emtrootrightrightrightrightleftexamples[1]base[1].first = 100
emtrootrightrightrightrightleftexamples[1]base[1].second = 0.062499993
emtrootrightrightrightrightleftexamples[1]base[2].first = 116060
emtrootrightrightrightrightleftexamples[1]base[2].second = 1
emtrootrightrightrightrightleftexamples[1]full.size() = 3
emtrootrightrightrightrightleftexamples[1]full[0].first = 2
...
```
should we separate upstream from field name with '.'?

ataymano:
> 

line 55 from screenshot is trying to dereference .end() iterator in case of empty example (i.e. "0 | x:0" with "--noconstant"). pushing fix.

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
jackgerrits: Yeah I think that would help readability a lot here

ataymano (approved):

jackgerrits (approved):

PR #4266, opened 2022-11-03
fix!: save/load entire tag in flat_example + bump version to 9.6

bassmang:


jackgerrits:
Because we know this is a breaking change just for KSVM models we should add an error if trying to load an affected model rather than a silent failure

jackgerrits (approved):

PR #4277, opened 2022-11-09
feat: explore eval example target rate

olgavrou:
We want to be able to tell explore_eval to target a specific number of examples to use while updating. The goal is to not go over the example count that we want

Added `block_size` argument to explore_eval. This should be calculated by doing:
 `#examples_in_file / #examples_to_use_during_update`

The examples will then be processed in blocks of `block_size`. In each block one example can be used to update. If in that block none of the examples pass the threshold then the example quota is rolled over to the next block and the next block can select 2 examples to update and so on


vowpalwabbit/core/src/reductions/explore_eval.cc
```diff
@@ -184,6 +198,10 @@ void do_actual_learning(explore_eval& data, multi_learner& base, VW::multi_ex& e
 
     if (data.random_state->get_and_update_random() < threshold)
     {
+      // if we reached the example quota for this example block just skip
+      if (data.block_ex_quota == 0) { return; }
```
ataymano, line 202: isn't it more readable to have it as elseif on line 172?

ataymano (approved):

vowpalwabbit/core/src/reductions/explore_eval.cc
olgavrou: yeah probably, updated

olgavrou: ah no if we skip too early then know_cost isn't changed

PR #4278, opened 2022-11-09
 refactor: [versioning] split options in scorer reduction

peterychang:


vowpalwabbit/core/src/reductions/scorer.cc
```diff
@@ -21,6 +22,11 @@ using namespace VW::config;
 
 namespace
 {
+struct options_scorer
```
bassmang, line 25: naming convention from the PR I have out would be `options_scorer_v1`, with the idea that new versions could be introduced later. Open to other convensions, thoughts @lokitoth ?

vowpalwabbit/core/src/reductions/scorer.cc
```diff
@@ -21,6 +22,11 @@ using namespace VW::config;
...
+{
+  std::string link;
```
lalo, line 27: it would be good to move away from string repr to another more restrictive type.

peterychang: I considered an enum class, but we still need access to the string itself regardless (for the name in this case..). I'm not sure what the final design of these objects will look like, but so far they all look like structs with public members so I'm sticking with that for now

PR #4279, opened 2022-11-10
feat: spin off automl predict_only_model to standard cb model

bassmang:


vowpalwabbit/core/src/reductions/automl.cc
```diff
@@ -92,16 +92,49 @@ void finish_example(VW::workspace& all, automl<CMType>& data, VW::multi_ex& ec)
...
+    if (data.cm->weights[index] != 0.0f)
+    {
+      uint32_t cb_ind = index / data.cm->wpp;
+      for (uint32_t stride = 0; stride < (static_cast<uint32_t>(1) << data.cm->weights.stride_shift()); ++stride)
+      { std::swap(data.cm->weights[index + stride], data.cm->weights[cb_ind + stride]); }
+    }
+  }
+  uint32_t num_bits =
+      options.was_supplied("bit_precision") ? options.get_typed_option<uint32_t>("bit_precision").value() : 18;
+  options.remove_option("automl");
```
lalo, line 122: do you have to remove all the other automl related options? if not when re-loading you'll get a warning/error about using un-used args. Try it out with --priority_challenger or something like that.

vowpalwabbit/core/src/reductions/automl.cc
```diff
@@ -92,16 +92,49 @@ void finish_example(VW::workspace& all, automl<CMType>& data, VW::multi_ex& ec)
...
+  std::swap(*data.cm->_sd_gravity, data.cm->per_live_model_state_double[2]);
+  std::swap(*data.cm->_cb_adf_event_sum, data.cm->per_live_model_state_uint64[0]);
+  std::swap(*data.cm->_cb_adf_action_sum, data.cm->per_live_model_state_uint64[1]);
+
+  clear_non_champ_weights(data.cm->weights, data.cm->estimators.size(), data.cm->wpp);
+
+  uint64_t multiplier = static_cast<uint64_t>(data.cm->wpp) << data.cm->weights.stride_shift();
+  for (uint32_t index = 0; index < data.cm->weights.mask(); index += multiplier)
+  {
+    if (data.cm->weights[index] != 0.0f)
```
lalo, line 113: this might not be compatible with ftrl/coin - you would need to check all the other values in the same chunk just like here https://github.com/VowpalWabbit/vowpal_wabbit/blob/417012d772bf26cecc8d234a820994aa65144d82/vowpalwabbit/core/src/reductions/gd.cc#L1062

vowpalwabbit/core/include/vw/core/learner.h
```diff
@@ -329,6 +337,13 @@ class learner
     if (_save_load_fd.base) { _save_load_fd.base->save_load(io, read, text); }
   }
 
+  // called to edit the command-line from a reduction.
+  inline void NO_SANITIZE_UNDEFINED output_single_model(VW::workspace& all)
+  {
+    if (_output_single_model_fd.output_single_model_f == nullptr) { return; }
+    _output_single_model_fd.output_single_model_f(all, _output_single_model_fd.data);
```
lalo, line 344: shouldn't you call recursively down the stack in case another reduction needs to do some cleanup? might be an extreme edge case for now - but similar to save_load.

test/core.vwtest.json
```diff
@@ -5464,5 +5450,59 @@
...
+    "vw_command": "--cb_explore_adf -d train-sets/cb_test.ldf --noconstant -f models/aml_spin_off_cb.m -b 17 --predict_only_model -q ::",
+    "diff_files": {
+      "stderr": "train-sets/ref/cb_similar_aml_spinoff_save.stderr"
+    },
+    "input_files": [
+      "train-sets/cb_test.ldf"
+    ]
+  },
+  {
+    "id": 426,
```
lalo, line 5493: you can also recycle this unit test: https://github.com/VowpalWabbit/vowpal_wabbit/blob/417012d772bf26cecc8d234a820994aa65144d82/test/unit_test/automl_weights_test.cc#L263

line 280 model is -b 20, line 279 is -b 18, so if you run --predict_only_model on 280 line one it should be equal to the other one 279 -- as in exactly the same bit by bit.

vowpalwabbit/core/src/reductions/automl.cc
lalo: also might be worth checking if GD writes to model if we have something like w[index] == 0.0f, w[index+1] != 0.0f, w[index+2] != 0.0f, w[index+3] != 0.0f.

vowpalwabbit/core/src/reductions/automl.cc
lalo: you might be able to automate it similar to how we extract the name of a reduction setup, but instead of registering the name you register all the options https://github.com/VowpalWabbit/vowpal_wabbit/blob/417012d772bf26cecc8d234a820994aa65144d82/vowpalwabbit/core/src/reduction_stack.cc#L113

bassmang: updated to check and remove all automl options

vowpalwabbit/core/include/vw/core/learner.h
bassmang: Updated to be recursive. I dont think it makes sense at the moment to have more than one of these in the stack, but could be useful in the future

vowpalwabbit/core/src/parse_regressor.cc
```diff
@@ -467,6 +466,7 @@ void dump_regressor(VW::workspace& all, io_buf& buf, bool as_text)
 {
   if (buf.num_output_files() == 0) { THROW("Cannot dump regressor with an io buffer that has no output files."); }
   std::string unused;
+  if (all.l != nullptr) { all.l->output_single_model(all); }
```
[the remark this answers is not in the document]
bassmang: updated to `convert_to_single_model`. I think it should say specify something about converting a multi-model reduction to single-model, open to other suggestions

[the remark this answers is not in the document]
lalo: maybe uninterleave_model or unravel_model, technically an interleaved model is a single model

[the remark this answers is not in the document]
olgavrou: `output_target_model`?

vowpalwabbit/core/src/reductions/automl.cc
```diff
@@ -92,16 +92,49 @@ void finish_example(VW::workspace& all, automl<CMType>& data, VW::multi_ex& ec)
...
+  std::swap(*data.cm->_gd_normalized, data.cm->per_live_model_state_double[0]);
+  std::swap(*data.cm->_gd_total_weight, data.cm->per_live_model_state_double[1]);
+  std::swap(*data.cm->_sd_gravity, data.cm->per_live_model_state_double[2]);
+  std::swap(*data.cm->_cb_adf_event_sum, data.cm->per_live_model_state_uint64[0]);
+  std::swap(*data.cm->_cb_adf_action_sum, data.cm->per_live_model_state_uint64[1]);
+
+  clear_non_champ_weights(data.cm->weights, data.cm->estimators.size(), data.cm->wpp);
+
+  uint64_t multiplier = static_cast<uint64_t>(data.cm->wpp) << data.cm->weights.stride_shift();
+  for (uint32_t index = 0; index < data.cm->weights.mask(); index += multiplier)
```
olgavrou, line 111: should this be `index *= multiplier`?

vowpalwabbit/core/src/reductions/automl.cc
```diff
@@ -92,16 +92,49 @@ void finish_example(VW::workspace& all, automl<CMType>& data, VW::multi_ex& ec)
...
+  {
+    if (data.cm->weights[index] != 0.0f)
+    {
+      uint32_t cb_ind = index / data.cm->wpp;
+      for (uint32_t stride = 0; stride < (static_cast<uint32_t>(1) << data.cm->weights.stride_shift()); ++stride)
```
olgavrou, line 116: do you want the stride here or the indexes between two strides? if it is the latter maybe using `stride` here for the indexing parameter is misleading

vowpalwabbit/core/src/reductions/automl.cc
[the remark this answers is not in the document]
lalo: https://github.com/VowpalWabbit/vowpal_wabbit/blob/417012d772bf26cecc8d234a820994aa65144d82/vowpalwabbit/core/include/vw/core/array_parameters_dense.h#L63

olgavrou: also maybe create a function call here that encapsulates this multiplier indexing

vowpalwabbit/core/src/reductions/automl.cc
```diff
@@ -92,16 +92,49 @@ void finish_example(VW::workspace& all, automl<CMType>& data, VW::multi_ex& ec)
...
+  if (!options.was_supplied("predict_only_model")) { return; }
+  // Clear non-champ weights first
+
+  std::swap(*data.cm->_gd_normalized, data.cm->per_live_model_state_double[0]);
+  std::swap(*data.cm->_gd_total_weight, data.cm->per_live_model_state_double[1]);
+  std::swap(*data.cm->_sd_gravity, data.cm->per_live_model_state_double[2]);
+  std::swap(*data.cm->_cb_adf_event_sum, data.cm->per_live_model_state_uint64[0]);
+  std::swap(*data.cm->_cb_adf_action_sum, data.cm->per_live_model_state_uint64[1]);
+
+  clear_non_champ_weights(data.cm->weights, data.cm->estimators.size(), data.cm->wpp);
```
olgavrou, line 108: could you write a comment here perhaps explaining what the code below is attempting to do?

test/core.vwtest.json
bassmang: made a test `automl_equal_spin_off_model` which is similar and shows the weights matching

vowpalwabbit/core/src/reductions/automl.cc
```diff
@@ -92,16 +92,74 @@ void finish_example(VW::workspace& all, automl<CMType>& data, VW::multi_ex& ec)
...
+  options.get_typed_option<uint32_t>("bit_precision").value(num_bits - static_cast<uint32_t>(std::log2(data.cm->wpp)));
+  all.num_bits = num_bits - static_cast<uint32_t>(std::log2(data.cm->wpp));
+
+  std::vector<std::string> interactions_opt;
+  for (auto& interaction : data.cm->estimators[0].first.live_interactions)
+  {
+    std::string interaction_string;
+    for (auto& ns : interaction)
+    {
+      if (ns == ' ') { interaction_string += "\\x20"; }
```
lalo, line 140: do we need to do this for ccb related ns?

vowpalwabbit/core/src/reductions/automl.cc
bassmang: it is supposed to be `+=`, maybe multiplier isnt the right word

test/unit_test/automl_weights_test.cc
```diff
@@ -325,3 +326,64 @@ BOOST_AUTO_TEST_CASE(automl_equal_no_automl_w_iterations)
...
+      "--cb_explore_adf --quiet --epsilon 0.2 "
+      "--random_seed 5 --predict_only_model ";
+  std::string vw_automl_arg =
+      "--automl 4 --priority_type favor_popular_namespaces "
+      "--oracle_type one_diff --global_lease 10 ";
+  int seed = 10;
+  // a switch happens around ~1756
+  size_t num_iterations = 1700;
+
+  auto* vw_qcolcol = VW::initialize(vw_arg + "-b 17 -q ::");
```
lalo, line 344: can we save the model here and then after calling vw::finish hash both and make sure its the same contents? - we would also have to delete the model file to be good citizens

vowpalwabbit/core/src/parse_regressor.cc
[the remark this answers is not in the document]
bassmang: > `output_target_model`?

I think this is good name

vowpalwabbit/core/src/reductions/automl.cc
bassmang: updated to `stride_ind`, also moved this function to `array_parameters_dense.cc`

vowpalwabbit/core/src/reductions/automl.cc
bassmang: added comments for each operation

vowpalwabbit/core/src/reductions/automl.cc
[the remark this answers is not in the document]
bassmang: this seems possible but might complicate things

vowpalwabbit/core/src/reductions/automl.cc
bassmang: I don't think so since ccb namespaces aren't included in the command line

test/unit_test/automl_weights_test.cc
bassmang: https://github.com/VowpalWabbit/vowpal_wabbit/pull/4279/files#diff-dbdeee57e850cde4835b15d683271658db36cb50cc9e288deaf1a6beff534a5fR5511

This test checks that the model files are identical

vowpalwabbit/core/src/parse_regressor.cc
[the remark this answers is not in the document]
rajan-chari: How about pre_save_load()?  Makes it more generic and hints at when vw might call this function.

vowpalwabbit/core/src/reductions/automl.cc
```diff
@@ -92,16 +92,65 @@ void finish_example(VW::workspace& all, automl<CMType>& data, VW::multi_ex& ec)
...
 }
 
+template <typename CMType>
+void convert_to_single_model(VW::workspace& all, automl<CMType>& data)
+{
+  options_i& options = *all.options;
+  if (!options.was_supplied("predict_only_model")) { return; }
+  // Clear non-champ weights first
+
+  std::swap(*data.cm->_gd_normalized, data.cm->per_live_model_state_double[0]);
```
rajan-chari, line 102: The rest of this is generic and would work with other non-gd base reductions.  Would be good to fix gd and cb_adf so they save per model state for these things.  I realize that's big in scope so add an issue to the backlog?

vowpalwabbit/core/src/array_parameters_dense.cc
```diff
@@ -89,6 +89,19 @@ void dense_parameters::clear_offset(const size_t offset, const size_t params_per
   }
 }
 
+void dense_parameters::adjust_weights_single_model(const size_t params_per_problem)
```
rajan-chari, line 92: Cool.

rajan-chari: Would be easy to change this to save model of any index.

rajan-chari: More comments here would be helpful.  How it works in conjunction with gd.save_load would be also useful for people looking at this code.  The reason for swap is subtle, a comment there would also be useful.

rajan-chari (approved):
Looks good.  Thanks.

vowpalwabbit/core/src/reductions/automl.cc
lalo: Two active PRs this one https://github.com/VowpalWabbit/vowpal_wabbit/pull/4023 blocked on search.cc bug https://github.com/VowpalWabbit/vowpal_wabbit/pull/4031

vowpalwabbit/core/src/array_parameters_dense.cc
bassmang: Added a comment explaining the main operation of changing weights. I removed added a model index argument and removed to swap because that relied on the assumption that the other models were zero'ed out

vowpalwabbit/core/src/parse_regressor.cc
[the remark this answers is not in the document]
jackgerrits: > How about pre_save_load()? Makes it more generic and hints at when vw might call this function.

I like that, it makes it clear when during the lifecycle the hook is called

test/unit_test/simulator.cc
```diff
@@ -12,13 +12,13 @@
 
 namespace simulator
 {
-cb_sim::cb_sim(uint64_t seed)
+cb_sim::cb_sim(uint64_t seed, bool use_default_ns)
```
lalo, line 15: do we still need use_default_ns?

vowpalwabbit/core/src/reductions/automl.cc
```diff
@@ -92,14 +92,66 @@ void finish_example(VW::workspace& all, automl<CMType>& data, VW::multi_ex& ec)
...
+  std::swap(*data.cm->_cb_adf_event_sum, data.cm->per_live_model_state_uint64[0]);
+  std::swap(*data.cm->_cb_adf_action_sum, data.cm->per_live_model_state_uint64[1]);
+
+  // Clear all non-champ weights
+  clear_non_champ_weights(data.cm->weights, data.cm->estimators.size(), data.cm->wpp);
+
+  // Adjust champ weights to new single-model space
+  data.cm->weights.adjust_weights_single_model(data.cm->wpp, 0);
+
+  std::set<std::string> aml_opts = {"automl", "global_lease", "cm_type", "priority_type", "priority_challengers",
```
lalo, line 114: add assert in the setup func of automl ensuring the count stays constant with a comment saying you have to update this bit over here and they always have to match to avoid future regressions.

test/core.vwtest.json
```diff
@@ -5486,5 +5472,71 @@
     "input_files": [
       "train-sets/explore_eval_egreedy.dat"
     ]
+  },
+  {
+    "id": 425,
+    "desc": "Spin off automl model to lower bitsize to cb_explore",
```
lalo, line 5478: add cubic test, or is this only for quadratics for now?

vowpalwabbit/core/src/reductions/automl.cc
bassmang: I was actually able to get rid of this by looking at the option group instead

test/core.vwtest.json
bassmang: Cubics should work, I can add a test for that

bassmang: add a unit test here
https://github.com/VowpalWabbit/vowpal_wabbit/pull/4279/files#diff-9932fc58f28480ef64a94f6fc4cb26b0b50e7d94c621a3edb00fd0131fcc14dfR400

test/core.vwtest.json
```diff
@@ -5486,5 +5472,71 @@
...
+      "stderr": "train-sets/ref/spin_off_aml_load.stderr",
+      "aml_spinoff_cb.inv": "pred-sets/ref/aml_spinoff.inv"
+    },
+    "input_files": [
+      "train-sets/automl_spin_off.txt",
+      "models/aml_spin_off_cb.m"
+    ],
+    "depends_on": [
+      427
+    ]
```
olgavrou, line 5528: it would be good if these model diffs were to happen with larger files with more features and interactions to allow for hash collisions to happen and check this more thoroughly (something like the `ccb_lots_of_interactions.dat` file)

vowpalwabbit/core/src/array_parameters_dense.cc
```diff
@@ -89,6 +89,26 @@ void dense_parameters::clear_offset(const size_t offset, const size_t params_per
...
+  {
+    uint32_t cb_ind = index / params_per_problem;
+    for (uint64_t stride_ind = 0; stride_ind < stride(); ++stride_ind)
+    {
+      if (_begin[index + stride() * model_num + stride_ind] != 0.0f)
+      {
+        // Move all strided weights of selected model to smaller weights array. Zero out
+        // previous weights.
+        _begin[cb_ind + stride_ind] = _begin[index + stride() * model_num + stride_ind];
+        _begin[index + stride() * model_num + stride_ind] = 0.f;
```
olgavrou, line 106: we are zero-ing out previous weights, but there is a cli option to set `initial_weight` to something other than zero, do we care about that in automl?

vowpalwabbit/core/src/array_parameters_dense.cc
```diff
@@ -89,6 +89,26 @@ void dense_parameters::clear_offset(const size_t offset, const size_t params_per
   }
 }
 
+void dense_parameters::adjust_weights_single_model(const size_t params_per_problem, const size_t model_num)
```
olgavrou, line 92: this method could also be unit tested in `weights_test.cc` seems like a prime candidate for some rigorous unit testing

vowpalwabbit/core/src/array_parameters_dense.cc
bassmang: that shouldnt matter since it shift over all non-zero features

test/core.vwtest.json
bassmang: this feature isnt compatable with ccb yet. Also it would be difficult to compare these in runtests since each interaction would need to be enumerated in the command line when comparing to a standard model

olgavrou: I wasn't suggesting we use the ccb file but a file like that one where there are multiple namespaces to explore and is a bit more complex than the one exercised here and could potentially surface issues

bassmang: updated the test file, now it has 290k interacted features across 6 namespaces
https://github.com/VowpalWabbit/vowpal_wabbit/pull/4279/commits/4fb69a8de60d5accab278b3bece8f9db9c820c9d#diff-94d0ac7a20310e17e8f59703288b6806f838c0d36c8893ac70385c2227ae62d8R1

vowpalwabbit/core/src/array_parameters_dense.cc
bassmang: The weights are tested here:
https://github.com/VowpalWabbit/vowpal_wabbit/pull/4279/files#diff-9932fc58f28480ef64a94f6fc4cb26b0b50e7d94c621a3edb00fd0131fcc14dfR333

It doesn't make much sense to test this without initializing weights with a standard and automl model since it converts one to the other

test/core.vwtest.json
olgavrou: cool

lalo (approved):

PR #4281, opened 2022-11-14
feat: Add avx2 implementation for one pass svd in large action spaces.

zwd-ms:


vowpalwabbit/core/src/reductions/cb/details/large_action/compute_dot_prod_avx2.cc
```diff
@@ -0,0 +1,176 @@
...
+      // Handle tail of the loop using scalar implementation.
+      compute1(features.values[i], features.indices[i], offset, weights_mask, column_index, seed, sum);
+    }
+  }
+  const auto& red_features = ex->ex_reduction_features.template get<VW::large_action_space::las_reduction_features>();
+  const auto& interactions =
+      red_features.generated_interactions ? *red_features.generated_interactions : *ex->interactions;
+  for (const auto& ns : interactions)
+  {
+    // TODO: other interactions, ignores & extent_interactions. The following only handles quadratics.
```
olgavrou, line 131: should we emit some sort of warning if users select LAS + simd + ignore/extend/cubics/other instead of silently ignoring?

olgavrou (approved):
approve with suggestion

rami-lv:
@zwd-ms  Did you benchmark the performance improvement of using `avx512` I would be so glad if you can share the numbers.

PR #4285, opened 2022-11-21
feat: explore eval target rate

olgavrou:
replacing block count with target rate

`rate_target` class is trying to estimate what multiplier to use to achieve the desired target rate by providing a multiplier either under or over the desired rejection rate

The other important change is that we don't just report the number of updated examples but the weighted number of updated examples. The reason is that if the threshold is > 1 then we change the weight of the example when learning to reflect that, so we need to update it in the number of examples that vw has "seen" that is reported in the end too


vowpalwabbit/core/src/reductions/explore_eval.cc
```diff
@@ -27,6 +27,53 @@ using namespace VW::config;
...
+    _sum_p += p;
+    _t++;
+  }
+
+public:
+  rate_target() : _target_rate(1), _sum_p(0), _t(0) {}
+
+  void set_target_rate(float rate) { _target_rate = rate; }
+
+  float get_target_rate() { return _target_rate; }
```
ataymano, line 63: const?

ataymano (approved):

PR #4288, opened 2022-11-24
Arm64 performance optimizations 

rami-lv:
Recently at @liftoff we have been leveraging arm64 machines to train our models using vowpal wabbit.
Using arm64 in our infrastructure has helped us reduce the cost of our ML infrastructure. 

To get the most out of vowpal wabbit on these machines, we added arm64 compiler optimization and transported the SIMD instructions used in vowpal wabbit via [sse2neon](https://github.com/DLTcollab/sse2neon).  To do that, we followed [AWS guide](https://github.com/aws/aws-graviton-getting-started/blob/main/c-c%2B%2B.md#enabling-arm-architecture-specific-features) on how to optimize builds for arm64 machines, there are probably more optimizations to apply which require a deeper knowledge of the training algorithm. 

These optimizations improved our ML pipeline time by roughly 20% (the pipeline contains steps other than training with vowpal wabbit so more experiments should be conducted if you are interested in getting the improvement on vowpal wabbit). 

The proper solution would be to fill the placeholder in `lda_core.cc` with the corresponding instructions but that would require a deeper understanding of the code.

### DIFF
--- a/cmake/VWFlags.cmake
+++ b/cmake/VWFlags.cmake
@@ -18,9 +18,14 @@ if("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64")
   endif()
 endif()
 
+set(LINUX_ARM64_OPT_FLAGS "")
+if("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "aarch64|arm64|ARM64")
+  set(LINUX_ARM64_OPT_FLAGS -mcpu=neoverse-n1)
+endif()
+
 # Add -ffast-math for speed, remove for testability.
 # no-stack-check is added to mitigate stack alignment issue on Catalina where there is a bug with aligning stack-check instructions, and stack-check became default option
-set(LINUX_RELEASE_CONFIG -fno-strict-aliasing ${LINUX_X86_64_OPT_FLAGS} -fno-stack-check -fomit-frame-pointer)
+set(LINUX_RELEASE_CONFIG -fno-strict-aliasing ${LINUX_X86_64_OPT_FLAGS} ${LINUX_ARM64_OPT_FLAGS} -fno-stack-check -fomit-frame-pointer)
 set(LINUX_DEBUG_CONFIG -fno-stack-check)
 
 #Use default visiblity on UNIX otherwise a lot of the C++ symbols end up for exported and interpose'able
--- a/vowpalwabbit/core/src/reductions/lda_core.cc
+++ b/vowpalwabbit/core/src/reductions/lda_core.cc
@@ -32,6 +32,10 @@ VW_WARNING_STATE_POP
 #include "vw/core/vw_versions.h"
 #include "vw/io/logger.h"
 
+#if defined(__ARM_NEON)
+#include <sse2neon/sse2neon.h>
+#endif
+
 #include <algorithm>
 #include <cassert>
 #include <cmath>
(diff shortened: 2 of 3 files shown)


rami-lv:
I couldn't properly add `sse2neno` as an external dependency, I would appreciate some help on that.

rami-lv:
@zwd-ms Thanks for the suggestion I didn't know about the project. 
However, integrating SIMDe means that we have to rename the intrinsics calls in `lda_core.cc` to match those of SIMDe which might be not preferable for the maintainers. 
Also, SIMDe has merged all `sse2neon` code so there is no advantage of using one or another. 

In my opinion, unless the maintainers decide to integrate SIMDe this should be temporary, and the long-term solution is to actually implement the corresponding NEON logic. 

As for `AVX2` I couldn't find any usage of AVX intrinsics through the code, did you?

rami-lv:
> As for `AVX2` I couldn't find any usage of AVX intrinsics through the code, did you?

Never mind I just found them.

cmake/VWFlags.cmake
```diff
@@ -18,9 +18,14 @@ if("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64")
   endif()
 endif()
 
+set(LINUX_ARM64_OPT_FLAGS "")
+if("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "aarch64|arm64|ARM64")
+  set(LINUX_ARM64_OPT_FLAGS -mcpu=neoverse-n1)
+endif()
```
jackgerrits, line 24: This is specific to this CPU, so we should not be adding it whenever we encounter an arm CPU. Generally for specific optimization flags like this it is better to add them when configuring your build.

So, in this case we should remove this but in your build you would add the following to your command line to achieve the same outcome:
```
-DCMAKE_CXX_FLAGS_RELEASE="-mcpu=neoverse-n1"
```

vowpalwabbit/core/src/reductions/lda_core.cc
```diff
@@ -32,6 +32,10 @@ VW_WARNING_STATE_POP
 #include "vw/core/vw_versions.h"
 #include "vw/io/logger.h"
 
+#if defined(__ARM_NEON)
+#include <sse2neon/sse2neon.h>
```
jackgerrits, line 36: This has not been made available to the `vw_core` cmake target, so the build won't be able to find this. We also don't have any CI which is going to cause this path to be exercised so we need to be careful.

The include should be added here: https://github.com/VowpalWabbit/vowpal_wabbit/blob/3ee966596575fbce66b9bef8e6d586da5b5a52af/vowpalwabbit/core/CMakeLists.txt#L372

jackgerrits (changes requested):

vowpalwabbit/core/src/reductions/lda_core.cc
rami-lv: >  I can push a commit to the branch with this change 

That would be nice of you

cmake/VWFlags.cmake
jackgerrits: I'm going to go ahead and push a change removing this piece so we can move forward here.

rami-lv: @jackgerrits FYI, I added `sse2neon` to vcpkg. Would you prefer that I add the package instead?

vowpalwabbit/core/src/reductions/lda_core.cc
rami-lv: @jackgerrits FYI, I added `sse2neon` to vcpkg. Would you prefer that I add the package instead?

jackgerrits: Thanks for adding that, that's super helpful! There was a tiny issue with the installed path, I went ahead and opened a fix here https://github.com/microsoft/vcpkg/pull/28268. When that gets merged I will update this PR to consume.

jackgerrits: I'll go ahead and merge this one and we can update to consume vcpkg later.

jackgerrits (approved):

PR #4296, opened 2022-12-01
refactor: implement scaffold for finish_example split and POC migrations

jackgerrits:


vowpalwabbit/core/include/vw/core/learner.h
```diff
@@ -366,19 +388,86 @@ class learner
...
+    }
   }
 
   inline void NO_SANITIZE_UNDEFINED print_example(VW::workspace& all, E& ec)
   {
     debug_log_message(ec, "print_example");
-
-    if (_finish_example_fd.print_example_f == nullptr)
-      THROW("fatal: learner did not register print example fn: " + _name);
+    if (!has_legacy_print_example()) { THROW("fatal: learner did not register print example fn: " + _name); }
```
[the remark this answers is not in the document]
jackgerrits: Yeah it will be superseded by output/record but until that happens I figured marking it legacy indicates its fate

vowpalwabbit/core/src/simple_label.cc
```diff
@@ -55,6 +55,22 @@ void VW::details::return_simple_example(VW::workspace& all, void*, VW::example&
   VW::finish_example(all, ec);
 }
 
+void VW::details::record_stats_simple_label(
```
olgavrou, line 58: maybe `update_stats` instead of `record_stats`? it is in sync with `sd.update` and it is clear that it has side effects on a shared object since record could just mean record the stats somewhere

vowpalwabbit/core/src/simple_label.cc
```diff
@@ -55,6 +55,22 @@ void VW::details::return_simple_example(VW::workspace& all, void*, VW::example&
...
+    const VW::workspace& /* all */, shared_data& sd, const VW::example& ec, VW::io::logger& /* logger */)
+{
+  const auto& ld = ec.l.simple;
+  sd.update(ec.test_only, ld.label != FLT_MAX, ec.loss, ec.weight, ec.get_num_features());
+  if (ld.label != FLT_MAX && !ec.test_only) { sd.weighted_labels += (static_cast<double>(ld.label)) * ec.weight; }
+}
+
+void VW::details::output_example_simple_label(
```
olgavrou, line 66: is this function specific to the label?

vowpalwabbit/core/src/simple_label.cc
jackgerrits: `update_stats` is better I agree

vowpalwabbit/core/src/simple_label.cc
jackgerrits: Yes, this is a general implementation for simple labels

olgavrou (approved):

PR #4298, opened 2022-12-01
feat: Enforce minimum probability for squarecb and update impl

jackgerrits:
Update implementation of `exploration::enforce_minimum_probability`

Mixture implementation can be found here: https://github.com/VowpalWabbit/vowpal_wabbit/compare/master...canondetortugas:vowpal_wabbit:uniform_exploration


marco-rossi29 (approved):

vowpalwabbit/core/src/reductions/cb/cb_explore_adf_squarecb.cc
```diff
@@ -225,6 +226,9 @@ void cb_explore_adf_squarecb::predict(multi_learner& base, VW::multi_ex& example
       total_weight += pa;
     }
     preds[a_min].score = 1.f - total_weight;
+
+    exploration::enforce_minimum_probability(_epsilon, true, begin_scores(preds), end_scores(preds));
+    // TODO: Do we need to sort here?
```
lalo, line 231: is this comment still valid?

jackgerrits: I think it can be removed, just need to check

lalo (approved):

PR #4308, opened 2022-12-05
refactor: split apart output and progressive log

jackgerrits:


vowpalwabbit/core/src/reductions/bfgs.cc
```diff
@@ -1200,7 +1201,8 @@ base_learner* VW::reductions::bfgs_setup(VW::setup_base_i& stack_builder)
                         .set_save_load(save_load)
                         .set_init_driver(init_driver)
                         .set_end_pass(end_pass)
-                        .set_output_example(VW::details::output_example_simple_label<bfgs>)
+                        .set_output_example_prediction(VW::details::output_example_prediction_simple_label<bfgs>)
```
olgavrou, line 1204: Should we use either `output` or `print` for both functions here?

jackgerrits: set_output_example_prediction usually goes to a file, whereas the update usually goes to stdout

olgavrou: true dat

olgavrou (approved):

PR #4309, opened 2022-12-05
refactor: split cache parser into separate lib

jackgerrits:


vowpalwabbit/core/src/example.cc
```diff
@@ -331,9 +331,9 @@ size_t read_model_field(io_buf& io, flat_example& fe, VW::label_parser& lbl_pars
   bytes += read_model_field(io, fe.num_features);
   bytes += read_model_field(io, fe.total_sum_feat_sq);
   unsigned char index = 0;
-  bytes += ::VW::details::read_cached_index(io, index);
+  bytes += ::VW::parsers::cache::details::read_cached_index(io, index);
   bool sorted = true;
-  bytes += ::VW::details::read_cached_features(io, fe.fs, sorted);
+  bytes += ::VW::parsers::cache::details::read_cached_features(io, fe.fs, sorted);
```
lalo, line 336: is the cache parser independent from the concept of single line and multi line examples?

jackgerrits: Cache parser is single line parser with sentinels for newlines that must be handled externally to it.

lalo (approved):

PR #4315, opened 2022-12-06
refactor: allow reduction to control print frequency

jackgerrits:
I've gone back and forth on this one, but because of things like search and audit_regressor I think we need to leave the control with the reduction to decide if it wants to print an update.


lalo (approved):

vowpalwabbit/core/src/reductions/conditional_contextual_bandit.cc
```diff
@@ -580,7 +580,10 @@ void output_example_prediction_ccb(
 void print_update_ccb(VW::workspace& all, shared_data& /* sd */, const ccb_data& data, const VW::multi_ex& ec_seq,
     VW::io::logger& /* unused */)
 {
-  if (!ec_seq.empty() && !data.no_pred)
+  const bool should_print_driver_update =
```
lalo, line 583: maybe throw this in SD as a default print update cadence

PR #4324, opened 2022-12-09
refactor: migrate topk finish_example

jackgerrits:


jackgerrits:
This one requires more thought:
- What is the correct prediction type here?
- What is the correct way to update shared data and do the printout here?

jackgerrits:
> This one requires more thought:
> 
> * What is the correct prediction type here?
> * What is the correct way to update shared data and do the printout here?

In general, the changes here maintain semantics while fixing some of the contract breaks.

- The reduction used to pass back predictions as scalars on each example, this has been converted to `scalars` on the 0th example.
- Updating shared data is the same as before, print update does a single print update per predict vs one per example as was done before

PR #4327, opened 2022-12-12
fix: [automl] config oracle edge cases

lalo:


vowpalwabbit/core/src/reductions/details/automl/automl_oracle.cc
```diff
@@ -187,6 +187,9 @@ void config_oracle<oracle_impl>::insert_config(set_ns_list_t&& new_elements,
 void oracle_rand_impl::gen_ns_groupings_at(const std::string& interaction_type,
     const interaction_vec_t& champ_interactions, const size_t, set_ns_list_t& new_elements, config_type)
 {
+  // If champ has no interactions, we can't generate any new configs.
+  if (champ_interactions.empty()) { return; }
```
bassmang, line 191: What's the point of this? If the champ has no interactions, we should still be able to generate new configs with additional namespaces, namely when the champ is initiated with no interactions

lalo: this would only be empty if you are excluding all the namespaces, so you don't have anything else to add to be excluded

lalo: this is only for the random impl

lalo: not longer applies, see latest

bassmang (approved):

PR #4330, opened 2022-12-12
refactor: Migrate cb_explore_adf reductions finish functions

peterychang:


vowpalwabbit/core/src/reductions/cb/cb_explore_adf_greedy.cc
```diff
@@ -147,8 +147,9 @@ VW::LEARNER::base_learner* VW::reductions::cb_explore_adf_greedy_setup(VW::setup
                 .set_input_prediction_type(VW::prediction_type_t::ACTION_SCORES)
                 .set_output_prediction_type(VW::prediction_type_t::ACTION_PROBS)
                 .set_params_per_weight(problem_multiplier)
-                .set_finish_example(explore_type::finish_multiline_example)
-                .set_print_example(explore_type::print_multiline_example)
```
jackgerrits, line 151: Until automl+others have been updated to use the new functions we will need to keep in place a `print_example` function which just calls `update_stats`, `output_example_prediction` and `print_update`

jackgerrits (approved):

PR #4343, opened 2022-12-13
feat: use strong type for no pred

jackgerrits:


olgavrou (approved):
wondering if something triggered this change?

jackgerrits:
No particular trigger, just seemed like an opportunity to be safer.

PR #4344, opened 2022-12-13
refactor: [versioning] split split options_config/setup 1

bassmang:


vowpalwabbit/core/src/reductions/audit_regressor.cc
```diff
@@ -28,10 +28,10 @@ static constexpr std::array<VW::column_definition, NUM_COLS> AUDIT_REGRESSOR_COL
 static const std::array<std::string, NUM_COLS> AUDIT_REGRESSOR_HEADER = {
     "example\ncounter", "values\naudited", "total\nprogress"};
 
-class audit_regressor_data
+class audit_regressor_data_obj
```
jackgerrits, line 31: A class shouldn't be named obj

vowpalwabbit/core/src/reductions/autolink.cc
```diff
@@ -82,17 +82,30 @@ void predict_or_learn(autolink& b, VW::LEARNER::single_learner& base, VW::exampl
...
 
-VW::LEARNER::base_learner* VW::reductions::autolink_setup(VW::setup_base_i& stack_builder)
+struct options_autolink_v1
 {
-  options_i& options = *stack_builder.get_options();
-  VW::workspace& all = *stack_builder.get_all_pointer();
   uint32_t d;
+};
+
+std::unique_ptr<options_autolink_v1> get_autolink_options_instance(
```
jackgerrits, line 90: What is the contract around this function? Not only does it define the options and parse them it seems it also needs to make the decision if the reduction is enabled or not?

vowpalwabbit/core/src/reductions/baseline_challenger_cb.cc
```diff
@@ -215,15 +220,27 @@ VW::LEARNER::base_learner* VW::reductions::baseline_challenger_cb_setup(VW::setu
 
   if (!options.was_supplied("cb_adf")) { THROW("cb_challenger requires cb_explore_adf or cb_adf"); }
 
-  bool emit_metrics = options.was_supplied("extra_metrics");
-  auto data = VW::make_unique<baseline_challenger_data>(emit_metrics, alpha, tau);
+  baseline_challenger_cb_opts->emit_metrics = options.was_supplied("extra_metrics");
```
jackgerrits, line 223: If this reduction is determining whether to emit metrics or not based on the options then it should be part of its option description

vowpalwabbit/core/src/reductions/bfgs.cc
```diff
@@ -1111,66 +1111,91 @@ void save_load(bfgs& b, io_buf& model_file, bool read, bool text)
...
-  if (bfgs_enabled)
+  bfgs_opts->bfgs_enabled = options.add_parse_and_check_necessary(bfgs_options);
+  if (!conjugate_gradient_enabled && !bfgs_opts->bfgs_enabled) { return nullptr; }
+  if (conjugate_gradient_enabled && bfgs_opts->bfgs_enabled)
+  {
+    THROW("'conjugate_gradient' and 'bfgs' cannot be used together.");
+  }
+  if (!all.holdout_set_off)
+  {
+    bfgs_opts->early_terminate = options.get_typed_option<uint64_t>("early_terminate").value();
```
jackgerrits, line 1155: we want to try and minimize these "digging out the value" operations we need to do. If a reduction has a dependency on an option it is best to put it in its option group

vowpalwabbit/core/src/reductions/binary.cc
```diff
@@ -52,16 +52,30 @@ void predict_or_learn(binary_data& data, VW::LEARNER::single_learner& base, VW::
...
 {
-  options_i& options = *stack_builder.get_options();
-
   bool binary = false;
+};
+
+std::unique_ptr<options_binary_v1> get_binary_options_instance(
+    const VW::workspace&, VW::io::logger&, options_i& options)
+{
+  auto binary_opts = VW::make_unique<options_binary_v1>();
```
jackgerrits, line 63: IMO we should allocate the options object after calling `add_parse_and_check_necessary` and then initialize and return the object all in one go. We create and delete a lot of options objects when we dont need to.

jackgerrits (changes requested):

vowpalwabbit/core/src/reductions/baseline_challenger_cb.cc
olgavrou: isn't metrics one of those options that is universal across reductions?

jackgerrits: Yeah, epsilon is too and is part of the option group so we aren't consistent. I don't feel super confident about these cross-reduction global options because right now they apply to all reductions. However, if the concept of option namespacing is brought in then the global/cross reduction options no longer work as before

vowpalwabbit/core/src/reductions/autolink.cc
bassmang: Yes it is supposed to define and parse the options, then return the finished configuration, or `nullptr` if the reduction is not enabled

vowpalwabbit/core/src/reductions/baseline_challenger_cb.cc
olgavrou: epsilon is reduction-specific it is exploration only, this is more close to `random_seed`

jackgerrits: epsilon is still defined in 11 reductions, I think at that point its no longer all that specific

olgavrou: you can't say that epsilon is universal though even though 11 exploration reductions use it, where as metrics can be used by every single reduction if they wish to do so

if we are going to make the change to move the universal reduction options into every reduction I think it should be done in a separate PR

vowpalwabbit/core/src/reductions/audit_regressor.cc
bassmang: I updated this because I have been naming the data instance `blah_data`, I can update the class name to `audit_regressor` which is more standard

vowpalwabbit/core/src/reductions/baseline_challenger_cb.cc
jackgerrits: Yeah it is out of scope of this and I am not saying we even should do it. I'm mostly pointing out that unless we put an option into a reductions option description there is no way for a user to know that the reduction can use that option

olgavrou: that is true

rajan-chari: Probably out of scope for this PR but wanted to mention here...

Some options seem to be vw_driver options... for example, threading, logging, metrics, prediction_file, input_file etc.  And some that are vw_ml (reduction_lib?) options, for example epsilon.  

Jacob mentioned possible support for vw_ml options, where individual reductions can have their own epsilon values, if specified, without any explicit code from the reduction author.

bassmang: I think with `extra_metrics` specifically it makes sense to parse the option elsewhere globally and add it to the all/logger object. I can open this in a separate PR, thoughts on this?

jackgerrits: Totally agree on clearer distinction between driver options and reduction options.

We can make metrics more of a formalized concept that reductions can subscribe to. I think that would be good. However, what I'd really like to avoid is just yet another boolean in the `workspace` object. There should be some level of encapsulation with it.

olgavrou: I like the idea of subscribing; something like a callback function registration that can be called after setup and activate any global functionality (i.e. callbacks create metrics related objects where they are needed etc)

communicating via workspace is temptingly easy :)

bassmang: https://github.com/VowpalWabbit/vowpal_wabbit/pull/4354
Opened this which does exactly the `workspace` fix. @jackgerrits how would you suggest we further encapsulate this? It seems like a more global field than many other members in workspace

jackgerrits: Just adding another boolean to `workspace` isn't really getting us anywhere in my opinion. I think unless we have a more holistic change we shouldn't change anything and just leave it as is. It takes a lot of work to get something out of the god object I really want to avoid doing it without a really good reason.

As for how we further encapsulate this I don't have a design or direction in mind at the moment.

vowpalwabbit/core/src/reductions/bfgs.cc
bassmang: I agree these should be in the option group

vowpalwabbit/core/src/reductions/baseline_challenger_cb.cc
bassmang: This has been resolved with the `metrics_collector`

vowpalwabbit/core/src/reductions/bfgs.cc
```diff
@@ -1141,7 +1141,7 @@ std::unique_ptr<options_bfgs_v1> get_bfgs_options_instance(
                .default_value(0.001f)
                .help("Termination threshold"))
       .add(make_option("early_terminate", bfgs_opts->early_terminate)
-               .default_value(0)
+               .default_value(3)
```
jackgerrits, line 1144: Why change this?

bassmang: The default in parse_args is 3 and before this was using 
`options.get_typed_option<uint64_t>("early_terminate").value();`
which would default to 3, so this keeps behavior the same

vowpalwabbit/core/src/reductions/binary.cc
bassmang: Not sure what you're suggesting here. `add_parse_and_check_necessary` has to be called after the option group is defined and the options are assigned to variables. Where would we assign the options if not the options_config object?

bassmang:
NOTE: Revisit as versioning progresses

PR #4347, opened 2022-12-14
refactor: migrate cbzo finish_example

jackgerrits:


vowpalwabbit/core/src/reductions/cbzo.cc
```diff
@@ -346,8 +350,8 @@ base_learner* VW::reductions::cbzo_setup(VW::setup_base_i& stack_builder)
   if (options.was_supplied("feature_mask")) { feature_mask_off = false; }
 
   uint8_t policy;
-  if (policy_str.compare("constant") == 0) { policy = CONSTANT_POLICY; }
-  else if (policy_str.compare("linear") == 0) { policy = LINEAR_POLICY; }
+  if (policy_str == "constant") { policy = CONSTANT_POLICY; }
```
lalo, line 353: I think this check already happens inside of the one_of parsing, but not sure.

lalo (approved):

vowpalwabbit/core/src/reductions/cbzo.cc
jackgerrits: Yeah the one_of check does confirm that but we still need to convert the string to the int here.

PR #4356, opened 2022-12-16
refactor: move text parser into its own lib

olgavrou:


vowpalwabbit/text_parser/include/vw/text_parser/parse_example_text.h
```diff
@@ -0,0 +1,29 @@
...
+
+namespace VW
+{
+namespace parsers
+{
+namespace text
+{
+void substring_to_example(VW::workspace* all, VW::example* ae, VW::string_view example);
+
+void read_line(VW::workspace& all, example* ex, VW::string_view line);  // read example from the line.
```
jackgerrits, line 20: Whats the difference between `substring_to_example` and `read_line`? Can we consolidate?

vowpalwabbit/text_parser/include/vw/text_parser/parse_example_text.h
```diff
@@ -0,0 +1,29 @@
...
+void read_line(VW::workspace& all, example* ex, const char* line);      // read example from the line.
+void read_lines(VW::workspace* all, const char* line, size_t len,
+    VW::multi_ex& examples);  // read examples from the new line separated strings.
+
+int read_features_string(VW::workspace* all, io_buf& buf, VW::multi_ex& examples);
+size_t read_features(io_buf& buf, char*& line, size_t& num_chars);
```
jackgerrits, line 26: These are probably details

vowpalwabbit/text_parser/include/vw/text_parser/parse_example_text.h
```diff
@@ -0,0 +1,29 @@
...
+namespace VW
+{
+namespace parsers
+{
+namespace text
+{
+void substring_to_example(VW::workspace* all, VW::example* ae, VW::string_view example);
+
+void read_line(VW::workspace& all, example* ex, VW::string_view line);  // read example from the line.
+void read_line(VW::workspace& all, example* ex, const char* line);      // read example from the line.
```
jackgerrits, line 21: We should remove this one and just require using the string view overload

vowpalwabbit/text_parser/src/parse_example_text.cc
```diff
@@ -54,6 +54,7 @@ int read_features_string(VW::workspace* all, io_buf& buf, VW::multi_ex& examples
   return static_cast<int>(num_bytes_consumed);
 }
 
+// TODO what to do about this is this details? probably yes
 template <bool audit>
```
jackgerrits, line 58: This class is only used in the cc file, so it should be wrapped in an anonymous namespace so its TU local

vowpalwabbit/text_parser/include/vw/text_parser/parse_example_text.h
olgavrou: I agree for `read_features` (only used in the json parser)

I am not sure about `read_features_string`, that and `read_line` is what is set in the `parser` and used to parse, is that a detail?

vowpalwabbit/text_parser/include/vw/text_parser/parse_example_text.h
olgavrou: `read_line` reads until a newline is found and then passes the string to `substring_to_example`

`substring_to_example` is used by the throughput tool and internally in other parser methods, could probably just be made `details` and only expose `read_line`

vowpalwabbit/text_parser/include/vw/text_parser/parse_example_text.h
jackgerrits: Ah I see its the "entrypoint". I guess it should not be a detail. However, it's purpose is mainly to facilitate the driver not necessary a user of the C++ API to parse input

vowpalwabbit/text_parser/include/vw/text_parser/parse_example_text.h
olgavrou: plain remove or deprecate? I'm fine with plain remove

jackgerrits: We don't need to deprecate it because this file is brand new

olgavrou: I meant from the original file

vowpalwabbit/text_parser/include/vw/text_parser/parse_example_text.h
olgavrou: yup, hence the "I'm not sure what we should do with this", maybe just leave it for now until we consolidate the parser interfaces

jackgerrits: I think the default for when we aren't sure should be to put it in details to be honest. It at least conveys that we consider this a detail for now

vowpalwabbit/text_parser/include/vw/text_parser/parse_example_text.h
jackgerrits: The approach I have been taking is if the thing moves to a non-details namespace then I leave a compatibility func or alias. If it moves to details I don't worry about that. So in this case since it isn't details leaving the func in is safer

olgavrou: yeah I agree I just removed

vowpalwabbit/text_parser/include/vw/text_parser/parse_example_text.h
olgavrou: the more I think about it the more I see it as in line with `read_line`, one just takes an `io_buf` input, the other a `string_view`, and I don't think either is a detail

vowpalwabbit/text_parser/include/vw/text_parser/parse_example_text.h
olgavrou: made a string_view version of read_lines too while we are at it

jackgerrits (approved):

PR #4360, opened 2022-12-16
refactor: cb_explore finish function

peterychang:


vowpalwabbit/core/src/reductions/cb/cb_explore.cc
```diff
@@ -238,9 +238,10 @@ void predict_or_learn_cover(cb_explore& data, single_learner& base, VW::example&
   ec.l.cs = VW::cs_label{};
 }
 
-void print_update_cb_explore(VW::workspace& all, bool is_test, VW::example& ec, std::stringstream& pred_string)
+void print_update_cb_explore(
```
jackgerrits, line 241: It looks like there are two `print_update_cb_explore` is one of them no longer needed?

jackgerrits: Oh I see, the other one calls into here.

jackgerrits (approved):

PR #4377, opened 2022-12-20
refactor: [automl/epsilon_decay] integrate robust confidence sequences

bassmang:


test/train-sets/ref/cb_similar_aml_spinoff_save.stderr
```diff
@@ -13,15 +13,15 @@ Input label = CB
 Output pred = ACTION_PROBS
 average  since         example        example        current        current  current
 loss     last          counter         weight          label        predict features
-0.333333 0.333333            1            1.0          0:1:1         0:0.33     8436
-0.175000 0.016667            2            2.0          1:1:1         2:0.97    18319
-0.333333 0.491667            4            4.0          1:1:1         0:0.97    23278
-0.293750 0.254167            8            8.0          0:1:1         2:0.97    13862
-0.273958 0.254167           16           16.0          0:1:1         2:0.97     2249
+-0.33333 -0.33333            1            1.0         0:-1:1         0:0.33     8436
```
lalo, line 16: why would loss change on the first example if the CS is not used for loss?

bassmang: I had to change the rewards because the new estimator can only use rewards between 0 and 1, implying a costs between -1 and 0

lalo (approved):

vowpalwabbit/core/tests/automl_test.cc
```diff
@@ -103,7 +102,7 @@ get_automl_data(VW::workspace& all)
   all.l->get_enabled_reductions(e_r);
   if (std::find(e_r.begin(), e_r.end(), "automl") == e_r.end())
   {
-    BOOST_FAIL("automl not found in enabled reductions");
+    EXPECT_THROW("automl not found in enabled reductions", char*);
```
jackgerrits:
```suggestion
    FAIL() << "automl not found in enabled reductions";
```

jackgerrits (changes requested):

jackgerrits (changes requested):
Main concern here is that simulator is test code so it should not be distributed in the core library.

vowpalwabbit/core/include/vw/core/simulator.h
```diff
@@ -4,11 +4,9 @@
 
 #pragma once
```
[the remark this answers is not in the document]
bassmang: moved to core tests

vowpalwabbit/core/tests/automl_test.cc
bassmang: `FAIL()` seems to only work for void functions, updated to `THROW`

jackgerrits (approved):

PR #4378, opened 2022-12-20
Cb to cb adf finish refactor

peterychang:


vowpalwabbit/core/src/reductions/cb/cb_to_cb_adf.cc
```diff
@@ -87,19 +87,12 @@ void predict_or_learn(cb_to_cb_adf& data, multi_learner& base, VW::example& ec)
...
-  }
-  else
-  {
-    c.adf_data.ecs[0]->pred.multiclass = std::move(ec.pred.multiclass);
-    c.adf_learner->print_example(all, c.adf_data.ecs);
-  }
-  VW::finish_example(all, ec);
+  if (c.explore_mode) { c.adf_data.ecs[0]->pred.a_s = std::move(ec.pred.a_s); }
+  else { c.adf_data.ecs[0]->pred.multiclass = std::move(ec.pred.multiclass); }
+  c.adf_learner->print_example(all, c.adf_data.ecs);
```
jackgerrits, line 95: `print_example` as it is today does more than just output prediction. So this output_prediction function should not be calling it.

`cb_to_cb_adf` should implement the 3 (update stats, output pred and print update) and be delegating the calls respectively to the underlying in much the same way it used to do for the the finish->print_example call beforre

peterychang:
#4398

PR #4384, opened 2022-12-21
fix: [automl] update print logic for new oracle

lalo:


test/train-sets/ref/automl_readable_qc.txt
```diff
@@ -0,0 +1,3345 @@
...
+_automl_config_manager_configs[164]_exclusions[0][1] = 105
+_automl_config_manager_configs[164]_exclusions[0][2] = 105
+_automl_config_manager_configs[164]_lease = 4000
+_automl_config_manager_configs[164]_state = New
+_automl_config_manager_configs[164]_type = Interaction
+_automl_config_manager_configs[165]_exclusions.size() = 1
+_automl_config_manager_configs[165]_exclusions[0].size() = 3
+_automl_config_manager_configs[165]_exclusions[0][0] = 105
+_automl_config_manager_configs[165]_exclusions[0][1] = 105
+_automl_config_manager_configs[165]_exclusions[0][2] = 106
```
lalo, line 1355: @bassmang should we just save the configs being used and re-generate from the champ instead of storing all of them? this might get out of control

vowpalwabbit/core/src/reductions/details/automl/automl_util.cc
```diff
@@ -79,13 +79,14 @@ std::string ns_to_str(unsigned char ns)
...
-std::string interaction_vec_t_to_string(
-    const std::vector<std::vector<namespace_index>>& interactions, const std::string& interaction_type)
+std::string interaction_vec_t_to_string(const std::vector<std::vector<namespace_index>>& interactions)
 {
   std::stringstream ss;
   for (const std::vector<VW::namespace_index>& v : interactions)
   {
-    interaction_type == "quadratic" ? ss << "-q " : ss << "-cubic ";
+    if (v.size() == 2) { ss << "-q "; }
+    else if (v.size() == 3) { ss << "-cubic "; }
```
jackgerrits, line 88:
```suggestion
    else if (v.size() == 3) { ss << "--cubic "; }
```

jackgerrits (approved):

test/train-sets/ref/automl_readable_qc.txt
lalo: unrelated to this PR but open question

bassmang: Not sure if this is possible, especially since randomness is used to generate challengers

PR #4395, opened 2022-12-27
fix!: resolve csoaa_ldf prediction return correctness

jackgerrits:
- `csoaa_ldf` probabilities return was being mixed in with other returns and being returned on each example
- multiclass return was also being returned on each example
- This retains the same output format for predictions files
- To reduce the scope of work I didn't migrate the finish functions in this PR
- Some of the compare files were incorrect which should be resolved.

## Breaking change

This is a breaking change for the `csoaa_ldf` reduction when using it in **library mode** (not outputting predictions on command line). This change is being done for correctness.

- When using `--probabilities`:
  - The probabilities are returned in the `scalars` slot of the 0th example instead of in `prob` for each example
  -  Probabilities are no longer returned in addition to the other return type but on their own
- When using default (not `--probabilities` or `--csoaa_rank`) the multiclass prediction used to be returned on each example with the predicted example having the class that was predicted set on it. Now only the 0th example has `multi` set and it's value is the predicted class.


test/train-sets/ref/multilabel_ldf.stderr
```diff
@@ -25,4 +25,4 @@ number of examples = 8
 weighted example sum = 8.000000
 weighted label sum = 0.000000
 average loss = 0.875000
-total feature number = 24
+total feature number = 66
```
olgavrou, line 28: do we understand why this changed?

vowpalwabbit/core/src/reductions/csoaa_ldf.cc
```diff
@@ -493,111 +466,164 @@ void output_example(
...
-  if (VW::is_cs_example_header(head_ec)) { return; }
-
-  all.sd->total_features += head_ec.get_num_features();
-
-  float loss = 0.;
-  VW::v_array<VW::action_score>& preds = head_ec.pred.a_s;
+  size_t num_features = 0;
+  for (const auto* ec : ec_seq)
+  {
+    if (VW::is_cs_example_header(*ec))
```
olgavrou, line 478: won't the features have been merged by feature merger at this point? isn't it enough to just count the number of features in each example?

test/train-sets/ref/multilabel_ldf.stderr
jackgerrits: It was treating each example separately despite being in a multiex and the featuer counting was off as a result

olgavrou (approved):

python/tests/test_pyvw.py
```diff
@@ -104,10 +104,10 @@ def test_prob_prediction_type():
         model.example("2:0.8  | a b c"),
     ]
     model.learn(multi_ex)
-    assert model.get_prediction_type() == vowpalwabbit.PredictionType.PROB
+    assert model.get_prediction_type() == vowpalwabbit.PredictionType.SCALARS
```
lalo, line 107: this diff is the breaking change

jackgerrits: To library callers this is one of them

lalo (approved):

PR #4397, opened 2022-12-27
refactor: migrate cb_adf finish_example

jackgerrits:

PR #4398, opened 2022-12-27
refactor: cb_to_cb_adf finish function

peterychang:
requires #4397


vowpalwabbit/core/include/vw/core/learner.h
```diff
@@ -473,25 +473,43 @@ class learner
     _finish_example_fd.print_example_f(all, _finish_example_fd.data, (void*)&ec);
   }
 
-  inline void NO_SANITIZE_UNDEFINED update_stats(VW::workspace& all, const E& ec)
+  inline void NO_SANITIZE_UNDEFINED update_stats(
```
jackgerrits, line 476: These are public learner functions. `data` should not be exposed as a parameter for the caller.

vowpalwabbit/core/src/reductions/cb/cb_to_cb_adf.cc
```diff
@@ -87,19 +87,46 @@ void predict_or_learn(cb_to_cb_adf& data, multi_learner& base, VW::example& ec)
...
+    c.adf_data.ecs[0]->pred.a_s = ec.pred.a_s;
   }
   else
   {
-    c.adf_data.ecs[0]->pred.multiclass = std::move(ec.pred.multiclass);
-    c.adf_learner->print_example(all, c.adf_data.ecs);
+    c.adf_data.ecs[0]->pred.multiclass = ec.pred.multiclass;
   }
-  VW::finish_example(all, ec);
+  c.adf_learner->update_stats(all, sd, &c, c.adf_data.ecs, logger);
```
jackgerrits, line 101: `adf_learner`'s data is not the c, but its own data object. Leaving data as private in the learner interface will fix this issue

jackgerrits (approved):

PR #4401, opened 2022-12-28
revert: "build: rely on gtest_main target instead of custom"

jackgerrits:
Reverts VowpalWabbit/vowpal_wabbit#4396

This doesn't work on windows which is why the main.cc file workaround is needed.

See here: https://github.com/google/googletest/issues/2157


jackgerrits:
Incorporated into #4399

PR #4404, opened 2022-12-29
refactor: fix namespacing of parameters

jackgerrits:


vowpalwabbit/core/include/vw/core/array_parameters_dense.h
```diff
@@ -9,6 +9,11 @@
 #include <cassert>
 #include <cstdint>
 #include <iterator>
+namespace VW
+{
+
+namespace details
```
peterychang, line 15: I assume the dense parameters/iterator should never be used outside of VW code?

peterychang (approved):

vowpalwabbit/core/include/vw/core/array_parameters_dense.h
jackgerrits: The dense_parameters::iterator typedef is the intended way to access the type

PR #4407, opened 2022-12-29
refactor: Add metrics collector

bassmang:
***NOTE***
Remember to remove metric_output_hooks from RLClientlib  when updading VW and replace with metrics_manager register_callback_function


vowpalwabbit/core/src/metrics_manager.cc
```diff
@@ -0,0 +1,34 @@
...
+}
+
+metric_sink metrics_manager::collect_metrics()
+{
+  VW::metric_sink sink;
+  while (!_metrics_callbacks.empty())
+  {
+    const metrics_callback_fn& callback = _metrics_callbacks.back();
+    callback(sink);
+    _metrics_callbacks.pop_back();
```
jackgerrits, line 30: Why remove the callbacks? If you just iterate them then this can also be a const method.

Keeping them around could for instance allow us to take progressive "snapshots" of metrics over time.

vowpalwabbit/core/src/reductions/baseline_challenger_cb.cc
```diff
@@ -215,8 +217,7 @@ VW::LEARNER::base_learner* VW::reductions::baseline_challenger_cb_setup(VW::setu
 
   if (!options.was_supplied("cb_adf")) { THROW("cb_challenger requires cb_explore_adf or cb_adf"); }
 
-  bool emit_metrics = options.was_supplied("extra_metrics");
-  auto data = VW::make_unique<baseline_challenger_data>(emit_metrics, alpha, tau);
+  auto data = VW::make_unique<baseline_challenger_data>(all.global_metrics.are_metrics_enabled(), alpha, tau);
```
jackgerrits, line 220: Rather than reductions accessing via the all pointer, we could expose the metrics_manager via the `setup_base_i` interface

vowpalwabbit/core/include/vw/core/learner.h
```diff
@@ -388,10 +388,12 @@ class learner
...
   // called when metrics is enabled.  Autorecursive.
-  void NO_SANITIZE_UNDEFINED persist_metrics(metric_sink& metrics)
+  void NO_SANITIZE_UNDEFINED persist_metrics(VW::metrics_manager& manager)
   {
-    _persist_metrics_fd.save_metric_f(_persist_metrics_fd.data, metrics);
-    if (_persist_metrics_fd.base) { _persist_metrics_fd.base->persist_metrics(metrics); }
+    auto metric_f = [this](metric_sink& sink) -> void
+    { _persist_metrics_fd.save_metric_f(_persist_metrics_fd.data, sink); };
+    manager.register_metrics_callback(metric_f);
+    if (_persist_metrics_fd.base) { _persist_metrics_fd.base->persist_metrics(manager); }
```
jackgerrits, line 396: Why change this function? Based on the current design it seems like the metric manager shouldn't be coupled with calling persist_metrics but just consume the output of it.

vowpalwabbit/core/src/reductions/metrics.cc
```diff
@@ -137,38 +138,44 @@ void predict_or_learn(metrics_data& data, T& base, E& ec)
...
+  metrics_manager& manager = all.global_metrics;
+  if (manager.are_metrics_enabled())
   {
-    std::string filename = all.options->get_typed_option<std::string>("extra_metrics").value();
-    VW::metric_sink list_metrics;
-
-    all.l->persist_metrics(list_metrics);
-
-    for (auto& metric_hook : all.metric_output_hooks) { metric_hook(list_metrics); }
+    all.l->persist_metrics(manager);
```
jackgerrits, line 144: This no longer actually collects metrics but registers callbacks. I don't think you should change the semantics of this call.

vowpalwabbit/core/src/reductions/metrics.cc
```diff
@@ -98,8 +97,10 @@ class json_metrics_writer : public VW::metric_sink_visitor
   Writer<FileWriteStream>& _writer;
 };
 
-void list_to_json_file(const std::string& filename, const VW::metric_sink& metrics, VW::io::logger& logger)
+void list_to_json_file(VW::metrics_manager& manager, VW::io::logger& logger)
```
jackgerrits, line 100: I think this function should remain unchanged

vowpalwabbit/core/src/reductions/metrics.cc
```diff
@@ -137,38 +138,44 @@ void predict_or_learn(metrics_data& data, T& base, E& ec)
...
+  new_options.add(make_option("extra_metrics", out_file)
                       .necessary()
                       .help("Specify filename to write metrics to. Note: There is no fixed schema"));
 
   if (!options.add_parse_and_check_necessary(new_options)) { return nullptr; }
 
-  if (data->out_file.empty()) THROW("extra_metrics argument (output filename) is missing.");
+  if (out_file.empty()) THROW("extra_metrics argument (output filename) is missing.");
+  VW::metrics_manager manager(true, out_file);
+  all.global_metrics = std::move(manager);
```
jackgerrits, line 178:
```suggestion
  all.global_metrics = VW::metrics_manager(true, out_file);
```

jackgerrits (changes requested):

vowpalwabbit/core/src/metrics_manager.cc
```diff
@@ -0,0 +1,34 @@
...
+bool metrics_manager::are_metrics_enabled() const { return _are_metrics_enabled; }
+
+std::string metrics_manager::get_filename() const { return _metrics_filename; }
+
+void metrics_manager::register_metrics_callback(const metrics_callback_fn& callback)
+{
+  _metrics_callbacks.push_back(callback);
+}
+
+metric_sink metrics_manager::collect_metrics()
```
jackgerrits, line 23: I'd rather see this function take a reference to the learner stack so that it can handle calling persist_metrics inside the function

vowpalwabbit/core/src/metrics_manager.cc
bassmang: yea that was for testing, reverted

vowpalwabbit/core/include/vw/core/learner.h
bassmang: reverted

vowpalwabbit/core/src/reductions/metrics.cc
bassmang: reverted `persiste_metrics` and moved call to `collect_metrics`

vowpalwabbit/core/src/reductions/baseline_challenger_cb.cc
bassmang: Is this necessary? Seems like that could bloat this interface

jackgerrits: Generally we don't want to force reductions to use fields from all in their setup (or impl), but this can be fixed later if it seems important enough. This works for now

vowpalwabbit/core/src/metrics_manager.cc
```diff
@@ -0,0 +1,34 @@
...
+
+void metrics_manager::register_metrics_callback(const metrics_callback_fn& callback)
+{
+  if (_are_metrics_enabled) { _metrics_callbacks.push_back(callback); }
+}
+
+metric_sink metrics_manager::collect_metrics(LEARNER::base_learner* l) const
+{
+  VW::metric_sink sink;
+  if (_are_metrics_enabled)
```
jackgerrits, line 27: it feels like calling this when metrics are not enabled is an error and perhaps we should throw?

vowpalwabbit/core/src/reductions/metrics.cc
```diff
@@ -137,38 +137,40 @@ void predict_or_learn(metrics_data& data, T& base, E& ec)
...
   {
-    std::string filename = all.options->get_typed_option<std::string>("extra_metrics").value();
-    VW::metric_sink list_metrics;
-
-    all.l->persist_metrics(list_metrics);
-
-    for (auto& metric_hook : all.metric_output_hooks) { metric_hook(list_metrics); }
-
-    list_metrics.set_uint("total_log_calls", all.logger.get_log_count());
+    auto additional_metrics = [&all](metric_sink& sink) -> void
```
jackgerrits, line 143: I don't think we should add this extra stuff as a callback. That means if you were to call output_metrics twice it would add the callback twice.

I think we should leave the callbacks as is and add these extra metrics outside of being a callback.

vowpalwabbit/core/src/metrics_manager.cc
bassmang: updated to throw

vowpalwabbit/core/src/reductions/metrics.cc
bassmang: I agree this should be decoupled from output_metrics. I think it will still need to be added as a callback, since there is no way to interact with the metrics_sink directly in the new design. I added a callback in parse_args so it will not be called multiple times (eg in `output_metrics` or `list_to_json_file`)

jackgerrits (approved):

olgavrou:
what is the objective of metrics manager? to provide global information on whether metrics are enabled and to provide a way for collection of metrics to be called in lib mode? e.g. python can collect metrics now both by calling persist_metrics and collect_metrics?

vowpalwabbit/core/include/vw/core/metrics_manager.h
```diff
@@ -0,0 +1,37 @@
...
+{
+class metrics_manager
+{
+public:
+  metrics_manager(bool enabled = false, std::string filename = "");
+
+  using metrics_callback_fn = std::function<void(VW::metric_sink&)>;
+
+  bool are_metrics_enabled() const;
+  std::string get_filename() const;
```
olgavrou, line 28: get_filename should not be here if the class never uses it, it should probably still live in metrics, or are there plans for it to be used somehow in the future?

vowpalwabbit/core/include/vw/core/metrics_manager.h
```diff
@@ -0,0 +1,37 @@
...
+#include <functional>
+#include <map>
+#include <memory>
+#include <set>
+#include <string>
+#include <vector>
+
+namespace VW
+{
+class metrics_manager
```
olgavrou, line 20: if we remove the filename this could be renamed to `metrics_collector` and restrict the object from potentially becoming another all object in the future

vowpalwabbit/core/include/vw/core/metrics_collector.h
```diff
@@ -0,0 +1,36 @@
...
+  using metrics_callback_fn = std::function<void(VW::metric_sink&)>;
+
+  bool are_metrics_enabled() const;
+  void register_metrics_callback(const metrics_callback_fn& callback);
+  VW::metric_sink collect_metrics(LEARNER::base_learner* l = nullptr) const;
+
+private:
+  bool _are_metrics_enabled;
+  std::vector<metrics_callback_fn> _metrics_callbacks;
+  std::string _metrics_filename;
```
olgavrou, line 34:
```suggestion
```

olgavrou (approved):

vowpalwabbit/core/include/vw/core/metrics_manager.h
bassmang: reverted back to `std::string filename = all.options->get_typed_option<std::string>("extra_metrics").value();`

vowpalwabbit/core/include/vw/core/metrics_manager.h
bassmang: removed filename and updated to `metics_collector`

PR #4408, opened 2022-12-29
refactor: standardize googletest naming

bassmang:
Remove `_tests` from group and `test_` from individual tests to shorten filtering and consistency


vowpalwabbit/slim/test/ut_util.cc
```diff
@@ -351,7 +351,7 @@ invalid_model_param invalid_model_param[] = {
 
 INSTANTIATE_TEST_SUITE_P(VowpalWabbitSlim, invalid_model_test, ::testing::ValuesIn(invalid_model_param));
 
-TEST(VowpalWabbitSlim, multiclass_data_4)
+TEST(vowpal_wabbit_slim, multiclass_data_4)
```
jackgerrits, line 354: If you change this you also need to update this: https://github.com/VowpalWabbit/vowpal_wabbit/blob/5c920b44f3f554820ff1fce01a7976dd1907814f/.github/workflows/build_vw_slim.yml#L31

vowpalwabbit/slim/test/ut_explore.cc
```diff
@@ -9,7 +9,7 @@
 
 using namespace ::testing;
 
-TEST(ExploreTestSuite, sampling_rank)
+TEST(explore_slim, sampling_rank)
```
jackgerrits, line 12: If you change this you also need to update this: https://github.com/VowpalWabbit/vowpal_wabbit/blob/5c920b44f3f554820ff1fce01a7976dd1907814f/.github/workflows/build_vw_slim.yml#L31

vowpalwabbit/slim/test/ut_opts.cc
```diff
@@ -4,15 +4,15 @@
 using namespace ::testing;
 using namespace vw_slim;
 
-TEST(CommandLineOptions, parsing)
+TEST(command_line_options_slim, parsing)
```
jackgerrits, line 7: If you change this you also need to update this: https://github.com/VowpalWabbit/vowpal_wabbit/blob/5c920b44f3f554820ff1fce01a7976dd1907814f/.github/workflows/build_vw_slim.yml#L31

jackgerrits (commented):
I didn't realize this but according to the gtest docs the test suite name can't include `_`

https://google.github.io/googletest/reference/testing.html#TEST

While you're here it might be a good time to fix that up?

bassmang:
> I didn't realize this but according to the gtest docs the test suite name can't include `_`
> 
> https://google.github.io/googletest/reference/testing.html#TEST
> 
> While you're here it might be a good time to fix that up?

Good catch, I'll update to camel case

jackgerrits (approved):

PR #4410, opened 2023-01-03
refactor: move vw.h functions to be defined in vw.cc

jackgerrits:


olgavrou (commented):
instead of moving everything into vw.cc it seems like some things should be shuffled around so that their declarations are moved from the vw.h file into respective .h files, and their definitions left as-is

thoughts?

jackgerrits:
I'd love to update the contents of vw.h, as I think as far as an API surface it needs a lot of work. In the past I think vw.h has formed somewhat of a public API, so I think that keeping the contents of it that same may make sense in this PR. In saying that I think we should evolve the API itself and how it is exposed. It is incomplete and hard to use as is right now.

olgavrou (approved):

PR #4411, opened 2023-01-03
feat: Enable learner type checks at build.

zwd-ms:
Also set some missing types in setup.


olgavrou (commented):
nice! can we get a test that exercises a throw?

jackgerrits (approved):

lalo:
promoted to feature from chore

PR #4413, opened 2023-01-03
refactor: migrate lda finish_example

jackgerrits:
Performance test shows no noticeable change in performance.


vowpalwabbit/core/src/reductions/lda_core.cc
```diff
@@ -1361,21 +1349,36 @@ base_learner* VW::reductions::lda_setup(VW::setup_base_i& stack_builder)
...
+    }
+  }
+
   ld->v.resize(all.lda * ld->minibatch);
 
   ld->decay_levels.push_back(0.f);
 
   all.example_parser->lbl_parser = VW::no_label_parser_global;
 
+  // If minibatch is > 1, then the predict function does not actually produce predictions.
```
peterychang, line 1367: We should probably fix this before we merge? Or at the very least get a John opinion on if this is ok for the time being

jackgerrits: Its existed this way for a long time (probably since the beginning), we just correctly report the type now. That is, that it produces no prediction.

olgavrou (approved):

vowpalwabbit/core/src/reductions/lda_core.cc
jackgerrits: Discussed offline, its fine since it is maintaining previous behavior

PR #4415, opened 2023-01-03
feat: constexpr uniform_hash and type fixes

jackgerrits:
Makes uniform hash constexpr.

It clarifies the fact we use x86 murmurhash 32bit, and fixes the types accordingly. Note, we were casting internally anyway which made for an odd API. Now it is clearer by the parameter types.


lalo (approved):
More explicit of current behavior.

byronxu99 (approved):

PR #4418, opened 2023-01-04
refactor: change initial pool size to 0

jackgerrits:
Forcing all consumers to have a preallocated pool is quite restrictive, especially now that the pool is dynamic (in the past it was a fixed size ring buffer).

Removing this initial allocation improves support for library scenarios, and in practice should have no effect on driver performance. (hyperfine shows this too)

This change from a default of 256 to an initial size of 0 means that base memory consumption is halved (An example is ~31kb) from 13.5MB to 6.2MB.

Startup should also be reduced. While it is hard to measure: `vw --no_stdin` using hyperfine shows:

```
Benchmark 1: ./build/vowpalwabbit/cli/vw --no_stdin
  Time (mean ± σ):       4.5 ms ±   2.1 ms    [User: 1.6 ms, System: 0.9 ms]
  Range (min … max):     3.0 ms …  30.6 ms    682 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 2: ./vw-master --no_stdin
  Time (mean ± σ):       5.8 ms ±   1.6 ms    [User: 2.0 ms, System: 1.5 ms]
  Range (min … max):     4.2 ms …  21.6 ms    658 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  './build/vowpalwabbit/cli/vw --no_stdin' ran
    1.29 ± 0.69 times faster than './vw-master --no_stdin'
```


jackgerrits:
Wait for this change for 10 release

PR #4422, opened 2023-01-04
refactor: migrate epsilon_decay finish_example

bassmang:


vowpalwabbit/core/src/reductions/epsilon_decay.cc
```diff
@@ -408,7 +410,7 @@ VW::LEARNER::base_learner* VW::reductions::epsilon_decay_setup(VW::setup_base_i&
                         .set_params_per_weight(model_count)
                         .set_output_prediction_type(base_learner->get_output_prediction_type())
                         .set_save_load(save_load_epsilon_decay)
-                        .set_finish(::finish)
+                        .set_output_example_prediction(output_example_epsilon_decay)
```
jackgerrits, line 413: Epsilon decay also needs to implement stats and update printing? Unless it can safely delegate to a lower reduction?

bassmang: safe to delegate to lower reductions

vowpalwabbit/core/src/reductions/epsilon_decay.cc
```diff
@@ -277,13 +277,15 @@ void save_load_epsilon_decay(
   else { VW::model_utils::write_model_field(io, epsilon_decay, "_epsilon_decay", text); }
 }
 
-void finish(VW::reductions::epsilon_decay::epsilon_decay_data& data)
+void output_example_epsilon_decay(VW::workspace& /* all */,
```
jackgerrits, line 280: `finish` is the correct place for this function.

PR #4426, opened 2023-01-04
refactor: remove learn and label references from ect predict

bassmang:


jackgerrits:
The separation of ect_predict returning the prediction was helpful for understanding the product of the operation. I think we should leave the function in there but replace the learn usages. This has the added benefit of less churn in this file.

bassmang:
> The separation of ect_predict returning the prediction was helpful for understanding the product of the operation. I think we should leave the function in there but replace the learn usages. This has the added benefit of less churn in this file.

sure just removed the stuff related to label and learning

vowpalwabbit/core/src/reductions/ect.cc
```diff
@@ -173,7 +173,6 @@ uint32_t ect_predict(ect& e, single_learner& base, VW::example& ec)
   uint32_t finals_winner = 0;
 
   // Binary final elimination tournament first
-  ec.l.simple = {FLT_MAX};
```
jackgerrits, line 176: This is necessary. Even though we are calling predict, the simple label should still be set since predict paths require a test_only label

jackgerrits (approved):

PR #4427, opened 2023-01-04
refactor: cleanup cb.h header

jackgerrits:


vowpalwabbit/core/src/cb.cc
```diff
@@ -84,13 +80,13 @@ void parse_label(CB::label& ld, VW::label_parser_reuse_mem& reuse_mem, const std
...
 
-VW::label_parser cb_label = {
+VW::label_parser VW::cb_label_parser_global = {
     // default_label
     [](VW::polylabel& label) { label.cb.reset_to_default(); },
     // parse_label
     [](VW::polylabel& label, VW::reduction_features& /*red_features*/, VW::label_parser_reuse_mem& reuse_mem,
         const VW::named_labels* /*ldict*/, const std::vector<VW::string_view>& words, VW::io::logger& logger)
-    { CB::parse_label(label.cb, reuse_mem, words, logger); },
+    { ::parse_label_cb(label.cb, reuse_mem, words, logger); },
```
bassmang, line 89: why not VW::parse_label_cb?

jackgerrits: This function is only used in this header so it's in an anonymous namespace making it translation unit local.

bassmang (approved):

PR #4432, opened 2023-01-05
refactor: deprecate is_from_pool

jackgerrits:


vowpalwabbit/core/include/vw/core/object_pool.h
```diff
@@ -92,6 +93,7 @@ class no_lock_object_pool
     return size;
   }
 
+  VW_DEPRECATED("Pools will no longer be able to check if an object is from the pool in VW 10.")
```
lalo, line 96: why is this longer not needed? what changed

jackgerrits: The removal of finish_example and VW's required managed pool makes this unnecessary. In fact, removing this makes the pool more flexible since you can add externally created object into the pool

lalo (approved):

PR #4434, opened 2023-01-06
refactor: migrate INTERACTIONS namespace

jackgerrits:


vowpalwabbit/core/include/vw/core/interactions.h
```diff
@@ -326,10 +338,10 @@ std::vector<std::vector<VW::namespace_index>> compile_interactions(
     }
     else { final_interactions.push_back(inter); }
   }
-  std::sort(final_interactions.begin(), final_interactions.end(), INTERACTIONS::sort_interactions_comparator);
+  std::sort(final_interactions.begin(), final_interactions.end(), VW::details::sort_interactions_comparator);
```
olgavrou, line 341: should `sort_and_filter_duplicate_interactions` also be a detail? (thinking if the comparator is a detail why not the sort and filter or vice versa?)

olgavrou (approved):

vowpalwabbit/core/include/vw/core/interactions.h
jackgerrits: Yeah this being a detail makes sense

PR #4435, opened 2023-01-06
refactor: migrate exploration namespace to VW::explore

jackgerrits:


vowpalwabbit/explore/include/vw/explore/explore_internal.h
```diff
@@ -14,8 +16,13 @@
...
 #include <vector>
 
-namespace exploration
+namespace VW
+{
+namespace explore
+{
+namespace details
 {
+
```
olgavrou, line 25: are these randomness generators only used during exploration?

olgavrou: yes looks like it

olgavrou (approved):

vowpalwabbit/explore/include/vw/explore/explore_internal.h
jackgerrits: Yeah, I want to converge on the same impl rather than a copy paste as is here. But I'll do that in another PR.

PR #4444, opened 2023-01-10
refactor: migrate to new initialize function

jackgerrits:


vowpalwabbit/core/include/vw/core/global_data.h
```diff
@@ -319,8 +319,8 @@ class workspace
   workspace(const VW::workspace&&) = delete;
   VW::workspace& operator=(const VW::workspace&&) = delete;
 
-  std::string get_setupfn_name(reduction_setup_fn setup);
-  void build_setupfn_name_dict(std::vector<std::tuple<std::string, reduction_setup_fn>>&);
+  std::string _get_setupfn_name(reduction_setup_fn setup);
```
olgavrou, line 322: why are these functions underscored? we don't follow that practice in the code base

jackgerrits: The workspace is somewhat of a public API and these functions are definitely internal. The underscore implies they are implementation details. Just an idea to make intent clearer to the user

olgavrou: it is a bit pythonic, if they are internal can we put them in the details namespace or make them private?

jackgerrits: You can't namespace a method and making them private isn't doable without making a large number of other places in the codebase friends with the workspace

olgavrou: yeah but they are private member variables, the underscore is just for developer convenience

jackgerrits: Do you think its better to leave the names as is here? The thing I am trying to avoid is an end user relying or even using these APIs

library/library_example.cc
```diff
@@ -12,7 +13,8 @@ inline VW::feature vw_feature_from_string(VW::workspace& v, const std::string& f
 
 int main(int argc, char* argv[])
 {
-  VW::workspace* model = VW::initialize("--hash all -q st --noconstant -f train2.vw --no_stdin");
+  auto model = VW::initialize(VW::make_unique<VW::config::options_cli>(
```
rajan-chari, line 16: Nit: Splitting into two lines seems more reading friendly to me.

jackgerrits: It's formatted with clang-format, I think it's better to leave it to the tool

vowpalwabbit/core/tests/eigen_memory_tree_test.cc
```diff
@@ -42,8 +42,6 @@ TEST(Emt, ParamsTest1)
   EXPECT_EQ(tree->scorer_type, emt_scorer_type::SELF_CONSISTENT_RANK);
   EXPECT_EQ(tree->router_type, emt_router_type::EIGEN);
   EXPECT_EQ(tree->bounder->max_size, 0);
-
```
rajan-chari, line 45: That's nice!

vowpalwabbit/core/include/vw/core/global_data.h
olgavrou: related to discussion: https://github.com/VowpalWabbit/vowpal_wabbit/pull/4456

vowpalwabbit/core/tests/interactions_test.cc
```diff
@@ -476,37 +462,30 @@ TEST(Interactions, ExtentInteractionExpansionTest)
 
 void do_interaction_feature_count_test(bool add_quadratic, bool add_cubic, bool combinations, bool no_constant)
 {
-  std::string char_cmd_line = "--quiet";
-  std::string extent_cmd_line = "--quiet";
+  std::vector<std::string> char_cmd_line{"--quiet"};
```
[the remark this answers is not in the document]
rajan-chari: Is it a bad idea to intoduce a utility funciton that splits a string into a string vector?  The cons are it's not going to work on some command line strings?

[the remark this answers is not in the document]
jackgerrits: For the standard usage in tests the shortcut is `vwtest::make_args("--quiet", ...)`. It gets a bit more verbose when tests build up command lines like in this one. I don't think we should alias something as common as a vector of strings though.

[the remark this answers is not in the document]
jackgerrits: There is a utility function `VW:: split_command_line`. There is just no reason to use it in tests when the input is known ahead of time.

vowpalwabbit/core/tests/simulator.cc
```diff
@@ -94,6 +95,7 @@ std::pair<std::string, float> cb_sim::get_action(VW::workspace* vw, const std::m
   auto const& scores = examples[0]->pred.a_s;
   std::vector<float> ordered_scores(scores.size());
   for (auto const& action_score : scores) { ordered_scores[action_score.action] = action_score.score; }
+  pmf.reserve(ordered_scores.size());
```
rajan-chari, line 98: Interesting optimization.  Why not declare the vector here and use the copy constructor?

jackgerrits: Yeah could do that too, would be better. This was just me applying a clang-tidy suggestion as I go.

vowpalwabbit/core/tests/tutorial_test.cc
```diff
@@ -8,18 +8,21 @@
...
+      std::vector<std::string>{"--cb_explore_adf", "--quiet", "--epsilon=0.2", "--random_seed=5"});
   EXPECT_GT(ctr.back(), 0.37f);
   EXPECT_LT(ctr.back(), 0.49f);
 }
 
 TEST(Tutorial, CppSimulatorWithInteraction)
 {
-  auto ctr = simulator::_test_helper("--cb_explore_adf -q UA --quiet --epsilon 0.2 --random_seed 5");
+  auto ctr = simulator::_test_helper(
+      std::vector<std::string>{"--cb_explore_adf", "--quadratic=UA", "--quiet", "--epsilon=0.2", "--random_seed=5"});
```
rajan-chari, line 20: Shouldn't this be two strings?

jackgerrits: which one?

vowpalwabbit/core/src/vw.cc
```diff
@@ -276,8 +284,8 @@ std::unique_ptr<VW::workspace> VW::initialize_experimental(std::unique_ptr<confi
 }
 
 std::unique_ptr<VW::workspace> VW::initialize(std::unique_ptr<config::options_i> options,
-    std::unique_ptr<VW::io::reader> model_override_reader, bool skip_model_load,
```
rajan-chari, line 279: Is there a reason to remove this param and have it always skip model load?

jackgerrits: The only use case for this parameter is an implementation detail of seed_vw_model and I cannot think of a single end user scenario. 

Removing this parameter means we don't ever skip_model_load since there is no reason to

vowpalwabbit/core/src/vw.cc
```diff
@@ -315,8 +323,11 @@ std::unique_ptr<VW::workspace> VW::seed_vw_model(VW::workspace& vw_model, const
 
   auto options = VW::make_unique<config::options_cli>(serialized_options);
 
-  auto new_model = VW::initialize(std::move(options), nullptr, true /* skip_model_load */, driver_output_func,
-      driver_output_func_context, custom_logger);
+  std::unique_ptr<config::options_i, options_deleter_type> options_custom_deleter(
+      new VW::config::options_cli(serialized_options), [](VW::config::options_i* ptr) { delete ptr; });
+
+  auto new_model = initialize_internal(std::move(options_custom_deleter), nullptr, false /* skip model load */,
```
rajan-chari, line 329: Is skip_model_load=true intended?

vowpalwabbit/core/src/reduction_stack.cc
```diff
@@ -250,7 +250,7 @@ void default_reduction_stack_setup::delayed_state_attach(VW::workspace& all, VW:
   _all_ptr = &all;
   _options_impl = &options;
   // populate setup_fn -> name map to be used to lookup names in setup_base
-  all.build_setupfn_name_dict(_reduction_stack);
+  all._build_setupfn_name_dict(_reduction_stack);
```
rajan-chari, line 253: Curious about the _ in fn_name.  What's the convention?

jackgerrits: These functions are implementation details, so the idea was to express that. This change is no longer needed though as it's superseded by: https://github.com/VowpalWabbit/vowpal_wabbit/pull/4456

vowpalwabbit/core/src/vw.cc
jackgerrits: Yes, we don't want to skip loading the model the user supplied

vowpalwabbit/core/src/reduction_stack.cc
rajan-chari: Got it.

vowpalwabbit/core/src/parser.cc
```diff
@@ -527,8 +527,9 @@ void VW::details::enable_sources(
         if (got_sigterm)
         {
           for (size_t i = 0; i < num_children; i++) { kill(children[i], SIGTERM); }
-          VW::finish(all);
-          exit(0);
+          all.finish();
+          delete &all;
```
rajan-chari, line 531: Is delete needed?

jackgerrits: Since we call `exit` immediately after we're making a best effort to free our resources. I don't think `exit` will cause destructors up the stack to call if I understand correctly.

library/test_search.cc
```diff
@@ -96,18 +98,19 @@ void train()
 {
   // initialize VW as usual, but use 'hook' as the search_task
   cerr << endl << endl << "##### train() #####" << endl << endl;
-  VW::workspace& vw_obj =
-      *VW::initialize("--search 4 --quiet --search_task hook --example_queue_limit 1024 -f my_model");
-  run(vw_obj);
-  VW::finish(vw_obj, false);
+  auto vw_obj = VW::initialize(VW::make_unique<VW::config::options_cli>(
+      std::vector<std::string>{"--search 4 --quiet --search_task hook --example_queue_limit 1024 -f my_model"}));
```
rajan-chari, line 102: Needs to be split into individual strings or vw_split_command?

library/test_search.cc
```diff
@@ -96,18 +98,19 @@ void train()
...
 }
 
 void predict()
 {
   cerr << endl << endl << "##### predict() #####" << endl << endl;
-  VW::workspace& vw_obj = *VW::initialize("--quiet -t --example_queue_limit 1024 -i my_model");
-  run(vw_obj);
-  VW::finish(vw_obj, false);
+  auto vw_obj = VW::initialize(VW::make_unique<VW::config::options_cli>(
+      std::vector<std::string>{"--quiet -t --example_queue_limit 1024 -i my_model"}));
```
rajan-chari, line 111: Needs to be split into individual strings or vw_split_command?

vowpalwabbit/json_parser/tests/dsjson_parser_test.cc
```diff
@@ -236,9 +230,8 @@ TEST(ParseDsjson, Cats)
   }
 }
 )";
-  auto vw = VW::initialize(
-      "--dsjson --chain_hash --cats 4 --min_value=185 --max_value=23959 --bandwidth 1 --no_stdin --quiet", nullptr,
-      false, nullptr, nullptr);
+  auto vw = VW::initialize(vwtest::make_args("--dsjson", "--chain_hash", "--cats", "4", "--min_value=185",
+      "--max_value=23959", "--bandwidth", "1", "--no_stdin", "--quiet"));
```
rajan-chari, line 234: need two strings for cmds with '=' ?

vowpalwabbit/json_parser/tests/dsjson_parser_test.cc
```diff
@@ -278,9 +270,8 @@ TEST(ParseDsjson, CatsNoLabel)
   }
 }
 )";
-  auto vw = VW::initialize(
-      "--dsjson --chain_hash -t --cats 4 --min_value=185 --max_value=23959 --bandwidth 1 --no_stdin --quiet", nullptr,
-      false, nullptr, nullptr);
+  auto vw = VW::initialize(vwtest::make_args("--dsjson", "--chain_hash", "-t", "--cats", "4", "--min_value=185",
```
rajan-chari, line 273: need two strings for cmds with '=' ?

vowpalwabbit/core/tests/tutorial_test.cc
rajan-chari: --quadratic=UA
--epsilon=0.2
--random_seed=5

library/test_search.cc
jackgerrits: Good catch thanks!

library/test_search.cc
jackgerrits: Thanks for finding!

vowpalwabbit/json_parser/tests/dsjson_parser_test.cc
jackgerrits: No, `--max_value=23959` is a valid single token. In fact it is preferred as it is unambiguous to the CLI parser.

rajan-chari (approved):
Looks good.  Thanks.

olgavrou (approved):

vowpalwabbit/core/tests/tutorial_test.cc
jackgerrits: Those are valid ways to pass the value of an option

PR #4451, opened 2023-01-11
feat: expose explore eval stats to python interface

olgavrou:


lalo (approved):

python/tests/test_pyvw.py
```diff
@@ -766,3 +766,14 @@ def test_merge_models_with_base():
...
     assert merged_model.get_weight_from_name("bar") != 0
+
+
+def test_get_explore_eval_stats():
+    vw = vowpalwabbit.Workspace("--cb_explore_adf --explore_eval --dsjson")
+
+    ex_l_str = '{"_label_cost":-0.9,"_label_probability":0.5,"_label_Action":1,"_labelIndex":0,"o":[{"v":1.0,"EventId":"38cbf24f-70b2-4c76-aa0c-970d0c8d388e","ActionTaken":false}],"Timestamp":"2020-11-15T17:09:31.8350000Z","Version":"1","EventId":"38cbf24f-70b2-4c76-aa0c-970d0c8d388e","a":[1,2],"c":{ "GUser":{"id":"person5","major":"engineering","hobby":"hiking","favorite_character":"spock"}, "_multi": [ { "TAction":{"topic":"SkiConditions-VT"} }, { "TAction":{"topic":"HerbGarden"} } ] },"p":[0.5,0.5],"VWState":{"m":"N/A"}}\n'
+    ex_l = vw.parse(ex_l_str)
+    vw.learn(ex_l)
+    vw.finish_example(ex_l)
```
lalo, line 777: you have to call finish example before fetching this metrics right?

lalo: maybe add comment for someone checking this test to make sure that they do that

python/pylibvw.cc
```diff
@@ -1102,6 +1102,19 @@ float get_total_sum_feat_sq(example_ptr ec) { return ec->get_total_sum_feat_sq()
 double get_sum_loss(vw_ptr vw) { return vw->sd->sum_loss; }
 double get_holdout_sum_loss(vw_ptr vw) { return vw->sd->holdout_sum_loss; }
 double get_weighted_examples(vw_ptr vw) { return vw->sd->weighted_examples(); }
+float get_explore_eval_weighted_update_count(vw_ptr vw)
+{
+  return vw->sd->reduction_stats.count("explore_eval_weighted_update_count")
```
jackgerrits, line 1107: If it doesn't exist then perhaps we should return `None`?

olgavrou: mmm yeah probably

rajan-chari:
Wondering if there should be an explict arg asking for stats.  Library can then avoid calculating stats ...

olgavrou:
> Wondering if there should be an explict arg asking for stats. Library can then avoid calculating stats ...

Explore Eval doesn't make sense without these stats really, we discussed yesterday and are going to move these to metrics so that they can be exposed to python via that

jackgerrits:
Stats cannot be skipped as GD's implementation depends on them currently.

vowpalwabbit/core/src/reductions/metrics.cc
```diff
@@ -137,6 +137,12 @@ void predict_or_learn(metrics_data& data, T& base, E& ec)
 
 void VW::reductions::additional_metrics(VW::workspace& all, VW::metric_sink& sink)
 {
+  sink.set_float("sum_loss", all.sd->sum_loss);
```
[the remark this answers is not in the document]
lalo: @jackgerrits  you ran into this with py bindings next

test/core.vwtest.json
```diff
@@ -5895,5 +5895,16 @@
     "input_files": [
       "train-sets/las_100_actions.txt"
     ]
+  },
+  {
+    "id": 456,
+    "desc": "Spin off epsilon decay model to lower bitsize to cb_explore with explore eval",
+    "vw_command": "--cb_explore_adf -d train-sets/automl_spin_off.txt --noconstant -b 18 --predict_only_model --epsilon_decay --model_count 2 --explore_eval",
```
jackgerrits, line 5902: Is this meant to be in this PR?

vowpalwabbit/core/src/reductions/metrics.cc
```diff
@@ -136,6 +136,12 @@ void predict_or_learn(metrics_data& data, T& base, E& ec)
 
 void additional_metrics(VW::workspace& all, VW::metric_sink& sink)
 {
+  sink.set_float("sum_loss", all.sd->sum_loss);
```
jackgerrits, line 139: This is already accessible in Python

https://vowpalwabbit.org/docs/vowpal_wabbit/python/latest/reference/vowpalwabbit.pyvw.pylibvw.html#vowpalwabbit.pyvw.pylibvw.vw.get_sum_loss

vowpalwabbit/core/src/reductions/metrics.cc
```diff
@@ -136,6 +136,12 @@ void predict_or_learn(metrics_data& data, T& base, E& ec)
...
+  sink.set_float("weighted_labeled_examples", all.sd->weighted_labeled_examples);
```
jackgerrits, line 140: Weighted examples is already exposed so it feels like this should just be added next to it

[double get_weighted_examples(vw_ptr vw) { return vw->sd->weighted_examples(); }](https://vowpalwabbit.org/docs/vowpal_wabbit/python/latest/reference/vowpalwabbit.pyvw.pylibvw.html#vowpalwabbit.pyvw.pylibvw.vw.get_weighted_examples)

vowpalwabbit/core/src/reductions/metrics.cc
```diff
@@ -136,6 +136,12 @@ void predict_or_learn(metrics_data& data, T& base, E& ec)
...
+  if (all.sd->weighted_labeled_examples != 0.0f)
+  {
+    sink.set_float("avg_loss", all.sd->sum_loss / all.sd->weighted_labeled_examples);
```
jackgerrits, line 143: Why add this as a metric when it can be trivially derived from the other two?

test/core.vwtest.json
lalo: I think so

jackgerrits: It's not related to metrics or explore eval though

vowpalwabbit/core/src/reductions/metrics.cc
lalo: sanity check - will remove

vowpalwabbit/core/src/reductions/metrics.cc
lalo: will remove

PR #4455, opened 2023-01-13
build: do not override externally set VW_CXX_STANDARD

jackgerrits:


cmake/VowpalWabbitUtils.cmake
```diff
@@ -2,7 +2,9 @@ include(CMakeParseArguments)
 include(GNUInstallDirs)
 include(VWFlags)
 include(DetectCXXStandard)
-set(VW_CXX_STANDARD 11)
+if(NOT VW_CXX_STANDARD)
```
rajan-chari, line 5: Curious.  In all the CI's will it be C++ 11 by default?

jackgerrits: Yes, unless otherwise specified 11 is the default.

PR #4461, opened 2023-01-18
refactor: Migrate raw function pointers to std::function

byronxu99:
This PR replaces several raw function pointers with `std::function` objects.

The biggest changes are to the `learner` class, which is now no longer templated. To erase the data type, learner builders will bind the data object to their input functions via a lambda capture. To erase the example type, the lambda will accept a `polymorphic_ex` wrapper around `VW::example*` and `VW::multi_ex*`, which implicitly converts back to the original type. The resulting lambda can be stored in a `std::function` object with a type which does not depend on DataT nor ExampleT.

This removes all undefined behavior around function pointer typecasting.


vowpalwabbit/core/src/learner.cc
```diff
@@ -312,12 +326,366 @@ bool ec_is_example_header(const example& ec, label_type_t label_type)
...
+  l->_subtract_with_all_f = nullptr;
+
+  _derived_learner_created_already = true;
+  return l;
+}
+
+learner& learner::safe_get_base_learner()
+{
+  if (_base_learner) { return *_base_learner; }
+  return *this;
```
rajan-chari, line 680: reminder:  If null, return a type that will produce errors instead of *this

vowpalwabbit/core/src/learner.cc
```diff
@@ -312,12 +326,366 @@ bool ec_is_example_header(const example& ec, label_type_t label_type)
...
+  l->_add_f = nullptr;
+  l->_add_with_all_f = nullptr;
+  l->_subtract_f = nullptr;
+  l->_subtract_with_all_f = nullptr;
+
+  _derived_learner_created_already = true;
+  return l;
+}
+
+learner& learner::safe_get_base_learner()
```
rajan-chari, line 677: Some docs here explaining what's meant by base_learner would make the expectation clear for the user

vowpalwabbit/core/src/loss_functions.cc
```diff
@@ -482,12 +483,18 @@ namespace VW
 std::unique_ptr<loss_function> get_loss_function(
     VW::workspace& all, const std::string& funcName, float function_parameter_0, float function_parameter_1)
 {
+  // For some loss functions, we want to know if all.set_minmax has been overwritten
+  // Check if the std::function object does not contain VW::details::noop_mm
+  using noop_mm_fn_ptr_type = typename std::add_pointer<decltype(VW::details::noop_mm)>::type;
+  auto* minmax_fptr = all.set_minmax.target<noop_mm_fn_ptr_type>();
+  bool set_minmax_is_not_noop = (minmax_fptr == nullptr || *minmax_fptr != VW::details::noop_mm);
```
peterychang, line 490: instead of this, couldn't you just do `set_minmax = nullptr` wherever we set it to noop_mm? Does that break some other logic elsewhere?

vowpalwabbit/core/src/learner.cc
byronxu99: I renamed the leaner immediately before the current learner as "previous". Now "base" refers specifically to the last learner of the entire reduction stack.

vowpalwabbit/core/src/loss_functions.cc
byronxu99: `noop_mm` isn't a new function here - it was there before and I just changed the signature. I think the purpose of having it was to avoid needing to do a nullptr check every time we want to call the set_minmax function.

The alternative way of just doing the nullptr check should also be fine. I'll try it out and see if it'll let me get rid of the ugly std::function target() call, which requires figuring out a specific "pointer to your function pointer" type.

byronxu99: `noop_mm` isn't a new function here - it was there before and I just changed the signature. I think the purpose of having it was to avoid needing to do a nullptr check every time we want to call the set_minmax function.

The alternative way of just doing the nullptr check should also be fine. I'll try it out and see if it'll let me get rid of the ugly std::function target() call, which requires figuring out a specific "pointer to your function pointer" type.

vowpalwabbit/core/src/learner.cc
```diff
@@ -312,77 +326,472 @@ bool ec_is_example_header(const example& ec, label_type_t label_type)
...
-            to_string(out_label_type), this_name, to_string(base_in_label_type), base_name);
-    THROW(message);
+    details::increment_offset(ec, increment, lo);
+    debug_log_message(ec, "multipredict");
+    for (size_t c = 0; c < count; c++)
+    {
+      _predict_f(safe_get_previous_learner(), ec);
+      if (finalize_predictions)
+      {
+        pred[c] = std::move(static_cast<VW::example&>(ec).pred);  // TODO: this breaks for complex labels because =
```
peterychang, line 390: The only label that doesn't do a correct copy is CCB (which contains a pointer). I think all the predictions should do the right thing. The big problem with that though, is that deep copies are really expensive for these large objects

jackgerrits: The CCB label implements the copy/move operators so it should work okay

vowpalwabbit/core/include/vw/core/reduction_stack.h
```diff
@@ -21,9 +19,9 @@ class default_reduction_stack_setup : public setup_base_i
 
   void delayed_state_attach(VW::workspace& all, VW::config::options_i& options) override;
 
-  // this function consumes all the reduction_stack until it's able to construct a base_learner
+  // this function consumes all the reduction_stack until it's able to construct a learner
   // same signature as the old setup_base(...) from parse_args.cc
-  VW::LEARNER::base_learner* setup_base_learner() override;
+  VW::LEARNER::learner* setup_base_learner() override;
```
jackgerrits, line 24: It seems like this should return a shared_pointer as well?

vowpalwabbit/core/include/vw/core/learner.h
```diff
@@ -297,386 +149,175 @@ class learner
...
-    assert((is_multiline() && std::is_same<multi_ex, E>::value) ||
-        (!is_multiline() && std::is_same<example, E>::value));  // sanity check under debug compile
-    details::increment_offset(ec, increment, i);
-    debug_log_message(ec, "predict");
-    _learn_fd.predict_f(_learn_fd.data, *_learn_fd.base, (void*)&ec);
-    details::decrement_offset(ec, increment, i);
-  }
+  /// example::pred. If the polymorphic_ex is ::multi_ex then the prediction is
+  /// set on the 0th item in the list.
+  void predict(polymorphic_ex ec, size_t i = 0);
```
jackgerrits, line 154: Perhaps we should rename `polymorphic_ex` to `polymorphic_ex_ref` to make it clear it is essentially a reference wrapper

vowpalwabbit/core/include/vw/core/learner.h
```diff
@@ -1054,206 +665,361 @@ class common_learner_builder
...
-          BaseLearnerT>
+    : public common_learner_builder<reduction_learner_builder<DataT, ExampleT>, DataT, ExampleT>
 {
 public:
-  using super =
-      common_learner_builder<reduction_learner_builder<DataT, ExampleT, BaseLearnerT>, DataT, ExampleT, BaseLearnerT>;
-  reduction_learner_builder(std::unique_ptr<DataT>&& data, BaseLearnerT* base, const std::string& name)
-      // NOTE: This is a copy of the base! The purpose is to copy all of the
+  using super = common_learner_builder<reduction_learner_builder<DataT, ExampleT>, DataT, ExampleT>;
+  reduction_learner_builder(std::unique_ptr<DataT>&& data, learner* prev, const std::string& name)
```
jackgerrits, line 674: `prev` being a shared_ptr here will allow us to propagate the lifetime all the way from he initial creation through the stack

vowpalwabbit/core/include/vw/core/learner.h
```diff
@@ -1054,206 +665,361 @@ class common_learner_builder
...
-    this->learner_ptr->_learn_fd.multipredict_f = nullptr;
-    // Don't propagate merge functions
-    this->learner_ptr->_merge_fn = nullptr;
-    this->learner_ptr->_merge_with_all_fn = nullptr;
-    this->learner_ptr->_add_fn = nullptr;
-    this->learner_ptr->_add_with_all_fn = nullptr;
-    this->learner_ptr->_subtract_fn = nullptr;
-    this->learner_ptr->_subtract_with_all_fn = nullptr;
+      : common_learner_builder<reduction_learner_builder<DataT, ExampleT>, DataT, ExampleT>(
+            prev->make_next_learner(), std::move(data), name)
```
jackgerrits, line 680: I don't love the rename to `prev`, I prefer the original naming of base. Base has also been in use for quite some time in this codebase so I think most people understand what it means

vowpalwabbit/core/src/learner.cc
```diff
@@ -312,77 +326,427 @@ bool ec_is_example_header(const example& ec, label_type_t label_type)
...
+{
+  if (l->is_multiline()) { return l; }
+  auto message = fmt::format("Tried to use a singleline reduction as a multiline reduction Name: {}", l->get_name());
+  THROW(message);
+}
 
-void VW::LEARNER::details::learner_build_diagnostic(VW::string_view this_name, VW::string_view base_name,
-    prediction_type_t in_pred_type, prediction_type_t base_out_pred_type, label_type_t out_label_type,
-    label_type_t base_in_label_type, details::merge_fn merge_fn_ptr, details::merge_with_all_fn merge_with_all_fn_ptr)
+learner* as_singleline(learner* l)
```
jackgerrits, line 336: Because these no longer actually change the type I don't think it should be called `as_X` anymore. If we are checking an invariant then we should rename to something like `require_singleline`

vowpalwabbit/io/include/vw/io/io_adapter.h
```diff
@@ -104,7 +105,7 @@ std::unique_ptr<writer> open_compressed_stdout();
 std::unique_ptr<reader> open_stdin();
 std::unique_ptr<writer> open_stdout();
 
-using write_func_t = ssize_t (*)(void*, const char*, size_t);
+using write_func_t = std::function<ssize_t(void*, const char*, size_t)>;
```
jackgerrits, line 108: Not everything needs to become a `std::function` in my opinion. This function seems fine as it was.

vowpalwabbit/io/include/vw/io/logger.h
```diff
@@ -37,8 +38,8 @@ enum class log_level
 
 log_level get_log_level(const std::string& level);
 
-using logger_output_func_t = void (*)(void*, VW::io::log_level, const std::string&);
-using logger_legacy_output_func_t = void (*)(void*, const std::string&);
+using logger_output_func_t = std::function<void(void*, VW::io::log_level, const std::string&)>;
+using logger_legacy_output_func_t = std::function<void(void*, const std::string&)>;
```
jackgerrits, line 42: Same comment for leaving these as is. These also are  a part of the interface and shouldn't be changed without proper deprecation.

vowpalwabbit/core/include/vw/core/learner.h
byronxu99: Fair point but `polymorphic_ex` already a pretty long name, so I'm hesitant to make it even longer.

Also, multi_ex holds example pointers inside and not example objects.

jackgerrits: That's true. Perhaps add a doc comment so at least we're clear that this is a reference type

vowpalwabbit/io/include/vw/io/io_adapter.h
byronxu99: Changed back to function pointer

vowpalwabbit/core/include/vw/core/learner.h
byronxu99: Rajan wanted this change because of the double meaning of base - as both a "base learner" that's the last learner in the stack, and a "base of a learner" meaning the learner created immediately before this learner.

I renamed the second use case as "prev" to avoid the overloaded term. I'll see if I can add a comment that clarifies this better.

vowpalwabbit/core/include/vw/core/learner.h
byronxu99: We use `std::enable_shared_from_this` to let us create a valid shared_ptr from a raw pointer. The code right now should already have correct lifetime of the shared_ptr from creation to finally getting stored in VW::workspace.

vowpalwabbit/core/include/vw/core/reduction_stack.h
byronxu99: Added comments clarifying this along with a wrapper function that returns shared_ptr

vowpalwabbit/core/include/vw/core/learner.h
jackgerrits: It's clearer to the reader though if this were to express the shared ownership being passed. We should never use raw pointers for ownership which seems to be happening here even though it is being promoted to a shared pointer later

vowpalwabbit/core/include/vw/core/learner.h
jackgerrits: That is a reasonable explanation for splitting them, im on board. I still find the work prev confusing as normally you consider examples traveling down the stack as the default direction. Saying prev in that context implies going up and not down. So it feels like prev is a misnomer

vowpalwabbit/core/include/vw/core/reductions/plt.h
```diff
@@ -5,10 +5,12 @@
...
 #include "vw/core/vw_fwd.h"
 
+#include <memory>
+
 namespace VW
 {
 namespace reductions
 {
-VW::LEARNER::base_learner* plt_setup(VW::setup_base_i& stack_builder);
+std::shared_ptr<VW::LEARNER::learner> plt_setup(VW::setup_base_i& stack_builder);
```
lalo, line 14: what's the intuition behind shared_ptr?

vowpalwabbit/core/include/vw/core/learner.h
byronxu99: Changed it to use shared_ptr now

vowpalwabbit/core/include/vw/core/reduction_stack.h
byronxu99: Actually, with the other changes made this now returns shared_ptr

vowpalwabbit/core/include/vw/core/learner.h
byronxu99: I think you're looking at it from the perspective of examples, and I'm describing it from the perspective of learner objects. We create reduction learners by copy-constructing a previously created learner. So each learner has a pointer to its previous, which was the one used to create it.

vowpalwabbit/core/include/vw/core/array_parameters_dense.h
```diff
@@ -154,10 +153,9 @@ class dense_parameters
 #endif
 
 private:
-  VW::weight* _begin;
+  std::shared_ptr<VW::weight> _begin;
```
peterychang, line 156: Why a shared_ptr? Seems more standard to just delete the pointer in the destructor, or at the very least use a unique_ptr (since the copy and assignment operators are deleted)

byronxu99: shared_ptr is necessary for shallow_copy() to work, which is used for seed_vw_model

peterychang: ick, i guess seeded models share the weights? We should add a comment about that, because it would cause all kinds of issues if we had multiple models sharing weights in a multi-threaded way

vowpalwabbit/core/include/vw/core/learner.h
jackgerrits: Yeah there are a few ways of looking at it. You should ask a few people to see what the default intuition would be. For me it is the perspective of examples since thats what the logic of reductions is designed around

vowpalwabbit/core/include/vw/core/array_parameters_dense.h
byronxu99: Seeded models share both weights and `shared_data`, and both were raw pointers. The base model used to seed other models was responsible for freeing the pointers, and you had to be sure to free that model last. I changed both to use `shared_ptr`, and now they can be freed in any order.

Even worse is daemon mode, which tries to `mmap` the same shared_data across all its child processes, and never `munmap`s the memory. I made that one a `shared_ptr` with custom destructor.

byronxu99: Looks like there's already a comment in `vw.cc` about `seed_vw_model` being unsafe

/// Creates a workspace based off of another workspace. What this means is that
/// the model weights and the shared_data object are shared. This function needs
/// to be used with caution. Reduction data is not shared, therefore this
/// function is unsafe to use for situations where reduction state is required
/// for proper operation such as marginal and cb_adf. Learn on a seeded instance
/// is unsafe, and prediction is also potentially unsafe.

rajan-chari:
> Yeah there are a few ways of looking at it. You should ask a few people to see what the default intuition would be. For me it is the perspective of examples since thats what the logic of reductions is designed around 

Maybe it's helpful to seperate the issues when thinking about naming things.  There are 2 tensions:
1) From the stack construction perspective we move in one direction (bottom of the stack to top).  When predict() is called, we move in the other direction (top of stack calls the next reduction down the stack).
2) The word 'base' has two meanings when going down the predict/learn stack.  (a) It's the next down stream reduction.  (b) It's the last reduction in the stack.

vowpalwabbit/core/include/vw/core/learner.h
```diff
@@ -297,386 +160,175 @@ class learner
...
   VW_ATTR(nodiscard) const std::string& get_name() const { return _name; }
-  VW_ATTR(nodiscard) const base_learner* get_learn_base() const { return _learn_fd.base; }
-  VW_ATTR(nodiscard) base_learner* get_learn_base() { return _learn_fd.base; }
-  /// If true, this specific learner defines a save load function. If false, it simply forwards to a base
-  /// implementation.
-  VW_ATTR(nodiscard) bool learner_defines_own_save_load() { return _learn_fd.data == _save_load_fd.data; }
+
+  // Returns a pointer to the previous learner in the reduction stack
+  // Returns nullptr if this is the base learner of the stack
+  VW_ATTR(nodiscard) const learner* get_previous_learner() const { return _previous_learner.get(); }
```
lalo, line 244: Let's revisit this rename, previous is not the right word here. Status quo is that a reduction only knows whats under it i.e. a generic promise of a base learner - which might have some extra layers on top but that is hidden away.

lalo (changes requested):

vowpalwabbit/core/include/vw/core/learner.h
rajan-chari: Maybe it's helpful to separate the issues when thinking about naming things. There are 2 tensions:

From the stack construction perspective, we move in one direction (bottom of the stack to top). When predict() is called, we move in the other direction (top of stack calls the next reduction down the stack).
The word 'base' has two meanings when going down the predict/learn stack. (a) It's the next down stream reduction. (b) It's the last reduction in the stack.

lalo: The stack gets built from the top down also not from the bottom to the top. https://github.com/VowpalWabbit/vowpal_wabbit/blob/2fe6f2fb97962040c674e1bee3665d086347f2a0/vowpalwabbit/core/src/reduction_stack.cc#L273

We pop and call setup func - but yes behind the scenes some internal impl detail is that the learner instance for the bottom setup function gets called first because to build a stack you need the previous element first before wrapping it up with the next but that's just normal recursion/stack side effect of how things get executed.

jackgerrits: Yeah I agree with that. It is still built top down, it is just recursive in how it calls.

byronxu99: I have now renamed base learner (bottom-most learner of the stack) to "foundation learner". It turns out that this is the option that minimizes the number of other changes that needed to be made, especially to public documentation.

Changes to documentation are mainly in this wiki page, for which I have created a new copy: https://github.com/VowpalWabbit/vowpal_wabbit/wiki/What-is-a-learner%3F-(new-version)

In addition, this page uses "base" in both ways, but mostly in the "base of a learner" sense. It will require one small change from the other meaning of base to "foundation". https://github.com/VowpalWabbit/vowpal_wabbit/wiki/Matching-Label-and-Prediction-Types-Between-Reductions

There are a few other wiki pages where base would get changed to foundation, but those are fairly minor changes.

I could not find any change needed on the VW site, but maybe I missed something. https://vowpalwabbit.org/docs/vowpal_wabbit/python/latest/index.html

In summary, we have 3 options
1. Rename the bottom-most learner to "foundation". This is the current state of this PR, and will require the changes described above to make our documentation consistent.
2. Rename a reduction's base to "previous". This was what I has first done, but it was ambiguous because the next learner that an example would travel to was named "previous". We could also name it "next", but that gives the counterintuitive situation where a reduction is created by copying its "next" learner, which already exists before itself is created.
3. Don't rename anything, and leave the double meaning of "base" as-is.

byronxu99: Also tagging @JohnLangford for any comments on this

vowpalwabbit/core/src/parse_args.cc
```diff
@@ -1651,17 +1651,17 @@ void VW::details::parse_sources(options_i& options, VW::workspace& all, VW::io_b
...
+  if (!all.quiet && !all.options->was_supplied("audit_regressor") && !enabled_learners.empty())
   {
     const char* const delim = ", ";
     std::ostringstream imploded;
     std::copy(
-        enabled_reductions.begin(), enabled_reductions.end() - 1, std::ostream_iterator<std::string>(imploded, delim));
+        enabled_learners.begin(), enabled_learners.end() - 1, std::ostream_iterator<std::string>(imploded, delim));
 
-    *(all.trace_message) << "Enabled reductions: " << imploded.str() << enabled_reductions.back() << std::endl;
+    *(all.trace_message) << "Enabled reductions: " << imploded.str() << enabled_learners.back() << std::endl;
```
lalo, line 1664: I'm guessing you skipped the string change purposely.

byronxu99: Actually this was accidental. It's changed now.

PR #4463, opened 2023-01-19
fix: [automl] allow multiple models underneath automl

bassmang:


test/core.vwtest.json
```diff
@@ -5700,5 +5700,71 @@
...
+      "models/aml_spin_off_bag_cb.m"
+    ],
+    "depends_on": [
+      443
+    ]
+  },
+  {
+    "id": 445,
+    "desc": "Check that binary model files are identical between standard cb_explore_adf and automl spinoff with bag",
+    "bash_command": "diff models/aml_spin_off_bag.m models/aml_spin_off_bag_cb.m",
```
lalo, line 5761: the naming of the files could be more descriptive, since one model doesnt have automl on yet it has aml on the name

bassmang: updating naming

lalo:
this PR should go in first, before https://github.com/VowpalWabbit/vowpal_wabbit/pull/4023

vowpalwabbit/core/include/vw/core/multi_model_utils.h
```diff
@@ -7,49 +7,92 @@
...
 namespace VW
 {
 namespace reductions
 {
 namespace multi_model
 {
-// ***** NOTE: params_per_problem must be of form 2^n *****
-inline void clear_offset(dense_parameters& weights, const size_t offset, const size_t params_per_problem)
+// ***** NOTE: ppw must be of form 2^n *****
+inline void clear_offset(dense_parameters& weights, const size_t offset, const size_t ppw, const size_t inner_ppw_size)
```
lalo, line 17: this technically should be recursive, and maybe even part of learner.h. in this impl we can only have two reductions that modify the wpp, if we add a third it would stop working, right?

lalo, line 17: nit: rename args

lalo (approved):

vowpalwabbit/core/include/vw/core/multi_model_utils.h
bassmang: updated

PR #4472, opened 2023-01-30
refactor: estimators ns in cressieread

lalo:


vowpalwabbit/core/include/vw/core/estimators/cressieread_estimator.h
```diff
@@ -8,6 +8,8 @@
 
 namespace VW
 {
+namespace estimators
+{
 class cressieread_estimator
```
olgavrou, line 13: super nit: this class could be named simply `cressieread` now

olgavrou (approved):

vowpalwabbit/core/include/vw/core/estimators/cressieread_estimator.h
lalo: true might as well

PR #4476, opened 2023-01-30
fix: [LAS] LAS not a cb adf common reduction, fixes metrics with LAS bug

olgavrou:
When LAS is enabled then two exploration algorithms are enabled and they both call `cb_explore_adf_common`'s `persist_metrics` function which overrides the same keys

Fixed by making LAS a standalone reduction and not a cb adf common one

### DIFF
--- a/vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
+++ b/vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
@@ -250,6 +250,11 @@ void generate_Z(const multi_ex& examples, Eigen::MatrixXf& Z, Eigen::MatrixXf& B
 }
 
 template <typename T, typename S>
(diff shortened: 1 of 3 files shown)


vowpalwabbit/core/src/reductions/cb/cb_explore_adf_large_action_space.cc
```diff
@@ -315,7 +320,7 @@ VW::LEARNER::base_learner* make_las_with_impl(VW::setup_base_i& stack_builder, V
                 .set_output_example_prediction(explore_type::output_example_prediction)
                 .set_update_stats(explore_type::update_stats)
                 .set_print_update(explore_type::print_update)
-                .set_persist_metrics(explore_type::persist_metrics)
+                .set_persist_metrics(VW::cb_explore_adf::persist_metrics)
```
lalo, line 323: if you remove this set, would it be equivalent to this change?

lalo (approved):

PR #4477, opened 2023-01-31
test: modify test a bit

lalo:


test/core.vwtest.json
```diff
@@ -5766,5 +5766,37 @@
...
+    "input_files": [
+      "models/aml_spin_off_bag.m",
+      "train-sets/automl_spin_off_small_2.txt"
+    ],
+    "depends_on": [
+      441
+    ]
+  },
+  {
+    "id": 447,
```
lalo, line 5787: both 446 and 447 should be equal therefore reference same diff_files

PR #4490, opened 2023-02-07
refactor: [all] remove cost_sensitive from workspace

lalo:


olgavrou (approved):

vowpalwabbit/core/src/reductions/cb/warm_cb.cc
```diff
@@ -608,6 +609,8 @@ VW::LEARNER::base_learner* VW::reductions::warm_cb_setup(VW::setup_base_i& stack
   VW::learner_output_example_prediction_func<warm_cb, VW::example>* output_example_prediction_func = nullptr;
   VW::learner_print_update_func<warm_cb, VW::example>* print_update_func = nullptr;
 
+  data->cost_sensitive = base->get_learner_by_name_prefix("cs");
```
lalo, line 612: a bit sketchy that we are leveraging the fact that they all start with the word "cs", in reality this should be augmented into some sort of capability/trait of a reduction

vowpalwabbit/core/src/reductions/cb/cb_explore_adf_cover.cc
```diff
@@ -323,11 +323,12 @@ VW::LEARNER::base_learner* VW::reductions::cb_explore_adf_cover_setup(VW::setup_
   }
 
   auto* scorer = VW::LEARNER::as_singleline(base->get_learner_by_name_prefix("scorer"));
+  auto* cost_sensitive = VW::LEARNER::as_multiline(base->get_learner_by_name_prefix("cs"));
```
jackgerrits, line 326: Should add a check in case there is no cost sensitive learner in the stack

jackgerrits (approved):

vowpalwabbit/core/src/reductions/cb/cb_explore_adf_cover.cc
lalo: it throws https://github.com/VowpalWabbit/vowpal_wabbit/blob/24f66f44a9020437cf9942392925bf7e9daea7b8/vowpalwabbit/core/include/vw/core/learner.h#L521

jackgerrits: ah okay

byronxu99 (approved):

PR #4493, opened 2023-02-07
refactor: [workspace] split 'all' into multiple structs. split config and runtime vars.

lalo:
global_data.h is the main file to review


cs/cli/vowpalwabbit.cpp
```diff
@@ -79,17 +79,17 @@ VowpalWabbitPerformanceStatistics^ VowpalWabbit::PerformanceStatistics::get()
 { // see parse_args.cc:finish(...)
   auto stats = gcnew VowpalWabbitPerformanceStatistics();
 
-  if (m_vw->current_pass == 0)
+  if (m_vw->pc.current_pass == 0)
```
jackgerrits, line 82: pc is pretty ambiguous to me

vowpalwabbit/core/include/vw/core/global_data.h
```diff
@@ -145,38 +217,23 @@ class workspace
   // May be nullptr, so you must check before calling it
   std::function<void(float)> set_minmax;
 
-  uint64_t current_pass;
+  details::passes_config pc;
```
jackgerrits, line 220: Since we dont want these to change - would could only expose these to the overall library via const ref and require using a getter function

cs/cli/vowpalwabbit.cpp
lalo: went for short names for now to avoid a longer name being sprinkled all over the place

vowpalwabbit/core/include/vw/core/global_data.h
lalo: change is interesting, passes for example keeps state of current pass. in this refactor I haven't put much thought on next steps, for now it's about classification and seeing where these sub structs get used with help of grep and the short name

jackgerrits (approved):

vowpalwabbit/core/include/vw/core/global_data.h
```diff
@@ -102,115 +102,13 @@ class invert_hash_info
...
-  // Generic parser interface to make it possible to use any external parser.
-  std::unique_ptr<VW::details::input_parser> custom_parser;
-
-  std::string data_filename;
-
-  bool daemon;
-
-  bool save_per_pass;
-  float initial_weight;
+  bool add_constant;
```
olgavrou, line 109: probs not part of the PR but might be worth evaluating which fields need to be public and which do not

vowpalwabbit/core/include/vw/core/global_data.h
```diff
@@ -239,77 +137,218 @@ class workspace
...
+  bool do_reset_source;
+  std::unique_ptr<all_reduce_base> all_reduce;
+  VW::details::generate_interactions_object_cache generate_interactions_object_cache_state;
+  uint64_t parse_mask;  // 1 << num_bits -1
+};
+
+class parser_runtime
+{
+public:
+  std::string data_filename;
```
olgavrou, line 247: could we expect a user in lib mode to change the data filename from here?

olgavrou: or set a different example_parser

vowpalwabbit/core/include/vw/core/global_data.h
```diff
@@ -239,77 +137,218 @@ class workspace
...
+  std::shared_ptr<details::trace_message_wrapper> trace_message_wrapper_context;
+  std::shared_ptr<std::ostream> trace_message;
 
-  size_t length() { return (static_cast<size_t>(1)) << num_bits; };
+  std::unique_ptr<VW::io::writer> stdout_adapter;
+
+  std::map<uint64_t, VW::details::invert_hash_info> index_name_map;
+  std::shared_ptr<std::vector<char>> audit_buffer;
+  std::unique_ptr<VW::io::writer> audit_writer;
+  VW::metrics_collector global_metrics;
```
olgavrou, line 282: same question for some of the loggers and metrics, could someone running in lib mode want to consume?

jackgerrits (approved):

PR #4494, opened 2023-02-10
ci: [epsilon_decay] benchmarks

bassmang:


test/benchmarks/benchmark_epsilon_decay.cc
```diff
@@ -0,0 +1,57 @@
...
+  using callback_map =
+      typename std::map<size_t, std::function<bool(simulator::cb_sim&, VW::workspace&, VW::multi_ex&)>>;
+  callback_map test_hooks;
+
+  for (auto _ : state)
+  {
+    const size_t num_iterations = 1000;
+    const size_t seed = 99;
+    const std::vector<uint64_t> swap_after = {500};
+    if (use_decay == "decay")
```
byronxu99, line 34: is this variable used like a boolean? if so maybe it should be of type bool.

bassmang: sure just did this get all the args in a consistent way, updated to bool

byronxu99 (approved):

PR #4502, opened 2023-02-16
fix: [epsilon_decay] output prob ACTION_PROBS, skip: 449

bassmang:


vowpalwabbit/core/src/reductions/epsilon_decay.cc
```diff
@@ -455,7 +455,7 @@ std::shared_ptr<VW::LEARNER::learner> VW::reductions::epsilon_decay_setup(VW::se
                  .set_input_label_type(VW::label_type_t::CB)
                  .set_output_label_type(VW::label_type_t::CB)
                  .set_input_prediction_type(VW::prediction_type_t::ACTION_PROBS)
-                 .set_output_prediction_type(VW::prediction_type_t::ACTION_SCORES)
+                 .set_output_prediction_type(VW::prediction_type_t::ACTION_PROBS)
```
olgavrou, line 458: maybe add a runtest with edecay and explore eval?

bassmang: yea good idea

olgavrou (approved):

PR #4504, opened 2023-02-21
chore: force eigen submodule path to output a message if submodule mi…

olgavrou:
…ssing


ext_libs/ext_libs.cmake
```diff
@@ -101,6 +101,10 @@ if(VW_EIGEN_SYS_DEP)
   add_library(eigen INTERFACE)
   target_include_directories(eigen INTERFACE ${EIGEN3_INCLUDE_DIR})
 else()
+  if(NOT EXISTS ${CMAKE_CURRENT_LIST_DIR}/eigen)
```
jackgerrits, line 104: A clone creates the directories for submodules but the directory is empty. We should check for an expected file inside of the submodule.

```suggestion
  if(NOT EXISTS ${CMAKE_CURRENT_LIST_DIR}/eigen/CMakeLists.txt)
```

jackgerrits (approved):

PR #4505, opened 2023-02-22
refactor: make flat_example an implementation detail of ksvm

jackgerrits:
The primary use of `flat_example` is to easily calculate the dot produce of two examples. Therefore, I added functionality to better reflect that. KSVM is pretty complex so I didn't go all out there, but `flat_example` is now an implementation detail.

Additionally, KSVM depends on the cache representation for its model file which is an issue because the cache by definition can change. So there is a TODO to fix that.


vowpalwabbit/core/src/reductions/memory_tree.cc
```diff
@@ -226,39 +227,39 @@ class memory_tree
   }
 };
 
-float linear_kernel(const VW::flat_example* fec1, const VW::flat_example* fec2)
-{
-  float dotprod = 0;
+// float linear_kernel(const VW::flat_example* fec1, const VW::flat_example* fec2)
```
bassmang, line 230: Can this be deleted?

jackgerrits: Yep, deleted

bassmang (approved):

PR #4506, opened 2023-02-23
fix: enforce force reductions to add metrics in their own namespace

### LINKED ISSUE #4498: metrics namespace per reduction
## Short description
Force reductions to only add metrics in their own namespace. The metrics output dictionary should then be some sort of dictionary[reduction_name_str, dictionary]

## **Possible solution/implementation details**
It can be enforced at learner.h, before calling into the func pointer of persist metrics. You create a new namespace based on the name of the learner and its just a dictionary in json terms (metrics.cc)

## **Example/links if any**
https://github.com/VowpalWabbit/vowpal_wabbit/blob/1098efd3e67d1ad913bec1712e6bc52074f884a3/vowpalwabbit/core/include/vw/core/learner.h#L488

freyam:
Fixes #4498 

This pull request modifies the `set_persist_metrics()` function in `learner.h ` to enforce that force reductions can only add metrics in their own namespace. The output dictionary for metrics is now a dictionary of dictionaries, where each sub-dictionary contains the metrics for a single learner. **The key for each sub-dictionary is the learner's name.**

Here are the steps taken to enforce this:

1. Create a new namespace for the learner based on its name:
```cpp
std::string learner_name = this->learner_ptr->_name;
std::map<std::string, float> learner_metrics;
```

2. Filter out metrics from other learners and add only the learner's metrics to the new namespace:
```cpp
for (auto& metric : metrics) {
  if (metric.first.find(learner_name) != std::string::npos) {
    learner_metrics[metric.first] = metric.second;
  }
}
```

3. Add the learner namespace to the output dictionary:
```cpp
std::map<std::string, std::map<std::string, float>> output_metrics;
output_metrics[learner_name] = learner_metrics;
```

4. Call the original function pointer with the new metrics dictionary:
```cpp
fn_ptr(*data, output_metrics);
```

Putting it all together, here's the full implementation:
```cpp
LEARNER_BUILDER_DEFINE(set_persist_metrics(void (*fn_ptr)(DataT&, metric_sink&)),
  assert(fn_ptr != nullptr);
  DataT* data = this->learner_data.get();
  this->learner_ptr->_persist_metrics_f = [fn_ptr, data](metric_sink& metrics) {
    // Create a new namespace for this learner
    std::string learner_name = this->learner_ptr->name();
    std::map<std::string, float> learner_metrics;
    
    // Only add metrics from this learner to the namespace
    for (auto& metric : metrics) {
      if (metric.first.find(learner_name) != std::string::npos) {
        learner_metrics[metric.first] = metric.second;
      }
    }
    
    // Add the learner namespace to the output dictionary
    std::map<std::string, std::map<std::string, float>> output_metrics;
    output_metrics[learner_name] = learner_metrics;
    
    // Call the original function pointer with the new metrics dictionary
    fn_ptr(*data, output_metrics);
  };
)
```

This currently fails as `metrics` is not an iterable object (which makes sense from its definition). I am currently trying to figure out a workaround.

@lalo Can you review this and suggest what I could do instead of Step 2? I would love to learn from you.

### DIFF
--- a/vowpalwabbit/core/include/vw/core/learner.h
+++ b/vowpalwabbit/core/include/vw/core/learner.h
@@ -482,10 +482,18 @@ public:
     this->learner_ptr->_cleanup_example_f = [fn_ptr, data](polymorphic_ex ex) { fn_ptr(*data, ex); };
   )
 
+  // Set the function pointer for persisting metrics for the learner.
+  // This function enforces that force reductions can only add metrics in their own namespace.
+  // The metrics output dictionary is a dictionary of dictionaries, where each sub-dictionary
+  // contains the metrics for a single learner. The key for each sub-dictionary is the learner's name.
   LEARNER_BUILDER_DEFINE(set_persist_metrics(void (*fn_ptr)(DataT&, metric_sink&)),
     assert(fn_ptr != nullptr);
     DataT* data = this->learner_data.get();
-    this->learner_ptr->_persist_metrics_f = [fn_ptr, data](metric_sink& metrics) { fn_ptr(*data, metrics); };
+    std::string learner_name = this->name;
+    this->learner_ptr->_persist_metrics_f = [fn_ptr, data, learner_name](metric_sink& metrics) {
+        metrics[learner_name] = {};
+        fn_ptr(*data, metrics[learner_name]);
+    };
   )
 
   LEARNER_BUILDER_DEFINE(set_pre_save_load(void (*fn_ptr)(VW::workspace& all, DataT&)),


vowpalwabbit/core/include/vw/core/learner.h
```diff
@@ -482,10 +482,18 @@ class common_learner_builder
...
+  // This function enforces that force reductions can only add metrics in their own namespace.
+  // The metrics output dictionary is a dictionary of dictionaries, where each sub-dictionary
+  // contains the metrics for a single learner. The key for each sub-dictionary is the learner's name.
   LEARNER_BUILDER_DEFINE(set_persist_metrics(void (*fn_ptr)(DataT&, metric_sink&)),
     assert(fn_ptr != nullptr);
     DataT* data = this->learner_data.get();
-    this->learner_ptr->_persist_metrics_f = [fn_ptr, data](metric_sink& metrics) { fn_ptr(*data, metrics); };
+    std::string learner_name = this->name;
+    this->learner_ptr->_persist_metrics_f = [fn_ptr, data, learner_name](metric_sink& metrics) {
+        metrics[learner_name] = {};
```
lalo, line 494: here you should be able to tell `metric_sink` that it should push this new context called `learner_name`

something like
```
{
 metrics.push_context(learner_name);
 fn_ptr(*data, metrics);
 metrics.pop_context();
}
```

back inside any fn_ptr() you should be able to only operate inside this subcontext of the metrics object - you should not be able to access other contexts that might also exist.

freyam: Thanks for the review. This makes sense.

RohitRathore1: Hi @lalo, I am having a question. I did not find any references for `push_context` and `pop_context` methods in the `VW::metric_sink` class. Am I missing something over here? Will we have to define it?

PR #4507, opened 2023-02-23
feat: support tags for options

jackgerrits:
An example use case is to tag options which are "driver" options


vowpalwabbit/config/include/vw/config/option.h
```diff
@@ -56,6 +56,25 @@ class base_option
   virtual void accept(typed_option_visitor& handler) = 0;
 
   virtual ~base_option() = default;
+
+  VW_ATTR(nodiscard) const std::vector<std::string>& get_tags() const { return _tags; }
```
lalo, line 60: should we do lowercase only?

jackgerrits: Yeah that seems reasonable. Maybe c++ identifier rules as well

lalo (approved):

olgavrou (approved):

PR #4508, opened 2023-02-24
fix: Fix model merging not reweighting input

jackgerrits:


byronxu99 (approved):

vowpalwabbit/core/src/reductions/gd.cc
```diff
@@ -74,15 +74,19 @@ void merge_weights_with_save_resume(size_t length,
     for (size_t i = 0; i < length; i++) { adaptive_totals[i] += (&(this_model[i << weights.stride_shift()]))[1]; }
   }
 
-  for (size_t i = 0; i < source.size(); i++)
+  std::vector<VW::dense_parameters> weight_copies;
+  weight_copies.reserve(source.size());
+  for (auto i : source)
   {
-    VW::details::do_weighting(output_workspace.normalized_idx, length, adaptive_totals.data(), weights);
+    // There is no copy constructor for weights, so we have to copy manually.
```
byronxu99, line 81: Might be worth adding a `deep_copy` to `dense_parameters`. It would make sense given that there's already a `shallow_copy`.

Could also make a copy constructor that performs a deep copy

jackgerrits: I considered this. Adding a function which copies is sort of annoying since you need to decide if you return a value or pointer. Although this really isn't a big deal now that we can move the type. Yeah, okay ill add a deep copy. I feel like in the past we have deleted copy constructors to avoid accidental copies, really not sure how valid that is as a general rule though

PR #4509, opened 2023-02-24
feat: [LAS] add example ft hash and cache and re-use rows of matrix if actions do not change

olgavrou:
Rough benchmarks on my PC

Best case scenario where actions never change (default number of threads):
- 8 times faster for 300 actions
- 9.7 times faster for 500 actions
- 5.3 times faster for 1k actions
- 6.4 times faster for 2k actions

Worst case scenario where actions change all the time (default number of threads):
- either the same or 1.02/1.18 times slower

TODO
- [x] check what kind of speedup the CI benchmarks will show
- [x] add some unit tests for example feature space hashes


vowpalwabbit/core/include/vw/core/example_predict.h
```diff
@@ -42,6 +42,8 @@ class example_predict
   example_predict(example_predict&& other) = default;
   example_predict& operator=(example_predict&& other) = default;
 
+  void calculate_feature_space_hash();
```
jackgerrits, line 45: Can you make `feature_space_hash` private and make this a lazily evaluated value?

vowpalwabbit/core/src/vw.cc
```diff
@@ -751,6 +751,8 @@ void VW::setup_example(VW::workspace& all, VW::example* ae)
   // Set the interactions for this example to the global set.
   ae->interactions = &all.interactions;
   ae->extent_interactions = &all.extent_interactions;
+
+  if (all.calculate_feature_space_hash) { ae->calculate_feature_space_hash(); }
```
jackgerrits, line 755: If this is lazily evaluated then we don't need to put it in setup_example (which would be a very good thing)

olgavrou: where should it go then?

olgavrou: if it is to go in the las reduction then we won't benefit from the parser thread, is my only thought

jackgerrits: You could make `calculate_feature_space_hash()` return the value and cache it if you'd like to avoid rehashing

olgavrou: I'm confused by your suggestion, it does get calculated and set in example_predict

jackgerrits: `calculate_feature_space_hash();` returns void and not the hash that was calculated. I just mean change `calculate_feature_space_hash()` to return the hash itself and if the hash was already calculated then just return the cached value.

olgavrou: how about just renaming `calculate_feature_space_hash` to `set_feature_space_hash` and if it is already calculated then the call will do nothing

jackgerrits: You still need to know it has been set or not then to use it?

If `calculate_feature_space_hash()` returns the hash then you are guaranteed to get the hash whether it was pre-calculated or not

olgavrou: fair I can see that, updated

vowpalwabbit/core/src/example_predict.cc
```diff
@@ -26,3 +26,21 @@ bool VW::example_predict::iterator::operator!=(const iterator& rhs) const { retu
 
 VW::example_predict::iterator VW::example_predict::begin() { return {feature_space.data(), indices.begin()}; }
 VW::example_predict::iterator VW::example_predict::end() { return {feature_space.data(), indices.end()}; }
+
+uint64_t VW::example_predict::calculate_feature_space_hash()
+{
+  if (feature_space_hash == 0)
```
jackgerrits, line 32: 0 is a valid hash value which makes it impossible to tell the difference between not yet set and a hash of 0.

vowpalwabbit/core/src/example_predict.cc
```diff
@@ -26,3 +26,21 @@ bool VW::example_predict::iterator::operator!=(const iterator& rhs) const { retu
 
 VW::example_predict::iterator VW::example_predict::begin() { return {feature_space.data(), indices.begin()}; }
 VW::example_predict::iterator VW::example_predict::end() { return {feature_space.data(), indices.end()}; }
+
+uint64_t VW::example_predict::calculate_feature_space_hash()
```
jackgerrits, line 30: Should this hash function be order independent or dependent?

olgavrou: since the hash is a sum of hashes and not a chain hash, the order does not make a difference (added a unit test to exercise this)

vowpalwabbit/core/src/example_predict.cc
olgavrou: this will only be zero if there are no features at all (with `--noconstant` set) e.g. `shared` example or an example with no features or features with zero weights

given that there are no features in that case a hash of `0` is fine IMO

vowpalwabbit/core/src/example_predict.cc
olgavrou: thinking through if this makes a difference or not for LAS

vowpalwabbit/core/src/example_predict.cc
jackgerrits: But 0 is a valid hash value, so std::hash could produce zero

vowpalwabbit/core/src/example_predict.cc
jackgerrits: Yeah it depends on if order should matter for the hash, as in are two things in different orders equivalent for the purposes of the hash

vowpalwabbit/core/src/example_predict.cc
olgavrou: Since any number is a valid hash, the thing we could do is add a boolean that is set when the hash is calculated

olgavrou: done

vowpalwabbit/core/src/example_predict.cc
olgavrou: renamed

vowpalwabbit/core/src/vw.cc
```diff
@@ -751,6 +751,11 @@ void VW::setup_example(VW::workspace& all, VW::example* ae)
   // Set the interactions for this example to the global set.
   ae->interactions = &all.interactions;
   ae->extent_interactions = &all.extent_interactions;
+
+  if (all.calculate_feature_space_hash)
```
lalo, line 755: isnt this LAS specific since it's invalidated in [one_pass_svd_impl.cc](https://github.com/VowpalWabbit/vowpal_wabbit/pull/4509/files#diff-a8aeef01b1fd9b1fecf12e7021c61ffd8208eb180cd97cba3643a966668d09cc) line 58

olgavrou: that's LAS's internal cache of example hashes that is has seen

vowpalwabbit/core/include/vw/core/global_data.h
```diff
@@ -116,6 +116,7 @@ class workspace
   std::unique_ptr<all_reduce_base> all_reduce;
 
   bool chain_hash_json = false;
+  bool calculate_feature_space_hash = false;
```
jackgerrits, line 119: Do we know by how much we improve performance by doing this in setup_example? Perhaps we should avoid adding to the global state until a point where it is proven to be a performance issue?

olgavrou: did a quick benchmark, the difference isn't really noticeable so moved calculation into LAS

vowpalwabbit/core/src/reductions/cb/details/large_action/one_pass_svd_impl.cc
```diff
@@ -46,6 +46,8 @@ namespace cb_explore_adf
 
 void one_pass_svd_impl::generate_AOmega(const multi_ex& examples, const std::vector<float>& shrink_factors)
 {
+  for (auto* ex : examples) { ex->feature_space_hash = ex->calculate_order_independent_feature_space_hash(); }
```
jackgerrits, line 49: Is this necessary at the beginning?  If `calculate_order_independent_feature_space_hash()` either calculates it or returns the cached value then you can just call it when the hash is needed

vowpalwabbit/core/include/vw/core/example_predict.h
```diff
@@ -42,6 +42,10 @@ class example_predict
   example_predict(example_predict&& other) = default;
   example_predict& operator=(example_predict&& other) = default;
 
+  // this hashing function does not take into account the order of the features
+  // an example with the exact same namespaces / features-values but in a different order will have the same hash
+  uint64_t calculate_order_independent_feature_space_hash();
```
jackgerrits, line 47:
```suggestion
  uint64_t get_or_calculate_order_independent_feature_space_hash();
```
Makes the caching of the value clearer

jackgerrits (approved):

PR #4513, opened 2023-02-28
fix: multi-model state for cb_adf

bassmang:


vowpalwabbit/core/include/vw/core/gen_cs_example.h
```diff
@@ -12,6 +12,12 @@ namespace VW
 {
 namespace details
 {
+class cb_to_cs_per_model_state
```
jackgerrits, line 15: Might be a good opportunity to move this and name is something more appropriate. Since it is only relevant for MTR

bassmang: Agree with the name change. I think moving cb_to_cs makes sense for a separate PR. @lalo thoughts on naming?

vowpalwabbit/core/src/reductions/cb/cb_adf.cc
```diff
@@ -234,6 +235,7 @@ void VW::reductions::cb_adf::learn(learner& base, VW::multi_ex& ec_seq)
   if (VW::test_cb_adf_sequence(ec_seq) != nullptr)
   {
     _offset = ec_seq[0]->ft_offset;
+    _offset_index = _offset / _all->weights.stride();
```
lalo, line 238: this should be something like _offset / self.increment (self being the learner that owns this cb_adf data instance) but we don't have access to that info at this point

lalo: @jackgerrits

vowpalwabbit/core/src/reductions/cb/cb_adf.cc
```diff
@@ -337,37 +340,50 @@ void save_load(VW::reductions::cb_adf& c, VW::io_buf& model_file, bool read, boo
...
 
 void cb_adf_add(
     const VW::reductions::cb_adf& data1, const VW::reductions::cb_adf& data2, VW::reductions::cb_adf& data_out)
 {
-  data_out.get_gen_cs().event_sum = data1.get_gen_cs().event_sum + data2.get_gen_cs().event_sum;
-  data_out.get_gen_cs().action_sum = data1.get_gen_cs().action_sum + data2.get_gen_cs().action_sum;
+  for (size_t i = 0; i < data1.get_gen_cs().per_model_state.size(); i++)
+  {
+    data_out.get_gen_cs().per_model_state[i].event_sum =
+        data1.get_gen_cs().per_model_state[i].event_sum + data2.get_gen_cs().per_model_state[i].event_sum;
```
lalo, line 371: this could become c.get_gen_cs(ft_offset=i).event_sum or not sure

lalo: it would look more similar to the left side of the diff

bassmang: move to seperate pr

lalo (approved):

vowpalwabbit/core/src/reductions/details/automl/automl_impl.cc
```diff
@@ -335,13 +335,9 @@ void interaction_config_manager<config_oracle_impl, estimator_impl>::do_learning
 {
   assert(live_slot < max_live_configs);
   // TODO: what to do if that slot is switched with a new config?
-  std::swap(*_cb_adf_event_sum, per_live_model_state_uint64[live_slot * 2]);
-  std::swap(*_cb_adf_action_sum, per_live_model_state_uint64[live_slot * 2 + 1]);
   for (example* ex : ec) { apply_config(ex, &estimators[live_slot].first.live_interactions); }
```
lalo, line 338: finally

bassmang: ik great to not have that anymore

vowpalwabbit/core/include/vw/core/gen_cs_example.h
lalo: no idea but yeah could be renamed based on that other pr of interleaved

PR #4520, opened 2023-03-03
feat: [LAS] with CCB

olgavrou:
- caching added also for ccb

speedup for ccb (200 actions 4 slots):

ccb with LAS with cache 13.7 ms 
ccb with LAS without cache 28.8 ms
ccb without LAS 10.3 ms

### DIFF
--- a/test/train-sets/ref/help.stdout
+++ b/test/train-sets/ref/help.stdout
@@ -639,6 +639,11 @@ Weight Options:
                                             with quadratic interactions. (x86 Linux only) (type: bool, experimental)
     --two_pass_svd                          A more accurate svd that is much slower than the default (one
                                             pass svd) (type: bool, experimental)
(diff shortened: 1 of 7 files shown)


olgavrou:
> Just to clarify, LAS should work out-of-box with CCB, and the action_cache_slack is a separate thing that can be used with both CB and CCB?

Yes exactly

`action_cache_slack` is mostly useful for CCB where we can have slots so in order to be able to keep the cache while ccb removes actions in each call but it also affects CB and can be changed/turned off for potential debugging purposes

jackgerrits:
How does the cache work for CCB when each action has the unique slot features added before calling into LAS?

olgavrou:
> How does the cache work for CCB when each action has the unique slot features added before calling into LAS?

Slots are added to the shared example so they are removed along with the shared features

the CCB example added checks this

PR #4527, opened 2023-03-13
test: [automl] add ccb test that checks for ft names

lalo:


.github/workflows/python_wheels.yml
```diff
@@ -70,8 +70,8 @@ jobs:
         shell: bash
         run: |
           pip install -r requirements.txt
-          pip install pytest twine
-          pip install built_wheel/*.whl
+          pip install pytest vw-executor twine
```
lalo, line 73: I can rewrite to avoid dependency -

python/tests/test_ccb.py
```diff
@@ -122,3 +122,130 @@ def test_ccb_non_slot_none_outcome():
...
+
+    automl_metrics = json.load(open(automl.outputs["--extra_metrics"][0]))
+
+    # we need champ switches to be zero for this to work
+    assert "total_champ_switches" in automl_metrics
+    assert automl_metrics["total_champ_switches"] == 0
+
+    fts_names_q = set([n for n in q[0].model9("--invert_hash").weights.index])
+    fts_names_automl = set(
+        [n for n in automl[0].model9("--invert_hash").weights.index if "[" not in n]
```
bassmang, line 241: leave comment that `[` implies a non-zero feature weight

lalo: done

test/train-sets/ref/automl_ccb.stderr
```diff
@@ -0,0 +1,23 @@
+using no cache
+Reading datafile = train-sets/ccb_reuse_small.data
+num sources = 1
+Num weight bits = 18
+learning rate = 0.5
+initial_t = 0
+power_t = 0.5
+cb_type = mtr
+Enabled learners: gd, generate_interactions, scorer-identity, csoaa_ldf-rank, cb_adf, automl, cb_explore_adf_greedy, cb_sample, shared_feature_merger, ccb_explore_adf, extra_metrics
```
lalo, line 9: generate_interactions should not be here, fix in the other PR

Issue #4511, opened 2023-02-27
Worse performance and less training time for the PLT reduction, when updating from VW 9.6 to 9.7


FabianKaiser:
### Describe the bug

The PLT reduction performs worse and trains faster in VW 9.7 when compared to VW 9.6.

The The difference can be seen also in the training logs:
- the loss behaves very different in both versions
- there are less passes used in VW 9.7 (which is likely due to the different loss)
- the performance is worse in VW 9.7 (this is not that drastic in the example data, but more so in larger datasets)

[reddit_data_sample.csv](https://github.com/VowpalWabbit/vowpal_wabbit/files/10837993/reddit_data_sample.csv)

# VW 9.6

PLT k = 1013
kary_tree = 2
creating cache_file = train.vw.cache
Reading datafile = train.vw
num sources = 1
Num weight bits = 30
learning rate = 0.5
initial_t = 0
power_t = 0.5
decay_learning_rate = 1
Enabled reductions: gd, scorer-identity, plt
Input label = MULTILABEL
Output pred = MULTILABELS
average  since         example        example        current        current  current
loss     last          counter         weight          label        predict features
1.000000 1.000000            1            1.0            254                     127
1.000000 1.000000            2            2.0            240                     103
1.000000 1.000000            4            4.0            960                      42
1.000000 1.000000            8            8.0            172                     199
1.000000 1.000000           16           16.0            645                     184
1.000000 1.000000           32           32.0            268                     236
1.000000 1.000000           64           64.0            207                     215
1.000000 1.000000          128          128.0            117                     118
1.000000 1.000000          256          256.0            777                      62
1.000000 1.000000          512          512.0            900                     216
1.000000 1.000000         1024         1024.0            852                     446
1.000000 1.000000         2048         2048.0            389                     190
1.000000 1.000000         4096         4096.0            545                     152
1.000000 1.000000         8192         8192.0            603                      83
1.000244 1.000488        16384        16384.0            730                     101
0.996429 0.996429        32768        32768.0            917                     124 h
0.994506 0.992584        65536        65536.0            977                      89 h
0.991005 0.987503       131072       131072.0            150                      68 h
0.984825 0.978646       262144       262144.0             29                     101 h
0.976190 0.967556       524288       524288.0            804                     108 h

finished run
number of examples per pass = 24300
passes used = 31
weighted example sum = 753300.000000
weighted label sum = 0.000000
average loss = 0.958889 h
total feature number = 103905211

[...]

PR #4534, opened 2023-03-16
feat: add a training loss calculation to the predict method of PLT reduction

mwydmuch:
This is a small change to PLT reduction that adds calculation of training loss also to the `predict` method when labels are available (e.g., when predicting for holdout dataset), instead of always returning 0. Which is a behavior that might cause confusion/problems to some users (as in #4511).

### DIFF
--- a/vowpalwabbit/core/src/reductions/plt.cc
+++ b/vowpalwabbit/core/src/reductions/plt.cc
@@ -139,12 +139,9 @@ void learn(plt& p, learner& base, VW::example& ec)
   auto multilabels = std::move(ec.l.multilabels);
   VW::polyprediction pred = std::move(ec.pred);
 
-  double t = p.all->sd->t;
-  double weighted_holdout_examples = p.all->sd->weighted_holdout_examples;
-  p.all->sd->weighted_holdout_examples = 0;
-
   get_nodes_to_update(p, multilabels);
 
+  double t = p.all->sd->t;
   float loss = 0;
   ec.l.simple = {1.f};
   ec.ex_reduction_features.template get<VW::simple_label_reduction_features>().reset_to_default();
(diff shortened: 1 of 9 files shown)


vowpalwabbit/core/src/reductions/plt.cc
```diff
@@ -139,6 +132,18 @@ void learn(plt& p, learner& base, VW::example& ec)
...
+}
+
+void learn(plt& p, learner& base, VW::example& ec)
+{
+  auto multilabels = std::move(ec.l.multilabels);
+  VW::polyprediction pred = std::move(ec.pred);
+
+  double t = p.all->sd->t;
+  double weighted_holdout_examples = p.all->sd->weighted_holdout_examples;
+  p.all->sd->weighted_holdout_examples = 0;
```
jackgerrits, line 144: Why reset this to zero?

mwydmuch: Good question. This is there from the initial implementation, it is not a change I introduced in this PR, but I don't remember why it's here. Does it somehow impact the update of the base classifier? That probably could be a reason for "resetting" this variable. If not, this can probably be removed.

mwydmuch: Ok, it seems that it doesn't impact the training so I think I will remove these.

jackgerrits: Sounds good!

jackgerrits (approved):
Thanks for this change! The loss being reported now is really helpful.

vowpalwabbit/core/src/reductions/plt.cc
```diff
@@ -154,8 +151,6 @@ void learn(plt& p, learner& base, VW::example& ec)
   for (auto& n : p.negative_nodes) { loss += learn_node(p, n, base, ec); }
 
   p.all->sd->t = t;
```
jackgerrits, line 153: Can you avoid setting t here?

mwydmuch:
Ok, all the checks have passed, so from my point, this is ready to be merged :)

PR #4549, opened 2023-04-04
feat: print features in version

jackgerrits:


lalo (approved):

vowpalwabbit/core/src/parse_args.cc
```diff
@@ -332,6 +333,7 @@ void parse_diagnostics(options_i& options, VW::workspace& all)
   if (version_arg)
   {
     std::cout << VW::VERSION.to_string() << " (git commit: " << VW::GIT_COMMIT << ")\n";
+    std::cout << "Enabled features: " << VW::ENABLED_FEATURES << "\n";
```
lalo, line 336: should be more explicit like Compiled features in binary:? Enabled sound like they are on - could cause confusion with flatbuffers that it is in there but not used/enabled

jackgerrits: yeah that is a good point

lalo (approved):

PR #4555, opened 2023-04-05
build: make lda optional with compile feature

jackgerrits:


cmake/VowpalWabbitFeatures.cmake
```diff
@@ -8,17 +8,12 @@
...
-#   CSV: Enable csv parser
-#   LAS_SIMD: Enable large action space with explicit simd (only work with linux for now)
 
-# set(VW_ALL_FEATURES "LAS_SIMD;FLATBUFFERS;CSV")
-set(VW_ALL_FEATURES "CSV;FLATBUFFERS")
+set(VW_ALL_FEATURES "CSV;FLATBUFFERS;LDA")
 
 option(VW_FEAT_FLATBUFFERS "Enable flatbuffers support" OFF)
 option(VW_FEAT_CSV "Enable csv parser" OFF)
+option(VW_FEAT_LDA "Enable lda reduction" ON)
```
bassmang, line 16: Is there a reason to not set this OFF by default?

ext_libs/README
```diff
@@ -0,0 +1,9 @@
+boost_math - optional
```
lalo, line 1: thoughts?

cmake/VowpalWabbitFeatures.cmake
jackgerrits: I wanted to maintain current behavior, so its on by default

ext_libs/README
jackgerrits: Seems reasonable

bassmang (approved):

PR #4556, opened 2023-04-05
feat: Add emt_initial flag to the EMT reduction

mrucker:
This pull request does: 

1. Adds the `--emt_initial` flag which selects the distance function used to initialize the scoring function.
2. Improves the runtime performance of EMT on datasets with large numbers of featres
3. Adds the new emt_initial flag to the model saving/loading subroutines
4. Adds unittests for all the above changes


vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -721,6 +880,10 @@ size_t read_model_field(io_buf& io, reductions::eigen_memory_tree::emt_tree& tre
   bytes += read_model_field(io, router_type);
   tree.router_type = static_cast<reductions::eigen_memory_tree::emt_router_type>(router_type);
 
+  uint32_t initial_type{};
+  bytes += read_model_field(io, initial_type);
+  tree.initial_type = static_cast<reductions::eigen_memory_tree::emt_initial_type>(initial_type);
+
```
jackgerrits, line 886: You don't need to save it like this since its a command line option, you can just use .keep() on the option

mrucker: Oh that's great to know. I've actually always wondered what `.keep()` did. I'll update the code.

mrucker: It has been removed (along with several other EMT parameters that could be removed from save/load due to .keep). 

I also added checks to the save load unittest to for all parameters I removed from explicit saving and loading.

vowpalwabbit/core/src/reductions/eigen_memory_tree.cc
```diff
@@ -710,34 +869,14 @@ size_t write_model_field(
 size_t read_model_field(io_buf& io, reductions::eigen_memory_tree::emt_tree& tree)
 {
   size_t bytes = 0;
-
-  bytes += read_model_field(io, tree.leaf_split);
```
jackgerrits, line 714: Since emt has been released we need to be careful about changed to save/load.

If you want to change this save/load code then please add a check  in your save load function which throws an error if a model is loaded which is of a version before these changes were made. You can find examples of this in other reductions

mrucker: Oh, that makes sense. The save changes aren't needed. My vote would be that maintaining two separate save versions probably isn't worth it in this case. Unless you feel differently, I've simply put back the old save logic without adding the new flag.

jackgerrits: Yeah i agree, leaving as is is the least disruption

mrucker:
Hey Guys,

I may have lost the thread here. Are you waiting on me to make further changes to this request? I can't remember. Thanks.

lalo (approved):

PR #4565, opened 2023-04-18
feat(CB_GF): correct update rule and simulation unit test

olgavrou:
- allow multiple costs if graph feedback

### DIFF
--- a/vowpalwabbit/core/include/vw/core/reductions/cb/cb_explore_adf_common.h
+++ b/vowpalwabbit/core/include/vw/core/reductions/cb/cb_explore_adf_common.h
@@ -87,7 +87,8 @@ class cb_explore_adf_base
 {
 public:
   template <typename... Args>
-  cb_explore_adf_base(bool with_metrics, Args&&... args) : explore(std::forward<Args>(args)...)
+  cb_explore_adf_base(bool with_metrics, Args&&... args)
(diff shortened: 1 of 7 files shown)


vowpalwabbit/core/include/vw/core/reductions/cb/cb_explore_adf_common.h
```diff
@@ -105,6 +105,7 @@ class cb_explore_adf_base
       const multi_ex& ec_seq, VW::io::logger& logger);
 
   ExploreType explore;
+  static bool allow_multiple_costs;
```
jackgerrits, line 108: Does this need to be static?

olgavrou: yes since it is used in predict/learn which are static

jackgerrits: If there are two workspace instances in the same process, one using cb with graph feedback and one not using it then this static would essentially break that scenario.

If this needs to be dynamic based on the setup then the variable and learn/predict should be non static

olgavrou: actually it doesn't need to be static and good point, updated

jackgerrits (approved):

PR #4568, opened 2023-04-20
feat(CB_GF): accept graph in json format

olgavrou:
and tests and graph validation

graph accepted as:

```
"_graph" : [{"row": 0, "col": 1, "val": 2}, {"row": 1, "col": 1, "val": 3}]
```


vowpalwabbit/core/src/reductions/cb/cb_explore_adf_graph_feedback.cc
```diff
@@ -421,11 +424,12 @@ void cb_explore_adf_graph_feedback::update_example_prediction(multi_ex& examples
       a_s.begin(), a_s.end(), [](const VW::action_score& a, const VW::action_score& b) { return a.score > b.score; });
 }
 
-arma::sp_mat get_graph(const VW::cb_graph_feedback::reduction_features& graph_reduction_features, size_t num_actions)
+arma::sp_mat get_graph(
+    const VW::cb_graph_feedback::reduction_features& graph_reduction_features, size_t num_actions, VW::workspace& all)
```
jackgerrits, line 428: Pass logger instead of all object

jackgerrits (approved):

PR #4574, opened 2023-04-26
feat: WASM initial CB model API

olgavrou:
For the cb model, the example is not opaque but a javasctipt object that the C++ side can inspect

Moved basic vw functionality out into `vw_model_basic` class which is used as a mixin for the `vw_model` which is generic vw model implementation and the newly added `cb_vw_model`

`vw_model_basic` can be enhanced in the future with any other generic functionality

### DIFF
--- a/wasm/wasm_wrapper.cc
+++ b/wasm/wasm_wrapper.cc
@@ -285,6 +285,8 @@ struct cb_vw_model : MixIn
     // TODO else if json_context else if embeddings vector of some kind, else throw
 
     auto example_list = parse(context);
+    assert(!example_list.empty());
+
     for (auto* ex : example_list) { VW::setup_example(*this->vw_ptr, ex); }
 
     this->vw_ptr->predict(example_list);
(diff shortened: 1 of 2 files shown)


wasm/wasm_wrapper.cc
```diff
@@ -235,28 +255,98 @@ struct vw_model
...
+    size_t index_offset = 0;
+    if (VW::LEARNER::ec_is_example_header(
+            *example_list[0], this->vw_ptr->parser_runtime.example_parser->lbl_parser.label_type))
+    {
+      index_offset = 1;
+    }
+
+    for (unsigned int i = 0; i < length; i++)
+    {
+      // TODO and test for index out of bounds i.e. incorrect label
```
olgavrou, line 312: I'll do this todo in this PR

wasm/wasm_wrapper.cc
```diff
@@ -40,6 +40,9 @@ void validate_options(const VW::config::options_i& options)
   if (!options.get_positional_tokens().empty()) { THROW("Positional options are not allowed") }
 }
 
+struct vw_model_basic;
```
cheng-tan, line 43: can we remove Line 43 - 45 and set the default MinxIn on Line183? to be consistent with cb_vw_model?

olgavrou: these need to be forward declared so that vw_model can be referenced in line 55

cheng-tan (approved):

lalo (approved):

PR #4575, opened 2023-04-27
feat: WASM VW JS wrapper file/classes

olgavrou:
- also added save load
- PR was getting big so thought I'd stop here

### DIFF
--- a/wasm/vw.js
+++ b/wasm/vw.js
@@ -1,12 +1,6 @@
 const VWWasmModule = require('./out/vw-wasm.js');
 const fs = require('fs');
 
-const ModelType =
-{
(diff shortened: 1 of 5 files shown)


lalo (approved):

wasm/vw.js
```diff
@@ -0,0 +1,133 @@
+const { log } = require('console');
+const fs = require('fs');
+const VWWasmModule = require('./out/vw-wasm.js');
+
+
+// internals
+
+const ModelType =
```
lalo, line 8: should this be ProblemType

olgavrou: sure why not, changed

PR #4581, opened 2023-05-08
fix(CB_GF): correct constraints, huge cleanup, and clip and normalize probabilities

olgavrou:
test results are a bit less dramatic


olgavrou:
reviewed with Paul offline and extensively tested with cvxpy version so will merge

PR #4591, opened 2023-05-17
test: [RLOS_2023][WIP] add test for regression weight 

michiboo:
Add constant function test for regression

### DIFF
--- a/python/tests/test_regression.py
+++ b/python/tests/test_regression.py
@@ -1,0 +1,62 @@
+from vw_executor.vw import Vw
+from vw_executor.vw_opts import Grid
+from numpy.testing import assert_allclose
+import pandas as pd
+import numpy as np
+import random
+
+with open("regression1.txt", "w") as f:
+    for i in range(10000):
+        x = random.uniform(1, 100)
+        y = 5
+        f.write(f"{y} |f x:{x}\n")
+
+
+def core_test(files, grid, outputs, job_assert):
+    vw = Vw(".vw_cache", reset=True)
+    result = vw.train(files, grid, outputs)
+    job_assert(result)
+
+
+def test_regression_weight():
+    options = Grid(
+        {
+            "#base": ["-P 50000 --preserve_performance_counters --save_resume "],
+            "#reg": ["", "--coin", "--ftrl", "--pistol"],
+        }
+    )
+
+    def assert_job(job):
+        tolerance = 1
+        constant = 5
+        for j in job:
+            assert j.status.value == 3
+            data = j.outputs["--readable_model"]
+            with open(data[0], "r") as f:
+                data = f.readlines()
+            data = [i.strip() for i in data]
+            weight_b = float(data[-1].split(":")[1].split(" ")[0])
+            weight_x1 = float(data[-2].split(":")[1].split(" ")[0])
+            print(f"Running for command {j.opts}")
+            print(weight_x1, weight_b)
+            err_msg = (
+                lambda w, we, wName: f"assert fail for {wName}: actual: {w}, expected {we}"
+            )
+            assert np.isclose(weight_b, constant, atol=tolerance), err_msg(
+                weight_b, constant, "Bias"
+            )
+            assert np.isclose(weight_x1, 0, atol=tolerance), err_msg(
+                weight_x1, 0, "w_x1"
+            )
+
+            # checking prediction
+            # predictions = j.outputs['-p']
+            # with open(predictions[0], "r") as f:
+            #     predictions = f.readlines()
+            #     predictions = [float(i) for i in predictions[1:]]
+            # assert assert_allclose(predictions, [5]*len(predictions)), "predicted value should be 5"
+
+    core_test("regression1.txt", options, ["--readable_model", "-p"], assert_job)
+
+
+test_regression_weight()


python/tests/test_regression.py
```diff
@@ -0,0 +1,62 @@
+from vw_executor.vw import Vw
+from vw_executor.vw_opts import Grid
+from numpy.testing import assert_allclose
+import pandas as pd
+import numpy as np
+import random
+
+with open("regression1.txt", "w") as f:
```
ataymano, line 8: Let's make this call either from test itself or from fixture.

python/tests/test_regression.py
```diff
@@ -0,0 +1,62 @@
...
+    for i in range(10000):
```
ataymano, line 9: Generating constant dataset seems generic enough function with data size/constant/... as parameters

python/tests/test_regression.py
```diff
@@ -0,0 +1,62 @@
...
+        x = random.uniform(1, 100)
```
ataymano, line 10: it is not great to have true randomness in tests. We should seed it before every call

python/tests/test_regression.py
```diff
@@ -0,0 +1,62 @@
...
+    def assert_job(job):
+        tolerance = 1
+        constant = 5
+        for j in job:
+            assert j.status.value == 3
+            data = j.outputs["--readable_model"]
+            with open(data[0], "r") as f:
+                data = f.readlines()
+            data = [i.strip() for i in data]
+            weight_b = float(data[-1].split(":")[1].split(" ")[0])
```
ataymano, line 38: https://github.com/VowpalWabbit/data-science/blob/b9cadfed0622e99050afa5685d77c33aee8703c5/from_mwt_ds/DataScience/vw_executor/artifacts.py#L220

you can have access to weights as job[0].Model9('--readable_model').weights

python/tests/test_regression.py
```diff
@@ -0,0 +1,62 @@
...
+            "#base": ["-P 50000 --preserve_performance_counters --save_resume "],
+            "#reg": ["", "--coin", "--ftrl", "--pistol"],
+        }
+    )
+
+    def assert_job(job):
+        tolerance = 1
+        constant = 5
+        for j in job:
+            assert j.status.value == 3
```
ataymano, line 33: let's use enum here:
https://github.com/VowpalWabbit/data-science/blob/b9cadfed0622e99050afa5685d77c33aee8703c5/from_mwt_ds/DataScience/vw_executor/vw.py#L30

python/tests/test_regression.py
```diff
@@ -0,0 +1,62 @@
...
+
+def test_regression_weight():
+    options = Grid(
+        {
+            "#base": ["-P 50000 --preserve_performance_counters --save_resume "],
+            "#reg": ["", "--coin", "--ftrl", "--pistol"],
+        }
+    )
+
+    def assert_job(job):
```
ataymano, line 29: This code seems quite generic - "compare weights with expected values with given tolerance", but not reusable at the moment.

python/tests/test_regression.py
ataymano, line 8: let's also write all data files in separate folder ('data' or something)

python/tests/test_regression.py
```diff
@@ -0,0 +1,62 @@
...
+
+with open("regression1.txt", "w") as f:
+    for i in range(10000):
+        x = random.uniform(1, 100)
+        y = 5
+        f.write(f"{y} |f x:{x}\n")
+
+
+def core_test(files, grid, outputs, job_assert):
+    vw = Vw(".vw_cache", reset=True)
```
ataymano, line 16: probably better to add handler=None. I think now it is trying to draw progress bars.

python/tests/pytest.json
```diff
@@ -0,0 +1,32 @@
...
+            "#base": ["-P 50000 --preserve_performance_counters --save_resume "],
+            "#reg": ["", "--coin"]
+        }
+    },
+    {
+        "data_func": "constant_function",
+        "data_func_args": [2000,5,1,100],
+        "assert_func": "assert_weight",
+        "assert_func_args": [[5, 0],1],
+        "grid":   {
```
ataymano, line 27: learning_rate is specific argument for sgd ("") - others learners have another hyperparmeters (i.e. ftrl has ftrl_alpha and ftrl_beta).

In python there is a way to use '+'/'*' operator over grids in the following way:
```
grid = Grid(
    {
        '#base': '-P 1000 ...'
    }) * (
        Grid({
            "#lr": "",
            "--learning_rate": [0.01, 0.1, 1]
            }) + Grid({
                '#lr': '--ftrl',
                '--ftrl_alpha': [0.01, 0.1, 1]
                }))
```
Would be good to consider being able to express that kind of things in the future.

python/tests/test_helper.py
```diff
@@ -0,0 +1,69 @@
...
+import os
+
+# Get the current directory
+current_dir = os.path.dirname(os.path.abspath(__file__))
+
+def json_to_dict_list(file):
+    res = []
+    with open(current_dir + "/" + file, 'r') as file:
+        # Load the JSON data
+        json_data = json.load(file)
```
ataymano, line 13: looks like just returning json_data should work here

python/tests/test_regression.py
```diff
@@ -0,0 +1,41 @@
...
+        job_assert(j, *job_assert_args)
+
+@pytest.mark.parametrize('test_dict', json_to_dict_list("pytest.json"), indirect=True)
+def test_all(test_dict):
+    options = Grid(
+      test_dict['grid'],
+    )
+    data = dynamic_function_call("data_generation", test_dict['data_func'], *test_dict["data_func_args"])
+    assert_job = get_function_object("assert_job", test_dict['assert_func'])
+    core_test(data, options, ["--readable_model", "-p"], assert_job, test_dict['assert_func_args'])
```
ataymano, line 41: Can we move output configuration (readable_model/p/...) to json description

python/tests/pytest.json
ataymano: this is just to consider for the future

python/tests/pytest.json
```diff
@@ -0,0 +1,32 @@
...
+            "#base": ["-P 50000 --preserve_performance_counters --save_resume "],
+            "#reg": ["", "--coin"]
+        }
+    },
+    {
+        "data_func": "constant_function",
+        "data_func_args": [2000,5,1,100],
+        "assert_func": "assert_weight",
```
ataymano, line 15: Please consider supporting list of assertions

python/tests/test_regression.py
```diff
@@ -0,0 +1,41 @@
+from vw_executor.vw import Vw
```
ataymano, line 1: this file seems not related to regression but more of a core infrasturcture.
Better to rename it to "core.py/test.py/..."

PR #4593, opened 2023-05-18
feat: WASM packaging

olgavrou:
node can consume the modularized version that the browser plugin consumes, no need to build both


wasm/vw.js
olgavrou, line 1: the diff here is just consuming the module in the modularized version which returns a promise that you need to check for instead of `onRuntimeInitialized` and also just moved the class definitions to after the module has been properly loaded

PR #4596, opened 2023-05-19
fix: WASM package scoping and cleanup error class check

olgavrou:


wasm/README.md
```diff
@@ -211,15 +211,15 @@ vwPromise.then((vw) => {
 
 ### How-To handle errors
 
-Some function calls with throw if something went wrong or if they were called incorrectly. There are two type of errors that can be thrown: native JavaScript errors and WebAssembly runtime errors.
+Some function calls with throw if something went wrong or if they were called incorrectly. There are two type of errors that can be thrown: native JavaScript errors and WebAssembly runtime errors which are wrapped in a VWError object.
```
lalo, line 214: only the latter one is wrapped in VWError or both? how should one read this last sentence?

olgavrou: updated

PR #4600, opened 2023-05-30
test: [RLOS_2023][WIP] updated test for regression weight 

michiboo:
@ataymano

### DIFF
--- a/python/tests/pytest.json
+++ b/python/tests/pytest.json
@@ -3,7 +3,10 @@
         "data_func": "constant_function",
         "data_func_args": [2000,5,1,100],
         "assert_func": "assert_prediction",
-        "assert_func_args": [5,1],
+        "assert_func_args": {
+            "expected_value": 5,
+            "threshold":0.5
+        },
         "grid":   {
             "#base": ["-P 50000 --preserve_performance_counters --save_resume "],
             "#reg": ["", "--coin"]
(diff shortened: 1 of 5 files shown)


python/tests/pytest.json
```diff
@@ -0,0 +1,54 @@
...
+            "expected_weights":[5, 0],
+            "atol": 100,
+            "rtol": 10
+        },
+        "grid":   {
+            "#base": ["-P 1 --preserve_performance_counters --save_resume"],
+            "#reg": ["", "--coin", "--ftrl", "--pistol"]
+        },
+        "output": ["--readable_model", "-p"],
+        "*" : {
```
ataymano, line 45: not sure if I follow the syntax here: seems like this is outside of "grid" object?

michiboo: so bascially * allow users to have multiple tests so for example, 
if  "--learning_rate": [0.01, 0.001, 0.1] is in *, the it will created a grid object with each of these values in --learning_rate.
if there are multiple parameters in *, then it will produce all combination for example:
 {parameters in grid}  --learning_rate 0.01 --loss_function absolute ..
 {parameters in grid}  --learning_rate 0.001 --loss_function absolute ..
 {parameters in grid}  --learning_rate 0.01 --loss_function quantile ..
 {parameters in grid}  --learning_rate 0.001 --loss_function quantile ..

ataymano (approved):

PR #4605, opened 2023-06-02
feat: flag to save and load per model state

bassmang:


vowpalwabbit/core/src/reductions/gd.cc
```diff
@@ -1461,7 +1486,10 @@ std::shared_ptr<VW::LEARNER::learner> VW::reductions::gd_setup(VW::setup_base_i&
       .add(make_option("l2_state", local_contraction)
                .allow_override()
                .default_value(L2_STATE_DEFAULT)
-               .help("Amount of accumulated implicit l2 regularization"));
+               .help("Amount of accumulated implicit l2 regularization"))
+      .add(make_option("per_model_save_load", per_model_save_load)
```
lalo, line 1490: mark as keep()

PR #4608, opened 2023-06-05
fix: vcpkg for windows ci

bassmang:


.github/workflows/build_windows_cmake.yml
```diff
@@ -24,38 +24,33 @@ jobs:
     env:
       CMAKE_BUILD_DIR: ${{ github.workspace }}/vw/build
       SOURCE_DIR: ${{ github.workspace }}/vw
-      VCPKG_ROOT: ${{ github.workspace }}/vcpkg
-      VCPKG_REF: f14984af3738e69f197bf0e647a8dca12de92996
+      VCPKG_ROOT_VW: ${{ github.workspace }}/vw/ext_libs/vcpkg
```
jackgerrits, line 27: I think `VCPKG_ROOT` is read by some of the tooling and used. What is the reason for changing it?

bassmang: It was getting reset to the wrong value in ilammy/msvc-dev-cmd@v1, I can try to look for another workaround

bassmang:
** approved offline

Issue #4597, opened 2023-05-25
can't get proper prediction score from python when --boosting N option used at training time


jprobichaud:
### Describe the bug

I've trained a small model using the `--boosting N` option, like for example `--loss_function logistic -b 18 --l1 0.1 --l2 0.0001 --nn 50 --boosting 5` using python or the command line.

Then, when I try to get prediction scores out of it, I get "1.0" all the time when using python or using `-p` on the command line. When I use the `-r` command line parameter with vw cli executable, I get the list of all features with their scores and a final meaninfgul score at the end.

It seems impossible to get these values from within the python classes, no matter what I try (playing with the various `PredictionType` )

This seems either a documentation bug or a more serious issue w.r.t. boosting reduction perhaps.

### How to reproduce

1. train with `--loss_function logistic -b 18 --l1 0.1 --l2 0.0001 --nn 50 --boosting 5`  
2. try to get predictions within python that matches the scores you get from using `vw -i model.vw -t test.txt -r /dev/stdout`

### Version

9.2.0

### OS

Linux

### Language

Python

ataymano:
Seems like a bug with boosting (and nn/loss function are not relevant here).
Simpler repro (can be done with binary and -p as well, so problem is not python specific):
```
from vowpalwabbit import pyvw
import numpy as np

vw = pyvw.Workspace('--boosting 1 -b 1')

yhat = lambda x: 2 * x + 3

for i in range(10000):
    x = np.random.rand()
    vw.learn(f'{yhat(x)} | 1:{x}')

print(f'w_x = {vw.get_weight(1, 0)}')
print(f'constant = {vw.get_weight(0, 0)}')

x = 2
print(f'y({x}) = {vw.predict("| 1:{x}")}')
```
Output:
```
w_x = 1.9999972581863403
constant = 3.000001907348633
y(2) = 1.0
```

ataymano:
yeah seems like some binary classification logic is hardcoded there:
https://github.com/VowpalWabbit/vowpal_wabbit/blob/63090d6f1b1810a9e0f0d00ca8573317a7363bbb/vowpalwabbit/core/src/reductions/boosting.cc#L170

JohnLangford:
This issue was fixed in PR #4611 (commit dfef80826) on September 1, 2023.

[...]

PR #4611, opened 2023-06-08
fix!: no binary stuff in boosting

ataymano:
1) no binary classification code in boosting
2) boosting below binary in reduction stack
3) **breaking change (since old "--boosting n" becomes "--boosting n --binary")**


ataymano:
#4597

jackgerrits (approved):

PR #4612, opened 2023-06-08
test: [RLOS_2023][WIP]  test for contextual bandit

michiboo:


python/tests/pytest.json
```diff
@@ -0,0 +1,54 @@
+[
+    {
+        "data_func": "constant_function",
+        "data_func_args": [2000,5,1,100],
+        "assert_func": "assert_prediction",
```
ataymano, line 5: looks like should be updated

python/tests/test_cb.json
```diff
@@ -0,0 +1,52 @@
...
+    {
+        "data_func": "generate_cb_data",
+        "data_func_args": {
+            "num_examples": 100,
+            "num_features": 1,
+            "num_action": 1
+        },
+        "assert_functions": [
+            {
+                "assert_func": "assert_loss",
```
ataymano, line 11: better to expose both reward and expected loss values as parameters to this config file

python/tests/core.py
```diff
@@ -0,0 +1,72 @@
...
+from numpy.testing import assert_allclose
+import pandas as pd
+import numpy as np
+import pytest
+import os
+from test_helper import json_to_dict_list, dynamic_function_call, get_function_object, generate_string_combinations
+
+CURR_DICT = os.path.dirname(os.path.abspath(__file__))
+
+def combine_list_cmds_grids(cmds, base_grid):
```
ataymano, line 12: seems like a lot of it is already implemented as __add__/__mul__ operators of Grid and can be reused

python/tests/reward_functions.py
```diff
@@ -0,0 +1,22 @@
...
+    return 1
+
+
+def constant_reward(**kwargs):
+    reward = kwargs["reward"]
+    chosen_action = kwargs["chosen_action"]
+    return reward[chosen_action - 1]
+
+
+def fixed_reward_for_diff_context(**kwargs):
```
ataymano, line 11: name seems to be misleading

python/tests/core.py
```diff
@@ -0,0 +1,70 @@
...
+
+@pytest.mark.parametrize("test_descriptions", TEST_CONFIG_FILES, indirect=True)
+def test_all(test_descriptions):
+    for test_description in test_descriptions:
+        mutiply = test_description.get("*", None)
+        plus = test_description.get("+", None)
+        grid = Grid(test_description["grid"])
+
+        if mutiply:
+            grid = grid.__mul__(mutiply)
```
ataymano, line 52:
```suggestion
            grid = grid * mutiply
```

python/tests/core.py
```diff
@@ -0,0 +1,70 @@
...
+        if plus:
+            grid.__mul__(plus)
```
ataymano, line 54:
```suggestion
            grid = grid + plus
```

python/tests/probability_functions.py
```diff
@@ -0,0 +1,7 @@
+def constant_probability(chosen_action, **kwargs):
```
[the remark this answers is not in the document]
ataymano: the fact that this function both can be represented as even_probability and not using any of its arguments is confuing.
Can we either remove it or make it with no arguments?

python/tests/reward_functions.py
```diff
@@ -0,0 +1,22 @@
+def fixed_reward(**kwargs):
+    return 1
+
+
+def constant_reward(**kwargs):
```
ataymano, line 5: kwargs seem to be replaceable with regular arguments

python/tests/test_cb.json
```diff
@@ -0,0 +1,151 @@
...
+        "data_func": "generate_cb_data",
+        "data_func_args": {
+            "num_examples": 100,
+            "num_features": 1,
+            "num_action": 1,
+            "reward_function": {
+                "name": "fixed_reward",
+                "params": {}
+            },
+            "probability_function": {
```
ataymano, line 13: better to rename to "logging_policy"

python/tests/reward_functions.py
ataymano, line 5:
```suggestion
def constant_reward(reward: float):
```

python/tests/reward_functions.py
```diff
@@ -0,0 +1,22 @@
+def fixed_reward(**kwargs):
```
ataymano, line 1: Let's do fixed signature here:
def reward_func(context: int, chosen_action: int)

python/tests/reward_functions.py
```diff
@@ -0,0 +1,22 @@
...
+def constant_reward(**kwargs):
+    reward = kwargs["reward"]
+    chosen_action = kwargs["chosen_action"]
+    return reward[chosen_action - 1]
+
+
+def fixed_reward_two_action(**kwargs):
+    chosen_action = kwargs["chosen_action"]
+    chosen_context = kwargs["chosen_context"]
+    if chosen_context == 1 and chosen_action == 2:
```
ataymano, line 14: better to expose this logic to config file somehow:
1) either via more descriptive function name
2) or via extra parameter (like reward matrix)

ataymano: but let's do it in next Pr

PR #4618, opened 2023-06-28
test: [RLOS_2023] [WIP] Support + and * expression for grids

michiboo:


python/tests/test_helper.py
```diff
@@ -14,6 +14,41 @@ def json_to_dict_list(file):
...
+
+    for i, item in enumerate(config):
+        if isinstance(item, dict):
+            if expression and (expression[-1].isdigit()):
+                expression += " * "
+            expression += "a" + str(config.index(item))
+
+        elif isinstance(item, str):
+            if item == "(":
+                if expression[-1].isdigit():
```
ataymano, line 41: will fail if expression is empty

python/tests/test_core.py
```diff
@@ -55,18 +61,13 @@ def init_all(test_descriptions):
...
-            if plus:
-                grid += plus
-            options = Grid(grid)
+            grid_expression = generate_mathematical_expression_json(
+                test_description["grids"]
+            )
+            variables = {}
+            for i, item in enumerate(test_description["grids"]):
+                variables["a" + str(i)] = Grid(item)
+            options = evaluate_expression(grid_expression, variables)
```
ataymano, line 70: create helper function like options = get_options(test_description["grids"])

ataymano (approved):

PR #4621, opened 2023-07-07
fix: update to reflect the fact that coin's learn does not return a prediction

jackgerrits:
The prediction it currently returns essentially represents what the prediction would have been post learn and therefore does not match the contract of learn_returns_prediction,


ataymano:
Just checked manually - example from above got passed.

PR #4627, opened 2023-07-23
test: [RLOS2023][WIP] add option for storing output and grid language redefinition

michiboo:


python/tests/test_framework/test_configs/cb.json
```diff
@@ -49,90 +49,56 @@
...
+                    ""
+                ]
+            },
+            {
+                "--softmax": [
+                    ""
+                ]
+            }
+        ],
+        "grids_expression": "g0 * (g1 + g2 + g3 + g4 + g5 +g6 + g7 + g8)",
```
ataymano, line 101: It is hard to track down which one is g3 for example. Maybe we can turn current list into disctionary with explicit namings?

ataymano (approved):

PR #4629, opened 2023-07-31
test: [RLOS2023] add test for slate

michiboo:


python/tests/test_framework/test_configs/slate.json
```diff
@@ -0,0 +1,130 @@
+[
+    {
+        "test_name": "slates",
+        "data_func": {
+            "name": "generate_slate_data",
+            "params": {
+                "num_examples": 1000,
+                "num_action": [
```
ataymano, line 8: looks like there is typo: num_action instead of num_actions and we are not detecting it since we are relying on *args call in test_core. Probably better to switch to **kwargs one and enforce named arguments always

python/tests/test_framework/test_configs/slate.json
```diff
@@ -0,0 +1,130 @@
...
+                    3,
+                    3,
+                    3
+                ],
+                "reward_function": {
+                    "name": "changing_reward",
```
ataymano, line 14: Non-stationarity of reward is not exposed to this file.

python/tests/test_framework/slate/data_generation.py
```diff
@@ -0,0 +1,78 @@
...
+def generate_slate_data(
+    num_examples,
+    num_actions,
+    reward_function,
+    logging_policy,
+    action_space,
+    num_slots=1,
+    num_context=1,
+    context_name=None,
+    slot_name=None,
```
ataymano, line 23: Probably better to remove slot_name and num_slots for now and start with supporting automatic slot naming (with inference of slots number from action_space output)

python/tests/test_framework/test_configs/slate.json
```diff
@@ -0,0 +1,130 @@
+[
+    {
+        "test_name": "slates",
+        "data_func": {
+            "name": "generate_slate_data",
+            "params": {
+                "num_examples": 1000,
+                "num_action": [
+                    3,
+                    3,
```
ataymano, line 10: [nit] maybe starting with 2x2 problem will make it more readable without actual coverage loss

python/tests/test_framework/test_configs/slate.json
```diff
@@ -0,0 +1,130 @@
...
+                    ""
+                ]
+            },
+            "g8": {
+                "--softmax": [
+                    ""
+                ]
+            }
+        },
+        "grids_expression": "g0 * (g1 + g2 + g3 + g4 + g5 +g6 + g7 + g8)",
```
ataymano, line 124: changing g0/g1/.. names to readable ones (like cover/bag/...) will make this expression more readable.

python/tests/test_framework/slate/data_generation.py
```diff
@@ -0,0 +1,72 @@
...
+script_directory = os.path.dirname(os.path.realpath(__file__))
+random.seed(10)
+
+
+def generate_slate_data(
+    num_examples,
+    reward_function,
+    logging_policy,
+    action_space,
+    num_context=1,
```
ataymano, line 14: num_context is conflicting with context_name. Can we allow context_name to be a list and infer no_context from it?

python/tests/test_framework/test_configs/slate.json
```diff
@@ -0,0 +1,144 @@
...
+                        ]
+                    }
+                }
+            }
+        },
+        "assert_functions": [
+            {
+                "name": "assert_loss",
+                "params": {
+                    "expected_loss": 0.8,
```
ataymano, line 60: expected loss should be slightly more than -2, but CI is still green

PR #4630, opened 2023-08-09
test: [RLOS2023] add test for cb with continous action 

michiboo:


python/tests/test_framework/test_configs/cb_cont.json
```diff
@@ -0,0 +1,189 @@
+[
+    {
+        "test_name": "cb_two_action",
```
ataymano, line 3: cb -> ca

python/tests/test_framework/test_configs/cb_cont.json
```diff
@@ -0,0 +1,189 @@
...
+                "num_examples": 100,
+                "num_features": 1,
+                "action_range": [
+                    0,
+                    2
+                ],
+                "reward_function": {
+                    "name": "constant_reward",
+                    "params": {
+                        "reward": [
```
ataymano, line 16: let's reflect the fact that it is piecewise constant in the name and move threshold value  either to name or to arguments list

python/tests/test_framework/cb_cont/data_generation.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    logging_policy_obj = get_function_object(
+        "cb_cont.logging_policies", logging_policy["name"]
+    )
+    features = [f"feature{index}" for index in range(1, num_features + 1)]
+    with open(os.path.join(script_directory, dataFile), "w") as f:
+        for _ in range(num_examples):
+            no_context = len(context_name)
+            if no_context > 1:
+                context = random.randint(1, no_context)
```
ataymano, line 36: context = context_names[context - 1] after that? (or using np.random.choice)

PR #4632, opened 2023-08-15
fix: stash and restore prediction when calling learn during learn_returns_prediction==false calls

jackgerrits:
- Fixes Igl's usage of the prediction in learn
- Causes one less call to predict for cats which means the PRNG sequence changes and so do the predictions.
- Seems that expreplay used to be returning the final prediction that was coming out of GD in its learn call. This fixes that to be the initial predict.


test/pred-sets/ref/cats_load.predict
```diff
@@ -1,10 +1,10 @@
...
-1.8475317,0.40625
-2.1299162,0.40625
-1.7430687,0.40625
-14.327164,0.00625
-1.7164373,0.40625
-1.748255,0.40625
-30.00431,0.00625
-0.51225615,0.40625
-2.3816023,0.40625
+25.271912,0.00625
```
ataymano, line 1: numbers here are crazily different - do we need this test?

test/train-sets/ref/0001-replay.stderr
```diff
@@ -18,18 +18,18 @@ loss     last          counter         weight          label        predict feat
...
+0.441406 0.449219          512          512.0         0.0000         0.0000      278
+0.453125 0.464844         1024         1024.0         1.0000         1.0000      170
 
 finished run
 number of examples per pass = 200
 passes used = 8
 weighted example sum = 1600.000000
 weighted label sum = 728.000000
-average loss = 0.047746
+average loss = 0.455000
```
lalo, line 32: so loss went up?

jackgerrits: Previous loss was cheating since it was using predictions that were coming out of learn.

ataymano (approved):

PR #4642, opened 2023-09-04
docs: Update python_Contextual_bandits_and_Vowpal_Wabbit.ipynb

Abinash-bit:


jackgerrits:
Since it is an interactive notebook we would prefer to leave it as the value being output

PR #4643, opened 2023-09-14
[feat] Enable CSV feature by default

jackgerrits:


derfugu1:
So my understanding is the csv parsing is not enabled and when I build from scratch I get an 'unrecognized option --csv' error. IF that is the case - should this tutorial https://vowpalwabbit.org/docs/vowpal_wabbit/python/latest/tutorials/cmd_csv_with_iris_dataset.html not be updated? Either include a warning that it is not going to work, or explain how to change the build process to include it? It's quite frustrating to be honest, working with a library and a tutorial that don't work well together. I am not familiar with build processes at all, so something like this can easily cost hours.

PR #4650, opened 2023-10-09
fix: prevent feature corruption in LRU cache

mrucker:
The LRU Cache in the EMT reduction maintains references to memories within the EMT tree. These references could become corrupted when the ownership of the memory pointer was transferred to the tree leaf. This update fixes this problem by waiting until after ownership is transferred before storing the memory reference in the cache.


ataymano:
That seems like workaround but not fix.
I tried to add (it is clearly not copyable since it is has vector of unique_ptrs):
```
emt_node(const emt_node&) = delete;
emt_node& operator=(const emt_node&) = delete;
```
and it broke my build (cmake + vs2019 + win)

ataymano (approved):

PR #4651, opened 2023-10-13
refactor: [sparse_weights] get for predict

bassmang:


vowpalwabbit/core/include/vw/core/array_parameters_sparse.h
```diff
@@ -109,7 +111,7 @@ class sparse_parameters
   void share(size_t /* length */);
 #endif
 
-private:
+public:
```
olgavrou, line 114: why make these vars public? did you just want to make the function public?

bassmang: yea that was for testing, reverted it

vowpalwabbit/core/include/vw/core/array_parameters_sparse.h
```diff
@@ -20,7 +20,7 @@ class sparse_parameters;
 namespace details
 {
 
-using weight_map = std::unordered_map<uint64_t, std::shared_ptr<VW::weight>>;
+using weight_map = std::map<uint64_t, std::shared_ptr<VW::weight>>;
```
jackgerrits, line 23: What's the reasoning behind making this a map?

bassmang: its showing better latency for some tests, but its primarily from trying to predict on with features that aren't in the model, so it's unclear if this is an optimization

vowpalwabbit/core/src/array_parameters_sparse.cc
```diff
@@ -26,6 +26,21 @@ VW::weight* VW::sparse_parameters::get_or_default_and_get(size_t i) const
   return iter->second.get();
 }
 
+VW::weight* VW::sparse_parameters::get_impl(size_t i) const
+{
+  static VW::weight default_value = 0.0f;
+
+  uint64_t index = i & _weight_mask;
+  auto iter = _map.find(index);
```
rajan-chari, line 34: bloom filter might make things faster here if find() is slow

bassmang: Bloom filter seems to not be deterministic which would probably be an issue. find() should have an average O(1) lookup but its slow since its looking for elements not in the map, which will always be O(n)

vowpalwabbit/core/include/vw/core/array_parameters.h
```diff
@@ -25,6 +25,12 @@ class parameters
     else { return dense_weights[i]; }
   }
 
+  inline VW::weight& get(size_t i)
```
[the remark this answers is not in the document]
bassmang: `get` will look for a weight and return a default if not found. `[]` will do the same but will insert also insert a default entry in the weights map if not found

vowpalwabbit/core/include/vw/core/array_parameters_sparse.h
bassmang: Max Latencies
Old: 17506
New: 8642
New ordered: 4168
 
Average Latencies
Old: 234.6244564186664
New: 279.0359539029642
New ordered: 192.9708312415632

These were the results for one dataset where there were lots of features in the dataset not present in the model.

bassmang: This change has been moved to #4654

test/core.vwtest.json
```diff
@@ -6021,7 +6021,7 @@
   {
     "id": 465,
     "desc": "cb_explore_adf with epsilon-greedy exploration using --sparse_weights and saving model",
-    "vw_command": "--cb_explore_adf --epsilon 0.1 -d train-sets/cb_test.ldf --noconstant --sparse_weights -f standard_sparse_model.vw",
+    "vw_command": "--cb_explore_adf --epsilon 0.1 -d train-sets/cb_test.ldf --noconstant --sparse_weights -f standard_sparse_model.vw -q::",
```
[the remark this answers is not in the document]
bassmang: benchmarks added, tests can be added in a separate PR

rajan-chari:
The one change that makes this PR side effect free is specialization of get for random init.  In that case, insert something in the map.

rajan-chari:
Based on latency tests done so far, this is definitely desirable.

.github/workflows/vendor_build.yml
```diff
@@ -48,7 +48,6 @@ jobs:
           -DVW_INSTALL=Off
           -DWARNINGS=On
           -DWARNING_AS_ERROR=On
-          -DUSE_LATEST_STD=On
```
olgavrou, line 51: why are we removing this from here?

PR #4655, opened 2023-10-17
fix: Fixing model version check for Active reduction

peterychang:


jackgerrits (approved):

peterychang:
Would it be worth adding another check so that the save_load checks if the version is either `(9.0 <= version < 9.4) || version >= 9.10`?

jackgerrits:
Hmm yeah true that could work

PR #4657, opened 2023-10-30
fix: Gtest fix

bassmang:


olgavrou (approved):

olgavrou:
Please create an issue and info about the problem with the latest ubuntu image so that we can revert to `USE_STD_LATEST` once fixed

PR #4660, opened 2023-11-08
prototype:  pytorch base learner

rajan-chari:
**Current todo**

- [x] –passes
- [x] fix memory leak during densification
- [x] VW reporting functions
	- [x] set_output_example_prediction(output_example_prediction)
	- [x] set_print_update(print_update)
	- [x] set_cleanup_example(cleanup)
- [x] Save prediction in example
	- [x] ec.partial_prediction, contraction
	- [x] ec.pred.scalar
- [x] batch optimizer step after gradient accumulation
- [x] Change the LinearLayer size as input size grows.  Use initial, double + copy strategy to reduce compute weight
- [x] n base_learners
- [x] Smoke test - unit test
- [ ] save/load
	- [ ] model
	- [ ] feature_dict
- [ ] Test - learning (cb)
- [ ] Full coverage - unit test
- [ ] Fixed random seed for reproducible results
- [ ] Experimental
- [ ] AdamW
-------------------------------------------------------------------------

- [ ] Use example.weight in training
- [ ] Check for test_only in learn()?
- [ ] apply l1?
- [ ] num_features calc
- [ ] apply gd finalize_prediction? nan/max/min
- [ ] interactions disabled
- [ ] Triage:
	- [ ] set_multipredict(nullptr)
	- [ ] set_update(nullptr)
	- [ ] set_save_load(nullptr)
	- [ ] set_end_pass(nullptr)
	- [ ] set_merge_with_all(nullptr)
	- [ ] set_add_with_all(nullptr)
	- [ ] set_subtract_with_all(nullptr)

**Current bugs**
- [ ] Parsing -0.00000

**Post MVP Wish list**
- [ ] accumulate error during minibatch instead of gradients
- [ ] Support arbitrary model created by TorchScript/Other mechanisms
- [ ] VW binary input format
- [ ] statically link torch dlls
- [ ] GPU
- [ ] Different optimizers (AdamW) for simple N layer network
- [ ] Profile allocations and ensure no-allocations in steady state
- [ ] Reduce Binary Size

**Debugging Notes:**
When building debug configuration in windows, and using system dependencies, use the debug version of libtorch libraries.  This is because release version of your app will be ABI incompatible with debug version of libtorch and vice-versa leading to very odd errors.  This is why std cpp types should not be used in library interfaces!


lalo:
update https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/ThirdPartyNotices.txt if you are planning on merging

vowpalwabbit/core/src/reductions/dnn/dnn.cc
```diff
@@ -0,0 +1,214 @@
...
+  config::options_i& options = *stack_builder.get_options();
+  VW::workspace& all = *stack_builder.get_all_pointer();
+
+  config::option_group_definition new_options("[Reduction] DNN");
+  bool use_dnn = false;
+  int num_inputs = 20;
+  int num_layers = 3;
+  int hidden_layer_size = 20;
+  int mini_batch_size = 10;
+  new_options.add(make_option("dnn", use_dnn).keep().necessary().help("Fully connected deep neural network base learner."))
```
ataymano, line 184: should we mark them as experimental()?

rajan-chari: Yep.

rajan-chari:
> update https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/ThirdPartyNotices.txt if you are planning on merging

Got it.

rajan-chari:
> update https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/ThirdPartyNotices.txt if you are planning on merging

Makes sense.

olgavrou:
To be re-opened in the future

PR #4671, opened 2023-11-29
feat: direct interface for active.cc and variable rename for understandability

beygel:


vowpalwabbit/core/src/reductions/active.cc
```diff
@@ -110,6 +120,37 @@ void predict_or_learn_active(active& a, learner& base, VW::example& ec)
...
 
+template <bool is_learn>
+void predict_or_learn_active_direct(active& a, learner& base, VW::example& ec)
+{ 
+  if (is_learn) { base.learn(ec); }
+  else { base.predict(ec); }
+    
+  if (ec.l.simple.label == FLT_MAX)
+  {
+    std::cout << "test label" << std::endl;
```
ataymano, line 131: this print is going to happen every example. Is it intentional?

vowpalwabbit/core/src/reductions/active.cc
```diff
@@ -110,6 +120,37 @@ void predict_or_learn_active(active& a, learner& base, VW::example& ec)
...
+      const float threshold = (a._shared_data->max_label + a._shared_data->min_label) * 0.5f;
+      // We want to understand the change in prediction if the label were to be
+      // the opposite of what was predicted. 0 and 1 are used for the expected min
+      // and max labels to be coming in from the active interactor.
+      ec.l.simple.label = (ec.pred.scalar >= threshold) ? a._min_seen_label : a._max_seen_label;
+      ec.confidence = std::abs(ec.pred.scalar - threshold) / base.sensitivity(ec);
+      std::cout << "query, threshold = " << threshold << " label = " << ec.l.simple.label << " prediction = " << ec.pred.scalar << " confidence = " << ec.confidence << " wue = " << a._shared_data->weighted_unlabeled_examples << std::endl;
+      ec.l.simple.label = FLT_MAX;
+      ec.pred.scalar = query_decision(a, ec.confidence, static_cast<float>(a._shared_data->weighted_unlabeled_examples));
+      std::cout << "query decision = " << ec.pred.scalar << std::endl;
```
ataymano, line 143: Let me double check my understanding: these prints are not diagnostic logs but essential output for this mode, right?

beygel: They were diagnostic logs.  I've just removed them, triggering an update to the pull request.  Can you please take a look now?  Thanks.

ataymano (approved):

PR #4677, opened 2024-01-02
fix: Update VcPkg to fix issue when installing Zlib

lokitoth:
The current VcPkg version's port of zlib relies on a msys2 dependency to be downloaded from the web (in this case libwinpthreads), which now errors out with 404. This has not been caught by CI because we cache vcpkg artifacts to avoid a very long build-time. The fix is to upgrade vcpkg to a newer version.

Changes:
* Updates vcpkg to release tag 2023.12.12
* Changes the build_windows_cmake action to be manually-triggered only
* Add .gitignore for default folder vcpkg will install to when triggered from CLI (rather than CMake)

In addition: We should consider adding a vcpkg-only test action on a schedule (or similar) to ensure that we catch these

### DIFF
--- a/.github/workflows/build_windows_cmake.yml
+++ b/.github/workflows/build_windows_cmake.yml
@@ -1,13 +1,14 @@
 name: Windows / C++ - CMake only
 
 on:
-  push:
-    branches:
-      - master
-      - 'releases/**'
-  pull_request:
-    branches:
-      - '*'
+  # push:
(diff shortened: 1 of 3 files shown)


lokitoth:
Force-push to fix semantic-title check

lokitoth:
Closing, since this is a duplicate (ish, later vcpkg version) of #4672

PR #4682, opened 2024-01-25
build: Fix building against RapidJSON 1.1.0 as sys dep

geofft:
The last release of RapidJSON, 1.1.0 from 2016 (see Tencent/rapidjson#1006 I guess), spells the variable in all caps: https://github.com/Tencent/rapidjson/blob/v1.1.0/RapidJSONConfig.cmake.in Use both spellings to accommodate both people on the last release and people who have picked a newer git commit.


jackgerrits (approved):

lokitoth:
Thanks for the fix @geofft!

PR #4684, opened 2024-02-14
feat: Support cleaning up spare examples correctly in read_span_flatbuffer()

lokitoth:
Consumers of VW as a library can provide their own event pools, etc. Previous parsers were always able to predict when an even would be needed ahead of time, so would only allocate when necessary. This was done by relying on a single incoming event preallocation to let the external host deallocate in the case of nothing to be parsed.

This does not work for the FB parser due to how it handles re-entrancy, and we do not want to spend the time re-architecting it to avoid this. The fix, in this case, is to expand the API to include a callback to return spare events back to the host's event pool.

### DIFF
--- a/vowpalwabbit/fb_parser/include/vw/fb_parser/parse_example_flatbuffer.h
+++ b/vowpalwabbit/fb_parser/include/vw/fb_parser/parse_example_flatbuffer.h
@@ -15,7 +15,7 @@ namespace VW
 
 namespace experimental
 {
-  class api_status;
+class api_status;
 }
 
 using example_sink_f = std::function<void(VW::multi_ex&& spare_examples)>;
@@ -26,9 +26,8 @@ namespace flatbuffer
 {
 int flatbuffer_to_examples(VW::workspace* all, io_buf& buf, VW::multi_ex& examples);
 
-
-int read_span_flatbuffer(
-    VW::workspace* all, const uint8_t* span, size_t length, example_factory_t example_factory, VW::multi_ex& examples, example_sink_f example_sink = nullptr, VW::experimental::api_status* status = nullptr);
+int read_span_flatbuffer(VW::workspace* all, const uint8_t* span, size_t length, example_factory_t example_factory,
+    VW::multi_ex& examples, example_sink_f example_sink = nullptr, VW::experimental::api_status* status = nullptr);
 
 class parser
 {
@@ -62,6 +61,19 @@ private:
   int parse_flat_label(shared_data* sd, example* ae, const Example* eg, VW::io::logger& logger,
       VW::experimental::api_status* status = nullptr);
   int get_namespace_index(const Namespace* ns, namespace_index& ni, VW::experimental::api_status* status = nullptr);
+
+  inline void reset_active_multi_ex()
+  {
+    _multi_ex_index = 0;
+    _active_multi_ex = false;
+    _multi_example_object = nullptr;
+  }
+
+  inline void reset_active_collection()
+  {
+    _example_index = 0;
+    _active_collection = false;
+  }
 
   void parse_simple_label(shared_data* sd, polylabel* l, reduction_features* red_features, const SimpleLabel* label);
   void parse_cb_label(polylabel* l, const CBLabel* label);
--- a/vowpalwabbit/fb_parser/include/vw/fb_parser/parse_example_flatbuffer.h
+++ b/vowpalwabbit/fb_parser/include/vw/fb_parser/parse_example_flatbuffer.h
@@ -16,13 +16,17 @@ namespace VW
 
 class api_status;
 
+using example_sink_f = std::function<void(VW::multi_ex&& spare_examples)>;
+
 namespace parsers
 {
 namespace flatbuffer
 {
 int flatbuffer_to_examples(VW::workspace* all, io_buf& buf, VW::multi_ex& examples);
+
+
 bool read_span_flatbuffer(
-    VW::workspace* all, const uint8_t* span, size_t length, example_factory_t example_factory, VW::multi_ex& examples);
+    VW::workspace* all, const uint8_t* span, size_t length, example_factory_t example_factory, VW::multi_ex& examples, example_sink_f example_sink = nullptr);
 
 class parser
 {
--- a/vowpalwabbit/fb_parser/include/vw/fb_parser/parse_example_flatbuffer.h
+++ b/vowpalwabbit/fb_parser/include/vw/fb_parser/parse_example_flatbuffer.h
@@ -4,7 +4,6 @@
 
 #pragma once
 
-#include "vw/core/api_status.h"
 #include "vw/core/example.h"
 #include "vw/core/multi_ex.h"
 #include "vw/core/shared_data.h"
@@ -14,7 +13,10 @@
 namespace VW
 {
 
-class api_status;
+namespace experimental
+{
+  class api_status;
+}
 
 using example_sink_f = std::function<void(VW::multi_ex&& spare_examples)>;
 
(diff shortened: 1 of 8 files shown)


vowpalwabbit/fb_parser/src/parse_example_flatbuffer.cc
```diff
@@ -59,16 +64,15 @@ bool read_span_flatbuffer(
...
   if (address % 8 != 0)
   {
     std::stringstream sstream;
     sstream << "fb_parser error: flatbuffer data not aligned to 8 bytes" << std::endl;
     sstream << "   span => @" << std::hex << address << std::dec << " % " << 8 << " = " << address % 8
             << " (vs desired = " << 0 << ")";
-    THROW(sstream.str());
-    return false;
+
+    RETURN_ERROR_LS(status, fb_parser_span_misaligned) << sstream.str();
```
rajan-chari, line 75: Nice.

vowpalwabbit/fb_parser/include/vw/fb_parser/parse_example_flatbuffer.h
```diff
@@ -14,15 +13,22 @@
...
 {
 namespace flatbuffer
 {
 int flatbuffer_to_examples(VW::workspace* all, io_buf& buf, VW::multi_ex& examples);
-bool read_span_flatbuffer(
-    VW::workspace* all, const uint8_t* span, size_t length, example_factory_t example_factory, VW::multi_ex& examples);
+
+
+int read_span_flatbuffer(
+    VW::workspace* all, const uint8_t* span, size_t length, example_factory_t example_factory, VW::multi_ex& examples, example_sink_f example_sink = nullptr, VW::experimental::api_status* status = nullptr);
```
rajan-chari, line 31: Nice way to avoid allocs.

vowpalwabbit/fb_parser/src/parse_example_flatbuffer.cc
```diff
@@ -79,17 +83,32 @@ bool read_span_flatbuffer(
...
+
+    RETURN_ERROR_LS(status, fb_parser_span_length_mismatch) << sstream.str();
   }
 
   VW::multi_ex temp_ex;
-  temp_ex.push_back(&example_factory());
+  auto scope_guard = VW::scope_exit([&temp_ex, &all, &example_sink]()
+  {
+    if (example_sink == nullptr) { VW::finish_example(*all, temp_ex); }
+    else { example_sink(std::move(temp_ex)); }
```
rajan-chari, line 94: sweet.

vowpalwabbit/fb_parser/src/parse_example_flatbuffer.cc
```diff
@@ -43,9 +45,12 @@ int flatbuffer_to_examples(VW::workspace* all, io_buf& buf, VW::multi_ex& exampl
   return static_cast<int>(status.get_error_code() == VW::experimental::error_code::success);
 }
 
-bool read_span_flatbuffer(
-    VW::workspace* all, const uint8_t* span, size_t length, example_factory_t example_factory, VW::multi_ex& examples)
+int read_span_flatbuffer(VW::workspace* all, const uint8_t* span, size_t length, example_factory_t example_factory,
+    VW::multi_ex& examples, example_sink_f example_sink, VW::experimental::api_status* status)
 {
+  int a = 0;
```
rajan-chari, line 51: Not sure what this is for.

vowpalwabbit/fb_parser/src/parse_example_flatbuffer.cc
```diff
@@ -79,17 +83,32 @@ bool read_span_flatbuffer(
...
+    if (example_sink == nullptr) { VW::finish_example(*all, temp_ex); }
+    else { example_sink(std::move(temp_ex)); }
+  });
+
+  // There is a bit of unhappiness with the interface of the read_XYZ_<format>() functions, because they often
+  // expect the input multi_ex to have a single "empty" example there. This contributes, in part, to the large
+  // proliferation of entry points into the JSON parser(s). We want to avoid exposing that insofar as possible,
+  // so we will check whether we already received a perfectly good example and use that, or create a new one if
+  // needed.
+  if (examples.size() > 0)
```
rajan-chari, line 102: This is a bit tricky but does reuse memory.

vowpalwabbit/fb_parser/src/parse_example_flatbuffer.cc
```diff
@@ -98,10 +117,7 @@ bool read_span_flatbuffer(
...
         break;
       default:
-        std::stringstream sstream;
-        sstream << "Error parsing examples: " << std::endl;
-        THROW(sstream.str());
-        return false;
+        RETURN_IF_FAIL(result);
     }
 
     has_more &= !temp_ex[0]->is_newline;
```
rajan-chari, line 123: A comment here would be useful for us mere mortals :)

vowpalwabbit/fb_parser/src/parse_example_flatbuffer.cc
```diff
@@ -79,17 +83,32 @@ bool read_span_flatbuffer(
...
+  // expect the input multi_ex to have a single "empty" example there. This contributes, in part, to the large
+  // proliferation of entry points into the JSON parser(s). We want to avoid exposing that insofar as possible,
+  // so we will check whether we already received a perfectly good example and use that, or create a new one if
+  // needed.
+  if (examples.size() > 0)
+  {
+    temp_ex.push_back(examples[0]);
+    examples.pop_back();
+  }
+  else { temp_ex.push_back(&example_factory()); }
```
rajan-chari, line 107: It looks like the first element from examples is being added to temp_ex but last element is removed from examples.  Is this intentional?

vowpalwabbit/fb_parser/src/parse_label.cc
```diff
@@ -2,6 +2,7 @@
 // individual contributors. All rights reserved. Released under a BSD (revised)
 // license as described in the file LICENSE.
 
+#include "vw/core/api_status.h"
```
rajan-chari, line 5: does not seem to be used in this file.

rajan-chari: never mind

vowpalwabbit/fb_parser/src/parse_example_flatbuffer.cc
lokitoth: I needed a code change to force a re-build. Forgot to remove it :)

vowpalwabbit/fb_parser/src/parse_example_flatbuffer.cc
lokitoth: I added some comments throughout

vowpalwabbit/fb_parser/src/parse_example_flatbuffer.cc
lokitoth: The expectation here is that elements[0] is also the last element. May be worth to add an assert here.

vowpalwabbit/fb_parser/src/parse_label.cc
lokitoth: It gets used for RETURN_ERROR at line 153/154:
 https://github.com/VowpalWabbit/vowpal_wabbit/pull/4684/files/1895dcca3e9ccd629e3972a024d46d6ce49e141f#diff-0a4181992ca5c75481b37b2f08796914a150ba67c47f01701f66bff87447ec48L153-R154

vowpalwabbit/fb_parser/src/parse_example_flatbuffer.cc
```diff
@@ -79,17 +80,33 @@ bool read_span_flatbuffer(
...
+        if (example_sink == nullptr) { VW::finish_example(*all, temp_ex); }
+        else { example_sink(std::move(temp_ex)); }
+      });
+
+  // There is a bit of unhappiness with the interface of the read_XYZ_<format>() functions, because they often
+  // expect the input multi_ex to have a single "empty" example there. This contributes, in part, to the large
+  // proliferation of entry points into the JSON parser(s). We want to avoid exposing that insofar as possible,
+  // so we will check whether we already received a perfectly good example and use that, or create a new one if
+  // needed.
+  if (examples.size() > 0)
```
rajan-chari, line 100: assert size == 1?

rajan-chari (approved):
Looks good.  Thanks Jacob.

PR #4695, opened 2024-06-06
fix: string-view-lite: Include `<string>` before using `std::basic_string`

StephanTLavavej:
I work on MSVC's STL, and we regularly build popular open-source projects, including yours, with development builds of the MSVC toolset. This allows us to find and fix toolset regressions before they affect users, and also allows us to provide advance notice of breaking changes, which is the case here.

We recently merged https://github.com/microsoft/STL/pull/4633 which will ship in VS 2022 17.11 Preview 3. This improved build throughput by refactoring `<string_view>` so that it no longer drags in `std::basic_string`. Your third-party dependency string-view-lite was affected by this source-breaking change, as it wasn't properly including `<string>`. When built with our updated STL, this will emit a compiler error:

```
C:/gitP/VowpalWabbit/vowpal_wabbit/ext_libs/string-view-lite\nonstd/string_view.hpp(134): error C2039: 'basic_string': is not a member of 'std'
```

The fix is simple and portable: simply include the necessary header, immediately before it's needed. I've submitted this upstream as https://github.com/martinmoene/string-view-lite/pull/58 .


jackgerrits:
Thanks for the fix @StephanTLavavej, and for all the work you do on the STL!

jackgerrits (approved):

PR #4702, opened 2024-10-13
suggested fixes

JohnLangford:
From here: https://github.com/VowpalWabbit/vowpal_wabbit/issues/4700 .


JohnLangford:
This is redundant after a general CI clean up.

PR #4708, opened 2025-01-14
Moving to SPDX ID for license id

hyandell:
With license ids becoming more standardized, I wanted to suggest using "BSD-3-Clause" here rather than the obscure "Vowpal Wabbit License" term. It also makes the id descriptive as "Vowpal Wabbit License" would apply no matter what you put in the license text.

### DIFF
--- a/java/pom.xml.in
+++ b/java/pom.xml.in
@@ -32,7 +32,7 @@
 
     <licenses>
         <license>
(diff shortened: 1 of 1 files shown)


JohnLangford:
Added into #4724 --- Thanks :-)

PR #4720, opened 2025-10-04
build: support building with eigen 5.0.0

cho-m:
Existing find_package won't detect Eigen3 5.0.0. Upstream added support for range to help with this https://gitlab.com/libeigen/eigen/-/merge_requests/1989

However, the support is only available in 3.4.1+ so version 3.4.0 needs to use old approach.

Also, EIGEN3_INCLUDE_DIR was removed so instead just use Eigen3::Eigen target.

Eigen 3.4.0 has:
```cmake
# Create imported target Eigen3::Eigen
add_library(Eigen3::Eigen INTERFACE IMPORTED)

set_target_properties(Eigen3::Eigen PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/eigen3"
)
```

Eigen 5.0.0 has:
```cmake
# Create imported target Eigen3::Eigen
add_library(Eigen3::Eigen INTERFACE IMPORTED)

set_target_properties(Eigen3::Eigen PROPERTIES
  INTERFACE_COMPILE_FEATURES "cxx_std_14"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/eigen3"
)
```


JohnLangford:
I've created #4728 which:

1. ✅ Rebases your changes on current master (includes recent ARM64 fixes)
2. ✅ Adds CMake version guard for CMake < 3.19 compatibility
3. ✅ Removes the `eigen3: 3.4.0` override from vcpkg.json

The third change is key: vcpkg baseline has **Eigen 5.0.1**, so removing the override means CI will actually test Eigen 5.x support (previously it was locked to 3.4.0, so the PR had no practical effect in CI).

This allows us to validate that the changes work with Eigen 5.x in all CI environments. If CI passes, we'll have confidence that the Eigen 5.x support is solid.

Credit to you for the original implementation - I've preserved your authorship in the commit. #4728 builds directly on your work with the improvements from my critique.

JohnLangford:
Merged via #4728 which includes:
- ✅ Rebase on current master
- ✅ CMake version guard for CMake < 3.19
- ✅ Removed Eigen 3.4.0 vcpkg override (tested with Eigen 5.0.1)
- ✅ All CI passed

Thank you for the original implementation! The changes are now in master.

PR #4734, opened 2026-01-14
fix: use raw strings for regex patterns to avoid invalid escape sequences

JohnLangford:
Fixes invalid escape sequence warnings in Python test script.

## Changes
- Convert regex pattern strings to raw strings (r"...") at lines 48 and 133
- Prevents SyntaxWarning: invalid escape sequence '\-'

## Testing
- Verified pattern matches still work correctly
- Python 3.12+ will no longer emit syntax warnings

Fixes issue #5 from CI warnings summary.


JohnLangford:
Closing in favor of combined PR #4738

PR #4755, opened 2026-01-16
fix: avoid C4353 warning by checking audit_func for null before calling

JohnLangford:
## Summary

Fixes 2 C4353 warnings in Windows 2019 builds by using MSVC pragma to suppress the warning.

## Problem

MSVC generates C4353 warning when constant 0 (nullptr) is used as a function expression:
```
warning C4353: nonstandard extension used: constant 0 as function expression.
Use '__noop' function intrinsic instead
```

The `depth_audit_func` lambda in `interactions_predict.h` calls `audit_func(dat, audit_str)`, where `audit_func` is a template parameter function pointer. When instantiated with nullptr, MSVC treats the call as `0(...)`, triggering this nonstandard extension warning.

## Solution

Added MSVC-specific pragma to disable C4353 warning around the lambda definition. This is safe because:
- The lambda is only called when `audit` template parameter is true
- When `audit` is false, `audit_func` is set to `dummy_func`, not nullptr
- The warning is a false positive from MSVC's template instantiation analysis

## Alternative Approaches Considered

1. **Runtime nullptr check**: Would trigger GCC warning `-Werror=address` because the function pointer is known at compile time
2. **`if constexpr`**: Requires C++17, but VW uses C++11 by default
3. **Template specialization**: Overly complex for this simple case

## Test plan

- [x] CI passes on Windows 2019/2022 builds without C4353 warnings
- [x] GCC/Clang builds unaffected (pragma is MSVC-only)
- [x] No functional changes - same behavior, just suppressed warning

### DIFF
--- a/vowpalwabbit/core/include/vw/core/interactions_predict.h
+++ b/vowpalwabbit/core/include/vw/core/interactions_predict.h
@@ -424,7 +424,17 @@ inline void generate_interactions(const std::vector<std::vector<VW::namespace_in
         dat, begin, end, ec.ft_offset, weights, value, index);
   };
 
+  // MSVC warns about using constant 0 as function expression (C4353) when audit_func is nullptr
+  // This is a nonstandard extension but is safe here since audit_func is always a valid function
+  // pointer (either an actual audit function or dummy_func) when this lambda is called
+#ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


JohnLangford:
Closing in favor of combined PR #4758

Issue #4729, opened 2026-01-14
Regenerate test references with CSV and FlatBuffer features enabled


JohnLangford:
## Problem

Currently, tests 275 and 276 (--help output tests) are skipped in the vcpkg build workflow because:
1. The vcpkg build includes CSV and FlatBuffer features (VW_FEAT_CSV, VW_FEAT_FLATBUFFERS)
2. The test reference files were generated without these features
3. This causes --help output to differ from the expected references

See: https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/.github/workflows/vcpkg_build.yml#L78-L80

## Proposed Solution

Regenerate test reference files with CSV and FlatBuffer features enabled. This would:
- Allow tests 275 and 276 to run in the vcpkg workflow
- Provide better test coverage of the Windows artifacts
- Better reflect the actual artifact configuration that users receive

## Implementation

1. Rebuild VW with `-DVW_FEAT_CSV=On -DVW_FEAT_FLATBUFFERS=On`
2. Regenerate affected test references (primarily tests 275, 276)
3. Remove `--skip_test 275 276` from vcpkg_build.yml
4. Consider whether other artifacts should also enable these features by default

## Context

These features are already enabled in:
- vcpkg builds (CMakePresets.json default preset)
- valgrind workflow
- Some CI workflows

But not in:
- Default CMake configuration
- Test reference generation

JohnLangford:
This issue was already addressed by commit ea35ca2d8 which:
1. Enabled VW_FEAT_CSV in vcpkg-debug and vcpkg-release presets
2. Removed the --skip_test 275 276 from vcpkg_build.yml

Tests 275 and 276 should now run in the vcpkg workflow.

PR #4770, opened 2026-01-19
fix: update comment for skipped --help tests in vcpkg workflow

JohnLangford:
## Summary
- Update the comment explaining why tests 275 and 276 are skipped in vcpkg builds

## Details
After investigation, the original comment was incorrect. The actual situation is:
- Test reference files (`help.stdout`) **include** CSV parser options
- The vcpkg build does **NOT** include CSV features
- Therefore, the vcpkg `--help` output is missing CSV options compared to the reference

The tests must remain skipped until either:
1. CSV features are enabled in the vcpkg build, or
2. Separate reference files are created for vcpkg tests without CSV options

Updates #4729 (clarifies the actual issue)

## Test plan
- [x] vcpkg workflow passes with tests 275/276 skipped


JohnLangford:
Superseded by #4773 (combined PR)

PR #4771, opened 2026-01-19
fix: use weak reference in py_log_wrapper to prevent memory leak

JohnLangford:
## Summary
- Uses Python weak reference (`weakref.ref`) in C++ `py_log_wrapper` instead of strong reference
- Breaks reference cycle: `Workspace -> py_log_wrapper -> _log_fwd`
- Allows `_log_fwd` to be safely garbage collected after `finish()`
- Saves log messages before clearing so `get_log()` works after `finish()`

### The Problem
Previously, `py_log_wrapper` held a strong `py::object` reference to the Python `_log_fwd` object. This created a reference cycle that prevented proper garbage collection:
- `Workspace` holds `_log_wrapper`
- `_log_wrapper` holds strong reference to `_log_fwd`
- Both are owned by `Workspace` but the cross-reference prevents GC

### The Solution
The `py_log_wrapper` now uses `weakref.ref` to hold a weak reference to the Python log object:

```cpp
py_log_wrapper(py::object py_log) {
    py::object weakref_module = py::import("weakref");
    weak_ref = weakref_module.attr("ref")(py_log);
}
```

In callbacks, we check if the weak reference is still valid:
```cpp
py::object py_log = inst->get_log_object();
if (py_log.is_none()) { return; }  // Object was collected, skip logging
py_log.attr("log")(message);
```

### Benefits
1. **No memory leak**: `_log_fwd` can be GC'd when cleared in `finish()`
2. **No segfaults**: Callbacks gracefully skip if object is collected
3. **Backward compatible**: Existing code works unchanged

### Notes
- `_log_wrapper` still cannot be cleared because C++ holds raw pointers to it
- However, `_log_wrapper` is small (~100 bytes) vs `_log_fwd` (can be MBs of log messages)
- The important memory (accumulated log messages) is now properly freed

## Test plan
- [x] Existing Python tests pass
- [x] No segfaults in multiprocessing tests (cibuildwheel e2e_v2)
- [x] `get_log()` returns correct output after `finish()`
- [x] Memory is properly freed after `finish()`

### DIFF
--- a/python/pylibvw.cc
+++ b/python/pylibvw.cc
@@ -242,15 +242,32 @@ public:
 class py_log_wrapper
 {
 public:
-  py::object py_log;
-  py_log_wrapper(py::object py_log) : py_log(py_log) {}
+  py::object weak_ref;  // Weak reference to Python log object (prevents reference cycles)
+
+  py_log_wrapper(py::object py_log)
+  {
+    // Create a weak reference to avoid preventing garbage collection
+    // This breaks the reference cycle: Workspace -> py_log_wrapper -> _log_fwd
+    py::object weakref_module = py::import("weakref");
+    weak_ref = weakref_module.attr("ref")(py_log);
+  }
+
+  // Get the Python log object, or None if it has been garbage collected
+  py::object get_log_object() const
+  {
+    // Call the weakref to get the actual object (returns None if collected)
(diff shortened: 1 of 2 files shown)


JohnLangford:

PR #4776, opened 2026-01-19
fix: add .keep() to --eval option to persist in model file

### LINKED ISSUE #4175: --eval corrupting model state
### Describe the bug

`--eval` flag has counterintuitive effects on the model

### How to reproduce

This crashes:
vw -d train-sets/rcv1_cb_eval --cb 2 --eval -f model
vw -i model

This doesn't crash:
vw -d train-sets/rcv1_cb_eval --cb 2 --eval -f model
vw -i model --eval

This doesn't crash:
vw -d train-sets/rcv1_cb_eval --cb 2 --eval -f model --predict_only_model
vw -i model

`--eval` needs to be specified when load if `--predict_only_model` is not

### Version

9.4

### OS

Linux

### Language

CLI

### Additional context

_No response_

JohnLangford:
## Summary
- Add `.keep()` to the `--eval` option in cb_algs.cc so it gets serialized to the model file

## Problem
Models saved with `--eval` would crash on reload:
```bash
vw -d train.txt --cb 2 --eval -f model    # Save with --eval
vw -i model                                # CRASH!
```

Workarounds that worked:
- `vw -i model --eval` (re-specify flag)
- `vw ... --predict_only_model -f model` (save without resume state)

## Root Cause
The `--eval` option was missing `.keep()`, so it wasn't serialized to the model file. On reload, the CB reduction didn't know the model was trained in evaluation mode, causing state mismatch and crash.

## Fix
Add `.keep()` to match other CB options (`--cb_type`, `--cb_force_legacy`).

Fixes #4175

## Test plan
- [ ] Save model with `--eval`, reload without specifying `--eval`
- [ ] Verify model works correctly after reload

### DIFF
--- a/test/train-sets/ref/help.stdout
+++ b/test/train-sets/ref/help.stdout
@@ -453,7 +453,7 @@ Weight Options:
                                             necessary)
     --cb_type arg                           Contextual bandit method to use (type: str, default: dr, choices
                                             {dm, dr, ips, mtr, sm}, keep)
-    --eval                                  Evaluate a policy rather than optimizing (type: bool)
+    --eval                                  Evaluate a policy rather than optimizing (type: bool, keep)
     --cb_force_legacy                       Default to non-adf cb implementation (cb_to_cb_adf) (type: bool,
                                             keep)
 [Reduction] Contextual Bandit with Action Dependent Features Options:


JohnLangford:

PR #4777, opened 2026-01-19
fix: support unsigned integers in sklearn tovw function

### LINKED ISSUE #4609: Sklearn adapter function `tovw` does not support unsigned integers in features
## Mitigation

A user should use signed integer types and not unsigned integer types when passing to the `sklearn` adapter functions.

## Details

The `tovw` function uses [`dump_svmlight_file`](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.dump_svmlight_file.html) to convert to a format that can easily construct VW text examples.

This function does not support input of unsigned integers, it requires signed due to the `pyx` code internally in sklearn.

Fails:
```
from vowpalwabbit.sklearn import VWClassifier
import numpy as np
import pandas as pd

X = pd.DataFrame({'a': [1]}, dtype='uint32')
y = pd.Series(np.zeros(1))

VWClassifier().fit(X, y)
```

Succeeds:
```
from vowpalwabbit.sklearn import VWClassifier
import numpy as np
import pandas as pd

X = pd.DataFrame({'a': [1]}, dtype='int32') # <-----
y = pd.Series(np.zeros(1))

VWClassifier().fit(X, y)
```

To fix this one way is to avoid using the `dump_svmlight_file` function. It is used currently as a way to easily convert the dataframe to vw text format.

JohnLangford:
## Summary
- Convert unsigned integer arrays to signed types automatically instead of raising an error

## Problem
The `tovw` function raised a `TypeError` when passed unsigned integer arrays because sklearn's `dump_svmlight_file` doesn't support them.

## Solution
Automatically convert unsigned types to signed types with sufficient range:
- `uint8` → `int16`
- `uint16` → `int32`  
- `uint32` → `int64`
- `uint64` → `int64` (may overflow for very large values, but rare in practice)

Works with both dense numpy arrays and sparse scipy matrices.

Fixes #4609

## Test plan
```python
from vowpalwabbit.sklearn import VWRegressor
import numpy as np
import pandas as pd

X = pd.DataFrame({'a': [1]}, dtype='uint32')
y = pd.Series(np.zeros(1))

VWRegressor().fit(X, y)  # Now works!
```

### DIFF
--- a/python/tests/test_sklearn_vw.py
+++ b/python/tests/test_sklearn_vw.py
@@ -44,12 +44,16 @@ def test_tovw():
     assert tovw(x=csr_matrix(x), y=y, sample_weight=w, convert_labels=True) == expected
 
 
-def test_tovw_raises_for_uint():
-    X = pd.DataFrame({"a": [1]}, dtype="uint32")
-    y = pd.Series(np.zeros(1))
+def test_tovw_handles_uint():
+    """Test that unsigned integers are properly converted to signed integers."""
+    X = pd.DataFrame({"a": [1, 2, 3]}, dtype="uint32")
+    y = pd.Series([0.0, 1.0, 0.5])
 
-    with pytest.raises(TypeError):
-        VWRegressor().fit(X, y)
+    # Should not raise - unsigned ints are converted to signed ints
+    model = VWRegressor()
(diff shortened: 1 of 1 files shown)


JohnLangford:

PR #4782, opened 2026-01-21
fix: disable dropout during inference in --nn reduction

### LINKED ISSUE #4067: inconsistent results when using vw daemon
### Describe the bug

I'm trying to get infer on a vw model trained with cats_pdf using a daemon. the problem is that I'm not getting the same result for a specific context : you can see below that out of 100 trials I got 2 different results

```
1-28.591745:0.006451613,28.591745-29.11138:1.5459961,29.11138-32:0.006451613
  
1-29.07612:0.006451613,29.07612-29.595755:1.5459961,29.595755-32:0.006451613
```

### How to reproduce

1 - train a vw model  :

`vw --cats_pdf 64 --min_value 1 --max_value 32 -b 20 --chain_hash --nn 120 --dropout --progress 10000 --learning_rate 0.00003 --data dummy_data.vw --final_regressor model_new_params.pkl --bandwidth 0.25981749230238499`

2- create a daemon :

`vw --daemon --initial_regressor model_new_params.pkl  --epsilon 0.2 --chain_hash --quiet --testonly --port 8888 --num_children 10
`
3- communicate with the daemon : `python ping_daemon.py`

### Version

9.1.0 (git commit: bd4998c9e)

### OS

Linux

### Language

CLI

### Additional context

_No response_

JohnLangford:
## Summary
- Disable dropout during inference (prediction/--testonly mode) in the `--nn` reduction
- During inference, use mean field approach (all neurons active) for deterministic predictions
- Follows the same pattern as `--lrq` dropout handling

## Problem
When using `--nn` with `--dropout`, predictions were non-deterministic even in `--testonly` mode because dropout was still randomly dropping neurons during inference. This caused issues like inconsistent daemon predictions (#4067).

## Solution
Only apply dropout during learning (`is_learn == true`). During inference, all neurons are active (mean field approximation), which is the standard practice for neural network dropout.

## Changes
- `nn.cc`: Added `do_dropout` flag that is only true when both `n.dropout` and `is_learn` are true
- Updated `dropscale`, dropout application, and CONVERSE block to use `do_dropout`

## Test plan
- [x] Added `nn_test.cc` with tests verifying:
  - Predictions are deterministic during inference with dropout models
  - Dropout still works during training

Fixes #4067

### DIFF
--- a/vowpalwabbit/core/CMakeLists.txt
+++ b/vowpalwabbit/core/CMakeLists.txt
@@ -503,6 +503,7 @@ set(vw_core_test_sources
       tests/minimal_custom_reduction.cc
       tests/model_util_test.cc
       tests/multiclass_label_parser_test.cc
+      tests/nn_test.cc
       tests/numeric_cast_test.cc
       tests/object_pool_test.cc
       tests/offset_tree_test.cc
--- a/vowpalwabbit/core/src/reductions/nn.cc
+++ b/vowpalwabbit/core/src/reductions/nn.cc
@@ -166,7 +166,9 @@ void predict_or_learn_multi(nn& n, learner& base, VW::example& ec)
     auto save_set_minmax = n.all->set_minmax;
     float save_min_label;
     float save_max_label;
-    float dropscale = n.dropout ? 2.0f : 1.0f;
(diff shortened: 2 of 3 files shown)


JohnLangford:
Merged into combined PR #4794

PR #4783, opened 2026-01-21
fix: make learn a no-op for unlabeled examples in cb_explore_adf

### LINKED ISSUE #4274: VW calls predict twice during learn for cb_explore_adf
### Describe the bug

If you call vw with `--cb_explore_adf` and an example that does not have a label then predict will be called twice, this seems sub-optimal and un-intuitive for reductions above cb_explore_adf and can cause bugs

e.g. when called with LAS, learn paths would not get full predictions if there is a label missing and `-p <file>` is specified (see #4273)

If a label is missing during learn what should the reduction contract be? Should learn be called anyway? Should prediction be called? Should we warn? Should we throw?

### How to reproduce

input modified from `cb_test.ldf` file:

```
shared | s_1 s_2
 | a_1 b_1 c_1
 | a_2 b_2 c_2
 | a_3 b_3 c_3
```

run with `--cb_explore_adf` and printing/breakpoint to see that predict gets called twice from `cb_explore_adf_common.h`

[code line](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/core/include/vw/core/reductions/cb/cb_explore_adf_common.h#L151)

### Version

9.5.0

### OS

Linux

### Language

CLI

### Additional context

_No response_

JohnLangford:
## Summary
- Make `learn()` a no-op for unlabeled examples in `cb_explore_adf`
- Remove the implicit `predict()` call that happened when `learn()` was called without a label

## Problem
When `learn()` was called with unlabeled examples in `cb_explore_adf`, it would internally call `predict()`. This caused:
1. Predict being called twice in some configurations (once by caller, once implicitly)
2. Confusing behavior for reductions above `cb_explore_adf`
3. Potential bugs (see #4273)

## Solution
Following the suggestion from @olgavrou: make `learn()` a no-op when there's no label. Callers should explicitly call `predict()` if they want predictions for unlabeled examples.

The `metric_predict_in_learn` counter is still incremented to track when this situation occurs, which can help identify code that may need to be updated.

## Breaking Change Note
This is a behavior change. Code that relied on getting predictions from `learn()` for unlabeled examples will need to explicitly call `predict()` instead.

Fixes #4274


JohnLangford:
Closing without merge - the silent no-op behavior for unlabeled examples during learning mode may cause issues for users who rely on getting predictions for those examples.

PR #4786, opened 2026-01-21
fix: downgrade duplicate interactions message from warning to info

### LINKED ISSUE #4496: Consider downgrading `-q ::` warning to info
When passing `-q ::` to VW you get a warning that isn't really a warning but more informational. This is such a common scenario, we should reduce the noise and just make this an info level log. Besides, I am not sure if there are many/any real use cases for `--leave_duplicate_interactions` any way.

Would there even be a duplicate created with `::` since we generate it on the fly based on input?

```sh
$ vw -q::
creating quadratic features for pairs: ::
[warning] Any duplicate namespace interactions will be removed
You can use --leave_duplicate_interactions to disable this behaviour.
using no cache
Reading datafile = none
num sources = 0
Num weight bits = 1
...
```

JohnLangford:
## Summary
- Downgrade the "Any duplicate namespace interactions will be removed" message from warning to info level
- Update 24 test reference files to expect `[info]` instead of `[warning]`

## Details
When using wildcard interactions like `-q ::`, VW outputs:
```
[warning] Any duplicate namespace interactions will be removed
You can use --leave_duplicate_interactions to disable this behaviour.
```

This message is informational rather than a warning - it describes expected and correct behavior. The warning level creates unnecessary noise for a very common use case.

## Test plan
- [x] Update test reference files
- [ ] CI tests pass

Fixes #4496

### DIFF
--- a/test/train-sets/ref/cats_load.stdout
+++ b/test/train-sets/ref/cats_load.stdout
@@ -1,3 +1,3 @@
 [warning] model file has set of {-q, --cubic, --interactions} settings stored, but they'll be OVERRIDDEN by set of {-q, --cubic, --interactions} settings from command line.
-[warning] Any duplicate namespace interactions will be removed
+[info] Any duplicate namespace interactions will be removed
 You can use --leave_duplicate_interactions to disable this behaviour.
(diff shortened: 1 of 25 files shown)


JohnLangford:
Merged into combined PR #4794

PR #4792, opened 2026-01-22
chore: remove ARM diagnostics and relax architecture flags

JohnLangford:
## Summary
Cleanup after ARM64 testing is confirmed working with pybind11.

### Changes
- **Remove** `arm_test_diagnostic.yml` workflow - redundant with cibuildwheel ARM64 job
- **Remove** `before-test` diagnostic echo commands from `pyproject.toml`
- **Relax** ARM64 compiler flags:
  - `armv8-a` → `armv8-a+crc` (CRC32 extension is widely available on ARM64 servers)
  - Remove `-mtune=generic` (use compiler default)
  - Keep `-mno-outline-atomics` (still needed to prevent LSE atomic calls)

### Kept
- `scripts/arm_diagnostic.sh` - useful for local debugging if issues recur

### Test plan
- [ ] ARM64 cibuildwheel job passes with relaxed flags


JohnLangford:
Merged into combined PR #4794

PR #4793, opened 2026-01-22
feat: use vcpkg native GitHub Actions binary cache

JohnLangford:
## Summary
- Switch from filesystem-based vcpkg binary caching to the native GitHub Actions cache backend (`x-gha`)
- Reduces transient network failures when downloading vcpkg packages (like the boost-cmake failure seen in PR #4783)

## Benefits
- More reliable caching through GitHub's internal infrastructure
- Granular per-package caching with content-addressed storage
- Better cache hit rates since packages are cached independently
- Reduced network failures from external downloads when packages are already cached

## Changes
- Updated `VCPKG_BINARY_SOURCES` to use `clear;x-gha,readwrite` instead of filesystem-based caching
- Added `actions/github-script` step to export required environment variables (`ACTIONS_CACHE_URL`, `ACTIONS_RUNTIME_TOKEN`)
- Removed local `vcpkg_binary_cache` directory creation since it's no longer needed

## Updated workflows
- `vcpkg_build.yml`
- `asan.yml`
- `run_benchmarks.yml`

## Test Plan
- [ ] Verify vcpkg builds succeed with the new caching mechanism
- [ ] Verify cache is being populated and restored correctly

### DIFF
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -14,7 +14,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  VCPKG_DEFAULT_BINARY_CACHE: ${{github.workspace}}/vcpkg_binary_cache
+  VCPKG_BINARY_SOURCES: 'clear;x-gha,readwrite'
 
 jobs:
   test_with_sanitizers:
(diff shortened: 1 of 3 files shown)


JohnLangford:
Merged into combined PR #4794

PR #4795, opened 2026-01-22
fix: create parent directories when writing files

### LINKED ISSUE #2686: VW should create parent dirs if needed when writing files
If any parents dirs are missing VW will fail. It should make all parent dirs if need be.

```
$ vw --no_stdin -f dir/model.vw --quiet

vw (io_adapter.cc:262): can't open: dir/model.vw.writing, errno = No such file or directory
```

JohnLangford:
## Summary
- When VW attempts to write a file (model, predictions, etc.) to a path where parent directories don't exist, it now automatically creates them instead of failing with "No such file or directory"
- Works on both Unix and Windows platforms

## Test plan
- [x] Verified compilation on Linux
- [ ] Test manually: `vw --no_stdin -f dir/subdir/model.vw --quiet` should work without pre-creating `dir/subdir/`

Fixes #2686

### DIFF
--- a/vowpalwabbit/io/src/io_adapter.cc
+++ b/vowpalwabbit/io/src/io_adapter.cc
@@ -12,6 +12,7 @@
 #  define ssize_t int64_t
 #  include <io.h>
 #  include <winsock2.h>
+#  include <direct.h>
 #else
 #  include <sys/socket.h>
 #  include <unistd.h>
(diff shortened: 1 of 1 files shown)


JohnLangford:
Closing in favor of #4798 which has the same fix plus Windows compatibility (S_ISDIR macro).

PR #4807, opened 2026-01-24
test: add unit tests for spanning_tree build_tree function

JohnLangford:
## Summary
- Add comprehensive unit tests for the `build_tree` algorithm in the spanning_tree module
- Move `build_tree` function to `VW::spanning_tree_internal` namespace to make it testable
- Tests cover single node through 100 node trees, offset handling, and tree connectivity

This addresses the 0% code coverage in the spanning_tree module identified via Codecov.

## Test plan
- [x] All 9 new tests pass locally
- [ ] CI passes

PR #4810, opened 2026-01-24
test: add unit tests for spanning_tree public interface

JohnLangford:
## Summary
Add unit tests for the `spanning_tree` module that test through the public interface without modifying source code.

Closes the approach from #4807 which required modifying source code to expose internal functions.

### Tests Added (7 tests)
- Construction with default port (26543)
- Construction with port 0 (OS-assigned ephemeral port)  
- `bound_port()` returns correct value
- Quiet mode construction
- Non-quiet mode construction
- Multiple instances with different ports
- Destructor cleanup without `start()`

### Note
The `start()`/`stop()` lifecycle methods are not tested here because they have a race condition that causes flaky tests when `stop()` is called quickly after `start()`. The core functionality (socket binding, port allocation) is tested through the constructor.

## Test plan
- [x] All 7 tests pass locally
- [ ] CI passes

### DIFF
--- a/vowpalwabbit/spanning_tree/CMakeLists.txt
+++ b/vowpalwabbit/spanning_tree/CMakeLists.txt
@@ -17,3 +17,9 @@ vw_add_library(
 if(WIN32)
   target_link_libraries(vw_spanning_tree PRIVATE wsock32 ws2_32)
 endif()
+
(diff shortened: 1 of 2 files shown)


JohnLangford:
Superseded by #4811 (combined PR)

PR #4834, opened 2026-01-28
fix: remove redundant predict call in cb_explore_adf_cover learn path

### LINKED ISSUE #4274: VW calls predict twice during learn for cb_explore_adf

JohnLangford:
## Summary
- Fixes the inverted condition in `cb_explore_adf_cover::predict_or_learn_impl` that caused predict to be called twice during learn
- When `base.learn_returns_prediction` is true, cb_adf's learn functions (e.g., `learn_mtr<true>`) already include an internal predict call, making cover's explicit predict redundant
- When `base.learn_returns_prediction` is false, cover was NOT calling predict, leaving predictions unpopulated

The fix flips the condition from `if (base.learn_returns_prediction)` to `if (!base.learn_returns_prediction)`.

Fixes #4274

## Test plan
- [ ] Verify existing cover-related tests pass (e2e tests in core.vwtest.json)
- [ ] Verify no regression in cb_explore_adf_cover output predictions
- [ ] Compare model outputs before/after to ensure algorithmic equivalence


JohnLangford:
Superseded by #4838 (combined PR)

PR #4835, opened 2026-01-28
fix: prevent --cover and --large_action_space from being used together

### LINKED ISSUE #4254: LAS + cover is failing on C# predict call
### Describe the bug

There is nondeterministic AccessViolation exception during C# Predict call with LAS and cover exploration turned on (I was not able to reproduce it after removing any of LAS/cover/C# property, but do not have intuition why it matters).

### How to reproduce

```
static void CbTest()
{
    string b64_model = null;
    using (
        var vw =
        new VW.VowpalWabbit($"--cb_explore_adf --large_action_space --max_actions 2 --cover 5 --full_predictions"))
    {
        var test = new List<string>();
        test.Add("shared | s_1");
        for (int i = 0; i < 200; ++i)
        {
            test.Add("| a_" + i.ToString());
        }
        using (var coll = new VW.VowpalWabbitMultiLineExampleCollection(vw, null, test.Select(x => vw.ParseLine(x)).ToArray()))
        {
            var prediction = coll.Predict(VW.VowpalWabbitPredictionType.ActionProbabilities);
            double sum = prediction.Select(x => x.Score).Sum();
            Console.WriteLine(sum);
        }
    }
}
```
From time to time (but quite often), I am able to get this:
image

### Version

9.5

### OS

Windows

### Language

C#

### Additional context

_No response_

JohnLangford:
## Summary
- Adds a validation check to prevent `--cover` and `--large_action_space` from being combined
- The cover algorithm indexes internal arrays using action IDs directly, but LAS filters the prediction set to non-contiguous action IDs, causing out-of-bounds buffer access
- This was causing nondeterministic `AccessViolationException` crashes on C# predict calls

Fixes #4254

## Test plan
- [ ] Verify that combining `--cover` and `--large_action_space` now produces a clear error message
- [ ] Verify existing cover tests pass without LAS
- [ ] Verify existing LAS tests pass without cover

### DIFF
--- a/vowpalwabbit/core/tests/cb_explore_adf_test.cc
+++ b/vowpalwabbit/core/tests/cb_explore_adf_test.cc
@@ -16,3 +16,10 @@ TEST(CbExploreAdf, ShouldThrowEmptyMultiExample)
   // An empty example collection is invalid and so should throw.
   EXPECT_THROW(vw->learn(example_collection), VW::vw_exception);
 }
+
+TEST(CbExploreAdf, CoverAndLargeActionSpaceIncompatible)
+{
+  EXPECT_THROW(VW::initialize(vwtest::make_args(
+                   "--cb_explore_adf", "--cover", "3", "--large_action_space", "--max_actions", "10", "--quiet")),
+      VW::vw_exception);
+}


JohnLangford:
Superseded by #4838 (combined PR)

PR #4935, opened 2026-08-17
ci: least-privilege token permissions and no credential persistence

JohnLangford:
Last of three PRs clearing the zizmor code-scanning alerts. **Stacked on #4934** (which is stacked on #4932) — the diff shown here is only this PR's changes, and GitHub retargets it as the others merge.

Additive only: 116 inserted lines, 0 deleted.

## `excessive-permissions` — 55 findings

Twenty of the twenty-one workflows declared no `permissions:` block, so every job ran with whatever the repository default grants — historically read-write across contents, issues, pull requests, packages and more. A compromised action, or a command injection in any build step, inherited all of it.

Each now declares `permissions: contents: read` at the top level, placed where `zizmor.yml` already puts it (after `concurrency:`). The jobs that genuinely need more keep their existing job-level grants, which override the default:

| job | grant | why |
|---|---|---|
| `codeql-analysis.analyze` | `security-events: write` | uploads SARIF |
| `zizmor.zizmor` | `security-events: write` | uploads SARIF |
| `run_benchmarks.benchmark` | `contents: write` | pushes `gh-pages` |
| `dotnet_nugets.benchmark_nuget` | `contents: write` | pushes `gh-pages` |

`check_pr_title.check` gets a **new** job-level grant of `pull-requests: read` — the action reads the PR and, with `validateSingleCommit`, its commit list. It checks nothing out, so it doesn't need `contents`.

I checked everything else that could touch the API with `GITHUB_TOKEN` before tightening:

- both `gh-pages` pushes use an explicit tokenized remote URL, and both `github-action-benchmark` steps run `auto-push: false`, so `contents: write` covers the alert comments they post
- every `actions/github-script` step in the tree only calls `core.exportVariable` — no API calls
- the docs deploy and the coverage upload authenticate with their own secrets (`DOCS_DEPLOY_TOKEN`, `CODECOV_TOKEN`), so `GITHUB_TOKEN` scope is irrelevant to them
- `java-publish.yml` has no `mvn deploy` / registry push — it builds and uploads artifacts

## `artipacked` — 48 findings

`actions/checkout` leaves the token in `.git/config` as an `http.extraheader` by default. It stays there for the rest of the job, so any later step — or an artifact that happens to sweep up `.git` — can read it. All 48 checkouts that didn't say otherwise now set `persist-credentials: false`.

Nothing here depended on the persisted credential:

- **submodules are public** — and the one relative URL, `ext_libs/vcpkg = ../../microsoft/vcpkg.git`, resolves to `github.com/microsoft/vcpkg`
- the `git fetch --unshallow --tags` calls in the nuget workflows read a public repository
- the two `git push` calls carry their token in the remote URL, not in git config

The one checkout that *does* need it — the `VowpalWabbit/docs` deploy in `python_checks.yml`, which pushes with `DOCS_DEPLOY_TOKEN` — already declared `persist-credentials: true` and is untouched.

Verified structurally after the edit by re-parsing every workflow: all 50 checkout steps declare `persist-credentials`, exactly one is `true`, and no non-checkout step picked the key up by accident.

## Verification

```
zizmor .github/workflows/    104 -> 1 finding
```

Across the three PRs: **314 → 1**. All 21 workflows still parse as YAML; CRLF preserved in `java-publish.yml` and `python_checks.yml`.

The single remaining finding is the `adhoc-packages` note explained in #4934 — unpinned `pip install` in the docs job, which can't be meaningfully fixed while VW's `requirements.txt` files are themselves unpinned.

## What to watch on merge

The permission tightening is the only part of this series that can break a workflow at runtime rather than at parse time, and it can only do so for a job that reaches the API in a way I didn't find. If something does fail, the fix is a job-level grant rather than a revert — the failure will name the missing scope.

### DIFF
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -20,6 +20,9 @@ env:
   NEEDRESTART_MODE: a
   VCPKG_BINARY_SOURCES: 'clear;x-gha,readwrite'
 
+permissions:
+  contents: read
+
 jobs:
   test_with_sanitizers:
     strategy:
@@ -42,6 +45,7 @@ jobs:
     steps:
       - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
         with:
+          persist-credentials: false
           submodules: false
           fetch-depth: 0
 
@@ -167,6 +171,7 @@ jobs:
     steps:
       - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
         with:
+          persist-credentials: false
           submodules: false
           fetch-depth: 0
 
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -13,6 +13,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.sha }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   build_macos:
     runs-on: ${{matrix.os}}
@@ -24,6 +27,7 @@ jobs:
     steps:
       - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
         with:
+          persist-credentials: false
           submodules: false
 
       - name: Init submodules with retry
--- a/.github/workflows/build_vw_slim.yml
+++ b/.github/workflows/build_vw_slim.yml
@@ -19,6 +19,9 @@ env:
   DEBIAN_FRONTEND: noninteractive
   NEEDRESTART_MODE: a
 
+permissions:
+  contents: read
+
 jobs:
   check:
     runs-on: ubuntu-latest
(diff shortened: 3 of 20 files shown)


JohnLangford:
Rebased onto the current master (via #4932) and retargeted: the base is now `jcl/vw-zizmor-hardening` rather than `jcl/vw-zizmor-semantic`, since #4934 has merged and that branch is gone.

The permissions and `persist-credentials` changes were regenerated on the new base rather than textually merged, because #4933 added top-level `env:` blocks in exactly the spot the `permissions:` blocks go. The result now reads `on:` → `concurrency:` → `env:` → `permissions:` → `jobs:`, and master's #4933 additions (21 `timeout-minutes: 90`, 12 `NEEDRESTART_MODE` blocks) are untouched.

Re-verified after the rebase:

- `zizmor .github/workflows/` → **1** finding (the `adhoc-packages` note explained in #4934)
- all 50 checkout steps declare `persist-credentials`; exactly one is `true` (the `VowpalWabbit/docs` deploy that pushes with `DOCS_DEPLOY_TOKEN`); no non-checkout step picked the key up
- every write-needing job kept its grant — `codeql-analysis.analyze` and `zizmor.zizmor` (`security-events: write`), `run_benchmarks.benchmark` and `dotnet_nugets.benchmark_nuget` (`contents: write`), plus the new `check_pr_title.check` (`pull-requests: read`)
- all 21 workflows parse; CRLF preserved in `java-publish.yml` and `python_checks.yml`

PR #4936, opened 2026-08-18
fix(ci): retry apt installs so a stalled mirror does not fail the job

JohnLangford:
## Problem

Hosted Linux runners have been stalling inside `sudo apt-get` for hours at a stretch. #4933 capped every job at 90 minutes, so a stall now costs 90 minutes instead of six hours — but **the job still fails**, and re-running is a coin flip while the condition lasts.

Four jobs across #4930/#4931 died at exactly `1h30m` with `conclusion: cancelled`, each killed by that new cap while sitting on an apt call:

| Job | Step it was killed on |
|---|---|
| `benchmark` | `Install build dependencies (Linux only)` |
| `Build python wheel (current)` | `Install build dependencies` |
| `Analyze (csharp, ubuntu-latest)` | `Build CSharp .NET Core` — begins `sudo apt update && sudo apt install` |
| `Test current models with master code` | `Install python` |

## Why the usual mitigations don't apply

**This is a hang, not an error.** `apt-get` never returns. So `-y`, `|| true`, and `Acquire::Retries` can none of them fire — there is no failure for them to act on. Only a wall-clock bound turns the hang into something retryable.

That is also why the previous attempt at this (the env vars in #4933) did nothing: `DEBIAN_FRONTEND` / `NEEDRESTART_MODE` address an interactive prompt, and two jobs then hung on the very commit that set them. Those vars are kept as hygiene; they are not the fix.

## Change

Route every apt install through `.github/scripts/apt-install.sh`:

- each attempt bounded with `timeout` (300s update / 600s install)
- up to **3 attempts**, so a stalled attempt costs ~5 min and retries instead of eating the job's whole budget
- index refresh failure is survivable (`|| true`); **install failure is not** — exhausting all three exits non-zero, so a genuine packaging problem still fails loudly rather than silently proceeding without the packages

`24 call sites` across 13 workflows now use it. The single bare `apt-get update` with no install (`python-docs`) gets the same bound inline — that is the step that hung for an hour, and its trailing `|| true` could not help because the process never returned.

Two jobs (`build-wheel-master`, `linux-python-sdist-build-test`) called apt *before* checking out the repo. They now `sparse-checkout` just `.github/scripts` first, rather than growing a second copy of the retry logic.

## Verification

The wrapper was exercised locally against a mock `apt-get` that hangs:

| Scenario | Result |
|---|---|
| hangs twice, succeeds on 3rd | retries past both, **exit 0** |
| hangs on all 3 attempts | **exit 1** — fails loudly, no silent success |
| invoked with no packages | **exit 2** |

Also checked mechanically: all workflows parse, no raw `sudo apt` invocations remain, CRLF/LF endings preserved per file, and **every** `apt-install.sh` call site is preceded by a checkout in its job.

## Scope

CI only. No product code. Net **-10 lines** in the workflows plus one 30-line script.

### DIFF
--- a/.github/scripts/apt-install.sh
+++ b/.github/scripts/apt-install.sh
@@ -7,6 +7,12 @@
 # returns and the job burns its whole timeout. `timeout` converts that hang
 # into a failure, which can then be retried.
 #
+# `timeout` must run *inside* sudo, not outside it. Run as the unprivileged
+# user it cannot signal the root apt-get at all, and without -k it then waits
+# forever for a child that never dies -- which is exactly the hang it was
+# meant to break. As root, with -k, it escalates to SIGKILL and always
+# returns.
+#
 # Usage: apt-install.sh <package> [package...]
 set -euo pipefail
 
@@ -17,9 +23,9 @@ fi
 
 for attempt in 1 2 3; do
   # An index refresh that fails is survivable; the install below is not.
-  timeout 300 sudo -E apt-get update -o Acquire::Retries=3 || true
+  sudo -E timeout -k 30 300 apt-get update -o Acquire::Retries=3 || true
 
-  if timeout 600 sudo -E apt-get install -y -o Acquire::Retries=3 "$@"; then
+  if sudo -E timeout -k 30 600 apt-get install -y -o Acquire::Retries=3 "$@"; then
     exit 0
   fi
 
--- a/.github/scripts/apt-install.sh
+++ b/.github/scripts/apt-install.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+# Install apt packages, tolerating the transient stalls that have been hanging
+# hosted Linux runners.
+#
+# The failure mode this exists for is a hang, not an error: apt blocks
+# indefinitely reaching a package mirror, so a plain `apt-get install` never
+# returns and the job burns its whole timeout. `timeout` converts that hang
+# into a failure, which can then be retried.
+#
+# Usage: apt-install.sh <package> [package...]
+set -euo pipefail
+
+if [[ $# -eq 0 ]]; then
+  echo "apt-install.sh: no packages given" >&2
+  exit 2
+fi
+
+for attempt in 1 2 3; do
+  # An index refresh that fails is survivable; the install below is not.
+  timeout 300 sudo -E apt-get update -o Acquire::Retries=3 || true
+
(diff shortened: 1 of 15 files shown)


JohnLangford:
zizmor's four findings on this PR were mine, from the two `actions/checkout` steps added so the apt retry helper is present in the two jobs that call apt before checking out:

| Alert | Rule | Location |
|---|---|---|
| 440, 442 | `unpinned-uses` (error) | `python_checks.yml`, `python_wheels.yml` |
| 438, 441 | `artipacked` (note) | same two steps |

Fixed in `fc3f1f6f` by matching the pattern this repo already uses in `zizmor.yml` itself — SHA-pinned action plus no credential persistence:

```yaml
- uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
  with:
    sparse-checkout: .github/scripts
    submodules: false
    persist-credentials: false
```

All four alerts are now in `fixed` state and the `zizmor` check passes. The remaining open zizmor alerts on this branch are pre-existing on `master` (unpinned `actions/*` references across the workflows) and are untouched here.

JohnLangford:
### Follow-up: the wrapper had the same bug it was meant to fix

The first version ran `timeout` *outside* `sudo`:

```bash
timeout 300 sudo -E apt-get update ...
```

That cannot work. `timeout` then runs as the unprivileged runner user while apt-get runs as root, so its SIGTERM is rejected (EPERM); and with no `-k` it waits forever for a child that never dies. So the wrapper reproduced the exact hang it was written to break, and the retry loop never iterated.

This PR's own CI caught it: `lint.c++.formatting` sat on the wrapped step from `03:35` to `05:05` and was killed by the 90-minute job cap, with this in the cleanup —

```
Terminate orphan process: pid (1835) (timeout)
```

— the `timeout` process still alive 88 minutes after its 300-second deadline.

Fixed in `f55631b1`: `timeout` now runs **inside** sudo so it is root and can signal apt-get, plus `-k 30` to escalate to SIGKILL when SIGTERM is ignored.

```bash
sudo -E timeout -k 30 600 apt-get install -y -o Acquire::Retries=3 "$@"
```

My original local test did not catch this because the mock `sudo` ran as the same user, so there was no privilege boundary to hit. Re-verified against a child that traps and ignores SIGTERM: the old form never returns; the new form returns `rc=137` once the kill grace expires; and the full wrapper retries past two such hangs to `exit 0`.

